PR #4824, opened 2015-10-27
Improvements to the sbt build

szeiger:
- Set the correct unmanagedResourceDirectories for the compiler and
  library projects. These projects already had includeFilter definitions
  to copy the correct resource files to the target dirs but they were
  meaningless without a source dir.
- Set a target path for ‘dist’ to make stream files go to the proper
  .gitignored dir.
- Set permissions ugo+rx on all generated script and batch files.

PR #4855, opened 2015-11-20
Create a full Scala distribution from sbt

szeiger:
This replicates the results of `ant -Dlocker.skip=1 distpack` with the following differences:
- The POM files produced by the ant build under dists/maven still contain unresolved variables like `@VERSION@`. The ant build resolves them in the actual publishing process. The sbt build generates the correct version numbers directly.
- Scaladoc sets in the ant build are created with the scaladoc tool from “quick”. The sbt build uses the STARR scaladoc.
- The scaladoc sets for scala-actors, scala-compiler, scala-reflect and scalap from the ant build contain documentation for the classes `Null`, `Nothing`, `Any` and `AnyRef`. The sbt build omits those
  classes from all doc sets (except for scala-library, of course) under the assumption that this is a bug in the ant build.

Other idiosyncracies of the ant build like declaring a packaging type of `jar` (instead of `bundle`) for OSGi bundles, or using slightly different module names and descriptions in scaladocs, OSGi bundles and
POMs, are reproduced faithfully for now.

Publishing from the sbt build is done in the usual sbt way. No actual publishing targets are currently defined in the build. In order to test stability of the publishing process, the `publishDists` task can be used
to publish to dists/maven-sbt with the same directory layout and naming conventions employed by the ant build for publishing to dists/maven.

Source and binary JARs are OSGi-enabled. For the source JARs this is done by hand-crafting the appropriate MANIFEST.MF entries (like the ant build does). Binary JARs are packaged with BND, using the same outdated BND version as the ant build for now to get identical results. We do not use sbt-osgi because it depends on a newer version of BND (probably OK to upgrade in the future) and does not allow a crucial bit of configuration that we need: Setting the classpath for BND. In sbt-osgi this is always `fullClasspath in Compile` whereas we really want `products in Compile in packageBin` to keep things simple.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -221,13 +221,19 @@ lazy val compiler = configureAsSubproject(project)
         streams.value.cacheDirectory) ++
       (mappings in Compile in packageBin in LocalProject("interactive")).value ++
       (mappings in Compile in packageBin in LocalProject("scaladoc")).value ++
-      (mappings in Compile in packageBin in LocalProject("repl")).value,
+      (mappings in Compile in packageBin in LocalProject("repl")).value ++
+      (mappings in Compile in packageBin in LocalProject("repl-jline")).value.filter(_._2 != "repl-jline.properties") ++
+      (mappings in Compile in packageBin in LocalProject("repl-jline-embedded")).value,
     includeFilter in unmanagedResources in Compile :=
       "*.tmpl" | "*.xml" | "*.js" | "*.css" | "*.html" | "*.properties" | "*.swf" |
       "*.png" | "*.gif" | "*.gif" | "*.txt",
     scalacOptions in Compile in doc ++= Seq(
       "-doc-root-content", (sourceDirectory in Compile).value + "/rootdoc.txt"
-    )
+    ),
+    // Generate the ScriptEngineFactory service definition. The ant build does this when building
+    // the JAR but sbt has no support for it and it is easier to do as a resource generator:
+    generateServiceProviderResources("javax.script.ScriptEngineFactory" -> "scala.tools.nsc.interpreter.IMain$Factory"),
+    managedResourceDirectories in Compile := Seq((resourceManaged in Compile).value)
   )
   .dependsOn(library, reflect)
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -122,21 +150,97 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ versionProperties
...
+        <developers>
+          <developer>
+            <id>lamp</id>
+            <name>EPFL LAMP</name>
+          </developer>
+          <developer>
+            <id>Typesafe</id>
+            <name>Typesafe, Inc.</name>
+          </developer>
+        </developers>
```
soc, line 178:
```
+          <developer>
+            <id>ScalaContributors</id>
+            <name>Scala Contributors</name>
+          </developer>
```

?

soc:
Amazing! Thanks @szeiger!

build.sbt
SethTisue, line 178: +1

lrytz:
LGTM! there are some details that have to be different in the 2.12 version, could you prepare a separate PR for 2.12.x? then we can skip over these commits when merging 2.11 to 2.12. you could also put a `[nomerge]` in the commit messages here.

build.sbt
szeiger, line 178: There are several metadata issues to clean up. For example, listing the organization as LAMP vs Scala Team, or using different names and descriptions for subprojects in Manifests, POMs and scaladocs. This PR merely brings the sbt up to the standard of the ANT build. IMHO further improvements to the metadata should be done separately, either after removing the ANT build or in both build systems together.

szeiger:
@lrytz This PR depends on #4824 which is not marked as `[nomerge]`. Will that be auto-merged to 2.12 or is it a manual process?

lrytz:
@szeiger I just created a PR that merges, amongst others, #4824 to 2.12.x. See details in the description of #4864, I took note of all conflicts in build.sbt.

szeiger:
OK, great. Updated with `[nomerge]` (the one commit should be mergeable)

build.sbt
lrytz, line 178: sounds reasonable

PR #4856, opened 2015-11-20
Refactor excluded qname test to be more data like

janekdb:
This refactoring extracts data from code into a form which is closer to
configuration data.

This is a step change toward making this configurable.


janekdb:
Review by @VladUreche .

VladUreche:
LGTM, thanks for the refactor Janek!

PR #4857, opened 2015-11-22
Fix List Scaladoc time & space formatting

janekdb:
The Performance section got sucked into a wormhole and popped up in the
example tag. The laws of physics differ in the attributes block resulting
in the loss of the line break between the Time and Space paragraphs.

Fixed by moving the section out of the example tag.


janekdb:
### Before

The performance section is in the example tag and the time and space sections run together.

scaladoc-list-performance-before

### After

scaladoc-list-performance-after

janekdb:
Review by @heathermiller , @dickwall , @som-snytt .

lrytz:
LGTM, thanks a lot!

PR #4846, opened 2015-11-13
Support specialized functions in closure optimizer

lrytz:
Includes the commits of #4837

PR #4837, opened 2015-11-06
Copy propagation, remove unused values (closures!) and local variables

lrytz:
See individual commit messages.

PR #4858, opened 2015-11-23
Eliminate non-escaping boxes, tuples and refs

lrytz:
This PR includes the commits of #4846.

Eliminate boxes, tuples and refs that are created and used within a
single method without escaping. For details on the implementation see
the doc comment in class BoxUnbox.


lrytz:
@retronym this is now ready for review. This PR includes the commits of https://github.com/scala/scala/pull/4837 and https://github.com/scala/scala/pull/4846, so you can review everything here if you like. I addressed the comments you gave in 4837. A community build is running here: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/269/parameters/

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -140,6 +145,14 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
     (result, map, hasSerializableClosureInstantiation)
   }
 
+  def getBoxedUnit: FieldInsnNode = new FieldInsnNode(Opcodes.GETSTATIC, coreBTypes.srBoxedUnitRef.internalName, "UNIT", coreBTypes.srBoxedUnitRef.descriptor)
+
+  private val anonfunAdaptedName = """.*\$anonfun\$\d+\$adapted"""
```
retronym: Precompile the regex here?

```
scala> val r = "foo..".r
r: scala.util.matching.Regex = foo..

scala> r.pattern.matcher("fooab").matches
res9: Boolean = true
```

lrytz: :+1:

lrytz:
Community build status
- ~~Shapeless failed because it has `-Xfatal-warnings` enabled and there's an invocation to `List.dropWhile` that cannot be inlined, producing the `there was 1 inliner warning` summary. The method cannot be inlined because it contains an `INVOKESPECIAL` of a private method (the `loop` method is private, see https://github.com/scala/scala/blob/v2.12.0-M3/src/library/scala/collection/immutable/List.scala#L354-L361). The fix here is to pass `-Yopt-warnings:none`.~~
- ~~The shapeless failure looks a bit more dramatic than it is because the compiler crashes after issuing the error, due to https://issues.scala-lang.org/browse/SI-9592.~~
- ~~Scala.js failed due to a compiler plugin they use to inject classes in JUnit tests. Fix on the way: https://github.com/scala-js/scala-js/pull/2120~~

lrytz:
Community build passed with `-Yopt:l:classpath` (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/277/)

retronym:
Any chance for a unit test for "Run DCE before the closure optimizer (fixes a crash)" ? Getting pretty close to :santa: on this PR otherwise!

lrytz:
I'll give it a try, have to think about it. The bug was discovered in the community build on scala-js. I didn't isolate / minimize it, the stack trace and some mental processing was enough to realize what is the problem.

src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala
```diff
@@ -311,13 +341,298 @@ object BytecodeUtils {
...
+      (0 to 22).map(base + _)
+    } ++ {
+      primitives.map(base + "0$mc" + _ + "$sp") // Function0
+    } ++ {
+      // return type specializations appear first in the name string (alphabetical sorting)
+      for (r <- ijfdzv; a <- ijfd) yield base + "1$mc" + r + a + "$sp" // Function1
+    } ++ {
+      for (r <- ijfdzv; a <- ijd; b <- ijd) yield base + "2$mc" + r + a + b + "$sp" // Function2
+    }
+  }
```
retronym: The string refs to symbols in this file, and to the specialized ones here in particular, are the main thing I think needs a cleanup before merging.

lrytz: ok, i'll look into cleaning this up

lrytz:
> Any chance for a unit test for "Run DCE before the closure optimizer (fixes a crash)" ?

pushed (amendeded last commit,  4eee5e0)

lrytz:
~~@retronym regarding the enumeration of specialized subclasses of `FunctionN` [(here)](https://github.com/scala/scala/blob/4eee5e088eb72ba00aa4b02a03a1bcf27b5d751b/src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala#L613-L628) / `TupleN` [(here)](https://github.com/scala/scala/blob/4eee5e088eb72ba00aa4b02a03a1bcf27b5d751b/src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala#L568-L575): the only related thing I could find in the compiler is the [`specializedClass`](https://github.com/scala/scala/blob/4eee5e088eb72ba00aa4b02a03a1bcf27b5d751b/src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala#L80) map. The problem is that it only contain classes that are referenced by the program being compiled, the map is populated by method [`specializeClass`](https://github.com/scala/scala/blob/4eee5e088eb72ba00aa4b02a03a1bcf27b5d751b/src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala#L539). So if I want to use that functionality I'd have to invoke `specializeClass` for the `FunctionN` / `TupleN` types. Do you think that's the best way?~~

lrytz:
~~a temporary problem: using specialization's infrastructure wouldn't work for [`scala/runtime/java8/JFunction` types](https://github.com/scala/scala/blob/4eee5e088eb72ba00aa4b02a03a1bcf27b5d751b/src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala#L630-L631) (written in Java, manually specialized)~~

lrytz:
@retronym this PR should be ready now, the last 4 commits are new

retronym:
`MULTIANEWARRAY` should also be retained for potential side effect, no?

Otherwise, this is ready to go!

retronym:
I think the `-YoptTrace` option ought to accept an argument that lets one limit the output to a particular class / method pair. As it stands, it is only really useful after you've extracted the method of interest to a standalone test case. Feel free to do this in a followup PR, though.

retronym:
Here's a script to post-process the optimizer trace into a series of diffs (via a temp git repo). I've used your test case from 1a9e649 as the example.

https://gist.github.com/retronym/0efa5cea463aced52fae

Here's another example from the central commit in this PR: https://gist.github.com/retronym/99a026883916abf2d309

retronym:
After looking at those diffs, I notice that `eliminateStaleStores` doesn't reduce `maxLocals`. This happens subsequently in `dce`, but I was wondering whether this is something that could/should be done more eagerly.

lrytz:
> the -YoptTrace option ought to accept an argument that lets one limit the output to a particular class / method pair

Added in 3dccf0e

> notice that eliminateStaleStores doesn't reduce maxLocals

This is intentional, see this comment: https://github.com/lrytz/scala/commit/29d772d0655a3a5af79c3d8eca91bb33a82c7194#diff-70147974eed4d5f7faa546e42504ce64R94

retronym:
LGTM

PR #4862, opened 2015-11-24
SI-9567 Fix pattern match on 23+ param, method local case class

retronym:
Typechecking constructor patterns of method local case classes
was only working because of the existence of the unapply method
in the companion, which is used if navigation to the case class
companion object fails.

We now support defintion of, and pattern matching on, case classes
with more than 22 parameters. These have no `unapply` method
in the companion, as we don't have a large enough tuple type to
return. So for such case classes, the fallback that we inadvertently
relied on would no longer save us, and we'd end up with a compile
error advising that the identifier in the constructor pattern was
neither a case class nor an extractor.

This is due to the propensity of `Symbol#companionXxx` to return
`NoSymbol` when in the midst of typechecking. That method should
only be relied upon after typechecking. During typechecking,
`Namers#companionSymbolOf` should be used instead, which looks in the
scopes of enclosing contexts for symbol companionship. That's
what I've done in this commit.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala
+++ b/src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala
@@ -504,14 +504,26 @@ trait MatchTranslation {
        */
       def treeMaker(binder: Symbol, binderKnownNonNull: Boolean, pos: Position): TreeMaker = {
         val paramAccessors = binder.constrParamAccessors
+        val numParams = paramAccessors.length
+        def paramAccessorAt(subPatIndex: Int) = paramAccessors(math.min(subPatIndex, numParams - 1))
         // binders corresponding to mutable fields should be stored (SI-5158, SI-6070)
         // make an exception for classes under the scala package as they should be well-behaved,
         // to optimize matching on List
+        val hasRepeated = paramAccessors.lastOption match {
+          case Some(x) => definitions.isRepeated(x)
+          case _ => false
+        }
         val mutableBinders = (
           if (!binder.info.typeSymbol.hasTransOwner(ScalaPackageClass) &&
-              (paramAccessors exists (_.isMutable)))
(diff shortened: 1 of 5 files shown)


lrytz:
https://github.com/scala/scala/blob/2.12.x/test/files/run/t9003.scala#L68 failed

retronym:
Looks like a latent bug. If we make the case class in that test case top-level, it also fails under 2.11.7

``` scala
case class CaseSequenceTopLevel(as: Int*)

object Test {

    val buffer1 = collection.mutable.Buffer(0, 0)
    CaseSequenceTopLevel(buffer1: _*) match {
      case CaseSequenceTopLevel(_, i) =>
        buffer1(1) = 1
        assert(i == 0, i) // fails in 2.11.7 -optimize
    }
}
```

Whereas with this patch the failures are uniform for top-level and method-owned case classes.

I'll investigate.

retronym:
I can crash patmat with a variation:

``` scala
case class CaseSequenceTopLevel(var x: Any, as: Int*)

object Test {

    val buffer1 = collection.mutable.Buffer(0, 0)
    CaseSequenceTopLevel("", buffer1: _*) match {
      case CaseSequenceTopLevel(_, _, i) =>
        buffer1(1) = 1
        assert(i == 0, i) // fails in 2.11.7 -optimize
    }

    // case class CaseSequence(as: Int*)
    // val buffer2 = collection.mutable.Buffer(0, 0)
    // CaseSequence(buffer2: _*) match {
    //   case CaseSequence(_, i) =>
    //     buffer2(1) = 1
    //     assert(i == 0, i)
    // }
  }
}
```

[This code](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala#L506-L515) doesn't account for repeated patterns.

lrytz:
LGTM

PR #4863, opened 2015-11-24
DRY-er trees in pattern matcher code gen.

retronym:
Rather than building a cascade of if/elses, push additional conditions
into a conjunction in the condition of a single if/else. This is
possible when emitting conditions for the list of arguments of a
pattern.

Here's an example of the improvement to post-pattern matcher trees:

  https://gist.github.com/retronym/0d8f7126157061d72b81

While we could try to rely on the optimizer to coalesce the repeated
else clauses, it seems wasteful to emit the code in that way in
the first place.

Review by @lrytz && @adriaanm


lrytz:
needs an update-check on run/t6288.scala

lrytz:
LGTM, very nice!

retronym:
Checkfile updated.

sjrd:
Why does this PR target 2.12.x and not 2.11.x?

retronym:
I was being conservative.

adriaanm:
Indeed -- nice improvement!

PR #4866, opened 2015-11-25
Clarify source of f function in TailCalls flatMap

CaseyLeask:
When reading this line I was repeatedly confused by where the `f` was coming from.
There is another `f` with an identical signature coming from the function definition for `flatMap`.

It's a small change, but I think it really helps you understand the context of where the `f` is coming from.

### DIFF
--- a/src/library/scala/util/control/TailCalls.scala
+++ b/src/library/scala/util/control/TailCalls.scala
@@ -55,7 +55,7 @@ object TailCalls {
         case Done(a) => Call(() => f(a))
         case c@Call(_) => Cont(c, f)
         // Take advantage of the monad associative law to optimize the size of the required stack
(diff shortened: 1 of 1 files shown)


SethTisue:
LGTM assuming Jenkins signs off

soc:
Minor nitpick:

I think it would be more consistent to do

```
c.f(x).flatMap(f)
```

instead of

```
c.f(x) flatMap f
```

CaseyLeask:
While I agree with you, I was trying to keep in the original style as much as possible.

CaseyLeask:
@soc Should I change it?

lrytz:
re-running a build only for that triggers my conscience, given the paris climate summit starting next week :)

SethTisue:
there is infix syntax throughout the file, so I think Casey's choice here made sense

PR #4868, opened 2015-11-26
SI-9542 Fix regression in value classes (served two ways)

retronym:
```
% git log -3 --oneline --graph
*   b1c522a Merge branch 'ticket/9542' into ticket/9542-combo
|\
| * d09f024 SI-9542 Fix regression in value classes with enclosing refs
* | e7980e4 SI-9542 Unify different reprs. of module type refs
|/
```

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/ExtensionMethods.scala
+++ b/src/compiler/scala/tools/nsc/transform/ExtensionMethods.scala
@@ -244,7 +244,10 @@ abstract class ExtensionMethods extends Transform with TypingTransformers {
 
           // These three lines are assembling Foo.bar$extension[T1, T2, ...]($this)
           // which leaves the actual argument application for extensionCall.
-          val sel        = Select(gen.mkAttributedRef(companion), extensionMeth)
+          // SI-9542 We form the selection here from the thisType of the companion's owner. This is motivated
+          //         by the test case, and is a valid way to construct the reference because we know that this
+          //         method is also enclosed by that owner.
+          val sel        = Select(gen.mkAttributedRef(companion.owner.thisType, companion), extensionMeth)
           val targs      = origTpeParams map (_.tpeHK)
           val callPrefix = gen.mkMethodCall(sel, targs, This(origThis) :: Nil)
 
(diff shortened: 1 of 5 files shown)


retronym:
Review by @adriaanm

src/compiler/scala/tools/nsc/transform/ExtensionMethods.scala
```diff
@@ -244,7 +244,7 @@ abstract class ExtensionMethods extends Transform with TypingTransformers {
 
           // These three lines are assembling Foo.bar$extension[T1, T2, ...]($this)
           // which leaves the actual argument application for extensionCall.
-          val sel        = Select(gen.mkAttributedRef(companion), extensionMeth)
+          val sel        = Select(gen.mkAttributedRef(companion.owner.thisType, companion), extensionMeth)
```
adriaanm, line 250: (based on hangout-review:) explain why this prefix is known to be valid (nesting is known by construction)

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -3967,14 +3967,15 @@ trait Types
...
-  def normalizePlus(tp: Type) = (
+  def normalizePlus(tp: Type): Type = {
     if (isRawType(tp)) rawToExistential(tp)
     else tp.normalize match {
-      // Unify the two representations of module classes
-      case st @ SingleType(_, sym) if sym.isModule => st.underlying.normalize
-      case _                                       => tp.normalize
+      // Unify the representations of module classes
+      case st@SingleType(_, sym) if sym.isModule => st.underlying.normalize
+      case st@ThisType(sym) if sym.isModuleClass && sym.isStatic => normalizePlus(st.underlying)
```
adriaanm: document why `isStatic`?

retronym:
@adriaanm I've removed the `isStatic` condition (I can't remember why it was there, and will see if the test suite helps jog my memory). I've also added some more internal documentation to justify the use of `thistype` in `ExtensionMethods`.

I've also removed the merge commit (originally added to show that the either of the commits served as a fix.)

adriaanm:
LGTM

PR #4869, opened 2015-11-26
Apply some static code analysis recommendations

janekdb:
Fix a batch of code inspection recommendations generated by IntelliJ 14.1.5.

Categories of fix,
- Unnecessary public modifier in interface
- Replace filter+size with count
- Replace filter+nonEmpty with exists
- Replace filter+headOption with find
- Replace `if (x != null) Some(x) else None` with `Option(x)`
- Replace `getOrElse null` with `orNull`
- Drop redundant semicolons
- Replace anon fun with PF
- Replace anon fun with method

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/JavapClass.scala
+++ b/src/repl/scala/tools/nsc/interpreter/JavapClass.scala
@@ -600,11 +600,11 @@ object JavapClass {
     def parents: List[ClassLoader] = parentsOf(loader)
     /* all file locations */
     def locations = {
-      def alldirs = parents flatMap (_ match {
+      def alldirs = parents flatMap {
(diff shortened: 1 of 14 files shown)


src/repl/scala/tools/nsc/interpreter/JavapClass.scala
```diff
@@ -600,11 +600,11 @@ object JavapClass {
     def parents: List[ClassLoader] = parentsOf(loader)
     /* all file locations */
     def locations = {
-      def alldirs = parents flatMap (_ match {
+      def alldirs = parents flatMap {
```
janekdb, line 603: Not entirely sure about this change. Why would the previous version have been written in such a baroque fashion when there is an obviously simpler approach?

lrytz:
LGTM

PR #4870, opened 2015-11-27
SI-9507 Make ArrayStack an IndexedSeqOptimized

ruippeixotog:
Just like `ArraySeq`, `ArrayBuffer` and all other collections that use an array as underlying data structure, `ArrayStack` should also be an instance of `IndexedSeq` and `IndexedSeqOptimized`. As expected by both of the traits, `ArrayStack` has constant-time random element access and length computation.

### DIFF
--- a/src/library/scala/collection/mutable/ArrayStack.scala
+++ b/src/library/scala/collection/mutable/ArrayStack.scala
@@ -64,9 +64,10 @@ object ArrayStack extends SeqFactory[ArrayStack] {
 class ArrayStack[T] private(private var table : Array[AnyRef],
                             private var index : Int)
 extends AbstractSeq[T]
-   with Seq[T]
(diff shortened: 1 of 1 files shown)


ruippeixotog:
Review by @Ichoran

szeiger:
Does https://issues.scala-lang.org/browse/SI-9605 apply here? Could this cause the same problem?

ruippeixotog:
In this case, the collection changed in this PR is now an `IndexedSeq`, and therefore the current `search` code will be fast after this is merged. Before this it would have the slow behavior - and rightly so, as there was no "evidence" that an `ArrayStack` had fast random access.

I'll submit a PR with a fix for the bug you mentioned, it seems easy to do.

Ichoran:
LGTM

PR #4822, opened 2015-10-27
SI-9178 Don't eta expand to an Function0-like SAM expected type

retronym:
Otherwise, we can end up with a subtle source incompatibility with
the pre-SAM regime. Arguably we should phase out eta expansion to
Function0 as well, but I'll leave that for another day.

Review by @adriaanm

PR #4871, opened 2015-11-27
[WIP] Turn on SAM by default

retronym:
Initial work to change settings and test by
Svyatoslav Ilinskiy ilinskiy.sv@gmail.com (Thanks!)

To avoid cycles during overload resolution (which showed up
during bootstrapping), and to improve performance, I've guarded
the detection of SAM types in `isCompatible` to cases when the
LHS is potentially compatible.

The changes to `isCompatible` might need some discussion and refinement, but I'm publishing this PR so we can we how close we are to flipping this switch.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -267,7 +267,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
       else {
         // Phase travel necessary. For example, nullary methods (getter of an abstract val) get an
         // empty parameter list in later phases and would therefore be picked as SAM.
-        val samSym = exitingPickler(definitions.findSam(classSym.tpe))
+        val samSym = exitingPickler(definitions.samOf(classSym.tpe))
         if (samSym == NoSymbol) None
         else Some(samSym.javaSimpleName.toString + methodSymToDescriptor(samSym))
       }
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -295,7 +295,7 @@ trait Infer extends Checkable {
         && !isByNameParamType(tp)
         && isCompatible(tp, dropByName(pt))
       )
-      def isCompatibleSam(tp: Type, pt: Type): Boolean = {
+      def isCompatibleSam(tp: Type, pt: Type): Boolean = (definitions.isFunctionType(tp) || tp.isInstanceOf[MethodType] || tp.isInstanceOf[PolyType]) &&  {
         val samFun = typer.samToFunctionType(pt)
         (samFun ne NoType) && isCompatible(tp, samFun)
       }
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -796,9 +796,7 @@ trait Definitions extends api.StandardDefinitions {
      * The class defining the method is a supertype of `tp` that
      * has a public no-arg primary constructor.
      */
-    def samOf(tp: Type): Symbol = if (!settings.Xexperimental) NoSymbol else findSam(tp)
-
-    def findSam(tp: Type): Symbol = {
+    def samOf(tp: Type): Symbol = {
       // if tp has a constructor, it must be public and must not take any arguments
       // (not even an implicit argument list -- to keep it simple for now)
       val tpSym  = tp.typeSymbol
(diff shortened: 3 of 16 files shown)


retronym:
See #4822 for previous work and discussion.

retronym:
Community build scheduled to start soon at https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/250

ilinum:
Great, thanks!

retronym:
The community build failure seems in `scodec-bits` seems to come to the new failure to typecheck:

```
class C {
  def foo(f: Int => Unit) = ()
  def foo(i: X) = ()

  foo(x => ())
}
trait X { def apply(a: Int)}
```

Before, the function literal argument would pre-filter the overloaded alternatives to a single candidate (based on the "shape type" rule in overload resolution). This would provide the expected type to drive inference of the lambda parameter type.

However, with SAM enabled, both variants are survive the preselection, so we are forced to typecheck the argument without an expected type, which leads to a "missing parameter type" failure.

ilinum:
So in this case we pre-filter results and get two types:
- `Int => Unit`
- `X`

So if there's already a good option: `Int => Unit`, why do we need to even try `X`?
What I'm saying is that if we have a function type to match our expression, why do we even look for SAM?

ilinum:
Does this happen because during shape check we have no access to types? Do we need to check both so the following works?

```
class C {
  def foo(f: Int => Unit) = ()
  def foo(i: X) = ()

  foo(x => x.charAt(0))
}
trait X { def apply(a: String)}
```

EDIT: Actually, the example I provided also doesn't work

retronym:
> So if there's already a good option: Int => Unit, why do we need to even try X?

Shape type filtering occurs before we typecheck the argument, so all we really know is the arity of function literal. At that point, we don't know whether the argument will be compatible with a formal parameter type, even if it is a function type of the right arity.

retronym:
While we ponder whether or not to tweak the interaction between inference, overloading, and SAMs, I've submitted a PR to `scodec-bits` to work around the problem. That will let us run more of the community build: https://github.com/scodec/scodec-bits/pull/44

retronym:
Scheduled a community build using my fork of scodec-bits: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/251/

ilinum:
Right, so my argument still stands in a way. 

Why can't we filter SAM out if there's a function literal with compatible arity?
As far as I understand, this is the case that is affected by that:

```
class C {
  def foo(f: String => Unit) = ()
  def foo(i: X) = ()

  foo((x: String) => x.charAt(0)) //doesn't compile
  foo(x => x.charAt(0)) //also doesn't compile
}
trait X { def apply(a: String)}
```

Currently this code doesn't compile but it will simply resolve to the non-SAM version of `foo`
Is it that bad? Am I missing any other consequences?

ilinum:
@retronym any progress on this issue?

adriaanm:
Your example is a nice stress test, but it also complicates things because it throws value discarding in the mix. You're passing in a `String => Char`, expecting a `String => Unit`. Overloading resolution does not take value discarding into account (so far that never mattered, since you can't have two Function1 overloads, as they erase to the same type). When you remove that from the example, it works as expected:

```
scala> :paste

trait X { def apply(a: String)}

def foo(f: String => Unit) = println("F1")
def foo(i: X) = println("X")

defined trait X
foo: (f: String => Unit)Unit <and> (i: X)Unit
foo: (f: String => Unit)Unit <and> (i: X)Unit

scala> foo((x: String) => ()) 
F1

scala> foo(x => ())
<console>:14: error: missing parameter type
       foo(x => ())
           ^
```

adriaanm:
BTW, I'm fleshing out the spec in https://github.com/scala/scala/compare/2.11.x...adriaanm:t8429?expand=1

ilinum:
All right I see!

So do you have a solution for this? What's your idea of fixing it?

adriaanm:
Sorry, which part? I'm reluctant to change overloading resolution further, so I'm not planning to add value-discarding in there. Plenty of other things (e.g., type inference) stop working when you add an overload, that's the best we can do, sadly.

ilinum:
I was asking about that problem:

```
class C {
  def foo(f: Int => Unit) = ()
  def foo(i: X) = ()

  foo(x => ())
}
trait X { def apply(a: Int)}
```

If you leave it as is right now, it will introduce source incompatibility from 2.11 to 2.12.

adriaanm:
I think we'll have to deal with the breakage. If scodec is the only project in the community build to break, I'd say we should do it. I prefer the regularity of the status quo over an adhoc solution for source compatibility in this corner case.

milessabin:
Have you pinged @mpilquist about this?

adriaanm:
Yes, Jason sent a PR: https://github.com/scodec/scodec-bits/pull/44

adriaanm:
Following up with a new PR later today.

adriaanm:
Moved to #4945

PR #4874, opened 2015-11-27
Upgrade JarJar in the ant build

szeiger:
This switches the JarJar dependency over to the Pantsbuild fork, which
supports Java 8 bytecode and is already used by the sbt build. Shading
of the embedded JLine is currently broken in 2.12 with the old version
(the shaded classes are not created).


lrytz:
LGTM

PR #4875, opened 2015-11-27
[SI-9454] Add isAlive method into sys.process.Process trait

slothspot:
review @SethTisue.
proposed patch for https://issues.scala-lang.org/browse/SI-9454

### DIFF
--- a/src/library/scala/sys/process/Process.scala
+++ b/src/library/scala/sys/process/Process.scala
@@ -26,11 +26,11 @@ import scala.language.implicitConversions
  *  make it possible for one to block until the process exits and get the exit value,
  *  or destroy the process altogether.
  *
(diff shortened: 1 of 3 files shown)


SethTisue:
thanks for the PR. saw your tweets too. returning from a US holiday, swamped for now, will look before too long.

SethTisue:
I did some manual testing and this seems to work fine.

but this comment in Process.scala should be removed:

``` text
*  Presently, one cannot poll the `Process` to see if it has finished.
```

slothspot:
@SethTisue Updated and rebased

SethTisue:
@Ichoran can you take a quick look?

lrytz:
LGTM

PR #4880, opened 2015-12-11
WIP: constant folding preserves tree, tracks constant in type

adriaanm:
This is motivated by scaladoc, presentation & incremental compilers all needing the original tree. There's no reason for the type checker to replace the tree -- this can wait until erasure. 

NOTE: this is a WIP to see what breaks and start the discusion --  no decision has been made about whether this can go into 2.11.x. It has to be a smooth, low-risk, transition.

### DIFF
--- a/src/compiler/scala/reflect/reify/codegen/GenTrees.scala
+++ b/src/compiler/scala/reflect/reify/codegen/GenTrees.scala
@@ -67,7 +67,7 @@ trait GenTrees {
     case global.EmptyTree             => reifyMirrorObject(EmptyTree)
     case global.noSelfType            => mirrorSelect(nme.noSelfType)
     case global.pendingSuperCall      => mirrorSelect(nme.pendingSuperCall)
-    case Literal(const @ Constant(_)) => mirrorCall(nme.Literal, reifyProduct(const))
+    case treeInfo.LiteralLike(const @ Constant(_)) => mirrorCall(nme.Literal, reifyProduct(const))
     case Import(expr, selectors)      => mirrorCall(nme.Import, reify(expr), mkList(selectors map reifyProduct))
     case _                            => reifyProduct(tree)
   }
--- a/src/compiler/scala/reflect/reify/codegen/GenTrees.scala
+++ b/src/compiler/scala/reflect/reify/codegen/GenTrees.scala
@@ -67,7 +67,7 @@ trait GenTrees {
     case global.EmptyTree             => reifyMirrorObject(EmptyTree)
     case global.noSelfType            => mirrorSelect(nme.noSelfType)
     case global.pendingSuperCall      => mirrorSelect(nme.pendingSuperCall)
-    case treeInfo.LiteralLike(const @ Constant(_)) => mirrorCall(nme.Literal, reifyProduct(const))
+    case Literal(const @ Constant(_)) => mirrorCall(nme.Literal, reifyProduct(const))
     case Import(expr, selectors)      => mirrorCall(nme.Import, reify(expr), mkList(selectors map reifyProduct))
     case _                            => reifyProduct(tree)
   }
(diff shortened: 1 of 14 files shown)


adriaanm:
@retronym points to:
- https://github.com/scala/scala/compare/2.11.x...retronym:topic/const-type-fold
- https://issues.scala-lang.org/browse/SI-9532
- https://issues.scala-lang.org/browse/SI-8384
- https://issues.scala-lang.org/browse/SI-9477
- https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaStructureBuilder.scala#L45

TODO: 
- [ ] incorporate tests for these
- [ ] check IDE

sjrd:
Note-to-self: this might break our assertions that arguments to `@JSExport` and `@JSName` annotations are literals ...
Ping @gzm0

gzm0:
Hmmm... Basically means that `@JSExport(a)` won't work anymore... The community build should trigger on this.

adriaanm:
I'll take a look at Scala-js. Migration should be simple: look for a ConstantType instead of a literal. The LiteralLike extractor introduced in tree info should do the trick.

adriaanm:
For context, we'll only merge to 2.11 if the migration is really simple. Would be great to fix scaladoc, IDE & incremental compilation handling of constants thought.

adriaanm:
> - [X] could you avoid a separate traversal and bake it into some existing transformation in ersaure?

I haven't found a good spot to hook this up to, but I'm still looking :-) In principle, constant folding can only affect Ident/Select/Apply.

gzm0:
> For context, we'll only merge to 2.11 if the migration is really simple.

Whether this goes into 2.11 doesn't matter that much to Scala.js. We'll probably have to add the `LiteralLike` extractor to our compat layer (since we use a single codebase for all Scala versions we support). Then it will probably just work when you patch the 2.11 branch.

gkossakowski:
More references:
https://issues.scala-lang.org/browse/SI-7173
https://github.com/sbt/sbt/issues/1543
https://github.com/sbt/sbt/blob/1.0.x/sbt/src/sbt-test/source-dependencies/constants/pending

adriaanm:
Maybe this will come back in time for 2.12.0-M5. Closing for now

odersky:
In dotty we keep the original tree in Typer, but replace with literals in the next phase after it.

adriaanm:
Interesting -- why not defer until erasure?

This PR defers constant folding until uncurry. I tried postponing until erasure, but for practical reasons for the current pipeline, it needs to happen before explicitouter, which shouldn't see a reference to a final val that will be constant folded away (or it will make the val non-private). Similar reasoning applies to tailcalls: constants don't influence code flow.

odersky:
The main reason not to defer further is Dotty's transformer architecture. Tree transformers automatically recompute types, but instead of using a full Typer they use a TypeAssigner which is much smaller and more reliable. TypeAssigners don't know how to do constant folding. So we risk changing (say) some part of a tree with a constant type in an inconsequential way (say, we duplicate it), and the type of the resulting tree would be no longer constant. By converting trees with constant types to literals early we protect against that. On the other hand, we definitely want to keep the original trees in Typer, for the reasons you mention.

adriaanm:
Thanks for the explanation!

retronym:
One reason we want to keep the terms in place so that a later compiler phase (in this case, SBT's dependency extraction phase) can see the original trees. By default, that phase in SBT runs right after `pickler.`

smarter:
Right, I'm working on adding an sbt phase to dotty and I don't see why it should be right after typer (right now our trees do not even always pass `-Ycheck` after typer, so it's excluded from `-Ycheck`) so I'm running it after pickler to imitate scalac. I made a patch to preserve `TypeDefs`: https://github.com/smarter/dotty/commit/d3742c8a4234ae4bad080b782612926d9ed67cc5 but ideally we should preserve constant-folded trees too somehow.

adriaanm:
it runs after picklers so it sees the same symbol info when compiling from source as when loading from classfiles -- this is important e.g., to avoid spurious incremental recompilation when mixing source files and class files

PR #4887, opened 2015-12-20
Allow @migration to be used outside of the scala package

janekdb:
Various Scaladoc guides list @migration as a useful tag however it only
become of use outside of the Standard Standard Library when public.

### DIFF
--- a/src/library/scala/annotation/migration.scala
+++ b/src/library/scala/annotation/migration.scala
@@ -13,7 +13,7 @@ package scala.annotation
  * between versions.  This is intended for methods which for one
  * reason or another retain the same name and type signature,
  * but some aspect of their behavior is different.  An illustrative
- * examples is Stack.iterator, which reversed from LIFO to FIFO
+ * example is Stack.iterator, which reversed from LIFO to FIFO
  * order between Scala 2.7 and 2.8.
  *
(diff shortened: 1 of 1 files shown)


sjrd:
How would that be used by another library? If I upgrade from Foo 1.1 to Foo 1.2, what compiler flag should I specify to get warnings relevant to that migration? `scalac -Xmigration:1.1` would not do what I want, now, would it?

janekdb:
Good point @sjrd. I will have to delete this PR as it does not account for this. From a documentary point of view allowing wider use of @migration would be useful as this example shows.

scaladoc-migration-example

soc:
I think it would be interesting to think about a way to divorce it from the versioning concerns, just like deprecated does (maybe just remove the ability to pass a version explicitly?).

soc:
Or maybe the version info could be made relative to a name? E. g. `@migration("Foo changed", "foolib-2.0")` and then `-Xmigration:foolib-2.0`, or `-Xmigration:foolib` for all warnings?

sjrd:
@soc Yes, I would love to have a per-library way to specify migration warnings.

janekdb:
SBT syntax allows a particular version of a library to be identified on a single line, i.e. CLI friendly, and already an adopted syntax.

PR #4877, opened 2015-12-07
add new look for scaladoc

felixmulder:
I'm working on a new look for scaladoc. My goal has been to make something similar to the beautiful Typesafe sites. For now this has been for my own amusement, but if there is interest - I'd like to make this good enough to replace the current (in my opinion, somewhat outdated) look.

screen shot 2015-12-08 at 22 42 33
screen shot 2015-12-08 at 22 42 35
screen shot 2015-12-08 at 22 42 37
screen shot 2015-12-08 at 22 42 41
screen shot 2015-12-08 at 22 42 45
screen shot 2015-12-08 at 22 42 48

There's still a lot of stuff to do:
- ~~Replace large "C"-, "O"- and "T"-icons with css based images.~~
- ~~Add hide index component (left portion)~~
- Filter with: ordering, inherited, implicitly, visibility (right portion)
- ~~Add the HiDPI arrows to "Linear Supertypes/Subtypes"~~

To name a few...

One of my goals is complete removal of the static images that are not HiDPI and replacing them with utf8 symbols or CSS.

If there is interest - I'll keep working on my branch then squish for merge.

PR #4888, opened 2015-12-24
fix mobile styling for scaladoc

felixmulder:
I replaced all occurrences of absolute font sizing with relative ones. The resulting changes should be minimal. Please verify by looking at the resulting docs on dropbox:

https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

:santa: :christmas_tree:

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
@@ -47,6 +47,8 @@ abstract class HtmlPage extends Page { thisPage =>
     val html =
       <html>
         <head>
+          <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+          <meta name="viewport" content="width=device-width, initial-scale=1"/>
           <title>{ title }</title>
           <meta name="description" content={ description }/>
           <meta name="keywords" content={ keywords }/>
@@ -269,7 +271,8 @@ abstract class HtmlPage extends Page { thisPage =>
           <div>
             Related Docs:
             <a href={relativeLinkTo(tpl.companion.get)} title={docEntityKindToCompanionTitle(tpl)}>{objClassTrait}</a>
-            | {templateToHtml(tpl.inTemplate, s"package ${tpl.inTemplate.name}")}
+            <span class="divider">|</span>
+            {templateToHtml(tpl.inTemplate, s"package ${tpl.inTemplate.name}")}
           </div>
         case None =>
           <div>Related Doc:
--- a/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.css
+++ b/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.css
@@ -23,7 +23,6 @@
 
 * {
   color: inherit;
-  font-size: 10pt;
   text-decoration: none;
   font-family: "Lato", Arial, sans-serif;
   border-width: 0px;
@@ -43,10 +42,11 @@ a:hover {
 }
 
 html {
-    box-sizing: border-box;
+  background-color: #364550;
+  box-sizing: border-box;
 }
 *, *:before, *:after {
-    box-sizing: inherit;
+  box-sizing: inherit;
 }
 
 textarea, input { outline: none; }
@@ -56,7 +56,7 @@ textarea, input { outline: none; }
 }
 
 #browser {
-  width: 20em;
+  width: 16.75em;
   top: 0px;
   left: 0;
   bottom: 0px;
@@ -66,7 +66,7 @@ textarea, input { outline: none; }
 }
 
 #browser.full-screen {
-  left: -18em;
+  left: -15em;
 }
 
 #filter {
@@ -80,7 +80,7 @@ textarea, input { outline: none; }
   border-left:0;
   border-right:0;
   z-index: 99;
-  min-height: 6.1em;
+  min-height: 5.5em;
 }
 
 #filter.scrolled {
@@ -110,16 +110,33 @@ textarea, input { outline: none; }
 
 #filter > span.toggle-sidebar {
   position: absolute;
-  top: 0.66em;
-  right: 0.4em;
-  font-size: 1.2em;
+  top: 0.8em;
+  right: 0.2em;
   color: #fff;
   z-index: 99;
+  width: 1.5em;
+  height: 1.5em;
 }
 
 #filter > span.toggle-sidebar:hover {
   cursor: pointer;
-  color: #c2c2c2;
+}
+
+/* Pseudo element replacing UTF8-symbol "Trigram From Heaven" */
+#filter > span.toggle-sidebar:before {
+  position: absolute;
+  top: -0.45em;
+  left: 0.45em;
+  content: "";
+  display: block;
+  width: 0.7em;
+  -webkit-box-shadow: 0 0.8em 0 1px #fff, 0 1.1em 0 1px #fff, 0 1.4em 0 1px #fff;
+  box-shadow: 0 0.8em 0 1px #fff, 0 1.1em 0 1px #fff, 0 1.4em 0 1px #fff;
+}
+
+#filter > span.toggle-sidebar:hover:before {
+  -webkit-box-shadow: 0 0.8em 0 1px #c2c2c2, 0 1.1em 0 1px #c2c2c2, 0 1.4em 0 1px #c2c2c2;
+  box-shadow: 0 0.8em 0 1px #c2c2c2, 0 1.1em 0 1px #c2c2c2, 0 1.4em 0 1px #c2c2c2;
 }
 
 #textfilter > .pre {
@@ -136,7 +153,7 @@ textarea, input { outline: none; }
   display: block;
   height: 20px;
   margin-top: 0.5em;
-  margin-bottom: 1em;
+  margin-bottom: 0.8em;
 }
 
 #textfilter > .input {
@@ -144,7 +161,7 @@ textarea, input { outline: none; }
   position: absolute;
   top: 0;
   left: 0;
-  padding: 0.2em 2em 0.2em 0.5em;
+  padding: 0.2em 1.8em 0.2em 0.5em;
   width: 100%;
 }
 
@@ -180,6 +197,7 @@ textarea, input { outline: none; }
 
 #textfilter > .input > input {
   font-family: "Open Sans";
+  font-size: 0.85em;
   height: 2em;
   padding: 0 0.5em;
   color: #fff;
@@ -195,8 +213,9 @@ textarea, input { outline: none; }
 #textfilter > .clear {
   display: none;
   position: absolute;
-  top: 0.58em;
-  right: 1.7em;
+  font-size: 0.9em;
+  top: 0.53em;
+  right: 1.8em;
   height: 23px;
   width: 21px;
   color: rgba(255, 255, 255, 0.4);
@@ -208,21 +227,23 @@ textarea, input { outline: none; }
 }
 
 #focusfilter {
+  font-size: 0.9em;
   position: relative;
   text-align: center;
   display: block;
   padding: 0.6em;
   background-color: #f16665;
   color: #fff;
+  margin-top: 0.65em;
 }
 
 #focusfilter .focuscoll {
   font-weight: bold;
 }
 
-#focusfilter img {
-  bottom: -2px;
-  position: relative;
+#focusfilter a.focusremove {
+  margin-left: 0.2em;
+  font-size: 0.9em;
 }
 
 #kindfilter-container {
@@ -231,7 +252,7 @@ textarea, input { outline: none; }
   z-index: 99;
   bottom: 0.5em;
   left: 0;
-  width: 20em;
+  width: 16.75em;
 }
 
 #kindfilter {
@@ -248,6 +269,7 @@ textarea, input { outline: none; }
   -ms-user-select: none;
   user-select: none;
   color: #fff;
+  font-size: 0.8em;
 }
 
 #kindfilter:hover {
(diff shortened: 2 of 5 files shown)


felixmulder:
cc: @VladUreche

VladUreche:
Thanks @felixmulder!

Do you have any idea why the page looks like this on my phone? Maybe I'm doing something wrong...

screenshot_2015-12-24-22-18-57

VladUreche:
So the problems I see are:
- disproportionately large font for the main text
- fixed size for the signatures (aka `abstract def`)
- too small font for the method/operator name
- too large font for the arguments and description

felixmulder:
screenshot_20151225-012758
screenshot_20151225-012803

Make sure it looks the same for you. This should preferably be tested on multiple devices. I've only managed to test this on a Nexus 5 (res: 1080x1920)

I added some media queries for making the right hand side a bit more responsive. For instance, if the screen is small enough - it will adjust the left margin of the member definition (this can be seen in a normal desktop browser).

VladUreche:
Amazing!!! In the name of everyone who ever tried browsing Scaladoc from a mobile, a huge THANK YOU! :+1:

There must be something wrong on my side, but I don't see the triple lines to open the side bar:

screenshot_2015-12-25-02-26-23

VladUreche:
Just noticed another minor issue: I can't read the page description, since it's hidden behind the title bar (this is the top of the page, if I try to pull down the page, it refreshes):

screenshot_2015-12-25-02-32-32

VladUreche:
Btw, ping @heathermiller: you'll be thrilled :)

felixmulder:
What phone/tablet are you using, @VladUreche?

VladUreche:
Sony Xperia Z1 Compact with Android 4.4.4 and the browser is Chrome 46.0.2490.76.

felixmulder:
Hmm, would have thought it was something bigger. I'll adjust the media queries to include rules for your device.

Going to Italy next week, this might lie in wait after tonight.

VladUreche:
Thanks for all the great work you've done! And have a wonderful holiday!
PS: Passing by Lausanne on your way? Let's grab a beer!

felixmulder:
Allright, I've relaxed the media-queries a bit. If you feel like experimenting - see line 1048 in `template.css` and downward. (If I haven't caught your phone/tablet, that's the two places to add their dimensions.)

Original plan was actually to go via Geneva, but now we'll be going through Turin instead. Rain-check Vlad! :smile:

EDIT: if your phone is caught by the media queries the then the text should be visible. Regarding the triple lines, it's simply a UTF8-font. But perhaps the default font on your system is missing that glyph. I'll check if it's possible to style it differently.

felixmulder:
So I think I managed to fix the missing trigram (tripple line menu thingy) by adding some black magic CSS. Should work on all browsers, tested it on my droid and friend's iPhone 6.

VladUreche:
Awesome! It works well on my phone now. Here's one last snag i caught while browsing: the annotations (e.g. `@throws@`, `@author`, `@see` ...) force one more indent, which for small resolutions is too much. It would be better if for mobiles the text would start on the next line, with no indentation. WDYT?

screenshot_2015-12-26-10-19-35

felixmulder:
No wifi at our apartment. Will fix it when I'm back next Saturday. Cheerios!

VladUreche:
Thanks! Have a wonderful vacation!

felixmulder:
@VladUreche: I've supposedly fixed the indentation of the annotations you mentioned above. This commit should also fix the scroll-bar appearing in code snippets on:

https://dl.dropboxusercontent.com/u/358427/scaladoc/reflect/index.html#scala.reflect.api.Universe

This screenshot is what the resulting changes look like on my Nexus 5:

screenshot_20160103-224116

As an added bonus, I added some space between the entity icons as per /u/predef's suggestion in the reddit thread, I'll get to the rest of his feedback in the week to come. Since it is not mobile related, I'll make a separate PR.

The rebased commit message has the full list of changes.

VladUreche:
Wohoooooooo! That's awesome, thanks @felixmulder!

VladUreche:
@felixmulder, I just noticed one more issue, still related to those pesky signatures at the top of the page:

screenshot_2016-01-04-00-02-09

Can you make the entire "abstract class Universe extends ..." scroll up with the rest of the page? Right now it's hiding some of the documentation and it's stuck at the top, rendering only 40% of the screen estate usable :(

VladUreche:
Oh, and by the way, it's covering the documentation again, I can't scroll to the lines above "link symbol" -- if I try, chrome refreshes the page.

VladUreche:
Found another snag, for long class names it's moving the title to the next line (which is over the signature):

screenshot_2016-01-04-02-03-32

@felixmulder, it's up to you if you want to still address these. The patch is a clear win over what's out there atm, so when you're tired of fixing snags, just tell me and I'll merge as-is (and we file the rest as bugs).

felixmulder:
I'll change the behavior of the entity definition to scroll with the rest of the view on mobile.

Since these bugs are so obvious that you keep finding them by browsing the docs I'll keep fixing them. That should be the fastest solution. Once the frequency of new issues drops or we're past RC1, filing them in bug reports might be the most transparent course of action. What do you think?

VladUreche:
@felixmulder That sounds amazing! Thanks!

felixmulder:
Your issues should now be fixed. In the latest commit:
- Dynamic offset for the entity signature 
- Resized header fonts for mobile browsers
- Removed "EXPERIMENTAL" button from mobile view in shortcomment divs - they're still there if the member is expanded or navigated to

felixmulder:
Hmm, a lot of pressure on the jenkins server? I wonder if I can force it to rebuild

felixmulder:
@VladUreche - feels like this PR is complete. Next up I'm going to fix the style of your inheritance graphs (separate PR) :smile:

VladUreche:
Awesome, thanks @felixmulder!

VladUreche:
Looking forward to the newly styled diagrams!

Btw, here's a tiny snag, but it may also be my phone -- I can't scroll to see the rightmost code in the code boxes. Can you try on your nexus?

screenshot_2016-01-05-12-01-05

felixmulder:
Hmm, I think there's something wrong with your phone @VladUreche, you better give me your credit card information :wink: 

It is working fine in Firefox on Android, but not Chrome as you said. I'll see if I can find a solution for it and sneak it into the next commit.

I've been thinking of doing something about search as well. One idea would be as someone else mentioned in #4877 to add a little drop-down menu when entering a search string in the top left filter box saying something like "Press enter to search for members". Upon pressing enter the relevant results would be displayed in an overlay on the right pane.

This is along the lines of what @heathermiller suggested regarding search. I'll need to generate a more thorough index - which sounds a lot more exciting than GUI-related things anyway :smile:

If you (both) have any suggestions on what you'd like it to contain I'm all ears!

VladUreche:
Thanks @felixmulder! Full member search would be an awesome addition! If you need any help on the technical side, let's chat directly on skype or hangouts.

heathermiller:
I wouldn't do a full member search in the way you suggest. It would confuse how the down key and enter key is already used to move down to select filtered entities. I tried to explain what I think would work better in the earlier PR in a weak hand-wavy way, but I don't think I was clear ;) What I think would be more intuitive would be:
- search results broken down into two sections: entities and members
- clear headers separating each section of results
- entities: should look like they look now
- members: should look like they look in the index. Take note of members like `result` which are implemented on _many_ classes. http://scala-lang.org/api/current/#index.index-r 

Do you see what I mean?

heathermiller:
RE:

> members: should look like they look in the index. Take note of members like result which are implemented on many classes. http://scala-lang.org/api/current/#index.index-r

Perhaps to reduce the amount of cognitive noise, we could have a clickable ellipsis that unfolds the full list of entities that that member is implemented on for search results that take up too much space...

felixmulder:
All right! I think I'm starting to grasp how you'd like it @heathermiller. Albeit, I'm still somewhat confused :sweat_smile: bear with me!

Do you want to keep the incremental filtering of top-level entities? If so - I reckon there are two use cases for the top left input:
1. Incremental filtering of the top-level entities (real time, same as now)
2. If the user presses enter after entering a search string, the full member search is performed and results are listed in the right pane

Or! Do you want to kill the incremental filtering? In that case, only number 2 above would apply.

Either way - the part that is difficult is presenting the full member search with intuitive results. I've read through your ideas, and I'm not sure how it should all hang together. For the sake of progress, I've created a mockup of results that I think are intuitive. It might not be exactly what you specified, but here it is:

screen shot 2016-01-06 at 09 31 18

As you can see it displays in order by package - then by entity (trait/class followed by object). It highlights the search string in the results. On a technical note, I quite like this ordering. It would allow for the search to be performed in an incremental way - i.e. allowing us to show results as they are found.

It also sort of eliminates the need for the A-Z index - but it could of course be kept if it's a popular feature or if it feels like a "quicker" way of getting from A to B.

If you have a better idea / improvements - would you mind drawing them? (Or illustrating them in some other way)

I'm sorry for any misunderstanding, gonna get my :coffee: now - let's create something great :)

VladUreche:
@felixmulder Thanks for the mockup, it looks like something I wanted to have in scaladoc for a very long time.

For the  entities, keep in mind that there may be several levels of nesting. For example, you have method `typeSignature` in `(a) Symbol` in `(t) Universe`: http://www.scala-lang.org/api/2.12.0-M3/scala-reflect/index.html#scala.reflect.api.Symbols$Symbol@typeSignature:Symbols.this.Type.

I guess there are two ways of showing this:
(1) use the package as top-level entity and several levels of nesting
(2) use `Universe` as the top-level entity and a single level of nesting

What do you think, which one would be easier to implement? @heathermiller wdyt?

felixmulder:
@heathermiller will probably have a better understanding of the entity indexing than I do. But for the sake of offering an outside perspective I'll say this:

I don't believe the results should be nested further than `package => entity => member`.

`Symbol` could be displayed as a member of `scala.reflect.api.Symbols`. `Universe` would also have it as a member.

Will having entities like `Symbol` stand on their own result in too many top-level entities? Otherwise that could also be done. We (or I) have time on our hands to try these things out luckily :smile:

VladUreche:
The thing is you want to show it in `Universe` and not in `Symbols` or `SymbolsApi`, otherwise the users won't know how to get their hands on a symbol.

heathermiller:
Hey @felixmulder! I totally understand the struggle of displaying something meaningful. If you didn't understand my suggestion, no worries – it's my fault for not making it clear, not yours :)

Here's a picture of what I mean:
fullsizerender 2

In the above image, there are two "sections" of the search results displayed in the following order:
- classes/traits/objects
- members

Under classes/traits/objects, I'd do the filtering in the sidebar exactly as we do it now. Order by package, and list in alphabetical order within the package. For the member section, I'd display as we do in the index – we don't worry about the signatures, we just show that "map" is implemented on the following entities: AbstractMap, etc, etc.

Do you see what I mean?

I was also thinking that we could add some way to toggle member-only searching in case people don't want the entities to show up as well or viceaversa. But I'm not sure how we'd add that without adding more annoying buttons

heathermiller:
PS, and yes, you're right, we could totally drop the index this way! (It's not popular at all, almost no one realizes how to use it!)

felixmulder:
All right, great - I now think I understand your suggestion semi-fully, two questions!
1. Would you want all entities and all members (i.e. the search results) to be displayed in the left pane initially? Or with members only on search?
2. The vertical ellipses you have drawn, are they what you meant by expandable & clickable?

Searching for something like "map" in the current API, yields roughly 60 results in top level entities from ~10 different packages. This would effectively hide the member results, with this approach. Which was one of the reasons behind the mockup I posted earlier.

On the other hand - with the approach I suggested, it has a different behavior compared to the one that is currently there + there'd probably have to be some rewriting of the index generation.

I think the approach you're suggesting is pretty much "all there" already - so it should be pretty simple to implement. As such, we could implement it - and then I could experiment with some ideas and we'll see what we like best in the end.

heathermiller:
> Would you want all entities and all members (i.e. the search results) to be displayed in the left pane initially? Or with members only on search?

Yes. I know it seems a bit weird from a "how do I design this interaction" standpoint. But for users, I think it's most intuitive. I think a lot of novice users likely search for members in the current sidebar without realizing that it only searches entities. I think being able to search for both in the same filter would be a lot more intuitive to new users.

> The vertical ellipses you have drawn, are they what you meant by expandable & clickable?

Good question. In my drawing, I only wanted to highlight that there could be many more entities with that member. What I had in mind for the expansion would be something like:

fullsizerender 3

> Searching for something like "map" in the current API, yields roughly 60 results in top level entities from ~10 different packages. This would effectively hide the member results, with this approach. Which was one of the reasons behind the mockup I posted earlier.

Yep, so this is the reason for the "toggle member-only searching" bit I had posted. I think it would be a good idea to think of a way to toggle between results for cases like these – so toggle "members only", "entities only", or the default which would be both. We could do a simple toggle button that appears on mouseover or something. I haven't worked out exactly how best to let the user interact with this. A keyboard shortcut would be nice, or something similar to what you had earlier suggested where, as someone is typing, a dropdown appears and gives them the choice? Maybe we could even work keyboard shortcuts in that dropdown, e.g., arrow left selects the members-only toggle, arrow right selects entities-only toggle? Just brainstorming :)

felixmulder:
I wonder how many people have tried writing something in the search box and then pressing enter. Google inbox has a pretty nice implementation of search, displaying stuff incrementally and then the full search results on enter. More spitballin' :smile:

heathermiller:
You're probably right about the pressing enter thing – I would bet that people only arrow down and hit enter, if they hit enter at all. In Gmail, I admit that I never hit enter at all from the search bar, I just arrow down to the message I'm looking for, and then hit enter. I seldom go to the full search result. 

Good ideas :)

heathermiller:
Btw, my above drawings etc are only a suggestion. By no means do I want to limit you to that. Just wanted to share what I came up with.

PR #4889, opened 2016-01-01
SI-8289 Scaladoc: Make the order of subclasses deterministic

kzys:
I intentionally used 2.12.x as a base branch since there is a huge improvement on Scaladoc in the branch :)


lrytz:
LGTM

SethTisue:
thank you Kato!

PR #4890, opened 2016-01-03
SI-8641 Fix right shift docs for integer types

linasm:
Docs for >> operation of integer types (from Byte to Long) had a wrong
direction saying that it is bit-shift left.


linasm:
@heathermiller please review

SethTisue:
nice catch! would you mind opening a new PR targeting the 2.11.x branch, so 2.11.8 will get the fix? (from there, the changes will get merged into 2.12.x too)

linasm:
PR for 2.11.x:: https://github.com/scala/scala/pull/4892

PR #4891, opened 2016-01-03
Trim complexity in CommentFactoryBase

janekdb:
1. Replace option handling with library call
2. Remove NumberFormatException catch presumed to be copy/paste legacy
3. It's Scaladoc
4. Parse trailing whitespace regex once instead of per line
5. Use string interpolation where it improves readability
6. Add missed alternative to block grammar production rule
7. Add regression test for tag requirement warnings

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
@@ -48,7 +48,7 @@ trait CommentFactoryBase { this: MemberLookupBase =>
     groupNames0:     Map[String,Body] = Map.empty,
     groupPrio0:      Map[String,Body] = Map.empty
   ) : Comment = new Comment{
-    val body           = if(body0 isDefined) body0.get else Body(Seq.empty)
+    val body           = body0 getOrElse Body(Seq.empty)
     val authors        = authors0
     val see            = see0
     val result         = result0
@@ -83,13 +83,9 @@ trait CommentFactoryBase { this: MemberLookupBase =>
     }
     val groupNames     = groupNames0 flatMap {
       case (group, body) =>
-        try {
-          body match {
-            case Body(List(Paragraph(Chain(List(Summary(Text(name))))))) if (!name.trim.contains("\n")) => List(group -> (name.trim))
-            case _                                                       => List()
-          }
-        } catch {
-          case _: java.lang.NumberFormatException => List()
+        body match {
+          case Body(List(Paragraph(Chain(List(Summary(Text(name))))))) if (!name.trim.contains("\n")) => List(group -> (name.trim))
+          case _                                                       => List()
         }
     }
 
@@ -165,11 +161,11 @@ trait CommentFactoryBase { this: MemberLookupBase =>
   private val SymbolTagRegex =
     new Regex("""\s*@(param|tparam|throws|groupdesc|groupname|groupprio)\s+(\S*)\s*(.*)""")
 
-  /** The start of a scaladoc code block */
+  /** The start of a Scaladoc code block */
   private val CodeBlockStartRegex =
     new Regex("""(.*?)((?:\{\{\{)|(?:\u000E<pre(?: [^>]*)?>\u000E))(.*)""")
 
-  /** The end of a scaladoc code block */
+  /** The end of a Scaladoc code block */
   private val CodeBlockEndRegex =
     new Regex("""(.*?)((?:\}\}\})|(?:\u000E</pre>\u000E))(.*)""")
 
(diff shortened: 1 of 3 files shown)


janekdb:
I need some advice on how to add a test for the modified warning code. The message should be unchanged. This object covers one path that needs testing,

```
$ cat DoubleVersionTag.scala

 /**
  * @version 1.0
  * @version 2.0
  */
object DoubleVersionTag
```

This is the warning,

```
$ ./build/quick/bin/scaladoc DoubleVersionTag.scala 
model contains 2 documentable templates
DoubleVersionTag.scala:2: warning: Only one '@version' tag is allowed
 /**
 ^
one warning found
```

I tried adding this to `./test/scaladoc/run/tag-requirements.scala`and using `$ ant test.scaladoc` but the output was `Done.` and not the expected warning,

```
import scala.tools.nsc.doc.base._
import scala.tools.nsc.doc.model._
import scala.tools.partest.ScaladocModelTest

object Test extends ScaladocModelTest {

  override def code =
    """
        /**
         * @version 1.0
         * @version 2.0
         */
        object Test {}
    """

  def scaladocSettings = ""
  def testModel(root: Package) = ()
}
```

Don't discount the possibility I've overlooked an existing test I could extend or that I am looking in completely the wrong area!

Could some kind soul steer me in the right direction?

janekdb:
Review by @VladUreche, @gourlaysama.

janekdb:
This is the build failure,

```
[WARNING] The requested profile "scala-2.11.x" could not be activated because it does not exist.
[ERROR] The goal you specified requires a project to execute but there is no POM in this directory (/home/jenkins/workspace/scala-2.11.x-integrate-ide/target/scalariform). Please verify you invoked Maven from the correct directory. -> [Help 1]
org.apache.maven.lifecycle.MissingProjectException: The goal you specified requires a project to execute but there is no POM in this directory (/home/jenkins/workspace/scala-2.11.x-integrate-ide/target/scalariform). Please verify you invoked Maven from the correct directory.
```

VladUreche:
@janekdb Thank you for the cleanup, it's always appreciated! LGTM

Regarding your question, I'm a bit surprised you don't get the warning... There [a redirect in `ScaladocModelTest`](https://github.com/scala/scala/blob/2.12.x/src/scaladoc/scala/tools/partest/ScaladocModelTest.scala#L62-L64) which should guarantee that anything printed ends up captured by the test.

Unfortunately atm I just don't have the bandwidth to debug this. Would you mind filing a bug and assigning it to me?

janekdb:
@VladUreche Thanks for the feedback. It transpires that extracting the comments triggers the expected warning.

```
  def testModel(root: Package) = {
    // get the quick access implicit defs in scope (_package(s), _class(es), _trait(s), object(s) _method(s), _value(s))
    import access._

    val base = root._package("scala")._package("test")._package("scaladoc")._package("tagrequirements")

    val TestObject = base._object("TestObject")
    val testObjectComment = extractCommentText(TestObject.comment.get)
    println("hi")
    println(testObjectComment)
    println("bye")
  }
```

Next step is for me to minimise the above code and then we should be good to add the test.

VladUreche:
@janekdb oh, yeah, the text is only parsed when the doc comment is requested. How stupid of me not to think about that... Thanks for debugging that!

VladUreche:
It should even be enough to do `TestObject.comment`, that triggers parsing.

test/scaladoc/run/tag-requirements.check
```diff
@@ -0,0 +1,16 @@
+newSource:3: warning: Only one '@version' tag is allowed
+        /**
+        ^
+newSource:9: warning: Tag '@param' must be followed by a symbol name
+          /**
+          ^
+newSource:9: warning: Tag '@param' is not recognised
```
janekdb, line 7: I think we could raise a later improvement which removes this warning on the basis that the `param` tag has already been warned on. It's not actually unrecognised but unused because of the missing symbol name.

VladUreche, line 7: Agreed.

SethTisue:
/nothingtoseehere

(integrate-ide failure is unrelated to the PR)

src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
```diff
@@ -334,15 +332,16 @@ trait CommentFactoryBase { this: MemberLookupBase =>
...
               case stk: SimpleTagKey if (stk.name == key.name) =>
-                reporter.warning(pos, "Tag '@" + stk.name + "' must be followed by a symbol name")
+                reporter.warning(pos, s"Tag '@${stk.name}' must be followed by a symbol name")
                 None
               case _ => None
             }
           val pairs: Seq[(String, Body)] =
             for (key <- keys) yield {
               val bs = (bodyTags remove key).get
+              // > 0 ?? - JDB
```
lrytz: @janekdb `> 1` seems correct to me

janekdb: I'll remove this comment. Code is correct as-is.

SethTisue:
wow, it is not easy to figure out where the actual failure is in 18MB of log. @sschaef, can you tell what's going on here?

kiritsuku:
Looks like a flaky test - at least I can't see how this PR should affect such a test. I rerun the tests once again. Maybe this time they pass.

SethTisue:
(I see Jenkins is backed up right now, looks like we might need to wait a few hours before this will re-run.)

lrytz:
LGTM, thanks @janekdb!

PR #4893, opened 2016-01-04
SI-9607 Deprecate /: and :\

soc:


povder:
If we're at deprecating symbolic method names, please consider doing something with `ProcessBuilder`, `Source` and `Sink` http://www.scala-lang.org/files/archive/nightly/2.12.x/api/2.12.x/#scala.sys.process.ProcessBuilder 

In my opinion there should at least be option not to use these symbolic method names, in `Sink` for example, there is no such option.

SethTisue:
@povder can't predict for sure in advance, but I would think that a PR adding non-symbolic names in ProcessBuilder would be something there would be consensus on accepting; want to make one? if you want to test the waters first, start a thread on scala-internals

SethTisue:
I'm in favor of this deprecation. @odersky, WDYT?

soc:
@SethTisue Done, thanks!

SethTisue:
we'll discuss at next Scala team meeting

SethTisue:
see scala/bug#9607 for discussion details. a PR with documentation changes instead would be welcome.

PR #4894, opened 2016-01-05
Provide simpler commented example for Future.map

mmynsted:
The `forComprehensionExamples` example is good for `flatMap` but if one does not understand `flatMap` when reading the documentation for `map` it might be confusing to separate out what is important just for `map`.   

Having a simpler example for `map` and the full, combined example for `flatMap` lets one better see how `map` works, but then also see how `map` and `flatMap` might work together in a typical for comprehension.

### DIFF
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -243,7 +243,17 @@ trait Future[+T] extends Awaitable[T] {
    *  this future. If this future is completed with an exception then the new
    *  future will also contain this exception.
    *
-   *  $forComprehensionExamples
+   *  Example:
+   *
+   *  {{{
+   *  val f = Future { "The future" }
+   *  val g = f map { x: String => x + " is now!" }
+   *  }}}
+   *
+   *  Note that a for comprehension involving a `Future` 
+   *  may expand to include a call to `map` and or `flatMap` 
+   *  and `withFilter`.  See [[scala.concurrent.Future#flatMap]] for an example of such a comprehension.
+   *
    *
    *  @tparam S  the type of the returned `Future`
    *  @param f   the function which will be applied to the successful result of this `Future`


src/library/scala/concurrent/Future.scala
```diff
@@ -243,7 +243,12 @@ trait Future[+T] extends Awaitable[T] {
    *  this future. If this future is completed with an exception then the new
    *  future will also contain this exception.
    *
-   *  $forComprehensionExamples
+   *  Example:
+   *
+   *  {{{
+   *  val f = Future { "The future" }
+   *  val g = f map { (x: String) => x + " is now!" }
+   *  }}}
```
He-Pin, line 251: why not 

```
for(str <- Future ("The future" ))  yield str + " is now!"

```

mmynsted, line 251: There is nothing wrong with that, but I wanted the example to be as simple to understand as possible.   By explicitly showing `map`, one can very easily see how `map` is being used in the example.

SethTisue:
I like having a more self-contained example first, but I'm a bit sorry to lose the information that there is a connection between `map` and `for`. (Though admittedly this is just general information about the Scala language...)

Perhaps add something like this?

> Note that a for comprehensions involving a `Future` may expand to include a call to `map` (and/or `flatMap` and `withFilter`). See `flatMap` for an example of such a comprehension.

(appropriately Scaladoc-ified)

mmynsted:
Oh I like that.  :-)

src/library/scala/concurrent/Future.scala
som-snytt, line 251: Sorry to nitpick:

```
val g = f map { x: String => s"$x is now!" }
val g = f map (x => s"$x is now!")
```

Only b/c doc is indoctrination.

mmynsted, line 251: While I agree that using the simple string interpolator would be a better practice I do not think it should be in this example.  I do not want to assume that somebody looking up how to use `map` also knows how to use the interpolator.   

I do like your removal of the extra parentheses though.

SethTisue:
LGTM. can you squash this into a single commit? then we can merge it.

mmynsted:
Should I just close this one and create a new PR with the single commit?  I am not sure what happens if I rebase and squash these...

lrytz:
You can rebase / squash and then `push -f` to your branch, that will update this PR.

mmynsted:
Oh nice!  I did not expect a rebase squash on a PR to work like this.  Very clean, and simple.  :-)

mpociecha:
@mmynsted FYI, there are still two commits.

mmynsted:
Heh, yeah.  I will have to fix that.

mmynsted:
Ok. fixed.

src/library/scala/concurrent/Future.scala
```diff
@@ -243,7 +243,17 @@ trait Future[+T] extends Awaitable[T] {
...
+   *  Example:
+   *
+   *  {{{
+   *  val f = Future { "The future" }
+   *  val g = f map { x: String => x + " is now!" }
+   *  }}}
+   *
+   *  Note that a for comprehension involving a {{{Future}}} 
+   *  may expand to include a call to {{{map}}} and or {{{flatMap}}} 
+   *  and {{{withFilter}}}.  See {{{flatMap}}} for an example of such a comprehension.
```
retronym: Triple `{` are for code blocks, could you please use a single back-ticks to enclose monospace phrases here?

The pointer to the example should also be a link. Here is the syntax for defining a link to a non-overloaded method:

```
/** Example [[scala.concurrent.Future#flatMap]] */
```

mmynsted:
Revised

lrytz:
Sorry to be so late to the party.. But now that there's only one usage of [`forComprehensionExamples`](https://github.com/scala/scala/blob/90ee1b871236402f8543bf424a4f38d909598b3a/src/library/scala/concurrent/Future.scala#L63-L79) left, I think it would be better to just inline it into the [doc comment of `flatMap`](https://github.com/scala/scala/blob/90ee1b871236402f8543bf424a4f38d909598b3a/src/library/scala/concurrent/Future.scala#L259)

retronym:
> But now that there's only one usage of forComprehensionExamples left, I think it would be better to just inline it into the doc comment of flatMap

Let's let this cleanup slide for now so as to get this change in. @mmynsted If you feel like implementing this in a followup PR, please do!

PR #4895, opened 2016-01-05
SI-9564 Make reusing of ArrayBuilder after clear() safe

linasm:
In some cases ArrayBuilder.elems was still pointing to the previously created array after a call to ArrayBuilder.clear(). As a result, reusing such ArrayBuilder would overwrite the contents of previously created array.

### DIFF
--- a/src/library/scala/collection/mutable/ArrayBuilder.scala
+++ b/src/library/scala/collection/mutable/ArrayBuilder.scala
@@ -100,6 +100,7 @@ object ArrayBuilder {
 
     def clear() {
       size = 0
+      resize(0)
     }
 
     def result() = {
@@ -165,6 +166,7 @@ object ArrayBuilder {
 
     def clear() {
       size = 0
+      resize(0)
     }
 
     def result() = {
@@ -230,6 +232,7 @@ object ArrayBuilder {
 
     def clear() {
       size = 0
+      resize(0)
     }
 
     def result() = {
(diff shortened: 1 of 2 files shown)


sjrd:
If that is a bug, then we have much worse bugs in the mutable collections. For example, an `ArrayBuffer`'s `result()` method returns itself (`this`). See https://github.com/scala/scala/blob/73d8306424931acb301c9c00b19d72d31f51037e/src/library/scala/collection/mutable/ArrayBuffer.scala#L177
I think this is true for most, if not all, mutable collections.

I don't think this is an implementation bug. I believe it is a _documentation_ bug. After a call to `Builder.result()`, the _whole_ instance must be considered invalid. Calling `clear()` does not fix it.

SethTisue:
see previous discussion at https://issues.scala-lang.org/browse/SI-9564. I had leaned towards Sébastien's position, otoh I normally defer to @Ichoran on this kind of thing. anyone else have an opinion?

SethTisue:
let's discuss at https://issues.scala-lang.org/browse/SI-9564, since that's where the bulk of the previous discussion is

Ichoran:
@linasm - Thanks for the PR, but I would like to try a more comprehensive approach to taming reusable and non-reusable Builders.  (Also, there was a more efficient way to make array builders reusable.)

https://github.com/scala/scala/pull/4904

linasm:
I'm closing this in favour of a more general solution in #4904

PR #4896, opened 2016-01-07
Use invokedynamic for structural calls, symbol literals, lambda ser.

retronym:
The previous encodings created static fields in the enclosing class
to host caches. However, this isn't an option once emit code in default
methods of interfaces, as Java interfaces don't allow static fields.

Luckily, we can emulate a static field by using invokedynamic: when
the call site is linked, our bootstrap methid can perform one-time
computation, and we can capture the result in the CallSite.

Review by @lrytz. 

I can't remember why I needed to change `visitHandle`, but I know that the ugliness there is a symptom of a somewhat ad-hoc representation of indy calls in our trees. Ideally, we'd be able to come up with something general purpose, so that a macro author could emit an arbitrary indy call, without needing special cases in the backend for each bootstrap method.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -310,6 +310,15 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
         case app : Apply =>
           generatedType = genApply(app, expectedType)
 
+        case app @ ApplyDynamic(qual, Literal(Constant(boostrapMethodRef: Symbol)) :: staticAndDynamicArgs) =>
+          val numStaticArgs = boostrapMethodRef.paramss.head.size - 3 /*JVM provided args*/
+          val (staticArgs, dynamicArgs) = staticAndDynamicArgs.splitAt(numStaticArgs)
+          val boostrapDescriptor = staticHandleFromSymbol(boostrapMethodRef)
+          val bootstrapArgs = staticArgs.map({case t @ Literal(c: Constant) => bootstrapMethodArg(c, t.pos)})
+          val descriptor = methodBTypeFromMethodType(qual.symbol.info, false)
+          genLoadArguments(dynamicArgs, qual.symbol.info.params.map(param => typeToBType(param.info)))
+          mnode.visitInvokeDynamicInsn(qual.symbol.name.encoded, descriptor.descriptor, boostrapDescriptor, bootstrapArgs : _*)
+
         case ApplyDynamic(qual, args) => sys.error("No invokedynamic support yet.")
 
         case This(qual) =>
@@ -1327,7 +1336,7 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
       val samName = sam.name.toString
       val samMethodType = methodBTypeFromSymbol(sam).toASMType
 
-      val flags = LambdaMetafactory.FLAG_SERIALIZABLE | LambdaMetafactory.FLAG_MARKERS
+      val flags = java.lang.invoke.LambdaMetafactory.FLAG_SERIALIZABLE | java.lang.invoke.LambdaMetafactory.FLAG_MARKERS
 
       val ScalaSerializable = classBTypeFromSymbol(definitions.SerializableClass).toASMType
       bc.jmethod.visitInvokeDynamicInsn(samName, invokedType, lambdaMetaFactoryBootstrapHandle,
@@ -1342,16 +1351,4 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
       indyLambdaHosts += cnode.name
     }
   }
-
-  lazy val lambdaMetaFactoryBootstrapHandle =
-    new asm.Handle(asm.Opcodes.H_INVOKESTATIC,
-      coreBTypes.jliLambdaMetafactoryRef.internalName, sn.AltMetafactory.toString,
-      MethodBType(
-        List(
-          coreBTypes.jliMethodHandlesLookupRef,
-          coreBTypes.StringRef,
-          coreBTypes.jliMethodTypeRef,
-          ArrayBType(ObjectRef)),
-        coreBTypes.jliCallSiteRef
-      ).descriptor)
 }
--- a/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
@@ -64,15 +64,13 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
   /**
    * Add:
-   * private static java.util.Map $deserializeLambdaCache$ = null
    * private static Object $deserializeLambda$(SerializedLambda l) {
-   *   var cache = $deserializeLambdaCache$
-   *   if (cache eq null) {
-   *     cache = new java.util.HashMap()
-   *     $deserializeLambdaCache$ = cache
-   *   }
-   *   return scala.runtime.LambdaDeserializer.deserializeLambda(MethodHandles.lookup(), cache, l);
+   *   return indy[scala.runtime.LambdaDeserialize.bootstrap](l)
    * }
+   *
+   * We use invokedynamic here to enable caching within the deserializer without needing to
+   * host a static field in the enclosing class. This allows us to add this method to interfaces
+   * that define lambdas in default methods.
    */
   def addLambdaDeserialize(classNode: ClassNode): Unit = {
     val cw = classNode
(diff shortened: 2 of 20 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1354,4 +1368,14 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
           ArrayBType(ObjectRef)),
         coreBTypes.jliCallSiteRef
       ).descriptor)
+
+  lazy val symbolLiteralBoostrapHandle =
+    new asm.Handle(asm.Opcodes.H_INVOKESTATIC,
+      "scala/runtime/SymbolLiteral", "bootstrap",
+      "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;)Ljava/lang/invoke/CallSite;")
```
lrytz: References to class names should go through a BType, see https://github.com/scala/scala/blob/fb22e2b0a73605d654c153e02d454e5cec21f355/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala#L74-L78

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1354,4 +1368,14 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
+
+  lazy val structuralCallSite_BoostrapHandle =
+    new asm.Handle(asm.Opcodes.H_INVOKESTATIC,
+      "scala/runtime/StructuralCallSite", "bootstrap",
+      "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;)Ljava/lang/invoke/CallSite;")
```
lrytz: ditto (references to class names should go through a BType)

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -54,17 +54,20 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
   class ProdConsAnalyzer(val methodNode: MethodNode, classInternalName: InternalName) extends AsmAnalyzer(methodNode, classInternalName, new Analyzer(new InitialProducerSourceInterpreter)) with ProdConsAnalyzerImpl
 
+  lazy val lambdaDeserializeBoostrapHandle =
+    new scala.tools.asm.Handle(scala.tools.asm.Opcodes.H_INVOKESTATIC,
+      "scala/runtime/LambdaDeserialize", "bootstrap",
+      "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;")
```
lrytz: ditto (references to class names should go through a BType)

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -77,37 +80,14 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
...
-      mv.visitVarInsn(ASTORE, 1)
-      mv.visitVarInsn(ALOAD, 1)
-      mv.visitFieldInsn(PUTSTATIC, classNode.name, "$deserializeLambdaCache$", mapDesc)
-      mv.visitLabel(l0)
-      mv.visitFieldInsn(GETSTATIC, srLambdaDeserializerRef.internalName, "MODULE$", srLambdaDeserializerRef.descriptor)
-      mv.visitMethodInsn(INVOKESTATIC, jliMethodHandlesRef.internalName, "lookup", nilLookupDesc, false)
-      mv.visitVarInsn(ALOAD, 1)
       mv.visitVarInsn(ALOAD, 0)
-      mv.visitMethodInsn(INVOKEVIRTUAL, srLambdaDeserializerRef.internalName, "deserializeLambda", lookupMapSerlamObjDesc, false)
+      mv.visitInvokeDynamicInsn("lambdaDeserialize", "(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object;", lambdaDeserializeBoostrapHandle)
```
lrytz: ditto (references to class names should go through a BType)

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -204,7 +184,8 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
     def visitAnnotationss(annotss: Array[java.util.List[AnnotationNode]]) = if (annotss != null) annotss foreach visitAnnotations
 
     def visitHandle(handle: Handle): Unit = {
-      visitInternalNameOrArrayReference(handle.getOwner)
+      if (handle.getOwner != "scala/runtime/LambdaDeserialize" && handle.getOwner != "scala/runtime/SymbolLiteral")
```
lrytz: why is this filter necessary? they are not nested classes

src/library/scala/runtime/MethodCache.scala
```diff
@@ -10,7 +10,8 @@ package scala
 package runtime
 
 
-import java.lang.reflect.{ Method => JMethod }
+import java.lang.invoke.MethodType
+import java.lang.reflect.{Method => JMethod, InvocationTargetException}
```
lrytz: these changes seem unnecessary

DarkDimius:
> Java interfaces don't allow static fields

Java does allow `public static final` fields in interfaces:

``` java
public interface I {
  static public final Object field = null;
}
```

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
retronym: Just pushed this. (I had done this already in a separate commit on the branch that I'd cherry-picked this PR from, but forgot to include it.)

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
retronym: I'll remove this and see what breaks to jog my memory :)

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -310,6 +310,20 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
+          val reflectiveParamsDescriptor = MethodBType(reflectiveParams, ObjectRef).toASMType.getDescriptor
+          mnode.visitInvokeDynamicInsn(qual.symbol.name.encoded, invokedType.getDescriptor, structuralCallSite_BoostrapHandle, reflectiveParamsDescriptor)
+
+        case app @ ApplyDynamic(qual, Literal(Constant(symname: String)) :: Nil) if qual.symbol == Symbol_apply =>
+          val symbol = app.symbol
+          val returnAsmType = typeToBType(symbol.info.resultType).toASMType
+          val invokedType = asm.Type.getMethodType(returnAsmType)
+          val name = "apply"
+          mnode.visitInvokeDynamicInsn(name, invokedType.getDescriptor, symbolLiteralBoostrapHandle, symname)
+
```
retronym, line 321: This is the ad-hoc part I was referring to; ideally the `ApplyDynamic` tree would have all the info we need to emit the `invokedynamic` instruction.

retronym:
> Java does allow public static final fields in interfaces:

Oh, good to know. I'll keep that in mind as an option (we can simulate a non final field, where needed, with a mutable box).

I think there is an argument that the `invokedynamic` approach still has some merit, as it exposes fewer implementation details as public static fields.

retronym:
I think it is also worthwhile to push forward to support `invokedynamic` generally in the backend, as we can use macros to expose it through to the user level.

sjrd:
Just pointing out that exposing `invokedynamic` at the user level would be a JVM-only feature. I'm pretty sure there's no way to implement such a thing in Scala.js, although I'm not very familiar with the details of `invokedynamic`.

retronym:
> I can't remember why I needed to change visitHandle,

Removing the hack in `visitHandle` led to a test failure in `t6287.scala`.

Turns out that this was related to a problem under multi-run compilation. I'd cached the `asm.Handle` for bootstrap methods in a place in the backend that meant that on a subsequent compilation run, emitting the boostrap method reference didn't populate the `ClassBType` from a compiler `Symbol`. Just before writing the classfile to disk, the backend then attempted to load the class bytes from the compiler classpath instead to determine if it is a nested class. Under the toolbox global, however, we can't (in general) get to classfiles this way, so we hit an assertion failure.

I've pushed an updated version of this PR that moves the lazy vals into the part of the backend that is created for each `Run`.

retronym:
I've pushed another version with the generalization I talked about before. See the new test `test/files/run/indy-via-macro` for an example of a use case that can be solved with this facility.

retronym:
I'll bring this PR back soon.

lrytz:
LGTM!

Very nice work! The generalization is great, and trading `--` in `src/compiler` for `++` in `src/library` is also excellent.

retronym:
Rebased.

SethTisue:
@retronym should we tweak https://github.com/scala/scala/blob/2.12.x/src/library/scala/language.scala#L68-L84 to reflect (no pun intended) this?

SethTisue:
I guess it's fine as is since we still ultimately call the method reflectively.

PR #4898, opened 2016-01-07
SI-8700 Exhaustiveness warning for enums from Java source

soc:
Until now, the warning was only emitted for enums from Java class files.

This commit fixes it by
- aligning the flags set in JavaParsers with the flags set in
  ClassfileParser (which are required by the pattern matcher to
  even consider checking exhaustiveness)
- adding the enum members as childs to the class holding the enum
  as done in ClassfileParser so that the pattern matcher sees the enum
  members when looking for the sealed children of a type

### DIFF
--- a/src/compiler/scala/tools/nsc/javac/JavaParsers.scala
+++ b/src/compiler/scala/tools/nsc/javac/JavaParsers.scala
@@ -800,16 +800,10 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
       val superclazz =
         AppliedTypeTree(javaLangDot(tpnme.Enum), List(enumType))
       val finalFlag = if (enumIsFinal) Flags.FINAL else 0l
-      val abstractFlag = {
-        // javac adds `ACC_ABSTRACT` to enum classes with deferred members
-        val hasAbstractMember = body exists {
-          case d: DefDef => d.mods.isDeferred
-          case _         => false
-        }
-        if (hasAbstractMember) Flags.ABSTRACT else 0l
-      }
       addCompanionObject(consts ::: statics ::: predefs, atPos(pos) {
-        ClassDef(mods | Flags.JAVA_ENUM | finalFlag | abstractFlag, name, List(),
+        // Marking the enum class SEALED | ABSTRACT enables exhaustiveness checking. See also ClassfileParser.
+        // This is a bit of a hack and requires excluding the ABSTRACT flag in the backend, see method javaClassfileFlags.
+        ClassDef(mods | Flags.JAVA_ENUM | Flags.SEALED | Flags.ABSTRACT | finalFlag, name, List(),
                  makeTemplate(superclazz :: interfaces, body))
       })
     }
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
@@ -671,7 +671,7 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
     // Primitives are "abstract final" to prohibit instantiation
     // without having to provide any implementations, but that is an
     // illegal combination of modifiers at the bytecode level so
-    // suppress final if abstract if present.
+    // suppress final if abstract is present.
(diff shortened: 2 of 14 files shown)


test/files/neg/t8700.check
```diff
@@ -0,0 +1,7 @@
+Bar.scala:2: warning: match may not be exhaustive.
```
soc: I have no idea what it emits "2" here ... when I run scalac it correctly returns "4".

soc:
Review by @retronym?

test/files/neg/t8700.check
retronym: That looks like the right line number to me. It's the same as this warning (before your patch):

```
⚡ scalac sandbox/test.scala
sandbox/test.scala:2: warning: match may not be exhaustive.
It would fail on the following input: List(_)
    def baz(foo: List[String]) = foo match {
                                 ^
one warning found
```

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -800,16 +800,8 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
         AppliedTypeTree(javaLangDot(tpnme.Enum), List(enumType))
       val finalFlag = if (enumIsFinal) Flags.FINAL else 0l
-      val abstractFlag = {
-        // javac adds `ACC_ABSTRACT` to enum classes with deferred members
-        val hasAbstractMember = body exists {
-          case d: DefDef => d.mods.isDeferred
-          case _         => false
-        }
-        if (hasAbstractMember) Flags.ABSTRACT else 0l
-      }
```
retronym, line 810: Why remove this logic? @lrytz added this in https://github.com/scala/scala/commit/63812c18b23de60039fd3267e4806449ea679972#diff-226e876495a438b9b5c6d407d5997368R801

test/files/neg/t8700/Foo.java
```diff
@@ -0,0 +1,4 @@
+public enum Foo {
+    A,
+    B
+}
```
retronym: Could add a test with enums of the form `enum Foo { A { public void foo() {} }; public void foo(); }` (related to the `ACC_ABSTRACT` change I asked about above.

test/files/neg/t8700.check
soc: Argh. You are right. I somehow compared with an older file that had a package statement and was confused.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
soc, line 810: The problem is that otherwise the pattern matcher–given  `public enum Foo { A, B }`–suggests that `Foo()` is missing in the exhaustiveness checks.

ClassfileParser does the same here: https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala#L547

test/files/neg/t8700/Foo.java
soc: Done. Pushed. Thanks!

soc:
@lrytz Does this look fine to you?

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
lrytz, line 810: OK, can you add the same comment as in the classfile parser (https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala#L545-L546), or a reference to it?

lrytz:
Looks good otherwise!

lrytz:
Actually, can you check if we have equivalent test for the separate compilation case (when the Java enum is read from the classfile)?

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
soc, line 810: Done!

soc:
@lrytz Done!

lrytz:
LGTM, thank you Simon!

SethTisue:
nice work, thank you!

PR #4204, opened 2014-12-14
Fix problems with a locale-dependent decimal mark in StringContextTest

mpociecha:
This commit corrects three tests which were failing for certain locale
due to the different decimal marks in the expected value and the
result (e.g. 2.50 and 2,50).

From now also the expected value is formatted in accordance with the
current locale.

I reported this issue some time ago (see
https://groups.google.com/forum/#!topic/scala-internals/P7mv3wmTSfs )
but nothing changed so far. That's why I made these changes.

PR #4897, opened 2016-01-07
Make StringContextTest pass on Windows

linasm:
One assertion in StringContextTest."f interpolator baseline"() was relying on platform specific newline encoding, making it fail when run on Windows.

PR #4900, opened 2016-01-10
Fix problems with a locale-dependent decimal mark in StringContextTest

mpociecha:
Depending on the current locale one can get e.g. 3,400000e+00 instead of 3.400000e+00 so the expected result should take this into account.

It's a similar change to one I made some time ago in 2.11.x: https://github.com/scala/scala/pull/4204

### DIFF
--- a/test/junit/scala/StringContextTest.scala
+++ b/test/junit/scala/StringContextTest.scala
@@ -1,6 +1,8 @@
 
 package scala
 
+import java.text.DecimalFormat
+
 import language.implicitConversions
 
 import org.junit.Test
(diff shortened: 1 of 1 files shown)


SethTisue:
LGTM. 2nd reviewer @som-snytt?

som-snytt:
LGTM

In an alternate universe, the fix-up of the expected string is a custom interpolator, so that it's easy to read the test.

```
      f"${3}%e"  -> locally"3.000000e+00",
      f"${3.14}%s,%<f"  -> s"3.14,${ locally"3.140000" }",
```

Similar issue

https://github.com/scala/scala/pull/4897

mpociecha:
@SethTisue @som-snytt Before someone will merge this, look at this version: https://github.com/mpociecha/scala/commit/545c08824be2590a963e6513399894c9abe0a94c

If you think it's better, I'll replace the current version with it. :)

som-snytt:
That's great, I was just about to offer a PR on your branch with that.

mpociecha:
Or such a bit simplified version: https://github.com/mpociecha/scala/commit/e9f7f49eb21a2f212ca4143f994971002332f874

Wihout any verifications that there's a number specified but without regexp. ;) You can choose which version you like.

mpociecha:
@som-snytt Finally I pushed a bit modified version of the above one.

som-snytt:
LGTM also.

Your solution is similar to mine; any comment would be that it isn't exactly mine.

Worth noting that this workaround is necessary because `f""` doesn't take a `Locale` -- there was previous discussion about it using an implicit Locale or `f"3.14"(Locale.FR)`.

Alternatively, require running the test with Locale set to `Locale.US`, the way line ending issues were resolved by requiring git checkout with unix EOL. I guess you can also set `-Duser.language` etc explicitly to force the locale?

So `LC_ALL=en_US.utf8 ./test/partest` and so on, or the `-Duser.X=` in build and partest.

mpociecha:
Just rebased.

SethTisue:
thank you Michał!

PR #4903, opened 2016-01-13
SI-9622 Native method may be private

som-snytt:
This was lost in a refactor.

https://github.com/scala/scala/commit/290f687fb6ab91b6aef62d871036ddc3829f12b4

Mutability and if/else. Classic.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -1592,11 +1592,7 @@ trait Namers extends MethodSynthesis {
       import SymValidateErrors._
       def fail(kind: SymValidateErrors.Value) = SymbolValidationError(sym, kind)
 
-      def checkWithDeferred(flag: Int) {
-        if (sym hasFlag flag)
-          AbstractMemberWithModiferError(sym, flag)
-      }
-      def checkNoConflict(flag1: Int, flag2: Int) {
+      def checkNoConflict(flag1: Int, flag2: Int) = {
         if (sym hasAllFlags flag1.toLong | flag2)
           IllegalModifierCombination(sym, flag1, flag2)
       }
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -1592,11 +1592,7 @@ trait Namers extends MethodSynthesis {
       import SymValidateErrors._
       def fail(kind: SymValidateErrors.Value) = SymbolValidationError(sym, kind)
 
-      def checkWithDeferred(flag: Int) {
-        if (sym hasFlag flag)
-          AbstractMemberWithModiferError(sym, flag)
-      }
-      def checkNoConflict(flag1: Int, flag2: Int) {
+      def checkNoConflict(flag1: Int, flag2: Int) = {
```
SethTisue, line 1595: better with explicit `: Unit` IMO (here and elsewhere in the commit)

som-snytt, line 1595: Apparently `-Ywarn-infer-any` doesn't snag on `if (cond) v` which is `if (cond) v else ()`. You must get Unit or AnyVal or Any.

retronym, line 1595: That option only considers type arguments inferred as `Any` / `AnyVal`, not the types of `if` / `match` / `try` expressions.

retronym:
LGTM

I'm not sure how `-Ywarn-infer-any` is related to the choice not to be explicit with `: Unit =` here. But in any case I think this one is good to merge as is, keeping Seth's suggestion in mind for next time.

som-snytt:
The useful compiler option or abide rule would warn if the value of `def f() = if (c) effect` was used. I wouldn't ask the procedure rewriting tool to add `Unit` on the local def. Similar for the inferred booleans, `def b = x || y`.

retronym:
FTR, a workaround is to provide a dummy method body:

```
class C {
  @native private def sampleMethod(argOne: Long): Int = ???
}
```

```
⚡ scalac sandbox/test.scala && javap -c -private C
Compiled from "test.scala"
public class C {
  private native int sampleMethod(long);

  public C();
    Code:
       0: aload_0
       1: invokespecial #14                 // Method java/lang/Object."<init>":()V
       4: return
}
```

som-snytt:
It would be elegant if the RHS communicated "I am supplied by a native func."  I think unicode ellipsis is punctuation, but something else. `def f = ...`. `def f = _?`. That's not unlike `macro impl`. Sorry to speculate in this forum.

PR #4904, opened 2016-01-15
Clarified and expanded which Builders were reusable

Ichoran:
This additionally fixes both SI-8648 and SI-9564.

Added documentation to Builder to clarify that in general Builders are NOT reusable.

Altered implementation of GrowingBuilder to use Growable instance's clear (not valid for a reusable builder, but this one isn't reusable).

Added a new marker trait ReusableBuilder that specifies that these builders should be reusable.

Made all Array builders Reusable in all cases (by setting capacity to 0 if the original array is returned), and documented that they are reusable.

(Fixed a poor implmentation of Array[Unit] builder along the way.)

Documented which other builders were already reusable (maps, sets, Vector, LazyBuilder, StringBuilder, ListBuffer, etc.).

### DIFF
--- a/src/library/scala/collection/immutable/ListSet.scala
+++ b/src/library/scala/collection/immutable/ListSet.scala
@@ -12,7 +12,7 @@ package immutable
 
 import generic._
 import scala.annotation.tailrec
-import mutable.Builder
+import mutable.{Builder, ReusableBuilder}
 
 /** $factoryInfo
  *  @define Coll immutable.ListSet
@@ -32,8 +32,10 @@ object ListSet extends ImmutableSetFactory[ListSet] {
    *  a time to a list backed set puts the "squared" in N^2.  There is a
    *  temporary space cost, but it's improbable a list backed set could
    *  become large enough for this to matter given its pricy element lookup.
+   *
+   *  This builder is reusable.
    */
-  class ListSetBuilder[Elem](initial: ListSet[Elem]) extends Builder[Elem, ListSet[Elem]] {
+  class ListSetBuilder[Elem](initial: ListSet[Elem]) extends ReusableBuilder[Elem, ListSet[Elem]] {
     def this() = this(empty[Elem])
     protected val elems = (new mutable.ListBuffer[Elem] ++= initial).reverse
     protected val seen  = new mutable.HashSet[Elem] ++= initial
--- a/src/library/scala/collection/mutable/ArrayBuilder.scala
+++ b/src/library/scala/collection/mutable/ArrayBuilder.scala
@@ -18,7 +18,7 @@ import scala.reflect.ClassTag
  *
  *  @tparam T    the type of the elements for the builder.
  */
-abstract class ArrayBuilder[T] extends Builder[T, Array[T]] with Serializable
+abstract class ArrayBuilder[T] extends ReusableBuilder[T, Array[T]] with Serializable
 
 /** A companion object for array builders.
  *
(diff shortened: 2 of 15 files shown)


Ichoran:
Note that if we think adding the `ReusableBuilder` trait is too big of an impact on binary compatibility between 2.12 and 2.11, we can drop that part of the change and the runtime behavior will be unchanged.  (You just won't be able to use pattern matching to understand what kind of builder you have.)

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -45,7 +47,16 @@ object ListSet extends ImmutableSetFactory[ListSet] {
       }
       this
     }
+    /** Discards any existing elements.
+      *
+      * If `result` was called previously, this will reset the builder to build a new instance.
+      */
```
sjrd: The indentation of the `*` is not the same as what I'm used to see in this codebase. Is it intentional? What I usually see is:

``` scala
/** First line.
 *
 *  Second line and rest.
 */
```

src/library/scala/collection/mutable/ArrayBuilder.scala
```diff
@@ -634,65 +726,37 @@ object ArrayBuilder {
...
-    def result() = {
-      if (capacity != 0 && capacity == size) elems
-      else mkArray(size)
-    }
+    /** Returns the newly created Array.
+      *
+      * After calling `result`, the result of all methods of this builder are undefined until `clear` is called.
+      * Following a `clear`, the builder will be ready to build an additional instance of the same kind of Array.
+      */
+    def result() = new Array[Unit](size)
```
sjrd: Technically, this changes the behavior of this builder. It will return an array filled with `null`s, whereas the previous implementation would return an array filled with `BoxedUnit.UNIT`.

sjrd:
Sounds good to me, otherwise.

src/library/scala/collection/mutable/ArrayBuilder.scala
Ichoran: Oops, good catch!  I'd forgotten that.  (Encouraged somewhat by the REPL showing `null` as () when in an array.)

Ichoran:
@sjrd - Thanks for the review.  I think I fixed the problems you uncovered.

sjrd:
Good for me.

src/library/scala/collection/mutable/ArrayBuilder.scala
lrytz: in that spirit (not related to this PR):

```
scala> (new Array[Unit](1))(0) == null
<console>:12: warning: comparing values of types Unit and Null using `==' will always yield false
       (new Array[Unit](1))(0) == null
                               ^
res8: Boolean = true
```

lrytz:
Instead of duplicating the same doc comment over again, could you override methods `clear` and `result` in the `ReusableBuilder` marker trait? I tried with the following example:

```
package p
trait Base {
  /** Not reusable. */
  def foo: Int
}
trait RBase extends Base {
  /** This one is reusable */
  override def foo: Int
}
class A extends Base { def foo = 1 }
class B extends RBase { def foo = 2 }
```

Scaladoc handles it as expected: the doc in `A#foo` says "Not reusable", the one in `B#foo` says "This one is reusable".

Ichoran:
@lrytz - Not all the doc comments are the same (because not all behaviors are the same), but I may be able to reword it so it is not misleading if the default doc comments come through (i.e. someone forgets to replace them).

Ichoran:
@lrytz - I think I got reasonable language for the generic Reusable case so I only had to re-document the unusual ones (e.g. where you can keep building without clearing).

lrytz:
LGTM, thank you!

PR #4777, opened 2015-09-30
Improve implicits wildcard imports in the IDE

retronym:
In 451cab967a, I changed handling of selection of members from
package objects so as to correctly use `somePackage.package.type`
as the prefix, rather than `somePackage`. This fixed generic
substitution for members inherited from superclasses of the
package object.

However, this has subtly changed the scope from which we collect
implicits given a wildcard import. It seems that the IDE gets into
a situation after a scaladoc lookup, which temporarily typechecks
the sources of a package object of a third party library, in which
the members of package object differ from the members of the enclosing
package.

The upshot of this was spurious type errors due to implicit search
discarding an candidate implicit whose symbol is not matched by
typechecking an identifier with the symbol's name at the implicit
usage site (this is how we discard shadowed implicits.)

I'd like to ge to the bottom of this, but in the meantime, I've found
that we can fix the regression by looking up the implicit member
symbols in the package, even while correctly using the package object
as the prefix.

PR #4905, opened 2016-01-16
disable flaky presentation compiler test on Windows

SethTisue:
see https://github.com/scala/scala-dev/issues/72 for details

review by @retronym


retronym:
LGTM

Scaladoc in the presentation compiler is a bit of a nightmare. I looked at this in the context of [another bug a while back](https://github.com/scala/scala/pull/4777#issuecomment-164587560) but didn't get to the bottom of things :cry:

PR #4906, opened 2016-01-16
Rename Stream to LazyList

soc:


soc:
The only issue seems to be a failure in test/files/neg/abstract-report2.scala due to

```
java.lang.NoSuchMethodError: scala.sys.process.ProcessBuilder.lineStream_$bang()Lscala/collection/immutable/Stream;
    at scala.tools.partest.nest.Runner.scala$tools$partest$nest$Runner$$$anonfun$13(Runner.scala:299)
```

which means that partest needs a recompile  (expected).

Commit message will be adapted.

Note that `LazyList` has not been added to `scala.package`.

Discuss!

kmizu:
@soc Why is this pull request needed? Most Scala users already know a `Stream` is a lazy list !

xuwei-k:
I agree with kmizu.

soc:
_Most Scala users already know a Stream is a lazy list!_ assumes that not a single new person will ever adopt Scala in the future.

This change is trading an ongoing pain against a one-time pain. With Java 8 streams, reactive streams, etc. the ongoing pain will only increase in the future. If we save every new person learning Scala only 5 minutes (averaged), this will amount to days or months of time spent on something more useful.

The name is unintuitive, inconsistent with everything else called "stream" on and off the platform, and caused havoc for a long time already.

While "lazy list" doesn't convey every single characteristic of the class, it seemed to be the option most people were happy with in the last debate about this topic.

He-Pin:
I think rename to `LazyList` should be better,now java8 have `Stream` too.
+1 for this.

nafg:
@soc I agree that we need to make things as streamlined as possible for newcomers. (Not sure "time saved aggregated across all users" is a useful metric though :wink:  )

Two points related to this.
1. There is a big issue with ease of entry into scala. I think at this point the problem is not any big hurdle, but that there are countless small hurdles that add up, many of which could be eliminated.
2. The programming scene today is very different than in Scala's younger years. Then, most programmers were used to much more limited languages. Today most of Scala's "innovations" are to a significant extent commonplace. Functional programming has lost its stigma and is now something lots of people are interested in, and lots of technologies promote. The upshot is that we can't attract people just by saying "with scala you can use first class functions" etc. etc. There aren't a lot of things that are so radical that they will take people through a tough learning curve. Scala's benefits are now _somewhat_ more nuanced. And so we need to make it easier for people on the outside to see to the inside.

Another thing, today people are more into trying out languages. ES6 has basically made all javascript programmers do it. People are switching to Go, trying out Elixir and Elm, experimenting with Rust. I think a lot more people are multilingual -- and want to reuse concepts across languages. So it's very helpful to beginners if Scala's terminology is aligned, when appropriate, with analogous concepts in other languages. If someone uses streams in a lot of languages and in Scala it means something completely different, that's a bump that perhaps didn't have to be there.

Hmm -- actually, given say node.js streams, JDK8 streams, and scala.Stream, which is most different? I mean they're all different, but wouldn't you say that JDK8 streams have more in common with scala.Stream than with node.js streams (IIUC the latter are somewhat like akka streams)? They're both collections aren't they? So maybe that defeats the last argument.

Anyway, what's the "prior art" here? Is there precedent for a data structure like scala.Stream being called stream? And do other languages have them, and if so what do they call them?

Ichoran:
Stream memoizes, but the head is eager.  So I would tend to call it MemoizedList (or maybe MemoList) instead, if it was going to change its name.  Scala doesn't have a fully lazy list (i.e. `head` is lazy too).

hepin1989:
On the recently Scala Meetup @BeiJing [#1](https://github.com/scalacn/scala-meetup/blob/master/Beijing/2016-01-17%231.md) we talked about the  learning curve and big hurdle of newcomer of Scala too,On that Meeting ,one guy from IBM said that `Scala is the most difficult PL he have learnt`,and we come up with that scala maybe split to 4 level :
- OO
- FP
- TypeLevel
- Meta/Macro ？？？

Anyway，I think we should try to simplfy the language really to achive a big more community,take a look at Elixir,it have more Star than Scala now.

SethTisue:
@hepin1989 sorry, but this is offtopic for this PR — post to scala-debate instead? let's keep this focused specifically on the PR

odersky:
I think instead of going around in circles here we should invest our energy in producing an automatic rewrite tool. Once we have that, we can do changes like this. But without it, we are just wasting our time with discussions that will lead nowhere.

soc:
@Ichoran That was my original proposal, but wouldn't `List` be a "memoized list", too?

@odersky I'm not sure I understand the relation between the lack of an automated rewrite tool and discussions that will lead nowhere. Isn't that something completely orthogonal? The last time this was discussed, there wasn't much bikeshedding at all.

paulp:
If you like the name LazyList, save it for an actual lazy list.

```
scala> Stream({ println("so lazy") ; 1 })
so lazy
res0: scala.collection.immutable.Stream[Int] = Stream(1, ?)

scala> res0 map { x => println("lazy mapping") ; x + 1 }
lazy mapping
res1: scala.collection.immutable.Stream[Int] = Stream(2, ?)
```

soc:
@paulp I know of the issues with Stream, and "LazyList" is certainly not perfect, but do you have a better name suggestion?

Ichoran:
`LazyTail` is basically what it is.

paulp:
Ten years spent teaching people Stream is eager in the head, and you already want to do it over again under another name? Why? How can this be among anyone's top 100K issues? If you can't fix it, leave it alone.

soc:
My issue is with the name, not the implementation.

Nevertheless, fixing is orthogonal to this–though maybe desirable. (Although I don't want to imagine to fun memory retention issues when Stream accepts functions/by-name args which close over the whole world accidentally.)

SethTisue:
I (rather sadly, since Simon's arguments aren't invalid) agree that the name change is not worth doing all by itself.

PR #4908, opened 2016-01-17
Remove duplicate words across the code base.

grimreaper:
This fixes a few minor duplicate words.

This is my first contribution and a test to make sure I get the process correct.


SethTisue:
You could safely target 2.11.x here instead of 2.12.x. (We periodically merge from 2.11 to 2.12.) Otherwise LGTM.

Unfortunately GitHub doesn't let you retarget an existing PR, so you'd have to make a new PR targeting 2.11.x.

If you can't be bothered, we'll merge this anyway; but if you're willing, a new PR would be welcome.

grimreaper:
Lets see if I did that correctly: https://github.com/scala/scala/pull/4910

PR #4911, opened 2016-01-19
SD-70 Don't share footnotes across multiple calls to universe.showRaw

lrytz:
Before this commit, multiple invocations of universe.showRaw used a
shared weak map that caches footnotes. If the two printed objects
have equal components printed as footnotes, e.g., an equal TypeRef,
the result of the second invocation depends on whether the object
has been collected (and removed from the weak map) or not.
See https://github.com/scala/scala-dev/issues/70#issuecomment-171701671


lrytz:
review by @SethTisue

SethTisue:
LGTM. I'll merge in another day or so if there are no other reviewer comments.

lrytz:
cc @xeno-by

PR #4912, opened 2016-01-21
add new search - featuring entity and member search

felixmulder:
## New search functionality

This commit adds a revamped search function for the scaladoc tool. It
also contains a number of small fixes for HTML-layout and JavaScript
issues.

The search is implemented by enhancing the scheduler and using
JavaScript promises. List of changes/additions:
- Revamped search functionality
  - Search members as well as entities
  - Preserve keyboard navigation
  - Scroll to selected entity if outside of viewport
  - Non-blocking, cancelable
- Display library name (top left)
- Refactored scheduler
- Cleanup of HTML layout
  - Removed left pane
  - Better mobile layout, no need for dynamic offsets
  - Remove unused element classes
## Tested On

Chrome 47.0.2526.106
Firefox 43.0.4
Safari  9.0.2 (11601.3.9)
OS X
## Preview

https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

**Review:** @heathermiller, @VladUreche


lrytz:
Wow, looks beautiful!

My first impression about the revamped search: mixing members and entities into the same result list seems problematic. Try navigating to class `List`: there's a large number of results because all collections have a `toList`. I think we need some way to narrow the search space; random ideas: checkboxes, ideally with keyboard shortcuts, and/or search flags like `c:list` or `m:toList`).

Another idea: it would be really awesome if search terms could be combined. It would be amazing to type "list flatm" and then `List#flatMap` shows up :)

heathermiller:
Hey @felixmulder! I'm very sorry about the delay in getting to this–I'm currently in the US at a meeting with a no laptop policy. I have a lot of thoughts about this–in general I think it's a good first step, but I think that there need to be some changes to the sidebar and some more streamlining needs to be done. I've hesitated in responding because I don't have any good suggestions at the moment. One thing that we need to think about is the old searching functionality. I find it now hard to get the old search functionality back–for example, I might just want to see all entities that are *Map-named classes/traits/objects. Now I have a big list of zillions of members that I have to sort through. So, I worry that longtime users of Scaladoc might be annoyed by this aspect of the change–that is, they go to Scaladoc to quickly search for entities and it's now a noisier search, even though they now have the ability to search members too. I think it's just a matter of figuring out how to visually organize these results for people. 

I think Lukas is making suggestions in the right direction, only problem is that filters with checkboxes etc are kind of a dated UI technique. Maybe we could do ecommerce-style tags? Github-style filters (like in issue searches) could also work, but I find that rather verbose and not so quick... 

Just brainstorming. I hope it helps, and again sorry for the delays.

heathermiller:
About the comment alluding to the sidebar above–the sidebar is now rather useless, no? We don't use it anymore, with the searchbar on top now doing the searching. Maybe we should get rid of it and change the page from fullwidth to a less expansive width? Maybe max like 1140 or something? We could even use a lightweight CSS grid system to keep things responsive.

felixmulder:
Thanks for the feedback @lrytz! And don't sweat it @heathermiller, I know you guys are busy people :smile:

What do you guys think of something like this:

## General search

Start typing and it searches for everything.

## Focused search

Start out by defining type i.e:
- `class:`
- `trait:`
- `def:`

I.e. when you hit a colon, it tries to turn it into a tag - if we keep to reserved words it will make things easier.

So you could type something like:
- `class:` List
- `def:` map
- `val:` Lulz

felixmulder:
Also about the grid system - I like it - sort of. I think to get an overview of packages, some people would still like the left pane. But maybe we could hide it by default? That gets rid of the vis noize

EDIT: something like angular + some CSS framework would probably be nice to have if this is going to keep evolving and becoming more intricate. I've added some object orientation to the scripts, but there is still somewhat of a structure lacking IMO

EDIT2: AND! I do like the idea of 1140 as a width - would be awesome to have a max width and allow centering! :drool:

felixmulder:
@lrytz of course the tags could have synonyms like `#` - only problem is, some members might be named  `#::` like in stream.

heathermiller:
So this is what I meant earlier when I said Github-style searching (such as in issues) is a bit convoluted. I don't think that `class:` etc is the right way. Also, as a user, I might not know if something is a class or a trait before I start searching for it, I might only be ignorantly searching for anything that is some kind of map. Also, no one knows that we (in Scaladoc world) refer to class/trait/objects/package objects "entities". So I don't advocate that approach. It's less user-friendly than it can be.

heathermiller:
I think the focused search (for entities) should be the default. And that member searching should be secondary given that close to 100% of returning Scaladoc users expect an entity-first search.

lrytz:
another possibility to approach it: split up the search results into sections, and show the first few results for each section. see for example https://chrome.google.com/webstore/detail/github-awesome-autocomple/djkfdjpoelphhdclfjhnffmnlnoknfnd?hl=en

> Also, I don't know if something is a class or a trait, I just want an entity that is a map. And no one knows that we (in Scaladoc world) refer to class/trait/objects/package objects "entities". So I don't advocate that approach

agreed!

heathermiller:
About the sidepanel–as it currently is proposed, it doesn't really help provide an overview of packages at all. It gives you 100% of everything. I can't filter anymore. So if I only care about an overview of one package, I have to endlessly scroll to find it in a large library like the stdlib. I don't think the sidebar adds much, I actually think it adds a lot of confusion. If we want to provide an overview of the library, tree-style, we should perhaps do it another way.

Right now, by leaving it where is and not changing how it looks, we're just going to annoy returning Scaladoc users because we left this thing there and took the filtering functionality away. So first thing people are going to do is get annoyed/confused about why it doesn't filter anymore. I think leaving it causes more problems than it solves.

heathermiller:
RE @lrytz:

> another possibility to approach it: split up the search results into sections, and show the first few results for each section. see for example https://chrome.google.com/webstore/detail/github-awesome-autocomple/djkfdjpoelphhdclfjhnffmnlnoknfnd?hl=en

Yes! Something like this is what I tried to suggest a few times in earlier PRs. Like in the screenshots in that link where you have different like sub-sections for "Your Repositories" or "Issues" etc, we should do the same. Sub-section for class/traits/objects, sub-section for members.

felixmulder:
All right - battle plan:
- Scrap the sidebar
- Split search results into panes for members and entities and search for both as a default

As a start - how does that sound?

(Trying to distill this into something I can work on)

felixmulder:
The only issue I have with this really, is how to present the members separately in a good way...I'll get there hopefully though :smile:

heathermiller:
@felixmulder That sounds good!

Also, if we do this all in a dropdown, we can also limit the number of entities or members are shown. And, say, selecting the section heading in the dropdown box "Classes/Traits/Objects" or "Members" takes you to the full lit of found entities or members on the corresponding search page.

Do you get what I mean? (I have no idea how cryptic I sound, haha)

felixmulder:
I get what you mean - but if we do that - would the user expect them to be sorted by relevance instead of lexicographically?

I'll complete the "full search" first and then we can iterate on the dropdown :+1:

felixmulder:
So, I've implemented the changes we decided to start with. One thing that I find interesting with this solution is that the docs feel much more like a homepage/wiki with a nice introduction to the standard library. What I mean is this:

https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html

Less visual noise => I actually read the package description. Hehe... anyway - feedback appreciated.

I have not (yet) implemented the keyboard shortcuts that were previously available - once we're sure that this is the direction we want to take, I'll get on it.

VladUreche:
:+1: here's the test failure:

```
fail - scalacheck/IndexTest.scala  [compilation failed]% scalac IndexTest.scala
IndexTest.scala:77: error: value browser is not a member of scala.tools.nsc.doc.html.page.Index
        (index.browser \ "script").size == 1
               ^
```

felixmulder:
I renamed the member to `search` and also moved the script to the `<head>` tag. That's what's breaking it no doubt - I'll patch it in the morning :smile:

VladUreche:
Nevermind the breakage, the new look is awesome!
I was just pointing it out so you don't have to dig into the jenkins logs.

felixmulder:
Thanks @VladUreche ! The tests passed on my setup now, so we should be good!

Once this is good enough for everyone - I suggest we merge this and add the drop-down proposed by @heathermiller in a separate commit - I'm not a fan of big PRs :smile:

felixmulder:
Or rather - I'm not a fan of the all or nothing situation that pertain to big PRs

VladUreche:
From my side, it's perfectly understandable. But let's see if @heathermiller agrees, as she's in the best position to review the PR.

felixmulder:
With #4949 merged, there is no longer the issue of duplicates in the Array class.

VladUreche:
LGTM, thanks Felix, awesome work! 
Ping @heathermiller and @scala/core: ready to merge?

Preview: https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

VladUreche:
Ping @felixmulder: I just noticed the "Linear supertypes" and "Known subclasses" no longer open in https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.collection.immutable.List. Any idea what's happening?

Yes, I've got the "magic touch" :grinning:

lrytz:
small thing: when pushing a key in the search box but not changing the text (e.g. push and release "shift", or "return") the search results are refreshed.

heathermiller:
Yeah, there's a few smallish things I'd like to see done before we merge this, including keyboard shortcuts :) List coming, gimme a couple minutes to write it down.

heathermiller:
Here goes:
- The search results no longer reveal companions. Sometimes it's useful to navigate directly to object Boolean rather than class Boolean, for example. In the old API docs, this was remedied by having two little icons next to the name of the entity that had a companion; either a little trait and a little object icon, or a little class and a little object icon. Both were clickable. Can this functionality be returned?
- In the old API docs, clicking on the package name brought you to the package object page. How do I get to a package object page now with the new search? It seems I can't any longer. (**blocker**)
- Keyboard shortcut: when in the search results page, can hitting "escape" be set to go back?
- I'm really happy that there's now this `doctitle` thing being displayed in the top left. But it would also be good if we could maybe display a version number as well. Would this be possible? Also, could you refresh my memory on how this needs to be set? This seems like something that should also be added to the build.

PS, do we really need the dropdown? The search result page is really quite nice!

heathermiller:
Actually, I just realized that I can't navigate to companion objects at all currently. Not unless I first go to the class and then click on the "see related docs: object Boolean" link. There should be a way to directly go to companion objects from the search results.

heathermiller:
Also, when clicking on the text in the top left, in this case, "Scala Standard Library", is it possible to make that a link that brings us back to the root package? That text also isn't selectable, btw. Might be good to remedy that now if we can.

heathermiller:
Sometimes member search results (especially when there's an implicit keyword) get a little visually crowded on the page. For example, try searching just "Java". You'll see a bunch of implicit conversions on JavaConversions that is a bit hard to visually parse. I don't feel strongly about this, but was just thinking that maybe a bit of vertical spacing would help? Not sure this concern is particularly important.

lrytz:
i really like the change.

one thing that i'm not sure about is that there's no longer an overview on the project's packages and classes that can be browsed. basically the search page with no search term (probably without member results), and maybe with the possibility to fold (`+` / `-`) packages ("hide" / "show" in current scaladoc).

another idea (opinions welcome): i think it would be nice if the search term didn't disappear. if i select a result, hit "esc" or click "< Back", the search term would remain. when i click again on the search box, the results overlay would come back. the search term should ideally remain also when navigating to a different class / companion / package. using this functionality we could also display the package overview mentioned above when focusing on the search bar (when there's no search term yet).

i agree with all of Heather's comments.

what do you all think?

heathermiller:
RE: @lrytz's point on there being no overview. He's right. Scala programmers have adopted the Java-based reliance on there being a list of packages somehow visible in the Javadoc. So I too find myself itching for some kind of outline of what's in the library. 

But at the same time, I really like how things are now. I feel that removing all clutter like you've done really helps the user to focus on the actual docs. As you said yourself, you actually read the top-level package docs which you realized was unusual for you to do.

I worry that adding an overview to the left side à la Java would bring that distracting clutter back that begs peoples' attention _away_ from the actual documentation. I get the feeling that as programmers, we use the overview only a handful of times (the first few times we encounter a library), instead spending more time in search and entity pages. Maybe there's some sort of middle ground to be had? Lukas's suggestion of the empty search box leading to a package overview would be great, but likely undiscoverable (at least not by intuition). @felixmulder, or anyone else, do you have any ideas?

@lrytz, I didn't quite understand your second suggestion. Could you try to clarify? What I understood was:
- search term is left in the search box after you select a result.
- when clicking the search bar with an old search term in it, the previous search results page returns.

I didn't understand this though: 

> the search term should ideally remain also when navigating to a different class / companion / package. using this functionality we could also display the package overview mentioned above when focusing on the search bar (when there's no search term yet).

Could you clarify?

felixmulder:
@heathermiller:

> - The search results no longer reveal companions. Sometimes it's useful to navigate directly to object Boolean rather than class Boolean, for example. In the old API docs, this was remedied by having two little icons next to the name of the entity that had a companion; either a little trait and a little object icon, or a little class and a little object icon. Can this functionality be returned?

As two results or as one? I.e.

```
(o)(t) Map
```

Or:

```
(t) Map
(o) Map
```

Either is fine, I like the first one. WDYT?

> -  In the old API docs, clicking on the package name brought you to the package object page. How do I get to a package object page now with the new search? It seems I can't any longer. _(blocker)_

That's a miss, it used to be a link before the result was split into two panes. I'll get right on it.

> - Keyboard shortcut: when in the search results page, can hitting "escape" be set to go back?

Go back to where? Clearing removing the search result view and revealing wherever the user was previously?

> - I'm really happy that there's now this doctitle thing being displayed in the top left. But it would also be good if we could maybe display a version number as well. Would this be possible? Also, could you refresh my memory on how this needs to be set? This seems like something that should also be added to the build.

You really don't want to see the version number generated by `ant docs`. Might be better to just adjust the title passed to `scaladoc`?

> PS, do we really need the dropdown? The search result page is really quite nice!

I was hoping you'd come to this conclusion :sweat_smile: I don't really think it would add anything.

> Actually, I just realized that I can't navigate to companion objects at all currently. Not unless I first go to the class and then click on the "see related docs: object Boolean" link. There should be a way to directly go to companion objects from the search results.

You should be able to click the large semi-folded entity icon?

> Also, when clicking on the text in the top left, in this case, "Scala Standard Library", is it possible to make that a link that brings us back to the root package?

That should be no problem - I'll see what I can do.

> Sometimes member search results (especially when there's an implicit keyword) get a little visually crowded on the page. For example, try searching just "Java". You'll see a bunch of implicit conversions on JavaConversions that is a bit hard to visually parse. I don't feel strongly about this, but was just thinking that maybe a bit of vertical spacing would help? Not sure this concern is particularly important.

I'll see what I can do about this, fonts should also be bold for `def`, `val`, `lazy` etc.

@lrytz:

> i really like the change.

Thanks :smile:

> one thing that i'm not sure about is that there's no longer an overview on the project's packages and classes that can be browsed. basically the search page with no search term (probably without member results), and maybe with the possibility to fold (+ / -) packages ("hide" / "show" in current scaladoc).

I sort of agree, it's easy to say - I need some sort of `List` type. But it's harder to just browse randomly. It would be fun to visualize the docs in some form of tree view! :evergreen_tree: 

> another idea (opinions welcome): i think it would be nice if the search term didn't disappear. if i select a result, hit "esc" or click "< Back", the search term would remain. when i click again on the search box, the results overlay would come back. the search term should ideally remain also when navigating to a different class / companion / package. using this functionality we could also display the package overview mentioned above when focusing on the search bar (when there's no search term yet).

This has more to do with the fact that we're basically reloading the page. One idea I've been toying with for the `dotty` docs is to simply generate JSON and then have the docs be completely made in angular. This would allow us to do some fun things and get a snappier experience.

heathermiller:
+:100: on generating JSON! That's useful not only for the JS framework but also for potentially delivering documentation to other tools! Would also allow people to use dottydoc to produce different documentation tools! :))

felixmulder:
Would something like this be interesting for showing an overview: http://visjs.org/examples/network/layout/hierarchicalLayout.html ?

lrytz:
> Could you clarify?
> 
> > the search term should ideally remain also when navigating to a different class / companion / package.

I just mean that the search term remains in the search box when navigating around (clicking on a link to a different class, or a companion), so when you focus again on the search box, you get your last result page.

> > using this functionality we could also display the package overview mentioned above when focusing on the search bar (when there's no search term yet).

this just repeats the idea that focusing on an empty search box would lead to a package overview

felixmulder:
@VladUreche thanks, fixed it! (haven't uploaded to DB yet though...will do that once I've addressed the other issues)

heathermiller:
RE: @felixmulder 

> As two results or as one? I.e.

`(o)(t) Map`

Yep, this would be great :)

> Go back to where? Clearing removing the search result view and revealing wherever the user was previously?

Actually, yes. But this already works. Wonder what happened when I tried it the first time /shrug

> You really don't want to see the version number generated by ant docs. Might be better to just adjust the title passed to scaladoc?

Well, no version number generated by ant. Maybe we can just use the user-set version number? I imagined we could change the styling for the version number slightly, e.g., italic gray while title of the docs remains white. Just brainstorming. I'd just like the docs to correctly say "Scala Standard Library v2.11.6" or so.

> You should be able to click the large semi-folded entity icon? 

Yes but not in the search results page. Let's say when I type in "Boolean" into the search bar, I really want to go directly to the Boolean object, not the Boolean class. How would I do that without clicking first through the class page? I think it's best if both options are available from the search results page.

See what I mean?

heathermiller:
RE: @felixmulder 

> Would something like this be interesting for showing an overview: http://visjs.org/examples/network/layout/hierarchicalLayout.html ?

That's nice, but it seems complicated. I wouldn't know where we'd put it. Also, we already have the generated class hierarchies that fill a similar sort of niche, so this seems like conceptual aliasing.

I would imagine something simpler and Java-like, i.e., just a list of packages would do the trick.

@lrytz has a good idea about the empty search bar resulting in an overview. I'm just not clear on how discoverable it would be–how do you kick it on for the first time? I may be misunderstanding, but if the suggestion that when the search bar focused/selected, you are immediately switched you to overview mode, that could be useful. But it could also be annoying. It might be too aggressive of a screen switch–with no keys pressed, the page you were looking at was hijacked. This could be me senselessly worrying though, so it might be best to just try it first to see how it behaves and then discuss from there...

VladUreche:
> @lrytz has a good idea about the empty search bar resulting in an overview. I'm just not clear on how discoverable it would be–how do you kick it on for the first time? I may be misunderstanding, but if the suggestion that when the search bar focused/selected, you are immediately switched you to overview mode, that could be useful. But it could also be annoying. It might be too aggressive of a screen switch–with no keys pressed, the page you were looking at was hijacked. This could be me senselessly worrying though, so it might be best to just try it first to see how it behaves and then discuss from there...

Why not bring back the trigram on the left, with the package list hidden away by default?

lrytz:
> Why not bring back the trigram on the left, with the package list hidden away by default?

I'd try to keep the simplicity of the current proposal if possible..

heathermiller:
RE: @VladUreche see above discussion, we believe it's noisy.

felixmulder:
I would love to test this on other Scala users. I mean, this is pretty much our ~~ugly~~ better looking baby right now that we're changing back and forth.

When it comes to the overview - let me try some stuff out the coming month. For now let's keep the current look and when M4 is released, I'd expect people to voice their opinions.

**EDIT:** btw, searching for "." will yield all possible entities.

felixmulder:
@heathermiller:

> > Go back to where? Clearing removing the search result view and revealing wherever the user was previously?
> 
> Actually, yes. But this already works. Wonder what happened when I tried it the first time /shrug

Found it, `escape` only works when the field is selected, when navigating the entities - it won't clear the results. I'll fix it.

heathermiller:
> Found it, escape only works when the field is selected, when navigating the entities - it won't clear the results. I'll fix it.

:clap: :tada:

felixmulder:
Just an FYI for @heathermiller:

An example of the doc version generated by `ant docs` is `2.12.0-20160208-123622-ced0dbd786`. As such, if we want to use the same value available we're in for a bad time. (That's what I meant earlier!)

So, two options. Either we add a new parameter to `scaladoc` (super simple to do) or we change the version parameter in the build script. WDYT?

heathermiller:
What I had in mind the whole time was a user-provided value, like you suggest :) Not the ant-generated thing. I was kind of wondering out loud whether it was possible to get that from the build script somehow in a non-brittle way.

Maybe someone like @SethTisue, @adriaanm, @lrytz, or @retronym could weigh in here? What would be the most convenient way for you guys to get the Scala version into Scaladoc (so it's displayed on the actual Scaladoc pages)

felixmulder:
Getting the version number is quite simple: https://github.com/scala/scala/blob/2.12.x/build-ant-macros.xml#L433

The question is just if we want to change that variable - or add a new one :)

**PS:** If we are using a ant to produce a nightly doc - then it would suck if the version number is too long (it'd look ugly on the generated site).

heathermiller:
Ah, I missed that field. I didn't realize we had a scaladoc specific version number to populate as well. Then we ought to use that one I imagine. Why would it be desirable to use any other version number?

szeiger:
Here's your choice of version numbers as produced by the sbt build for PR validation:

```
[info] Canonical: 2.12.0-20160208-152005-fbba819, Maven: 2.12.0-fbba819-SNAPSHOT, OSGi: 2.12.0.v20160208-152005-fbba819, github: fbba819
```

We currently use the canonical version for scaladoc but that's easy to change (globally or for an individual build type).

felixmulder:
@szeiger: My concerns in this matter are purely aesthetic. Ideally I'd like the task `ant docs` to generate a scaladoc with a version number like:

```
2.12-M4
```

```
2.12.0
```

```
2.12-NIGHTLY
```

As they are short and will fit nicely on the new nav-bar: https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

WDYT, should we change the default `ant docs` task's `docversion`?

I mean, we could warn if the `docversion` is too long and turn it into an empty string...just spitballin'

@heathermiller: what's the approach when generating for scala-lang.org?

szeiger:
Personally I'd take a precise version number in a nightly build over a short one. Whatever you end up doing, it should go into the sbt build. I don't know how the nightly docs are built (there are no nightly Scala builds at the moment and the PR validation builds don't produce scaladocs) but we're in the process of switching everything to sbt.

heathermiller:
The precise version number can go elsewhere, like in `<title></title>` etc. Felix is right, there should be a human-friendly version number for the front page. 

Does anyone else from the Scala team have any suggestions about where to add such a version number?

heathermiller:
Note that this is for every future build, ever, of Scaladoc and not just the Scala stdlib nightlies...

felixmulder:
Precise version number in the `<title>` section sounds great. I could add an optional `shortdocversion` parameter to `scaladoc`. That could be optional - i.e. it won't show in the navbar unless set. Or - I could omit the version number on the navbar if it's too long and warn the user.

felixmulder:
@heathermiller I fixed all the issues (except version) and upped it to: https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

heathermiller:
> @heathermiller I fixed all the issues (except version) and upped it to: https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

Amazing! Looks good Felix!! Thank you!
Super stoked by this :)

Ah another thought–maybe it'd also be a good idea to put the full version number in a footer somewhere, just so it's easily copy/pasteable (wouldn't be in the title field...)

kzys:
On JSON, the reason Scaladoc was not using that is, XMLHttpRequest doesn't work on file:///. But I do agree that 99% of our customers are using Scaladoc on http://.

felixmulder:
@kzys: you're right! There are ways around this however. Currently the docs have the index contained within a `index.js` which is included after the main `index.js` file. We could easily include files in the same manner - but loading them dynamically may be preferable performance-wise which is possible with functions like: `$.getScript`! That would make it possible to load different types of data in a more modular sense :)

szeiger:
The title looks quite redundant at the moment: 
screen shot 2016-02-08 at 18 42 38

felixmulder:
@szeiger: this is also present in the version on www.scala-lang.org/api but not if hosted locally:

screen shot 2016-02-09 at 10 40 11

I think this has to do with that the document title is set from an `iframe`. I'll file an SI for it.

**UPDATE:** filed here: https://issues.scala-lang.org/browse/SI-9649

felixmulder:
@heathermiller - so this is what I ended up with: the version string will be shown in the `<title>` no matter what. However, if its length is leq to 12 characters (allowing for `XX.XX.XX-XXX`), the version is also shown in the nav-bar. As will be the case for `2.12.0-M4` as shown here:

https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.Boolean

Would be fun to get the search feature in M4! :smile: 

**EDIT:** scaladoc will warn if the user supplied a version number greater than 12 characters in length

heathermiller:
Ah, that's beautiful :) Looks great!!

heathermiller:
Just to determine the status of this, @felixmulder can you recap on what's left to do? I _believe_ that the only biggish unresolved issue brought up in this PR is finding a way to still support package outlines (thx, @lrytz!). Since this PR removes the old package outline all together (i.e., the filtering side bar), it might be best to also deal with replacing it in this PR as well. Unless we decide to not bother... Thoughts?

All of the other issues appear to be addressed, am I right?

heathermiller:
Just to clarify... By "replacing it" I simply mean proposing some way to provide the desirable bits of functionality that this PR removes. E.g., outline view of packages, but not necessarily the filtering that the left side bar used to do. And not necessarily by re-adding a left side bar, we all agree that life is better now that it's gone.

jrudolph:
> the version is also shown in the nav-bar

It seems to overlap the search bar for some widths. E.g. for 1246px width:

image

Can be fixed by adjusting the css selector to e.g. `@media only screen and (max-width: 1300px)`

felixmulder:
All issues (except the outline) have been resolved. Apparently I borked some tests though, that's all that remains. I'll get it done within the hour.

I would like one more chance to argue my suggestion for the tree-view. Let me mock it up for you to show what I mean.

felixmulder:
@jrudolph thanks! I'll get that one in as well. What device & browser?

heathermiller:
Ok, sounds good @felixmulder!
And of course thanks x :100: :smile: :tada:

lrytz:
Just tried on my phone (iPhone 6, Chrome):

image

> proposing some way to provide the desirable bits of functionality that this PR removes

I'm fine removing the sidebar completely for now, instead of spending time on a temporary solution that we don't really want to keep.

BTW, we should keep track of TODOs for scaladoc, i.e., work that is supposed to happen in follow-up PRs. Since JIRA is a bit hard to use, we use the issue tracker at https://github.com/scala/scala-dev/issues for tracking work in progress – @felixmulder, if you like that workflow I can give you permissions over there.

jrudolph:
> @jrudolph thanks! I'll get that one in as well. What device & browser?

Chrome on a Linux Desktop with arbitrarily sized horizontal tiling. Pure chance I hit that size exactly.

I like the new design :)

Two comments regarding the functionality:
- I agree that the outline isn't useful in many use cases. On the other hand, there are cases like new users or users that want to expand their knowledge of a library by browsing in a semi-directed fashion who would benefit from some kind of outline. I guess some library authors even arranged classes and documentation in a way that assumes an outline is available to discover an API.
- The global search itself would benefit from adding a summary line of context to each entry. Especially, if you start looking for something globally, you don't know exactly what to look for and may need to find out quickly if an entry is what you are looking for.
- The search bar still captures all function keys. Navigation with <PageDown> etc. don't work.

lrytz:
While on my phone I found another issue: I cannot actually scroll the contents of a page (like Boolean), instead swiping just drags the header bar around. See https://dl.dropboxusercontent.com/u/4881107/scroll.mp4

felixmulder:
@lrytz sure, I could try my hand at that.

Thanks for the video, all these finds are crucial - I believe it has something to do with absolute positioning on mobile devices. I'll look into it!

> - The search bar still captures all function keys. Navigation with etc. don't work.

@jrudolph: you mean the F1-FX keys? Navigation with what?

lrytz:
Two more things on the phone:
- When focus enters the search bar, the browser slightly zoomes in; that's a bit annyoing because the search result page is not fully displayed after
- Swiping in the search result page doesn't continue scrolling, the scroll abruptly stops when lifting the finger.

Both shown here: https://dl.dropboxusercontent.com/u/4881107/swipe.mp4

felixmulder:
@heathermiller: so this is what I had in mind, in the first picture notice the icon on the upper left hand side that is supposed to be slightly visible against the `body` bg.

screen shot 2016-02-10 at 15 31 37

When the icon is clicked, it reveals a tree overview of the entire lib. (Imagination needed for the next pic to replace the numbers by class names)

screen shot 2016-02-10 at 15 38 41

However, I'd like this in a separate PR :)

szeiger:
I tried it with Chrome on Android:
- Scrolling works fine
- Can't focus the search field (probably hidden behind the overlapping text)
- Opening and closing the individual sections feels sluggish. You don't really see any animation, it's just slow. Might be best to disable the animation on mobile browsers. (With Firefox on OSX I can see the animation but the text inside the expanding section jumps around)
- The down triangle for opened sections is rendered much smaller than the right triangle for closed sections.
- The big distance between the triangle and the method name (with possible modifiers and a lot of whitespace in between) pushes the method names and args too far to the right on a small screen, causing overflow

jrudolph:
> Navigation with what?

Oops, github ate the angle brackets... I meant PageDown and End

felixmulder:
@szeiger would you mind posting a screenshot of this one please:

> - The big distance between the triangle and the method name (with possible modifiers and a lot of whitespace in between) pushes the method names and args too far to the right on a small screen, causing overflow

?

szeiger:
@felixmulder here's an example: http://i.imgur.com/IQuCpbt.png

felixmulder:
@szeiger, hmm yeah that was supposed to be fixed a while back. I'll check on it.

Thanks for the screenie!

felixmulder:
Ok @jrudolph, @lrytz, @szeiger: I've dealt with the issues you guys posted. Most of them should be fixed now - but since some of you have devices I cannot emulate/access, some things might not be 100%.

Please check the fixed issues below:

## Fixed issues
- Overlapping of version and title (@jrudolph)
- Smooth scrolling the search (swiping results page stops abruptly finger release)
- Fix version overlap on mobile browsers (i.e remove)
- Cannot scroll page (iOS)
- Make sure text input field is not hiding search input (droid)
- Fix zoom-in on iOS when selecting search field
- Remove animations on mobile for opening/closing sections

## Remaining issues
- PgDown/up fix
- Replace down-right triangle (with one that doesn't change size!)
- fix @szeiger's indented methods

Regarding the remaining issues, all of them (except the PageUp/Down one) are unrelated to this PR. I'll see what I can do about the page/up down thingy.

lrytz:
cool, i can confirm those fixes. another one I just discovered: when clicking on a member to unfold on mobile, the page jumps to the top (chrome on ios). can you reproduce that?

also, the fold / unfold arrows are inconsistent, i get ▶️ and ▼

jrudolph:
> PgDown/up fix

Actually, I just figured this isn't a regression but normal behavior that PgUp/Down won't work while a text field is focused. It would be nice if it worked but it's strictly no bug.

felixmulder:
@jrudolph: true - but I agree with you. I'll put it on a todo / file a bug report

@lrytz: yeah, for some reason iOS insists on replacing that UTF8-symbol. It's a bug on their part IMO, but I'll look into replacing the arrows with something from the material design font we're using for other things.

Thanks for looking over the issues guys!

felixmulder:
@VladUreche @heathermiller seems like things are nearing completion here. Once the checks have been made we should be good to merge.

lrytz:
Tried a few more things
- [android 6, chrome] when enteringe the search box, the version number shows up again. it disappears when leaving the search box.
- [all platforms] search seems to be case-sensitive (to some degree); the phone was set up to capitalize the first letter, so i naturally typed "Fold" to search for "foldLeft", and don't get a result (not mobile-specific)

https://dl.dropboxusercontent.com/u/4881107/ios.mp4 shows the ios issues.
- [ios, chrome and safari] the scrolling issue seems to be still there, at least some times. i tried on more than one phone, sometimes i can scroll, sometimes it just moves the header around.
- [ios, chrome and safari] some pages take a really long time to display; it's not a connection problem, i'm opening the page for the second time, content should be cached. works fine on android.

some more here: https://dl.dropboxusercontent.com/u/4881107/jumping.mp4
- [ios, chrome] when clicking on a member, the page often jumps to the top
- [ios, chrome and safari] some page bodies are wider than the screen, for example "object Boolean", so i have to scroll vertically.

so things are not really good on iPhone for me. i tried on a iphone 6 and an iphone 5, the problems are the same.

lrytz:
maybe we shouldn't go too deep into the rabbit hole in this PR. if you'd prefer to focus on the desktop version for now and finish up this PR that's fine with me. we should keep track of the issues people found though. i think JIRA is a bit too heavy for this, how do you feel about using https://github.com/scala/scala-dev/issues?

felixmulder:
Looking at the things you found now. I think some of them deserve discussion. I'm putting them all in an issue on the `scala-dev` repo anyhoo :smile:

About the auto-capitalization. I think that it shouldn't be auto-capitalized (fixing ATM). But also - should we differentiate between `flatmap` and `Flatmap`?

felixmulder:
Regarding going "deep into the rabbit hole in this PR" - I'll look at some things tonight. But I strongly suggest we merge this for M4. I believe that we'll be getting a lot more eyes on it that way.

felixmulder:
Auto-capitalization has been removed - also there should be no object / trait header that is too wide for the display @lrytz

VladUreche:
@felixmulder Regarding M4, on the technical side I think the patch is ready to merge (it was good to merge even 2 weeks ago :+1:). I'll let @heathermiller and @lrytz comment on the design side, and when they're happy, we can merge.

felixmulder:
Thanks @VladUreche, I'm keeping the branch up-to-date till they decide :)

Just a thought - now that we have eliminated the entity listing, is there really any point to keeping this structure:

``` html
<div id="search">
  <!-- Search bar here -->
</div>
<div id="search-results">
  <!-- Search results end up here -->
</div>
<div id="content" style="-webkit-overflow-scrolling: touch;">
  <iframe id="template" name="template" src="entity.html">
    <!-- Entity documentation here loaded from a different HTML-file -->
  </iframe>
</div>
```

I mean, there really is no point in having an iframe. We could just as well be placing the search-related divs in the file loaded by the iframe and load that directly. This would also simplify permalinking, but sure the URI would be somewhat uglified - WDYT @heathermiller?

**EDIT:** the reason I want to kick the iframe is two-fold. Number one: the `iframe` is the reason for the weird scroll issue on iOS. Number two: when the time comes, I think we should build the doc dynamically from JSON.

felixmulder:
@heathermiller, @VladUreche, @lrytz: I tried out the no-iframe idea. Everything is so much better. I found one small issue, iOS will push around the view in Safari (not Chrome) when showing the keyboard for search; when the keyboard is hidden the view will resume its original position. Hopefully, this solves all the other problems you guys had:

https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/index.html

Note that this reworks the structure of scaladoc. There are no longer any "package.html" - instead each package has an `index.html` (as it's supposed to be). There is no special top "index.html", instead everything is treated as an `EntityPage`.

Please also note that I have not pushed these changes to this PR. It was done purely for experimentation. If this looks better to everyone involved - I say we drop the `iframe` and go with this approach.

VladUreche:
@felixmulder LGTM in my corner of the universe (Firefox + Ubuntu).

While we're at the styling part, can you please change `Implicit information` to just `Implicit`:

It's so ugly when it pushes the right column...

felixmulder:
@VladUreche: done, reuploading.... :smile:

VladUreche:
One more thing: can you hijack `Ctrl+F` and place the cursor in the member filter bar? Also maybe `Ctrl+F` twice to jump to the upper search tool?

In fact, it would be great if you hijacked `Ctrl+F` to the member filter bar and showed a little hint somewhere that "Hit Ctrl+F again for global search". Then, hitting `Ctrl+F` again would take you to the upper bar, with a hint "Hit Ctrl+F again for member search".

The hint would serve a double purpose:
1. Jump between search boxes
2. Showing users where the cursor is positioned

WDYT?

And one more spacing issue, from [`Predef`](https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/scala/Predef$.html#%3C:%3C[-From,+To]extendsFrom=%3ETowithSerializable) this time:

It should be `sealed abstract class` but `s` gets cut off.

VladUreche:
@felixmulder wow, that was fast!

felixmulder:
I'll add the Ctrl-F to the ToDo-list at: https://github.com/scala/scala-dev/issues/84 :+1:

felixmulder:
Regarding the predef thingy - it's the issue that is plaguing mobile view. We need a more responsive layout for the members. It's on my ToDo :)

VladUreche:
Thanks a lot Felix! :+1: :tada:

felixmulder:
@VladUreche: here's them without the "information" bit: https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/scala/Float.html#isWhole():Boolean

VladUreche:
Yaay, thanks @felixmulder!

felixmulder:
Had to fix some tests with the new structure. This latest commit is about as far down the rabbit hole as I'd like to go with this PR. Mobile browsing should be OK on all platforms now.

The final version is available on: https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/index.html

Ping: @heathermiller, @VladUreche, @lrytz

VladUreche:
@felixmulder, regarding the member indentation, seems it's even more broken now:



https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/scala/collection/mutable/index.html

VladUreche:
I added a link to the comment above to https://github.com/scala/scala-dev/issues/84, for reference.

felixmulder:
@VladUreche let me get back to you on that in 30 min.

felixmulder:
So here's the deal: large modifiers (aka `protected abstract def`) will f the alignment.

So options:
- adopt a max size and force the modifiers to that size. Drawback - looks ugly on mobile devices (old method)
- use css-table to allow for a dynamic left offset (this PR), drawback - if a group contains different types of members these won't be aligned to the same point
- hide modifiers on mobile?
- max sizes and align everything to the right?

I'll try out the last two tomorrow. Looks like M4 slipped anyway so I figure I have some time.

lrytz:
> Looks like M4 slipped anyway

There's still time for M4, at least a week (will keep you posted, we have a meeting on Tuesday where we'll probably reach a decision).

"Filter members" doesn't seem to work on the current version (eg https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/scala/collection/TraversableLike.html)

felixmulder:
Fixed filter members :+1:

lrytz:
I looked at it again, things feel _much_ more smooth now!

Here's a log of some things I noticed. I'll leave the decision to you if you want to fix any of them here or log them at https://github.com/scala/scala-dev/issues/84. Since they are mostly about mobile, which is not a priority, this PR LGTM in its current state.
- clicking on a package doesn't open its doc
- Mobile: when rotating the phone to landscape the font becomes quite large (compare [1] and [2])
- Desktop and Mobile: there's a triangle when on top of the modifiers when looking at the "Full Signature" (TraversableLike.foreach), see [3]
- Mobile: unfolding / folding the doc of a member requires two taps. the first seems to do what hovering does on the desktop version.
- Mobile: unfolding the doc comment of TraversableOnce.reduceRight: the text is too large and overflow, not everything is readable
- Mobile: There's no more get the `abstract` modifier (TraversableLike.foreach, package mutable) - is that by design?
- The modifiers alignment issue that Vlad mentions. I think on Desktop members should be aligned, and the row for modifiers should take as much space as actually required by one of the members. On Mobile it's probably better to not waste space.

[1]
image

[2]
image

[3]
image

felixmulder:
@lrytz : as I've said before, I'm not a fan of big PRs :smile: But that being said, I think that anything merged has to work on its own. So IMO let's merge this once I've fixed:

> - clicking on a package doesn't open its doc

Where are you clicking? Can't seem to reproduce this on Chrome OS X, is this a mobile issue?

> - Desktop and Mobile: there's a triangle when on top of the modifiers when looking at the "Full Signature" (TraversableLike.foreach), see [3]

I'll get right on this.

After that my priority will be refactoring the members to a table element instead of a list. This will allow us to have the same width for all modifiers (i.e. `abstract def` and `val` will have the same width) - since they'll end up in the same column.

I would like to do add the table in a separate PR which addresses the mobile issues that have been found. I'm adding these to the https://github.com/scala/scala-dev/issues/84 ToDo.

felixmulder:
@lrytz: I couldn't see the emoji arrow now - has it disappeared for you?

lrytz:
> > clicking on a package doesn't open its doc
> > Where are you clicking? Can't seem to reproduce this on Chrome OS X, is this a mobile issue?

For me it doesn't work on desktop nor mobile. Open https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/scala/collection/TraversableLike.html, enter "mutable" in the search box, scroll down in the serach results to package "scala.collection.mutable" and click on it. For me it just adds `#scala.collection.mutable.package` to the URL, nothing else happens.

> I couldn't see the emoji arrow now - has it disappeared for you?

I still get the ▶️  on mobile

felixmulder:
Thanks - fixed the package links :+1: 

EDIT: I'm adding unit tests for these in a later PR.

felixmulder:
@lrytz: so I fixed both package links and the full signature arrow. I also reverted the behaviour of the modifier alignment - so it should look OK now on both mobile and desktop.

I will fix the rest of the issues in separate PRs once this is merged.

felixmulder:
ping: @lrytz

lrytz:
To me it looks all good, but let's wait for a few others to take a final look!

VladUreche:
@lrytz LGTM from my side. @heathermiller?

retronym:
The new search facility looks great!

I'm somewhat uncomfortable with removing the sidebar altogether. I think it is quite common to poke around the package tree to get a feel for how a library is organized. When you're completely new to a library, and don't know what things are called, you don't know what terms to search for. I'm fine with merging this PR so long as we agree to revisit this choice before for 2.11.0-RC1, perhaps after getting some feedback from the community.

Related to this: when I search for "collection" (to find the package), all entities within that package. I'm not sure if this is information overload or not.

The links to the packages itself is a bit hard to discover: they don't have the "P" icon next to them in the same was as class/object/trait, and the link isn't underlined. We should try to steer people towards package level documentation. I know when reading Java or Scaladoc of a new library I sometimes feel lost until I stumble across a great summary in the package docs.

Not sure how much this would help, but I was thinking that it could be useful to search the text of the doc comments in addition to the method names. Could be fun to play with that idea.

Another idea for future refinement: Is it possible to add the search page to the browser history when you click through to a search result? Otherwise, I need to re-enter the search term.

A more immediate problem in the same vein: I would probably work around the history problem by Option-clicking to open the search result in a new tab. While this does open the result in a new tab, the search is cleared from the original tab (Chrome 48.0.2564.109 on Mac OS 10.11). Would be good to find a fix for that in this PR.

I notice that the tooltip for the search results is just the name of the result. Could we put something more useful in there? Perhaps the headline of the doc comment of the result? e.g. something like:

image

(Problems: do we use the class or companion object documentation here? Do we need to truncate?)

This would be really good to display in the grouped packages, too:

image

Anyway, aside from the issue of the search results disappearing on "open in new tab", this is all fodder for future PRs and discussions. Thanks for your work so far!

felixmulder:
Hey @retronym, thanks for the feedback - greatly appreciated!

> so long as we agree to revisit this choice before for 2.11.0-RC1, perhaps after getting some feedback from the community.

Couldn't agree more (I assume you meant 2.12)!  As @lrytz said further up: "I'm fine removing the sidebar completely for now, instead of spending time on a temporary solution that we don't really want to keep."

> The links to the packages itself is a bit hard to discover...sometimes feel lost until I stumble across a great summary in the package docs.

The idea here was that visually something should separate the search results in each category (entity and member results); e.g. I don't want to mistakenly find the mutable `Map`. I totally get where you're coming from, they should appear clickable at the very least (i.e. underlined) - I'm not totally convinced about giving them an icon but this is something we could definitely try out.

> Related to this: when I search for "collection" (to find the package), all entities within that package. I'm not sure if this is information overload or not.

Search works by looking at the qualified name of an entity i.e. `scala.collection.immutable.Seq`, thus if you search for `immutable`, you'll get all the entries of said package. I find this pretty useful :) 

> Not sure how much this would help, but I was thinking that it could be useful to search the text of the doc comments in addition to the method names. Could be fun to play with that idea.

Let's try it out, it might slow down the search somewhat, but could be scheduled with a lower priority than the other work - might work :)

> Another idea for future refinement: Is it possible to add the search page to the browser history when you click through to a search result? Otherwise, I need to re-enter the search term.

Just to be clear, what you want is to be able to click an entry in the search results - then press the back button and end up with the same search results? This is good! Let's do this - adding it to the https://github.com/scala/scala-dev/issues/84 todo.

> ...open the result in a new tab, the search is cleared from the original tab (Chrome 48.0.2564.109 on Mac OS 10.11). Would be good to find a fix for that in this PR.

These are remnants of the `<iframe>` design - previously we needed to set the `<iframe>` source to the entity clicked in the search results. I'll make these links `<a>` tags instead, should be done in this PR as you said.

> I notice that the tooltip for the search results is just the name of the result. Could we put something more useful in there? Perhaps the headline of the doc comment of the result?
> (Problems: do we use the class or companion object documentation here? Do we need to truncate?)

Putting it on the https://github.com/scala/scala-dev/issues/84 todo! We previously added a `@hideImplicitConversion` annotation to the docstring - one idea could be to do the same for something like a `@shortDescription` if it proves problematic to truncate etc. Perhaps @VladUreche has input on this?

VladUreche:
> Putting it on the scala/scala-dev#84 todo! We previously added a @hideImplicitConversion annotation to the docstring - one idea could be to do the same for something like a @shortDescription if it proves problematic to truncate etc. Perhaps @VladUreche has input on this?

Or just take the first sentence of the description, like we do for members. WDYT?

felixmulder:
@VladUreche: as long as the first sentence is a good explanation of the Entity/member - that seems like the easiest approach!

VladUreche:
> as long as the first sentence is a good explanation of the Entity/member - that seems like the easiest approach!

Assuming a project has a well-written documentation, the first sentence is the most succinct presentation of the entity. Assuming it doesn't, it's unlikely devs will think about the `@shortDescription` annotation, they'd better do the main documentation first. So I'd stick to picking the first sentence.

VladUreche:
Or, better yet, have both, with a bias for `@shortDescription` when it's available.

felixmulder:
added it to the https://github.com/scala/scala-dev/issues/84 :+1:

SethTisue:
I love the new search functionality — this is a really really nice improvement.

I'm sorry to be raising this so late, but: 

> I'm somewhat uncomfortable with removing the sidebar altogether

I'm really uncomfortable with this. I very much want to browse, too, not just search. I could live with the browser being collapsed by default, I guess, if doing it that way is a real win, but I would like it to be at most one click away.

VladUreche:
@felixmulder, I also feel that we should have some kind of package listing, hidden or not. What is your feeling, should we sprint for a quick restore of the current sidebar (in this PR) or do you feel confident you'll add a better package listing if you aim at M5?

By the way, M4 has been pushed to Feb 26, so there's still a bit of time :+1:

lrytz:
I'd keep this PR as-is. The goal is to have a package overview, not necessarily a sidebar, that's to be discussed and desigend, and a new PR will be the right place for this work.

SethTisue:
I should probably have said above that I (& the rest of the Scala team @ Typesafe) are just fine with losing the overview for one milestone if that's the way this goes — getting this PR into M4 is worth it.

heathermiller:
Let's keep this PR as-is. 

I 100% agree that we need an overview of package structure, but I don't think faking people out with a left sidebar that looks the same as before but is suddenly non-searchable is the best way to provide it. So as @felixmulder suggests, I say we put it on hold and deal with it in a follow-up PR. 

@felixmulder, I don't know if you got around to the little things that @retronym suggested (e.g., dealing with iframe remnants). Let us know. 

But in general, this LGTM. I say merge on @felixmulder's go-ahead.

lrytz:
So it seems the last thing to do for this PR is the issue that Jason mentioned: command-clicking on a search result opens the result page in a new tab (as it should), but it also closes the search page - it should remain open. BTW, command-clicking on a package opens it in the same tab, so the link functionality seems to be implemented somehow differently.

felixmulder:
Business up front: I've refactored the search results to display links - i.e. cmd/ctrl+click for opening stuff without closing the search results is now possible for all entities. Uploaded the docs here: https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/index.html

Please verify that it works as intended - then this PR is good to go.

> @felixmulder, I don't know if you got around to the little things that @retronym suggested (e.g., dealing with iframe remnants). Let us know.

@heathermiller: everything's been added to the https://github.com/scala/scala-dev/issues/84
I'll be seeing to it during the coming weeks.

## Discussion fodder

So I agree with @SethTisue on the matter of being "uncomfortable" - but I also think it's interesting. To me it raises the question - is a listing the best way to display an overview of the library? Can we use an overview to push people toward package documentation as @retronym mentioned?

What do you guys think about a tree view? (mockup further up somewhere)

Regarding the timeframe. I was originally going to be finished with my thesis on March 18 and heading for Lausanne on April 1; but it now appears that we've finished early (currently attending to feedback from supervisors). If I have until Feb 26 to get in some form of overview functionality - that should be entirely possible depending on the scope.

felixmulder:
@lrytz: the latest version of this PR removes the variations of the clicking functionality :+1:

lrytz:
LGTM :star2::star2::star2::star2::star2:

VladUreche:
:+1: :+1: :tada: :balloon: :+1: :+1: 

Thanks @felixmulder!! If there's any other feature you'll be adding before M4, please make another PR, since this one is MERRRRRRRRRRGED!!!

felixmulder:
Yay :smile: :tada: I'll be burning through the todos, and looking at the overview, ideas welcome!

heathermiller:
woooo congrats :) :tada: 

One small issue I just found while showing this off :( 
Filtering doesn't work on some pages. For example, if you go to the immutable package and try to search for members, the search bar doesn't do anything
https://dl.dropboxusercontent.com/u/358427/scaladoc-noiframe/library/scala/collection/immutable/index.html

heathermiller:
Correction: the "filter members"*\* bar doesn't do anything. Not the main search bar. That's fine! :)

felixmulder:
This is on the to-do, should be working for things like Boolean - so something's a miss. I'll look into it asap

PR #4914, opened 2016-01-22
Add partial appearance update of inheritence graphs in scaladoc

felixmulder:
This commit updates color and shape of the inheritence graphs. Also adds
a dropshadow on hover instead of opacity change.

Also removes some unneeded assets and corresponding license details.
### Known issues

Graphviz broke upon trying to include svg buttons instead of png.
Looking for a solution, but good enough for now.
### Preview

https://dl.dropboxusercontent.com/u/358427/scaladoc-graphs/library/index.html#scala.util.Try

**Review**: @VladUreche


VladUreche:
@felixmulder there must be something wrong -- the current class is missing:


Then, when I hover over a box (so it should be highlighted), it disappears:

VladUreche:
Regarding svg-s instead of png-s, there's a problem that makes graphviz non-deterministically drop images from the html output. There's a lot of ugly code that aims at restoring those images in the diagram generator, so make sure it's not actually removing the svg-s :))

VladUreche:
Despite the issues, the new style looks very nice, thanks a lot for updating the diagrams! :+1:

lrytz:
@VladUreche I could reproduce your issue only in Firefox. The images / boxes appear in Chrome and Safari.

VladUreche:
Thanks @lrytz! Indeed, I'm using Firefox 43.0.4 on Ubuntu 14.04 :) 

@felixmulder, do you know why the `List` diagram does not appear? https://dl.dropboxusercontent.com/u/358427/scaladoc-graphs/library/index.html#scala.collection.immutable.List

felixmulder:
@VladUreche - too bad regarding the firefox issue. As @lrytz said it appears in the webkit browsers. I'll get right on it.

Regarding the fixes for the stochastic removal of images - I think I might have patched one of them to replace PNGs with SVGs - which would be a disaster if it did spontaneously. I'll fix that one too.

Most likely - `List`s diagram does not appear because Graphviz crashes too many times. I'm not sure why this is. I'll see if I can get it working if you want to see it specifically.

felixmulder:
So I found the issue with the "invisible" class - it is an incompatibility with the SVG drop shadow filter :cry:  I'll see what I can do.

VladUreche:
Graphviz is very moody, especially so if you put in complex html in the diagram nodes. That's what we learned the hard way with diagrams :stuck_out_tongue:

felixmulder:
So, due to a bug in webkit (https://bugs.webkit.org/show_bug.cgi?id=65344) I can't seem to get the filter working cross-browser if I don't screw around with the JS, which IMHO is not worth it for a little drop shadow :+1: 

Anyhow, this solution is the most cross-browser and at least enables the functionality in Firefox without breaking Chrome. If the webkit fix should happen, then this will resolve itself in time.

I pushed the changes and after the checks, we should be fine.

https://dl.dropboxusercontent.com/u/358427/scaladoc-graphs/library/index.html#scala.util.Try

VladUreche:
Thanks @felixmulder and sorry for the delayed answer. I was finally able to record my screen and I'm seeing some weird blur when hovering over a diagram node. I made a video of this, and you'll notice how the node's caption becomes blurred and everything shifts to the right:

https://www.dropbox.com/s/7tfynahbr07ifpa/scaladoc.mkv?dl=0

Any idea if this can be addressed? I tested in Chromium and Firefox, and it only occurs in Firefox...

felixmulder:
Fixed?

VladUreche:
Yes!!! Thanks @felixmulder! Now the only problem that remains is the constant crashes from Graphviz that prevent `List` from being shown... Can you please run with `-doc-diagram-debug`?

VladUreche:
By the way, did you change the dot file code, the one fed to Graphviz?

felixmulder:
So I generated only the `List` class using the bin from the `quick.bin` task. No issues here:

screen shot 2016-01-23 at 20 34 22

Here's the output from `ant docs.clean && ant docs`: https://gist.github.com/felixmulder/8d1841ec9aef71994885

I don't think this is any different from when I've previously generated the docs. Could be my Graphviz version or the fact that I'm running generating the docs on OS X. Could try it on Archlinux on Monday after I get home.

**EDIT:** although, the `Serializable` super seems to land somewhat outside...I assume this is a problem? :stuck_out_tongue:

felixmulder:
Regarding the DOT code - not that much has been changed. Small things like `"filled"` to `"filled,rounded"`have been done in `DotDiagramGenerator.scala`

VladUreche:
Thanks @felixmulder!

Regarding the Graphviz version, I assume it must be isolated to your machine. I took the input that crashed dot on your machine and could render it successfully on mine...

Regarding `List`, it must be a problem with down-scaling the diagram (it's done automatically when the diagram is larger than the width of the page). Also, could you generate the diagram with implicits as well (pass `-implicits` to the scaladoc command line). Here's how it looked before:

list-prev

felixmulder:
There we go :) Fixed the down-scaling issue in 9b165f7 - because of new margins it would not resize properly

screen shot 2016-01-23 at 21 42 59

VladUreche:
LGTM! Thanks @felixmulder!

VladUreche:
I'm not sure about the other features, but the new scaladoc look is enough reason for me to upgrade to 2.12! Congratulations @felixmulder :+1:

felixmulder:
You're far too kind! Thanks for all the help :smile: :tada:

VladUreche:
@felixmulder There's one thing that would be nice to improve on diagrams, if you have the time and motivation: When the diagram is very large, for example for https://dl.dropboxusercontent.com/u/358427/scaladoc/library/index.html#scala.collection.package@content-diagram, it's automatically scaled and it gets the option to zoom to full screen. Now, there are three things that would make this nicer:
- zooming the diagram to the entire window without creating a new window at all
- in the zoomed version you should be able to use the scroll wheel/gestures to zoom in/out and dragging to move around (Google maps-like)
- we should have a button to make the diagram full-screen, in case someone wants to explore it on the entire screen

What do you think?

felixmulder:
> - zooming the diagram to the entire window without creating a new window at all
> - in the zoomed version you should be able to use the scroll wheel/gestures to zoom in/out and dragging to move around (Google maps-like)

For the first two something like this jQuery plugin should work: http://timmywil.github.io/jquery.panzoom/demo/

> - we should have a button to make the diagram full-screen, in case someone wants to explore it on the entire screen

For this, I suggest just use an overlay with a duplicate of the diagram.

How does this sound?

VladUreche:
> For the first two something like this jQuery plugin should work: http://timmywil.github.io/jquery.panzoom/demo/

It sounds awesome! The last panzoom example is exactly what we need for the diagrams, so people can drag them around and explore the relations between classes. 

> For this, I suggest just use an overlay with a duplicate of the diagram.

An overlay similar to the search one (which takes over the window and has a back button) sounds great!

felixmulder:
Great! I'll get to it sometime this week-end hopefully :+1:

felixmulder:
@VladUreche: so, I've implemented this in a branch that builds on #4912 

Current state: https://dl.dropboxusercontent.com/u/358427/panzoom-test/index.html#scala.collection.immutable.List

This is the branch I'm talking about: https://github.com/felixmulder/scala/tree/topic/scaladoc-panzoom-diagrams

VladUreche:
Felix, thanks a lot for the update. Awesome, as always!

Here are some minor issues I found (Firefox 44, Ubuntu 14.04):
- Stray `1.` entry on the left of example:
  
- The zoom: the text inside the buttons and the bar are misaligned:
  
- The diagram can go off-screen when it's zoomed out. Can we have this behavior
  - when zoomed in and larger than its window, you can move it outside
  - when zoomed out and smaller than its window, you can't move it outside:
    

WDYT?

felixmulder:
> - Stray 1. entry on the left of example:

I'll get right on that one - I thought I patched it, well anyhoo - I know how to fix it

> - The zoom: the text inside the buttons and the bar are misaligned:

It's hard to get all browsers looking nice when styling these elements as they all have different selectors / ideas. No good standard for those AFAIK, I'll see if there's a way to get a uniform look for them

> - The diagram can go off-screen when it's zoomed out. Can we have this behavior
>   - when zoomed in and larger than its window, you can move it outside
>   - when zoomed out and smaller than its window, you can't move it outside: 

Hopefully I can address this using the jquery-plugin I used. Otherwise, I'll look at something else...

felixmulder:
@VladUreche, there was something called "inverted" containment. I uploaded an example:

https://dl.dropboxusercontent.com/u/358427/panzoom-test/index.html#scala.collection.immutable.List

I think that's principally what you're suggesting in your third point, however, it might need some tweaking to work as expected in the full-screened case.

VladUreche:
Yep, that's exactly what I had in mind. Thanks Felix!
Now I'm wondering if we should allow zooming out beyond 1:1. It just makes everything complicated. WDYT?

felixmulder:
Agreed - uploaded a new version. How about the max-zoom? 5x?

https://dl.dropboxusercontent.com/u/358427/panzoom-test/index.html#scala.collection.immutable.List

VladUreche:
Awesome! :tada:

I think the max zoom should depend on the original scale-down factor (e.g. if we scaled by 1/2 you should be able to zoom to 2x or maybe 4x). If not, put something large, say 20x, so we cover large packages as well.

One more issue: can you set the z-ordering of the "Type Diagram" text above the diagram? Here's what I get now:



Thanks a lot for your wonderful work Felix!!!

felixmulder:
Should be no problem :+1:

VladUreche:
And two more suggestions:
- could we just have `(+)` and `(-)` but not the slider?
- could we give a name to the `[ ]` button? Maybe "Expand"?

felixmulder:
Yes! Vlad, you make me so happy :+1:

PR #4915, opened 2016-01-25
SI-9599 Multiple @todo formatted with comma on separate line

petermz:
Hi!

[This is my first Scala fix, so please forgive any mistakes]

From `bodyTags` definition at CommentFactoryBase.scala:318 and WikiParser.document(), it seems we only generate block-like structures for "@todo", so the comma joiner can be removed as unneeded.

I don't think this fix deserves a test so I didn't write one, or should I?

Thanks!

### DIFF
--- a/test/scaladoc/scalacheck/HtmlFactoryTest.scala
+++ b/test/scaladoc/scalacheck/HtmlFactoryTest.scala
@@ -819,4 +819,11 @@ object Test extends Properties("HtmlFactory") {
     }
 
   }
+
+  property("SI-9599 Multiple @todo formatted with comma on separate line") = {
+    createTemplates("SI-9599.scala")("X.html") match {
+      case node: scala.xml.Node => node.text.contains("todo3todo2todo1")
+      case _ => false
+    }
+  }
 }
--- a/src/scaladoc/scala/tools/nsc/doc/html/page/Template.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/page/Template.scala
@@ -624,7 +624,7 @@ class Template(universe: doc.Universe, generator: DiagramGenerator, tpl: DocTemp
             <dt>To do</dt>
             <dd>{
               val todoXml: List[NodeSeq] = (for(todo <- comment.todo ) yield <span class="cmt">{bodyToHtml(todo)}</span> )
-              todoXml.reduceLeft(_ ++ Text(", ") ++ _)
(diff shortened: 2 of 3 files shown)


petermz:
review @VladUreche  

Hi Vlad -- I've picked you as reviewer as per contributing guide; please let me know if there's better idea.
Thanks!

SethTisue:
/nothingtoseehere

("Found earlier commit(s) marked failure: ae891b" — CLA hadn't been signed yet. has been now, so, clearing the build status)

VladUreche:
Hi @petermz, thank you for your contribution! 

The patch looks good, but let's take care of two little details before we merge:
- can you please rebase on the latest `2.12.x`?
- can you add a test? You can use [this test](https://github.com/scala/scala/blob/2.11.x/test/scaladoc/run/tag-requirements.scala) and [its check file](https://github.com/scala/scala/blob/2.11.x/test/scaladoc/run/tag-requirements.check) as an inspiration (use `ant test.scaladoc` to run the scaladoc tests, which should succeed)

Thanks!

petermz:
Hi @VladUreche !
The code I'm fixing is invoked after the model is built, so I believe I cannot test it with a ScaladocModelTest. Instead, I've added a test case to HtmlFactoryTest which has the necessary infrastructure.

I also hope I've figured out the rebase thing correctly.
Thanks!

test/scaladoc/scalacheck/HtmlFactoryTest.scala
```diff
@@ -819,4 +819,13 @@ object Test extends Properties("HtmlFactory") {
     }
 
   }
+
+  property("SI-9599 Multiple @todo formatted with comma on separate line") = {
+    createTemplates("SI-9599.scala")("X.html") match {
+      case node: scala.xml.Node =>
+        node.toString.contains(
```
VladUreche: Here, could you please check if `node.text.contains("todo1todo2todo3")` instead of hard-coding the html backend output?

VladUreche:
Aside from the one issue I pointed out, LGTM. When you tried the `.text` approach, please ping me again and I will merge. Thanks for your contributions @petermz!

petermz:
Hi Vlad -- all done!

VladUreche:
Awesome! LGTM!

VladUreche:
@petermz congratulations on your first contribution to the Scala code base :+1:

petermz:
Wow! Thanks Vlad!

PR #4917, opened 2016-01-25
SI-9629 Emit missing 'pattern must be a value' error

retronym:
The error used to only be emitted for top-level patterns.

This commit moves it into `typedInternal` so it works
for nested patterns. It uses the typer mode to know
when to fire.


retronym:
Review by @adriaanm

adriaanm:
LGTM

PR #4919, opened 2016-01-27
Pattern Matching analysis improvements

retronym:


retronym:
Might be high value enough for a 2.11 backport once we're happy with these changes.

retronym:
Review by @adriaanm

retronym:
Note that this also closes SI-9399 and SI-9411, in addition to the tickets mentioned in the commit titles.

adriaanm:
Needs more parens:

```
fail - run/t3126.scala  [compilation failed]% scalac t3126.scala
t3126.scala:7: error: ';' expected but 'match' found.
    try (v: @unchecked) match { case Some(1) => } catch { case _: MatchError => }
                        ^
one error found
```

src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
```diff
@@ -84,11 +84,13 @@ trait TreeAndTypeAnalysis extends Debugging {
     tp <:< tpImpliedNormalizedToAny
   }
 
-  // TODO: improve, e.g., for constants
-  def sameValue(a: Tree, b: Tree): Boolean = (a eq b) || ((a, b) match {
-    case (_ : Ident, _ : Ident) => a.symbol eq b.symbol
-    case _                      => false
-  })
+  def equivalentPath(a: Tree, b: Tree): Boolean = (a, b) match {
+    case (Select(qual1, _), Select(qual2, _)) => equivalentPath(qual1, qual2) && a.symbol == b.symbol
```
adriaanm: Should we also cover `Apply` here? (For case class patterns.)

retronym: I checked that we covered all cases covered by compiling the standard library. But I'll try some variations manually and see if we can get here with an `Apply`.

adriaanm: The `Apply` case came up while running `strap.done` for me.

adriaanm: Looking at the specific cases, could be red herring: looks like it's indexing a vararg pattern.

adriaanm:
```
[strap.compiler] equivalentPath? x3.apply(0) / x3.apply(0)
```

retronym:
@adriaanm I've added the parens and a case for `Apply` nodes.

adriaanm:
LGTM! Great catch of my misuse of `correspondsStructure`.

dwijnand:
Very nice. Interesting to note all the tests that needed changing due to this.

paulp:
@retronym Cool thanks.

japgolly:
:+1: :+1: :+1: 
@retronym :beers: :bow:

japgolly:
Would it be possible to backport this for the soon-to-be-released 2.11.8? Also, is there anything I can do to help?

PR #4921, opened 2016-01-27
SI-9534 Use BoxedUnit in all cases for creating Array[Unit]

szeiger:
Calling `wrap` or one of the higher-dimension Array factory methods on
the `Manifest` for `Unit` led to an exception because it tried to use
`void` as a primitive type. Unlike all other primitive Scala types,
`Unit` needs to be boxed. The basic `newArray` method was not affected
by this bug because it was already special-cased. The fix is to also
special-case `arrayClass`.


denisrosset:
Thanks Stefan!

retronym:
LGTM

PR #4923, opened 2016-01-27
Ensure bytecode stability in the closure optimizer

lrytz:
https://github.com/scala/scala-dev/issues/77

Previously, the order in which closure invocations were re-written
depended on the callGraph's closureInstantiations map, which is not
sorted / linked.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
@@ -27,6 +27,22 @@ class ClosureOptimizer[BT <: BTypes](val btypes: BT) {
   import backendUtils._
   import ClosureOptimizer._
 
+  private object closureInitOrdering extends Ordering[ClosureInstantiation] {
+    override def compare(x: ClosureInstantiation, y: ClosureInstantiation): Int = {
+      val cls = x.ownerClass.internalName compareTo y.ownerClass.internalName
+      if (cls != 0) return cls
+
+      val mName = x.ownerMethod.name compareTo y.ownerMethod.name
+      if (mName != 0) return mName
+
+      val mDesc = x.ownerMethod.desc compareTo y.ownerMethod.desc
+      if (mDesc != 0) return mDesc
+
+      def pos(inst: ClosureInstantiation) = inst.ownerMethod.instructions.indexOf(inst.lambdaMetaFactoryCall.indy)
+      pos(x) - pos(y)
+    }
+  }
+
   /**
    * If a closure is allocated and invoked within the same method, re-write the invocation to the
    * closure body method.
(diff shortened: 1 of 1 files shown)


lrytz:
regressed in https://github.com/scala/scala/commit/4eee5e088eb72ba00aa4b02a03a1bcf27b5d751b

lrytz:
review by @SethTisue ?

SethTisue:
let's see what Jenkins thinks: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/286/consoleFull (404 til the worker spins up)

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
```diff
@@ -27,6 +27,22 @@ class ClosureOptimizer[BT <: BTypes](val btypes: BT) {
...
+      if (mName != 0) return mName
+
+      val mDesc = x.ownerMethod.desc compareTo y.ownerMethod.desc
+      if (mDesc != 0) return mDesc
+
+      def pos(inst: ClosureInstantiation) = inst.ownerMethod.instructions.indexOf(inst.lambdaMetaFactoryCall.indy)
+      pos(x) - pos(y)
+    }
+  }
+
```
SethTisue, line 45: wondered if there's a cleaner approach to coding this, but nothing at http://stackoverflow.com/questions/14695833/scala-idiom-for-ordering-by-multiple-criteria seems obviously preferable

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
```diff
@@ -57,27 +73,12 @@ class ClosureOptimizer[BT <: BTypes](val btypes: BT) {
...
-      }
-    }
-
-    // Use collections that keep insertion order, ensures order of closure rewrites (bytecode stability)
-    val toRewrite = mutable.LinkedHashMap.empty[ClosureInstantiation, mutable.ArrayBuffer[(MethodInsnNode, Int)]]
-    def addRewrite(init: ClosureInstantiation, invocation: MethodInsnNode, stackHeight: Int) = {
-      val calls = toRewrite.getOrElseUpdate(init, mutable.ArrayBuffer.empty[(MethodInsnNode, Int)])
-      calls += ((invocation, stackHeight))
+    // sort all closure invocations to rewrite to ensure bytecode stability
+    val toRewrite = new java.util.TreeMap[ClosureInstantiation, mutable.ArrayBuffer[(MethodInsnNode, Int)]](closureInitOrdering)
```
SethTisue, line 78: no standard Scala collection will do here?

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
lrytz, line 45: I didn't use `Ordering.by` because of the additional redirections and the tuple allocation, but maybe it's premature optimization?

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
lrytz, line 78: I could use an immutable TreeMap, but http://stackoverflow.com/questions/4531856/why-is-there-no-mutable-treemap-in-scala led me to use Java's - I don't feel strongly about it, but I kind of prefer to avoid unnecessary overhead.

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
SethTisue, line 45: perhaps, but the resulting code isn't really that much clearer, so I'd say leave it

SethTisue:
LGTM if the bootstrap run passes

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
soc, line 78: @lrytz https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/mutable/TreeMap.scala was added a while ago...

lrytz, line 78: wow, thanks for pointing it out, i only chcked the 2.11 scaladoc :-)
I only saw your comment after merging the PR, will submit a follow-up.

PR #4924, opened 2016-01-27
SI-9452: Extend BigDecimal with Ordered for java interop

ShaneDelmore:


retronym:
`BigInt` should be be changed as well.

ShaneDelmore:
Thanks for the reminder retronym

SethTisue:
review by @Ichoran

src/library/scala/math/BigDecimal.scala
```diff
@@ -539,19 +539,19 @@ extends ScalaNumber with ScalaNumericConversions with Serializable {
 
   /** Less-than-or-equals comparison of BigDecimals
    */
-  def <= (that: BigDecimal): Boolean = compare(that) <= 0
+  override def <= (that: BigDecimal): Boolean = compare(that) <= 0
```
Ichoran: These all re-implement exactly the same operations as in `Ordered`.  Is there any compelling reason to keep them?

src/library/scala/math/BigInt.scala
```diff
@@ -178,19 +183,19 @@ final class BigInt(val bigInteger: BigInteger) extends ScalaNumber with ScalaNum
 
   /** Less-than-or-equals comparison of BigInts
    */
-  def <= (that: BigInt): Boolean = compare(that) <= 0
+  override def <= (that: BigInt): Boolean = compare(that) <= 0
```
Ichoran: Again, the symbolic comparison operators are all defined in Ordered, and they're defined exactly like this.  Do we need to override them at all, instead of just deleting them?

src/library/scala/math/BigDecimal.scala
ShaneDelmore: That is a great point.  I will remove them.

src/library/scala/math/BigInt.scala
ShaneDelmore: Embarassed I missed that :D  And that's why we code review.

Ichoran:
Could you add a test that verifies that the Java interop works?  Normally we recommend using junit, though in this case there aren't many Java tests to built off of in (tests/junit).

ShaneDelmore:
@Ichoran Are there any existing examples of testing Java interop?  I have never actually used Java so I am a bit out of my element here.

Ichoran:
Well, I see

```
test/junit$ find . | grep 'java'
./scala/tools/testing/ClearAfterClass.java
./scala/runtime/LambdaDeserializerTest.java
```

plus a bunch more in `test/files/run` (that are not JUnit, but I guess that's okay).

Ichoran:
Or you could just write a Scala test and test isInstanceOf[java.lang.Comparable[_]].  That'd do also.  (It's not like there's any doubt that the trait's in there now, but tests help prevent regressions.)

ShaneDelmore:
I'll throw in the test for isInstanceOf[java.lang.Comparable.

Ichoran:
LGTM

Thanks Shane!

PR #4925, opened 2016-01-28
Update link pointing to Scala development tutorials

naferx:


mpociecha:
As for me the previous version anyway redirects to the new one. Also it would be nice to have these two commits squashed.

SethTisue:
Thanks for calling this to our attention.

But, there's no need for rootdoc.txt to exist at all anymore, IMO. I'll make a PR removing it from the 2.11.x branch, relocating content to README.md as necessary (I think it might already be there).

SethTisue:
(I haven't done this yet, but I haven't forgotten, either)

SethTisue:
https://github.com/scala/scala/pull/5513

PR #4926, opened 2016-01-28
mutable.TreeMap instead of java.util.TreeMap in closure optimizer

lrytz:
Follow-up for https://github.com/scala/scala/pull/4923

Review by @soc


Ichoran:
LGTM

(If @soc had specific concerns, I am unaware what they are, but this seems to me to replicate the original behavior.)

soc:
LGTM, sorry for the delay.

PR #4927, opened 2016-01-28
SI-9572 Check for illegal tuple sizes in the parser

szeiger:
This commit adds explicit checks with syntax errors for tuple literals
and types of more than 22 elements. An alternative approach to fixing
SI-9572 would be to revert to the old failure mode of Scala 2.10 where
references to arbitrary `scala.TupleXY` would be generated in the
parser, which then leads to “type/object not found” errors in the
typechecker. This fix here is more intrusive but arguably provides a
better user experience.

Methods `stripParens` and `makeBinop` are moved from `TreeBuilder` to
`Parsers` because they can now generate syntax errors. New methods
`makeSafeTupleType` and `makeSafeTupleTerm` implement the error checking
on top of `makeTupleType` and `makeTupleTerm`. They are overridden with
no-op versions in the quasiquotes parser because it also overrides `makeTupleType` and `makeTupleTerm` in a way that supports arbitrary tuple sizes.

### DIFF
--- a/src/compiler/scala/reflect/quasiquotes/Parsers.scala
+++ b/src/compiler/scala/reflect/quasiquotes/Parsers.scala
@@ -59,6 +59,10 @@ trait Parsers { self: Quasiquotes =>
 
       override implicit lazy val fresh: FreshNameCreator = new FreshNameCreator(nme.QUASIQUOTE_PREFIX)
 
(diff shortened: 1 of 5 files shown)


retronym:
LGTM

Small tip for next time to help lazy reviewers like me. For changes that necessitate moving chunks of existing code around, split it into two commits: one to move things around without changing behaviour, and another to make the actual change.

PR #4928, opened 2016-01-28
Document that `scala -e` starts/uses a compilation daemon

szeiger:
We should consider changing the default in the long term. For now, make it explicit that `-e` starts and uses fsc.

### DIFF
--- a/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
+++ b/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
@@ -87,7 +87,11 @@ A file argument will be run as a scala script unless it contains only
 self-contained compilation units (classes and objects) and exactly one
 runnable main method.  In that case the file will be compiled and the
 main method invoked.  This provides a bridge between scripts and standard
-scala source.%n"""
+scala source.
+
+When running a script or using -e, an already running compilation daemon
+(fsc) is used, or a new one started on demand.  The -nc option can be
+used to prevent this.%n"""
 }
 
 object GenericRunnerCommand {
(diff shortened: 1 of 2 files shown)


SethTisue:
I think the addition of this text is an improvement, but I suggest it go at the top level, rather than only under the `-e` option, because the same remarks apply to the script runner, too (i.e., when "a file containing scala source" is the target).

adriaanm:
Yes, we should mention script compilation uses the compiler demon. It shouldn't be the first thing people see though -- I think this is really only something advanced users will worry about. We shouldn't confuse the casual passerby.

> In that case the file will be compiled **(using the compiler demon, unless disabled with -nc)** and the main method invoked.

szeiger:
Updated to mention script compilation, too.

src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
```diff
@@ -87,7 +87,11 @@ A file argument will be run as a scala script unless it contains only
 self-contained compilation units (classes and objects) and exactly one
 runnable main method.  In that case the file will be compiled and the
 main method invoked.  This provides a bridge between scripts and standard
-scala source.%n"""
+scala source.
+
+When running a script or using -e, an already running compilation daemon
+(fsc) is used, or a new one started on demand.  The -nc option is required
```
SethTisue: s/is required/can be used/

src/manual/scala/man1/scala.scala
```diff
@@ -135,6 +139,11 @@ object scala extends Command {
     "line.  Headers can be used to make stand-alone script files, as shown " &
     "in the examples below.",
 
+    "When running a script or using " & Mono("-e") & ", an already running " &
+    "compilation daemon (fsc) is used, or a new one started on demand.  The " &
+    Mono("-nocompdaemon") & " or " & Mono("-nc") & " option is required to " &
```
SethTisue: s/is required/can be used/

szeiger:
Updated accordingly and squashed.

SethTisue:
LGTM

PR #4929, opened 2016-01-29
SI-5365 Enable exhaustivity analysis of matches with guards

retronym:
Rather than disabling the analysis altogether, rewrite
guard to `true` (by default) or to `false` under a new
compiler option.

The default option errs on the side of avoiding spurious
warnings where the guard writer knows better. However,
there was a convincing argument made in the ticket that
in these cases it would not be to onerous to require
the code to be rewritten from

```
case P if g1 =>
case P if g2 =>
```

to:

```
case P if g1 =>
case P       =>
```

Or to:

```
(scrut: @unchecked) match {
  case P if g1 =>
  case P if g2 =>
```

So perhaps we can turn on the strict version by
default, after running against the community build
as a sanity check.

Extractor patterns had the same limitation as guards.
This commit also enables the analysis for them.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
+++ b/src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
@@ -516,11 +516,15 @@ trait MatchAnalysis extends MatchApproximation {
       //    - there are extractor calls (that we can't secretly/soundly) rewrite
       val start = if (Statistics.canEnable) Statistics.startTimer(patmatAnaExhaust) else null
       var backoff = false
+      val strict = settings.XstrictPatmatAnalysis.value
 
       val approx = new TreeMakersToPropsIgnoreNullChecks(prevBinder)
       val symbolicCases = approx.approximateMatch(cases, approx.onUnknown { tm =>
         approx.fullRewrite.applyOrElse[TreeMaker, Prop](tm, {
           case BodyTreeMaker(_, _) => True // irrelevant -- will be discarded by symbolCase later
+          case GuardTreeMaker(_) => if (strict) False else True
+          case ExtractorTreeMaker(extractor, _, _) if extractor.symbol.name != nme.unapplySeq =>
+            if (strict) False else True
           case _ => // debug.patmat("backing off due to "+ tm)
             backoff = true
             False
@@ -761,6 +765,7 @@ trait MatchAnalysis extends MatchApproximation {
       def chop(path: Tree): List[Symbol] = path match {
         case Ident(_) => List(path.symbol)
         case Select(pre, name) => chop(pre) :+ path.symbol
+        case Apply(fun, args) => chop(fun) :+ path.symbol
         case _ =>
           // debug.patmat("don't know how to chop "+ path)
           Nil
--- a/test/files/neg/t5365b.check
+++ b/test/files/neg/t5365b.check
@@ -1,0 +1,31 @@
+t5365b.scala:2: warning: match may not be exhaustive.
+It would fail on the following inputs: None, Some(_)
+  def nonExhautiveIfWeAssumeGuardsTrueOrFalse(x: Option[Int]): Int = x match {
+                                                                     ^
+t5365b.scala:6: warning: match may not be exhaustive.
+It would fail on the following input: Some(_)
+  def nonExhautiveIfWeAssumeGuardsFalse(x: Option[Int]): Int = x match {
+                                                               ^
+t5365b.scala:11: warning: match may not be exhaustive.
+It would fail on the following input: Some(_)
+  def inverseGuards(x: Option[Int]): Int = x match {
+                                           ^
+t5365b.scala:17: warning: match may not be exhaustive.
+It would fail on the following inputs: None, Some(_)
+  def extractor(x: Option[Int]) = x match {
+                                  ^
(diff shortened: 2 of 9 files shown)


retronym:
TODO:
- [ ] Test the strict version against the community build
- [x] Approximate extractor patterns, too, rather than bailing out of exhaustivity.
- [ ] Check for new warnings in our build and address them
  - [x] Compile library/reflect/compiler with `-Xstrict-patmat-analysis`: https://gist.github.com/453b894f164ce7e23665

xuwei-k:
s/exhautivity/exhaustivity ?

test/files/neg/t5365b.check
```diff
@@ -0,0 +1,31 @@
...
+It would fail on the following inputs: None, Some(_)
+  def repeatedExtractor(x: Option[Int]) = x match {
+                                          ^
+t5365b.scala:23: warning: match may not be exhaustive.
+It would fail on the following input: Some(_)
+  def extractorStrict(x: Option[Int]) = x match {
+                                        ^
+t5365b.scala:27: warning: match may not be exhaustive.
+It would fail on the following input: Some(_)
+  def repeatedExtractorStrict(x: Option[Int]) = x match {
```
retronym, line 27: This warning message is imprecise, it should technically say `Some(x forSome x not in Extractor(x))`.

retronym:
I just found another hole in our exhaustivity checking.

```
class Test {
  def foo {
    case class Foo(a: Any)
    val xs: List[Foo] = Nil
    xs match {
      case Foo(a) :: Nil =>
    }
  }
}
```

The constructor pattern `Foo` gets translated into a call to `unapplySeq`, rather than to a extractor of the case class product elements. This hampers exhaustivity checking (disabled it completely prior to this PR, and forces us to make an assumption even after this PR). It also means that the code is less efficient, and needlessly allocates a tuple.

This root cause of this was fixed recently as https://github.com/scala/scala/pull/4862 (which isn't merged yet to the 2.12.x branch).

durban:
I understand this still doesn't issue a warning if the type of `scrut` is not a sealed trait (unless I'm missing something). That is, code like this still compiles  without a warning:

``` scala
trait Z
final case class Q(i: Int) extends Z
def f(z: Z) = z match {
  case Q(_) =>
}
```

Would it be possible to get a warning in cases like this too? I realize, that this is even more strict than the `-Xstrict-patmat-analysis` in this PR, so maybe behind an `-Xlint` option?

Also, after playing a bit more with this code, the warnings for extractors seem to behave somewhat surprisingly. For example, code like this compiles without any warnings:

``` scala
sealed trait T
final case class C(i: Int) extends T
final case class D(i: Int) extends T

object X {
  def unapply(t: T): Option[Int] = None
}

def g(t: T): Unit = t match {
  case X(_) =>
}
```

However, if I put in another case, thus increasing (or at least not decreasing) the exhaustiveness of the match, I get an (expected) warning:

``` scala
def g(t: T): Unit = t match {
  case X(_) =>
  case C(_) =>
}
```

To be clear, the (to me) surprising thing is the _absence_ of the warning in the first case. Is this behavior intentional?

adriaanm:
Let's take another look at this after M5. Scheduling for RC1.

retronym:
I don't have time to push this one forward right now, so I'm going to close the PR for now.

sellout:
I’d like to pick this up, if it’s ok with @retronym.

My own opinions: it’d be nice to make “false” the default (@adriaanm – how likely is that?). Also, we should warn whenever the checker gives up (which I think would cover @durban’s `X(_)` case?) and I like the -Xlint idea for the unsealed trait case.

adriaanm:
Great! I can be convinced of whatever strategy results in the best experience for the majority of Scala programmers.

src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
```diff
@@ -516,11 +516,15 @@ trait MatchAnalysis extends MatchApproximation {
...
       val start = if (Statistics.canEnable) Statistics.startTimer(patmatAnaExhaust) else null
       var backoff = false
+      val strict = settings.XstrictPatmatAnalysis.value
 
       val approx = new TreeMakersToPropsIgnoreNullChecks(prevBinder)
       val symbolicCases = approx.approximateMatch(cases, approx.onUnknown { tm =>
         approx.fullRewrite.applyOrElse[TreeMaker, Prop](tm, {
           case BodyTreeMaker(_, _) => True // irrelevant -- will be discarded by symbolCase later
+          case GuardTreeMaker(_) => if (strict) False else True
+          case ExtractorTreeMaker(extractor, _, _) if extractor.symbol.name != nme.unapplySeq =>
```
sellout, line 526: @retronym I thought the guard here was because of SI-9567 (fixed in #4862, which seems to have been merged into 2.12.x at this point). However, removing it in order to get the correct `repeatedExtractor` warnings causes other tests to fail (e.g., run/patmatnew.scala:58, which is an exhaustive match on `List[Int]`).

Wondering if you think there’s a middle ground that can be achieved, or if the `repeatedExtractor` warnings are a bit too ambitious (at least for me).

LogicalTime:
We should at least let this be available behind a compiler flag. Without this people will avoid using guards so they can be sure of the exhaustive warning/compile error.

PR #4930, opened 2016-01-29
SI-9634: Avoiding NullPointerException's when making Options from java.lang values

akauppi:
Introducing `AnyVal` seems to have caused a problem in the use of `java.lang` values, together with `Option`s. The current behaviour is somewhat surprising, and there is really no need for having Options of java basic values. 

This PR handles the normal `Option` and `Some` use cases, converting non-null `java.lang` values to an `Option` of corresponding `AnyVal`. This should be what the user wants, though they might not know it. :)

A further idea could be to ban `Option[java.lang.Long]` and others from existing, at all, but my type programming skills were not up to suggesting how that could be done, without adding complexities.

The bigger explanation of this is naturally in the [SI-9634](https://issues.scala-lang.org/browse/SI-9634) ticket. I hope to get feedback on this - even if the PR were not accepted, for some reason.

### DIFF
--- a/src/library/scala/Option.scala
+++ b/src/library/scala/Option.scala
@@ -24,6 +24,22 @@ object Option {
    */
   def apply[A](x: A): Option[A] = if (x == null) None else Some(x)
 
+  /** Option factories for values received from some Java API. Without this,
+    * nulls can trigger NullPointerException, in some scenarios. Also, convert
+    * to appropriate Scala AnyVals that guarantee the inner value won't be null.
+    *
+    *  @param  x the value (java.lang.xxx)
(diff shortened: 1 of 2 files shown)


jozic:
i would def want to have flexibility  of having Option of java.lang.Long (and friends) and not sure i would like Option `magically` convert my java types to scala types

SethTisue:
Let's discuss the advisability of this over on https://issues.scala-lang.org/browse/SI-9634. If it's agreed to be advisable, then we can use the PR to discuss the implementation.

lrytz:
closing for now, discussion on the ticket is still open

PR #4931, opened 2016-01-29
SI-9623 Avoid unnecessary hasNext calls in JoinIterator & ConcatIterator

szeiger:
These iterator implementations are used to concatenate two (JoinIterator)
or more (ConcatIterator) other iterators with `++`. They used to perform
many unnecessary calls to the child iterators’ `hasNext` methods.

This improved state machine-based implementation reduces that number to
the bare minimum, i.e. iterating over concatenated iterators with
`foreach` calls the children's `hasNext` methods a total of (number of
children) + (number of elements) times, the same as when iterating over
all children separately.

### DIFF
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -10,7 +10,7 @@ package scala
 package collection
 
 import mutable.ArrayBuffer
-import scala.annotation.migration
+import scala.annotation.{tailrec, migration}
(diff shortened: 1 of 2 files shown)


szeiger:
Should be binary compatible. I'll retarget to 2.11.

szeiger:
New PR: https://github.com/scala/scala/pull/4937

PR #4933, opened 2016-01-29
when running tests during PR validation, use sbt not ant [ci: last-only]

SethTisue:
this is the next step after #4884


SethTisue:
(not done yet: making it respect `scalac.args.optimise`)

SethTisue:
hello permgen our old friend you've come to talk to us again

```
sbt.ForkMain$ForkError: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Error running /home/jenkins/workspace/scala-2.11.x-validate-test/test/files/scalacheck/Ctrie.scala
Caused by: sbt.ForkMain$ForkError: java.lang.OutOfMemoryError: PermGen space
```

SethTisue:
if I run `test/partest test/files/scalacheck/Ctrie.scala` by itself, locally, it passes even with `-XX:MaxPermSize=48M`, so it seems this isn't an unusually PermGen hungry test. which has me worried scala-partest-interface might be doing something PermGen-unfriendly, perhaps holding on to classloader references between tests.

SethTisue:
increasing PermGen from 128M to 192M didn't help — and failed on the exact same test.

note also `# 111/117 passed, 6 failed in jvm` ("compilation failed" errors)

retronym:
https://github.com/sbt/sbt/issues/1223 might offer some clues. SBT forces GC and finalization periodically if the number of objects pending finalization exceeds a threshold. It appears that this is only done after a task has executed, so it probably won't help if a single task like `tests/it:test` internally leaks.

SethTisue:
The permgen failure also happens when running locally on Mac OS X. It sat a long time before failing; sample stack trace not long before failing: https://gist.github.com/SethTisue/8db45ef617a0542057b2

szeiger:
Strange, I can run partest locally from sbt without problems.

SethTisue:
> sbt/sbt#1223 might offer some clues

I don't think so, because partest runs in a forked JVM.

> Strange, I can run partest locally from sbt without problems

@szeiger on Java 6?

szeiger:
Hah, good point, I probably only ran it on Java 8.

SethTisue:
as I study a log of the old test/validate script a little closer, I see that ant was invoking partest multiple times, once per category of test. whereas in sbt we're doing one single giant partest invocation. so it's actually not surprising this might cause a PermGen regression, and I don't think there's any need to investigate the details.

I'll just change it to match the old setup. the easiest way, as @retronym suggests, is simply not to invoke partest through sbt at all. at some point, we might like to change `test/it:partest` to do the multiple invocations, but for now, we can just bypass sbt-partest-interface altogether. this still accomplishes our main goal which is to get ant out of the picture.

szeiger:
@SethTisue you can run individual categories from sbt, e.g. `test/it:testOnly -- --run`. I hope the syntax is right. I'm on Windows right now and I just noticed that partest in sbt tries to run pull-binary-libs for some reason. I need to look into that.

SethTisue:
> you can run individual categories from sbt

for now, I'm just going to invoke partest directly, but I'm including a comment in that part of that script that says we should consider switching to sbt/sbt-partest-interface later.

SethTisue:
> you can run individual categories from sbt

good to know. ultimately, we should switch to that.

SethTisue:
it didn't fail the job, but:

```
Archiving artifacts
```

SethTisue:
sigh, `test/partest` actually runs `ant -q test.suite.init`, so we're not cutting ant out of the picture by calling `test/partest` unless that's changed, and it also then reads `partest.properties` to construct a classpath, and that classpath uses the ant build directory, not the sbt build directory. so this needs more work.

retronym:
> and it also then reads partest.properties to construct a classpath

The SBT build now generates this file, albeit under `./build-sbt`.

See  https://github.com/scala/scala/commit/2a57f828b2ab911ab70f20a9eb4db1f754a62e47

We'd still need to modify the `partest` script to avoid calling `ant`, perhaps when provided with `--no-ant`.

szeiger:
That explains why it fails on Windows. I don't think partest or sbt-partest-interface should call any script at all. All that suite.init does is to build the classpath for partest in build/pack/partest.properties, which triggers a full ant build and pulling the libraries with the shell script, because all that stuff goes on the classpath. I wonder why I haven't noticed this before. sbt can build a correct partest.properties (but it goes into build-sbt/quick) which works for running partest from the command line. We should pass that to partest from sbt instead of calling ant.

szeiger:
It looks like the problem with partest in sbt is really just https://github.com/scala/scala/blob/9fb85af78034c117c46dcd3d2d6e73f6abd081ec/build.sbt#L536. Remove that line and it runs successfully.

I tested this on a Windows system that doesn't even have ant installed. This includes running partest from the shell (MSYS bash) after sbt dist/mkQuick and cp build-sbt/quick/partest.properties build/pack/. No ant and no JARs in lib required.

SethTisue:
> [...] the problem with partest in sbt [...]

@szeiger sorry, which problem?

SethTisue:
MiMa failure:

```
[info] Checking backward binary compatibility
[error] (run-main-1) java.lang.ArrayIndexOutOfBoundsException: 1477
[error] (run-main-0) java.lang.ArrayIndexOutOfBoundsException: 1477
java.lang.ArrayIndexOutOfBoundsException: 1477
    at com.typesafe.tools.mima.core.BufferReader.nextByte(BufferReader.scala:33)
    at com.typesafe.tools.mima.core.BufferReader.nextInt(BufferReader.scala:53)
    at com.typesafe.tools.mima.core.ClassfileParser$$anonfun$skipAttributes$1.apply$mcVI$sp(ClassfileParser.scala:269)
```

doesn't happen when I run locally. perhaps the 2.11.0 jar it's trying to use isn't a real jar, perhaps because retrieving it failed because validation stuff is running with special resolver settings, that kind of thing. I'll look into it (but not right away).

scripts/jobs/validate/test
```diff
@@ -1,17 +1,46 @@
...
+    runsbt "test/it:testOnly -- pos"
+    runsbt "test/it:testOnly -- neg"
+    runsbt "test/it:testOnly -- run"
+    runsbt "test/it:testOnly -- jvm"
+    runsbt "test/it:testOnly -- res"
+    runsbt "test/it:testOnly -- scalap"
+    runsbt "test/it:testOnly -- specialized"
+    runsbt "test/it:testOnly -- scalacheck"
+    runsbt "test/it:testOnly -- instrumented"
+    runsbt "test/it:testOnly -- presentation"
```
retronym, line 39: TODO (pending a new partest release to honour `--srcpath`)

```
runsbt "test/it:testOnly -- --srcpath scaladoc --run"
runsbt "test/it:testOnly -- --srcpath scaladoc --scalacheck"
```

project/MiMa.scala
```diff
@@ -0,0 +1,81 @@
...
+        def runOnce(prev: java.io.File, curr: java.io.File, isForward: Boolean): Unit = {
+          val direction = if (isForward) "forward" else "backward"
+          log.info(s"Checking $direction binary compatibility")
+          runMima(
+            prev = if (isForward) curr else prev,
+            curr = if (isForward) prev else curr,
+            // TODO: it would be nicer if each subproject had its own whitelist, but for now
+            // for compatibility with how Ant did things, there's just one at the root.
+            // once Ant is gone we'd be free to split it up.
+            filter = baseDirectory.value / ".." / ".." / s"bincompat-$direction.whitelist.conf",
```
retronym: I'd opt for `(baseDirectory in ThisBuild).value / s"bincompat-$direction.whitelist.conf",`

project/MiMa.scala
```diff
@@ -0,0 +1,81 @@
...
+    taskKey[Unit]("run Migration Manager to detect binary incompatibilities")
+
+  lazy val settings =
+    Seq(
+      mimaOldVersion := s"${scalaBinaryVersion.value}.0",
+      mima := {
+        val log = streams.value.log
+        def runOnce(prev: java.io.File, curr: java.io.File, isForward: Boolean): Unit = {
+          val direction = if (isForward) "forward" else "backward"
+          log.info(s"Checking $direction binary compatibility")
```
retronym, line 26: `log.debug` with the file names.

project/MiMa.scala
```diff
@@ -0,0 +1,81 @@
+// It would be nice to use sbt-mima-plugin here, but the plugin is missing
+// at least two features we need:
```
szeiger, line 2: It would be nice to feed this back to MiMa so we can rid of the duplication in the future.

project/MiMa.scala
```diff
@@ -0,0 +1,81 @@
...
+        val log = streams.value.log
+        def runOnce(prev: java.io.File, curr: java.io.File, isForward: Boolean): Unit = {
+          val direction = if (isForward) "forward" else "backward"
+          log.info(s"Checking $direction binary compatibility")
+          runMima(
+            prev = if (isForward) curr else prev,
+            curr = if (isForward) prev else curr,
+            // TODO: it would be nicer if each subproject had its own whitelist, but for now
+            // for compatibility with how Ant did things, there's just one at the root.
+            // once Ant is gone we'd be free to split it up.
```
szeiger, line 33: Once ANT is gone, I'd just put it into build.sbt and get rid of the extra files.

scripts/jobs/validate/test
```diff
@@ -1,17 +1,46 @@
-#!/bin/bash -e
+#!/bin/bash -e -v -x
+
+baseDir=${WORKSPACE-`pwd`}
+scriptsDir="$baseDir/scripts"
+. $scriptsDir/common
+
+function runsbt() {
+  $SBT_CMD \
+     -Dstarr.version=$scalaVersion \
```
szeiger, line 9: We should introduce an sbt SettingKey for starr.version. Once ant is gone we can remove the old versions.properties stuff but the key could already be used now (so you'd call an sbt set command here instead of overriding it through a system property).

SethTisue:
> MiMa failure: [...]

turned out to be transient, so I think the MiMa part of this is done (& I can move on to the other remaining parts)

szeiger:
Closing in favor of https://github.com/scala/scala/pull/5190

PR #4936, opened 2016-02-01
[SI-9638] Size aware flatMap

pfcoperez:
Added `sizedFlatMap` to `TraversableLike` and `Iterator` which provides an implementation for a `flatMap` operation which receives a  `f: (A, Option[Int]) ⇒ GenTraversableOnce[B]` or `f: (A, Int) ⇒ GenTraversableOnce[B]` instead of just ``f: (A) ⇒ GenTraversableOnce[B]`. Where the second `f` parameter were the accumulated result size so far.

Issue: https://issues.scala-lang.org/browse/SI-9638


Ichoran:
Thank you for your interest in improving the Scala library!  Unfortunately, this kind of change is too special-purpose to be included.  Also, it seems that the second argument is `Option[Int]` only because it is not convenient to implement the method in such a way that the size can always be detected, which is an indication that the library isn't doing its job properly.  Further discussion can take place on the ticket that describes the feature request.

PR #4937, opened 2016-02-01
SI-9623 Avoid unnecessary hasNext calls in JoinIterator & ConcatIterator

szeiger:
These iterator implementations are used to concatenate two (JoinIterator)
or more (ConcatIterator) other iterators with `++`. They used to perform
many unnecessary calls to the child iterators’ `hasNext` methods.

This improved state machine-based implementation reduces that number to
the bare minimum, i.e. iterating over concatenated iterators with
`foreach` calls the children's `hasNext` methods a total of (number of
children) + (number of elements) times, the same as when iterating over
all children separately.

### DIFF
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -10,7 +10,7 @@ package scala
 package collection
 
 import mutable.ArrayBuffer
-import scala.annotation.migration
+import scala.annotation.{tailrec, migration}
 import immutable.Stream
 import scala.collection.generic.CanBuildFrom
 import scala.annotation.unchecked.{ uncheckedVariance => uV }
@@ -168,8 +168,10 @@ object Iterator {
   private[scala] final class ConcatIterator[+A](private[this] var current: Iterator[A], initial: Vector[() => Iterator[A]]) extends Iterator[A] {
     @deprecated def this(initial: Vector[() => Iterator[A]]) = this(Iterator.empty, initial) // for binary compatibility
     private[this] var queue: Vector[() => Iterator[A]] = initial
+    private[this] var currentHasNextChecked = false
     // Advance current to the next non-empty iterator
     // current is set to null when all iterators are exhausted
+    @tailrec
     private[this] def advance(): Boolean = {
       if (queue.isEmpty) {
         current = null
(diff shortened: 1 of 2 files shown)


SethTisue:
review by @Ichoran plz

Ichoran:
Well, this is an improvement but the algorithms, especially for ConcatIterator, are still not very efficient.  The primary culprit is the incredibly expensive :+ and head/tail decomposition on a Vector.  This just isn't a good way to switch things rapidly (though at least it scales well).  Furthermore, it checks when adding a long LHS to a short RHS, but `++` doesn't check the RHS to see if maybe that is a `ConcatIterator` too, and it should really merge the lists.

I'm not sure what the goal here is.  Better?  Yes.  Looks correct?  Yes.  Achieved ideal performance?  No.

It does seem to fix the "too many hasNext" problem, which is great!

(But on the other hand one could also argue that iterators with an expensive `hasNext` that don't cache the result are poorly-written iterators.  Why is it the consumer's job to maintain extra state just in case?)

szeiger:
> The primary culprit is the incredibly expensive :+ and head/tail decomposition on a Vector. This just isn't a good way to switch things rapidly (though at least it scales well).

There should be a better way. I'll do some benchmarking to find out.

> Furthermore, it checks when adding a long LHS to a short RHS, but ++ doesn't check the RHS to see if maybe that is a ConcatIterator too, and it should really merge the lists.

Right, ConcatIterator originally came to pass to prevent consuming this iterator from SOEing:

```
(1 to n).foldLeft(start)((res, _) => res ++ Iterator(1))
```

But we have the same situation on the RHS. This still causes a SOE:

```
(1 to n).foldLeft(start)((res, _) => Iterator(1) ++ res)
```

> (But on the other hand one could also argue that iterators with an expensive hasNext that don't cache the result are poorly-written iterators. Why is it the consumer's job to maintain extra state just in case?)

That was my first thought as well, but repeatedly calling `hasNext` on an already exhausted Iterator looks like a rather unexpected call pattern to me. I wouldn't necessarily expect every implementation to be optimized for this case.

szeiger:
Here's a much faster version: https://github.com/scala/scala/commit/07db07113f6576af1097780cc14f8af21c12e32e

It's not binary compatible though. I suggest we use the fix here for 2.11.8 and consider the new implementation for 2.12.

Ichoran:
LGTM

Agreed--use this for 2.11, and the even better version for 2.12.  (The other one also looks good to me, save for one place where it looks like `last` doesn't release its contents.)

PR #4939, opened 2016-02-02
SI-9624 Improve documentation for TraversableOnce

szeiger:
- Move the doc comment for `hasDefiniteSize` up from TraversableLike
  to GenTraversableOnce and improve it.
- Add a note to `GenTraversableOnce.isEmpty` that implementations must
  not consume elements.

Review by @Ichoran

### DIFF
--- a/src/library/scala/collection/GenTraversableOnce.scala
+++ b/src/library/scala/collection/GenTraversableOnce.scala
@@ -67,6 +67,23 @@ trait GenTraversableOnce[+A] extends Any {
     */
   def foreach[U](f: A => U): Unit
 
+  /** Tests whether this $coll is known to have a finite size.
+    *  All strict collections are known to have finite size. For a non-strict
+    *  collection such as `Stream`, the predicate returns `'''true'''` if all
+    *  elements have been computed. It returns `'''false'''` if the stream is
+    *  not yet evaluated to the end. Non-empty Iterators usually return
+    *  `'''false'''` even if they were created from a collection with a known
+    *  finite size.
+    *
+    *  Note: many collection methods will not work on collections of infinite sizes.
+    *  The typical failure mode is an infinite loop. These methods always attempt a
+    *  traversal without checking first that `hasDefiniteSize` returns `'''true'''`.
+    *  However, checking `hasDefiniteSize` can provide an assurance that size is
+    *  well-defined and non-termination is not a concern.
+    *
+    *  @return  `'''true'''` if this collection is known to have finite size,
+    *           `'''false'''` otherwise.
+    */
   def hasDefiniteSize: Boolean
 
   def seq: TraversableOnce[A]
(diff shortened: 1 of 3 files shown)


src/library/scala/collection/GenTraversableOnce.scala
```diff
@@ -80,6 +94,9 @@ trait GenTraversableOnce[+A] extends Any {
   def size: Int
 
   /** Tests whether the $coll is empty.
+    *
+    * Note: Implementations in subclasses that are not repeatedly traversable must take
```
Ichoran: Wouldn't it be better to phrase this as a guarantee to the consumer of the API that `isEmpty` does not traverse a collection or consume any elements (so may freely be used on a TraversableOnce)?

src/library/scala/collection/GenTraversableOnce.scala
```diff
@@ -67,6 +67,20 @@ trait GenTraversableOnce[+A] extends Any {
...
+  /** Tests whether this $coll is known to have a finite size.
+    *  All strict collections are known to have finite size. For a non-strict
+    *  collection such as `Stream`, the predicate returns `'''true'''` if all
+    *  elements have been computed. It returns `'''false'''` if the stream is
+    *  not yet evaluated to the end. Non-empty Iterators usually return
+    *  `'''false'''` even if they were created from a collection with a known
+    *  finite size.
+    *
+    *  Note: many collection methods will not work on collections of infinite sizes
+    *  (but they do not require this method to return `'''true'''`).
```
Ichoran: Do you mean something like (but will work on finite collections whose size is not known)?  Thinking about what it would mean to require this method to return true seems like it requires more mental gymnastics than strictly necessary.  (Assuming that the person thinking about it has never seen this method before.)

src/library/scala/collection/GenTraversableOnce.scala
szeiger: I phrased it like this because the consumer side is already documented in the overridden method in Iterator, the only non-Traversable subclass of TraversableOnce. You only really see the version of the doc in TraversableOnce when you want to subclass it directly.

src/library/scala/collection/GenTraversableOnce.scala
szeiger: It's not about knowing the size but about knowing whether the collection has a finite size. And in particular, it _is_ about calling this specific method. I can't come up with an equally short version that makes it clearer. How about this longer explanation instead?

> Note: many collection methods will not work on collections of infinite sizes. The typical failure mode is an infinite loop. These methods always attempt a traversal without checking first that `hasDefiniteSize` returns `true`. The latter is sufficient for a terminating traversal but it is not required.

src/library/scala/collection/GenTraversableOnce.scala
Ichoran: Okay, fair enough.

src/library/scala/collection/GenTraversableOnce.scala
Ichoran: Yes, that longer explanation looks good.  It clearly and fairly succinctly explains the issues without assuming mastery of the paragraph that was presumably just read a few seconds before.  Maybe the last sentence would be clearer, "However, checking `hasDefiniteSize` can provide an assurance that `size` is well-defined and non-termination is not a concern."  (Because it's not actually the `hasDefiniteSize == true` that _causes_ the termination of the traversal.  It's just a promise.)

szeiger:
Updated accordingly and squashed.

lrytz:
LGTM

PR #4944, opened 2016-02-04
SI-9571 Avoid boxing primitives in string concatenation

lrytz:


lrytz:
@melezov thanks a lot for your patch, very nice work! I refined it a bit and added some tests.

Review by @soc. For the record, [@melezov signed the CLA](https://www.typesafe.com/contribute/cla/scala/check/melezov).

soc:
Maybe we should rename `genStringConcat` to something like `genReferenceConcat` given the name of `genPrimitiveConcat`?

soc:
Apart from that, LGTM.

lrytz:
Thanks for the review, @soc - with fresh eyes I managed to get rid of some duplicate code, now things look better.

src/compiler/scala/tools/nsc/backend/jvm/BCodeIdiomatic.scala
```diff
@@ -197,16 +197,19 @@ abstract class BCodeIdiomatic extends SubComponent {
...
-    final def genStringConcat(el: BType, pos: Position): Unit = {
-      val jtype = el match {
+    def genConcat(elemType: BType, pos: Position): Unit = {
+      val paramType = elemType match {
         case ct: ClassBType if ct.isSubtypeOf(StringRef).get          => StringRef
         case ct: ClassBType if ct.isSubtypeOf(jlStringBufferRef).get  => jlStringBufferRef
         case ct: ClassBType if ct.isSubtypeOf(jlCharSequenceRef).get  => jlCharSequenceRef
-        case rt: RefBType                                             => ObjectRef
-        case pt: PrimitiveBType                                       => pt  // Currently this ends up being boxed in erasure
+        // Don't match for `ArrayBType(CHAR)`, even though StringBuilder has an such an overload
```
esfand-r: sorry :sweat:  really minor  "has an such an overload" -> "has such an overload"

lrytz: fixed, thanks!

lrytz:
@soc does this LGTY now?

soc:
LGTM!

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -999,9 +999,17 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
           bc.genStartConcat(tree.pos)
           for (elem <- concatenations) {
-            val kind = tpeTK(elem)
-            genLoad(elem, kind)
-            bc.genStringConcat(kind, elem.pos)
+            val loadedElem = elem match {
+              case Apply(boxOp, value :: Nil) if currentRun.runDefinitions.isBox(boxOp.symbol) =>
+                // Eliminate boxing of primitive values. Boxing is introduced by erasure because
+                // there's only a single synthetic `+` method "added" to the string class.
+                value
```
retronym, line 1006: This will incorrectly transform:

```
"xx" + {println("side effect"); Int}.box(1)
```

to 

```
val sb = new StringBuilder(); sb.append("xx"); sb.append(1)
```

However, this isn't the only place in the backend that has this bug:

```
scala> {???; Int}.box(1)
res4: Integer = 1
```

We can treat this as a separate bug: https://github.com/scala/scala-dev/issues/85

lrytz, line 1006: nice find..!

PR #4945, opened 2016-02-05
Turn on sam conversion by default [ci: last-only]

adriaanm:
Cleaned up while I was in the area (code & spec). Minimal spec for SAM conversion.
- [ ] setting to disable sam conversion
- [ ] check overloading spec, add test for ToString example to test/files/pos/sammy_overload.scala
- [ ] update comments in test/files/pos/t9449.scala
- [ ] more tests based on https://gist.github.com/lrytz/ec4da087dc65e421eb7d
- [ ] remove mention of <static> from spec

### DIFF
--- a/test/files/pos/t9449.scala
+++ b/test/files/pos/t9449.scala
@@ -1,0 +1,19 @@
+trait II {
+  def apply(x: Int): Int
+}
+
+object Test {
+  def ii(x: Int): Int = x
+  def test = {
+    val ii1: II = x => ii(x) // works
+    val ii2: II = ii         // works (adapting `ii` to `II`)
+    val ii3: II = ii _       // fails -- should work
+                             // typedTyped({ii : (() => <empty>)})
+                             //    typedEta(ii, pt = II)
+                             //       adapt(ii, pt = (? => ?))
+                             //          instantiatedToMethodType(ii, pt = (? => ?))
+    // val ii3: II = ii _ // error:
+    // found   : Int => Int
+    // required: II
+  }
+}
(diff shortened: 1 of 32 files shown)


adriaanm:
Review by @retronym

adriaanm:
Review by @lrytz, bitte.

lrytz:
Should we have a setting to turn the feature off?

test/files/pos/t9449.scala
```diff
@@ -0,0 +1,19 @@
+trait II {
+  def apply(x: Int): Int
+}
+
+object Test {
+  def ii(x: Int): Int = x
+  def test = {
+    val ii1: II = x => ii(x) // works
+    val ii2: II = ii         // works (adapting `ii` to `II`)
+    val ii3: II = ii _       // fails -- should work
```
lrytz, line 10: it's a bit confusing to have the comment saying "fails", as it's actually fixed

lrytz:
Should have some tests for overloading resolution. The following used to compile:

```
scala> abstract class A[T, R]{ def apply(x: T): R }
defined class A

scala> object O {
     |   def m(a: A[Int, Int]) = 0
     |   def m(a: Int => Int) = 1
     | }
defined object O

scala> O.m(x => x)
<console>:15: error: missing parameter type
       O.m(x => x)
           ^
```

I haven't looked at the details in overloading resolution that cause this change. Is this expected? I don't remember if we already talked about it. Maybe we can have a specific warning for when this happens?

lrytz:
here's a few things i tried, can be included in a test: https://gist.github.com/lrytz/ec4da087dc65e421eb7d

adriaanm:
incorporated into #4971

PR #4946, opened 2016-02-05
Scaladoc: add pan & zoom feature to Type Inheritance graphs

felixmulder:
This commit adds the pan & zoom feature requested in #4914 by @VladUreche.

Preview: https://dl.dropboxusercontent.com/u/358427/panzoom-test/index.html#scala.collection.immutable.List

EDIT: before anyone asks, it supports pinch-to-zoom on mobile devices :+1:


VladUreche:
Wooohooo, thanks Felix! :tada:
The buttons look and work great!

I found two issues while happily zooming in and out of the diagrams:
- zooming out below 1:1 seems to work again, and the diagram goes outside the screen :(
- when expanding to full screen, any attempt to pan the diagram makes it jump to the top of the screen.

These are the two issues caught on tape: https://www.dropbox.com/s/3s62ji33ssl6b3q/scaladoc-diagrams.mkv?dl=0

felixmulder:
@VladUreche: updated the PR now, should fix the issue of 1:1 zoom. I'm not sure about your second issue. I tried replicating it on OS X (chrome & FF), couldn't break it. Then on Arch Linux with FF, still can't break it. Maybe I don't have the magic touch? Or maybe the latest changes fixed it?

EDIT: the dropbox link has the latest changes :+1:

VladUreche:
@felixmulder thanks for the fix! Indeed, you fixed both issues with a single commit!
LGTM

VladUreche:
I'll merge the PR as soon as all tests pass.

felixmulder:
Great! I was thinking of having a look at SI-9620 today. As that would resolve the issues with the duplicates in the docs.

VladUreche:
:+1: thanks Felix!

PR #4947, opened 2016-02-05
Scaladoc: fix stray number in codeblocks for firefox

felixmulder:
Fix the stray number in firefox

review: @VladUreche 

doc example with patch: https://dl.dropboxusercontent.com/u/358427/stray-number-test/index.html#scala.collection.immutable.List


VladUreche:
LGTM! Thanks @felixmulder!

PR #4948, opened 2016-02-05
SI-9574 Prevent illegal overrides of members with module types

szeiger:
Commit f32a32b1b33c9d7ccd62467e3e10cb69930023c8 introduced the ability
to override objects with other objects. The exception that allows
these overrides (where the usual subtyping check fails) was applied to
all members whose type is a module class. This is too broad, however,
because it not only applies to members of the form `object foo` but
also `def foo: bar.type` (where `bar` is an `object`).

The fix is to restrict the exception to those cases where both
definitions actually are objects.

Review by @adriaanm

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -88,17 +88,19 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
     if (sym.hasAccessBoundary) "" + sym.privateWithin.name else ""
   )
 
-  def overridesTypeInPrefix(tp1: Type, tp2: Type, prefix: Type): Boolean = (tp1.dealiasWiden, tp2.dealiasWiden) match {
+  def overridesTypeInPrefix(tp1: Type, tp2: Type, prefix: Type, isModuleOverride: Boolean): Boolean = (tp1.dealiasWiden, tp2.dealiasWiden) match {
     case (MethodType(List(), rtp1), NullaryMethodType(rtp2)) =>
       rtp1 <:< rtp2
     case (NullaryMethodType(rtp1), MethodType(List(), rtp2)) =>
       rtp1 <:< rtp2
     case (TypeRef(_, sym, _),  _) if sym.isModuleClass =>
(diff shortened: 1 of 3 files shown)


adriaanm:
Excellent work, @szeiger!

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -520,7 +519,7 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
         }
         def checkOverrideTerm() {
           other.cookJavaRawInfo() // #2454
-          if (!overridesTypeInPrefix(lowType, highType, rootType)) { // 8
+          if (!overridesTypeInPrefix(lowType, highType, rootType, low.isModuleOrModuleClass || high.isModuleOrModuleClass)) { // 8
```
adriaanm: Could you include a comment with the reasoning of grouping `low` || `high` together as far as them being a module? It would be good to extend test coverage to all possible combos (module overridden by val of module type, val of module type by module, module overridden by val of non-module type, val of non-module type by module).

szeiger: Only in the special case where a module type is replaced by another (by overriding an object with another object with a conforming type) does the subtyping exception make sense. You couldn't replace a val with a singleton type by an object because the object would not also replace the singleton type.

Overriding an object by a val is possible but doesn't require any subtyping exception:

```
scala> class Foo { object X { def y: Int = 42 }; def x = X }
defined class Foo

scala> class Bar extends Foo { override val X: super.X.type = x }
defined class Bar

scala> (new Bar).X
res1: Bar#X.type = null
```

Since https://github.com/scala/scala/commit/f32a32b1b33c9d7ccd62467e3e10cb69930023c8 requires overriding vals to be non-lazy, this doesn't look very useful. At first glance, also allowing lazy vals to override objects may be sound but wouldn't add much utility.

I'll add some more test cases for these situations.

szeiger: D'oh! This needs to be `low && high` of course, not `low || high`. Those extra tests are already paying off...

adriaanm:
LGTM!

PR #4949, opened 2016-02-06
Scaladoc: fix duplicate implicits in Array

felixmulder:
This commit fixes the duplicate methods in Array in Scaladoc. The new `scala.Predef._*ArrayOps` added the methods originally supressed in `src/scaladoc/scala/tools/nsc/doc/Settings.scala`

This fix will be improved when SI-9620 is resolved, which will remove the need for hardcoded ignores.

review: @VladUreche


felixmulder:
Despite finding this, I still think it is a good idea to implement SI-9620!

VladUreche:
LGTM! That's an easy way to fix the problem... Of course, if someone renames them, they will come back. So a complete fix for SI-9620 would be greatly appreciated :)

By the way, the fix for SI-9620 should be based on an annotation instead of a doc comment tag, since doc comments are lost in the class file. Or, even better, it would be good to mask implicit conversions at the client (e.g. `Array` should `@ignoreImplicitConversion Predef._arrayXXX`). This way, a client may keep implicit conversions while another would drop them. WDYT?

felixmulder:
My thoughts exactly :)

Good to know about the annotation, I'll dive into the code now :smile:

VladUreche:
Thanks Felix! :+1:

PR #4950, opened 2016-02-06
SBT build improvements

retronym:
Usability improvements: aliases and tab completion for partest.

Review by @szeiger

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -200,7 +200,18 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   },
   // Remove auto-generated manifest attributes
   packageOptions in Compile in packageBin := Seq.empty,
-  packageOptions in Compile in packageSrc := Seq.empty
+  packageOptions in Compile in packageSrc := Seq.empty,
+
+  // Lets us CTRL-C partest without exiting SBT entirely
+  cancelable in Global := true,
+  // When we fork subprocesses, use the base directory as the working directory.
+  // This enables `sbt> partest test/files/run/t1.scala` or `sbt> scalac sandbox/test.scala`
+  baseDirectory in Compile := (baseDirectory in ThisBuild).value,
+  baseDirectory in Test := (baseDirectory in ThisBuild).value,
+
+  // Don't log process output (e.g. of forked `compiler/runMain ...Main`), just pass it
+  // directly to stdout
+  outputStrategy in run := Some(StdoutOutput)
 )
 
 /** Extra post-processing for the published POM files. These are needed to create POMs that
@@ -418,7 +429,6 @@ lazy val repl = configureAsSubproject(project)
   .settings(
     connectInput in run := true,
     publishArtifact := false,
-    outputStrategy in run := Some(StdoutOutput),
     run <<= (run in Compile).partialInput(" -usejavacp") // Automatically add this so that `repl/run` works without additional arguments.
   )
   .dependsOn(compiler, interactive)
@@ -465,7 +475,8 @@ lazy val replJlineEmbedded = Project("repl-jline-embedded", file(".") / "target"
       val outdir = (classDirectory in Compile).value
       JarJar(inputs, outdir, config)
     }),
-    publishArtifact := false
+    publishArtifact := false,
+    connectInput in run := true
   )
   .dependsOn(replJline)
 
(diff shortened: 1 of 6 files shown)


dwijnand:
I don't know sbt's parsing infrastructure, but I'd look at how scripted is coded and/or ask @eed3si9n.

I'll try and have a look at this this weekend though.

retronym:
Thanks. Don't go too far down the rabbit hole in my behalf... :)

I took a look at the scripted parser as inspiration but didn't figure it out.

retronym:
Just added a commit to mimic what the IntelliJ SBT import would do (were it able to handle our build), so that we can get IDE smarts when editing the build definition.

retronym:
> Thanks. Don't go too far down the rabbit hole in my behalf... :)

Or, if rabbit holes are your thing, you might like to write an option parser for the `{scalac, scala, scaladoc}` commands in the same manner as https://github.com/scala/scala-tool-support/blob/master/bash-completion/src/main/scala/BashCompletion.scala

First version would be a static parser, based on the current set of options that enumerate by running the code above against 2.11.7. The extra credit version would fire up the just-built compiler to get the options-du-jour. That would need to be done via reflection, of course (in the same way that `runMain` works).

dwijnand:
Yeah I've been looking at this since yesterday.. and I'm throwing in the towel, sorry. It's an area I've just skim read in the docs and in the sources, not anything I'm well-versed on.

retronym:
I've just pushed an additional commit with my first attempt at the Scala option parser. There are a few TODOs in there, but I'd like to page this out for a while and get back to my regularly scheduled job of trait surgery, so I'm hoping we can book the progress so far and address these later on.

build.sbt
```diff
@@ -200,7 +200,18 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   },
   // Remove auto-generated manifest attributes
   packageOptions in Compile in packageBin := Seq.empty,
-  packageOptions in Compile in packageSrc := Seq.empty
+  packageOptions in Compile in packageSrc := Seq.empty,
+
+  // Lets us CTRL-C partest without exiting SBT entirely
+  cancelable in Global := true,
```
szeiger, line 206: You may want to reference https://github.com/sbt/sbt/issues/1442 here. Seeing the brute force `in Global` in a subproject definition made me wonder why you need it, leading to said sbt ticket.

szeiger:
Awesome improvements. I can't really offer any advise on the sbt parser combinator code. I learned the basics only a few weeks ago when working on the Scala build (and then abandoned the solution that would have required me to write parsers due to other complications).

retronym:
A neat trick built on top of this is to use continuous execution in SBT like:

```
// local.sbt (.gitignored)
watchSources ++= (((baseDirectory in ThisBuild).value / "sandbox") ** ("*.java" | "*.scala")).get
```

```
> ~ ; scalac sandbox/test.scala; scala Test
```

That will retrigger the continuous execution when you change the compiler or when you change the `test.scala` file.

eed3si9n:
This reminds me of https://github.com/mentat-labs/sbt-nsa, which tries to strongly type scalac options.

retronym:
> Could be risky if you have whitespace or escaped characters in the command line and partestParser doesn't agree with sbt's own parser on tokenization.

Yep, that's definitely a hole. Passing a single command string from our custom task to SBT's "remainingCommands" is super easy, but it would be more principled for us to delegate to the internals of `runMain` and `testOnly` programatically. However it seems we might have to copy/paste a bit of code from those tasks to make that happen.

```
  def runMainTask(classpath: Initialize[Task[Classpath]], scalaRun: Initialize[Task[ScalaRun]]): Initialize[InputTask[Unit]] =
    {
      import DefaultParsers._
      val parser = loadForParser(discoveredMainClasses)((s, names) => runMainParser(s, names getOrElse Nil))
      Def.inputTask {
        val (mainClass, args) = parser.parsed
        toError(scalaRun.value.run(mainClass, data(classpath.value), args, streams.value.log))
      }
    }
```

```
def inputTests(key: InputKey[_]): Initialize[InputTask[Unit]] = inputTests0.mapReferenced(Def.mapScope(_ in key.key))
  private[this] lazy val inputTests0: Initialize[InputTask[Unit]] =
    {
      val parser = loadForParser(definedTestNames)((s, i) => testOnlyParser(s, i getOrElse Nil))
      Def.inputTaskDyn {
        val (selected, frameworkOptions) = parser.parsed
        val s = streams.value
        val filter = testFilter.value
        val config = testExecution.value

        implicit val display = Project.showContextKey(state.value)
        val modifiedOpts = Tests.Filters(filter(selected)) +: Tests.Argument(frameworkOptions: _*) +: config.options
        val newConfig = config.copy(options = modifiedOpts)
        val output = allTestGroupsTask(s, loadedTestFrameworks.value, testLoader.value, testGrouping.value, newConfig, fullClasspath.value, javaHome.value, testForkedParallel.value)
        val taskName = display(resolvedScoped.value)
        val trl = testResultLogger.value
        val processed = output.map(out => trl.run(s.log, out, taskName))
        Def.value(processed)
      }
    }
```

szeiger:
Ah, and half the stuff you'd call from the duplicated code is probably private.

SethTisue:
further improvements are always possible, but there isn't any reason not to go ahead and merge this, is there?

retronym:
I think this is ready to go. We might reconsider the scoping of the `baseDirectory in {Compile, Test}` setting (e.g. I could temporarily set that in the custom commands, rather than set it in the build proper), but let's wait until evidence that something breaks.

Separately, we should release a new version of partest (I merged my change to honour `--srcpath`) and update `versions.properties`), and then merge all of this to 2.12.x.

PR #4951, opened 2016-02-07
SI-9632: Close file handles correctly in FileZipArchive

hamishdickson:
Prior to this change, ZipFileArchive was not releasing the streams to files it had opened. This can be a particularly bad problem on Windows as the OS locks the file and then won't allow updates or deletion

This was found and reported by @fommil in the ENSIME project. He has since added this fix into ENSIME as a monkey patch and while there are other issues he is still investigating, this fix resolves his original issue. (see https://issues.scala-lang.org/browse/SI-9632 for more details)

I think @retronym is probably the best person to review this


fommil:
:+1: This is primarily a Windows problem, so I hope the tests are running in a Windows environment? (although we have had ENSIME users complain that, previously, the `scala-compiler` was eating up their file handler allocation on OS X).

fommil:
for posterity, this is where the ensime monkeys live https://github.com/ensime/ensime-server/tree/master/monkeys/src/main :monkey:

hamishdickson:
@fommil Ah yes, asserting delete actually works would be a good test. It appears I've messed up the build so I'll add that test in and then redo this pull request (it's my first PR to Scala...)

Yes, I ran the tests on Windows 8 (I don't have 10 anywhere, but I don't think they changed the file system?) and they work on there :smile: 

> although we have had ENSIME users complain that, previously, the scala-compiler was eating up their file handler allocation on OS X

oh I've never seen that before

fommil:
re: OS X, yeah and I can see why it would happen. If you have, say, 10 open projects, and ENSIME running on all of them, each with a few hundred jars, then that is a lot of file handles just being eaten up. OS X has a hard limit, so in combination with `sbt`'s `scala-compiler` instances effectively spiking that during runs (plus, potentially stale compiler instances that haven't been garbage collected yet) you can easily get to 10,000 file handles. Of course, with this fix, the handlers go to zero.

I'd recommend doing a performance test on this patch. It's not a big deal for ENSIME, but I can imagine the need to open / close zip files having an impact on the compile runs. Apache VFS caches the files, which is maybe something you might want to look at, with a time limit cut off for cache eviction / file close (which is better than what we have today, which is at the whim of the JVM running finalizers)

hamishdickson:
Ah yes, that does make sense.

Also, very good point about opening/closing files maybe causing a performance hit. I've not done any performance testing, so that's something I'll 100% need to check out - I don't want to be "that guy that made compile times even worse" :grimacing:

hamishdickson:
I'm going to close this PR for now. There is a minor build issue and I'd like to do some performance testing

PR #4952, opened 2016-02-08
SI-9620: add doc annotation to hide specific conversions

felixmulder:
This commit will introduce the doc annotation `@hideImplicitConversion`.
By specifying which conversions to hide, the user can "toggle" which
conversions are shown be default.

This implementation is a better workaround than hardcoding which ones to
ignore when running scaladoc.

An example of how to use this:

``` scala
package a

trait Foo[S] {
  def foo(t: S): Int = 123
}

/** Boo with only one foo method, hopefully!
 * @hideImplicitConversion BooShouldNotAppearIsFoo
 */
trait Boo[T]

object Boo {
  sealed trait ShouldNotAppear
  implicit class BooShouldNotAppearIsFoo(boo: Boo[ShouldNotAppear]) extends Foo[ShouldNotAppear]
  implicit class BooLongIsFoo(boo: Boo[Long]) extends Foo[Long]
}
```

Review: @VladUreche


felixmulder:
All right - updated with your input @VladUreche

VladUreche:
Thanks for the quick update Felix and sorry for my belated answers.

Thinking more about this, I'm not sure it will eliminate the `Array` problem. The reason is that it hides the implicit conversions at the very last moment, when the html is generated. But the JSON is still generated with the multiple members.

At the moment, I guess your previous PR which amended the hardcoded object masks this problem. But I expect once you remove the hardcoding, you will see the duplicate methods in search again.

So, the way I think this should work is to prune the implicit conversions directly in `ModelFactoryImplicitSupport`, before any implicitly added members are created for the current entity. WDYT?

felixmulder:
Damnit, you're right!

Regarding pruning in `ModelFactoryImplicitSupport`, this will remove the entries completely. Are we fine with that? I mean - the current solution in this PR hides them from view - pruning them in `ModelFactoryImplicitSupport` will completely remove them.

So IMO there are two options:
- Prune in `ModelFactoryImplicitSupport` (they won't show up)
- Keep the current implementation, but also prune in `IndexScript.scala` where the JSON is generated (they won't show up in search, but will be toggle-able on the entity page)

WDYT?

VladUreche:
Yes, both solutions do the trick. Though I'm more in favor of the `ModelFactoryImplicitSupport` one, which looks more principled.

felixmulder:
All right - then I'll go for # 1 :+1:

felixmulder:
At first I thought I couldn't get the comment from the `ModelFactoryImplicitSupport` class - then I noticed the `thisFactory` variable and the party was back on :tada:

This should be along the lines of what you intended for the implementation @VladUreche

felixmulder:
Pulled changes from upstream into my branch - once it's been through Jenkins it should be good to go.

VladUreche:
@felixmulder great! Can you please add a test case on this?
You can use this as a template: https://github.com/scala/scala/blob/2.12.x/test/scaladoc/run/SI-3448.scala

VladUreche:
Maybe a good think to test is that adding this annotation to `Array` and un-hardcoding the implicit conversions to be ignored produces the right scaladoc output.

felixmulder:
I've added the test and I'll look at Array now. Will ping you once I know that everything's working :smile:

VladUreche:
:+1:

felixmulder:
Ok - so this is where I'm at. I've added the annotation to the Array class. Removed the hardcoded ignores from the `Settings.scala` file in scaladoc. I've created a test for SI-9620.

It generates the Array documentation without the implicits, huzzah! :tada:

HOWEVER! I can't get my stupid test to work :cry: 

@VladUreche

test/scaladoc/run/SI-9620.scala
```diff
@@ -0,0 +1,43 @@
...
+  // no need for special settings
+  def scaladocSettings = ""
+
+  def testModel(rootPackage: Package) = {
+    import access._
+
+    // Assert Boo only has one implicit conversion
+    val boo = rootPackage._package("a")._trait("Boo")
+    val conversions = boo._conversions("a.Boo.BooShouldNotAppearIsFoo") ++ boo._conversions("a.Boo.BooLongIsFoo")
+    assert(conversions.length == 1, conversions.length + " == 1")
```
felixmulder, line 33: @VladUreche: the length here is 0 for some reason. I also checked `boo.conversions.length` which is 0.

test/scaladoc/run/SI-9620.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    object Boo {
+      sealed trait ShouldNotAppear
+      implicit class BooShouldNotAppearIsFoo(boo: Boo[ShouldNotAppear]) extends Foo[ShouldNotAppear]
+      implicit class BooLongIsFoo(boo: Boo[Long]) extends Foo[Long]
+    }
+  """
+
+  // no need for special settings
+  def scaladocSettings = ""
```
VladUreche: @felixmulder add `-implicits` here, otherwise the test won't perform implicit inference.

test/scaladoc/run/SI-9620.scala
VladUreche, line 33: I think it's probably the fact that `-implicits` is not set in the scaladoc flags.

VladUreche, line 33: Yep, most likely. Here's a test for implicits:
https://github.com/scala/scala/blob/2.12.x/test/scaladoc/run/implicits-known-type-classes.scala#L11

felixmulder:
@VladUreche thanks so much! It works :smile: Thanks a bunch for the help!

Squished and commited!

VladUreche:
LGTM, thanks a lot for your tireless work @felixmulder!
Ping @janekdb: can you please add the `hideImplicitConversion` to the scaladoc tutorial?

janekdb:
@VladUreche : Added to pending list on https://github.com/scala/scala-lang/pull/394

VladUreche:
:+1:

PR #4955, opened 2016-02-09
SI-9650 Refchecks on case apply transform

som-snytt:
Apply checks for unsavoriness when transforming
a case apply.

This is narrower in scope than SI-8685. It just makes sure check is applied for case apply, including prefix of qual.

### DIFF
--- a/test/files/neg/t8685.scala
+++ b/test/files/neg/t8685.scala
@@ -1,0 +1,54 @@
+
+
+@deprecated("class C is depr", since="now")
+case class C(i: Int)
+
+case class D @deprecated("ctor D is depr", since="now") (i: Int)
+
+@deprecated("class E is depr", since="now")
+case class E(i: Int)
+@deprecated("module E is depr", since="now")
+object E
+
+@deprecated("module F is depr", since="now")
+object F {
+  case class G(i: Int)
+}
+
+object G {
+  case class H(i: Int)
+}
+
+object Extra {
+  @deprecated("Extra module F is depr", since="now")
+  object F {
+    case class G(i: Int)
+  }
+}
+
+object J {
+  @deprecated("Inner K is depr", since="now")
+  case class K(i: Int)
+}
+
+trait Applies {
+  def f = C(42)
+  def g = D(42)
+  def h = E(42)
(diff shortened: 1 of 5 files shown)


som-snytt:
Updates the `test.check` for `compileTimeOnly`, which is triggered similarly to deprecation.

test/files/neg/t8685.scala
```diff
@@ -0,0 +1,29 @@
...
+@deprecated("module F is depr", since="now")
+object F {
+  case class G(i: Int)
+}
+
+trait Applies {
+  def f = C(42)
+  def g = D(42)
+  def h = E(42)
+  def i = F.G(42)
```
retronym, line 38: How about:

```
object G { case class H(i: Int) }
trait Applies {
  @deprecated val gg = G
  this.gg.G(0)
}
```

It's not clear to me whether the `Select(... gg)` will sneak through the net.

som-snytt, line 38: Evil retronym!

lrytz:
LGTM, very nice work!

test/files/neg/t8685.scala
retronym, line 38: Don't mean to drip feed the critique, but when I noticed the match on `Select | Ident`, this variation came to mind. Seems like a general bug in refchecks:

```
scala> trait T { @deprecated("", "") type U = Int }
defined trait T

scala> type V = T#U
defined type alias V
```

Should be fixed by generalising the match to `RefTree`.

som-snytt, line 38: Actually I was drip-feeding the fix. I realized that it's difficult/impossible as a casual contributor to visualize the pattern matching space for trees -- there's a comment somewhere that enumerates the cases -- but it would be handy if one could ask, What shapes of trees can I expect here? I didn't have time to research yesterday b/c day job. But I bet a tool could automate that (rather than debug lines).

som-snytt, line 38: BTW, the current code does have calls to `checkUndesired` with the match on `RefTree`, so I could have pushed the ball over the line. hashtag americanfootball

PR #4956, opened 2016-02-10
Fix SBT tab completion of scala command

retronym:
`scala -deprecation` (without a trailing script argument)
wasn't allowed. Now it is.

I also supported trailing whitespace on all commands.


retronym:
Review by @SethTisue

retronym:
Found a bug, will be back soon...

SethTisue:
LGTM

PR #4799, opened 2015-10-10
Use jarlister in build

rjolly:
The goal of this change is to exercize the "manifest classpath" mechanism,
meant to bring the compiler its needed classes as resources, listed
in jar manifests, as opposed to files, thus enabling to use the compiler
in sandboxed environments (and also the scripting engine for that matter).

This is a follow-up to https://github.com/scala/scala/pull/4456 . For the build to succeed, the jarlister should be added to the bintray under its sha1. It is available from http://repo.maven.apache.org/maven2/com/github/rjolly/jarlister_2.11/1.0/jarlister_2.11-1.0.jar

PR #4959, opened 2016-02-11
Use jarlister in build

rjolly:
The goal of this change is to exercize the "manifest classpath" mechanism,
meant to bring the compiler its needed classes as resources, listed
in jar manifests, as opposed to files, thus enabling to use the compiler
in sandboxed environments (and also the scripting engine for that matter).

This is a follow-up to #4799 which was closed prematurely.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -351,6 +351,7 @@ lazy val library = configureAsSubproject(project)
     products in Compile in packageBin ++=
       (products in Compile in packageBin in forkjoin).value,
     Osgi.headers += "Import-Package" -> "sun.misc;resolution:=optional, *",
+    Osgi.jarlist := true,
     fixPom(
       "/project/name" -> <name>Scala Library</name>,
       "/project/description" -> <description>Standard library for the Scala Programming Language</description>,
@@ -420,13 +421,15 @@ lazy val compiler = configureAsSubproject(project)
     scalacOptions in Compile in doc ++= Seq(
       "-doc-root-content", (sourceDirectory in Compile).value + "/rootdoc.txt"
     ),
-    Osgi.headers +=
+    Osgi.headers ++= Seq(
       "Import-Package" -> ("jline.*;resolution:=optional," +
                            "org.apache.tools.ant.*;resolution:=optional," +
                            "scala.util.parsing.*;version=\"${range;[====,====];"+versionNumber("scala-parser-combinators")+"}\";resolution:=optional," +
                            "scala.xml.*;version=\"${range;[====,====];"+versionNumber("scala-xml")+"}\";resolution:=optional," +
                            "scala.*;version=\"${range;[==,=+);${ver}}\"," +
                            "*"),
+      "Class-Path" -> "scala-reflect.jar scala-library.jar"
+    ),
     // Generate the ScriptEngineFactory service definition. The ant build does this when building
     // the JAR but sbt has no support for it and it is easier to do as a resource generator:
     generateServiceProviderResources("javax.script.ScriptEngineFactory" -> "scala.tools.nsc.interpreter.IMain$Factory"),
--- a/build.xml
+++ b/build.xml
@@ -279,6 +279,10 @@ TODO:
         <dependency groupId="com.googlecode.jarjar" artifactId="jarjar" version="1.3"/>
       </artifact:dependencies>
 
+      <artifact:dependencies pathId="jarlister.classpath">
+        <dependency groupId="com.github.rjolly" artifactId="jarlister_2.11" version="1.0"/>
+      </artifact:dependencies>
+
       <!-- JUnit -->
       <property name="junit.version" value="4.11"/>
       <artifact:dependencies pathId="junit.classpath" filesetId="junit.fileset">
@@ -867,6 +871,11 @@ TODO:
       <path refid="aux.libs"/>
     </path>
 
+    <path id="pack.lib.path">
+      <pathelement location="${library.jar}"/>
+      <path refid="jarlister.classpath"/>
+    </path>
+
     <path id="pack.bin.tool.path">
       <pathelement location="${library.jar}"/>
       <pathelement location="${xml.jar}"/>
@@ -1230,7 +1239,10 @@ TODO:
 <!-- ===========================================================================
                                   PACKED QUICK BUILD (PACK)
 ============================================================================ -->
-  <target name="pack.lib"    depends="quick.lib, forkjoin.done"> <staged-pack project="library"/></target>
+  <target name="pack.lib" depends="quick.lib, forkjoin.done"> <staged-pack project="library"/>
+    <taskdef resource="scala/tools/ant/antlib.xml" classpathref="pack.lib.path"/>
+    <jarlister file="${library.jar}"/>
+  </target>
 
   <target name="pack.reflect" depends="quick.reflect"> <staged-pack project="reflect"/> </target>
 
--- a/project/Osgi.scala
+++ b/project/Osgi.scala
@@ -1,6 +1,7 @@
 import aQute.lib.osgi.Builder
 import aQute.lib.osgi.Constants._
 import java.util.Properties
+import java.util.jar.Attributes
 import sbt._
 import sbt.Keys._
 import scala.collection.JavaConversions._
@@ -16,6 +17,7 @@ object Osgi {
   val bundleName = SettingKey[String]("osgiBundleName", "The Bundle-Name for the manifest.")
   val bundleSymbolicName = SettingKey[String]("osgiBundleSymbolicName", "The Bundle-SymbolicName for the manifest.")
   val headers = SettingKey[Seq[(String, String)]]("osgiHeaders", "Headers and processing instructions for BND.")
+  val jarlist = SettingKey[Boolean]("osgiJarlist", "List classes in manifest.")
 
   def settings: Seq[Setting[_]] = Seq(
     bundleName := description.value,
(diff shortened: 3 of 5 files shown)


adriaanm:
I'll let @retronym take a look as well, since he's been involved over the whole course of this work, but the ant part LGTM -- thanks for figuring it out! /cc @szeiger / @sethtisue for sbt build follow up...

szeiger:
I'm inclined to leave this one to @SethTisue who's been involved in the previous PRs (unless he would like to pass it on to me). The combination of this PR with the upcoming validate-test changes will break. PR validation still succeeds at the moment because validate-test rebuilds everything with ant (using the sbt-built artifacts from validate-publish-core as STARR).

I see that the previous decision was to have a separate Jarlister tool and use that from Scala's build. I think a closer integration would make sense for the sbt build because we're already doing the JAR writing on our own (in `Osgi.bundleTask`). It may be possible to patch the JAR with only a few lines of code before it's written to disk.

SethTisue:
the integrate-ide failure is unrelated; I reported it to the Scala IDE folks (they're having some trouble with download.scala-ide.org)

build.sbt
```diff
@@ -321,6 +321,18 @@ lazy val library = configureAsSubproject(project)
     products in Compile in packageBin ++=
       (products in Compile in packageBin in forkjoin).value,
     Osgi.headers += "Import-Package" -> "sun.misc;resolution:=optional, *",
+    packageOptions in (Compile, packageBin) +=  {
```
szeiger: This won't have any effect because it modifies the default packaging options but `scala-library.jar` gets built with BND, not with sbt's built-in packaging task. The right JAR can be built by `library/packageBin::packagedArtifact`, you can run `publishDists` to build a full distribution (with output similar to the ant build's). The correct place to add this should be `Osgi.bundleTask` (perhaps as a new setting or by honoring `packageOption` there, so it can be turned on specifically for `scala-library.jar` without affecting other projects).

rjolly:
This change still allows the test to be exercised because it is performed before BND. So I think it is worth merging, and I will make another PR to do what you suggest. What do you think ?

szeiger:
Actually, the tests pose a bigger problem. They do not work in this version when run from sbt:

```
> partest test/files/run/t7843-jsr223-service.scala
[...]
Partest version:
Compiler under test: $baseDir/compiler
Scala version is:    Scala compiler version 2.11.9-20160519-165401-80d46f8 -- Copyright 2002-2016, LAMP/EPFL
Scalac options are:
Compilation Path:    /Users/szeiger/code/scala/target/test/it-classes:$baseDir/test:$baseDir/compiler:$baseDir/library:/Users/szeiger/code/scala/build-sbt/libs/classes/forkjoin:$baseDir/reflect:$baseDir/interactive:$baseDir/actors:$baseDir/repl-jline-embedded:$baseDir/repl-jline:$baseDir/repl:$baseDir/scalap:$baseDir/partest-extras:/Users/szeiger/code/scala/build-sbt/pack/lib/scala-partest-javaagent.jar:$baseDir/scaladoc:$sourceDir/lib/annotations.jar:$sourceDir/lib/enums.jar:$sourceDir/lib/genericNest.jar:$sourceDir/lib/jsoup-1.3.1.jar:$sourceDir/lib/macro210.jar:$sourceDir/lib/methvsfield.jar:$sourceDir/lib/nest.jar:/Users/szeiger/code/scala/test/lib:/Users/szeiger/.ivy2/cache/org.apache.ant/ant/jars/ant-1.9.4.jar:/Users/szeiger/.ivy2/cache/org.apache.ant/ant-launcher/jars/ant-launcher-1.9.4.jar:/Users/szeiger/.ivy2/cache/org.scala-lang.modules/scala-asm/bundles/scala-asm-5.0.4-scala-3.jar:/Users/szeiger/.ivy2/cache/org.scala-lang.modules/scala-xml_2.11/bundles/scala-xml_2.11-1.0.4.jar:/Users/szeiger/.ivy2/cache/org.scala-lang.modules/scala-parser-combinators_2.11/bundles/scala-parser-combinators_2.11-1.0.4.jar:/Users/szeiger/.ivy2/cache/jline/jline/jars/jline-2.12.1.jar:/Users/szeiger/.ivy2/cache/org.scala-lang.modules/scala-partest_2.11/jars/scala-partest_2.11-1.0.13.jar:/Users/szeiger/.ivy2/cache/com.googlecode.java-diff-utils/diffutils/jars/diffutils-1.3.0.jar:/Users/szeiger/.ivy2/cache/org.scala-sbt/test-interface/jars/test-interface-1.0.jar:/Users/szeiger/.ivy2/cache/org.scalacheck/scalacheck_2.11/jars/scalacheck_2.11-1.11.6.jar:/Users/szeiger/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-agent-0.13.11.jar:/Users/szeiger/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-interface-1.0.jar
Java binaries in:    /Library/Java/JavaVirtualMachines/jdk1.8.0_60.jdk/Contents/Home/jre/bin
Java runtime is:     Java HotSpot(TM) 64-Bit Server VM (build 25.60-b23, mixed mode)
Java options are:    -Xmx1024M -Xms64M -XX:MaxPermSize=128M
baseDir:             /Users/szeiger/code/scala/build-sbt/quick/classes
sourceDir:           /Users/szeiger/code/scala/test/files

Selected 1 tests drawn from specified tests

# starting 1 test in run
!! 1 - run/t7843-jsr223-service.scala            [non-zero exit code]
# 0/1 passed, 1 failed in run

partest --update-check \
  /Users/szeiger/code/scala/test/files/run/t7843-jsr223-service.scala

[error] Failed: Total 1, Failed 1, Errors 0, Passed 0
[error] Failed tests:
[error]     partest
[error] (test/it:testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 27 s, completed May 20, 2016 2:47:47 PM
```

Let's take a step back and look at the ant build:
- Publishing is done by building JARs first, repackaging with BND, and repackaging again with jarlister
- The test job uses the previously published "starr", rebuilds the JARs for "pack" in the same way (ant, BND, jarlister)
- It then runs partest on this new "pack"

This should work fine with the changes to the ant build in this PR.

Here's how it will work for sbt:
- Publishing is done directly from individual files with BND. sbt's internal packaging tasks are not used. jarlister postprocessing has to be done on the JARs produced by BND.
- The test job uses the previously published "starr", rebuilds it as "quick" but does not run any packaging to get "pack"
- It then runs partest directly on the new "quick"

The "starr" publishing job has already been switched over to sbt, so its JAR artifacts to not get the new Manifest entries. You do not notice this yet because the test job still runs on ant and publishes a new "pack" for partest. The final publishing of the release artifacts is also done in the same way.

We're close to being able to switch everything over to sbt though and we want to do this before 2.12.0 gets released. So if you want the proper artifacts in 2.12.0, it has to be done in the sbt build.

The first step for jarlister is to hook into our Osgi publishing (as I described in a previous comment).

Making the tests work could be more difficult. As shown above they do not work in this PR and even when you hook into the Osgi publishing in the correct way, they still won't work. The reason for this, as mentioned above, is that sbt runs partest from "quick" instead of "pack", so no packaging takes place at all.

One solution would be to go back to running partest from "pack" but I'd be hesitant to do this just for the sake of these two tests here. Running it from "quick" makes for much faster turnaround times.

The best solution I can think of is adding a new test project (ideally based on junit, not partest) for "pack"-based tests. This would have to be run on CI and could be run manually but wouldn't be part of a normal compile/test cycle during development. It shouldn't be too hard to add this to the sbt build but adding it also to ant would be waste of time at this point. I suggest we add the tests as disabled for now (or disable them when switching to sbt) and add the new "pack"-based testing to sbt later (after ant is gone).

rjolly:
Stefan, it sounds good to me. I have moved the change to bundleTask.

build.sbt
```diff
@@ -420,13 +421,15 @@ lazy val compiler = configureAsSubproject(project)
...
     ),
-    Osgi.headers +=
+    Osgi.headers ++= Seq(
       "Import-Package" -> ("jline.*;resolution:=optional," +
                            "org.apache.tools.ant.*;resolution:=optional," +
                            "scala.util.parsing.*;version=\"${range;[====,====];"+versionNumber("scala-parser-combinators")+"}\";resolution:=optional," +
                            "scala.xml.*;version=\"${range;[====,====];"+versionNumber("scala-xml")+"}\";resolution:=optional," +
                            "scala.*;version=\"${range;[==,=+);${ver}}\"," +
                            "*"),
+      "Class-Path" -> "scala-reflect.jar scala-library.jar"
```
szeiger, line 431: What is this for? I can't find any documentation for `Class-Path` in the BND manual and it does not make it into the `MANIFEST.MF`.

rjolly: It is a standard manifest attribute, note BND. It goes in scala-compiler, not scala-library. This is to be able to type a simple "jrunscript -classpath scala-compiler.jar -l scala" to launch the script engine. This is in par with what the ant build does.

rjolly, line 431: It is a standard manifest attribute, note BND. It goes in scala-compiler, not scala-library. This is to be able to type a simple "jrunscript -classpath scala-compiler.jar -l scala" to launch the script engine. It is in par with what the ant build does.

szeiger, line 431: Ah, I was looking at the wrong manifest. According to http://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#classpath this looks rather application-specific. Does it make sense to include it in a library that gets published to a Maven or Ivy repo? A `Class-Path` would require a specific directory layout locally. The command line use case doesn't look very compelling to me (why use `jrunscript` instead of `scala` from the command line if you have a scala distribution installed?)

rjolly, line 431: In short : because this is what to expect from a JSR-223 compliant language. It is still useful when run from the distribution as obtained from the website. In cases where the directory layout is not right, it costs nothing to have the attribute, it will just not be used.

szeiger:
LGTM

PR #4960, opened 2016-02-11
Fix off-by-one documentation error in Product<N>

janekdb:
Sanity check:

```
scala> case class P2(i: Int, j: Int)
defined class P2

val p2 = P2(1, 3)

p2.productElement(0)
res0: Any = 1

p2.productElement(1)
res1: Any = 3

p2.productElement(2)
java.lang.IndexOutOfBoundsException: 2
  at P2.productElement(<console>:10)
```

### DIFF
--- a/src/build/genprod.scala
+++ b/src/build/genprod.scala
@@ -15,7 +15,6 @@ import scala.language.postfixOps
  *      where the argument is the desired output directory
  *
  *  @author  Burak Emir, Stephane Micheloud, Geoffrey Washburn, Paul Phillips
- *  @version 1.1
  */
 object genprod extends App {
   val MAX_ARITY = 22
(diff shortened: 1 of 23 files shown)


janekdb:
Review by @som-snytt, @heathermiller, @dickwall.

src/build/genprod.scala
```diff
@@ -353,7 +352,7 @@ object ProductTwo extends Product(2)
 
 class Product(val i: Int) extends Group("Product") with Arity {
   val productElementComment = """
-  /** Returns the n-th projection of this product if 0 < n <= productArity,
+  /** Returns the n-th projection of this product if 0 <= n < productArity,
    *  otherwise throws an `IndexOutOfBoundsException`.
    *
    *  @param n number of the projection to be returned
```
janekdb, line 358: Hmm, wondering now if the component at zero-based index 7 is properly called the 7th or the 8th projection.

If it is called the 8th projection then the main comment needs to be,

   Returns the (n+1)-th projection of this product if 0 <= n < productArity,

janekdb, line 358: This version of the main comment avoids having to decide if projection indexing starts at 0 or 1. It is also consistent with the commenting on the `_n` defs,

```
Returns the projection of element n + 1 of this product if 0 <= n < productArity,
```

Technically accurate but possibly less clear. This change could just be left as-is and would be an improvement.

lrytz:
LGTM, thanks! I agree the doc is clear now after your fix.

PR #4961, opened 2016-02-12
Update to partest 1.0.13

retronym:
Which lets us remove the dependency on sbt-partest-interface,
as this has been incorporated into scala-partest itself.

The new version of partest is still in transit between Sonatype and Maven Central.

I've tested this locally by explicitly adding Sonatype releases as a resolver.

``` scala
resolvers in ThisBuild += sbt.Opts.resolver.sonatypeReleases
```

```
> partest --srcpath scaladoc --run --terse
[info] Packaging /Users/jason/code/scala2/build-sbt/pack/lib/scala-partest-javaagent.jar ...
[info] Done packaging.
Selected 57 tests drawn from 1 named test categories

# starting 57 tests in run
```

Review by @szeiger


retronym:
/nothingtoseehere (Only changes the SBT build which isn't yet used by CI)

PR #4962, opened 2016-02-12
Update to partest 1.0.13

retronym:
Which lets us remove the dependency on sbt-partest-interface,
as this has been incorporated into scala-partest itself.

Review by @szeiger


retronym:
Targetting at 2.12 only.

szeiger:
Changes LGTM. What's the reason for not updating 2.11.x as well?

retronym:
We'd need to back port Lukas ' cleanup of invalid flag files, which are not allowed by partest anymore.

SethTisue:
that backport is now https://github.com/scala/scala/pull/4975

PR #4963, opened 2016-02-12
Generate leaner code for branches

lrytz:
GenBCode used to generate more bytecode for branching instructions than
GenASM. A simple method

```
def f(x: Int, b: Boolean) = if (b) 1 else 2
```

would generate

```
 ILOAD 2
 IFNE L1
 GOTO L2
L1
 ICONST_1
 GOTO L3
L2
 ICONST_2
L3
 IRETURN
```

If the conditional branch is negated (IFEQ) the GOTO is unnecessary.

While -Yopt:l:method would clean this up, it's also not too hard to
generate the leaner bytecode in the first place.


retronym:
The code sample in the commit comment doesn't appear to be quite the same as the bytecode

lrytz:
Review by @retronym

lrytz:
Added another commit to prevent translating `null` to `ACONST_NULL; POP; ACONST_NULL`.

retronym:
LGTM, nicely explained and tested.

PR #4964, opened 2016-02-12
fix smhasher url

xuwei-k:
google code is dead
http://google-opensource.blogspot.jp/2015/03/farewell-to-google-code.html


SethTisue:
LGTM.

this could have targeted 2.11.x, but nbd either way. (2.11.x changes get merged onto 2.12.x periodically; commits don't move in the other direction unless explicitly backported)

lrytz:
re-submitted at https://github.com/scala/scala/pull/4967

PR #4965, opened 2016-02-13
SI-8790 test case

lrytz:
Tuples created for pattern matching are eliminated since
https://github.com/scala/scala/pull/4858


retronym:
LGTM

PR #4966, opened 2016-02-14
SD-79 don't issue spurious inliner warnings under l:project

lrytz:
When enabling `-Yopt:inline-project` (or `-Yopt:l:project`), the inliner
would spuriously warn about callsites to methods marked `@inline` that
are read from the classpath (not being compiled currently).

This patch introduces yet another field to the `Callsite` class, which
is growing a bit too large. But the call graph representation will get
an overhaul when implementing the new inliner heuristics (2.12.0-M5), so
this is just a temporary fix that would be nice to have in M4.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
@@ -132,12 +132,13 @@ class CallGraph[BT <: BTypes](val btypes: BT) {
               (declarationClassNode, source) <- byteCodeRepository.classNodeAndSource(declarationClass): Either[OptimizerWarning, (ClassNode, Source)]
             } yield {
                 val declarationClassBType = classBTypeFromClassNode(declarationClassNode)
-                val CallsiteInfo(safeToInline, safeToRewrite, annotatedInline, annotatedNoInline, samParamTypes, warning) = analyzeCallsite(method, declarationClassBType, call.owner, source)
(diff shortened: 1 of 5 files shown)


lrytz:
huh:

```
scala.collection.mutable.ArrayOps$ofRef cannot be cast to scala.collection.mutable.IndexedSeq
```

not obvious how (if?) it's related to my changes..

retronym:
```
/home/jenkins/workspace/scala-2.12.x-validate-test/build-ant-macros.xml:349: java.lang.ClassCastException: scala.collection.mutable.ArrayOps$ofRef cannot be cast to scala.collection.mutable.IndexedSeq
    at scala.collection.mutable.IndexedSeqLike$class.thisCollection(IndexedSeqLike.scala:39)
    at scala.collection.mutable.ArrayOps$class.seq(ArrayOps.scala:172)
    at scala.collection.mutable.ArrayOps$ofRef.seq(ArrayOps.scala:184)
    at scala.collection.mutable.ArrayOps$ofRef.seq(ArrayOps.scala:184)
```

Looks like `scala.collection.mutable.ArrayOps$class.seq` is not indirecting the call to `thisCollection` through the trait forwarder `ArrayOps.thisCollection`, which would get to the override in `ofRef`, rather than the implementation in `scala.collection.mutable.IndexedSeqLike`. No idea why.

retronym:
Maybe the calculation of `isRewritableTraitCall` is wrong, and this change exposes the latent bug.

lrytz:
The problem was that I messed up the pattern match here https://github.com/scala/scala/pull/4966/files#diff-5158a9f0244fae0dd67b78969168cb03L86. Too many booleans in `Callee`. As said in the commit comment, I'll clean this up with the new inliner heuristics.

retronym:
LGTM

PR #4967, opened 2016-02-15
fix smhasher url

lrytz:
re-submission of #4964 for 2.11.x


SethTisue:
LGTM, thanks @xuwei-k (and @lrytz)

PR #4968, opened 2016-02-15
Remove -Y settings that are no longer used in 2.12

lrytz:
Added a deprecation warning for `-optimize`.

Later we'll also graduate `-Yopt` to `-opt`, probably for 2.12.0-M5.


soc:
@lrytz Wouldn't it be reasonable to keep `-optimize`, so that people have less need to distinguish between Scala versions in their build files?

lrytz:
Two reasons against:
- `-Yopt` is a multi-choice, so if we just rename it to `-optimize` users get an error. We'd need to set a default choice (eg `l:classpath`), which I'm not sure is a good idea.
- I'd love to get over the `-optimize` vs `-optimise` bikeshed :)

soc:
What I meant is that we could pick a default for optimize, so that code written for < 2.12 doesn't fail compilation on 2.12.

retronym:
`-optimi{sz}e` could be deprecated alias for `-opt:l:classpath`. The deprecation message could point to a doc about the new optimizer that spells out the cost/benefits between the different optimization levels.

soc:
@retronym That sounds like a nice approach! I think that would help users so that we don't needlessly incur the SBT config overhead we had when scala-xml was moved to a module.

lrytz:
OK, I implemented that, see
- https://github.com/lrytz/scala/commit/b7626d998ee415271105450f125c507629b8eae2#diff-3da64c3c223da7bbf9970ac26f9fb9b8L361
- https://github.com/lrytz/scala/commit/b7626d998ee415271105450f125c507629b8eae2#diff-23914508e085783a817c17e550b96b70 
- https://github.com/lrytz/scala/commit/b7626d998ee415271105450f125c507629b8eae2#diff-f90d44bc1be8e3167835ea5445b5de8eR383

soc:
Nice cleanups! This LGTM!

adriaanm:
LGTM!

PR #4970, opened 2016-02-17
SI-9658 Fix crosstalk between partial fun. and GADT match

retronym:
When typechecking the synthetic default case of a pattern matching
anonymous partial function, we failed to create a new `Context`.
This led to crosstalk with the management of the saved type bounds
of an enclosing GADT pattern match.

This commit avoids the direct call to `typeCase` and instead
indirects through `typedCases`, which spawns a new nested typer
context, and hence avoids the crosstalk when `restoreSavedTypeBounds`
runs.


retronym:
Review by @adriaanm

adriaanm:
LGTM

retronym:
Rebased (trivially) to fix the reported conflict.

PR #4972, opened 2016-02-17
Add package member filtering

felixmulder:
Previously only entities like classes and traits would allow for filter
of members. This commit removes that restriction by always including
the `<div id="order">` in the entity template.

As was discovered by @heathermiller in #4912.

Review: @VladUreche


VladUreche:
LGTM, thanks @felixmulder!
Will merge when the tests are done.

heathermiller:
Nice! LGTM :)

felixmulder:
Thanks guys :)

PR #4973, opened 2016-02-17
Scaladoc: Replace UTF8 glyphs with the Material Typeface

felixmulder:
This commit removes arrows and other symbols previously provided by
UTF8-symbols. Instead it uses the Material Typeface that we've used
for the search icon and graph symbols.

review: @VladUreche, @lrytz

I'm pulling Lukas into this one since he's the one whose phone was affected by this bug.

Docs for this PR available at: https://dl.dropboxusercontent.com/u/358427/scaladoc-glyphs/scala/Boolean.html


VladUreche:
@heathermiller may be in a better position to comment on this patch.

lrytz:
LGTM!

heathermiller:
Nice :) LGTM!

PR #4974, opened 2016-02-17
More conservative optimization for unnecessary outer ref checks

szeiger:
Outer ref checks in pattern matches are elided if the tested prefix is the same as the expected prefix or the expected type is defined in a "static" location (i.e. directly in a package or inside a chain of nested objects).

The `@unchecked` annotation can be used to suppress the warning when an outer ref check is required but no outer ref is available (like for nested final classes).

Includes a spec update to clarify the use of prefixes in type patterns.

For review only. This needs to be squashed before merging.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/patmat/MatchTreeMaking.scala
+++ b/src/compiler/scala/tools/nsc/transform/patmat/MatchTreeMaking.scala
@@ -316,7 +316,7 @@ trait MatchTreeMaking extends MatchCodeGen with Debugging {
       trait TypeTestCondStrategy {
         type Result
 
-        def outerTest(testedBinder: Symbol, expectedTp: Type): Result
+        def withOuterTest(orig: Result)(testedBinder: Symbol, expectedTp: Type): Result = orig
         // TODO: can probably always widen
         def typeTest(testedBinder: Symbol, expectedTp: Type): Result
         def nonNullTest(testedBinder: Symbol): Result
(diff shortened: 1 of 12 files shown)


adriaanm:
Sorry, I'm really deep in INDY Sammy land. Will emerge before I fly to Old Country land (Wed) with feedback.

src/compiler/scala/tools/nsc/transform/patmat/MatchTreeMaking.scala
```diff
@@ -336,18 +336,34 @@ trait MatchTreeMaking extends MatchCodeGen with Debugging {
...
+              else if (tp.typeSymbol.isModuleClass) isStatic(tp.prefix)
+              else false
+            tp.typeSymbol.owner == tp.prefix.typeSymbol && isStatic(tp.prefix)
           }
 
-          // ExplicitOuter replaces `Select(q, outerSym) OBJ_EQ expectedPrefix` by `Select(q, outerAccessor(outerSym.owner)) OBJ_EQ expectedPrefix`
-          // if there's an outer accessor, otherwise the condition becomes `true` -- TODO: can we improve needsOuterTest so there's always an outerAccessor?
-          val outer = expectedTp.typeSymbol.newMethod(vpmName.outer, newFlags = SYNTHETIC | ARTIFACT) setInfo expectedTp.prefix
-
-          (Select(codegen._asInstanceOf(testedBinder, expectedTp), outer)) OBJ_EQ expectedOuter
```
adriaanm, line 350: Are there well-typed matches where a statically known expected type could never match (and thus the test is constant `false` instead of `true`). I can't think of any, but it's worth exploring.

szeiger, line 350: I can't think of any, either, and I think there should not be any in principle. Such a match could never succeed, so if we can detect it at compile-time, it should fail to compile in the first place.

szeiger:
Added the test case that @lrytz ran into yesterday

szeiger:
Squashed and cleaned up. This should be good to go.

adriaanm:
LGTM -- good stuff, @szeiger

PR #4977, opened 2016-02-18
SBT build tweaks

retronym:
I accumulated these changes while seeing if the SBT build could be used in place of Ant in the bootstrap script. These changes aren't sufficient, but I thought I'd get the ball rolling with this PR.

The motivation for a custom setting for `skipDoc` setting key is that `set every publishArtifact in packageDoc` seemed not to work for me. (`set every` has some limitations in SBT).

Future work:
- Add a task equivalent to `publishCoreNoDocs` in the Ant build
- Figure out how to get SBT to pull a Scala version from the local Ivy version (I tried `sbt publishLocal` / `sbt 'set scalaVersion = ...' update`), but ran into problems.

Review by @szeiger

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -150,6 +150,7 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   cleanFiles += (classDirectory in Compile).value,
   cleanFiles += (target in Compile in doc).value,
   fork in run := true,
+  skipDoc := false,
   scalacOptions in Compile in doc ++= Seq(
     "-doc-footer", "epfl",
     "-diagrams",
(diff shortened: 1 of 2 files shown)


szeiger:
If by `publishCoreNoDocs` you mean the same as used in `scripts/validate/publish-core` but without optimization, I think the right way to go is to add another setup command similar to `ScriptCommands.setupPublishCore`. This is the one that is already used by `publish-core`. It does all the setup, including disabling docs generation (the correct magic invocation for which is `publishArtifact in (Compile, packageDoc) in ThisBuild := false`).

The reason why I added a single command to do the whole setup procedure is because there doesn't seem to be a way (at least not an easy or documented one) to programmatically alter the sbt state, starting from the _current_ state. You can find examples for performing sort of the equivalent of `set` from a command (and that's what `setupPublishCore` does as well) but these commands always start with the _initial_ state from the build script, so you cannot execute more than one from the command line. The real `set` command (which _can_ be run multiple times from the command line to update the _current_ state) requires access to private sbt internals.

The dependency version overriding makes sense. I originally had custom SettingKeys in mind for this in the long run, but that's probably overkill. Even if we end up removing `versions.properties` to put those values into `build.sbt` instead, we could still check the System properties there.

Summing up: I don't think we need `skipDoc`. `publish-core-no-docs` should work the same as `setupPublishCore`. The version override makes sense but it would be nice if you could target 2.11.x with that. All the sbt build and validation script work has been in 2.11.x so far.

/cc @SethTisue

PR #4978, opened 2016-02-18
Scaladoc: Add better hover-text in search results

felixmulder:
This commit adds hover text to entities in the search results. It does this by taking the `comment.short` and turning it into a `String` which is then placed in the JSON blob generated by `IndexScript.scala`.

Generated docs: https://dl.dropboxusercontent.com/u/358427/scaladoc-hover/index.html

Review: @VladUreche


VladUreche:
LGTM. Thanks @felixmulder!

PR #4979, opened 2016-02-18
Scaladoc: Add progress bar for search

felixmulder:
As suggested by @jrudolph, here's a simple implementation of the progress bar. It the `setProgress` function is run each time a package has been searched and the results added to the DOM. It seems to be accurate enough to not warrant a more advanced implementation.

Preview: https://dl.dropboxusercontent.com/u/358427/scaladoc-progress/index.html
Review: @heathermiller

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
@@ -63,6 +63,9 @@ trait EntityPage extends HtmlPage {
     <body>
       { search }
       <div id="search-results">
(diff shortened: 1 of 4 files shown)


heathermiller:
Nice! LGTM

felixmulder:
Merge me @VladUreche :)

VladUreche:
Done, thanks @felixmulder!

jrudolph:
Cool, that was fast! Thanks, @felixmulder. A progress bar was just what I imagined as it neatly covers both the fact that something is ongoing and then, that it finished.

PR #4981, opened 2016-02-19
[sbt] Allow system properties to override entries in verions.props

retronym:
Resubmmision of #4977

Review by @szeiger


szeiger:
LGTM

PR #4983, opened 2016-02-19
Add package view to scaladoc

felixmulder:
Happy Friday everyone!

So, this is what we have arrived at after trying out different ideas. This commit will add a package view on the right hand side of the main documentation. It will list these packages: parent, current, children, and siblings. 

Review: @heathermiller 
Preview: https://dl.dropboxusercontent.com/u/358427/scaladoc/scala/index.html
CC: @VladUreche, @SethTisue, @retronym, @lrytz 

EDIT: and before anybody says anything about mobile - this is for the desktop. We're still figuring out how best to display it on mobile. Therefore, it's hidden on mobile - for now.

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
@@ -71,9 +71,74 @@ trait EntityPage extends HtmlPage {
           <div id="member-results"></div>
         </div>
       </div>
-      <div id="content-container" style="-webkit-overflow-scrolling: touch;">
-        <div id="content">
-          { content }
+      <div id="content-scroll-container" style="-webkit-overflow-scrolling: touch;">
+        <div id="content-container" style="-webkit-overflow-scrolling: touch;">
+          <div id="subpackage-spacer">
+            <div id="packages">
+              <h1>Packages</h1>
+              <ul>
+                {
+                  def entityToUl(mbr: TemplateEntity with MemberEntity, indentation: Int): NodeSeq =
+                    <li class={"current-entities indented" + indentation}>
+                      {
+                        mbr match {
+                          case dtpl: DocTemplateEntity =>
+                            dtpl.companion.fold(<span class="separator"></span>) { c: DocTemplateEntity =>
+                              <a class="object" href={relativeLinkTo(c)} title={c.comment.fold("")(com => inlineToStr(com.short))}></a>
+                            }
+                          case _ => <span class="separator"></span>
+                        }
+                      }
+                      <a class={mbr.kind} href={relativeLinkTo(mbr)} title={mbr.comment.fold("")(com => inlineToStr(com.short))}></a>
+                      <a href={relativeLinkTo(mbr)} title={mbr.comment.fold("")(com => inlineToStr(com.short))}>
+                        {mbr.name}
+                      </a>
+                    </li>
+
+                  // Get path from root
+                  val rootToParentLis = tpl.toRoot
+                    .tail
+                    .reverse
+                    .zipWithIndex
+                    .map { case (pack, ind) =>
+                      memberToHtml(pack, tpl, indentation = ind, isParent = (pack eq tpl.toRoot.tail.head))
+                    }
+
+                  val parent = tpl.toRoot match {
+                    case _ :: parent :: _ if !parent.isRootPackage => Some(parent)
+                    case _ => None
+                  }
+
+                  val parentSub = parent.fold(Seq[TemplateEntity with MemberEntity](tpl)) { p =>
+                    p.templates.filter(_.isPackage).sortBy(_.name)
+                  }
+
+                  // If current entity is a package, take its containing entities - otherwise take parent's containing entities
+                  val currentPackageTpls =
+                    if (tpl.isPackage) tpl.templates
+                    else parent.fold(Seq.empty[TemplateEntity with MemberEntity])(p => p.templates)
+
+                  val (subsToTpl, subsAfterTpl) = parentSub.partition(_.name <= tpl.name)
+
+                  val subsToTplLis = subsToTpl.map(memberToHtml(_, tpl, indentation = rootToParentLis.length))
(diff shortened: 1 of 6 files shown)


soc:
Very nice!

I like the more light-weight look of the bar at the right side, but I think the huge class/object/trait/... name bar for the object/class/trait name looks very out-of-place now, especially considering that the search bar has exactly the same color, and the name bar is not full-width anymore.

Ideas: 
- Could we drop the large name bar, and just use the plain background of the page?
- Could we integrate the name into the search bar? (Maybe beneath it), and do some CSS/JS magic to shrink the bar to something smaller when you scroll down?

felixmulder:
Hey @soc! Thanks for the input. That's our thinking as well. It will be changed. Check the scala-dev issues for more of our todos!

lrytz:
Hi @felixmulder, I'm :mountain_cableway: :mount_fuji: :ski: for a week, so I hopefully won't be around much :)

My first impression is very positive.
- Once in a sub-package there's no easy way to get back up (how do you get from "immutable" back to "scala"?)
- maybe that's specific to collections: once in package "immutable", i do see the classes in that packe, but it looks quite cluttered with all the type signatures.. see https://dl.dropboxusercontent.com/u/358427/scaladoc/scala/collection/immutable/index.html

felixmulder:
Refactored the tree structure to include a path back up to the top package (i.e. scala in the standard library).

https://dl.dropboxusercontent.com/u/358427/scaladoc/scala/collection/parallel/index.html

Regarding the cluttering of type signatures in packages - I agree. I'll see what I can do about it once we start the rest of the cleanup.

felixmulder:
Refactored the tree structure to include the path back up to the top package (i.e. scala in the standard library).

https://dl.dropboxusercontent.com/u/358427/scaladoc/scala/collection/parallel/index.html

Regarding the cluttering of type signatures in packages - I agree. I'll see what I can do about it once we start the rest of the cleanup.

Have fun on your swoop-swoop Lukas :)

soc:
Maybe the signatures could be limited to one line and the rest could be folded by default, like the description?

felixmulder:
@soc - this is an older mockup (as can be seen by the package view) of what's to come:
screen shot 2016-02-18 at 17 43 19

I'm thinking of removing some of the extraneous type information in the `Type Members` section. WDYT?

Example:

``` scala
class HashMap[A, +B] extends AbstractMap[A, B] with Map[A, B] with MapLike[A, B, HashMap[A, B]] with Serializable with CustomParallelizable[(A, B), ParHashMap[A, B]]
```

Perhaps we could cut away everything after `HashMap[A, +B]` but still show the full type on its entity page and perhaps also on hover in the package.

SethTisue:
re: the CLA checking, perhaps there is some issue here relating to the company name change. I'll look into it.

felixmulder:
Updated the PR with two things:
- Out of bounds keyboard navigation
- Hover text on packages
- Entity url spacing

felixmulder:
@SethTisue, @VladUreche: Heather is currently on deadline and as such asked me to grab someone else for review. One of you guys up for it?

VladUreche:
Hi @felixmulder. I like the approach but here are some nitpicks:
- I can't get to the root package documentation anymore :(
- Sometimes it shows all subpackages (e.g. in `scala.collection`) and at other times just the tree that leads to the current package (e.g. in `scala.collection.immutable`). To me, it's rather counter-intuitive, since from the first screen I get the idea that it always shows all packages + the expanded tree for the current package
- You mentioned we shouldn't ask about mobile, but what's the plan there?

I actually would recommend adding an extra type of diagram, **Package Overview**, which would serve the purpose of browsing through the packages. That would only appear in the root package and would list all packages in the current scaladoc site, using the infrastructure already in place. WDYT?

felixmulder:
Hey @VladUreche !

> - I can't get to the root package documentation anymore :(

Is this better? https://dl.dropboxusercontent.com/u/358427/scaladoc/scala/collection/immutable/index.html

I omitted the root package purposefully, thought it looked ugly - but as you and @soc pointed out, it gets confusing. I concede! :)

> - Sometimes it shows all subpackages (e.g. in `scala.collection`) and at other times just the tree that leads to the current package (e.g. in `scala.collection.immutable`). To me, it's rather counter-intuitive, since from the first screen I get the idea that it always shows all packages + the expanded tree for the current package

This is what it always shows: path to `root` package, entity parent package, all sibling packages of entity and all child packages of entity.

For `scala.collection` it has both siblings and children - hence this view:

screen shot 2016-02-24 at 16 29 15

`scala.collection.immutable` only has sibling packages - no children. Hence a path to root is shown, as well as the sibling packages.

screen shot 2016-02-24 at 16 29 28

This is the way that both Rust and Elm do it - just to give you a reference if you want to have a look at where we've gotten our inspo from!

> - You mentioned we shouldn't ask about mobile, but what's the plan there?

Some form of breadcrumbs, like a file system basically. Example: 

Regarding a view with a full list of entities in the root package, that is entirely possible. But I think if we are to implement that, perhaps we want both that and this? WDYT?

If anyone besides @VladUreche is reading this, stop idling and voice your opinions! :smile:

VladUreche:
@FelixMulder Thanks for the answers and for adding `root` to the hierachy!

For the packages shown, I don't know the context in which the Rust/Elm documentations hide siblings, but for me, this violates the principle of least surprise... Am I the only one in bothered? Please speak up!

felixmulder:
On @heathermiller's recommendation I've added the currently selected package's entities. This is a definite improvement over the old filter panel - but doesn't go as far as to alienate all Java-brethren.

Latest version: https://dl.dropboxusercontent.com/u/358427/scaladoc-package-ents/scala/collection/index.html

heathermiller:
This LGTM for M4.

I think there are more tweaks and experiments to be had, but I think this is good for now for people to experiment with. 

@VladUreche, if you agree, would you merge?

SethTisue:
> This LGTM for M4.
> 
> I think there are more tweaks and experiments to be had, but I think this is good for now for people to experiment with.

:+1:

felixmulder:
Great, squished and awaiting the tests to run through. :tada:

VladUreche:
Since Heather LGTM-ed it, here goes :+1:

PR #4987, opened 2016-02-20
Fix var spelling in WeakHashSet

janekdb:
Fixed spelling of private def only.

`var threshhold` remains in need of fix.

FWIW, the project builds `$ ant build` with the var access restricted,

```
  /**
   * the limit at which we'll increase the size of the hash table
   */
  private[this] var threshhold = computeThreshold
```

The package is also internal strongly hinting it should not be depended on,

```
package scala
package reflect.internal.util
```

Would those two facts allow the var to be made private and renamed or just renamed in 2.12?

The alternatives are,
- Leave it as-is
- Deprecate it in 2.12 and make private in 2.13


janekdb:
Review by @xeno-by.

xeno-by:
Wdyt @retronym?

SethTisue:
I think this should target 2.12.

janekdb:
Replaced by #5014.

PR #4988, opened 2016-02-21
SI-9665 Backquoted vbar in extractor pattern

som-snytt:
Allow an infix extractor named `|`, when backquoted.

Because why not.

Not sure if `isIdentExcept` and `isIdentOf` see any usages.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -1966,8 +1966,8 @@ self =>
           case _ => EmptyTree
         }
         def loop(top: Tree): Tree = reducePatternStack(base, top) match {
-          case next if isIdentExcept(raw.BAR) => pushOpInfo(next) ; loop(simplePattern(badPattern3))
-          case next                           => next
+          case next if isIdent && !isRawBar => pushOpInfo(next) ; loop(simplePattern(badPattern3))
+          case next                         => next
         }
         checkWildStar orElse stripParens(loop(top))
       }
--- a/test/files/pos/t9665.scala
+++ b/test/files/pos/t9665.scala
@@ -1,0 +1,7 @@
+
+object | { def unapply(x: (Any, Any)) = Some(x) }
+
+trait Test {
+  def f() = (1,2) match { case 1 `|` 2 => }
+  def g() = 2 match { case 1 | 2 => }
+}


src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -654,6 +654,8 @@ self =>
     def isRawBar   = isRawIdent && in.name == raw.BAR
     def isRawIdent = in.token == IDENTIFIER
 
+    def isIdentExceptRaw(except: Name) = isIdent && !(isRawIdent && in.name == except)
```
adriaanm: I think we can simplify a bit as follows

```
// inline definitions
(in.token == IDENTIFIER || in.token == BACKQUOTED_IDENT) && !(in.token == IDENTIFIER && in.name == except)
// distribute negation
(in.token == IDENTIFIER || in.token == BACKQUOTED_IDENT) && (in.token != IDENTIFIER || in.name != except)
// commutativity of ||
(in.token == IDENTIFIER || in.token == BACKQUOTED_IDENT) && (in.name != except || in.token != IDENTIFIER)
// exclude middle
(in.token == IDENTIFIER || in.token == BACKQUOTED_IDENT) && (in.name != except || in.token == BACKQUOTED_IDENT)
// pull out `|| in.token == BACKQUOTED_IDENT`
(in.token == IDENTIFIER && in.name != except) || in.token == BACKQUOTED_IDENT
```

I guess we can reabstract now and get to `in.token == BACKQUOTED_IDENT || isIdentExcept(name)`.

Also, are there other cases where we should be allowing backquoted idents? (modulo #4935).

som-snytt: Do you still want me to let you know when you're being a bit pedantic?

som-snytt: We neglected to say, `isIdent && !isRawBar` at 1971.

Once we'd said that, "raw oyster bar" could not be far behind.

som-snytt:
Adriaan's irrefutable logic has once again led me to use a boolean operator. And an existing method.

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -1966,8 +1966,8 @@ self =>
           case _ => EmptyTree
         }
         def loop(top: Tree): Tree = reducePatternStack(base, top) match {
-          case next if isIdentExcept(raw.BAR) => pushOpInfo(next) ; loop(simplePattern(badPattern3))
-          case next                           => next
+          case next if isIdent && !isRawBar => pushOpInfo(next) ; loop(simplePattern(badPattern3))
```
adriaanm, line 1969: Nice! Better than the buffet of operators I marched past.

som-snytt:
This is a totally ragged-edge case, but maybe @adriaanm will green-light it. The scope is exceptionally narrow.

adriaanm:
LGTM I'm sorry, not sure how I got interrupted between the approving comment and the subsequent LGTM

som-snytt:
"buffet of operators" -> "food" -> "go visit vending machine."  Hey, it happens.

adriaanm:
A good thing I didn't try to tip it (over).

PR #4990, opened 2016-02-22
SI-9666: Use inline group names in Regex

som-snytt:
Delegate `Match group name` to the underlying `matcher`.

If that fails, try explicit group names as a fall back.

Because no attempt is made to correlate inline and explicit
names, explicit names are deprecated.

In the following case, either name is accepted:

```
new Regex("a(?<Bar>b*)c", "Bee")
```

But if names are reversed, the error is undetected:

```
new Regex("a(?<Bee>b*)(?<Bar>c)", "Bar", "Bee")
```

### DIFF
--- a/src/library/scala/util/matching/Regex.scala
+++ b/src/library/scala/util/matching/Regex.scala
@@ -182,6 +182,9 @@ class Regex private[matching](val pattern: Pattern, groupNames: String*) extends
    *  val namedYears = for (m <- namedDate findAllMatchIn dates) yield m group "year"
    *  }}}
    *
+   *  Group names supplied to the constructor are preferred to inline group names
+   *  when retrieving matched groups by name. Not all platforms support inline names.
+   *
    *  This constructor does not support options as flags, which must be
    *  supplied as inline flags in the pattern string: `(?idmsux-idmsux)`.
    *
@@ -578,6 +581,9 @@ object Regex {
    */
   trait MatchData {
 
+    /** Basically, wraps a platform Matcher. */
+    protected def matcher: Matcher
+
     /** The source from which the match originated */
     val source: CharSequence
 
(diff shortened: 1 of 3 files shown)


som-snytt:
This has the virtue of throwing `NoSuchElementException` on a bad name, for compatibility; but it would be nice to throw `IllegalArg` uniformly and only check explicit names if any were provided.

sjrd:
This could be damaging for Scala.js. Currently, one can use Scala's explicit group names in Scala.js: http://www.scala-js-fiddle.com/gist/6dcd4feac239f1988f25
But JS regexps do not have implicit group names, and therefore Java regexps don't support group names in Scala.js. The alternative of using implicit group names is therefore not possible in Scala.js.

som-snytt:
@sjrd thx for the heads up, I'll back out the deprecation, and check out the scala.js interaction.

som-snytt:
`t5045` uses the alt ctor, though it doesn't use `group(name)`

adriaanm:
Ping! We hope to close the M4 queue this week.

som-snytt:
Squashed, rebased, with tweaks. Not JS tested yet.

src/library/scala/util/matching/Regex.scala
```diff
@@ -190,6 +190,23 @@ class Regex private[matching](val pattern: Pattern, groupNames: String*) extends
...
+   *  val namedDate  = new Regex("""(\d\d\d\d)-(\d\d)-(\d\d)""", "year", "month", "day")
+   *  val namedYears = for (m <- namedDate findAllMatchIn dates) yield m group "year"
+   *  }}}
+   *
+   *  This constructor does not support options as flags, which must be
+   *  supplied as inline flags in the pattern string: `(?idmsux-idmsux)`.
+   *
+   *  @param regex      The regular expression to compile.
+   */
+  def this(regex: String) = this(Pattern.compile(regex))
```
som-snytt: If the other ctor isn't deprecated, this is extraneous.

adriaanm:
@sjrd does this LGT.js?

som-snytt:
I see I didn't clean up a few lines yet. Also, I assume `matcher.group(String)` must compile against scala.js. I assume a facade can just throw? I haven't looked yet. :shaved_ice: It doesn't autocomplete an emoji for shame, so shaved ice will have to do.

som-snytt:
I cleaned it up and adjusted the doc a bit.

sjrd:
Right now, this will fail to _link_ if anyone uses Scala's `Regex.MatchData.group(String)`, because [our implementation of `Matcher`](https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/regex/Matcher.scala) does not declare [the method `def group(String)`](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#group%28java.lang.String%29).

However, we can fix that on Scala.js' side by defining

``` scala
def group(name: String): String = throw new IllegalArgumentException
```

which technically respects the spec since there cannot be any capturing group with the given name.

So yes, LGT.js.

som-snytt:
Thanks, @sjrd . At some point, I had code to strip inline group names, probably to support old Java versions; something like that would be needed to let code using group names to cross-compile with JS, but oh well.

lrytz:
LGTM

PR #4788, opened 2015-10-06
explicitly specify insertion-order feature in docs

dk14:
As I can see, many developers see it as insertion-order tolerant collection: http://stackoverflow.com/questions/9313866/immutable-scala-map-implementation-that-preserves-insertion-order

However, some of them not so sure, see comments here: http://stackoverflow.com/a/32872750/1809978,  http://stackoverflow.com/a/32872791/1809978

So, how about making it clear?

PR #4994, opened 2016-02-23
explicitly specify insertion-order feature in docs

dk14:
It's based on this very minor pull request: https://github.com/scala/scala/pull/4788

Just added similar details for mutable version of `ListMap`.

P.S. Next time I'll try to not split that kind of minor changes into different PRs.

### DIFF
--- a/src/library/scala/collection/mutable/ListMap.scala
+++ b/src/library/scala/collection/mutable/ListMap.scala
@@ -15,7 +15,7 @@ package mutable
 import generic._
 import annotation.tailrec
 
-/** A simple mutable map backed by a list.
+/** A simple mutable map backed by a list, so it preserves insertion order.
  *
  *  @tparam A    the type of the keys contained in this list map.
  *  @tparam B    the type of the values assigned to keys in this list map.


dk14:
{"user":"dk14","signed":true,"version":"1.0","currentVersion":"1.0"}

SethTisue:
(we're having some trouble with the CLA checker because of the company name change)

SethTisue:
thank you Dmytro!

SethTisue:
LGTM

SethTisue:
the claim we inserted here is actually false since 2.10, as per https://twitter.com/travisbrown/status/735201036873400321

but I think it _should_ preserve order, as it did in 2.9, so I opened https://issues.scala-lang.org/browse/SI-9893. let's have any further discussion there

dk14:
Sorry, I didn't actually check it for mutable version as I didn't notice that  unlike [immutable one](https://github.com/scala/scala/blob/2.12.x/test/junit/scala/collection/immutable/ListMapTest.scala), there is no test for it: https://github.com/scala/scala/tree/2.12.x/test/junit/scala/collection/mutable.

PR #4997, opened 2016-02-25
Unclutter scaladoc entity members

felixmulder:
This commit removes clutter from the entity view, as shown here: https://dl.dropboxusercontent.com/u/358427/scaladoc-unclutter/scala/collection/index.html

Review: @VladUreche


VladUreche:
Is removing all distinctions between the title, description and other info desirable?



Ping @heathermiller, is this something you had in mind?

felixmulder:
@VladUreche it's along the lines of some things we've been discussing - but H-dawg has not been shown more than mockups :smile:

My redundancy alarm went off when I saw the `root package` with a sub-header of `package root`.  But perhaps I'm overreacting. Here's a version with the sub-header not removed for package entities:

https://dl.dropboxusercontent.com/u/358427/scaladoc-unclutter-vlad-edition/index.html

felixmulder:
Admittedly it looks better on packages that are not the root package e.g: https://dl.dropboxusercontent.com/u/358427/scaladoc-unclutter-vlad-edition/scala/collection/index.html

Haven't updated the PR yet - let me know if this fixes your concerns.

heathermiller:
I think that @felixmulder has a point in trying to reduce redundancy. Maybe we could go with something like this in order to still keep the indication that p = package?:

img_1867

I'm not sure. :/ Thinking about it...

VladUreche:
Okay, what I'm puzzled about is the lack of a special background for the description. I made more sense (to me) when the background behind the main text was white, so the description stood out clearly. But IANAD.

Also, important thing, we need to group the "stuff" in `package scala`:
- `Any`, `AnyVal`, `AnyRef`, `Null`, `Nothing` - group them as `Core Types`
- `Int` and co, group as `Primitive Types`
- `FunctionX` - group as `Functions`
- exceptions
- aliases (`Vector`, `::`, `List`, `Nil` etc)
- products and tuples
- other cr^Hlassses

And we need to do the same for `scala.Predef`, which is also a mess.

felixmulder:
@VladUreche: I'll file the `package scala` groupings as a separate PR momentarily. Regarding The separation of ~~church and state~~ definition and description - WDYT about entities like this: https://dl.dropboxusercontent.com/u/358427/scaladoc-unclutter-vlad-edition/scala/Boolean.html ?

In those entities I don't think there's the same problem. Just the packages.

@heathermiller: I like the idea - we should do something like that no matter what IMO! :+1: The other idea I've been toying with is breadcrumbs instead of package hierarchy tree thingy in #4983 - since that PR seems to be going nowhere.

heathermiller:
Yeah, I saw the breadcrumbs thing. I would guess that it's probably best, for mobile, just to move the sidebar to the bottom of the page rather than going the breadcrumbs route because it means you lose siblings on mobile which is weird. 

RE: #4983: I do very much like the package view! I think we need it! The only things I'd change about it right now are aesthetic – like width of the sidebar (I think slightly wider would look better), and maybe the fonts (make them a bit bigger?) But otherwise it looks good to me. I need to maybe mess with it more intensely to see if it's missing anything or doing anything surprising, but as far as I see right now, I do very much like this view!

heathermiller:
RE: @VladUreche, I don't understand what you don't like? You don't like that behind "(p) collections" is now no longer blue? I like that the blue went away – the interface reads less like a boxy old iPhone 3 UI, so now it draws more attention to the top-level documentation I think.

Also, it makes a lot more sense visually for it to be this way (no blue top bar) when a side bar is added. Again, the goal is to be less iframey/boxy, like old UIs had to be. 

Here's a visual example of what I mean. I feel I'm better drawn to the actual text in the documentation in an article like this (non boxy):
http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Random
Than this:
http://www.scala-lang.org/files/archive/api/current/#scala.collection.immutable.List

In the 2nd case above (Scala's List class) I'm distracted by all the other things crammed into boxes on the page. It makes me look at the top-level text last. Do you see what I mean?

VladUreche:
The entities look better, but I'm still puzzled about the missing blue behind the `(c) Boolean`. Now I realize the white background behind the main description (where the doc comment goes) is long gone, but I didn't notice it because the big blue title bar served as a visual anchor point, delimiting the beginning of the page. Now, with the big blue title bar gone, it feels a bit more difficult to figure out where the description starts.

Not that it's bad, but it takes some time to adapt to, at least for me, since it's not a 1:1 mapping like the boxy look was back when Heather introduced it. Again, IANAD.

heathermiller:
@VladUreche I understand, but it's only problematic because you've looked at old Scaladoc 99,000 times and as you say, you're looking for the same old visual anchor points that you've seen the last 98,999 times. I think it just takes getting used to. For new users that aren't used to old Scaladoc, I would imagine that it's less cluttered and more streamlined this way.

VladUreche:
@heathermiller you're right, I did get attached to the visual cues I used before. Let's keep this new design and see if other users complain.

What I'd like to avoid is the annoying feeling that the UI changed and it doesn't support the older workflows, which makes people feel that functionality has been restricted. For example, we have to think what will a user that previously looked for a class in the sidebar do -- whether they quickly find their way around or if we have to write a text on the left-hand side that instructs them to use the top search bar.

heathermiller:
@VladUreche I agree, and for that reason we'll still have a package overview in a sidebar somehow. Definitely. Without it, you're essentially stuck on the page you're on, unless you search, which is less than ideal.

heathermiller:
@VladUreche The key, though, is to streamline that sidebar. For example, I don't think it's necessary to include every package and class/trait/object that exists in the sidebar like we did before. 

I think @felixmulder is on to something with the sidebar that he's proposed.

Question though – what about also displaying all top-level entities in the current package somewhere on the package page? Do you think we could also do that with a sidebar? Or a list in the middle of the page? We've lost the ability to see entities in a package now with the new sidebar, which is one of these situations that I think will frustrate experienced users

felixmulder:
This PR is now stable enough to merge: https://dl.dropboxusercontent.com/u/358427/scaladoc-unclutter/index.html

heathermiller:
LGTM :)

VladUreche:
Thanks @felixmulder for the update!

PR #4841, opened 2015-11-11
SI-4331 Fix memory leaks across REPL :reset command

retronym:
When replacing the classloader, which retains all values created
in the prior REPL lines, we must be careful not to keep around
references to the previous classloader.

We were doing this in two places: the thread context classloader and
in an runtime reflection mirror.

Both of those references to the old classloader would have been
correctness problems, and would likely have meant some degraded
functionality for a post :reset REPL. Furthermore, they served to
retain memory allocated up to the first :reset forever.

I changed handling of the thread context classlaoder to only set and
restore it around execution of code. This avoids leaking the
classloader to whatever thread happend to initialize the REPL.
We actually call the initialization code from within in a background
thread that allows the user to start typing the first command sooner.

I trawled the git logs to understand the history of how and when
we set the context classloader. It was originally added in
7e617efa8f. A more recent change to
add support for JSR-223 (Scripting) also wrapped this around the
execution 3a30af154e62. I believe the second commit is the correct
approach, and that it safe to remove the old code.

While this commit makes :reset perform its advertised function,
it doesn't go further to making it a particulary useful command.
Ideally we'd offer a mechanism to transport some data across a :reset
boundary. The user could do something like this manually by stashing
data in a `java.util.Map` hosted in some class in the same classloader
as IMain. They would also have to restrict this to sending "pure data"
rather than instances of data types defined in the REPL session
itself.

I located the leaks with YourKit, and have tested manually by
repeatedly running:

```
 qscala -J-XX:+HeapDumpOnOutOfMemoryError -J-Xmx512M
Welcome to Scala 2.12.0-20151110-111920-44ae563d37 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_51).
Type in expressions for evaluation. Or try :help.

scala> {val b = new Array[Byte](256 * 1024* 1024); () => b}
res0: () => Array[Byte] = $$Lambda$1781/1479140596@4218500f

scala> :reset
Resetting interpreter state.
Forgetting this session history:

{val b = new Array[Byte](256 * 1024* 1024); () => b}

Forgetting all expression results and named terms: $intp

scala> {val b = new Array[Byte](256 * 1024* 1024); () => b}
res0: () => Array[Byte] = $$Lambda$1828/1370495328@6831d8fd

scala> :reset
Resetting interpreter state.
Forgetting this session history:

{val b = new Array[Byte](256 * 1024* 1024); () => b}

scala> {val b = new Array[Byte](256 * 1024* 1024); () => b}
[...]

PR #4998, opened 2016-02-25
SI-7898 Read user input during REPL warmup

som-snytt:
The compiler is created on main thread and user input is read
on an aux thread (opposite to currently).

Fixes completion when `-i` is supplied.

Adds an `-I` option to paste init files.

The temporary reader uses postInit to swap in the underlying
reader.

Completion is disabled for the temporary reader, rather than
blocking while it waits for a compiler. But manically hitting
tab is one way of knowing exactly when completion is live.

### DIFF
--- a/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
+++ b/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
@@ -79,6 +79,7 @@ Other startup options:
 
  -howtorun    what to run <script|object|jar|guess> (default: guess)
  -i <file>    preload <file> before starting the repl
+ -I <file>    preload <file>, enforcing line-by-line interpretation
  -e <string>  execute <string> as if entered in the repl
  -save        save the compiled script in a jar for future use
  -nc          no compilation daemon: do not use the fsc offline compiler
--- a/src/repl/scala/tools/nsc/interpreter/ILoop.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ILoop.scala
@@ -908,9 +908,6 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
       // we can get at it in generated code.
       intp.quietBind(NamedParam[IMain]("$intp", intp)(tagOfIMain, classTag[IMain]))
 
-      // add a help function for anyone who types "help" instead of ":help". Easily shadowed.
-      //addHelp()
-
       // Auto-run code via some setting.
       ( replProps.replAutorunCode.option
           flatMap (f => File(f).safeSlurp())
@@ -938,9 +935,24 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
         }
       case _ =>
     }
-    // TODO: wait until after startup to enable obnoxious settings
-    def withSuppressedSettings[A](body: =>A): A = {
-      body
+    // wait until after startup to enable noisy settings
+    def withSuppressedSettings[A](body: => A): A = {
+      val ss = this.settings
+      import ss._
+      val noisy = List(Xprint, Ytyperdebug)
+      val noisesome = noisy.exists(!_.isDefault)
+      val current = (Xprint.value, Ytyperdebug.value)
+      if (isReplDebug || !noisesome) body
+      else {
+        this.settings.Xprint.value = List.empty
+        this.settings.Ytyperdebug.value = false
+        try body
+        finally {
+          Xprint.value       = current._1
+          Ytyperdebug.value  = current._2
+          intp.global.printTypings = current._2
+        }
+      }
     }
     def startup(): String = withSuppressedSettings {
       // starting
--- a/src/repl/scala/tools/nsc/interpreter/ILoop.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ILoop.scala
@@ -1,5 +1,5 @@
 /* NSC -- new Scala compiler
- * Copyright 2005-2015 LAMP/EPFL
+ * Copyright 2005-2016 LAMP/EPFL
  * @author Alexander Spoon
  */
 package scala
@@ -15,7 +15,7 @@ import scala.tools.asm.ClassReader
 import scala.util.Properties.{ jdkHome, javaVersion, versionString, javaVmName }
 import scala.tools.nsc.util.{ ClassPath, Exceptional, stringFromWriter, stringFromStream }
 import scala.reflect.classTag
-import scala.reflect.internal.util.{ BatchSourceFile, ScalaClassLoader }
+import scala.reflect.internal.util.{ BatchSourceFile, ScalaClassLoader, NoPosition }
 import ScalaClassLoader._
 import scala.reflect.io.{ File, Directory }
 import scala.tools.util._
@@ -181,9 +181,9 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
     out.flush()
   }
   // turn off intp reporter and our echo
-  def mumly[A](op: =>A): A =
+  def mumly[A](op: => A): A =
     if (isReplDebug) op
-    else intp beSilentDuring {
+    else intp beQuietDuring {
       val saved = mum
       mum = true
       try op finally mum = saved
@@ -576,9 +576,9 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
     }
   }
 
-  def withFile[A](filename: String)(action: File => A): Option[A] = {
+  def withFile[A](filename: String)(action: File => A): Option[A] = intp.withLabel(filename) {
     val res = Some(File(filename)) filter (_.exists) map action
-    if (res.isEmpty) echo("That file does not exist")  // courtesy side-effect
+    if (res.isEmpty) intp.reporter.warning(NoPosition, s"File `$filename' does not exist.")  // courtesy side-effect
     res
   }
 
@@ -715,6 +715,7 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
    */
   def pasteCommand(arg: String): Result = {
     var shouldReplay: Option[String] = None
+    var label = "<pastie>"
     def result = Result(keepRunning = true, shouldReplay)
     val (raw, file, margin) =
       if (arg.isEmpty) (false, None, None)
@@ -735,6 +736,7 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
       }
     val code = (file, margin) match {
       case (Some(name), None) =>
+        label = name
         withFile(name) { f =>
           shouldReplay = Some(s":paste $arg")
           val s = f.slurp.trim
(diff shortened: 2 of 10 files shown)


som-snytt:
@SethTisue `-i` is so useful, I wonder if this is worth squeezing into 2.11.8 in order not to cripple the great new completion feature.

Sry for eleventh hour; this is code from a few years ago, etc.

The old version had a JLine completer that blocked until the compiler was ready, then swapped in the real completer; this approach is to use a dummy InteractiveReader that swaps in the underlying reader at postInit time.

Also, http://stackoverflow.com/a/35594151/1296806

som-snytt:
Apparently, I added a test once that shows `-i` output. The other note is that the `-i` noise is suppressed. You can use `-Dscala.repl.debug=true` to see errors in init scripts.

SethTisue:
I've opened https://github.com/scala/scala-dev/issues/96 on the (unrelated) integrate-ide failure

SethTisue:
> I wonder if this is worth squeezing into 2.11.8

I don't we think we can risk it :-\

som-snytt:
It would be nice is this got merged forward to 2.12.

som-snytt:
If this got merged, I could use REPL for my tool at work, for which I want some joe to type something in and use autocomplete, but I need either -i for set-up or go totally custom.

src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
```diff
@@ -79,6 +79,7 @@ Other startup options:
 
  -howtorun    what to run <script|object|jar|guess> (default: guess)
  -i <file>    preload <file> before starting the repl
+ -I <file>    preload <file> as though pasted
```
lrytz: to an ordinary user, it's probably not clear what `preload <file> as though pasted` means, and how it behaves differently from `-i` - maybe add something like "i.e., interpreting all at once instead of line-by-line"? you can probably come up with something better..

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -175,10 +175,18 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
+  protected def echo(msg: String) = if (!mum) {
     out println msg
     out.flush()
   }
+  // turn off intp reporter and our echo
+  def mumly[A](op: =>A): A =
+    if (isReplDebug) op
+    else intp beSilentDuring {
+      mum = true
+      try op finally mum = false
```
lrytz: you could save the current state of `mum` in a local and finally restore its original value, so nested calls of `mumly` calls would also work

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -408,11 +416,13 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
-  @tailrec final def loop(): LineResult = {
+  final def loop(): LineResult = loop(readOneLine())
+
+  @tailrec final def loop(line: String): LineResult = {
     import LineResults._
-    readOneLine() match {
+    line match {
       case null => EOF
-      case line => if (try processLine(line) catch crashRecovery) loop() else ERR
+      case _    => if (try processLine(line) catch crashRecovery) loop(readOneLine()) else ERR
```
lrytz: i think an `if else if else` would look cleaner here

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -885,49 +895,104 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
+    /* Actions to cram in parallel while collecting first user input at prompt.
+     * Run with output muted both from ILoop and from the intp reporter.
+     */
+    def loopPostInit(): Unit = mumly {
+      // Bind intp somewhere out of the regular namespace where
+      // we can get at it in generated code.
+      intp.quietBind(NamedParam[IMain]("$intp", intp)(tagOfIMain, classTag[IMain]))
+
+      // add a help function for anyone who types "help" instead of ":help". Easily shadowed.
+      //addHelp()
```
lrytz: leftover from some other patch?

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -885,49 +895,104 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
+
+      // Auto-run code via some setting.
+      ( replProps.replAutorunCode.option
+          flatMap (f => File(f).safeSlurp())
+          foreach (intp quietRun _)
+      )
+      // classloader and power mode setup
```
lrytz: stale comment about "classloader setup"? or did you remove `intp.setContextClassLoader()` accidentally?

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -885,49 +895,104 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
+    /** Reader to use before interpreter is online. */
+    def preLoop = {
+      val sr = SplashReader(newReader) { r =>
+        in = r
+        in.postInit()
+      }
+      in = sr
+      SplashLoop(sr, prompt)
     }
-    loadFiles(settings)
```
lrytz, line 920: looks like method `loadFiles` is now unused and can be removed

lrytz:
the idea to preload files in a separate thread is nice, but there's a drawback of no longer emitting any errors that stem from the loaded file. in current scala

```
$ scala -i Test.scala
Loading Test.scala...
<console>:11: error: not found: value bluberish
       val x = bluberish
               ^

Welcome to Scala 2.12.0-20160502-155348-d550006 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_77).

scala>
```

after this change:

```
$ ~/scala/scala3/build-sbt/quick/bin/scala -i Test.scala
Welcome to Scala 2.11.8-20160225-221846-903998c (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_77).

scala> x
<console>:12: error: not found: value x
       x
       ^
```

thoughts? we could put background loading behind another flag.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -885,49 +895,104 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
+          addReplay(s":load $f")
+        }
+        for (f <- settings.pastefiles.value) {
+          pasteCommand(f)
+          addReplay(s":paste $f")
+        }
+      case _ =>
+    }
+    // TODO: wait until after startup to enable obnoxious settings
+    def withSuppressedSettings[A](body: =>A): A = {
```
lrytz: not sure what the future intensions are, but maybe remove that method until they are actually implemented?

lrytz:
another question: would we break anything if we just changed `-i` to paste mode?

src/repl/scala/tools/nsc/interpreter/ILoop.scala
som-snytt: the whole PR is leftover.

som-snytt:
I commented earlier, "You can use -Dscala.repl.debug=true to see errors in init scripts." But I've forgotten details by now and must respool. The general idea was not to see start-up cruft, including -print output on start-up which isn't relevant to user.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
som-snytt: I'll check that out. It was to delay setting `-print:typer` for example.

som-snytt:
That's a good question about -i. I guess line-by-line means a line can fail independently. Rarely one does want to limp along. Failure could be due to missing jar for something yagni.

src/compiler/scala/tools/nsc/GenericRunnerCommand.scala
som-snytt: Reversed the sense. -I is preload line-by-line.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
som-snytt: I'd become if-averse.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
som-snytt: Removal was https://github.com/scala/scala/pull/4841/files#diff-f31fa4d21a39bc6ed3c30fd38bcbc9d4L892 where the comment was correct, of course.

som-snytt:
@lrytz Pushed feedback fix-up from last week. I realize @lrytz is busy on weekends. I'd wanted to show a colleague a line from this PR, but was foiled b/c hadn't pushed yet. Push early!

som-snytt:
Adds quiet startup. Edit: what a killer feature that is.

Edit: but no "help" function, which bumps the line package name in checkfiles; and is reported on reset.

lrytz:
I really like all the changes here. I'm just still a bit worried about making `-i` quiet by default, I can imagine it causing frustration, so I'd prefer to have that as a separate option. How about adding `-iq` and `-Iq`?

An even better option might be to only report errors and warnings, but skip the repl log output. Not sure how hard that would be to implement.

lrytz:
I understand that it is annoying to get error messages printed while the user already has a prompt at his hand, but then the solution is "simply" to fix the init script not to produce errors / warnings. or we could have the quiet options on top.

som-snytt:
Yes, you're right.

I pushed another commit to improve reporting around script loading and pasting. It doesn't show up yet on github, but hopefully eventually.

The output situation is still confusing: `ILoop.echo` is possibly extraneous; everything should go through `ReplReporter`; there's a notion of "forcing" INFO messages, IIRC, if that's the idea there. The filtering has `quiet during` to turn off results printing and `silent during` to turn off all output. The new `mum` flag was to turn off `ILoop.echo`. In this case, if reporter had a "threshold" setting like loggers, we'd filter "info and below" to see only warnings and errors at startup. Maybe that style of filtering will make it into the next reporting interface.

som-snytt:
There was one so-called Session test that demonstrated a pasted error.

Apparently, the way to update a SessionTest (with inline transcript) is to paste to REPL, which then reproduces the transcript. However, you can't paste a `:paste` command.

lrytz:
i was still getting `<console>` for errors in the `-i` file, here's what made it work: https://github.com/som-snytt/scala/compare/issue/7898-i...lrytz:pr4998?expand=1 - that change is without understanding anything from the control flow, just grepp'd a bit around

som-snytt:
@lrytz Apparently, I didn't try a parse error.

lrytz:
LGTM!

dwijnand:
Cross-linking, this has broken the REPL in sbt: https://issues.scala-lang.org/browse/SI-9804, specifically this commit: https://github.com/scala/scala/commit/99dad60d984d3f72338f3bad4c4fe905090edd51.

som-snytt:
@dwijnand Thanks.

Is there an issue somewhere that `publishLocal` should not enforce using java 6?

```
java.lang.RuntimeException: javac version 1.8.0_60 may not be used to publish, it has to start with 1.6 due to javacVersionPrefix setting
```

dwijnand:
sbt must work on Java 6, so officially that's required, but you're free to drop that in sbt: https://github.com/sbt/sbt/blob/642b091bf68c3f4f770ae043524c08491881074d/project/Release.scala#L23

But I don't think the fix should be in sbt - can't just break the repl for 2.12.0-M5+ for versions of sbt < whatever version ships with a fix (eg. 0.13.12).

som-snytt:
I was building locally because usually some debug is required; but I've looked at the console before, because I remember that initialCommands run very early.

https://github.com/sbt/sbt-javaversioncheck/issues/2

PR #5001, opened 2016-02-28
SI-9656 Fix immutable Range toString method

stevorobs3:
Change the toString method on Range and NumericRange to be more concise.


src/library/scala/collection/immutable/Range.scala
```diff
@@ -401,18 +401,12 @@ extends scala.collection.AbstractSeq[Int]
    *  equals contract.
    */
 
-  override def toString() = {
-    val endStr =
-      if (numRangeElements > Range.MAX_PRINT || (!isEmpty && numRangeElements < 0)) ", ... )" else ")"
-    take(Range.MAX_PRINT).mkString("Range(", ", ", endStr)
-  }
+  override def toString = s"Range($start, $end, $step)"
```
sjrd: This does not take `isInclusive` into account.
I suggest delegating to the `description` method, which already implements a nice string representation for ranges.

stevorobs3: Thanks - sure I'll change that. That method isn't available in NumericRange though. Is it OK to just duplicate the description code?

src/library/scala/collection/immutable/NumericRange.scala
```diff
@@ -250,10 +250,7 @@ extends AbstractSeq[T] with IndexedSeq[T] with Serializable {
       super.equals(other)
   }
 
-  override def toString() = {
-    val endStr = if (length > Range.MAX_PRINT) ", ... )" else ")"
-    take(Range.MAX_PRINT).mkString("NumericRange(", ", ", endStr)
-  }
+  override def toString = "%d %s %d by %s".format(start, if (isInclusive) "to" else "until", end, step)
```
soc: I think this would be the perfect place to use the `f` interpolator.

Ichoran: Indeed, it would be, as that would catch that `start` and `end` are not printable as `%d`s.  (Oops--deleted previous wrong comment.)

stevorobs3: I'm having trouble coming up with a format that makes sense for both BigInt and BigDecimal.

I've been playing around with the f interpolator and I've come up with this:

``` scala
abstract class NumericRange[T]
  (val start: T, val end: T, val step: T, val isInclusive: Boolean)
  (implicit num: Integral[T])
...
override def toString = f"${num toDouble start}%.2f ${if (isInclusive) "to" else "until"} ${num toDouble end}%.2f by ${num toDouble step}%.2f"
```

But it feels wrong for BigInt to print out the decimal digits. Is there an easy way to know if the type contains Integral numbers?

Alternatively we could just print them as strings but that doesn't seem to format well (BigDecimal has way too many digits)

(It also seems odd that the class takes in an implicit num: Integral[T] when BigDecimal, etc aren't Integral but Fractional. Is this because NumericRange only needs an Integral[T] for its implementation?)

sjrd: I wouldn't even use the `f` formatter. Just use `s`. Let the numbers display themselves as it pleases them.

stevorobs3: OK - I can change it to use the s-interpolator, but it does print BigDecimals wierdly:

``` scala
println(Range.BigDecimal(new BigDecimal(1.3),new BigDecimal(100.3),new BigDecimal(1.3)))
```

prints out:
1.3000000000000000444089209850062616169452667236328125 until 100.2999999999999971578290569595992565155029296875 by 1.3000000000000000444089209850062616169452667236328125

SethTisue:
should the `by 1` be elided, since it's the default?

SethTisue:
I think there should be test coverage for the `to` case, and for what happens for a range like `10 to 0`

test/files/jvm/serialization.check
```diff
@@ -97,12 +97,12 @@ x = Queue(a, b, c)
 y = Queue(a, b, c)
 x equals y: true, y equals x: true
 
-x = Range(0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
-y = Range(0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
+x = 0 until 10 by 1
```
janekdb, line 100: If I was browsing a log file `Range(0 until 10 by 1)` would be of greater benefit.

test/files/jvm/serialization.check
```diff
@@ -97,12 +97,12 @@ x = Queue(a, b, c)
...
+y = 0 until 10 by 1
 x equals y: true, y equals x: true
 
-x = NumericRange(0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
-y = NumericRange(0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
+x = 0 until 10 by 1
+y = 0 until 10 by 1
```
janekdb, line 105: As a user I would prefer `NumericRange(0 until 10 by 1)`

som-snytt:
If I supply an `Integral[FiniteDuration]`, currently

```
> NumericRange(Duration.Zero, Duration(10, Seconds), Duration(1, Seconds))
NumericRange(0 days, 1 second, 2 seconds, [snip], 9 seconds)
```

what if

```
0 days until 9 seconds by 1 second of FiniteDuration
```

Just a thought.

`Range.Double` is interesting because it's really a `Range.BigDecimal` with a `.toDouble` view of values. But `0.0 until 1.0 by 0.1` is expressive. Wait, `BigDecimal` prints a trailing zero? So currently,

```
NumericRange(0.0, 0.05, 0.10, [snip])  // BigDecimal
NumericRange(0.0, 0.05, 0.1, [snip])   // Double
```

So maybe including the name of `T` in the string isn't crazy.

som-snytt:
I updated with my idea to specify the step type (which is the range type), and also whether the range is empty or exact.

https://github.com/scala/scala/pull/5175

szeiger:
Closing in favor of https://github.com/scala/scala/pull/5175 (which is based on this PR)

PR #4675, opened 2015-07-31
SI-9425 Leave Companion.apply if constructor is less accessible

retronym:
Calls to synthetic case class apply methods are inlined to the
underlying constructor invocation in refchecks.

However, this can lead to accessibility errors if the constructor
is private.

This commit ensures that the constructor is at least as accessible
as the apply method before performing this tranform.

I've manually checked that other the optimization still works in other
cases:

scala> class CaseApply { Some(42)  }
defined class CaseApply

```
scala> :javap -c CaseApply
Compiled from "<console>"
public class CaseApply {
  public CaseApply();
    Code:
       0: aload_0
       1: invokespecial #9                  // Method java/lang/Object."<init>":()V
       4: new           #11                 // class scala/Some
       7: dup
       8: bipush        42
      10: invokestatic  #17                 // Method scala/runtime/BoxesRunTime.boxToInteger:(I)Ljava/lang/Integer;
      13: invokespecial #20                 // Method scala/Some."<init>":(Ljava/lang/Object;)V
      16: pop
      17: return
}
```

PR #5002, opened 2016-03-01
SI-9546 Fix regression in rewrite of case apply to constructor call

retronym:
In SI-9425, I disabled the rewrite of `CaseClass.apply(x)` to
`new CaseClass(x)` if the constructor was was less accessible
than the apply method. This solved a problem with spurious
"constructor cannot be accessed" errors during refchecks for
case classes with non-public constructors.

However, for polymorphic case classes, refchecks was persistent,
and even after refusing to transform the `TypeApply` within:

```
CaseClass.apply[String]("")
```

It _would_ try again to transform the enclosing `Select`, a
code path only intended for monomorphic case classes. The tree has
a `PolyType`, which foiled the newly added accessibility check.

I've modified the guard to specifically exclude `PolyType` typed
trees (which only reach that code after we decline to transform
an enclosing `TypeApply`). I've also rewritten the check to use
`finalResultType` rather than `resultType`, which would have also
avoided the problem.

```
scala> case class CaseClass[A](a: A)
defined class CaseClass

scala> typeOf[CaseClass.type].decl(TermName("apply")).info
res0: $r.intp.global.Type = [A](a: A)CaseClass[A]

scala> res0.resultType
res1: $r.intp.global.Type = (a: A)CaseClass[A]

scala> res0.resultType.typeSymbol
res2: $r.intp.global.Symbol = <none>

scala> res0.finalResultType
res3: $r.intp.global.Type = CaseClass[A]

scala> res0.finalResultType.typeSymbol
res4: $r.intp.global.Symbol = class CaseClass
```

I've manually checked that the rewrite is still applied for a
generic case class with a public constructor.

```
% ./build-sbt/pack/bin/scalac -Xprint:refchecks test/files/run/t9546.scala 2>&1 | grep "def mkOpt"
      def mkOpt: foo.Opt[String] = Opt.apply[String]("");
/code/scala2 on ticket/9546*
% ./build-sbt/pack/bin/scalac -Xprint:refchecks test/files/run/t9546b.scala 2>&1 | grep "def mkOpt"
      def mkOpt: foo.Opt[String] = new foo.Opt[String]("");
```

If the case class was not also a value class, the bug still existed,
but never manifest as a crash. In `run/t9546c.scala`, the following
nonsensical tree was created in refchecks, but the problem was washed
away during erasure.

```
def mkOpt = new foo.Opt[A][String]("");
```

If `Opt` is a value class, `posterasure` needs to inspect and transform
this tree, and the crash ensued.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -1503,9 +1503,8 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
       }
     }
 
-    private def transformCaseApply(tree: Tree, ifNot: => Unit) = {
+    private def isSimpleCaseApply(tree: Tree): Boolean = {
       val sym = tree.symbol
-
       def isClassTypeAccessible(tree: Tree): Boolean = tree match {
         case TypeApply(fun, targs) =>
           isClassTypeAccessible(fun)
@@ -1513,31 +1512,26 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
           ( // SI-4859 `CaseClass1().InnerCaseClass2()` must not be rewritten to `new InnerCaseClass2()`;
             //          {expr; Outer}.Inner() must not be rewritten to `new Outer.Inner()`.
             treeInfo.isQualifierSafeToElide(module) &&
-            // SI-5626 Classes in refinement types cannot be constructed with `new`. In this case,
-            // the companion class is actually not a ClassSymbol, but a reference to an abstract type.
-            module.symbol.companionClass.isClass
-          )
+              // SI-5626 Classes in refinement types cannot be constructed with `new`. In this case,
+              // the companion class is actually not a ClassSymbol, but a reference to an abstract type.
+              module.symbol.companionClass.isClass
+            )
       }
 
-      val doTransform =
-        sym.isSourceMethod &&
+      sym.isSourceMethod &&
         sym.isCase &&
         sym.name == nme.apply &&
         isClassTypeAccessible(tree) &&
         !tree.tpe.resultType.typeSymbol.primaryConstructor.isLessAccessibleThan(tree.symbol)
+    }
 
-      if (doTransform) {
-        tree foreach {
-          case i@Ident(_) =>
-            enterReference(i.pos, i.symbol) // SI-5390 need to `enterReference` for `a` in `a.B()`
-          case _ =>
-        }
-        toConstructor(tree.pos, tree.tpe)
+    private def transformCaseApply(tree: Tree) = {
+      tree foreach {
+        case i@Ident(_) =>
+          enterReference(i.pos, i.symbol) // SI-5390 need to `enterReference` for `a` in `a.B()`
+        case _ =>
       }
-      else {
-        ifNot
-        tree
-      }
+      toConstructor(tree.pos, tree.tpe)
     }
 
     private def transformApply(tree: Apply): Tree = tree match {
@@ -1577,12 +1571,15 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
         // term should have been eliminated by super accessors
         assert(!(qual.symbol.isTrait && sym.isTerm && mix == tpnme.EMPTY), (qual.symbol, sym, mix))
 
-      transformCaseApply(tree,
+      if (isSimpleCaseApply(tree)) {
+        transformCaseApply(tree)
+      } else {
         qual match {
           case Super(_, mix)  => checkSuper(mix)
           case _              =>
         }
-      )
+        tree
+      }
     }
     private def transformIf(tree: If): Tree = {
       val If(cond, thenpart, elsepart) = tree
@@ -1706,7 +1703,10 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
 
           case TypeApply(fn, args) =>
             checkBounds(tree, NoPrefix, NoSymbol, fn.tpe.typeParams, args map (_.tpe))
-            transformCaseApply(tree, ())
+            if (isSimpleCaseApply(tree))
+              transformCaseApply(tree)
+            else
+              tree
 
           case x @ Apply(_, _)  =>
             transformApply(x)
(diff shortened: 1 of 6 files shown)


retronym:
IDE build failure: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/2200/

SethTisue:
I've opened https://github.com/scala/scala-dev/issues/96 on the integrate-ide regression

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1520,11 +1520,12 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
       }
 
       val doTransform =
+        !tree.tpe.isInstanceOf[PolyType] && // SI-9546 If we've bailed out of the transform for a TypeApply, don't try again for the enclosed Select
```
lrytz: i'd leave this line out given that the fix below is enough

retronym: This one actually makes more sense as the primary check. Otherwise, in the case of a public constructor, we are relying on the details of the transform (e.g. that parents are transformed before children).

lrytz: hmm, i don't see why this is, sorry..

retronym: I'll expand the internal docs tomorrow to help explain this.

retronym:
@lrytz Hopefully this commit progression will make a bit more sense. Includes some Scala 2.6 archeonostalgia, too!

```
% cat sandbox/test.scala
case class Foo(a: Any)

class Test {
  Foo("blerg")
}

% (java_use 1.6; ~/scala/2.6/bin/scalac -Xprint:typer  sandbox/test.scala ) | grep blerg
    Foo("blerg")

% (java_use 1.6; ~/scala/2.7/bin/scalac -Xprint:typer  sandbox/test.scala ) | grep blerg
    Foo.apply("blerg")
```

SethTisue:
community build: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/242/consoleFull

SethTisue:
@lrytz if you could review this sometime today, that'd be good — I'd like to get the 2.11.8 release train rolling again

lrytz:
I think we still need the `finalResultType` in https://github.com/retronym/scala/commit/1a59c55ed5c6a0d53acd0259090cf57328748451#diff-e2cfc09e4d7c753232666c6e31bd1926R1525

Example:
```
case class A private (x: Int)
case class B private (x: Int)(y: Int)

class C {
  def f = A(1)
  def g = B(1)(2)
}
```

With 2.11.7:
```
➜  sandbox git:(pr5002) scalac11 -version
Scala compiler version 2.11.7 -- Copyright 2002-2013, LAMP/EPFL
➜  sandbox git:(pr5002) scalac11 Test.scala
Test.scala:5: error: constructor A in class A cannot be accessed in class C
  def f = A(1)
          ^
Test.scala:6: error: constructor B in class B cannot be accessed in class C
  def g = B(1)(2)
          ^
two errors found
``


➜  sandbox git:(pr5002) ✗ git diff | cat
➜  sandbox git:(pr5002) qsc -version
Scala compiler version 2.11.8-20160302-221155-fe5bd09861 -- Copyright 2002-2016, LAMP/EPFL
➜  sandbox git:(pr5002) qsc Test.scala
Test.scala:6: error: constructor B in class B cannot be accessed in class C
  def g = B(1)(2)
          ^
one error found
```

lrytz:
I think we still need the `finalResultType` in https://github.com/retronym/scala/commit/1a59c55ed5c6a0d53acd0259090cf57328748451#diff-e2cfc09e4d7c753232666c6e31bd1926R1525

Example:

```
case class A private (x: Int)
case class B private (x: Int)(y: Int)

class C {
  def f = A(1)
  def g = B(1)(2)
}
```

With 2.11.7:

```
➜  sandbox git:(pr5002) scalac11 -version
Scala compiler version 2.11.7 -- Copyright 2002-2013, LAMP/EPFL
➜  sandbox git:(pr5002) scalac11 Test.scala
Test.scala:5: error: constructor A in class A cannot be accessed in class C
  def f = A(1)
          ^
Test.scala:6: error: constructor B in class B cannot be accessed in class C
  def g = B(1)(2)
          ^
two errors found
```

With fe5bd09861994734bc394813d069ea40c89d39de

```
➜  sandbox git:(pr5002) qsc -version
Scala compiler version 2.11.8-20160302-221155-fe5bd09861 -- Copyright 2002-2016, LAMP/EPFL
➜  sandbox git:(pr5002) qsc Test.scala
Test.scala:6: error: constructor B in class B cannot be accessed in class C
  def g = B(1)(2)
          ^
one error found
```

With `finalResultType`

```
➜  sandbox git:(pr5002) ✗ git diff | cat
diff --git a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
index 517271e..c19e11b 100644
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -1522,7 +1522,7 @@ abstract class RefChecks extends InfoTransform with scala.reflect.internal.trans
         sym.isCase &&
         sym.name == nme.apply &&
         isClassTypeAccessible(tree) &&
-        !tree.tpe.resultType.typeSymbol.primaryConstructor.isLessAccessibleThan(tree.symbol)
+        !tree.tpe.finalResultType.typeSymbol.primaryConstructor.isLessAccessibleThan(tree.symbol)
     }

     private def transformCaseApply(tree: Tree) = {
➜  sandbox git:(pr5002) qsc Test.scala
➜  sandbox git:(pr5002)
```

lrytz:
(Though that's not related to the fix of SI-9546, it's rather an improvement of #4675 that fixed the original issue SI-9425)

retronym:
> (Though that's not related to the fix of SI-9546, it's rather an improvement of #4675 that fixed the original issue SI-9425)

Good catch, I've pushed another commit with your test case that switches to `finalResultType`.

retronym:
@lrytz I've used this text:

```
Rewrite eligible calls to monomorphic case companion apply methods to the equivalent constructor call.

Note: for generic case classes the rewrite needs to be handled at the enclosing `TypeApply` to transform
`TypeApply(Select(C, apply), targs)` to `Select(New(C[targs]), <init>)`. In case such a `TypeApply`
was deemed ineligible for transformation (e.g. the case constructor was private), the refchecks transform
will recurse to this point with `Select(C, apply)`, which will have a type `[T](...)C[T]`.

We don't need to perform the check on the Select node, and `!isHigherKinded will guard against this
redundant (and previously buggy, SI-9546) consideration.
```

lrytz:
LGTM, thanks!

SethTisue:
community build of the merge commit: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/245/consoleFull (404 while Jenkins catches up)

SethTisue:
newer: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/246/consoleFull

PR #5003, opened 2016-03-01
New trait encoding using default methods [ci: last-only]

retronym:
Until now, concrete methods in traits were encoded with "trait implementation classes".
- Such a trait would compile to two class files
  - the trait interface, a Java interface, and
  - the implementation class, containing "trait implementation methods"
- trait implementation methods are static methods has an explicit self parameter. 
- some methods don't require addition of an interface method, such as private methods. Calls to these directly call the implementation method
- classes that mixin a trait install "trait forwarders", which implement the abstract method in the interface by forwarding to the trait implementation method.

The new encoding:
- no longer emits trait implementation classes or trait implementation methods.
- instead, concrete methods are simply retained in the interface, as JVM 8 default interface methods (the JVM spec changes in [JSR-335](http://download.oracle.com/otndocs/jcp/lambda-0_9_3-fr-eval-spec/index.html) paves the way)
- use `invokespecial` to call private or particular super implementations of a method (rather `invokestatic`)
  - in cases when we `invokespecial` to a method in an indirect ancestor, we add that ancestor redundantly as a direct parent. We are investigating alternatives approaches here.
- we still emit trait fowrarders, although we are [investigating](https://github.com/scala/scala-dev/issues/98) ways to only do this when the JVM would be unable to resolve the correct method using its rules for default method resolution.

Here's an example:

```
trait T {
  println("T")
  def m1 = m2
  private def m2 = "m2"
}
trait U extends T {
  println("T")
  override def m1 = super[T].m1
}
class C extends U {
  println("C")
  def test = m1
}
```

The old and new encodings are displayed and diffed here: https://gist.github.com/retronym/f174d23f859f0e053580
## Bootstrapping Plan
- [x] Update the bootstrap script to generate docs on the second pass of the modules bootstrap when a custom STARR is used.
- [x] Trigger [scala-2.12.x-integrate-bootstrap](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/), with `publishToSonatype=true` and `STARR_REF=`c8e6050 and `SCALA_VER_SUFFIX=-M3-dc9effe` ("Nuke implementation classes"): [scala-2.12.x-integrate-bootstrap/351](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/351/console)
- [x] Tag dc9effe ("   Remove manual mixins in JFunctionN") as [v2.11.0-M3-dc9effe](https://github.com/scala/scala/releases/tag/v2.12.0-M3-dc9effe)
- [x] Promote the resulting release (including modules) to Maven Central.
- [x] [Ask](https://github.com/rickynils/scalacheck/issues/225) @rickynils to publish a ScalaCheck release against this build. (We used a locally built version during the bootstrap script, but that won't be available when we use this as a STARR)
- [x] Rework ("Update STARR to freshly bootstrapped build") to set `starr.version=v2.11.0-M4-089b836` and remove the temporary build scaffolding to use the release-temp repository.
- [x] Merge this PR. dc9effe will be the only commit that doesn't build directly (it only builds with `bootstrap STARR_REF=HEAD^1 ..`)
- [x] [Update the references to 2.12.0-M3](https://github.com/scala/community-builds/pull/232) in the `scalaVersion` override for `scala-{xml, parser-combinators}` in the community build.

Variations to discuss:
- [x] Squash the last commits in this PR back into "Nuke implementation classes" before doing this.
- [x] Squash "Nuke implementation classes", "Remove manual mixins in JFunctionN", and "Update STARR to freshly bootstrapped build" before merging to 2.12.x, with a commit message that points to the tag v2.11.0-M4-089b836 which contains the finer grained commits. (**Verdict**: we don't need to do this.)
- [x] Does the tag v2.11.0-M3-089b836 make more sense? Ideally, Maven's ordering scheme would consider `2.11.0-M3 < XXXXX-089b836 < 2.11.0-M4`. Decision: Use `M3-`


retronym:
Note that the commits are ordered as follows:

```
⚡ git log --oneline 2.12.x..
083ee4f Update STARR to freshly bootstrapped build
3ecc141 Remove manual mixins in JFunctionN.
902bb54 Nuke trait implementation classes
```

retronym:
- [x] [Community Build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/335/)

```
[zinc] java.lang.NoClassDefFoundError: scala/Product$class
[zinc]  at scala.util.parsing.json.JSONArray.<init>(Parser.scala:103)
[zinc]  at scala.tools.nsc.doc.html.page.IndexScript.scala$tools$nsc$doc$html$page$IndexScript$$$anonfun$2(IndexScript.scala:45)
[zinc]  at scala.tools.nsc.doc.html.page.IndexScript$$Lambda$2929/1019011979.apply(Unknown Source)
[zinc]  at scala.collection.TraversableLike.scala$collection$TraversableLike$$$anonfun$3(TraversableLike.scala:234)
[zinc]  at scala.collection.TraversableLike$$Lambda$124/404499565.apply(Unknown Source)
[zinc]  at scala.collection.immutable.Map$Map3.foreach(Map.scala:176)
```

adriaanm:
- [x] [New community build run](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/338/console)
  - [x] Worked other than ScalaJS !
  - [x] ScalaJS support
    - [x] https://github.com/scala-js/scala-js/issues/2285, requesting requisite changes from ScalaJS.
    - [x] Voila: https://github.com/scala-js/scala-js/pull/2288
    - [x] New [community build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/340/console) with @sjrd's patch.
    - [x] Merge changes to ScalaJS

retronym:
- [ ] https://github.com/scala-js/scala-js/issues/2285, requesting requisite changes from ScalaJS

adriaanm:
LGTM!

retronym:
Note to admins: this PR can't be merged as is, we need to manually pull a few levers to publish to publish the new STARR and bootstrap.

adriaanm:
How about adding the https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/ repo instead of the release-temp one and using 2.12.0-73e4038-SNAPSHOT for STARR?

retronym:
Sounds like a good plan. I've just rebased, so I've triggered a build of the latest version of the "Nuke impl classes" commit, [scala-2.12.x-validate-publish-core/1504](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-validate-publish-core/1504/console). That can serve as the STARR for the subsequent commits.

We can revert the change to include pr-validation-snapshots as a Ant/SBT resolver after we re-STARR again (either immediately after merging this PR, or after the release of 2.12.0-M4)

retronym:
No wait I'm confusing myself again.

adriaanm:
Your comment seems fine, except that we can't remove the resolver until we have a release on maven. My comment should also be amended to use the sha of the commit before the one that introduces the new STARR, if we decide to drop the old approach of using scala-release-temp (instead of just tacking on another commit). I'm fine with keeping a few commits in history that are suboptimal, as long as they all build.

retronym:
We need to run the `bootstrap` job to get builds of the modules, it is not enough to use the result of `publishCore` as STARR. If we can get [that job](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/build?delay=0sec) to publish into `pr-validation-snapshots` rather than `release-temp`, however, life might be simpler.

lrytz:
In the interest of reproducible builds there shouldn't be revisions in the git history that depend on artifacts stored in transient repositories (pr-validation-snapshots / scala-release-temp). Is this the case here?

sjrd:
Stupid question: why not simply publish M4 with whatever new STARR you need, _then_ publish M5 with these changes, depending on M5?

Wasn't there some agreement at some point that builds should only ever use a published version as STARR?

retronym:
I agree it would be preferable to have the new STARR published to Maven. We could give this an arbitrary name `2.12.0-M4-<shortsha>`, if we'd prefer not to "use up" a milestone number.

adriaanm:
Right, forgot about the modules. Let's publish `M4-<shortsha>`.

adriaanm:
That'll also be more easily reproducible than my suggestion.

retronym:
[moved release plan to PR description]

adriaanm:
Sounds like a plan!

retronym:
I'm currently running the tip of this PR through the [release build](https://scala-ci.typesafe.com/job/scala-2.12.x-release-main/324/console). This is a preview of what M4 would look like. The release will end up in [scala-2.12.0-af98223-nightly.tgz](http://www.scala-lang.org/files/archive/nightly/2.12.x/scala-2.12.0-af98223-nightly.tgz) within an hour or so.

I'm going to use this to do a few sanity checks (e.g. for performance).

lrytz:
@retronym am i right that the fix for SD-86 (available here https://github.com/retronym/scala/pull/19) is not yet in this commits? that's fine with me, i just need to remember to submit it to scala/scala once this is merged, before m4.

retronym:
@lrytz I had already squashed that into d1ea397e748c8b111d219c86db16afc10e67cff0

lrytz:
the plan sounds good, squashing is not necessary IMO.

good point about the tag name, `2.12.0-M3-089b836` might be safer. otherwise `2.12.0-M4-089b836` might show up above `2.12.0-M4`, for example on search.maven.org.

lrytz:
> @lrytz I squashed that into d1ea397

the version you squashed in is an older one where SD-86 is not yet fixed, I force-pushed to https://github.com/retronym/scala/pull/19 at some point. but it's fine for this PR, I'll handle it in a follow-up.

retronym:
Okay! I'll put the plan in motion tomorrow using tag `v2.12.0-M3-089b836`. I'll also write a better high level description of the change in this PR description and in the main commit.

retronym:
Performance sanity check: comparing M3 vs http://www.scala-lang.org/files/archive/nightly/2.12.x/scala-2.12.0-af98223-nightly.tgz

This compares the aggregate of:
- the different work performed by the new compiler (e.g. no longer emitting separate impl classes)
- the fact that the new compiler is bootstrapped through the new encoding
- the fact that the new compiler is bootstrapped through @lrytz's optimizer improvements.

The numbers seems fairly consistent. They suggest are somewhere between on-par and slightly better.

I also compared against 2.11.8, similar story.

```
% time /Users/jason/Downloads/scala-2.12.0-af98223-nightly/bin/scalac -d /tmp  -J-Xmx2G src/library/**/*.scala
warning: there were 129 deprecation warnings; re-run with -deprecation for details
one warning found

real    0m38.038s
user    2m34.539s
sys 0m3.242s

% time ~/scala/2.12.0-M3/bin/scalac -d /tmp  -J-Xmx2G src/library/**/*.scala
warning: there were 129 deprecation warnings; re-run with -deprecation for details
one warning found

real    0m42.439s
user    2m59.309s
sys 0m3.546s

% time ~/scala/2.11.8/bin/scalac -d /tmp  -J-Xmx2G src/library/**/*.scala
warning: there were 129 deprecation warnings; re-run with -deprecation for details
one warning found

real    0m39.425s
user    2m44.753s
sys 0m3.826s
```

scottcarey:
The jars are ~ 12% smaller.

 :grinning:

retronym:
@scottcarey Thanks, I had meant to include that info. We can shave some more off by being less conservative about emitting trait forwarders (https://github.com/scala/scala-dev/issues/98), assuming we're happy with the tradeoffs. This will be around 5%.

adriaanm:
/nothingtoseehere (the main job was pending due to a jenkins issue -- i manually reran the test job)

adriaanm:
Congrats, @retronym -- it was a long slog, but so long implementation classes, at last! I don't think we need to squash those old commits (they are useful history). I think this is good to merge!

retronym:
> I don't think we need to squash those old commits (they are useful history)

Agreed. @adriaanm please pull the triggers (this one and https://github.com/scala/community-builds/pull/232)!

adriaanm:
Gladly! :confetti_ball:

smarter:
Congrats! @retronym: Are you aware of any differences between the 2.12 trait encoding and Dotty trait encoding? It'd be great if we could align them.

retronym:
@smarter Thanks! Our encoding of fields (and the way that they are desugared in the compilation pipeline) is still somewhat different. @adriaanm is working towards bringing that closer together in https://github.com/scala/scala/pull/5035 (although we don't have the intention of replicating the `initialize` methods, just the place that the desugaring happens in the pipeline.)

PR #5005, opened 2016-03-01
Document -Xxml:coalescing in scalac man page

janekdb:
The formatting style is based on -g and -target.

Test with `$ ant pack.doc && man build/pack/man/man1/scalac.1`

This is how it looks,

scalac-man-page-xxml


janekdb:
Review by @som-snytt, @heathermiller, @dickwall.

som-snytt:
LGTM

But isn't there a general problem with keeping it up-to-date? I see CPS in the phases list, for instance.

janekdb:
I agree there is a wider opportunity to place some limits on the gap that can open between implementation and documentation via derivation of documentation from in-use code. The initial question would be _"Are the various *Settings classes sufficient as-is to allows sections of the man pages models to be derived?"_ A first step towards this could be to script a consistency check which checks that each implemented option is documented and any documented option has an implementation.

This would surface some nuances. For example `-toolcp` is supported differently to `javaextdirs`. `-toolcp` is commented out in `ScalaSettings` but is present in a shell script.

```
$ grep -r -- -toolcp *
project/ScalaOptionParser.scala:  private def pathSettingNames = List("-bootclasspath", "-classpath", "-extdirs", "-javabootclasspath", "-javaextdirs", "-sourcepath", "-toolcp")
src/compiler/scala/tools/nsc/settings/ScalaSettings.scala:  /*val toolcp =*/ PathSetting("-toolcp", "Add to the runner classpath.", "")
src/compiler/scala/tools/ant/templates/tool-windows.tmpl:if not [%~1]==[-toolcp] (
src/compiler/scala/tools/ant/templates/tool-unix.tmpl:    -toolcp)
```

```
$ grep -r -- -javaextdirs *
project/ScalaOptionParser.scala:  private def pathSettingNames = List("-bootclasspath", "-classpath", "-extdirs", "-javabootclasspath", "-javaextdirs", "-sourcepath", "-toolcp")
src/compiler/scala/tools/nsc/settings/StandardScalaSettings.scala:  val javaextdirs =       PathSetting ("-javaextdirs", "Override java extdirs classpath.", Defaults.javaExtDirs)
```

Some progress could be made by restricting the scope of the deduplication to say just the `-X` options which appear to be more straightforwardly organised that some other classes of option. Even in this case there are subtleties. For example how would the default setting for `-Xxml` be known to differ between `2.11` and `2.12`?

SethTisue:
thank you Janek!

PR #5007, opened 2016-03-03
fix bootstrap regression (module docs weren't being published)

SethTisue:
reverts the "skip docs on first module build" part of
c4fc2fd42457a87fea9e7af94e0ba2f57e533854 since it was resulting in
module docs never being published at all, which then would later
cause release-website-archives to fail. see
https://github.com/scala/scala-dev/issues/89

review by @retronym


retronym:
LGTM, you are right to revert this to get things back to green.

Here's an alternative that we can test out that re-runs the full module builds (including scaladoc) in a second round when we're not publishing to Sonatype: https://github.com/scala/scala/compare/2.12.x...retronym:ticket/SD-89

PR #5008, opened 2016-03-03
Remove deprecated Predef.error

janekdb:
This was deprecated in 2.9.0.

I posted a question about the impact of this on `sbt` here: https://groups.google.com/forum/#!topic/sbt-dev/BNA692wX7Sk. Probably worth checking that thread in a few days - no objections so far but that could be because the post just hasn't be viewed widely enough yet.
## TODO
- [x] Wait for #5009 to be merged
- [x] Wait for #5049 to be merged

### DIFF
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -188,13 +188,7 @@ object Predef extends LowPriorityImplicits with DeprecatedPredef {
   /** @group utilities */
   @inline def locally[T](x: T): T  = x    // to communicate intent and avoid unmoored statements
 
-  // errors and asserts -------------------------------------------------
-
-  // !!! Remove this when possible - ideally for 2.11.
-  // We are stuck with it a while longer because sbt's compiler interface
-  // still calls it as of 0.12.2.
-  @deprecated("Use `sys.error(message)` instead", "2.9.0")
-  def error(message: String): Nothing = sys.error(message)
+  // assertions ---------------------------------------------------------
 
   /** Tests an expression, throwing an `AssertionError` if false.
    *  Calls to this method will not be generated if `-Xelide-below`
--- a/test/files/pos/t4365/a_1.scala
+++ b/test/files/pos/t4365/a_1.scala
@@ -9,7 +9,7 @@ trait SeqViewLike[+A,
   trait Transformed[+B] extends super[GenSeqViewLike].Transformed[B]
 
   abstract class AbstractTransformed[+B] extends Seq[B] with Transformed[B] {
-    def underlying: Coll = error("")
+    def underlying: Coll = sys.error("")
   }
 
   trait Reversed extends Transformed[A] with super[GenSeqViewLike].Reversed
--- a/test/files/pos/t4365/b_1.scala
+++ b/test/files/pos/t4365/b_1.scala
@@ -10,7 +10,7 @@ self =>
 
   trait Transformed[+B] {
     def length: Int = 0
-    def apply(idx: Int): B = error("")
+    def apply(idx: Int): B = sys.error("")
   }
 
   trait Reversed extends Transformed[A] {


retronym:
The build failure is checking whether SBT 0.12.4's compiler interface can still be compiled. You can bump `sbt.latest.version` to something more recent in `build.xml` to reset this test to something more modern.

janekdb:
@retronym - Thanks. I've updated to the latest sbt release, temporarily dropped the removal of `error` and pushed to see the outcome of that change in isolation.

Update: Downgraded to `0.13.9` to avoid this error with `0.13.11`,

`[get] Can't get http://dl.bintray.com/typesafe/ivy-releases/org.scala-sbt/compiler-interface/0.13.11/jars/compiler-interface-src.jar to /home/jdb/workspaces/public/scala/build/sbt-interface/0.13.11/src/compiler-interface-src.jar`

adriaanm:
Looks like they moved to http://dl.bintray.com/typesafe/ivy-releases/org.scala-sbt/compiler-interface/0.13.11/srcs/compiler-interface-sources.jar

adriaanm:
We've been talking for a while now about creating a dedicated project in scala/scala that has sbt's compiler-specific logic, and coming up with a stable api. FYI, here's one of our discussions on that: https://github.com/scala/scala/compare/2.12.x...adriaanm:sbt-api-consolidate#commitcomment-16551474

In that light, maybe we shouldn't worry too much about testing sbt in our project like this, and just remove this from our build. I believe PR validation's Scala IDE job builds all of sbt.

SethTisue:
@janekdb let's wait to merge this until we can get all the way to 0.13.11

janekdb:
Sure. 0.13.11 update started: #5049.

janekdb:
Review by @SethTisue, @adriaanm.

SethTisue:
LGTM

in general we are waiting until 2.13 to remove deprecated stuff, but this seems to me like a reasonable exception to make: it's been deprecated for donkey's years, only this one weird sbt issue was holding up removal, it's only a forwarder, and getting things out of Predef is always especially attractive.

in the immortal words of the Supreme Court in Bush v. Gore, "Our consideration is limited to the present circumstances"

SethTisue:
as I was merging this, @adriaanm raised a doubt about whether it's actually definitely OK to drop sbt 0.12 support in Scala 2.12 — it's conceivable those circles in the Venn diagram of the community might actually intersect? though one certainly hopes not...

but I thought about it more afterwards, and I don't think it matters, because sbt 0.12 uses Scala 2.9, which doesn't work on Java 8 anyway, and 2.12 is 8-only, so I think the circles _can't_ intersect

/cc @eed3si9n

eed3si9n:
LGTM too. I don't know anyone who uses sbt 0.12 actively, and I don't think it's unreasonable to tell them to upgrade to sbt 0.13+ if they want to use Scala 2.12.

SethTisue:
@eed3si9n thanks. so then the other question is, do you happen to know exactly when sbt 0.13 became compatible with this change? both Adriaan and I were looking at commits in the sbt history that replaced `error` with `sys.error`, but it wasn't totally obvious to us which commits were relevant. is only the `compile/interface` directory relevant? code in that directory defines its own method named `error`, so it isn't immediately obvious where it might actually be calling `Predef.error`.

SethTisue:
@adriaanm I now don't think this needs to be release-noted, even.

adriaanm:
Cool! Thanks for doubling down on the due diligence.

eed3si9n:
@SethTisue Yes https://github.com/sbt/sbt/tree/0.12/compile/interface should be the only relevant bit.
I was looking for it too, and found `def error(msg: String) = throw new RuntimeException(msg)` myself here - https://github.com/sbt/sbt/blob/0.12/compile/interface/API.scala#L23

PR #5009, opened 2016-03-04
Update sbt version to most recent that includes the compiler interface

janekdb:
compiler-interface-src.jar was available for 0.13.9 but not 0.13.10 or 0.13.11
at the time this commit was tested.

This in preparation for removing Predef#error which was deprecated in 2.9.0.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -461,7 +461,7 @@ TODO:
     </fail>
 
     <!-- Allow this to be overridden simply -->
-    <property name="sbt.latest.version"    value="0.12.4"/>
+    <property name="sbt.latest.version"    value="0.13.9"/>
 
     <property name="sbt.src.dir"           value="${build-sbt.dir}/${sbt.latest.version}/src"/>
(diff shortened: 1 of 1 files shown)


janekdb:
Review by @retronym, @eed3si9n.

janekdb:
This has past the CI build but locally I am seeing a failure,

```
$ rm -r build && ant build test
...
test.sbt:
[stopwatch] [quick.sbt-interface.timer: 0.229 sec]

bc.init:

bc.prepare:

BUILD FAILED
/home/jdb/workspaces/public/scala/build.xml:1545:
  java.lang.ClassCastException:
    org.codehaus.plexus.DefaultPlexusContainer cannot be cast to org.codehaus.plexus.PlexusContainer
```

Update: This also happens for me on the current `2.12.x` so it will be a local environment problem I have. I'll leave this note here in case someone might see this same local build problem.

retronym:
LGTM

PR #5012, opened 2016-03-05
SI-9687 - Fix buggy deserialization of immutable.HashMap.HashMap1

lgieron:
When scala.collection.immutable.HashMap.HashMap1 got deserialized, its 'kv'
field was errorneously left as 'null', which caused NullPointerException
in calls to 'merged' function. This commit changes the deserialization to set
'kv' to its proper value.

PR also contains a regression test for this issue.

### DIFF
--- a/test/junit/scala/collection/immutable/HashMap1Test.scala
+++ b/test/junit/scala/collection/immutable/HashMap1Test.scala
@@ -1,0 +1,29 @@
+package scala.collection.immutable
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import java.io.{ByteArrayOutputStream, ObjectOutputStream, ByteArrayInputStream, ObjectInputStream}
+
+@RunWith(classOf[JUnit4])
+class HashMap1Test {
+
+  @Test
+  def deserializesCorrectly() {
+  	val kv = (1->1)
+    val map = new HashMap.HashMap1(kv._1,kv._1.hashCode,kv._2,kv)
+
+ 	val baos = new ByteArrayOutputStream()
+    val oos = new ObjectOutputStream(baos)
+    oos.writeObject(map)
+    oos.close()
+    
+    val bais = new ByteArrayInputStream(baos.toByteArray())
+    val ois = new ObjectInputStream(bais)
+    val deserializedMap = ois.readObject().asInstanceOf[HashMap$HashMap1]
+    ois.close()
+    
+    assertEquals(kv, deserializedMap.kv)
+  }
+}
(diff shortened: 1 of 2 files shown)


test/junit/scala/collection/immutable/HashMap1Test.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import java.io.{ByteArrayOutputStream, ObjectOutputStream, ByteArrayInputStream, ObjectInputStream}
+
+@RunWith(classOf[JUnit4])
+class HashMap1Test {
+
+  @Test
+  def deserializesCorrectly() {
+  	val kv = (1->1)
```
janekdb, line 14: Format code.

test/junit/scala/collection/immutable/HashMap1Test.scala
```diff
@@ -0,0 +1,29 @@
...
+    val oos = new ObjectOutputStream(baos)
+    oos.writeObject(map)
+    oos.close()
+    
+    val bais = new ByteArrayInputStream(baos.toByteArray())
+    val ois = new ObjectInputStream(bais)
+    val deserializedMap = ois.readObject().asInstanceOf[HashMap$HashMap1]
+    ois.close()
+    
+    assertEquals(kv, deserializedMap.kv)
```
janekdb, line 27: This could be an opportunity to refactor serialisation test code into a utilities object. There's a few places where this is done. This package looks like a suitable location,

scala-testing

Here's one example of where that could be used,

scala-testing-example

I am not suggesting you expand the scope of this commit beyond moving the ser/des code into a shared location but it could be good to at least put the code there as part of this fix and then return in a later commit to replace other occurrences of the ser/des pattern with the common code. WDYT?

retronym:
> This commit changes the deserialization to set 'kv' to its proper value.

This commit changes `updated0`, which is used more widely than in deserialization. Can you explain in some more detail why you think the existing code used `null` there and why your change is necessary. Is there some other correctness problem outside of serialization?

An alternative fix for this bug (and, perhaps, similar ones) might be to instead switch to using the serialization proxy pattern, in we instead serialize an `Array[(K, V])`, and deserialization calls `HashMap.apply(...)` with this sequence. Arguably we should also do this in other spots like `immutable.Set`

lgieron:
@retronym Thank you for your comment, it prompted me to go a bit deeper into this issue. Here are my findings:
1. It seems that the `kv` is lazily initialized in HashMap1 on purpose (likely for performance/memory usage optimization). The lazy initialization is performed by `ensurePair` method. So, the original issue I reported is actually not a bug (as having `null` for `kv` does not mean the object is malformed).
2. However, there definitely is a bug in `merge` logic - when merging, there is no call to `ensurePair` in the call path and in result trying to `merge` HashMap1 objects  which don't have `kv` set throws `NullPointerException`. I've analysed code HashMap.scala file and it seems to me that `merge` is the only use case unable to handle objects with `null` `kv`.
I'll close this ticket and PR and create a new ticket and PR for the bug in `merge`.

lgieron:
I've created a new ticket and PR (as described in previous comment) for this issue:
https://issues.scala-lang.org/browse/SI-9688
https://github.com/scala/scala/pull/5017

PR #5013, opened 2016-03-05
Format collection conversions Scaladoc as code blocks instead of bullets

janekdb:
- Makes the list of conversions easier to scan
- Makes main comment formatting internally consistent

This is how the documentations looks with this change,

scaladoc-2 12-conversions-block-formatted

Before the change showing the use of bullets.

scaladoc-2 12-conversions-list-formatted

Ignore the package list discrepancy. I wasn't rebased when taking the first captures.

### DIFF
--- a/src/library/scala/collection/JavaConversions.scala
+++ b/src/library/scala/collection/JavaConversions.scala
@@ -16,12 +16,12 @@ import convert._
  *
  *    The following conversions are supported:
  *{{{
- *    scala.collection.Iterable <=> java.lang.Iterable
- *    scala.collection.Iterable <=> java.util.Collection
- *    scala.collection.Iterator <=> java.util.{ Iterator, Enumeration }
+ *    scala.collection.Iterable       <=> java.lang.Iterable
+ *    scala.collection.Iterable       <=> java.util.Collection
+ *    scala.collection.Iterator       <=> java.util.{ Iterator, Enumeration }
  *    scala.collection.mutable.Buffer <=> java.util.List
- *    scala.collection.mutable.Set <=> java.util.Set
- *    scala.collection.mutable.Map <=> java.util.{ Map, Dictionary }
+ *    scala.collection.mutable.Set    <=> java.util.Set
+ *    scala.collection.mutable.Map    <=> java.util.{ Map, Dictionary }
  *    scala.collection.concurrent.Map <=> java.util.concurrent.ConcurrentMap
  *}}}
  *    In all cases, converting from a source type to a target type and back
--- a/src/library/scala/collection/JavaConverters.scala
+++ b/src/library/scala/collection/JavaConverters.scala
@@ -19,14 +19,14 @@ import convert._
  *  Scala and Java collections using `asScala` and `asJava` methods.
  *
  *  The following conversions are supported via `asJava`, `asScala`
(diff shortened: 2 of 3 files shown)


janekdb:
This should not be merged until the issue around code block styling is resolved so the Scaladoc can be properly eyeballed :eyes: . The code blocks under 2.12 are no longer using a monospaced style. @felixmulder, is this unintentional?

scaladoc-2 12-pre-not-monospaced

Monospacing is absent with 4dccfbdd27729bc79c57530ae40a683fcf1649f7.

I don't have the commit id but back on 2016-Feb-18 code blocks did have monospacing and the blocked formatting was viable,

scaladoc-2 12--monospaced-a-while-back-2016-feb-18

felixmulder:
Check out the nightly built docs on scala-lang.org, my guess is that this issue is not present there. Rather the issue is with Firefox on localhost.

I could of course be wrong - this is my gut feeling.

I'm currently in Prague (beer, carbs, pork :beers:) - I'll check on this tomorrow or early this week. Cheers!

janekdb:
As you guessed no problem with the nightly build. Strange as this was working with FF/localhost when I first tried it. I'll try some other browser. :beers:

felixmulder:
So I'm only getting this for the material font (the one that provides us with the cool magnifying glass and whatnot). But apparently, it's a feature in firefox that disallows unsafe URI for font loading on localhost - as such, that is the only place where this is apparent.

There's a workaround - if you just want to know that everything's working and that's to set:

`security.fileuri.strict_origin_policy` to `true` in `about:config`. It's not something we'd want to advocate, but it gets the job done. I'll see if I can do anything about it during the week.

felixmulder:
One idea would be to provide a web-based fallback. In other words, if the font is not available locally (because of security restrictions in FF) then the font will be loaded from the web (Google's CDN or cloudflare), WDYT?

janekdb:
Thanks for the workaround - much appreciated.

Setting `security.fileuri.strict_origin_policy` to `false` in `about:config` restores the monospacing in the code blocks.

scaladoc-2 12-monospace-working-localhost-with-policy-off

My preference for Scaladoc would be for it to work offline without reconfiguring the browser but I'm not aware of what would be lost to achieve that goal so can't comment on the trade-offs yet.

janekdb:
Review by @felixmulder , @heathermiller, @dickwall.

felixmulder:
Well there is a third option - defining the typeface like this:

``` css
font-family: "Source Code Pro", "Monaco", monospace;
```

I will do this in a separate PR (#5016) , this will make the typeface fall back on `Monaco` then the default `monospace` typeface if the fallback typeface does not exist. Should make for a pleasant experience without needing online resources. Perhaps `Ubuntu Mono Regular` as well for looking nice on most Linux dists. Is there a nice looking default Windows monospaced typeface? I haven't touched Windows in 10 years (bless).

Regarding the review, it looks fine to me but - I'll leave that to the authorities @heathermiller and @dickwall :+1: 

**EDIT:** also, I'm not having any issues with FF (44.0.2) on OS X

janekdb:
This is ready for review.

SethTisue:
LGTM, thanks Janek!

PR #5016, opened 2016-03-07
Add fallback typefaces for monospaced sections in scaladoc

felixmulder:
The PR adds fallback typefaces to all monospaced areas. This should ensure that users with some versions of firefox get a good looking monospaced typeface despite firefox's local filesystem security rules.

This commit will also fix the `Filter All Members` input appearing if
there are no members - please see root for an example.

There's also a small CSS fix that addresses a vertical scrollbar on
smaller viewports (footer was not responsively resized).

preview: https://dl.dropboxusercontent.com/u/358427/scaladoc-monosp/index.html
review: @VladUreche

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.css
+++ b/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.css
@@ -40,7 +40,7 @@ body {
   text-align: center;
   color: #858484;
   bottom: 0;
-  height: 20px;
+  min-height: 20px;
   margin: 0 1em 0.5em;
 }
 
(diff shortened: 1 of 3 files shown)


VladUreche:
LGTM, thanks @felixmulder

felixmulder:
Thanks @VladUreche :)

src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.css
```diff
@@ -593,7 +593,7 @@ div#definition > h4#signature > span.modifier_kind > i.unfold-arrow,
   background-color: #fff;
   margin: 5px 0;
   display: block;
-  font-family: "Source Code Pro";
+  font-family: "Source Code Pro", "Monaco", "Ubuntu Mono Regular", "Lucida Console", monospace;
```
janekdb, line 596: Yep, Lucida Console for Windows. Seemed to always be available. Always the first change I make in Git Bash.

lucida-console-regular

PR #5017, opened 2016-03-07
SI-9688 Make merge in immutable HashMap1 work with null kv.

lgieron:
The kv field of scala.collection.immutable.HashMap.HashMap1 can be null. This
commit corrects the behavior of updated0 (which is on call path for merged) to
work in such cases, instead of throwing NPE.

Commit contains regression test.

### DIFF
--- a/src/library/scala/collection/immutable/HashMap.scala
+++ b/src/library/scala/collection/immutable/HashMap.scala
@@ -197,7 +197,7 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
           if (this.value.asInstanceOf[AnyRef] eq value.asInstanceOf[AnyRef]) this
           else new HashMap1(key, hash, value, kv)
         } else {
-          val nkv = merger(this.ensurePair, if(kv != null) kv else (key->value))
+          val nkv = merger(this.ensurePair, if(kv != null) kv else (key, value))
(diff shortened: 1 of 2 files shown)


SethTisue:
review by @retronym and/or @Ichoran ?

lgieron:
Anyone willing to review this? This fixes a serious bug in core functionality.

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -197,7 +197,7 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
           if (this.value.asInstanceOf[AnyRef] eq value.asInstanceOf[AnyRef]) this
           else new HashMap1(key, hash, value, kv)
         } else {
-          val nkv = merger(this.kv, kv)
+          val nkv = merger(this.ensurePair, if(kv != null) kv else (key->value))
```
lrytz: since this is relatively low-level code, could you create the pair as `(key, value)` instead of using `->`?

lrytz:
sorry this took so long..! other than my tiny remark, this looks good.

lrytz:
LGTM. the IDE failure is expected, it would go away if you'd rebase on current 2.11.x.

lgieron:
Cool, thanks for checking this out!

PR #5018, opened 2016-03-07
Remove mention of deprecated I/O methods from Predef main comment

janekdb:
The read\* functions are deprecated so should not be highlighted.

### DIFF
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -28,13 +28,11 @@ import scala.io.StdIn
  *  constructors ([[scala.collection.immutable.::]] and
  *  [[scala.collection.immutable.Nil]]).
  *
- *  === Console I/O ===
- *  Predef provides a number of simple functions for console I/O, such as
- *  `print`, `println`, `readLine`, `readInt`, etc. These functions are all
- *  aliases of the functions provided by [[scala.Console]].
+ *  === Console Output ===
+ *  For basic console output, `Predef` provides convenience methods [[print(x:Any* print]] and [[println(x:Any* println]],
+ *  which are aliases of the methods in the object [[scala.Console]].
  *
  *  === Assertions ===
- *
  *  A set of `assert` functions are provided for use as a way to document
  *  and dynamically check invariants in code. Invocations of `assert` can be elided
  *  at compile time by providing the command line option `-Xdisable-assertions`,
@@ -304,9 +302,37 @@ object Predef extends LowPriorityImplicits with DeprecatedPredef {
 
   // printing -----------------------------------------------------------
 
+  /** Prints an object to `out` using its `toString` method.
+   *
+   *  @param x the object to print; may be null.
+   */
   def print(x: Any) = Console.print(x)
+
+  /** Prints a newline character on the default output.
+   */
   def println() = Console.println()
+
+  /** Prints out an object to the default output, followed by a newline character.
+   *
+   *  @param x the object to print.
+   */
   def println(x: Any) = Console.println(x)
+
+  /** Prints its arguments as a formatted string to the default output,
+   *  based on a string pattern (in a fashion similar to printf in C).
+   *
+   *  The interpretation of the formatting patterns is described in
+   *  <a href="" target="contentFrame" class="java/util/Formatter">
+   *  `java.util.Formatter`</a>.
+   *
+   *  Consider using the [[scala.StringContext.f f interpolator]] as more type safe and idiomatic.
+   *
+   *  @param text the pattern for formatting the arguments.
+   *  @param args the arguments used to instantiating the pattern.
+   *  @throws java.lang.IllegalArgumentException if there was a problem with the format string or arguments
+   *
+   *  @see [[scala.StringContext.f StringContext.f]]
+   */
   def printf(text: String, xs: Any*) = Console.print(text.format(xs: _*))
 
   // views --------------------------------------------------------------


soc:
I'd probably drop printf too, considering we have the `f` interpolator.

janekdb:
`printf` removed from main comment.

If the aim is to move attention to the `f` interpolator then deprecation of `Predef.printf` needs to be considered (in a future commit).

Is that too strong? I wonder how much use `printf` gets. Enough to justify remaining in `Predef`?

som-snytt:
I never use `Predef.printf`, even though I call what I do `printf`, out of habit.

Sorry to bikeshed, or sorry to raze the shed, but the whole paragraph is now wrongish. 

```
For basic console output, `Predef` provides convenience methods `print` and `println`,
which are aliases of the methods in the object `scala.Console`.
```

If anybody knows how to turn `print` into a link, it should be one. Brackets? Bracken?

Probably the section title should be `Console Output` or something innocuous.

In a separate PR, really the whole doc could be rewritten, to point out `any2stringadd` and how to disable, `???`, etc. Also mention the compiler option `-Yno-predef`, maybe at the end in a footnote or implementation note.

janekdb:
I'll linkify `print` and `println` shorty but in the meanwhile could you be more specific about how the paragraph is wrongish? Even provide a full rewrite of the paragraph it that's easier.

som-snytt:
My comment did provide a rewrite in triple-backticks.

janekdb:
@som-snytt - Used your text. Decided to copy over the method Scaladoc from `Console` instead of making the method mentions links. I think this is more likely to benefit a user in a hurry.

This is how it looks,

scaladoc-2 12-predef-no-read-methods-main-comment

scaladoc-2 12-predef-no-read-methods-methods

som-snytt:
I actually meant links to `Predef.println` from the text, which is even more useful now that you have added doc there. :)  I'm thinking of the new user who doesn't know to tab to the search box etc.  I suggest that link only if you really want to see your face reflected by the high polish of the PR; I don't mean to nit-pick; and the other methods referenced up there don't have links. It looks great.

janekdb:
Links have been added.

@felixmulder - There is a cosmetic improvement that could be made here. When the link is clicked the page scrolls down positioning the method at the top but it's hidden by the search bar.

This video shows the above scenario: https://youtu.be/IgDnakzj4Lg.

@som-snytt - I'm now imagining a harmonious world where `double2Double` and `long2Long` stand side-by-side,  where `shortWrapper` and `floatWrapper` jostle along lock step and `require` and `assume` are best buddies all facilitated by our good friend `@group`. I ask a simple question: _Is Grouptopia too much to ask_? The answer comes back: _That's a separate PR_.

felixmulder:
Thanks @janekdb, I've put it on the to-do here: https://github.com/scala/scala-dev/issues/84

Fever and lots of tea atm, will have a look sometime next week :)

janekdb:
@som-snytt The note about `f` being idiomatic has been added. Over to reviewers now.

src/library/scala/Predef.scala
```diff
@@ -304,9 +302,38 @@ object Predef extends LowPriorityImplicits with DeprecatedPredef {
 
   // printing -----------------------------------------------------------
 
+  /** Prints an object to `out` using its `toString` method.
+    *
+    *  @param x the object to print; may be null.
+    */
```
som-snytt: The asterisk should align with the first of the double stars. See the top comment. Javadoc style (see second comment in source) seems to sneak by, but alt-Scaladoc style is out of favor. I'm not the comment cop, but since this is `Predef`, uniformity is worth enforcing. Is there a SO Q&A, "How many ways are there to write Scaladoc comments in Scala?" Luckily, April 1 is just around the corner.

janekdb:
Alignment problems fixed. IDEA being overly helpful and reformatting copy/paste. Grrrr!

With that blemish eradicated the time has come for a fresh dusting of fingerprint power. Have we yet commissioned the perfect crime?

som-snytt:
LGTM

I might have put guidance ("Consider using something else") after the method description; but that's a matter of emphasis.

janekdb:
Moved the guidance to after the method description.

som-snytt:
You are the Terminator! "He never gets tired, he'll never stop coming after you!" Actually, the quote I'm thinking of is T2, "It would never stop, it would never leave him, it would never say it was too busy to make one more doc change in a PR."  I amended that last part. Well, Sarah Connor goes on to say, "In an insane world, it was the sanest choice."

src/library/scala/Predef.scala
```diff
@@ -304,9 +302,37 @@ object Predef extends LowPriorityImplicits with DeprecatedPredef {
...
   def println(x: Any) = Console.println(x)
+
+  /** Prints its arguments as a formatted string to the default output,
+   *  based on a string pattern (in a fashion similar to printf in C).
+   *
+   *  The interpretation of the formatting patterns is described in
+   *  <a href="" target="contentFrame" class="java/util/Formatter">
+   *  `java.util.Formatter`</a>.
+   *
+   *  Consider using the [[scala.StringContext.f f interpolator]] as more idiomatic.
```
SethTisue: suggest: "more typesafe and idiomatic"

janekdb:
Made the "type safe and" improvement. Thanks!

SethTisue:
thank you, Janek!

src/library/scala/Predef.scala
som-snytt: Now more Typomatic!

Is it too late to change from lightbend to typomatic?

SethTisue:
note that discussion on the possible deprecation of `Predef.println` is at https://groups.google.com/forum/#!topic/scala-user/1NLGIvVbQdg

PR #5020, opened 2016-03-08
[WIP] add exception filtering capability to scala.util.Try

axos88:
This PR adds the ability to specify which throwables should or should not be catched and turned into a Failure. This enables us to catch Errors (which have been statically considered fatal previously, and always bubbled up), and handle them in the same monadic way as we could handle NonFatal exceptions previously.

Please review the PR, keepeng an eye out for performance considerations.

### DIFF
--- a/src/library/scala/util/Try.scala
+++ b/src/library/scala/util/Try.scala
@@ -202,13 +202,40 @@ sealed abstract class Try[+T] extends Product with Serializable {
 }
 
 object Try {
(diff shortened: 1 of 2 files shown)


axos88:
Signed the CLA, shouldn't it turn green by itself?

axos88:
Refactored to avoid limitation SI-2383

axos88:
Too many design possibilities, closing to discuss them.

PR #4848, opened 2015-11-17
Improve annotation support

soc:
- Drop Java requirement for declaring "real" annotations
- Read Java annotations from Java source files to set
  Retention information correctly
- Introduce {Runtime,Constant}Annotation

PR #5022, opened 2016-03-08
SI-3236 try to infer constant types for simple final static literal java expressions

jrudolph:
In a mixed Java/Scala project this will allow simple final static constants to be used in annotations from both Java and Scala. The workaround would be compiling the constants in a separate compilation run (which btw will prevent the original error from appearing in a usual developer setup involving separate compilation).

This is not yet finished, I just wanted to put that here for Jenkins to ponder upon it and to get some feedback if this would be a possible way to go. Of course, it still isn't as powerful as in Java where real constants can reference each other and would still be constants to be used in annotations (What would be necessary to get there? @soc told me something about constant folding / annotations that might have been related but I cannot remember the details).

I guess the real thing would have to be targeted against 2.12.x, I just put it on 2.11.x for now to be able to test it immediately against existing code.

/cc @magro


soc:
Hey @jrudolph! Thanks for your work! @retronym just helped me to get the annotation improvements unstuck, and I'll rebase your changes on top of it.

Do you have any test cases that show before vs. after?

I think the biggest issue will be that arrays of constant values also count as constants (although this might be restricted to array initializer literals right in the annotation and not references to one ... would have to try that) and constants which are constants because they only refer to other constants and arithmetic operations.

Nevertheless, I think it would make sense to integrate your improvements, and maybe bail out with a warning and not write the annotation if scalac can not figure out that something is in fact a constant.

jrudolph:
> Do you have any test cases that show before vs. after?

Not yet here in the PR but in a separate project, I'll have to extract them from. I'll do that when I find time this week.

> I think the biggest issue will be that arrays of constant values also count as constants (although this might be restricted to array initializer literals right in the annotation and not references to one ... would have to try that) and constants which are constants because they only refer to other constants and arithmetic operations.

Yes, for the first steps I'd try to support cases which can be supported syntactically.

In which branch are your changes?

soc:
@jrudolph https://github.com/scala/scala/pull/4848

lrytz:
@jrudolph i think this looks promising and would be very valuable to get in, do you have time to push this forward?

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -567,10 +567,49 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
+          def lit(negate: Boolean): Tree =
+            if (in.lookaheadToken == SEMI)
+              in.token match {
+                case INTLIT if tpt.tpe == IntTpe => constantTpe(in.intVal(negate).toInt)
+                case LONGLIT if tpt.tpe == LongTpe => constantTpe(in.intVal(negate))
+                case FLOATLIT if tpt.tpe == FloatTpe => constantTpe(in.floatVal(negate).toFloat)
+                case DOUBLELIT if tpt.tpe == DoubleTpe => constantTpe(in.floatVal(negate))
+                case STRINGLIT =>
+                  tpt match {
+                    case Ident(TypeName("String")) => constantTpe(in.name.toString)
```
lrytz, line 602: do we need to check the type here? if there's a string literal on the rhs its type can only be `String`. The test here would not cover qualified references to `String`.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -567,10 +567,49 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
+          finally in.nextToken()
+
+        if (mods.hasFlag(Flags.STATIC) && mods.isFinal) {
+          def lit(negate: Boolean): Tree =
+            if (in.lookaheadToken == SEMI)
+              in.token match {
+                case INTLIT if tpt.tpe == IntTpe => constantTpe(in.intVal(negate).toInt)
+                case LONGLIT if tpt.tpe == LongTpe => constantTpe(in.intVal(negate))
+                case FLOATLIT if tpt.tpe == FloatTpe => constantTpe(in.floatVal(negate).toFloat)
+                case DOUBLELIT if tpt.tpe == DoubleTpe => constantTpe(in.floatVal(negate))
```
lrytz, line 599: i guess we'd have to complete the list of primitive types here?

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -567,10 +567,49 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
+                    case _ => tpt1
+                  }
+                case _ => tpt1
+              }
+            else tpt1
+
+          in.token match {
+            case MINUS =>
+              in.nextToken()
+              lit(negate = true)
```
lrytz, line 612: could be extended to support `public static final int ONE = -(-1);` - not that it's all that common..

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -567,10 +567,49 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
+            case other => lit(negate = false)
+          }
+        } else tpt1
+      }
+
+      val tpt2: Tree =
+        if (in.token == EQUALS && !mods.isParameter)
+          try optConstantTpe() finally skipTo(COMMA, SEMI)
```
lrytz: somehow an elegant pattern, but it's also a bit confusing because it makes me thing there are exceptions involved. and it generates handlers in the bytecode.

jrudolph:
What would be the deadline to get this into 2.11.9?

lrytz:
According to https://github.com/scala/scala/milestones, 2.11.9 is currently dated Sept 30, so there would be time.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
lrytz, line 602: I see now that we'll miss a check, `public static final NotString s = "hi"` would pass.

jrudolph:
I updated this with more tests and a more complete implementation that spans all primitive types.

More "complex" expressions than literals and simple negation are still not supported. IMO even if support for simple arithmetic on literals would be nice it isn't strictly necessary because the programmer can do the constant folding manually as an easy workaround.

In current Scala, you cannot define constants in Java or Scala and use them in annotations from both Scala and Java (e.g. in a migration scenario from Java code to Scala code). This state seems to be enough to unblock this situation by allowing you to define constants in Java files and use them both from Java and Scala in annotations in the same (mixed) compilation run. 

It would be interesting to support support references to other constants and arithmetic on them. This would actually increase flexibility in organizing constants. However, this seems to be much harder to do. scalac would have to do actual constant folding on Java code to achieve this (right-hand sides of definitions are currently skipped in scalac's Java parser, etc.).

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
jrudolph, line 602: To pass javac, I think it could only be `public static final Object s = "hi"` which would be narrowed to a constant String from the view of Scala without the check. This would probably be ok in most cases but those where overloads or implicits would be picked differently for Object vs String. So, I guess it makes sense to keep it somewhat conservative.

Wrt qualified reference, yes, right, I guess we should support them. There's only one possible way to qualify them in Java which would be `java.lang.String`, right? Shouldn't be hard to support that here as well.

soc:
@jrudolph Are you at ScalaDays Berlin? Maybe we could look into what's missing from the larger annotation improvement change, address the remaining issues (maybe with the help of some Scala compiler devs), integrate your improvements and push this across the finish line?

As far as I remember there was only a minor issue remaining but I couldn't figure out what was causing it.

lrytz:
re-submitted as #5487

PR #5026, opened 2016-03-09
upgrade to newer scala-xml (namely 1.0.5)

SethTisue:
there shouldn't be user-visible changes since 1.0.4, but regardless,
it's good to depend on the latest

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -25,7 +25,7 @@ scala.binary.version=2.11
 scala.full.version=2.11.7
 
 # external modules shipped with distribution, as specified by scala-library-all's pom
-scala-xml.version.number=1.0.4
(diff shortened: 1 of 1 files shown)


SethTisue:
integrate-ide failure is unrelated; it's being tracked at https://github.com/scala/scala-dev/issues/97

SethTisue:
review by @lrytz ?

lrytz:
LGTM!

SethTisue:
1.0.6 was released in September 2016. there aren't reports of regressions, but maybe most people aren't using the new version, either. it's a judgment call, but I would lean towards just sticking with 1.0.5 for 2.11.9

PR #5034, opened 2016-03-11
Organise Predef methods into Scaladoc groups

janekdb:
This change uses @group\* tags to declutter Predef from a Scaladoc point of view.

It would be excellent to have feedback on any aspect of this but in particular the group descriptions which where oddly the most thought provoking part of this. It's difficult to be concise and useful!

> Not in scope for this PR but Predef is a bit odd because it contains methods that a user should know about and have documentation for, and other methods and things which live here to tie other parts of the system together. Maybe there should be a `InternalsDef` where these chaps can live out of view.

Preview of the page is shown below.
## V4

By squinting carefully its possible to see the group descriptions. Click the image to go large.

predef-with-groups-v4

### DIFF
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -65,6 +65,49 @@ import scala.io.StdIn
  *  are provided for the "widening" of numeric values, for instance, converting a
  *  Short value to a Long value as required, and to add additional higher-order
  *  functions to Array values. These are described in more detail in the documentation of [[scala.Array]].
+ *
+ * @groupname utilities Utility Methods
+ * @groupprio utilities 10
+ *
+ * @groupname assertions Assertions
+ * @groupprio assertions 20
+ * @groupdesc assertions These methods support program verfication and runtime correctness.
+ *
+ * @groupname console-output Console Output
+ * @groupprio console-output 30
+ * @groupdesc console-output These methods provide output via the console.
+ *
+ * @groupname type-constraints Type Constraints
+ * @groupprio type-constraints 40
+ * @groupdesc type-constraints These entities allows constraints between types to be stipulated.
+ *
+ * @groupname aliases Aliases
+ * @groupprio aliases 50
+ * @groupdesc aliases These aliases bring selected immutable types into scope without any imports.
+ *
+ * @groupname conversions-string String Conversions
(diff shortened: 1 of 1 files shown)


SethTisue:
This looks marvelous. I (shamefacedly) admit I didn't even know Scaladoc _had_ this feature.

Someone on Gitter said you can attach a "description" to each group? If that's true, that sounds like a good way to indicate that some of the groups contain methods that are not intended to be called directly from user code.

janekdb:
@groupdesc allows descriptions to be added. Suggestions for a sentence or two for each group welcomed.

janekdb:
Added some @groupdesc tags - screenshot updated.They are merely placeholders at this point so please lean in with suggestions.

janekdb:
Review by @SethTisue, @som-snytt, @felixmulder, @heathermiller,

SethTisue:
LGTM (and can always be tweaked further, anyway, if anyone sees any further improvements to be made)

VladUreche:
:+1: :100: :tada:

> This looks marvelous. I (shamefacedly) admit I didn't even know Scaladoc had this feature.

I must shamefacedly admit I never came around to documenting it :(

som-snytt:
I forgot to say this looks great.

janekdb:
Shamefacedly I have added a _groups_ TODO item to the Scaladoc documentation issue but not have not yet done it: https://github.com/scala/scala-lang/pull/394.

Although I do have my notes on @groups\* which without too much of a shameface could be described as not unadjacent to more than marginally useful.

I also wonder if we could take a systematic view to applying some carefully selected groups to collections. Transformers, Views, Traversers, Mutators??? I seem to remember seeing something like this somewhere else but don't recall where now.

VladUreche:
> Although I do have my notes on `@groups*` which without too much of a shameface could be described as not unadjacent to more than marginally useful.

Well, that's a good start nevertheless!

> I also wonder if we could take a systemic view to applying some carefully selected groups to collections. Transformers, Views, Traversers, Mutators??? I seem to remember seeing something like this somewhere else but don't recall where now.

Yes, in theory (modulo possible bugs) the groups attribute is inherited, so we only need to add it in one place :)

PR #5035, opened 2016-03-12
WIP New traits lambda [ci: last-only]

adriaanm:
traits now do the lambada without fields


adriaanm:
```
class MethodNesting {
  def method2method(a: Int) = {
    def intermediate = {
      def localmethod = a + 1
      localmethod
    }
    intermediate
  }
}

trait TraitNesting { def foo(x: Int) = { trait Nested { def bar = x }; (new Nested{}).bar } }

class ClassUseOuter {
  def local(x: Int) = {
    class StoreCapture {
      class AccessViaOuter {
        def y = x
      }
    }
    val s = new StoreCapture
    val via = (new s.AccessViaOuter)
    // TODO: use reflection to ensure via's class has only an $outer field
    via.y
  }
}

class ConstructorCall(val x: Any)
object ConstructorCall {
  def m(UGH: Any): ConstructorCall = {
    class X extends ConstructorCall(Option(null).getOrElse(UGH))
    new X
  }
}

object Test extends App {
//  def main(args: Array[String]): Unit = {
    assert((new MethodNesting).method2method(1) == 2)
    assert((new TraitNesting {}).foo(2) == 2)
    assert((new ClassUseOuter).local(2) == 2)
    assert(ConstructorCall.m(2).x == 2)
//  }
}

trait EarlyInitImplRestriction {
  trait Tree {
    def orElse(alt: => Tree): Tree = ???
  }
  trait MacroContext {
    val bla: Tree
    def meh: Tree
  }
  def macroContext(local: Tree): MacroContext = {
    new {
      // the anon class in bla captures local
      // note that bla's owner is the anon subclass of MacroContext
      // we'll need to mark `local` free in things nested in the ctor (premature self weirdness)
      val bla = local orElse local
    } with MacroContext {
      def meh = local
    }
  }
}

```

adriaanm:
The test failure shows we of course can't escape traits that capture locals that easily. The capture could happen in the initializer or, equivalently, the RHS of a val.

adriaanm:
More specifically, dotty's approach of pushing down captures in traits to its methods (to avoid adding fields for them) doesn't work for us because our trait initialization works differently. (Dotty's lifts RHS of fields into methods, which can then get the captured locals via arguments.)

For us, to compile something like this

```
object TraitValCapturing {
  def bar(ev: Any) = {
    trait Y { val ev2 = ev }

    (new Y{}).ev2
  }
}
```

we'll need to have a getter for `ev` in `Y` and implement it in each subclass of `Y` (we see all those because Y is local).

PR #5037, opened 2016-03-13
Fix Scaladoc links in scala.io.Source

janekdb:
The link to `iter` will work if the restriction to showing only public members
is lifted in the visibility section.


adriaanm:
LGTM

PR #5039, opened 2016-03-14
Scaladoc: change behavior of clear-button on input fields

felixmulder:
This commit changes the behavior to only remove the "x" if there is no
input in the input field. The old behavior was to bind the blur event to
be fired later - but this had the drawback of not working properly on
fast hardware as the timeout would actually beat the click event.

preview: https://dl.dropboxusercontent.com/u/358427/scaladoc-x-button/scala/Boolean.html
review: @VladUreche


VladUreche:
Works for me :)

PR #5042, opened 2016-03-15
SI-9539 Specify charset when reading ScalaSignatures, ...

soc:
... without it we would use the platform's default charset.

### DIFF
--- a/src/reflect/scala/reflect/runtime/JavaMirrors.scala
+++ b/src/reflect/scala/reflect/runtime/JavaMirrors.scala
@@ -613,7 +613,7 @@ private[scala] trait JavaMirrors extends internal.SymbolTable with api.JavaUnive
         loadBytes[String]("scala.reflect.ScalaSignature") match {
           case Some(ssig) =>
             info(s"unpickling Scala $clazz and $module, owner = ${clazz.owner}")
-            val bytes = ssig.getBytes
+            val bytes = ssig.getBytes(java.nio.charset.StandardCharsets.UTF_8)
             val len = ByteCodecs.decode(bytes)
             assignAssociatedFile(clazz, module, jclazz)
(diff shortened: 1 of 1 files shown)


soc:
What's the best way to test this?

retronym:
@soc does it pass the manual test I used https://issues.scala-lang.org/browse/SI-9539?focusedCommentId=73556&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-73556 ?

I don't think we have a way to set startup JVM options in partest or our JUnit suite for a particular test. It might be a stretch, but it would be interesting to have a nightly build that switched used something esoteric like `-Dfile.encoding=Cp1047` as the JVM options (including in forked JVMs) and runs the whole build and test suite.

soc:
@retronym I just tested it using your exmaple, and the change results in the right result.

retronym:
LGTM

velkan1984:
I am a Mainframe guy and just trying to get the Apache Spark for z/OS installed on my mainframe.
When I run the ./bin/spark-shell I am getting the following error with a 
"Exception in thread "main" scala.reflect.internal.MissingRequirementError: error while loading package, Scala signature package has wrong version expected: 5.0 found: 45.0 in scala.package" and get some lines of trace.

I tried the options of   -Dfile.encoding=Cp1047   and  -Dfile.encoding=UTF8 on the spark-env.sh conf parms. (eg: SPARK_DAEMON_JAVA_OPTS="-Dfile.encoding=Cp1047") .
But still getting the errors.
Can you help me get over this issue?

Warm Regards,
Velkan

SethTisue:
@velkan1984 your question isn't related to this pull request; please use another forum. see https://spark.apache.org/community.html

SethTisue:
@velkan1984 I see there you also posted this same question at https://issues.scala-lang.org/browse/SI-9539. I suggest continuing there rather than here

PR #5044, opened 2016-03-16
WIP SD-98 Avoid redundant mixin forwarders [ci: last-only]

retronym:
Builds on top of the trait encoding change PR.

```
% cat sandbox/test.scala
class A {
  def aaa = "A"
}
trait T extends A {
  private def bar = ""
  def foo = bar
  override def aaa = "T"
}
class C extends T

% qscalac sandbox/test.scala && javap -private -classpath . A T C
Compiled from "test.scala"
public class A {
  public java.lang.String aaa();
  public A();
}
Compiled from "test.scala"
public interface T {
  private java.lang.String bar();
  public java.lang.String foo();
  public java.lang.String aaa();
  public void $init$();
}
Compiled from "test.scala"
public class C extends A implements T {
  public java.lang.String aaa();
  public C();
}
```

TODO:
- [ ] Check for breakage in test suite or community build
- [ ] Add test case to show when forwarders are emitted.
- [ ] Discuss possible binary compatibility tradeoffs (e.g. if a separately compiled trait changes a method from deferred to concrete, how does absence of the forwarder change behaviour?)


lrytz:
here's a test that breaks with this patch:

```
trait T1 {
  def f = 1
}

trait T2 { self: T1 =>
  override def f = 2
}

class C extends T1 with T2

object Test extends App {
  println(new C().f)
}
```

```
$ qsc Test.scala && qs Test
java.lang.IncompatibleClassChangeError: Conflicting default methods: T1.f T2.f
    at C.f(Test.scala)
    at Test$.delayedEndpoint$Test$1(Test.scala:12)
    at Test$delayedInit$body.apply(Test.scala:11)
    at scala.Function0.apply$mcV$sp(Function0.scala:34)
```

lrytz:
re-submitted in https://github.com/scala/scala/pull/5085

PR #5048, opened 2016-03-17
flat classpath, wip

lrytz:


retronym:
Looking good!

I'm sort of surprised that `test/files/run/t6502.scala` (the test of REPL `:require`) works.

lrytz:
the test [hard-codes `ClassPathRepresentationType.Recursive`](https://github.com/scala/scala/blob/2.12.x/test/files/run/t6502.scala#L21), without that it fails.

PR #5050, opened 2016-03-18
Fix scaladoc scroll and auto-expand on permalinks

dhoepelman:
Small chances that fix [anchor links scroll](https://github.com/scala/scala/pull/5018#issuecomment-195020665) and auto-expand member description if an anchor link is used.

Suggested review by @VladUreche, @heathermiller

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.js
+++ b/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.js
@@ -16,18 +16,6 @@ $(document).ready(function() {
         $zoomOut: $('#diagram-zoom-out'),
     });
 
-    $("#template > div > div > ol > li > span.permalink").click(function(e) {
-        e.preventDefault();
-        var href = $("a", this).attr("href");
-        if (href.indexOf("#") != -1) {
-            location.hash = href.split("#").pop();
-            $("#template > div > div > ol > li").removeClass("selected");
-            var parent = $(this).parent().addClass("selected");
-            $("#content-container").animate({scrollTop: $("#content-container").scrollTop() + $(this).offset().top - $("#search").height() - 23}, 500);
-        }
-        return false;
-    });
-
     var oldWidth = $("div#subpackage-spacer").width() + 1 + "px";
     $("div#packages > ul > li.current").click(function() {
         $("div#subpackage-spacer").css({ "width": oldWidth });
@@ -65,21 +53,22 @@ $(document).ready(function() {
     function exposeMember(jqElem) {
         var jqElemParent = jqElem.parent(),
             parentName = jqElemParent.attr("name"),
-            linearizationName = /^([^#]*)(#.*)?$/gi.exec(parentName)[1];
+            ancestorName = /^([^#]*)(#.*)?$/gi.exec(parentName)[1];
 
         // switch visibility filter if necessary
         if (jqElemParent.attr("visbl") == "prt") {
             toggleVisibilityFilter(controls.visibility.all, controls.visibility.publicOnly);
         }
 
-        // toggle appropriate linearization buttons
-        if (linearizationName) {
-            $("#linearization li.out[name='" + linearizationName + "']").removeClass("out").addClass("in");
+        // toggle appropriate ancestor filter buttons
+        if (ancestorName) {
+            $("#filterby li.out[name='" + ancestorName + "']").removeClass("out").addClass("in");
         }
 
         filter();
         jqElemParent.addClass("selected");
-        $("#content-container").animate({scrollTop: jqElemParent.offset().top - $("#search").height() - 5 }, 1000);
+        commentToggleFct(jqElemParent);
+        $("#content-scroll-container").animate({scrollTop: $("#content-scroll-container").scrollTop() + jqElemParent.offset().top - $("#search").height() - 23 }, 1000);
     }
 
     var isHiddenClass = function (name) {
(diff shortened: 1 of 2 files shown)


VladUreche:
Thank you for the contribution @dhoepelman! The best reviewer for this would be @felixmulder, since he's been working on the layout side of scaladoc lately.

felixmulder:
Thanks for the contribution! I'll have a look tonight :)

felixmulder:
Also, please upload a build to somewhere I can have a look!

felixmulder:
For visible members - this works well - good job!

But there is still a problem when linking to hidden members like in this url:

```
.../scala/build/scaladoc/library/scala/Array.html#+(other:String):String
```

This member is shown by clicking `implicitly -> by any2stringadd`

If you'd like to fix that as well - that'd be great. Otherwise, this is good as is. Good job :+1:

src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.js
```diff
@@ -20,10 +20,16 @@ $(document).ready(function() {
...
-            location.hash = href.split("#").pop();
-            $("#template > div > div > ol > li").removeClass("selected");
-            var parent = $(this).parent().addClass("selected");
-            $("#content-container").animate({scrollTop: $("#content-container").scrollTop() + $(this).offset().top - $("#search").height() - 23}, 500);
+            var hash = href.split("#").pop()
+            try {
+                window.history.pushState({}, "", "#" + hash)
+            } catch (e) {
+                // fallback for file:// URLs
+                // pushState gives a better user experience because the browser doesn't reload
```
felixmulder: Kill this line or move it above the `window.history.pushState` line :)

felixmulder:
And before I forget - squish the commits once you consider this PR done :smiley:

dhoepelman:
Will take a look at it. Also noticed that clicking the permalink button doesn't work on members with `@group` (click handler doesn't get attached)

pimverkerk:
Thanks for the review, next time we will upload a build somewhere, any preferred place?

felixmulder:
> any preferred place?

@pimverkerk: no preferred place in particular - I usually upload it to the public folder on my dropbox on recommendation from others. Use whatever's easiest for you!

src/scaladoc/scala/tools/nsc/doc/html/resource/lib/template.js
```diff
@@ -298,13 +287,29 @@ $(document).ready(function() {
...
 
-    // highlight and jump to selected member
+    // highlight and jump to selected member if an anchor is provided
     if (window.location.hash) {
         var jqElem = findElementByHash(window.location.hash);
         if (jqElem.length > 0)
             exposeMember(jqElem);
     }
 
+    $("#template span.permalink").click(function(e) {
```
dhoepelman, line 297: Moved down to here so it will attach to the DOM elements created by `initInherit()`

dhoepelman:
[Build here](https://dl.dropboxusercontent.com/u/2911691/html/scaladoc/library/index.html)

@felixmulder
I’ve fixed [the hidden member problem](https://dl.dropboxusercontent.com/u/2911691/html/scaladoc/library/scala/Array.html#+%28other:String%29:String) (at least for implicits, inherited and private members already worked). I would say it can be merged if it LGTY.

On [pages with groups](https://dl.dropboxusercontent.com/u/2911691/html/scaladoc/reflect/scala/reflect/api/Mirrors.html#FieldMirrorextendsAnyRef) the scroll & expand on page enter is still broken.
The core of this problem is that `initInherit()` causes multiple anchors with the same ID to be put on the page.
For example `<a id="FieldMirrorextendsAnyRef"></a>` occurs twice on that page, if you remove the first one the scroll works as expected.

A hack would be to adjust the code to match the last id instead of the first, but it would probably be best to prevent duplicate id’s from existing by removing the statically generated elements or altering id's.
Didn’t feel confident enough to make those adjustments, might have effects I don’t foresee.

felixmulder:
Well - ideally we would group the members statically in `Entity.scala` - but since it's being done in `initInherit()` it should be done correctly. The `initInherit()` method seems to be **copying** the grouped members to their specific groups - as such duplicating the `<a id="...` as well as the list items they are contained within.

~~If they were **moved** instead, this might resolve the problem...do you mind fixing this as well? :sweat_smile:~~

**EDIT:** Changing the behaviour poses a different problem for us, since we're allowing the members to be sorted by `Grouped`, `Alphabetical` and by `Inheritance` - we need to allow for duplicates. So perhaps we should scrap anchor `id` and use something like `data-permalink`.

But this should be done in a different PR. So - LGTM :)

VladUreche:
Then, when switching between {grouped, normal, all-at-once} the procedure should be moving them around.

dhoepelman:
@felixmulder 
Changing `<a id=` into something like `<a data-permalink=` won't really solve the problem though, as there will still be two identically named permalink targets. Generating a new anchor/id might be a solution, but that needs some additional logic to hide/unhide the correct set when a permalink is used.
@VladUreche's solution sounds like a better one

felixmulder:
I agree, go for it! If it's too slow, we'll go back to the drawing board for statically generating the groups and so on.

dhoepelman:
I might take a look at it next week, if I do it I'll open a separate PR though

felixmulder:
That's fine by me - this PR LGTM. I think it's best if @VladUreche decides how you should rebase the PR in order for it to be merged (i.e. if it's ok as is, with two commits or if it has to be a single commit).

VladUreche:
Since the commits have a different authors, it's fine to keep them both.
Thanks for the contribution @dhoepelman and @pimverkerk!

VladUreche:
I'll merge as soon as the tests pass.

pimverkerk:
@VladUreche We merged the latest commits and all tests pass. So i guess you're good to go.

VladUreche:
Thanks for the contributions @pimverkerk and @dhoepelman!

PR #5051, opened 2016-03-19
Improve documentation of `GenSeqLike#length`

kmizu:
- [Range#length](https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/Range.scala#L140) throws Exception in the case when [Range#longLength](https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/Range.scala#L72) returns a Long greater than `Int.MaxValue`

### DIFF
--- a/src/library/scala/collection/GenSeqLike.scala
+++ b/src/library/scala/collection/GenSeqLike.scala
@@ -58,6 +58,7 @@ trait GenSeqLike[+A, +Repr] extends Any with GenIterableLike[A, Repr] with Equal
    *  Note: `xs.length` and `xs.size` yield the same result.
    *
    *  @return     the number of elements in this $coll.
+   *  @throws     IllegalArgumentException if the length of the sequence cannot be represented in an `Int`, for example, `(-1 to Int.MaxValue).length`.
    */
   def length: Int
 


src/library/scala/collection/GenSeqLike.scala
```diff
@@ -55,6 +55,10 @@ trait GenSeqLike[+A, +Repr] extends Any with GenIterableLike[A, Repr] with Equal
    *
    *  $willNotTerminateInf
    *
+   *  If the number of elements included in this is greater than 
+   *  `Int.MaxValue`, Exception will be thrown.
+   *  For example, `(-1 to Int.MaxValue).length` throw Exception.
+   *
```
retronym: Documentation of exceptions is usually presented in a `@throws` tag.

```
   *  @throws   IllegalArgumentException if the length of the length of the sequence cannot be represented in an `Int`, for example, `(-1 to Int.MaxValue).length`.
```

kmizu: OK.  I wil fix it.

lrytz:
would you mind squashing the two commits into one? you can use `git rebase -i 99a82be91cbb85239f70508f6695c6b21fd3558c`.

kmizu:
@lrytz Thanks. I squashed the two commits into one just now.

src/library/scala/collection/GenSeqLike.scala
```diff
@@ -58,6 +58,7 @@ trait GenSeqLike[+A, +Repr] extends Any with GenIterableLike[A, Repr] with Equal
    *  Note: `xs.length` and `xs.size` yield the same result.
    *
    *  @return     the number of elements in this $coll.
+   *  @throws   IllegalArgumentException if the length of of the sequence cannot be represented in an `Int`, for example, `(-1 to Int.MaxValue).length`.
```
SethTisue: there's a repeated word here: "of of"

kmizu: I fixed it just now.

adriaanm:
Thank you! For next time, please consider a shorter title for the commit (the PR's title was perfect, the commit title should be under 70 chars -- half a tweet :-)).

kmizu:
> please consider a shorter title for the commit

OK.

PR #5053, opened 2016-03-21
SI-9314 Marginal edge case to warn-missing-interp

som-snytt:
Don't warn on `"$package_name"` or `"${}"`.  Detecting empty expression also nullifies the test. As in "jury nullification."

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -105,7 +105,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
   // that are turned private by typedBlock
   private final val SYNTHETIC_PRIVATE = TRANS_FLAG
 
-  private final val InterpolatorCodeRegex  = """\$\{(.*?)\}""".r
+  private final val InterpolatorCodeRegex  = """\$\{\s*(.*?)\s*\}""".r
   private final val InterpolatorIdentRegex = """\$[$\w]+""".r // note that \w doesn't include $
 
   abstract class Typer(context0: Context) extends TyperDiagnostics with Adaptation with Tag with PatternTyper with TyperContextErrors {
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -105,7 +105,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
   // that are turned private by typedBlock
   private final val SYNTHETIC_PRIVATE = TRANS_FLAG
 
-  private final val InterpolatorCodeRegex  = """\$\{.*?\}""".r
+  private final val InterpolatorCodeRegex  = """\$\{(.*?)\}""".r
```
lrytz: Could we simplify `.*?` to `.*`

i won't add a ? at the end of the above question to avoid confusion..  :)

som-snytt:
```
scala> val r = """\$\{(.*?)\}""".r
r: scala.util.matching.Regex = \$\{(.*?)\}

scala> r findFirstIn "x ${a} y ${b}"
res0: Option[String] = Some(${a})

scala> val r = """\$\{(.*)\}""".r
r: scala.util.matching.Regex = \$\{(.*)\}

scala> r findFirstIn "x ${a} y ${b}"
res1: Option[String] = Some(${a} y ${b})
```

Ideally, we'd parse and then detect if it parses and any idents are in context.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -5206,16 +5206,18 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
           def suspiciousSym(name: TermName) = context.lookupSymbol(name, _ => true).symbol
-          def suspiciousExpr                = InterpolatorCodeRegex findFirstIn s
+          val suspiciousExpr                = InterpolatorCodeRegex findFirstMatchIn s
           def suspiciousIdents              = InterpolatorIdentRegex findAllIn s map (s => suspiciousSym(TermName(s drop 1)))
 
           if (suspiciousExpr.nonEmpty)
-            warn("detected an interpolated expression") // "${...}"
+            suspiciousExpr filter (!_.group(1).trim.isEmpty) foreach (_ =>
+              warn("detected an interpolated expression") // "${...}"
+            )
```
som-snytt: I thought you were going to apply the `suspiciousSym` test if the text is an identifier. Then `${junk}` gets a pass like `$pass`. That might solve most conflicts with templating.

som-snytt:
Added not warning on `"${ RANDOM }"` where the simple ident is not in scope. Maybe this covers the templating use case. It takes the edge off the edge case.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
lrytz: got it thanks.

lrytz:
LGTM, thank you!

PR #5056, opened 2016-03-22
SI-9488 - adds the same default toString format to Scala Futures as 2…

viktorklang:
….12.x

Includes tests to verify the toString representations.

### DIFF
--- a/bincompat-backward.whitelist.conf
+++ b/bincompat-backward.whitelist.conf
@@ -217,6 +217,12 @@ filter {
     {
         matchName="scala.concurrent.impl.Promise$DefaultPromise"
         problemName=MissingTypesProblem
+    },
+    // SI-9488: Due to SI-8362 above, toString was silently changed to the AtomicReference toString implementation,
+    // This is fixed by SI-9488, and this should be safe since the class in question is stdlib internal.
+    {
+        matchName="scala.concurrent.impl.Promise.toString"
+        problemName=MissingMethodProblem
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


viktorklang:
@retronym Could you have a look? (https://issues.scala-lang.org/browse/SI-9488)

retronym:
LGTM

Thanks!

lrytz:
IDE failure due to https://github.com/scala/scala-dev/issues/104

viktorklang:
Thanks Lukas!

lrytz:
btw, since i went trough the exercise of understanding the binary compatibility impact, here's why i think it's safe. adding a concrete method to a trait requires re-compiling all sublcasses, here's an what happens otherwise:

```
➜ cat A.scala
trait A {
  override def toString = "hjai"
}
➜ cat B.scala
class C extends A
object Test extends App {
  println((new C).toString)
}
➜ scalac A.scala B.scala
➜ scala Test
hjai
➜ rm A*.class
➜ cat A1.scala
trait A {
  //override def toString = "hjai"
}
➜ scalac A1.scala
➜ scala Test
java.lang.ClassNotFoundException: A$class
```

`trait Promise[T]` is `private[concurrent]` and its subtypes are
- either also private (and being recompiled as part of the same codebase)
-  `class DefaultPromise` is a class, so the mix-in is generated

so it looks safe.

viktorklang:
@lrytz yes, exactly. That's why I put the code in that trait and not the Future-trait itself :)

PR #5058, opened 2016-03-23
Inline super calls, as they are statically resolved

lrytz:
Ensures that mixin methods of `@inline` annotated concrete trait methods
inline the trait method.

Fixes https://github.com/scala/scala-dev/issues/86

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
@@ -132,7 +132,7 @@ class CallGraph[BT <: BTypes](val btypes: BT) {
               (declarationClassNode, source) <- byteCodeRepository.classNodeAndSource(declarationClass): Either[OptimizerWarning, (ClassNode, Source)]
             } yield {
                 val declarationClassBType = classBTypeFromClassNode(declarationClassNode)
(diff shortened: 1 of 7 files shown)


lrytz:
follow-up for new trait encoding, review by @retronym

src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
```diff
@@ -295,6 +295,7 @@ class CallGraph[BT <: BTypes](val btypes: BT) {
           // TODO: type analysis can render more calls statically resolved. Example:
           //   new A.f  // can be inlined, the receiver type is known to be exactly A.
           val isStaticallyResolved: Boolean = {
+            isNonVirtualCall(call) || // SD-86: super calls (invokespecial) can be inlined
```
retronym, line 298: Noting that although this will include constructors, they are explicitly excluded below with `!BytecodeUtils.isConstructor(calleeMethodNode)`

retronym:
My reading of this is that we will now inline `INVOKESTATIC`.

Typically we wouldn't expect static methods to be `@inline` annotated, but assuming one was, or that heuristics tried to inline it anyway, would this mean we omit side effect in `<clinit>` of the enclosing class? We've had a similar discussion about omitting side effects from module loads.

lrytz:
There's actually no change in this PR with respect to static methods: already before, `isStaticallyResolved` was true because `methodInlineInfo.effectivelyFinal` is true.

But you're right of course about skipping the class initializer; the inliner does delay them currently. I noted this at https://github.com/scala/scala-dev/issues/112

retronym:
LGTM

PR #5059, opened 2016-03-23
SI-9702 Fix backend crash with classOf[T] annotation argument

lrytz:
A `classOf[List[Int]]` annotation argument is translated to a constant
with the `AliasArgsTypeRef` as constant value.

To obtain the corresponding ASM Type to emit as annotation argument, we
now go directly to the class symbol using `typeSymbol`, instead of
calling the generic `typeToBType` method, which expects only
non-parameterized types.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -595,17 +595,6 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
       val classSym = if (sym.isJavaDefined && sym.isModuleClass) exitingPickler(sym.linkedClassOfClass) else sym
       classBTypeFromSymbol(classSym).internalName
     }
-
-    /**
-     * The jvm descriptor of a type.
-     */
-    final def descriptor(t: Type): String = typeToBType(t).descriptor
-
-    /**
-     * The jvm descriptor for a symbol.
-     */
-    final def descriptor(sym: Symbol): String = classBTypeFromSymbol(sym).descriptor
-
   } // end of trait BCInnerClassGen
 
   trait BCAnnotGen extends BCInnerClassGen {
@@ -613,6 +602,35 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
     private lazy val AnnotationRetentionPolicySourceValue  = AnnotationRetentionPolicyModule.tpe.member(TermName("SOURCE"))
     private lazy val AnnotationRetentionPolicyClassValue   = AnnotationRetentionPolicyModule.tpe.member(TermName("CLASS"))
     private lazy val AnnotationRetentionPolicyRuntimeValue = AnnotationRetentionPolicyModule.tpe.member(TermName("RUNTIME"))
+
+    /**
+     * Annotations are not processed by the compilation pipeline like ordinary trees. Instead, the
+     * typer extracts them into [[AnnotationInfo]] objects which are attached to the corresponding
+     * symbol (sym.annotations) or type (as an AnnotatedType, eliminated by erasure).
+     *
+     * For Scala annotations this is OK: they are stored in the pickle and ignored by the backend.
+     * Java annoations on the other hand are additionally emitted to the classfile in Java's format.
+     *
+     * This means that [[Type]] instances within an AnnotaionInfo reach the backend non-erased. Examples:
+     *   - @(javax.annotation.Resource @annotation.meta.getter) val x = 0
+     *     Here, annotationInfo.atp is an AnnotatedType.
+     *   - @SomeAnnotation[T] val x = 0
+     *     In principle, the annotationInfo.atp is a non-erased type ref. However, this cannot
+     *     actually happen because Java annotations cannot be generic.
+     *   - @javax.annotation.Resource(`type` = classOf[List[_]]) val x = 0
+     *     The annotationInfo.assocs contains a LiteralAnnotArg(Constant(tp)) where tp is the
+     *     non-erased existential type.
+     */
+    def erasedType(tp: Type): Type = enteringErasure {
+      // make sure we don't erase value class references to the type that the value class boxes
+      // this is basically the same logic as in erasure's preTransform, case Literal(classTag).
+      tp.dealiasWiden match {
+        case tr @ TypeRef(_, clazz, _) if clazz.isDerivedValueClass => erasure.scalaErasure.eraseNormalClassRef(tr)
+        case tpe => erasure.erasure(tpe.typeSymbol)(tpe)
+      }
+    }
+
+    def descriptorForErasedType(tp: Type): String = typeToBType(erasedType(tp)).descriptor
 
     /** Whether an annotation should be emitted as a Java annotation
       * .initialize: if 'annot' is read from pickle, atp might be uninitialized
@@ -650,7 +668,6 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
         ca(idx) = b.asInstanceOf[Char]
         idx += 1
       }
-
       ca
     }
 
@@ -713,9 +730,10 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
               case StringTag  =>
                 assert(const.value != null, const) // TODO this invariant isn't documented in `case class Constant`
                 av.visit(name, const.stringValue)  // `stringValue` special-cases null, but that execution path isn't exercised for a const with StringTag
-              case ClazzTag   => av.visit(name, typeToBType(const.typeValue).toASMType)
+              case ClazzTag   =>
+                av.visit(name, typeToBType(erasedType(const.typeValue)).toASMType)
               case EnumTag =>
-                val edesc  = descriptor(const.tpe) // the class descriptor of the enumeration class.
+                val edesc  = descriptorForErasedType(const.tpe) // the class descriptor of the enumeration class.
                 val evalue = const.symbolValue.name.toString // value the actual enumeration value.
                 av.visitEnum(name, edesc, evalue)
             }
@@ -740,7 +758,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
         case NestedAnnotArg(annInfo) =>
           val AnnotationInfo(typ, args, assocs) = annInfo
           assert(args.isEmpty, args)
-          val desc = descriptor(typ) // the class descriptor of the nested annotation class
+          val desc = descriptorForErasedType(typ) // the class descriptor of the nested annotation class
           val nestedVisitor = av.visitAnnotation(name, desc)
           emitAssocs(nestedVisitor, assocs)
       }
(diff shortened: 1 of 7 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
```diff
@@ -713,7 +713,8 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
               case StringTag  =>
                 assert(const.value != null, const) // TODO this invariant isn't documented in `case class Constant`
                 av.visit(name, const.stringValue)  // `stringValue` special-cases null, but that execution path isn't exercised for a const with StringTag
-              case ClazzTag   => av.visit(name, typeToBType(const.typeValue).toASMType)
```
retronym, line 716: Do we need the same treatment in `genConstant`, which contains the following?

``` scala
        case ClazzTag   =>
          val toPush: BType = {
            typeToBType(const.typeValue) match {
              case kind: PrimitiveBType => boxedClassOfPrimitive(kind)
              case kind => kind
            }
          }
          mnode.visitLdcInsn(toPush.toASMType)

```

retronym:
Looks like we've also got a kindred bug in [erasure](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/transform/Erasure.scala#L1113-L1119):

```
scala> class VC(val a: Any) extends AnyVal; type VCAlias = VC
defined class VC
defined type alias VCAlias

scala> classOf[VCAlias]
res1: Class[VCAlias] = class java.lang.Object

scala> classOf[VC]
res2: Class[VC] = class VC
```

src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
lrytz, line 716: It turns out that code works fine. The reason is that erasure visits the `Literal(Constant(..))` AST and replaces the type by a `ClassNoArgsTypeRef` to `List`. In the case of an annotation arg, the `Constant(..)` is part of a symbol's `annotations` and not touched by erasure.

lrytz:
the value class type alias issue is easy to fix, will push another commit.

lrytz:
review by @retronym. it includes a sketch for having run tests within junit, for now using toolboxes. maybe we should not rely on toolboxes, as java symbols are loaded through java reflection instead of the classfile parser - i fixed one bug in there.

lrytz:
it's a bit of a rabbit hole, here's another one that crashes with current 2.12.x (only when `final`!)

```
scala> class T { final val k = classOf[List[_]] }
java.lang.AssertionError: assertion failed: type List
    at scala.tools.nsc.backend.jvm.BTypesFromSymbols.nonClassTypeRefToBType$1(BTypesFromSymbols.scala:166)
```

probably won't finish that stuff today.

lrytz:
as so often, once you start looking more details, there's plenty of little surprises :-)

here's what happens for `final val k = classOf[List[_]]`

pre-req. erasure has two tree transformers: `preTransform` basically sets all tree's `tpe` to `null` (and does a few other adjustments), the second phase runs `exitingErasure(newTyper().typed(tree)`.
- type-checking a `classOf[..]`-`TypeApply` goes to `typedClassOf` and creates a `Literal(Constant(tp)) setType ConstantType(Constant(tp))`, where `tp` is the existential `List[_]`
- since it's a final val, the namer assigns the `ConstantType` to the `ValDef`'s `tpt` (without widening), and equally to the generated getter
- for the getter, UnCurry replaces the `DefDef`'s `rhs` by a `Literal` (because the return type is a `ConstantType`)
- erasure's `preTransform` has a special case for `Literal` trees where the constant holds a type (i.e., `classOf`-literals): the `Type` in the constant is replaced by its erased version. this applies for both the `ValDef`'s and the getter `DefDef`'s `rhs`.
  - interesting: `treeCopy.Literal(tree, Constant(erased))` will copy the original `ConstantType(Constant(tp))` (where tp is still the existential) onto the newly created literal
  - but at the end, `preTransform` invokes `tree.clearType()`, so the type is set to `null`

now we're in the re-typing phase of erasure
- the `Literal` is assigned type `t1 = ConstantType(Constant(List))`
- `typedValDef` computes the expected type to type-check the `rhs`. this expected type is basically `typedType(vdef.tpt)`.
  - the `vdef.tpt` is a `t2 = TypeTree(ConstantType(Constant(List[_])))`: the existential is still there. The reason is that the erasure type-maps return `ConstantType`s unchanged. this should probably be fixed
  - at the end of type-checking the `rhs`, the typer invokes `adapt`
  - because `t1` does not conform to `t2`, erasure's `adapt` method introduces a cast to `t1`.
  - the new tree `Literal(t2).asInstOf[t2]` is type-checked and assigned type `t2`. at the end the `adapt` method of the typer sees a tree with `ConstantType(Constant(t2))`, so `adaptConstant` replaces the tree by a new `Literal(Constant(t2))`. So we get the existential back!
- the same happens for the getter `DefDef`.

that's how we end up with a non-erased (existential) type in the backend.

if the `val` is not final, the namer assigns `ConstantType(..).widen` to the `ValDef.tpt`, which is a `TypeRef` for `Class[List[_]]`. This one is correctly erased by the erasure type map, so the expected when re-typing `rhs` during erasure is simply `Class`.

```
scala> :power

scala> val nonErasedConstantTp = ConstantType(Constant(typeOf[List[_]]))
scala> val erasedConstantTp = ConstantType(Constant(erasure.scalaErasure(typeOf[List[_]])))
scala> val valDefTptForNonFinal = nonErasedConstantTp.widen

scala> exitingErasure { erasedConstantTp <:< nonErasedConstantTp }
res15: Boolean = false // therefore, erasure inserts the cast

scala> exitingErasure { erasedConstantTp <:< erasure.scalaErasure(valDefTptForNonFinal) }
res16: Boolean = true // no cast inserted for non-final val
```

lrytz:
I think I went into all the glory details of this topic now, so ping @retronym.

retronym:
What a rabbit hole! This looks thoroughly thought through and tested. I'll review this in more detail tomorrow.

retronym:
LGTM

@adriaanm Might be a contender for M4

adriaanm:
Agreed -- merged! Now let's ship this puppy :shipit:

src/reflect/scala/reflect/internal/transform/Erasure.scala
```diff
@@ -112,8 +112,9 @@ trait Erasure {
     protected def eraseDerivedValueClassRef(tref: TypeRef): Type = erasedValueClassArg(tref)
 
     def apply(tp: Type): Type = tp match {
-      case ConstantType(_) =>
-        tp
+      case ConstantType(ct) =>
+        if (ct.tag == ClazzTag) ConstantType(Constant(apply(ct.typeValue)))
+        else tp
```
sjrd, line 117: I think this broke Scala.js. In _some_ cases, a `classOf[Unit]` becomes `classOf[BoxedUnit]` after erasure in 2.12.0-M4, but it did not up and until March 31, before this PR was merged (it stayed a `classOf[Unit]`). The "some" cases include when compiling `java.lang.Void` and `scala.runtime.BoxedUnit` in Scala.js; but does not include when compiling the hello world or the test suite. I did not try to minimize this, so far :-s

This causes the following 3 Scala.js tests to fail:
- org.scalajs.testsuite.javalib.lang.ClassTest
- org.scalajs.testsuite.scalalib.ArrayBuilderTest
- org.scalajs.testsuite.scalalib.ClassTagTest

See the full log here: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/364/consoleFull (grep for `[scala-js] [error] Failed tests:`).

At least for `ClassTest`, this happens because `java.lang.Void.TYPE` and `scala.runtime.BoxedUnit.TYPE` return the `class scala.runtime.BoxedUnit` instead of `void`. And that, because, when those classes were compiled, `classOf[Unit]` was transformed into `classOf[BoxedUnit]` during erasure. The source for these files are at https://github.com/scala-js/scala-js/blob/master/javalanglib/src/main/scala/java/lang/Void.scala and https://github.com/scala-js/scala-js/blob/master/library-aux/src/main/scala/scala/runtime/BoxedUnit.scala

Any idea?

sjrd, line 117: OK I can reproduce it. It happens precisely for `final val`s whose right-hand-side is a `classOf[Unit]`. Compiling

``` scala
object Foo {
  final val C = classOf[Unit]
}
```

gives, after erasure:

``` scala
  object Foo extends Object {
    def <init>(): helloworld.Foo.type = {
      Foo.super.<init>();
      ()
    };
    final private[this] val C: Class(classOf[scala.runtime.BoxedUnit]) = classOf[scala.runtime.BoxedUnit];
    final <stable> <accessor> def C(): Class(classOf[scala.runtime.BoxedUnit]) = classOf[scala.runtime.BoxedUnit]
  };
```

lrytz, line 117: I'm trying to reproduce:

```
$ cat Test.scala
object Test {
  final val a              = classOf[Unit]
  final val b: Class[Unit] = classOf[Unit]
  val c                    = classOf[Unit]
  val d      : Class[Unit] = classOf[Unit]

    val bu = classOf[runtime.BoxedUnit]
    println(a)
    println(bu)
    println(a eq b)
    println(a eq c)
    println(a eq d)
    println(a eq bu)
  }
}

$ scala -version
Scala code runner version 2.12.0-M4 -- Copyright 2002-2016, LAMP/EPFL
$ scalac Test.scala
$ scala Test
void
class scala.runtime.BoxedUnit
true
true
true
false
```

lrytz, line 117: i do see the difference in the AST, i wonder if that's expected and what happens in the backend so that the two fields hold the same value:

```
$ cat Test.scala
object Test {
  final val a              = classOf[Unit]
  final val b: Class[Unit] = classOf[Unit]
}
$ scalac Test.scala -Xprint:erasure
[[syntax trees at end of                   erasure]] // Test.scala
package <empty> {
  object Test extends Object {
    def <init>(): Test.type = {
      Test.super.<init>();
      ()
    };
    final private[this] val a: Class(classOf[scala.runtime.BoxedUnit]) = classOf[scala.runtime.BoxedUnit];
    final <stable> <accessor> def a(): Class(classOf[scala.runtime.BoxedUnit]) = classOf[scala.runtime.BoxedUnit];
    final private[this] val b: Class = classOf[scala.Unit];
    final <stable> <accessor> def b(): Class = Test.this.b
  }
}
```

lrytz, line 117: it seems to be specific to `Unit`, doesn't happen for `Int`.

sjrd, line 117: Yes, in Scala/JVM you can only see it in the AST. Because of constant-folding for `final val`s coming from Scala-defined classes, the `final val` is not actually accessed at run-time. You can fully reproduce the bug by implementing an inherited `def` with the `final val`:

``` scala
Welcome to Scala 2.12.0-M4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_60).

scala> trait Foo { def C: Class[_] }
defined trait Foo

scala> object Bar extends Foo { final val C = classOf[Unit] }
defined object Bar

scala> Bar.C // constant-folded, works
res0: Class[Unit] = void

scala> val foo: Foo = Bar
foo: Foo = Bar$@16cb0050

scala> foo.C // not constant-folded, does not work
res1: Class[_] = class scala.runtime.BoxedUnit
```

lrytz, line 117: https://github.com/scala/scala/pull/5094

PR #5060, opened 2016-03-23
Fix a bug in multi-dimensional array creation

DarkDimius:
`']' * 5` is not `]]]]]` but `245`.

I guess this code is never executed, because it would never work.


lrytz:
We track the fact that we're not emitting `multinewarray` at https://github.com/scala/scala-dev/issues/106

lrytz:
LGTM

PR #5061, opened 2016-03-23
JMH-based benchmark framework for contributors

performantdata:
Turns `test/benchmarks` into an SBT project that uses SBT-JMH.

This is in follow-up to [my plans mentioned on scala-internals](https://groups.google.com/forum/#!topic/scala-internals/OloyOAf3NH0).


SethTisue:
presumably the integrate-ide failure is https://github.com/scala/scala-dev/issues/104 rather than anything specific to this PR

retronym:
Great stuff. I had toyed with this idea in https://github.com/retronym/scala-jmh-suite, which includes some plumbing to enable collection of Java Flight Recorder profiles of benchmarks.

performantdata:
Thanks. Thanks for the URL. I'll need to track memory usage soon. I also want to build my Gnuplot graphs automatically. More to come after this PR.

performantdata:
@retronym, I'm going with JOL, since it gives me easy access to memory usage from the teardown code. Can't figure how I'd do that with Flight Recorder.

retronym:
The more tools in the box, the better. I'm mostly after a convienient way to turn on Flight recorder when analysing benchmarks manually.

adriaanm:
/nothingtoseehere -- The failure is due to scala/scala-dev#104, not this PR.

performantdata:
review by @Ichoran or @axel22 ?

performantdata:
@retronym would you care to review?

test/benchmarks/README.md
```diff
@@ -0,0 +1,97 @@
+# Scala library benchmarks
+
+This directory is a standalone SBT project, within the Scala project,
+that makes use of the [SBT plugin](https://github.com/ktoso/sbt-jmh) for [JMH](http://openjdk.java.net/projects/code-tools/jmh/).
+
+## running a benchmark
+
+The benchmarks require first building Scala into `../../build/pack`, using Ant.
```
Ichoran: Does it matter whether it's Ant or SBT?

test/benchmarks/README.md
```diff
@@ -0,0 +1,97 @@
...
+    jmh:runMain scala.collection.mutable.OpenHashMapRunner
+
+in SBT.
+SBT should be run _from this directory_.
+
+The JMH results can be found under `target/jmh-results/`.
+`target` gets deleted on an SBT `clean`,
+so you should copy these files out of `target` if you wish to preserve them.
+
+## creating a benchmark and runner
```
Ichoran: Capitalize the title ("Creating a benchmark...")

test/benchmarks/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+The benchmarking classes use the same package hierarchy as the classes that they test
+in order to make it easy to expose, in package scope, members of the class under test,
+should that be necessary for benchmarking.
+
+There are two types of classes in the source directory:
+those suffixed "`Benchmark`" and those suffixed "`Runner`".
```
Ichoran: Quotes and backticks are redundant.  Just use backticks.

test/benchmarks/README.md
```diff
@@ -0,0 +1,97 @@
+# Scala library benchmarks
+
+This directory is a standalone SBT project, within the Scala project,
+that makes use of the [SBT plugin](https://github.com/ktoso/sbt-jmh) for [JMH](http://openjdk.java.net/projects/code-tools/jmh/).
+
+## running a benchmark
```
Ichoran: Capitalize first word of title (here and all following titles)

test/benchmarks/README.md
```diff
@@ -0,0 +1,97 @@
...
+These require the `-XX:+UnlockDiagnosticVMOptions` JVM option.
+
+#### viewing inlining events
+Add `-XX:+PrintInlining`.
+
+#### viewing the disassembled code
+To show the assembly code corresponding to the code generated by the JIT compiler for specific methods,
+add `-XX:CompileCommand=print,scala.collection.mutable.OpenHashMap::*`,
+for example, to show all of the methods in the `scala.collection.mutable.OpenHashMap` class.
+If you're running OpenJDK, you may need to install the disassembler library (`hsdis-amd64.so` for the `amd64` architecture).
```
Ichoran: This is necessary for the Oracle JVM also (I just verified).

test/benchmarks/README.md
```diff
@@ -0,0 +1,97 @@
...
+In Debian, this is available in <a href="https://packages.debian.org/search?keywords=libhsdis0-fcml">the `libhsdis0-fcml` package</a>.
```
Ichoran: This will work for the OpenJDK, but you have to move the file to the right path to get it to work for the Oracle JVM.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+    //    val jvm = params.getJvm
+    //    val jvmArgs = params.getJvmArgs.mkString(" ")
+
+    val f = new PrintWriter(outputFile, "UTF-8")
+    try {
+      datasetByName.foreach(_ match { case (label: String, dataset: Iterable[RunResult]) ⇒ {
+        f.format("# [%s]\n", label)
+
+        val isMemoryUsageDataset = label.contains(memoryDatasetQualifer)
+        dataset.foreach { result ⇒
```
Ichoran: Use two characters for `=>`.  The occasional problems with ⇒ are not worth it.  (People not set up to type it, random unicode mangling, etc..)

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+          val size = result.getParams.getParam(sizeParamName)
+          val secondaryResults = result.getSecondaryResults
+          if (isMemoryUsageDataset) {
+            val memoryResult = secondaryResults.get(memoryAuxCounterName)
+            val entriesResult = secondaryResults.get(entriesAuxCounterName)
+            f.format("%s %f %f %f %f\n", size,
```
Ichoran: This would be a lot easier to follow if you used plain `f.println` and the f- or s-interpolator to generate the results.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+              Double.box(entriesResult.getScore), Double.box(entriesResult.getStatistics.getStandardDeviation),
```
Ichoran: Can you factor this pattern out?  At least `getStatistics.getStandardDeviation` could be a call to a local `std` method, but you might just make a `statStr` method or something that  does the whole `f"${x.getScore} ${x.getStatistics.getStandardDeviation}"` thing.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /** Put elements into the given map. */
+  private def put_Int(map: OpenHashMap[Int,Int], from: Int, to: Int) {
+    var i = from
+    while (i <= to) {  // using a `for` expression instead adds significant overhead
+      map.put(i, i)
+      i += 1
+    }
+  }
```
Ichoran, line 127: You end here without doing anything with your constructed map.  This is generally safe enough, but it's probably good practice to include a dump of the result to Blackhole (e.g. the size) to be sure that the work isn't optimized away.  (Especially if someone copies and pastes this code for another benchmark where the putting is changed into a more elidable operation.)

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
```diff
@@ -0,0 +1,204 @@
...
+      // remove every other entry
+      var i = base + 1
+      while (i <= base + blockSize) {
+        map.remove(i)
+        i += 2
+      }
+
+      base += blockSize
+    }
+  }
```
Ichoran, line 150: Blackholing would probably be good again.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  /** Get elements from the given map. */
+  def get_Int(map: OpenHashMap[Int,Int], size: Int, bh: Blackhole) {
+    var i = 1
+    while (i <= size) {
+      bh.consume(map.get(i).getOrElse(0))
```
Ichoran, line 156: I think you'll have less overhead if you accumulate a running sum and push that into `bh` instead.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
```diff
@@ -0,0 +1,204 @@
...
+@Fork(6)
+@Threads(1)
+@Warmup(iterations = 20)
+@Measurement(iterations = 6)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class OpenHashMapBenchmark {
+  import OpenHashMapBenchmark._
+
+  @Param(Array("50", "100", "250", "1000", "2500", "10000", "25000", "100000", "250000", "1000000", "2500000",
```
Ichoran: Unless you declare that a map is only to be used for big collections, it's also valuable to test the smaller sizes.  This generally involves making an array of maps and doing something to each of them, as there is too much overhead to measure accurately the timings in a map of e.g. 2 elements.  You can just run all your benchmarks that way, since it's perfectly okay to have an array of length 1.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+/** Replacement JMH application that runs the [[OpenHashMap]] benchmark.
+  * 
+  * Outputs the results in a form consumable by a Gnuplot script.
+  */
+object OpenHashMapRunner extends JmhRunner {
+  /** File that will be created for the output data set. */
+  private[this] val outputFile = new File(outputDirectory, "OpenHashMap.dat")
+
+  /** Qualifier to add to the name of a memory usage data set. */
+  private[this] val memoryDatasetQualifer = " memory"
```
Ichoran: Do you mean to have a space in front here?  Is it _necessary_?  If it's not necessary but you need a separator, I'd go for `-` or `_` instead, simply because it otherwise looks like an error.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  /** Name of the JMH parameter for the number of map entries per invocation. */
+  private[this] val sizeParamName = "size"
```
Ichoran: Is this actually accomplishing anything except bulking up the code?  Abstracting out constants is fine if you might get them wrong or they might change, but this is unlikely to change and hard to get wrong.  Also, you're not really abstracting the common operation here (which is `getParam.getParam("size")`).  Same deal for all the names: they're used to accomplish something, so abstract the action rather than just make the name longer.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+  /** Name of the JMH auxiliary counter that collects the number of map entries. */
+  private[this] val entriesAuxCounterName = "mapEntries"
+
+  def main(args: Array[String]) {
+    import scala.collection.JavaConversions._
+    import scala.language.existentials
+  
+    val opts = new CommandLineOptions(args: _*)
+    var builder = new OptionsBuilder().parent(opts)
+      .jvmArgsPrepend("-Xmx6000m")
```
Ichoran: There is no reason to put this on a separate line

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    /** Map from data set name to data set. */
+    val datasetByName = Map.empty[String, Set[RunResult]]
+
+    /** Ordering for the results within a data set. Orders by increasing number of map entries. */
+    val ordering = Ordering.by[RunResult, Int](_.getParams.getParam(sizeParamName).toInt)
+
+    def addToDataset(result: RunResult, key: String): Unit =
+      datasetByName.get(key)
+        .getOrElse({ val d = SortedSet.empty(ordering); datasetByName.put(key, d); d }) += result
```
Ichoran: I generally prefer dot-on-previous-line because it works better when e.g. pasting into the REPL.  The indentation is already plenty of indication that it's a continuation of the previous command.  Also, the `+= result` bit is so far to the right it's easy to miss.  I'd just assign a temporary variable.  And are you not using `getOrElseUpdate` for some reason?  That might make it short enough so a single line is fine.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
Ichoran: Also, "Qualifier" is misspelled.

test/benchmarks/README.md
performantdata: What does "move the file to the right path" mean precisely? What should I tell them?

Ichoran: "If you are using the Oracle JVM or another compatible JVM that is not included with your distribution, you may need to copy or link the `hdis-amd64.so` file from its installation location for an OpenJDK install to the `jre/lib/amd64` directory inside the Java installation that is being used."

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
performantdata: I was previously overriding other options as well.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
performantdata, line 156: Maybe so, but I plan to test other type parameters than numerics, and would like the tests to be comparable.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
performantdata: That's how the tests are constructed already. See the `BulkPutState` class. I'd only need to add more `size` values. But the values are _sort of_ limited to multiples of 25 by the current `blocks` value in `put_remove_Int()`.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
performantdata: > Do you mean to have a space in front here?

Yes. I label each dataset with its corresponding method name, which can't (wouldn't normally) contain a space. So a `-` would also work.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
performantdata, line 127: The better place to blackhole results is in the `@Benchmark` methods; these are here as utility methods. But I believe that the maps being passed into those methods as external state prevents them from being optimized away.

The difficulty of eliding `put` calls wouldn't prevent the whole map from being optimized away. Conversely, if the `put` were changed to `get`, blackholing a `map.size()` would make the `get` no less elidable. I guess I don't have much sympathy either for someone who's copy-pasting the wrong example, e.g., `put_Int()` instead of `get_Int()`.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
retronym, line 156: Or perhaps, put the `Option[Int]` into the blackhole, rather than calling `Option#getOrElse`. Or is the intention to test the combined performance of `OHM#get` / `Option#getOrElse`?

It is also worth thinking about what assumptions these benchmarks make about the scalac compiler options used to compile the tests themselves. Under `-Yopt:l:classpath`, `Option#getOrElse` is inlined 

```
% ~/scala/2.12/bin/scala -Yopt:l:classpath
Welcome to Scala 2.12.0-M4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_71).

scala> class OptInline { def foo(a: Option[Int]) = a.getOrElse(0) }
defined class OptInline

scala> :javap -c OptInline#foo
  public int foo(scala.Option<java.lang.Object>);
    Code:
       0: aload_1
       1: ifnonnull     6
       4: aconst_null
       5: athrow
       6: aload_1
       7: invokevirtual #27                 // Method scala/Option.isEmpty:()Z
      10: ifeq          23
      13: invokestatic  #31                 // Method $line4$$read$$iw$$iw$OptInline$$$anonfun$1:()I
      16: invokestatic  #37                 // Method scala/runtime/BoxesRunTime.boxToInteger:(I)Ljava/lang/Integer;
      19: astore_2
      20: goto          28
      23: aload_1
      24: invokevirtual #41                 // Method scala/Option.get:()Ljava/lang/Object;
      27: astore_2
      28: aload_2
      29: invokestatic  #45                 // Method scala/runtime/BoxesRunTime.unboxToInt:(Ljava/lang/Object;)I
      32: ireturn
```

```
 ~/scala/2.12/bin/scala
Welcome to Scala 2.12.0-M4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_71).

scala> class OptInline { def foo(a: Option[Int]) = a.getOrElse(0) }
defined class OptInline

scala> :javap -c OptInline#foo
  public int foo(scala.Option<java.lang.Object>);
    Code:
       0: aload_1
       1: invokedynamic #43,  0             // InvokeDynamic #0:apply$mcI$sp:()Lscala/runtime/java8/JFunction0$mcI$sp;
       6: invokevirtual #49                 // Method scala/Option.getOrElse:(Lscala/Function0;)Ljava/lang/Object;
       9: invokestatic  #55                 // Method scala/runtime/BoxesRunTime.unboxToInt:(Ljava/lang/Object;)I
      12: ireturn
```

As a starting point, I would suggest adding `-Yopt:l:classpath` to the `scalacOptions` for this project. Perhaps we'd like to parameterize this down the track, or maybe have separate sub projects with different optimization levels so we could have fine grained control over the way that benchmark code is compiled.

performantdata, line 156: > is the intention to test the combined performance of OHM#get / Option#getOrElse?

Yes, because I think that's a more common usage, to subsequently unpack the `Option`.

performantdata:
I just rebased to pick up #5139.

performantdata:
@SethTisue, any idea what's wrong with the 2.11 build now?

adriaanm:
It looks like our CI was really overloaded because each commit was being validated at the same time and you pushed about 10 of them :-)

adriaanm:
We have limits on how many jobs Jenkins runs at the same time. Looks like they could do with some more tuning.

adriaanm:
Hmm, at least that was my first impression. Looks like several commits took >40min just to build the compiler. Not sure what's going on, but I haven't followed the history of this PR.

adriaanm:
I've added the suffix to tell the CI to only build the last commit, to help in load balancing. Let's do a rebuild when the other commits are done.

performantdata:
@adriaanm, it's still red. Do we ignore those commits, or how do I rebuild just them?

adriaanm:
I removed the suffix that tells the CI to only build the last commit and then `/rebuild`. Hopefully all-green in a couple hours. ⏳

performantdata:
Oh, it's right [there](https://github.com/scala/scabot/blob/master/README.md#commands).

performantdata:
Green. Back to you, @Ichoran.

performantdata:
@Ichoran, are you still reviewing this, Rex?

Ichoran:
@performantdata - I'm snowed under with commitments but I'll try to get to it sometime this weekend (probably Sunday afternoon).

performantdata:
@Ichoran OK thanks. It was just that I'm having to cherry-pick these commits whenever I want to test a branch, so I can just keep doing that until you're free, not a big deal.

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,105 @@
+package scala.collection.mutable
+
+import java.io.BufferedWriter
+import java.io.File
+import java.io.FileOutputStream
+import java.io.OutputStreamWriter
+import java.io.PrintWriter
+import scala.collection.JavaConversions
+import scala.language.existentials
```
Ichoran, line 9: This import is kind of lost here, which is contrary to the point of requiring the import in the first place.  Can you move it to the top?

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    results.foreach { result =>
+      addToDataset(result.label, result)
+
+      // Create another data set for trials that track memory usage
+      if (result.memory.isDefined)
+        addToDataset(result.label + memoryDatasetQualifier, result)
+    }
+
+    //TODO Write out test parameters
```
Ichoran, line 80: If we want to do this, can we do it?  If not, can we decide not and remove or change the comment?  (E.g. "If we wanted to write out the test parameters, we'd do here / like this: ...")

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    val f = new PrintWriter(outputFile, "UTF-8")
+    try {
+      datasetByName.foreach(_ match { case (label: String, dataset: Iterable[RunResult]) => {
+        f.println(s"# [$label]")
+
+        val isMemoryUsageDataset = label.endsWith(memoryDatasetQualifier)
+        dataset.foreach { r =>
+          f.println(r.size + " " + (
+            if (isMemoryUsageDataset)
```
Ichoran, line 92: Can you have one line for the if and one for the else so the logic is easier to follow at a glance?  Or maybe break out the entire thing as a helper method?  `def reportStats = ...`

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
```diff
@@ -0,0 +1,204 @@
...
+    }
+  }
+
+  /** Test getting elements from a map of `Int` to `Int`. */
+  @Benchmark
+  def put_get_Int(state: BulkGetState, bh: Blackhole) = OpenHashMapBenchmark.get_Int(state.map, size, bh)
+
+  /** Test getting elements from a map of `Int` to `Int` from which elements have been removed. */
+  @Benchmark
+  def put_remove_get_Int(state: BulkRemovedGetState, bh: Blackhole) = OpenHashMapBenchmark.get_Int(state.map, size, bh)
```
Ichoran, line 203: Is this the wrong call here?  Also, can you put the result on a second line so it doesn't disappear off the right margin?

``` scala
def put_blah =
  OHMB.blah(blah)
```

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
```diff
@@ -0,0 +1,204 @@
...
+@Fork(6)
+@Threads(1)
+@Warmup(iterations = 20)
+@Measurement(iterations = 6)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class OpenHashMapBenchmark {
+  import OpenHashMapBenchmark._
+
+  @Param(Array("25", "50", "100", "250", "1000", "2500", "10000", "25000", "100000", "250000", "1000000", "2500000",
```
Ichoran, line 173: I'm all for not adhering to margin widths set by 1980s hardware, but since you're wrapping the list of numbers anyway, can you do it more evenly?

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapRunner.scala
performantdata, line 80: This benchmark is still a work in progress. I have a Gnuplot script that I use that ought to become part of the toolkit, but haven't decided how to incorporate it yet. Some parameters would be written out to some file for inclusion in that script. This is a reminder to me.

Ichoran:
Getting close!  I think there's just one real bug--or maybe it's just unclear, in which case you should leave a comment explaining why the right method is being benchmarked.  The rest is just random style stuff.  The last step will be to squash the commits into one; there doesn't seem to be any particular value to leaving them separate is there?  (You might want to leave a branch with them separate in your repository, but it seems unnecessary for the main one.)

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
performantdata, line 203: > Is this the wrong call here?

No, it's correct. It's testing a `get()` call under different circumstances.

Ichoran, line 203: Can you try to do something with the method names to make it clearer that that is the intention?  `get_Int_on_removed` or something like that?  You do try to explain it in the comment, but it is not terribly obvious that the `OpenHashMapBenchmark.get_Int` call is correct.

performantdata, line 203: > can you put the result on a second line so it doesn't disappear off the right margin?

Admittedly it's a long line, but it fits in my IDE window. Have a specific right margin guideline, then?

Ichoran, line 203: I have it be pretty ad-hoc.  My rule is, "The more important it is, the less far to the right it should be."

In this case, what you're actually benchmarking is as important as anything, so it ought not get too far out there IMO.

performantdata, line 203: > Can you try to do something with the method names to make it clearer that that is the intention?

I'll address this when my outstanding PRs are in. Currently I have a number of data files hanging around, waiting on these PRs to close, that contain these method names, which are also in the Gnuplot script.

performantdata:
So, it's not a bug. I can address the style stuff in a subsequent commit in the PRs that I'll soon be benchmarking but haven't opened yet. Otherwise...when would you finish review?

test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala
Ichoran, line 203: There's no reason to make the commit history of the main repository messy just because you happen to have some open PRs now.  If those PRs are waiting on this, just branch.  If you need to revert, apply the reversion to the branch after rebasing on this.

(If you don't think the names should ever change, that's a different issue.)

performantdata, line 203: I'll be adding other key types, probably other value types, to the benchmarks. This isn't the last time I'll be editing this file.

> If those PRs are waiting on this

They're not. For other branches I have these files sitting around untracked. But I've generated benchmark data files containing these method names, for those PRs, and intend to keep the data until the PRs close.

> If you don't think the names should ever change

No, I agree with that, and most of your other personal syntactic preferences. I just don't think that should hold up a PR.

Ichoran, line 203: I agree with not holding up the PR.  My thought was that my attention is very limited, but I did spend some of it here, so why not take care of the problems now (if you agree there are improvements to be made)?  (Possible reason: because someone else will be reviewing the other PRs, so you won't be held back by how busy I am.)

performantdata, line 203: I'm working on the follow-up PR to #5121 right now, which is more time-sensitive.

Ichoran:
Anyway, the tests look technically correct.  There doesn't seem to be any reason to keep the separate commits, so you should squash them for the PR.  The style could use a little improvement, and I'm a little wary of a promissory note instead of an actual fix (at least for most of them--the single `get_Int` call is probably fine, though again, I'm not sure the reason to not change it is very compelling).  But if someone wants to merge this on technical soundness, it's ready.

Ichoran:
And thank you for writing a JMH harness for testing library performance!  It's long overdue in my opinion.

performantdata:
LGTRex

Ichoran:
It almost does, but not completely, which is why I avoided the normal phrase.  It would look good if the last few issues were fixed; as it is it's technically sound so could be merged anyway.

retronym:
I'm going to merge this one. Thanks for this effort, @performantdata. I think the commit history is clean enough to merge without squashing in this instance.

performantdata:
Uh...OK, thanks. I'm currently working on the fixes that Rex asked for, and adding `AnyRef` key tests. Another PR, I guess.

Thanks, Rex!

SethTisue:
we don't normally release-note contributor-facing things, but perhaps we really want to draw attention to this? not sure, untag if you disagree

SethTisue:
>  I also want to build my Gnuplot graphs automatically. More to come after this PR.

@performantdata @szeiger well I know quite some time has passed, but I'm wondering if you, or anyone else watching the repo, has anything that does this. even if it's not fully automatic, any sort of instructions I could follow. I'd love to be able to generate nice graphs like Stefan did over at https://github.com/scala/scala/pull/8534

SethTisue:
ah, @retronym has directed me towards https://github.com/lightbend/benchdb

I'll take it for a spin. (and perhaps if I'm feeling especially virtuous, I will add something about it to the benchmarking readme...)

PR #5062, opened 2016-03-24
Add back to search for Scaladoc

felixmulder:
Add basic support for navigating back to search results. This should work on web-servers as well as firefox local.

Permalinking to search results works on all browsers both local and on http.

There is one known issue. When doing `window.history.replaceState(...)` (to set the url to `.../Boolean.html?search=X`) on Chrome/Safari local, it will yield a security error. This is the reason as to why it only works on web-servers / firefox. There is an alternative - writing to `window.location.search`, unfortunately this will reload the page. Since most scala  docs will be viewed online, this should only be an issue prevalent to a tiny portion of our users.

An alternative implementation might be to switch out the hash to `#doc-search=X` - unfortunately this would break the permalink navigation. Not sure if there's a better approach to this.

preview: https://dl.dropboxusercontent.com/u/358427/scaladoc-unclutter/index.html?search=bool
review: @heathermiller, @VladUreche

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.js
+++ b/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.js
@@ -21,6 +21,20 @@ var Index = {};
     }
 })(Index);
 
+/** Find query string from URL */
+var QueryString = function(key) {
+    if (QueryString.map === undefined) { // only calc once
+        QueryString.map = {};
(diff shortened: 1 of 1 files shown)


heathermiller:
LGTM :) (when it merges cleanly, that is)

felixmulder:
@heathermiller - guessing you wanted me to update the branch? :)

heathermiller:
Thanks :)

PR #5063, opened 2016-03-25
Add summary reporting to Scaladoc

felixmulder:
This commit adds summary reporting for scaladoc. This eliminates the infinite list of `warning: doc-version (2.12.0-20160323-144128-6173274) is too long to be displayed in the webview`.

review: @VladUreche

### DIFF
--- a/src/scaladoc/scala/tools/ant/Scaladoc.scala
+++ b/src/scaladoc/scala/tools/ant/Scaladoc.scala
@@ -14,8 +14,8 @@ import org.apache.tools.ant.Project
 import org.apache.tools.ant.types.{Path, Reference}
 import org.apache.tools.ant.util.{FileUtils, GlobPatternMapper}
 
+import scala.tools.nsc.ScalaDocReporter
 import scala.tools.nsc.doc.Settings
-import scala.tools.nsc.reporters.ConsoleReporter
 
 /** An Ant task to document Scala code.
  *
@@ -666,7 +666,7 @@ class Scaladoc extends ScalaMatchingTask {
   /** Performs the compilation. */
   override def execute() = {
     val (docSettings, sourceFiles) = initialize
-    val reporter = new ConsoleReporter(docSettings)
+    val reporter = new ScalaDocReporter(docSettings)
     try {
       val docProcessor = new scala.tools.nsc.doc.DocFactory(reporter, docSettings)
       docProcessor.document(sourceFiles.map (_.toString))
--- a/src/scaladoc/scala/tools/nsc/ScalaDoc.scala
+++ b/src/scaladoc/scala/tools/nsc/ScalaDoc.scala
@@ -8,7 +8,8 @@ package scala.tools.nsc
 
 import scala.tools.nsc.doc.DocFactory
 import scala.tools.nsc.reporters.ConsoleReporter
-import scala.reflect.internal.util.FakePos
+import scala.reflect.internal.Reporter
+import scala.reflect.internal.util.{ FakePos, NoPosition, Position }
 
 /** The main class for scaladoc, a front-end for the Scala compiler
  *  that generates documentation from source files.
@@ -38,23 +39,43 @@ class ScalaDoc {
       reporter.echo(command.usageMsg)
     else
       try { new DocFactory(reporter, docSettings) document command.files }
-    catch {
-      case ex @ FatalError(msg) =>
-        if (docSettings.debug.value) ex.printStackTrace()
-        reporter.error(null, "fatal error: " + msg)
-    }
-    finally reporter.printSummary()
+      catch {
+        case ex @ FatalError(msg) =>
+          if (docSettings.debug.value) ex.printStackTrace()
+          reporter.error(null, "fatal error: " + msg)
+      }
+      finally reporter.printSummary()
 
     !reporter.reallyHasErrors
   }
 }
 
+/** The Scaladoc reporter adds summary messages to the `ConsoleReporter`
+ *
+ *  Use the `summaryX` methods to add unique summarizing message to the end of
+ *  the run.
+ */
 class ScalaDocReporter(settings: Settings) extends ConsoleReporter(settings) {
+  import scala.collection.mutable.LinkedHashMap
 
   // need to do sometimes lie so that the Global instance doesn't
   // trash all the symbols just because there was an error
   override def hasErrors = false
   def reallyHasErrors = super.hasErrors
+
+  private[this] val delayedMessages: LinkedHashMap[(Position, String), () => Unit] =
+    LinkedHashMap.empty
+
+  /** Eliminates messages if both `pos` and `msg` are equal to existing element */
+  def addDelayedMessage(pos: Position, msg: String, print: () => Unit): Unit =
+    delayedMessages += ((pos, msg) -> print)
+
+  def printDelayedMessages(): Unit = delayedMessages.values.foreach(_.apply())
+
+  override def printSummary(): Unit = {
+    printDelayedMessages()
+    super.printSummary()
+  }
 }
 
 object ScalaDoc extends ScalaDoc {
@@ -69,5 +90,21 @@ object ScalaDoc extends ScalaDoc {
 
   def main(args: Array[String]): Unit = sys exit {
     if (process(args)) 0 else 1
+  }
+
+  implicit class SummaryReporter(val rep: Reporter) extends AnyVal {
+    /** Adds print lambda to ScalaDocReporter, executes it on other reporter */
+    private[this] def summaryMessage(pos: Position, msg: String, print: () => Unit): Unit = rep match {
+      case r: ScalaDocReporter => r.addDelayedMessage(pos, msg, print)
+      case _ => print()
+    }
+
+    def summaryEcho(pos: Position, msg: String): Unit    = summaryMessage(pos, msg, () => rep.echo(pos, msg))
+    def summaryError(pos: Position, msg: String): Unit   = summaryMessage(pos, msg, () => rep.error(pos, msg))
+    def summaryWarning(pos: Position, msg: String): Unit = summaryMessage(pos, msg, () => rep.warning(pos, msg))
+
+    def summaryEcho(msg: String): Unit    = summaryEcho(NoPosition, msg)
+    def summaryError(msg: String): Unit   = summaryError(NoPosition, msg)
+    def summaryWarning(msg: String): Unit = summaryWarning(NoPosition, msg)
   }
 }
--- a/src/scaladoc/scala/tools/nsc/doc/html/Doclet.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/Doclet.scala
@@ -7,14 +7,19 @@ package scala.tools.nsc
 package doc
 package html
 
+import scala.reflect.internal.Reporter
 import doclet._
 
 /** The default doclet used by the scaladoc command line tool
   * when no user-provided doclet is provided. */
-class Doclet extends Generator with Universer {
+class Doclet(reporter: Reporter) extends Generator with Universer {
 
-  def generateImpl() {
-    new html.HtmlFactory(universe, new ScalaDocReporter(universe.settings)).generate()
-  }
+  @deprecated("Doclets should be created with the Reporter constructor. Otherwise logging reporters will not be shared by the creating parent", "2.12.0")
+  def this() = this(null)
 
+  def generateImpl() =
+    new html.HtmlFactory(
+      universe,
+      if (reporter != null) reporter else new ScalaDocReporter(universe.settings)
+    ).generate()
 }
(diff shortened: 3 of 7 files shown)


felixmulder:
This is breaking `xsbt.ScaladocInterface`, which is used by jenkins to perform a scaladoc compilation as far as I understand. As such, changing the `DocFactory`'s constructor is not a good idea.

I'll rework this.

felixmulder:
Unrelated test failed: https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/1396/consoleFull

I added an implicit conversion to `SummaryReporter` that checks if the given type is `ScalaDocReporter` and if such adds lambdas to a `delayedMessages` variable. _This passed the CI tests this morning._

I then changed one line - if the type is not of `ScalaDocReporter`, the message should be printed immediately. For some reason an unrelated test failed - not sure how I could've affected this, but re-building now...

felixmulder:
@VladUreche - no more borken tests :)

VladUreche:
@felixmulder, can you please explain what you aim to do with this patch? It may be that I'm very tired, but to me it seems wrong: you're eliminating duplicates on the message, whereas normally reporters eliminate duplicates based on (position, message). 

One more thing: Does the `mutable.Map` give you a stable order? I remember having bumped into this problem in `DiagramFactory` -- where iterating through some map was producing unstable results, and that made testing a nightmare. In the end I forced order by using `List`s.

felixmulder:
@VladUreche: Yeah sure!

So the idea is this: sometimes you'd like to issue warnings/errors/echoes that you'd like to be displayed **once** at the end of the documentation run (e.g. failing to write the `doc-version` to the webview since it's too large). The order in this instance does not really matter since it's all being shown in the end.

`mutable.Map` should not be giving me a stable order - but it wasn't something I was particularly interested in for this usecase. But WDYT - should I implement it in a fashion that prints these messages in a particular order (i.e. based on position instead)? I get that it will cover a broader range of usecases.

Something like a sorted collection of positions with messages attached might be a better idea in this case and could be made to fit elimination by `(position, message)` heuristic.

felixmulder:
I think you're right - I re-implemented according to the heuristic you specified - principle of least surprise right? :smile:

I'm on vacay in Italy - so I'll check back once a day. :ski: :mountain_cableway: :snowboarder:

VladUreche:
Have a nice vacation Felix!
Rgd the patch, there's already a similar mechanism for `deprecated`, `unchecked` and `language` features. Maybe you can reuse some of it?

felixmulder:
So this is what happens when the `deprecated` warnings appear (to my understanding): 

`compiler/scala/tools/nsc/Reporting.scala` has a `trait Reporting` that contains a class `PerRunReporting` which in turn contains a **private** `ConditionalReporting` class. The innermost class is used by `deprecated` etc. If the `-feature -deprecation` etc flags are enabled, then this class will report when a feature or deprecation is encountered in the AST. In the end they will summarize the encountered deprecations stating how many there were and how to get details on them by re-running with the correct flags (triggered from `Global.scala`).

So I can't really reuse that part of the code. However it would be feasible to override `protected def PerRunReporting` (in `Reporting.scala`) with our own `ScaladocPerRunReporting` and add a class inside it called `DelayedWarning`. But we'd still need the implicit conversion I added not to break the API - as well as the methods contained within it.

WDYT @VladUreche?

VladUreche:
Felix, are there any other usecases aside from the long version number? I mean, is the extra complexity really worth it?

felixmulder:
The only other usecase I can think of that we would like perhaps is letting people know that graphviz failed too many times (but that's not an issue on Linux with the correct version of dot). In general if we would like summary reporting of errors encountered during the parsing this would be nice.

That being said - I won't be too sad If you think this isn't worth it :)

What do the others think?

src/scaladoc/scala/tools/nsc/ScalaDoc.scala
```diff
@@ -38,23 +39,50 @@ class ScalaDoc {
...
+  import scala.collection.mutable
 
   // need to do sometimes lie so that the Global instance doesn't
   // trash all the symbols just because there was an error
   override def hasErrors = false
   def reallyHasErrors = super.hasErrors
+
+  private[this] implicit val orderingByPos: Ordering[Position] = Ordering.by(p => if (p.isDefined) p.start else -1)
+  private[this] val delayedMessages: mutable.SortedMap[Position, mutable.Map[String, () => Unit]] =
+    mutable.SortedMap.empty[Position, mutable.Map[String, () => Unit]]
```
lrytz: type args here are not necessary

src/scaladoc/scala/tools/nsc/ScalaDoc.scala
```diff
@@ -38,23 +39,50 @@ class ScalaDoc {
...
+  def addDelayedMessage(pos: Position, msg: String, print: () => Unit): Unit =
+    delayedMessages.get(pos).map { posMap =>
+      posMap += (msg -> print)
+    } getOrElse {
+      delayedMessages(pos) = mutable.Map(msg -> print)
+    }
+
+  def printDelayedMessages(): Unit = for {
+    (_, map) <- delayedMessages
+  } yield map.values.foreach(_.apply())
```
lrytz: to be pedantic: if there are multiple messages for a postion the order is still non-deterministic. i think to keep it simple you could just use ordinary maps and then get all messages into one list here and sort it.

src/scaladoc/scala/tools/nsc/ScalaDoc.scala
```diff
@@ -70,4 +98,20 @@ object ScalaDoc extends ScalaDoc {
   def main(args: Array[String]): Unit = sys exit {
     if (process(args)) 0 else 1
   }
+
+  implicit class SummaryReporter(rep: Reporter) {
```
lrytz: could `extend AnyVal`

src/scaladoc/scala/tools/nsc/doc/html/Doclet.scala
```diff
@@ -12,9 +12,17 @@ import doclet._
 /** The default doclet used by the scaladoc command line tool
   * when no user-provided doclet is provided. */
 class Doclet extends Generator with Universer {
+  import scala.reflect.internal.Reporter
 
-  def generateImpl() {
-    new html.HtmlFactory(universe, new ScalaDocReporter(universe.settings)).generate()
```
lrytz, line 17: the change here (and the type change from `ScalaDocReporter` to `Reporter` in various places`) seems to be unrelated to delayed reporting - what was the motivation?

lrytz:
it doesn't look too complex, i think it would be a valuable change

src/scaladoc/scala/tools/nsc/doc/html/Doclet.scala
felixmulder, line 17: When creating a `Doclet` the user would get a different reporter than the one specified when creating a `DocFactory`, this caused incorrect behaviour when printing the summary messages on the external reporter (i.e. no messages).

Since we're not forcing the user to create a `DocFactory` with a`ScalaDocReporter`, I thought it best to change the members to `Reporter` instead. WDYT?

src/scaladoc/scala/tools/nsc/ScalaDoc.scala
felixmulder: Just so I'm following - you're proposing a single map `mutable.Map[(Position, String), () => Unit]` and then sorting it in the end?

Or should I simplify it to `mutable.SortedMap[(Position, String), () => Unit]` that is ordered by the position?

lrytz: actually you could even do a `LinkedHashMap[(Position, String), () => Unit]`, as the insertion order should be deterministic, wdyt?

src/scaladoc/scala/tools/nsc/doc/html/Doclet.scala
lrytz, line 17: sounds right.

could we make the reporter a constructor argument? i see that Doclet is created reflectively. but maybe we can change that code to do something like

```
docletClass. getConstructors()
// find the one that takes a reporter, invoke it
// if it doesn't exist, invoke the no-args one, and maybe emit a deprecation warning saying
// that the Generator should now have a 1-arg constructor
```

src/scaladoc/scala/tools/nsc/ScalaDoc.scala
felixmulder: Good idea! I'll make the changes and ping you :)

src/scaladoc/scala/tools/ant/Scaladoc.scala
```diff
@@ -666,7 +666,7 @@ class Scaladoc extends ScalaMatchingTask {
   /** Performs the compilation. */
   override def execute() = {
     val (docSettings, sourceFiles) = initialize
-    val reporter = new ConsoleReporter(docSettings)
+    val reporter = new ScalaDocReporter(docSettings)
```
felixmulder, line 669: `ant docs` runs this file - so we should use `ScalaDocReporter` here instead to get the new behaviour.

felixmulder:
@lrytz: suggested changes in place :+1:

I must say though, that I don't really like the reflection instantiation - not sure what the usecase is. Nevertheless, this should be backwards compatible if anyone is using the `-doc-generator` flag.

src/scaladoc/scala/tools/nsc/doc/html/Doclet.scala
```diff
@@ -7,14 +7,19 @@ package scala.tools.nsc
...
 /** The default doclet used by the scaladoc command line tool
   * when no user-provided doclet is provided. */
-class Doclet extends Generator with Universer {
+class Doclet(reporter: Reporter) extends Generator with Universer {
 
-  def generateImpl() {
-    new html.HtmlFactory(universe, new ScalaDocReporter(universe.settings)).generate()
-  }
+  @deprecated("Doclets should be created with the Reporter constructor. Otherwise logging reporters will not be shared by the creating parent", "2.12.0")
+  def this() { this(null) }
```
som-snytt: It's seems more natural to me to supply the default reporter in the ctor, rather than checking for null in the method.

Also, `def this() = this(new ScalaDocReporter)` looks better.

felixmulder: I could change it to `def this() = this(null)`, however, I think it'd be hard to change it to:

`def this() = this(new ScalaDocReporter)` since the reporter constructor takes a settings value and that setting is created by the `universe` variable, which in turn isn't set until somebody calls `setUniverse` :(

som-snytt:
I think the difficulty of extending the summary mechanism goes to the big TODO about making it configurable (or at least extensible). 2.12 might be the time to at least tweak it so you can add a custom summary warning. These API were still in flux after refactoring. This might be a test case.

lrytz:
LGTM, maybe @VladUreche would like to take another look?

VladUreche:
@felixmulder, can you please un-import global.reporter in all scaladoc files, to make sure you use the doclet reporter?

Otherwise LGTM, thanks for the contribution!

VladUreche:
Btw, needs rebase :(

felixmulder:
@VladUreche - rebased!

adriaanm:
Still needs rebase, it looks like -- I usually do `git pull --rebase https://github.com/scala/scala.git 2.12.x`

felixmulder:
@adriaanm I usually do:

``` bash
$ scalarebase # alias: git fetch upstream && git rebase -i upstream/2.12.x
```

But eyyy waddayano :)

felixmulder:
@VladUreche - I'm not sure which reporters you want me to remove from the imports - the ones specified in the changes need to be there in order to get `Reporter` in scope...or am I missing something?

som-snytt:
Just a guess, but Vlad might mean to `import global.{reporter => _, _}` where appropriate. Sorry to offer a guess.

felixmulder:
@som-snytt - don't be sorry :) I'll wait for his confirmation, there's no rush on my part with this PR

VladUreche:
@som-snytt read my mind :)

The problem I see is that now we have two reporters:
- one from `import global._` imported into scope
- one from `doclet.reporter` which is in scope

Now, if we `import global._` we're using the `global`'s reporter. I don't know if that's a problem or not, all I'm saying it's that someone, sometime may stumble upon this irregularity and hate us for that.

An even better solution is to call the doclet's reporter `docletReporter` and replace all calls to `global.reporter.warning` and `reporter.warning` by `docletReporter.warning`. This will make it clear which reporter we're using and where.

WDYT?

adriaanm:
ping @felixmulder

felixmulder:
@adriaanm: I changed the reporters to `docletReporter`. Anything else that was on your mind? :)

adriaanm:
I was scanning through the PR queue and saw a question from Vlad but no response, so I pinged. I guess you said it with a commit, thanks :-)

som-snytt:
Is that the pink "Women in Programming" label? Lightbend is the thought leader. I get `SJW` mixed up with `SPJ`.

adriaanm:
Where's the badge for best retroactive acronym when you need it? Here's some more ice cream for you 🍨  (scooped, not shaved)

PR #5065, opened 2016-03-27

ktully:
Minor fixup to documentation in package.scala to recommend JavaConverters over the now deprecated JavaConversions.

### DIFF
--- a/src/library/scala/collection/package.scala
+++ b/src/library/scala/collection/package.scala
@@ -78,13 +78,13 @@ package scala
  *
  * === Converting between Java Collections ===
  *
- * The [[scala.collection.JavaConversions]] object provides implicit defs that
- * will allow mostly seamless integration between APIs using Java Collections
- * and the Scala collections library.
- *
- * Alternatively the [[scala.collection.JavaConverters]] object provides a collection
+ * The [[scala.collection.JavaConverters]] object provides a collection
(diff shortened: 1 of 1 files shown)


som-snytt:
Thanks, I missed that one. Who knew the docs were so thorough?

I wouldn't bother mentioning the deprecated API here, however. Out of sight, out of mind. Also, you won't have to come back in 18 months to change the doc again where it's removed.

If you prefer, I'll make the change on my PR, or you can do it. Be aware that doc edits break reflection tests.

ktully:
Thanks for your feedback, withdrawing my PR, happy for you to update in your PR.

Fair point about rewording to minimise future maintenance on the doc. I did consider a more aggressive edit, but wasn't sure what usual practice is for this project.

I didn't realise the tests would fail when documentation was updated, I see that now they've completed on my PR. 

I need to look into why my local build appeared to succeed (I'd never built Scala before today :).

som-snytt:
Just joking about tests failing. I did see some weird reflection test failure, but no idea what that was about...  I'll tack an edit onto my PR and credit you.

som-snytt:
Added to https://github.com/scala/scala/pull/5064 with the suggested edit.

ktully:
@som-snytt, thanks for the credit on your PR. I guess the challenge now is to sneak a test in that really will fail when a ScalaDoc is edited :)

PR #5067, opened 2016-03-29
SI-9724 don't cache hash codes in OpenHashMap

performantdata:
This includes the the changes from #5061, since benchmarking is required, so that PR should go in first. See [the JIRA ticket](https://issues.scala-lang.org/browse/SI-9724) for the performance graphs.


retronym:
Keys of type `java.lang.Integer` would represent a best case for the non-caching strategy, the key itself "caches" its hashcode by defining it as equal the the wrapped value. 

It isn't zero-overhead: calling `key.##`, which incurs a few type tests in `BoxesRuntime.hashXxx` will incur a little bit of overhead. Avoiding that overhead is the motivation for `scala.collection.mutable.AnyRefMap`.

Could you add a benchmark for keys with increasingly expensive `hashCode` operations? This could be simulated with `BlackHole.consumeCPU(tokens)` in JMH.

performantdata:
I think that  the cost of `Key.==`, which is also probably cheap for `Integer`s, will have more of an impact. `Key.##` is only avoided on resizing the table, and on `clone()`.

This is more than I want to benchmark right now, because I'd need to explore both methods, and the benchmark run time is still around eight hours. I'll try to reduce the benchmarking time first.

PR #5068, opened 2016-03-29
Accomodate and exploit new library, lang features JDK 8

retronym:


sjrd:
Speaking of `unsignedCompare`, there's a better implementation: `(i ^ Int.MinValue) < (j ^ Int.MinValue)`. For longs, replace `Int.MinValue` by `Long.MinValue`

soc:
@retronym 
- Regarding https://github.com/scala/scala/pull/5068/commits/d40f1b81372e1cdb101790d14c5bc8c7c304c128, the use of `addSupressed` looks fine, but the test seems to check all items in the list of issues I wrote for the ticket orignally:
  
  > How many exceptions are observed before the operation is aborted depends on the machine, the available cores and hyper-threading, the operating system, the threadpool implementation and configuration, the size of the collection and the runtime itself.
- Regarding https://github.com/scala/scala/pull/5068/commits/eb6e29d27eb9cac0430fb34c485788cf3108fe00, what's preventing us from just starting to use `writableStackTrace = false` for `NoStackTrace`? The config option?
- https://github.com/scala/scala/pull/5068/commits/71aab33d462de90dd4831523d7161fd7b8678d9b looks fine, thanks!

retronym:
> Regarding eb6e29d, what's preventing us from just starting to use writableStackTrace = false for NoStackTrace? The config option?

Just the fact that `NoStackTrace` and `ControlThrowable` are traits, and they can't express the pattern that all subclasses should call a particular super constructor with a particular parameter. `ControlThrowable` at least would have to persist as it can be used to pass through that category of exceptions in a handler.

soc:
@retronym oh, right, sorry!

fommil:
nice :smile: btw, in your travels of Java 7/8 APIs did you find any replacements for `URLClassLoader` that don't keep `JarFile`s lying around? I'm _seriously_ considering writing one because it's giving me a headache.

retronym:
@fommil I've pushed a commit to close and recreate the URL classloader created by the compiler to invoke macros at the end of each compilation run. This is a first step towards better resource hygiene.

fommil:
@retronym meanwhile, in sbt... https://github.com/fommil/sbt-big-project/blob/master/src/main/scala/BigProjectSettings.scala#L421 (search for `ClassLoaderHack`).

If there was an alternative to `URLClassLoader`, that would be awesome. I'm seriously considering writing a new classloader but deriving anything directly is subject to the GPLv2 licence of OpenJDK and that won't be palatable to scala-compiler or sbt. I'll probably just stick with the hack and hope our general tooling hygiene improves as you're paving the way.

fommil:
(btw, I think if it was possible to hotswap an implementation of `URLClassPath` that would get us there, it's not a final class so it is **possible**)

fommil:
@retronym I don't think writing our own `URLClassPath` (backend of `URLClassLoader`) to use NIO instead of `JarFile` would be all that hard. Thoughts here: https://github.com/fommil/sbt-big-project/issues/30#issuecomment-203414024 It might even be the sort of thing that gets contributed back to OpenJDK

lrytz:
LGTM

mkurz:
It looks like this also fixed https://issues.scala-lang.org/browse/SI-9853

PR #5069, opened 2016-03-30
Add type and class/trait member search, fixes SI-9721 and SI-9722

felixmulder:
This patch allows searching of more kinds of members - namely type aliases and inner classes/traits.

It also fixes searchability for members like `#::` and `##`.

issues: https://issues.scala-lang.org/browse/SI-9721, https://issues.scala-lang.org/browse/SI-9722
preview: https://dl.dropboxusercontent.com/u/358427/scaladoc-search-fixes/index.html
review: @VladUreche, @tpolecat


VladUreche:
LGTM!

PR #5070, opened 2016-03-31
Fix typo in the docs for the ++ method of Stream

markus1189:
Fixes a typo in the scala docs for `Stream.++`, more precisely in the docs for the param `that`:

> @param that The [[scala.collection.GenTraversableOnce]] **the** be concatenated to this `Stream`.

changed to

> @param that The [[scala.collection.GenTraversableOnce]] **to** be concatenated to this `Stream`.


lrytz:
LGTM, thank you!

PR #5071, opened 2016-03-31
Add initial unit test for Catch and augment documentation

janekdb:
- Add unit test for andFinally
- Reduce code duplication in andFinally
- Extend documentation


lrytz:
LGTM!

PR #5072, opened 2016-04-01
SI-6710 Clarify stub methods in primitive value classes

lrytz:
- Replaces the implementations of box/unbox in AnyVal companions by
  `???`, the methods are only stubs, and the impls did not correspond
  to the actual behavior. The doc comment already points to the actual
  implementation in BoxesRunTime.
- Replaces the body of `getClass` from `null` to `???` and clarifies in
  a comment why the overrides exist.

JIRA: https://issues.scala-lang.org/browse/SI-6710

### DIFF
--- a/src/compiler/scala/tools/cmd/gen/AnyVals.scala
+++ b/src/compiler/scala/tools/cmd/gen/AnyVals.scala
@@ -200,7 +200,8 @@ import scala.language.implicitConversions"""
     def classLines: List[String]
     def objectLines: List[String]
     def commonClassLines = List(
-      "override def getClass(): Class[@name@] = null"
+      "// Provide a more specific return type for Scaladoc",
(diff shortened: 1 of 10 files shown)


adriaanm:
That was fast :rocket:

demobox:
Yay, thanks @lrytz! Really glad to see this is finally happening after all!

lrytz:
> That was fast

Yeah, the issue was only filed on Nov 25 2012 :hear_no_evil:

lrytz:
review by @szeiger ?

szeiger:
LGTM. We should automate Codegen in the sbt build (similar to what I've done for `genprod`). I'll create a PR for that. (Eventually these different generators should be unified and run automatically but that's best done after getting rid of ant)

PR #5074, opened 2016-04-02
SI-9732 remove the use of Option from OpenHashMap

performantdata:
See [the JIRA ticket](https://issues.scala-lang.org/browse/SI-9732) for the performance graphs.

### DIFF
--- a/src/library/scala/collection/mutable/OpenHashMap.scala
+++ b/src/library/scala/collection/mutable/OpenHashMap.scala
@@ -17,16 +17,23 @@ package mutable
  *  @since 2.7
  */
 object OpenHashMap {
-
   def apply[K, V](elems : (K, V)*) = new OpenHashMap[K, V] ++= elems
   def empty[K, V] = new OpenHashMap[K, V]
 
   final private class OpenEntry[Key, Value](val key: Key,
                                             val hash: Int,
-                                            var value: Option[Value])
-                extends HashEntry[Key, OpenEntry[Key, Value]]
+                                            var value: Value)
 
   private[mutable] def nextPositivePowerOfTwo(i : Int) = 1 << (32 - Integer.numberOfLeadingZeros(i - 1))
+
+  // Hash table slot states
+
+  /** The hash table slot is empty. */
+  private val Empty: Byte = 0
+  /** The hash table slot was deleted. */
+  private val Deleted: Byte = 1
+  /** The hash table slot is occupied. */
+  private val Occupied: Byte = 2
 }
 
 /** A mutable hash map based on an open hashing scheme. The precise scheme is
@@ -52,7 +59,7 @@ extends AbstractMap[Key, Value]
    with Map[Key, Value]
    with MapLike[Key, Value, OpenHashMap[Key, Value]] {
 
-  import OpenHashMap.OpenEntry
+  import OpenHashMap._
   private type Entry = OpenEntry[Key, Value]
 
   /** A default constructor creates a hashmap with initial size `8`.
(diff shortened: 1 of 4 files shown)


src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -64,7 +72,22 @@ extends AbstractMap[Key, Value]
   private[this] val actualInitialSize = OpenHashMap.nextPositivePowerOfTwo(initialSize)
 
   private var mask = actualInitialSize - 1
-  private var table : Array[Entry] = new Array[Entry](actualInitialSize)
+
+  /** The state of each of the slots in the hash table.
+    * One of `Empty`, `Deleted` or `Occupied`.
+    */
+  private[this] var slotStates = new Array[Byte](actualInitialSize)
```
retronym, line 78: Did you try to change `OpenEntry` to have a `Value` + `Boolean`, rather than introducing this? On my architecture, I think you'd get that for free:

```
$ sbt
sbt> libraryDependencies += "org.openjdk.jol" % "jol-core" % "0.4"

sbt> consoleQuick
...
scala> println(VMSupport.vmDetails())
WARNING: Unable to attach Serviceability Agent: Error: Could not find or load main class org.openjdk.jol.util.HS_SA_Support

WARNING: VM details, e.g. object alignment, reference size, compressed references info will be guessed.

Running 64-bit HotSpot VM.
Using compressed oop with 3-bit shift.
Using compressed klass with 3-bit shift.
Objects are 8 bytes aligned.
Field sizes by type: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]
Array element sizes: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]

scala> ClassLayout.parseClass(Class.forName("scala.collection.mutable.OpenHashMap$OpenEntry")).toPrintable()

"scala.collection.mutable.OpenHashMap.OpenEntry object internals:
 OFFSET  SIZE   TYPE DESCRIPTION                    VALUE
      0    12        (object header)                N/A
     12     4    int OpenEntry.hash                 N/A
     16     4 Object OpenEntry.key                  N/A
     20     4 Option OpenEntry.value                N/A
     24     4 Object OpenEntry.next                 N/A
     28     4        (loss due to the next object alignment)
Instance size: 32 bytes (estimated, the sample instance is not available)
Space losses: 0 bytes internal + 4 bytes external = 4 bytes total
```

performantdata, line 78: I didn't try to add a `Byte` to `OpenEntry`. The thought crossed my mind, since that would preserve memory locality. But the direction that I'm headed is to eliminate the overhead of objects like `OpenEntry`, to support higher capacities than a chained hash map in a JVM. Because, if this class isn't good for that, then what is it good for?

I didn't realize that `OpenEntry`'s inheritance from `HashEntry` was adding an extra word. I assumed that it wasn't adding overhead, since it wasn't adding feature. The response to that, though, is to remove that dependence.

SethTisue:
if you merge the latest 2.11.x changes into this PR, it should pass validation

lrytz:
oh, I didn't see this PR before; my comment here https://github.com/scala/scala/pull/5124#issuecomment-216172287 is related.

src/library/scala/collection/mutable/OpenHashMap.scala
performantdata, line 78: I removed `HashEntry` trait here, rather than open another PR. I need to retest the performance over the next couple evenings.

performantdata, line 78: I updated the performance graphs in the JIRA ticket. It's gotten a bit better, and a bit worse.

What bothers me is why the `Some` isn't being elided in the `getOrElse(0)` call. It's sticks around also if I use the `AbstractMap.getOrElse(i, 0)`. Building tests with `-Yopt:l:classpath`.

performantdata:
review by @retronym ? You seem to have started already.

performantdata:
I'm going to rewrite this substantially, once #5183 is in.

SethTisue:
should we keep this open?

lrytz:
let's close it since there was no activity in a very long time. @performantdata feel free to re-open once you think it should be reviewed.

performantdata:
There are more substantial changes that I want to do that require API changes, which requires a 2.13.x branch, since they missed 2.12.0-M5. More satisfying to cut to the chase. Was planning to treat issues like this as a backport, if anyone else cared.

SethTisue:
note that `OpenHashMap` is now deprecated

PR #5078, opened 2016-04-04
Generate AnyVal source stubs from sbt

szeiger:
The sbt command `generateSources` now generates both the AnyVal
sources and the tuple/function sources (previously done by `genprod`).

Source generation is part of the sbt build, the `scala.tools.cmd.gen`
package is removed from `scala-compiler`. This simplifies bootstrapping.
Generated sources are still in the same location and checked into git.


build.sbt
```diff
@@ -649,11 +649,13 @@ lazy val root = (project in file("."))
...
     publishLocal := {},
     commands ++= ScriptCommands.all,
-    // Generate (Product|TupleN|Function|AbstractFunction)*.scala files. Once the ANT build is gone,
-    // we should move them into a managedSources dir instead of overwriting sources checked into git.
-    commands += Command.command("genprod") { state =>
-      val dir = (((baseDirectory in ThisBuild).value) / "src" / "library" / "scala").getPath
-      genprod.main(Array(dir))
+    // Generate (Product|TupleN|Function|AbstractFunction)*.scala files and scaladoc stubs for all
+    // AnyVal sources. Once the ANT build is gone, we should move them into a managedSources dir instead
+    // of overwriting sources checked into git.
```
lrytz: Not sure about this TODO, because there's a link to the source file on GitHub from the generated ScalaDoc (http://www.scala-lang.org/files/archive/api/2.12.0-M4/scala/Int.html)

szeiger: Ah, that's unfortunate. We could easily include the generated sources in source JARs but if the links go directly to github, we're out of luck. Is there a way to disable source links selectively? It's only a stub that doesn't really add any value on top of the scaladoc page, so we could just remove the link if that's possible.

szeiger: Another option would be to include the generated sources in the scaladoc jars and rewrite the generated links to point to them. The sbt build should make custom processing like this reasonably simple to implement.

sjrd: Not checking files into git, although desirable in theory, would be problematic for the source maps of Scala.js.

lrytz: @szeiger can you update the comment saying "in principle we could generate them in src_managed, but for scaladoc and scala-js we don't"?

szeiger:
Updated comments.

lrytz:
LGTM, worked for me.

PR #5080, opened 2016-04-04
WIP^2: Remove repl import wrappers

adriaanm:
@retronym figured out a much nicer way to enforce proper scoping in the repl, which removes some of the wrapping we need to do.

This would be nice to have for M5, but no show stopper.

I'm throwing this PR out there because I don't have time to work on this right now, but I think this would make for a nice improvement. Would be grateful if someone could pick this up.

### DIFF
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -665,12 +665,6 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     final def isLazyAccessor       = isLazy && lazyAccessor != NoSymbol
     final def isOverridableMember  = !(isClass || isEffectivelyFinal) && safeOwner.isClass
 
-    /** Does this symbol denote a wrapper created by the repl? */
-    final def isInterpreterWrapper = (
-         (this hasFlag MODULE)
-      && isTopLevel
-      && nme.isReplWrapperName(name)
-    )
 
     /** In our current architecture, symbols for top-level classes and modules
      *  are created as dummies. Package symbols just call newClass(name) or newModule(name) and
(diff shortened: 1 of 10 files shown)


som-snytt:
WIP^2 looks like git notation. I was going to try to help via dotty, but dotty also took some weekend hours. I'll try to follow up, but otherwise let me +1.

dragos:
I'll test this on the Spark Shell as soon as it builds.

adriaanm:
It's alive! Test failures expected -- hadn't looked at those yet.

som-snytt:
A wrapping fix for class-based at https://github.com/scala/scala/pull/5090 or I guess we should call it a context fix now.

som-snytt:
WIP: PIW. Purge Import Wrappers.

Some appreciation over at https://github.com/scala/scala/pull/5194, still in progress.

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -852,7 +846,6 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     def isEmptyPrefix = (
          isEffectiveRoot                      // has no prefix for real, <empty> or <root>
       || isAnonOrRefinementClass              // has uninteresting <anon> or <refinement> prefix
-      || nme.isReplWrapperName(name)          // has ugly $iw. prefix (doesn't call isInterpreterWrapper due to nesting)
```
som-snytt, line 855: I guess results in `$read.Foo` and `INSTANCE.Bar`. Hopefully restoring the check with `$read` is kosher and sufficient. I didn't know about these machinations. It makes me wonder if it can be generalized; people want to customize printing of types (simple name, or not; or how about always seeing `collection.Seq` etc).

adriaanm:
👾  pew! pew! 👾  there goes this pr

over to you

PR #5081, opened 2016-04-04
Bump STARR to 2.12.0-M4.

adriaanm:

PR #5082, opened 2016-04-04
Cleanups related to the removal of trait impl classes

lrytz:


lrytz:
Fixes https://github.com/scala/scala-dev/issues/90

adriaanm:
Let's base this on top of #5081 (should be ready to merge in 20 min or so) to standardize on M4 as STARR.

lrytz:
> Let's base this on top of #5081

why does this matter?

adriaanm:
I guess it doesn't really matter, just my OCD :-)

lrytz:
review by @adriaanm

src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
```diff
@@ -132,16 +132,16 @@ class CallGraph[BT <: BTypes](val btypes: BT) {
               (declarationClassNode, source) <- byteCodeRepository.classNodeAndSource(declarationClass): Either[OptimizerWarning, (ClassNode, Source)]
             } yield {
                 val declarationClassBType = classBTypeFromClassNode(declarationClassNode)
-                val CallsiteInfo(safeToInline, safeToRewrite, canInlineFromSource, annotatedInline, annotatedNoInline, samParamTypes, warning) = analyzeCallsite(method, declarationClassBType, call, source)
+                val info = analyzeCallsite(method, declarationClassBType, call, source)
+                import info._
```
adriaanm, line 136: neat pattern for multi-value return

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
```diff
@@ -592,11 +578,10 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
 
   /**
    * True for module classes of modules that are top-level or owned only by objects. Module classes
-   * for such objects will get a MODULE$ flag and a corresponding static initializer.
+   * for such objects will get a MODULE$ field and a corresponding static initializer.
    */
   final def isStaticModuleClass(sym: Symbol): Boolean = {
-    /* (1) Phase travel to to pickler is required to exclude implementation classes; they have the
-     * lateMODULEs after mixin, so isModuleClass would be true.
+    /* (1) Phase travel to pickler to exclude the lateMODULEs flag.
```
adriaanm: so do we no longer need this?

lrytz: yeah i was wondering and didn't dare.. but looking again at the details again it should be fine, i'll give it a try.

adriaanm:
LGTM

PR #5083, opened 2016-04-05
use Future#foreach instead of onSuccess in example code

xuwei-k:
bacause Future#onSuccess is deprecated
https://github.com/scala/scala/commit/fa0743c32338f147eaf7a


retronym:
LGTM, thanks.

PR #5084, opened 2016-04-05
SI-9734 Narrow type when import REPL history

som-snytt:
Under `-Yrepl-class-based`, imports from historical `$read`
instances must be singleton-typed so that path-dependent types
remain so.


som-snytt:
IDE build failure. I'll let someone else request rebuild when it's safe to do so.

SethTisue:
tracking the IDE thing at scala/scala-dev#104 (investigation underway)

lrytz:
LGTM!

PR #5086, opened 2016-04-06
SI-9735 REPL prefer standard escapes for code text

som-snytt:
When constructing code text for compilation, the REPL
should prefer standard escape sequences, in case unicode
escapes are disabled.


lrytz:
LGTM!

PR #5089, opened 2016-04-07
Documents date fixed + tool for fixing

FelAl:
heathermiller 13:29
you should talk to @adriaanm about this. several people have made pull requests to fix this. it's been maybe 1.5 years since the last one. as far as i understand, the lightbend scala team would rather wait until someone writes a tool for this
but it's been some time and i have not seen one

SethTisue
@FelAl Adriaan figures if a source file hasn’t been changed there’s no need to bump the copyright year. we try to remember to bump the date on files that actually change. I believe a PR that just updated every year to 2016 would be rejected, but one that brought the dates in line with the last-modified years according to Git might be accepted, especially if it included a script that could be used to do the same in the future

Usage example:
http://pastebin.com/JDL324XA

I appreciate any comments. Thank you for your attention.

### DIFF
--- a/scripts/jobs/integrate/windows
+++ b/scripts/jobs/integrate/windows
@@ -4,12 +4,13 @@
 
 export ANT_OPTS="-Dfile.encoding=UTF-8 -server -XX:+AggressiveOpts -XX:+UseParNewGC -Xmx2G -Xss1M -XX:MaxPermSize=512M -XX:ReservedCodeCacheSize=128M"
 
-# TODO: don't hardcode this path, which is just where we currently have
-# ant manually installed on jenkins-worker-windows-publish.
-PATH=/cygdrive/c/apache-ant-1.9.6/bin:$PATH
-
+# TODO: don't hardcode this path -- should be in scala/scala-jenkins-infra
+export PATH='/cygdrive/c/java/jdk-1.6/bin:/cygdrive/c/apache-ant-1.9.6/bin:/cygdrive/c/Program Files (x86)/Git-2.5.3/Cmd:/bin:/usr/bin:'
 export JAVA_HOME='C:/java/jdk-1.6'
-export PATH="$JAVA_HOME/bin:$PATH"
+
+java -version
+javac -version
+ant -version
 
 ant \
   -Dstarr.version=2.11.8 \


adriaanm:
Cool, thank you! Would be great to have this integrated into the sbt build so that we could run it during CI.

FelAl:
Ok, I'll fix it.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    } else {
+      return true
+    }
+  }
+
+  def analyzeDates(): Array[(String, String, String)] = {
+    val files_array = readFilesList()
+    var data = Array[(String, String, String)]()
+    for(file_path <- files_array) {
+      if(readFile(file_path).contains("(c) 2003-") && ifFileWasInGit(file_path)) {
```
Ichoran: What if a file wasn't written in 2003?

FelAl: It will be ignored.

For the most files we have next situation.

```
alfe@rt:~/mkdev/bigdata/scala/scala/src/library/scala/collection/generic$ git log --format=%cd Clearable.scala
Fri May 3 11:36:37 2013 -0700
Fri Nov 2 18:12:57 2012 +0100
Wed May 2 18:23:14 2012 -0700
```

First date in git - 2012, inside the file - 2003.
For future files, I may fix this. But now I don't know, what is the point?

Ichoran: The point is accuracy. One cannot legitimately claim copyright from 2003 on something that wasn't created until 2015.

That is why your tool is so helpful--people were updating copyrights without evidence that work was done!  This is way better.

Because the git history is not exhaustively complete, one can't set the early dates automatically, but if someone writes the correct date in rather than incorrectly backdating it, it would be great if the tool could still update the most recent date.

What is the error rate if you just search for "(c) 20"?  Would that work?  (Maybe with an additional check that the next three characters match the regex `"\\d\\d-"`?)

FelAl: Yeah, I see my error. Thank you.
https://github.com/scala/scala/blob/2.12.x/src/library/scala/math/ScalaNumericConversions.scala
It's (c) 2007 and I didn't consider such files.

What is the error rate if you just search for "(c) 20"? Would that work? (Maybe with an additional check that the next three characters match the regex "\d\d-"?)

I'll update my solution and we'll see it. But I think (c) 20 will do.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    fixWrongDates - actual fixing
+
+    From root folder:
+    Usage: scala tools/CheckDates.scala analyzeDates
+           scala tools/CheckDates.scala fixWrongDates
+
+  """
+
+  def readFilesList(): Array[String] = {
+    val myval =  ("find . -name *.scala ! -path ./tools/*  -type f") !!
```
janekdb: Single quotes around the pattern?

```
'*.scala'
```

This will prevent the find from failing when there is a stray .scala file at the root.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    val files_array = myval.split("\n").map(_.toString).distinct
+    
+    return files_array
```
janekdb: `return` keyword superfluous.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)
```
janekdb: Any reason to not be explicit about the character encoding?

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    val lines = try source.mkString finally source.close()
+
+    return lines
```
janekdb: `return keyword superfluous.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    return data
+  }
+
+  def fixWrongDates(): Unit = {
+    val files = analyzeDates()
+    files.foreach {file =>
+      if (file._2.toInt != file._3.toInt) {
+        val actual_file = readFile(file._1)
+        val new_file = actual_file.replace(s"(c) 2003-${file._2}", s"(c) 2003-${file._3}")
+        scala.tools.nsc.io.File(file._1).writeAll(new_file)
```
janekdb: The file encoding should be supplied explicitly.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+  }
+
+  def resultAnalyzeDates(): Unit = {
+    val data = analyzeDates
+    var bool = false
```
janekdb: A less generic local var name would be preferable.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    data.foreach {file =>  
+      if (file._2.toInt != file._3.toInt) {
+        println(file + " wrong date ")
+        bool = true
+      } else {
+        println(file + " ok ")
```
janekdb: Do you think it might be better to only print when there is a problem? Less console output.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+    if(bool) println("Oops...We have some error dates")
+    else println ("Everything is ok")
+  }
+
+  def main(args: Array[String]) {
```
janekdb: Extend `App`?

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    if (args.length == 0) println(usage)
+    if (args.length > 0) {
+      val operation = args(0)
+      operation match {
+        case "analyzeDates" => resultAnalyzeDates
```
janekdb: Consider returning a status value from `resultAnalyzeDates` and `fixWrongDates`. Use that as the command result code. This will make the tool more useful in shell scripts.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+        println(file + " ok ")
+      }
+    }
+    if(bool) println("Oops...We have some error dates")
+    else println ("Everything is ok")
+  }
+
+  def main(args: Array[String]) {
+    if (args.length == 0) println(usage)
+    if (args.length > 0) {
```
janekdb: If you want to be stricter fail when `args.length != 1`

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    return lines
+  }
+
+  def githubDate(pathToFile: String): String = {
+    val buf_dates = (s"git log -3  --format=%cd $pathToFile" !!)
+    val buf_messages = (s"git log -3  $pathToFile" !!)
+    val messages = buf_messages.split("commit").distinct.drop(1)
+    val dates = buf_dates.split("\\n").distinct
+    var gits_info = Array[(String, String)]()
+    for (a <- 0 to dates.length - 1) {
```
janekdb: zip messages and dates then map over the resulting list?

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+      gits_info = gits_info :+ (messages(a),dates(a))
+    }
+    val res = gits_info.filter(!_._1.contains("Documents date fixed"))
```
janekdb: Deploying a case class will allow the _n syntax to be dropped resulting in more maintainable code.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+        val new_file = actual_file.replace(s"(c) 2003-${file._2}", s"(c) 2003-${file._3}")
+        scala.tools.nsc.io.File(file._1).writeAll(new_file)
+      }
+    }
+  }
+
+  def resultAnalyzeDates(): Unit = {
+    val data = analyzeDates
+    var bool = false
+    data.foreach {file =>
```
janekdb: Consider partitioning the data list by the predicate on line 87. Then you can drop `bool` in exchange for a `notEmpty` call on one side of the partition.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
+import sys.process._
```
janekdb: Add copyright header? :smile:

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+
+object CheckDates {
+  val usage = """
+
+    This tool compares a copyright date in files with the date of last change in git.
+    If the message of last change contains "Documents date fixed" it will be ignored and
```
janekdb: Suggestion for the way commits are identified as being copyright date updates. Instead of requiring the string "Documents data fixed" somewhere in the commit body adopt a less variable approach by requiring a prefix in the commit subject,

```
[Sync copyright date to last commit]
```
- It's in the imperative voice
- The square brackets give the feeling of a category label

An additional benefit of relying only on the subject line is this will halve the number of calls `git log` as a format specifier can be used to include the subject in a string with a parsable format.

tools/CheckDates.scala
```diff
@@ -0,0 +1,110 @@
...
+    }
+    if(bool) println("Oops...We have some error dates")
+    else println ("Everything is ok")
+  }
+
+  def main(args: Array[String]) {
+    if (args.length == 0) println(usage)
+    if (args.length > 0) {
+      val operation = args(0)
+      operation match {
```
janekdb: Convert args to a List and the number of code paths can be reduced.

```
val argList = args.toList
argList match {
  case "analyzeDates" :: Nil =>  resultAnalyzeDates
  case "fixWrongDates" :: Nil  => fixWrongDates
  case _ => println(usage)
 }
```

FelAl:
Ichoran, I made fix for creation dates.
Also, I have not fixed code style yet.(janekdb comments)

FelAl:
Maybe I need to split my PR in 3-4 smaller PRs?

FelAl:
@janekdb janekdb added a note 2 days ago
Single quotes around the pattern?

'*.scala'
This will prevent the find from failing when there is a stray .scala file at the root.

Scala REPL

```
scala> ("find . -name *.scala ! -path ./tools/*  -type f") !!
res0: String =
"./test/script-tests/jar-manifest/src/jar-test.scala
./test/osgi/src/BasicReflection.scala
./test/osgi/src/ScalaOsgiHelper.scala
./test/osgi/src/ReflectionToolboxTest.scala
./test/osgi/src/BasicLibrary.scala
./test/osgi/src/BasicTest.scala
./test/checker-tests/fail2.scala
./test/checker-tests/fail6.scala
./test/checker-tests/fail12.scala
./test/checker-tests/fail7.scala
./test/checker-tests/fail10.scala
./test/checker-tests/fail1.scala
./test/scaladoc/resources/SI_5054_q1.scala
./test/scaladoc/resources/code-indent.scala
./test/scaladoc/resources/Trac4358.scala
./test/scaladoc/resources/SI_4421.scala
./test/scaladoc/resources/SI_5054_q5.scala
./test/scaladoc/resources/SI-6511.scala
./test/scaladoc/resources/SI-3314.scala
./test/scaladoc/resources/Trac4180.scala
./test/sca...
scala> ("find . -name '*.scala' ! -path ./tools/*  -type f") !!
res1: String = ""
```

Linux find

``` sh
alfe@rt:~/mkdev/bigdata/scala/scala$ find -name *.scala ! -path ./tools/*  -type f
find: paths must precede expression: ./tools/binary-repo-lib.sh
Usage: find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]
Goes on 1 depth.
alfe@rt:~/mkdev/bigdata/scala/scala$ find ! -path ./tools/ -type f -name *.scala
./test/files/jvm/value-class-boxing/test.scala
./test/files/jvm/opt_value_class/test.scala
./test/files/jvm/t7253/test.scala
./test/files/specialized/test.scala
./test/files/pos/t3486/test.scala
./test/files/pos/t1263/test.scala
./test/files/run/indylambda-boxing/test.scala
./test/files/run/t9200/test.scala
./test/files/run/t7106/test.scala
./test/files/run/t9349/test.scala
./test/files/run/t9489/test.scala
./test/files/scalacheck/testdir/test.scala
./test/files/scalacheck/test.scala
./test.scala

Next works fine, finds all .scala files
alfe@rt:~/mkdev/bigdata/scala/scala$ find ! -path ./tools/ -type f -name '*.scala' 
./test/script-tests/jar-manifest/src/jar-test.scala
./test/osgi/src/BasicReflection.scala
./test/osgi/src/ScalaOsgiHelper.scala
./test/osgi/src/ReflectionToolboxTest.scala
./test/osgi/src/BasicLibrary.scala
./test/osgi/src/BasicTest.scala
./test/checker-tests/fail2.scala
./test/checker-tests/fail6.scala
./test/checker-tests/fail12.scala
./test/checker-tests/fail7.scala
....(Very long output)
```

"This will prevent the find from failing when there is a stray .scala file at the root."
("find . -name _.scala ! -path ./tools/_  -type f") !! - It's ok for this case.

``` scala
scala> ("find . -name *.scala ! -path ./tools/*  -type f") !!
res4: String =
"./test/script-tests/jar-manifest/src/jar-test.scala
./test/osgi/src/BasicReflection.scala
./test/osgi/src/ScalaOsgiHelper.scala
.....
./project/ParserUtil.scala
./test.scala      - !!!
"
scala> 
```

janekdb:
Confirming the problem I anticipated does not exist.

```
scala> (("find . -name *.scala ! -path ./tools/*  -type f") !!).split("\n").size
res8: Int = 6574
```

```
$ > Test.scala
$ ll *.scala
Test.scala
```

```
scala> (("find . -name *.scala ! -path ./tools/*  -type f") !!).split("\n").size
res10: Int = 6574
```

Disregard my review comment.

tools/CheckDates.scala
FelAl: Extend App?

```
The argument of the scala command has to be a top-level object. If that object extends trait App, then all statements contained in that object will be executed; otherwise you have to add a method main which will act as the entry point of your program.
```

I choosed `main`, so now I don't see a point in the `Extend App`. Am I wrong?

janekdb: I guess I was surprised to see a `main` method after working with the more idiomatic `object C extends App` choice for so long. My preference would be to take advantage of the opportunity to declutter my class by dropping the need for `main`. What do you see as the benefit of `main` over `extends App`?

Ichoran: Technically, the mechanism of `extends App` is a kludge (it is implemented via `DelayedInit`, which has bugs in certain unusual cases that aren't applicable here).  I do not, therefore, really view it as "idiomatic" but rather "a convenience for beginners for now, that may eventually go away or be replaced by something else".  I always use the `main` method.  I think it's fine to leave it as it is.

tools/CheckDates.scala
FelAl: In the beginning I used
`git log --format=%s ./src/library/scala/runtime/ScalaNumberProxy.scala`
And split it via ("\n"), but I found that some commits contain `\n` in the subject, and `messages.count` may not be equal to `dates.count` for one commit in this way.

So, I decided to parse entire body, and it's more precise now.

janekdb:
@FelAl - you suggested,

> Maybe I need to split my PR in 3-4 smaller PRs?

I think this is a good idea because it would solve the problem of GitHub not being able to display the tool code in the "Files changed". This is a problem because I would like to respond to your comment about `partition` but the code is no longer available to view.

What do you think of this as the final set of PRs?

```
Add tool code - one commit
Add result of using tool - one commit
```

This has the benefit of ensuring each PR is doing one type of thing and will allow the tool code to be reviewed more easily.

janekdb:
Failure was,

```
[ERROR] Failed to execute goal on project typesafeDummy:
  Could not resolve dependencies for project com.typesafe:typesafeDummy:war:1.0-SNAPSHOT:
  Could not find artifact com.typesafe.sbt:incremental-compiler:jar:0.13.8-on-2.11.9-0038aee-SNAPSHOT-for-IDE-SNAPSHOT
  in pr-scala (https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/) -> [Help 1]
  org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project typesafeDummy:
  Could not resolve dependencies for project com.typesafe:typesafeDummy:war:1.0-SNAPSHOT:
  Could not find artifact com.typesafe.sbt:incremental-compiler:jar:0.13.8-on-2.11.9-0038aee-SNAPSHOT-for-IDE-SNAPSHOT
  in pr-scala (https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/)
```

FelAl:
Failed to execute goal on project typesafeDummy:

I don't know, how did I break this?

What do you think of this as the final set of PRs?

I think it's ok. But also, I want to integrate this into sbt build(CI)

adriaanm:
The IDE breakage is unrelated to this PR. I've opened a ticket to get this fixed: https://github.com/scala/scala-dev/issues/132

janekdb:
@FelAl - I've you are happy with this as the target set of PRs then I will watch out for when the tooling code is isolated and continue the review comments then.

```
Add tool code - one commit
Add result of using tool - one commit
```

FelAl:
Yes, It's ok!

adriaanm:
/nothingtoseehere -- The failure is due to scala/scala-dev#104 (and jenkins running out of inodes), not this PR.

PR #5090, opened 2016-04-08
SI-9740 Repl import fix -Yrepl-class-based

som-snytt:
Under `-Yrepl-class-based`, templating must follow the same scoping
as under traditional object-based. The new test shows a typical
case where two values of the same simple name must be imported in
different scopes.

JIRA: https://issues.scala-lang.org/browse/SI-9740


som-snytt:
There may be an ulterior motive for the existing behavior in spark?

som-snytt:
Sharper eyes than mine would have noticed that there is no space after the slashes in the javap 6 output:

```
//String Hello, delayed world.
```

I only checked against java 7 and 8.

src/repl/scala/tools/nsc/interpreter/package.scala
```diff
@@ -199,10 +199,11 @@ package object interpreter extends ReplConfig with ReplStrings {
   }
 
   /* debug assist
+  debug assist */
   private[nsc] implicit class `smart stringifier`(val sc: StringContext) extends AnyVal {
     import StringContext._, runtime.ScalaRunTime.stringOf
     def ss(args: Any*): String = sc.standardInterpolator(treatEscapes, args map stringOf)
-  } debug assist */
+  }
```
dragos, line 205: I'm not sure what's the meaning of these comments, but after uncommenting this class, maybe this empty comment should go away as well?

src/repl/scala/tools/nsc/interpreter/Imports.scala
```diff
@@ -127,7 +127,8 @@ trait Imports {
           case rh :: rest if !keepHandler(rh.handler) => select(rest, wanted)
           case rh :: rest                             =>
             import rh.handler._
-            val newWanted = wanted ++ referencedNames -- definedNames -- importedNames
+            val augment = rh match { case ReqAndHandler(_, _: ImportHandler) => referencedNames ; case _ => Nil }
```
dragos: This part is unclear to me... the intention here is to add referenced names _only_ from other imports? Say, we discovered we need to add `import foo.bar`, because `bar` is wanted`, but we also need to import`foo`, in case it comes from a previous request?

src/repl/scala/tools/nsc/interpreter/package.scala
som-snytt, line 205: Thx, after a couple of clean-up push -f, I was too tired when I saw that one. I figured to wait until a follow-up...

src/repl/scala/tools/nsc/interpreter/Imports.scala
som-snytt: Yes, that's the new test, I think. I couldn't think of a reason to do it for arbitrary defs. I was hoping a broken test would tell me.

dragos:
As far as I can tell this fixes the problem. I think there's another factor at play, the broken traverser that attempts to collect free names (it collects all `Idents` instead) when calculating `referenced`, but that's orthogonal. This LGTM, thank you @som-snytt!

dragos:
(pending a successful validation run, of course)

som-snytt:
Yes, the traverser causes too many imports, and relies on scoping to hide them; but extra wrapping is not free.  I was concerned that there's a reason spark needed it a certain way, so correctness doesn't help if it breaks spark shell.

dragos:
Any idea why the build can't find the compiler interface (source)?

```
sbt.ResolveException: download failed: org.scala-sbt#compiler-interface;0.13.8-dbuildx3a6b4130db221ca99a9a36399f2c80a80049b7e5!compiler-interface.jar(src)
```

lrytz:
ide validation on 2.11.x has be broken for a while, https://github.com/scala/scala-dev/issues/104

dragos:
@skyluc maybe you can ping @eed3si9n to fix this.

corruptmemory:
I built a version of Scala with this PR and the test suite passed for me.  I manually confirmed that the problem observed in SI-9740 seems addressed.

som-snytt:
Tweaked comments per @dragos

adriaanm:
/nothingtoseehere -- The failure is due to scala/scala-dev#104, not this PR.

dragos:
LGTM

lrytz:
this needs a rebase, unfortunately..

som-snytt:
NP-complete as in no problemo.

PR #5054, opened 2016-03-21
Update IntelliJ build for use with sbt

lrytz:
Re-submission of #5046 targeting 2.11.x. I will create a merge-2.11-to-2.12 follow-up PR.

PR #5091, opened 2016-04-08
Automate Eclipse settings and update project files:

dragos:
- remove `M2_REPO`. All dependencies are picked up from `build/deps`
- add script to update an existing workspace directory with the required path variables
- add the default Scala library to several projects for better out-of-the-box experience. This means
  that changes in the scale-library project may not be visible in the other projects, but makes it
  way easier to get a working config. If you really need that, you probably know what you’re doing
  anyway.

### DIFF
--- a/src/eclipse/README.md
+++ b/src/eclipse/README.md
@@ -7,21 +7,20 @@ The following points describe how to build Scala using Eclipse.
 
 0. Download the [Scala IDE bundle](http://scala-ide.org/download/sdk.html). It comes preconfigured for optimal performance.
 
-0. Run `ant init` to download some necessary jars.
+0. Run `ant build` to download some necessary jars and see a successful build.
 
-0. Import the project (in `src/eclipse`) via `File` → `Import Existing Projects` and navigate to `scala/src/eclipse`. Check all projects and click ok.
-
-0. You need to define a `path variable` inside Eclipse. Define `SCALA_BASEDIR` in
-`Preferences/General/Workspace/Linked Resources`. The value should be the absolute
-path to your Scala checkout. All paths in the project files are relative to this one,
-so nothing will work before you do so.
-
-  The same `SCALA_BASEDIR` variable needs to be defined as a `classpath variable` in
+0. You need to define a `path variable` and a `classpath variable` inside Eclipse, both pointing to the Scala checkout directory:
+  - (experimental): run `./update-workspace.sh scala_checkout_dir [workspace_dir]`. This should update your workspace settings
+ (restart Eclipse if it was running). For example:
+ ```
+ ./update-workspace.sh $HOME/git/scala ~/Documents/workspace-scalac
+ ```
+  - If the above didn't work, you can perform these steps manually: Define `SCALA_BASEDIR` in `Preferences/General/Workspace/Linked Resources`. The value should be the absolute
+path to your Scala checkout. All paths in the project files are relative to this one, so nothing will work before you do so.
+The same `SCALA_BASEDIR` variable needs to be defined **also** as a `classpath variable` in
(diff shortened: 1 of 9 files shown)


lrytz:
Unfortunately, the `ant` build and therefore `build/deps` is on its way out, so this will break soonish.

I recently added support for updating the IntelliJ module classpaths from our sbt build (https://github.com/scala/scala/pull/5054), I think this could be very easily extended to update the eclipse `.classpath` files. Would you mind taking a look?

dragos:
I will look at the Sbt build, but in the mean time I think this might help some people. /cc @sschaef @mlangc

lrytz:
sure, i don't mind merging it. i only took a quick look, didn't try it out. maybe @sschaef or @mlangc want to try it out and post an LGTM here? or we can just merge now if you prefer, that's fine too.

kiritsuku:
I didn't test the new script. Scaladoc is also broken for some reason. I get errors like `Couldn't create file for class scala/tools/ant/Scaladoc$lambda$$docSettings$1 null` but I'm not interested right now to figure out where that is coming from. partest-extras also can't be compiled because of `scala-partest_2.11-1.0.13.jar of partest-extras build path is cross-compiled with an incompatible version of Scala (2.11.0). In case this report is mistaken, this check can be disabled in the compiler preference page.` I don't know how to get partest for 2.11.8. The other more important project work fine. So a LGTM from me, since most projects compile without any changes. Not sure if that was the case previously. sbt-eclipse working for the Scala repo would be optimal of course.

src/eclipse/README.md
```diff
@@ -7,21 +7,20 @@ The following points describe how to build Scala using Eclipse.
...
-0. You need to define a `path variable` inside Eclipse. Define `SCALA_BASEDIR` in
-`Preferences/General/Workspace/Linked Resources`. The value should be the absolute
-path to your Scala checkout. All paths in the project files are relative to this one,
-so nothing will work before you do so.
-
-  The same `SCALA_BASEDIR` variable needs to be defined as a `classpath variable` in
+0. Run `ant build` to download some necessary jars and see a successful build.
+
+0. You need to define a `path variable` and a `classpath variable` inside Eclipse, both pointing to the Scala checkout directory:
+  - (experimental): run `./update-workspace.sh scala_checkout_dir [workspace_dir]`. This should update your workspace settings
```
mlangc, line 13: Hmm, this doesn't seem to work for me:

```
$ ./update-workspace.sh  /home/mlangc/Development/git/scala /home/mlangc/Development/eclipse-workspace/mars
Using metadata directory /home/mlangc/Development/eclipse-workspace/mars and Scala checkout /home/mlangc/Development/git/scala
Couldn't find /home/mlangc/Development/eclipse-workspace/mars/.plugins/org.eclipse.core.runtime/.settings/org.eclipse.core.resources.prefs. Is /home/mlangc/Development/eclipse-workspace/mars an Eclipse workspace?
```

I was more successful with

```
$ ./update-workspace.sh  /home/mlangc/Development/git/scala /home/mlangc/Development/eclipse-workspace/mars/.metadata
```

though (notice `.metadata` at the end of the last argument). Note that the same applies to `eclipse-luna`.

mlangc:
Except for the minor issue with the script above, everything seems to compile fine in the IDE. I also was able to start the `repl` from within Eclipse.

src/eclipse/README.md
lrytz, line 13: @dragos can you push that change?

dragos, line 13: OK, I'll double-check that thing.

dragos:
@sschaef scaladoc and partest-extra build fine for me.
@mlangc I fixed the path problem.

PR #5093, opened 2016-04-11
[Build] Include missing web assets in scaladoc

jodersky:
Running a Scala version that is built with sbt will fail when generating docs, since some required assets are not included in the packaged Scala artifacts. This PR adds the missing assets to the sbt build definition.

cc @SethTisue, the changes are related to the sbt build

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -507,7 +507,7 @@ lazy val scaladoc = configureAsSubproject(project)
     name := "scala-compiler-doc",
     description := "Scala Documentation Generator",
     libraryDependencies ++= Seq(scalaXmlDep, scalaParserCombinatorsDep, partestDep),
-    includeFilter in unmanagedResources in Compile := "*.html" | "*.css" | "*.gif" | "*.png" | "*.js" | "*.txt"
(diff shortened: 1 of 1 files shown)


lrytz:
could you re-submit this to 2.11.x? this way it will also make it into 2.12.x, we merge forward regularly.

szeiger:
LGTM

lrytz:
this needs a rebase to fix the merge conflict. @szeiger, is it that these files are new in 2.12 scaladoc, so it's not needed in 2.11?

szeiger:
@lrytz I ported the original filter expressions over from ant in 2.11, so I would expect them to be accurate there, but @jodersky should be able to confirm why there are needed in 2.12.

jodersky:
If you publish scala 2.12 with sbt, some assets are not included in the resulting jars and hence, scaladoc will fail.
@lrytz I never checked if the issue is also present in Scala 2.11 just running some tests now

jodersky:
Can't reproduce this on 2.11, I assume the new assets were introduced for scaladoc 2.12 (which looks pretty cool btw :))

PR #5096, opened 2016-04-12
Ensure ClassBTypes constructed from symbol and classfile are identical

lrytz:


sjrd:
I did not really read/understand the changes, but I'm guessing this could cause the opposite "problem" in Scala.js: that classes (symbols) read from classfiles do have the synthetic parents, but that symbols being compiled do _not_ have the synthetic parents at the time Scala.js runs.

The Scala.js IR does not have the restriction that super calls may only call directly inherited interfaces, so the Scala.js back-end would not add synthetic parents to fix the problem.

I'm not exactly sure whether what I'm saying makes any sense, but maybe it will trigger some thought.

lrytz:
> classes (symbols) read from classfiles do have the synthetic parents

Nothing has changed for symbols and frontend types. the additional interfaces only show up in `ClassBTypes`, which scala.js is probably not using at all.

we have two ways of filling up the details of a `ClassBType`: either we get the information (parents, flags, etc) from the frontend symbol or from the classfile. to ensure we have all the necessary information in the latter case, some metadata is saved in the classfile in the `InlineInfoAttribute`, for example what methods have the `@inline` annotation. The goal is basically to get a consistent `ClassBType` without requiring the corresponding symbol, because it's not always straightforward to get the correct symbol for a given class internal name (name mangling, synthetic classes, etc).

sjrd:
> Nothing has changed for symbols and frontend types. the additional interfaces only show up in ClassBTypes, which scala.js is probably not using at all.

OK then my concern is moot. Scala.js does not use `ClassBTypes`, indeed. Thanks for the explanation.

lrytz:
here's an observation while working on this PR: there's no good way to generate a valid tree for a static method defined in scala.

for comparison, the selection of a java-defined static method `A.f` is `Select(owner, "f")`, `owner.symbol` is the companion `ModuleSymbol` for `A`.

assume we have a static method in a scala class:
- if the owner is an ordinary `ClassSymbol` `c`, `Select(c.thisType, "m")` is not really valid as it's not a selection on the instance. a selection on the companion (if it exists) is also not valid because the static member is in the class, not the companion - for the backend, a member of a companion object would end up in the companion's `ModuleClass`.
- if the owner is a `ModuleClassSymbol` we can encode it as a selection on the `ModuleSymbol`, but that is also inconsistent: the member is static in the module class, and we don't need to load the module to access it.

in any case, the backend will identify the selection of a static method and not generate any bytecode for the `Select`'s qualifier.

these are not blockers, just a little weird. we should keep them in mind while adding more static methods (https://issues.scala-lang.org/browse/SI-9390)

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -675,80 +649,67 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
+              else InvokeStyle.Virtual
+
+            if (invokeStyle.hasInstance) genLoadQualifier(fun)
+            genLoadArguments(args, paramTKs(app))
+
+            val Select(qual, _) = fun // fun is a Select, also checked in genLoadQualifier
+            if (sym == definitions.Array_clone) {
+              // Special-case Array.clone, introduced in 36ef60e. The goal is to generate this
+              // call as "[I.clone" instead of "java/lang/Object.clone" (which would also work).
+              // Not sure why this is useful, maybe for performance. The commit message says
```
sjrd: I think that's because `Object.clone()` is protected. I guess trying to call it would be an illegal access on the JVM.

lrytz: I tested without the special case:

```
object Test {
  def foo(a: Array[Int]) = a.clone()
    println(foo(Array(1,2,3)).toList)
  }
}
```

gives

```
public final class Test$ {
  public foo([I)[I
    ALOAD 1
    INVOKEVIRTUAL java/lang/Object.clone ()Ljava/lang/Object;
    CHECKCAST [I
    ARETURN
}
```

and it runs fine. but i'd have to check the spec of protected again..

retronym:
>  there's no good way to generate a valid tree for a static method defined in scala.

I remember noticing the same problem once before when working in `Delambdafy`. I think I tried `Select(EmptyTree, "m")`, which I thought was more honest than the version useing `This`, but that broke something down the line, so I backed out of the change. We could try that again, perhaps things work now that `GenASM` is gone.

Another possibility would be be to use a null constant for the qualifier.

sjrd:
> if the owner is a ModuleClassSymbol we can encode it as a selection on the ModuleSymbol, but that is also inconsistent: the member is static in the module class, and we don't need to load the module to access it.

FYI, currently the Scala.js backend expects this. It does not do anything special to identify static methods. It compiles them as calls to methods on the companion object, just as the tree says.

That's what it's supposed to do, because Java statics end up being re implemented in Scala as methods on the companion object.

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
retronym: I found this part of the JVM spec hard to follow. `clone` is overriden as public in each of the `Array` classes (which also implement `Clonable`). Seems like its fine to refer the the protected method in `Object` in the invokevirtual in this case.

Toying around with this via reflection:

```
scala> java.lang.invoke.MethodHandles.publicLookup.findVirtual(classOf[Object], "clone", java.lang.invoke.MethodType.methodType(classOf[Object]))
java.lang.IllegalAccessException: member is protected: java.lang.Object.clone()Object/invokeVirtual, from java.lang.Object/public
  at java.lang.invoke.MemberName.makeAccessException(MemberName.java:852)

scala> def lookupOf(cls: Class[_]) = { import java.lang.invoke._; val constructor = classOf[MethodHandles.Lookup].getDeclaredConstructors.head; constructor.setAccessible(true); constructor.newInstance(cls).asInstanceOf[MethodHandles.Lookup] }
lookupOf: (cls: Class[_])java.lang.invoke.MethodHandles.Lookup

scala> lookupOf(classOf[Array[Int]]).findVirtual(classOf[Object], "clone", java.lang.invoke.MethodType.methodType(classOf[Object]))
res4: java.lang.invoke.MethodHandle = MethodHandle(int[])Object
```

Perhaps related:

https://bugs.openjdk.java.net/browse/JDK-4750641
https://bugs.openjdk.java.net/browse/JDK-4962591

retronym: Seems like OpenJDK has a special case in method resolution to allow this. Probably best not to rely on it.

http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/tip/src/share/vm/interpreter/linkResolver.cpp#l456

retronym:
@sjrd Here's a case where Scala 2.11 currently emits a static method within a class. This is done as a special case in lambdalift to avoid a verify error. Might be interesting to look at this from the ScalaJS perspective.

```
scala> class C(a: Any); class D extends C({ def foo = 1; foo})
defined class C
defined class D

scala> :javap -private -c D
Compiled from "<console>"
public class D extends C {
  private static final int foo$1();
    Code:
       0: iconst_1
       1: ireturn

  public D();
    Code:
       0: aload_0
       1: invokestatic  #11                 // Method foo$1:()I
       4: invokestatic  #17                 // Method scala/runtime/BoxesRunTime.boxToInteger:(I)Ljava/lang/Integer;
       7: invokespecial #20                 // Method C."<init>":(Ljava/lang/Object;)V
      10: return
}
```

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
retronym: A similar special case exists in `java.lang.invoke`: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/lang/invoke/MethodHandles.java#l1479

sjrd:
@retronym That's ... interesting ^^ I never saw that pop up. It turns out it works fine because the method is also _emitted_ as non-static, and that is not a problem for the Scala.js IR:

```
> helloworld/scalajsp helloworld/D.sjsir
class Lhelloworld_D extends Lhelloworld_C {
  def foo$1__p2__I(): int = {
    1
  }
  def init___() {
    this.Lhelloworld_C::init___O(this.foo$1__p2__I())
  }
}
```

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -69,12 +61,14 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
       tree match {
-        case Assign(lhs @ Select(_, _), rhs) =>
+        case Assign(lhs @ Select(qual, _), rhs) =>
           val isStatic = lhs.symbol.isStaticMember
           if (!isStatic) { genLoadQualifier(lhs) }
           genLoad(rhs, symInfoTK(lhs.symbol))
           lineNumber(tree)
-          fieldStore(lhs.symbol)
+          // receiverClass is used in the bytecode to access the field. using sym.owner may lead to IllegalAccessError, SI-4283
+          val receiverClass = qual.tpe.typeSymbol
```
retronym, line 70: This looks consistent with javac.

``` java
public class Test {
    private static class A {
        int i;
        static int j;
        void m() {};
        static void n() {};
    }
    static class B extends A {};

        B b = new B();
        C.test(b);
        assert(b.i == 1);
    }
}
class C {
    public static void test(Test.B b) {
        b.i = b.i + 1;
        b.j = b.j + 1;
        b.m();
        b.n();
    }
}

```

```
  public static void test(Test$B);
    Code:
       0: aload_0
       1: aload_0
       2: getfield      #2                  // Field Test$B.i:I
       5: iconst_1
       6: iadd
       7: putfield      #2                  // Field Test$B.i:I
      10: aload_0
      11: pop
      12: aload_0
      13: pop
      14: getstatic     #3                  // Field Test$B.j:I
      17: iconst_1
      18: iadd
      19: putstatic     #3                  // Field Test$B.j:I
      22: aload_0
      23: invokevirtual #4                  // Method Test$B.m:()V
      26: aload_0
      27: pop
      28: invokestatic  #5                  // Method Test$B.n:()V
      31: return
```

retronym, line 70: Scala doesn't allow static references via an instance or a subclass qualifier. I just noticed that this makes it impossible to express some static field/method selections.

``` scala
class D {
  def test(b: Test.B) {
    b.i = b.i + 1;

    b.j = b.j + 1; // j not a member
    Test.B.j = Test.B.j + 1; // j not a member
    Test.A.j = Test.A.j + 1; // A is not accessible

    b.m();
    b.n(); // n is not a member
    Test.B.n(); // n is not a member
    Test.A.n(); // A is not accessible
  }
}
```

retronym, line 70: For static access, the `qual.tpe.typeSymbol` will be the (fictional) module class representing the static perspective of the Java class. Looks like this is correctly handled by `def internalName`, so this looks correct.

retronym:
We may need a related change in `erasure` related to the [casts inserted to implement refinements](https://github.com/scala/scala/blob/96f230a0e9c254c45dc91c71b5929639e6add1f0/src/compiler/scala/tools/nsc/transform/Erasure.scala#L663-L665) (is it just refinements that require this treatment?).

``` java
public class Test {
    private static class A {
        int i;
    }
    static class B extends A {
    };

```

``` scala
class D {
  def test(b: Other with Test.B) {
    b.i = b.i + 1;
  }
}

class Other
```

``` scala
// scalac-2.11.8 -Xprint:posterasure  sandbox/test.scala
[[syntax trees at end of               posterasure]] // test.scala
package <empty> {
  class D extends Object {
    def test(b: Other): Unit = b.$asInstanceOf[Test.A]().i = b.$asInstanceOf[Test.A]().i.+(1)
  };
}
```

Here, we have cast the qualifier to the owner of the field.

I expect this would be pretty hard to fix, though, and it is something of a corner case so we might just have to leave it.

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
lrytz: Checked again the JVM spec

[Access Control](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.4): protected method R, declared in C, is accessible to a class or interface D if
- D is a subclass of C, and
- R must contain a symbolic reference to a class T, such that T is either a subclass of D, a superclass of D, or D itself

"symbolic reference to class T" means the receiver class in the invocation instruction.

But there's more: "This discussion omits a related restriction [...] is checked as part of the verification process (§4.10.1.8); it is not part of link-time access control."

[4.10.1.8. Type Checking for protected Members](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.10.1.8) encodes that a protected method C.m can be accessed in a subclass D only if the receiver object's type conforms to D. This corresponds to the [Java spec 6.6.2.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.6.2.1).

Assume we have

``` scala
class C { override def clone(): Object = "hi" }
```

We cannot emit `def f(c: C) = c.clone()` as `Object.clone()`. he JVM would perform the following checks:
- [Method Resolution (5.4.3.3)](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.3) finds `Object.clone`
- [Access Control (5.4.4)](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.4) is checked at the end of resolution and succeeds (every class and interface is a subclass of `Object`, so `Object.clone` can always passes)
- [Verification (4.10.1.8)](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.10.1.8) will fail the `passesProtectedCheck`

I tested this with a compiler hack, we get `java.lang.VerifyError` indeed.

So the reason `Object.clone` works for arrays is really the special cases you point to above. Given that javac also emits `[I.clone` we better stick to that.

Finally, to make the discussion complete, here's the example with traits:

```
trait T { override def clone(): Object = "hu" }
trait U extends T
class K extends U
object T1 { def f1(u: U) = u.clone() }
```

We have to emit `u.clone` as `T.clone`. Using `U.clone` gives `NoSuchMethodError: U.clone`. Looking at [5.4.3.4. Interface Method Resolution](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.4) I actually don't see why this is the case, it might be a bug in the JVM.

lrytz: To stress that point ("it might be a bug in the JVM"), I can get javac to produce bytecode that crashes with a `NoSuchMethodError`:

``` java
public class A {
  public interface I1 {
    default Object clone() { return "hi"; }
  }

  public interface I2 extends I1 { }

  public static class K implements I2 {
    public Object clone() { return "Ka"; }
  }

  public static void f(I2 i) { System.out.println(i.clone()); }

    f(new K());
  }
}
```

```
$ javac A.java
$ java A
Exception in thread "main" java.lang.NoSuchMethodError: A$I2.clone()Ljava/lang/Object;
    at A.f(A.java:12)
    at A.main(A.java:15)
```

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
lrytz, line 70: > Scala doesn't allow static references via an instance or a subclass qualifier

I didn't know that was possible in Java..!

lrytz:
Good observation..! I managed to turn this into a test:

```
$ cat B.java
package p1;

interface A {
  int i();
}

public interface B extends A { }

$ cat Test.scala
package p2 {
  class K
}

object Test {
  def t(x: p2.K with p1.B) = x.i()
    println(t(new p2.K with p1.B { def i() = 1 }))
  }
}

$ rm -rf p1 p2 *.class
$ javac B.java -d .
$ qsc Test.scala
$ qs Test
java.lang.IllegalAccessError: tried to access class p1.A from class Test$
    at Test$.t(Test.scala:6)
```

lrytz:
Echoing the earlier discussion about `[I.clone` vs `Object.clone`, because it's hidden now after a force-push:

Jason:

> Seems like OpenJDK has a special case in method resolution to allow this. Probably best not to rely on it.
> http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/tip/src/share/vm/interpreter/linkResolver.cpp#l439

The special case basically considers `Object.clone` as public if the dynamic receiver is an array.

Lukas:

> Checked again the JVM spec
> 
> [Access Control](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.4): protected method R, declared in C, is accessible > to a class or interface D if
> - D is a subclass of C, and
> - R must contain a symbolic reference to a class T, such that T is either a subclass of D, a superclass of D, or D itself
> 
> "symbolic reference to class T" means the receiver class in the invocation instruction.
> 
> But there's more: "This discussion omits a related restriction [...] is checked as part of the verification process (§4.10.1.8); it is not > part of link-time access control."
> 
> [4.10.1.8. Type Checking for protected Members](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.10.1.8) encodes that a > protected method C.m can be accessed in a subclass D only if the receiver object's type conforms to D. This corresponds to the [Java spec > 6.6.2.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.6.2.1).
> 
> Assume we have
> 
> ``` scala
> class C { override def clone(): Object = "hi" }
> ```
> 
> We cannot emit `def f(c: C) = c.clone()` as `Object.clone()`. he JVM would perform the following checks:
> - [Method Resolution (5.4.3.3)](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.3) finds `Object.clone`
> - [Access Control (5.4.4)](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.4) is checked at the end of resolution > and succeeds (every class and interface is a subclass of `Object`, so `Object.clone` can always passes)
> - [Verification (4.10.1.8)](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.10.1.8) will fail the > `passesProtectedCheck`
> 
> I tested this with a compiler hack, we get `java.lang.VerifyError` indeed.
> 
> So the reason `Object.clone` works for arrays is really the special cases you point to above. Given that javac also emits `[I.clone` we > better stick to that.
> 
> Finally, to make the discussion complete, here's the example with traits:
> 
> ```
> trait T { override def clone(): Object = "hu" }
> trait U extends T
> class K extends U
> object T1 { def f1(u: U) = u.clone() }
> ```
> 
> We have to emit `u.clone` as `T.clone`. Using `U.clone` gives `NoSuchMethodError: U.clone`. Looking at [5.4.3.4. Interface Method Resolution](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.4) I actually don't see why this is the case, it might be a bug in the JVM.

Lukas:

> To stress that point ("it might be a bug in the JVM"), I can get javac to produce bytecode that crashes with a `NoSuchMethodError`:
> 
> ``` java
> public class A {
>   public interface I1 {
>     default Object clone() { return "hi"; }
>   }
> 
>   public interface I2 extends I1 { }
> 
>   public static class K implements I2 {
>     public Object clone() { return "Ka"; }
>   }
> 
>   public static void f(I2 i) { System.out.println(i.clone()); }
> 
>   public static void main(String[] args) {
>     f(new K());
>   }
> }
> ```
> 
> ```
> $ javac A.java
> $ java A
> Exception in thread "main" java.lang.NoSuchMethodError: A$I2.clone()Ljava/lang/Object;
>   at A.f(A.java:12)
>   at A.main(A.java:15)
> ```

lrytz:
@retronym I updated the comments [here](https://github.com/scala/scala/pull/5096/commits/ad2138397778fd718579bc9b2f40f982571057eb#diff-2e0fd5042aa83407a80a7557cc2a666bR668) and [here](https://github.com/scala/scala/pull/5096/commits/ad2138397778fd718579bc9b2f40f982571057eb#diff-2e0fd5042aa83407a80a7557cc2a666bR1036) and filed https://issues.scala-lang.org/browse/SI-9761, so this PR should be ready now.

retronym:
LGTM

PR #5097, opened 2016-04-12
SI-9749 REPL strip lead ws on dot continuation

som-snytt:
Permit leading whitespace before `.` for continued selection.

This is just to handle pastes, which will typically include
indented text, and not to make dot-continuation especially robust.


lrytz:
LGTM!

PR #5099, opened 2016-04-14
SI-9390 Emit local defs that don't capture this as static

retronym:
This avoids unnecessary memory retention, and allows lambdas
that call the local methods to be serializable, regardless of
whether or not the enclosing class is serializable.

The second point is especially pressing, given that the enclosing
class for local methods defined in a used to be the (serializable)
anonymous function class, but as of Scala 2.12 will be the enclosing
class of the lambda.

This change is similar in spirit to SI-9408 / 93bee55e.


retronym:
See https://gist.github.com/JoshRosen/8aacdee0162da430868e7f73247d45d8 for an analysis of this in the context of Spark. /cc @joshrosen

retronym:
Review, additional tests by @adriaanm | @lrytz, please.

retronym:
Bytecode diff from 2.12.0-M4 for a simple example:

``` diff
 scala> :paste -raw

 class StaticLift { def foo = { def bar(x: Any) = true; bar _ } }

 scala> :javap -c -private StaticLift
 Compiled from "<pastie>"
 public class StaticLift {
   public scala.Function1<java.lang.Object, java.lang.Object> foo();
     Code:
-       0: aload_0
-       1: invokedynamic #37,  0             // InvokeDynamic #0:apply:(LStaticLift;)Lscala/Function1;
-       6: areturn
+       0: invokedynamic #36,  0             // InvokeDynamic #0:apply:()Lscala/Function1;
+       5: areturn

-  private final boolean bar$1(java.lang.Object);
+  private static final boolean bar$1(java.lang.Object);
     Code:
        0: iconst_1
        1: ireturn

-  public final boolean StaticLift$$$anonfun$1(java.lang.Object);
+  public static final boolean StaticLift$$$anonfun$1(java.lang.Object);
     Code:
        0: aload_0
-       1: aload_1
-       2: invokespecial #46                 // Method bar$1:(Ljava/lang/Object;)Z
-       5: ireturn
+       1: invokestatic  #45                 // Method bar$1:(Ljava/lang/Object;)Z
+       4: ireturn

   public StaticLift();
     Code:
        0: aload_0
-       1: invokespecial #50                 // Method java/lang/Object."<init>":()V
+       1: invokespecial #49                 // Method java/lang/Object."<init>":()V
        4: return

-  public final java.lang.Object StaticLift$$$anonfun$1$adapted(java.lang.Object);
+  public static final java.lang.Object StaticLift$$$anonfun$1$adapted(java.lang.Object);
     Code:
        0: aload_0
-       1: aload_1
-       2: invokevirtual #52                 // Method StaticLift$$$anonfun$1:(Ljava/lang/Object;)Z
-       5: invokestatic  #58                 // Method scala/runtime/BoxesRunTime.boxToBoolean:(Z)Ljava/lang/Boolean;
-       8: areturn
+       1: invokestatic  #51                 // Method StaticLift$$$anonfun$1:(Ljava/lang/Object;)Z
+       4: invokestatic  #57                 // Method scala/runtime/BoxesRunTime.boxToBoolean:(Z)Ljava/lang/Boolean;
+       7: areturn

   private static java.lang.Object $deserializeLambda$(java.lang.invoke.SerializedLambda);
     Code:
        0: aload_0
-       1: invokedynamic #70,  0             // InvokeDynamic #1:lambdaDeserialize:(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object;
+       1: invokedynamic #69,  0             // InvokeDynamic #1:lambdaDeserialize:(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object;
        6: areturn
-}
+}
\ No newline at end of file

```

test/junit/scala/tools/nsc/backend/jvm/opt/InlineWarningTest.scala
```diff
@@ -141,7 +141,7 @@ class InlineWarningTest extends ClearAfterClass {
 
     val warn =
       """M::f()I is annotated @inline but could not be inlined:
-        |The callee M::f()I contains the instruction INVOKESPECIAL M.nested$1 ()I
+        |The callee M::f()I contains the instruction INVOKESTATIC M.nested$1 ()I
```
lrytz, line 116: a nice thing about static methods is that we can in principle make them public for the purpose of the optimizer and don't need to name-mangle

src/compiler/scala/tools/nsc/transform/Delambdafy.scala
```diff
@@ -336,6 +352,9 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
         // They'll be of the form {(args...) => this.anonfun(args...)}
         // but we do need to make note of the lifted body method in case it refers to 'this'
         if (currentMethod.exists) liftedMethodReferences(currentMethod) += targetMethod(fun)
+      case Apply(sel @ Select(This(_), _), args) if currentMethod.exists && sel.symbol.isLiftedMethod =>
+        if (currentMethod.exists) liftedMethodReferences(currentMethod) += sel.symbol
```
lrytz, line 374: this condition looks redundant

JoshRosen:
Good news: I tested out this PR as of https://github.com/scala/scala/pull/5099/commits/e6a566e0b7a1c85c796bda44c809c6ea38cffd57 and it appears to have fixed the serializability issues with local `def`s that I found in Spark.

Bad news: in more local testing, I discovered a related problem affecting local classes. For example:

``` scala
class MyClass {
  Seq().map { x: Any => new Object {} }
}
```

In this example, the `map`'s closure will become an instance method of `MyClass` (rather than a static method) because the anonymous class created by `new Object {}` takes an instance of `MyClass` in its constructor.

This sounds like [SI-1419](https://issues.scala-lang.org/browse/SI-1419); would it be possible to target that issue for 2.12?

retronym:
@JoshRosen That's a tricky one. As of [SI-9408](https://issues.scala-lang.org/browse/SI-9408) / 93bee55, the constructor of the anonymous class no longer stores retains a reference to the enclosing class, but its constructor does still include this:

``` scala
class MyClass {
  def test = {
    { (x: Any) => new Object {} }
  }
}
```

``` scala
  final class anon$1 extends Object {
    def <init>($outer: MyClass): <$anon: Object> = {
      anon$1.super.<init>();
      // $outer is not stored in a field
      ()
    }
  }
```

The transform responsible for this keeps the constructor signature intact because in general, another compilation unit might be calling the constructor of an inner class, and we need to have an agreed constructor signature given that client and class might be separately compiled.

However, for local classes like in your example, we could in fact totally remove `$outer` from the constructor signature (we'd need to rewrite calls to the constructor in the same method, of course). This won't be trivial to implement, but might be worthwhile.

retronym:
> However, for local classes like in your example, we could in fact totally remove $outer from the constructor signature (we'd need to rewrite calls to the constructor in the same method, of course). This won't be trivial to implement, but might be worthwhile.

Here's a first cut at this change: https://github.com/retronym/scala/compare/ticket/9390...retronym:ticket/9390-2

lrytz:
LGTM!

retronym:
Let's run this one by @adriaanm next week before merging.

adriaanm:
LGTM -- should we look into unmangling names of members that end up being static, to keep them shorter? (If I understood Lukas's comment correctly, the name mangling we're talking about emulates static binding, i.e., these members should not be late bound/overridable, but they need not be private.)

adriaanm:
Sorry about the delay here -- unfortunately this will need a rebase.

retronym:
@lrytz @adriaanm I've rebased this on top of another open PR that also touches `Delambdafy`.

Additionally, I have added support for avoiding capturing the enclosing class for lambdas that only need to instantiate local classes elided outer fields.

retronym:
Looks like that null is getting used somewhere.

```
[quick.partest-extras] Exception in thread "main" java.lang.NullPointerException
[quick.partest-extras]  at scala.collection.LinearSeqLike$$anon$1.hasNext(LinearSeqLike.scala:43)
[quick.partest-extras]  at scala.collection.TraversableOnce.collectFirst(TraversableOnce.scala:144)
[quick.partest-extras]  at scala.tools.util.PathResolver$Environment$.searchForBootClasspath(PathRes
```

lrytz:
```
trait LinearSeqLike .. { self =>
  def iterator: Iterator = new AbstractIterator {
    var these = self
    def hasNext = !these.isEmpty
  }
}
```

The anonymous class constructor has a parameter `$outer: LinearSeqLike` which is only used in the constructor (stored in field `these`), so no field for `$outer` is needed; but its value is still required..

retronym:
@lrytz Thanks for investigating. Hopefully it will be as easy as adding a check to `Constructors` that the outer param is not referenced in any of the constructors, other than to forward to the primary constructor.

adriaanm:
Adding the WIP label since scabot will insist on adding "reviewed" based on earlier comments.

retronym:
Still trying:

```
[quick.library] Exception in thread "main" java.lang.NullPointerException
[quick.library]     at scala.tools.nsc.transform.Constructors$TemplateTransformer.scala$tools$nsc$transform$Constructors$TemplateTransformer$$$anonfun$34(Constructors.scala:718)
[quick.library]     at 
```

retronym:
@adriaanm @lrytz The last commit here needs review now.

lrytz:
@adriaanm fyi, the first three commits here are the same as in https://github.com/scala/scala/pull/5157

lrytz:
the last commit message seems incomplete:

```
However, the status quo means that a lambda that
This in turn makes lambdas that refer to such classes serializable
even when the outer class is not.
```

I'll updating this to explicitly pass the implicit phrase "itself serializable" :)

src/compiler/scala/tools/nsc/transform/Delambdafy.scala
```diff
@@ -368,6 +371,9 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
       case Apply(sel @ Select(This(_), _), args) if sel.symbol.isLiftedMethod =>
         if (currentMethod.exists) liftedMethodReferences(currentMethod) += sel.symbol
         super.traverseTrees(args)
+      case Apply(fun, outer :: rest) if fun.symbol.isConstructor && treeInfo.isQualifierSafeToElide(outer) && fun.symbol.hasAttachment[OuterArgCanBeElided.type] =>
```
lrytz: maybe extract this condition into a local method, as it's used in two places

lrytz:
looks good otherwise!

adriaanm:
ping @retronym -- lets apply a quick round of polish based on Lukas's feedback and get this merged today

lrytz:
note that merging this will also merge the commits it shares with #5157

adriaanm:
Ok, I'm going to let you coordinate these two then.

src/compiler/scala/tools/nsc/transform/Delambdafy.scala
retronym: done

lrytz:
LGTM

PR #5100, opened 2016-04-15
SI-6710 / PR 5072 follow-up: fix Unit.box / Unit.unbox

lrytz:
The backend replaces .box / .unbox methods by corresponding invocations
to BoxesRunTime, but not for Unit.

~~This commit extends the backend to also intercept box/unbox on Unit.~~

### DIFF
--- a/test/junit/scala/BoxUnboxTest.scala
+++ b/test/junit/scala/BoxUnboxTest.scala
@@ -1,0 +1,119 @@
+package scala
+
+import org.junit.Test
+import org.junit.Assert._
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+import scala.tools.testing.AssertUtil._
+
+@RunWith(classOf[JUnit4])
+class BoxUnboxTest {
+  def genericNull[T] = null.asInstanceOf[T] // allowed, see SI-4437, point 2
+
+  @Test
+  def boxUnboxInt(): Unit = {
+    val b = new Integer(1)
+    val u = 1
+
+    assertEquals(1.toInt, u)
+
+    assertEquals(Predef.int2Integer(1), b)
+    assertEquals(1: Integer, b)
+    assertEquals(Int.box(1), b)
+    assertEquals(1.asInstanceOf[Object], b)
+
+    assertThrows[ClassCastException]("".asInstanceOf[Integer])
+
+    assertEquals(Predef.Integer2int(b), u)
+    assertEquals(b: Int, u)
+    assertEquals(Int.unbox(b), u)
+    assertEquals(b.asInstanceOf[Int], u)
+    assertEquals(b.intValue, u)
+    assertEquals(b.toInt, u)
+    intWrapper(b).toInt
+
+    assertThrows[ClassCastException](Int.unbox(""))
+    assertThrows[ClassCastException]("".asInstanceOf[Int])
+
+    // null unboxing in various positions
+
+    val n1 = Int.unbox(null)
+    assertEquals(n1, 0)
+    val n2 = Predef.Integer2int(null)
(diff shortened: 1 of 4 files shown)


sjrd:
Hum ... why complicate the back-end when the user-space implementation of these methods was perfectly fine?

Adding code in the compiler to avoid code in the library seems like entirely the wrong trade-off to me.

lrytz:
This PR only fixes the regression of #5072.

There's a lot more that could be done around this area: if you play a bit with `null.asInstanceOf[X]` you'll quickly find surprises. I tried to collect the known ones into a test and make sure that at least we don't accidentally change anything.

Yet another question is whether we actually need to intercept the box/unbox methods and re-wire to the box/unbox implementations in `BoxesRunTime`. We could just put the implementations in the primitive classes companion objects. The only disadvantage I see is that instead of a static call we'd have a module load and an instance call.

lrytz:
> Hum ... why complicate the back-end when the user-space implementation of these methods was perfectly fine?

yeah, it's just one way to do it, and it was more consistent. i had the same thought of course, see my other comment: i think we could just get rid of hijacking box/unbox altogether.

lrytz:
it seems the logic was initially added to support the .net backend: https://github.com/scala/scala/commit/8e56e0e

sjrd:
Or erasure could just generate calls to `BoxesRunTime` directly, rather than the `box/unbox` methods. We implement `box/unbox` in user-space calling `BoxesRunTime`, and we deprecate those.

Another advantage of keeping the actual implem in `BoxesRunTime` rather than `Int` and co., is that `BoxesRunTime` is already platform-dependent, and overridden in Scala.js.

lrytz:
I changed the patch: now it restores the implementations of `Unit.box` / `Unit.unbox`, which are not re-wired to BoxesRunTime to in the backend.

Note that the implementation of `Unit.unbox` is changed:

```
  def unbox(x: java.lang.Object): Unit = x.asInstanceOf[scala.runtime.BoxedUnit]
```

Initially the method was just returning `()`. This is however inconsistent, all other `unbox` methods cast to the box type. Also, the Scaladoc was always saying "it accepts any Object, but will throw an exception if the argument is not a scala.runtime.BoxedUnit".

Longer-term, I think we should remove the re-wiring to BoxesRunTime and just put the implementations in the primitive companions. However, I think we need to implement the `@static` support first, otherwise every box/unbox operation involves a module load.

I think we should not deprecate `Int.box`, as that is the recommended (and most obvious) API for users.

sjrd:
LGTM

test/junit/scala/BoxUnboxTest.scala
```diff
@@ -0,0 +1,119 @@
...
+    val n1 = null.asInstanceOf[Unit]              // SI-9066: should be UNIT, but currently null
+    assertThrows[AssertionError](assert(n1 == b)) // should not throw
+
+    val n2 = null.asInstanceOf[Unit] == b    // SI-9066: should be true, but currently false
+    assertThrows[AssertionError](assert(n2)) // should not throw
+
+    def f(a: Any) = "" + a
+    val n3 = f(null.asInstanceOf[Unit])                  // "null", should be "()". probably same cause as SI-602.
+    assertThrows[AssertionError](assertEquals(n3, "()")) // should not throw
+  }
```
retronym, line 118: Sorry I didn't catch this before hand, but shouldn't this be a `partest` test, or a `JUnit` test which includes the code above as a `String`? This is a test of the compiler, not the library.

lrytz, line 118: right, i the entire test should be converted, i'll do that. note to slef: also do it for the inliner test (array.foreach)

PR #5101, opened 2016-04-15
Simplify TraversableLike.isEmpty?

jxtps:
It looks like several methods (isEmpty, forall, exists, find, head, and copyToArray) in TraversableLike could be greatly simplified by using a `for (x <- this) return Y` style, instead of the `breakable { for(x <- this) { ... break}}` style? 

If there are performance implications of using breakable, one could consider also modifying sliceInternal and takeWhile (though that would require calling b.result from two different points in the method). 

@Ichoran @axel22 (listed as handling collections on https://github.com/scala/scala )

### DIFF
--- a/src/library/scala/collection/TraversableLike.scala
+++ b/src/library/scala/collection/TraversableLike.scala
@@ -128,14 +128,8 @@ trait TraversableLike[+A, +Repr] extends Any
    *  @return    `true` if the $coll contain no elements, `false` otherwise.
    */
   def isEmpty: Boolean = {
-    var result = true
-    breakable {
-      for (x <- this) {
-        result = false
-        break
-      }
-    }
-    result
+    for (x <- this) return false
+    true
   }
 
   def hasDefiniteSize = true


jxtps:
(Happy to provide a full PR with all the methods edited, but I figured I'd show an example and check for any interest before doing that)

Ichoran:
I have benchmarked it in the past and the payload-free version (breakable/break) is in faster in my testing in this case.  I don't think I tested sliceInternal.  I tested takeWhile on something, but I don't think it was TraversableOnce.

Alefas:
I think for performance reasons 'while' would be much faster than 'for' or 'breakable'.

axel22:
@Alefas  Agreed, `while` with a `return` should result in optimal code. However, we need to be more generic in `TraversableLike`, since all we can use to implement a collection operation is `foreach`.

jxtps:
```
package com.example

import scala.collection._
import generic._
import mutable.Builder
import scala.annotation.{bridge, migration, tailrec}
import scala.annotation.unchecked.{uncheckedVariance => uV}
import parallel.ParIterable
import scala.util.control.Breaks

object Hello {

  val seq = Seq("Hello", "World", "!")

  val N = 100 * 1000

  def printTime(t0: Long, t1: Long, msg: String): Unit = {
    println("%,9d ns %s".format(t1 - t0, msg))
  }

  val breaks: Breaks = new Breaks
  def isEmptyExisting[A](seq: Seq[A]): Boolean = {
    import breaks._
    var result = true
    breakable {
      for (x <- seq) {
        result = false
        break
      }
    }
    result  }

  def testExisting(): Unit = {
    val t0 = System.nanoTime()
    var i = 0
    while (i < N) {
      isEmptyExisting(seq)
      i += 1
    }
    printTime(t0, System.nanoTime(), "isEmptyExisting")
  }

  def isEmptyNew[A](seq: Seq[A]): Boolean = {
    for (s <- seq) return false
    true
  }

  def testNew(): Unit = {
    val t0 = System.nanoTime()
    var i = 0
    while (i < N) {
      isEmptyNew(seq)
      i += 1
    }
    printTime(t0, System.nanoTime(), "isEmptyNew")
  }

    for (i <- 0 until 10) {
      println("\n---- %d ----".format(i))
      testExisting()
      testNew()
    }
  }
}
```

This produces the output: 

```
[info] Running com.example.Hello

---- 0 ----
18,438,779 ns isEmptyExisting
10,542,083 ns isEmptyNew

---- 1 ----
5,903,378 ns isEmptyExisting
1,860,693 ns isEmptyNew

---- 2 ----
  628,388 ns isEmptyExisting
1,579,655 ns isEmptyNew

---- 3 ----
  606,284 ns isEmptyExisting
1,425,320 ns isEmptyNew

---- 4 ----
  582,207 ns isEmptyExisting
1,399,269 ns isEmptyNew

---- 5 ----
  663,913 ns isEmptyExisting
1,481,371 ns isEmptyNew

---- 6 ----
  637,467 ns isEmptyExisting
1,453,345 ns isEmptyNew

---- 7 ----
  601,547 ns isEmptyExisting
1,484,529 ns isEmptyNew

---- 8 ----
  674,175 ns isEmptyExisting
1,447,030 ns isEmptyNew

---- 9 ----
  656,413 ns isEmptyExisting
1,462,819 ns isEmptyNew
[success] Total time: 3 s, completed Apr 25, 2016 1:56:28 PM
```

That microbenchmark may or may not be any good, but if anything it indicates that the current code is >2x faster than the visually simpler code.

axel22:
Could you retry the benchmark so that each version is run in a separate VM, and increase `N` to `10 * 1000 * 1000`?

jxtps:
```
[info] Running com.example.Hello
138,000,723 ns isEmptyExisting
97,158,240 ns isEmptyExisting
96,335,256 ns isEmptyExisting
69,096,672 ns isEmptyExisting
55,118,977 ns isEmptyExisting
[success] Total time: 3 s, completed Apr 25, 2016 2:26:49 PM
```

vs

```
[info] Running com.example.Hello
202,476,055 ns isEmptyNew
170,826,517 ns isEmptyNew
116,564,468 ns isEmptyNew
183,195,742 ns isEmptyNew
116,697,092 ns isEmptyNew
[success] Total time: 4 s, completed Apr 25, 2016 2:27:23 PM
```

axel22:
Thanks. New implementation indeed seems slower. You might additionally try increasing the number of repetitions, but it seems unlikely that there would be any further changes.

Ichoran:
This agrees with my previous observations.

jxtps:
It does indeed seem slower. 

I don't know how `return` is implemented in Scala closures, but I can imagine it requiring throwing an exception (or equivalent), and the exception having to be re-created from scratch on each call. 

Using Break enables the existing code to create the exception only once.

axel22:
Possible - after a quick glance, can't see if this class has any singletons:

https://github.com/scala/scala/blob/2.12.x/src/library/scala/runtime/NonLocalReturnControl.scala

Would be a good idea to run `javap` on that benchmark and inspect the bytecode.

jxtps:
```
  public <A> boolean isEmptyExisting(scala.collection.Seq<A>);
    Code:
       0: iconst_1
       1: invokestatic  #77                 // Method scala/runtime/BooleanRef.create:(Z)Lscala/runtime/BooleanRef;
       4: astore_2
       5: aload_0
       6: invokevirtual #79                 // Method breaks:()Lscala/util/control/Breaks;
       9: new           #81                 // class com/example/Hello$$anonfun$isEmptyExisting$1
      12: dup
      13: aload_1
      14: aload_2
      15: invokespecial #84                 // Method com/example/Hello$$anonfun$isEmptyExisting$1."<init>":(Lscala/collection/Seq;Lscala/runtime/BooleanRef;)V
      18: invokevirtual #90                 // Method scala/util/control/Breaks.breakable:(Lscala/Function0;)V
      21: aload_2
      22: getfield      #94                 // Field scala/runtime/BooleanRef.elem:Z
      25: ireturn
```

vs

```
  public <A> boolean isEmptyNew(scala.collection.Seq<A>);
    Code:
       0: new           #4                  // class java/lang/Object
       3: dup
       4: invokespecial #133                // Method java/lang/Object."<init>":()V
       7: astore_2
       8: aload_1
       9: new           #135                // class com/example/Hello$$anonfun$isEmptyNew$1
      12: dup
      13: aload_2
      14: invokespecial #137                // Method com/example/Hello$$anonfun$isEmptyNew$1."<init>":(Ljava/lang/Object;)V
      17: invokeinterface #143,  2          // InterfaceMethod scala/collection/Seq.foreach:(Lscala/Function1;)V
      22: iconst_1
      23: goto          39
      26: astore_3
      27: aload_3
      28: invokevirtual #147                // Method scala/runtime/NonLocalReturnControl.key:()Ljava/lang/Object;
      31: aload_2
      32: if_acmpne     40
      35: aload_3
      36: invokevirtual #151                // Method scala/runtime/NonLocalReturnControl.value$mcZ$sp:()Z
      39: ireturn
      40: aload_3
      41: athrow
    Exception table:
       from    to  target type
           8    26    26   Class scala/runtime/NonLocalReturnControl
```

axel22:
Looks like, in addition to instantiating the exception in the `Hello$$anonfun$isEmptyNew$1` lambda, emulating `return` with an exception requires instantiating an additional dummy `Object` that does not get escape-analyzed.

jxtps:
Is there an optimization opportunity in the byte code emitted by scalac?

To an outsider, it does seem counterintuitive that 

```
  def isEmpty: Boolean = {
    for (x <- this) return false
    true
  }
```

would lead to slower and more verbose byte code than 

```
  def isEmpty: Boolean = {
    var result = true
    breakable {
      for (x <- this) {
        result = false
        break
      } 
    }
    result
  }
```

retronym:
@jxtps That's in interesting question. I created a synthetic benchmark to try to understand this better: https://github.com/retronym/scala-jmh-suite/commit/102bbc3c7c4b26aa2284a07575ee6e097c85fcee

Results and analysis in the commit comment. `break` has a slight edge there, but not by much.

I think the difference is non-local return's allocate key for each method that contains them to be sure that you are catching one thrown from that invocation of method. With more complicated code, this is important to preserve semantics. But it isn't needed here, as we don't execute any unknown callbacks that might themselves be using exceptions as control flow that we might accidentally catch.

The use of `break` within collections is made safe by using `Traversable.breaks`, a collections-private instance of the `Breaks`. If it didn't do this, code like:

```
scala> val trav = new Traversable[String] { def foreach[U](a: String => U): Unit = a("")}
trav: Traversable[String] = ()

scala> trav.forall(_ => scala.util.control.Breaks.break)
```

would give the wrong answer (`false`, rather than throwing the `NonLocalReturnControl`.

The usual caveats about trusting benchmarks apply! I've been trying to improve my abilities in writing and analysing them, so would welcome any comments from the benchmark gurus watching this PR about the approach.

lrytz:
it looks like nothing is to be gained - should we close this PR?

adriaanm:
Thanks for exploring this, @jxtps! Closing this as it seems we should stay with the status quo.

PR #3388, opened 2014-01-20
SI-7445 ListMap.tail is returning wrong result

rklaehn:
Reverted the commit that introduced the bug, and modified HashMap to no
longer assume that tail is O(1).

Review by @Ichoran, @soc

PR #5103, opened 2016-04-17
Improve performance and behavior of ListMap and ListSet

ruippeixotog:
Makes the immutable `ListMap` and `ListSet` collections more alike one another, both in their semantics and in their performance.

In terms of semantics, makes the `ListMap` iterator return the elements in reverse order, as `ListSet` already does (improving its performance as a side-effect). While, as mentioned in SI-8985, `ListMap` and `ListSet` doesn't seem to make any guarantees in terms of iteration order, I believe users expect `ListSet` and `ListMap` to behave in the same way, particularly when they are implemented in the exact same way. The `next` method `ListSet`'s iterator now throws an exception when called after the last element, a behavior consistent with all the other collections.

In terms of performance, `ListMap` is given a custom builder that avoids creation in O(N^2) time using a strategy similar to the one already applied in `ListSet`. `ListSet`'s element removal method was not tail-recursive as the `ListMap` one, so a tail-recursive implementation was added.


ruippeixotog:
A Gist containing some naive benchmarks (together with the source code for them) is available here: https://gist.github.com/ruippeixotog/20f961055a94345ad24c2fd58079d26c

test/junit/scala/collection/immutable/ListMapTest.scala
```diff
@@ -0,0 +1,16 @@
+package scala.collection.immutable
```
janekdb, line 1: Add copyright header.

test/junit/scala/collection/immutable/ListSetTest.scala
```diff
@@ -0,0 +1,16 @@
+package scala.collection.immutable
```
janekdb, line 1: Add copyright header.

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -119,12 +119,8 @@ sealed class ListSet[A] extends AbstractSet[A]
     var that: ListSet[A] = self
     def hasNext = that.nonEmpty
     def next: A =
-      if (hasNext) {
-        val res = that.head
-        that = that.tail
-        res
-      }
-      else Iterator.empty.next()
+      if (!hasNext) throw new NoSuchElementException("next on empty iterator")
```
janekdb: Revert to delegation to `Iterator.empty.next()`. This avoids duplicating exception messages.

```
  /** The iterator which produces no values. */
val empty: Iterator[Nothing] = new AbstractIterator[Nothing] {
  def hasNext: Boolean = false
  def next(): Nothing = throw new NoSuchElementException("next on empty iterator")
}
```

Delegating to an empty instance to trigger an exception is also seen in other places in the std. lib. for example from `Range`,

```
final override def tail: Range = {
  if (isEmpty)
      Nil.tail

  drop(1)
}
```

Here `Nil.tail` is used for it's side effect,

```
override def tail: List[Nothing] =
  throw new UnsupportedOperationException("tail of empty list")
```

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -174,9 +170,17 @@ sealed class ListSet[A] extends AbstractSet[A]
...
-    override def -(e: A): ListSet[A] = if (e == head) self else {
-      val tail = self - e; new tail.Node(head)
-    }
+    override def -(e: A): ListSet[A] = removeInternal(e, this, Nil)
+
+    @tailrec private def removeInternal(k: A, cur: ListSet[A], acc: List[ListSet[A]]): ListSet[A] =
+      if (cur.isEmpty)
+        acc.last
+      else if (k == cur.head)
+        (cur.tail /: acc) {
```
janekdb: `/:`

http://docs.scala-lang.org/style/method-invocation.html

> Finally, the use of the /: and :\ should be avoided in preference to the more explicit foldLeft and foldRight method of Iterator. The right-associativity of the /: can lead to extremely confusing code, at the benefit of saving a few characters.

`:(`

Ichoran: I disagree.  I find the `/:` form much clearer, as it keeps the accumulated value always-on-the-left.  A high level of expertise can be assumed when reading or writing library code, so I think it's fine.  (Note that the style guide is not definitive.)

src/library/scala/collection/immutable/ListSet.scala
ruippeixotog: You're right, I misunderstood what the current code was doing. Using `Iterator.empty.next()` is indeed a good way to do this!

test/junit/scala/collection/immutable/ListMapTest.scala
ruippeixotog, line 1: Most, if not all, JUnit tests lack a copyright header. I already wrote several tests for this repository before and the copyright never was a requirement... Are you sure it's mandatory? @Ichoran, can you tell what is the "official" policy in these cases?

src/library/scala/collection/immutable/ListSet.scala
ruippeixotog: To be honest, it uses `/:` because I just copy-pasted and adjusted the `ListMap` code. If I was to write the code from scratch I'd probably use `foldLeft` (I also find `/:` a bit confusing) or even an imperative-style loop for performance reasons. As I also agree that standard library code is expected to be read by people with some expertise in the language (compared to other languages, Scala is actually very beginner-friendly on that matter!), I'm perfectly ok with both leaving this as it is now or rewriting it with a different style.

test/junit/scala/collection/immutable/ListMapTest.scala
janekdb, line 1: Confirming this was an assumption I had made. Please wait for the official policy.

src/library/scala/collection/immutable/ListMap.scala
```diff
@@ -32,13 +33,45 @@ object ListMap extends ImmutableMapFactory[ListMap] {
...
+            seen += kv
+          }
+        case None =>
+          elems += kv
+          seen += kv
+      }
+      this
+    }
+
+    def clear() = { elems.clear() }
```
lrytz: also `seen.clear()`?

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -174,9 +170,17 @@ sealed class ListSet[A] extends AbstractSet[A]
...
+          case (t, h) => val tt = t; new tt.Node(h.head)
```
lrytz: according to the comment in ListMap, `tt` exists due to https://issues.scala-lang.org/browse/SI-7459, which is marked as fixed now. could you remove `tt` in both places?

lrytz:
looks good otherwise. the build failure looks weird - let's just try again (amend your changes and `git push -f`).

ruippeixotog:
@lrytz thanks for the review, I just fixed both of the details you mentioned. The build failure didn't look like it was due to the modifications of this PR; I can try rebasing this into the current master if it doesn't work.

ruippeixotog:
I just rebased the branch on top of 2.12.x, but the error still occurs and Jenkins does not seem to be giving any helpful error messages. However, looking at the pull request list it seems that I'm the only one having this problem... Can anyone help me figuring out what's wrong with this PR?

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -174,9 +170,15 @@ sealed class ListSet[A] extends AbstractSet[A]
...
     /** `-` can be used to remove a single element from a set.
      */
-    override def -(e: A): ListSet[A] = if (e == head) self else {
-      val tail = self - e; new tail.Node(head)
-    }
+    override def -(e: A): ListSet[A] = removeInternal(e, this, Nil)
+
+    @tailrec private def removeInternal(k: A, cur: ListSet[A], acc: List[ListSet[A]]): ListSet[A] =
+      if (cur.isEmpty)
+        acc.last
```
Ichoran: These lines are pretty short.  It's easier to follow the logic if the result is on the same line as the conditional.  (You already went this way with the `hasNext` thing above, which is more debatable.)

src/library/scala/collection/immutable/ListMap.scala
```diff
@@ -32,13 +33,45 @@ object ListMap extends ImmutableMapFactory[ListMap] {
...
+    * A custom builder because forgetfully adding elements one at a time to a list backed map puts
+    * the "squared" in N^2. There is a temporary space cost, but it's improbable a list backed set
+    * could become large enough for this to matter given its pricy element lookup.
+    *
+    * This builder is reusable.
+    **/
+  class ListMapBuilder[A, B](initial: ListMap[A, B]) extends ReusableBuilder[(A, B), ListMap[A, B]] {
+    def this() = this(empty[A, B])
+    protected val elems = (new mutable.ListBuffer[(A, B)] ++= initial).reverse
+    protected val seen = new mutable.HashMap[A, B] ++= initial
```
Ichoran: Uh--this could get super, super expensive with small maps, which arguably is what everyone ought to have been using ListMap for to begin with.  Have you benchmarked how bad this is for lists of size 3 or so?

Ichoran:
I have one major concern about performance.  I wasn't able to spot why the test is failing, though.

lrytz:
For the compilation failure, if I re-add `.toList.reverseIterator` to `ListMap.iterator`, compilation succeeds. This means that some part of the compiler (i guess the part checking forward references) depends on the current iteration order.

Honestly, when i started looking at this PR, I was very surprised to learn that "`ListMap` and `ListSet` doesn't seem to make any guarantees in terms of iteration (SI-8985)". I think this situation creates a lot of confusion, e.g., http://stackoverflow.com/questions/3481925/insertion-ordered-listset. Just like `LinkedHashSet` / `LinkedHashMap` for mutable collections, I think that `ListSet` / `ListMap` should guarantee that "all traversal methods of this class visit elements in the order they were inserted".

src/library/scala/collection/immutable/ListMap.scala
ruippeixotog: Hi @Ichoran, I have just run some benchmarks for this `Builder` with small collections. The results are here: https://gist.github.com/ruippeixotog/199e72cfcf26577ff3c07bc16dec4e42.

From those results, it seems that the two algorithms even out after 4-5 elements. I do not have that much experience with microbenchmarking code, so I do not know if the code above produces stable enough results. What do you think of them?

ruippeixotog:
I was also surprised to see that `ListMap` and `ListSet` did not provide such guarantees - I always thought of them as an immutable alternative to `LinkedHashMap` and `LinkedHashSet`. If even the Scala compiler relies on their ordering, then I would rather make both `ListMap` and `ListSet` be insertion-ordered and document that properly in Scaladoc.

As you saw, iterating in the "correct" order requires the collection to be converted to a list and reversed, which has a performance hit. However, as @Ichoran mentioned, these collections are really only useful with a small number of elements, so I think it would be OK to do that.

Ichoran:
The benchmarking you did isn't really reliable enough to do anything with.  You should at least use Thyme; JMH would be better.

From this, it looks like you need to special-case 0-4.  Having a 2x performance hit in the most commonly used cases doesn't justify the gain in the cases which the data structure isn't suited for anyway.

But I'd do some more robust benchmarking first.

ruippeixotog:
Thanks for the framework suggestions! I benchmarked again `ListMap` using JMH. I also benchmarked `ListSet`. The full results are available here: [ListMap](https://gist.github.com/ruippeixotog/027684902621e395efe6b193195833a8), [ListSet](https://gist.github.com/ruippeixotog/95819bba5bd869c0f02f2ae736ccf0ba).

If I did not anything wrong in the benchmarks, it seems the custom builder is significantly slower for every size equal to or below 6... Considering again that the collection should only be used for very small sizes due to its overall performance, I don't think it's worth it to keep the custom builder and write special cases for small sizes.

If you take a look the test results, it seems that the custom `ListSet` builder is also significantly slower than the default one (a `mutable.SetBuilder`). It is expected, as the two collections are pretty much equal. Should I remove `ListSetBuilder` then?

Ichoran:
My recommendation would be to remove `ListSetBuilder` if it's not an improvement.

ruippeixotog:
Yes, I agree it's better that way. I have just pushed the updated commit.

lrytz:
OK, I agree the latest change looks good. A few small things to do:
- can you specify the iteration order in the scaladoc, similar to LinkedHashMap/Set?
- move the existing test case [t3822.scala](https://github.com/scala/scala/blob/4db1de6531933c927280226642235344d7f1de6a/test/files/run/t3822.scala) to the new junit test, so that they are all in the same place
- also [t7445](https://github.com/scala/scala/blob/4db1de6531933c927280226642235344d7f1de6a/test/files/run/t7445.scala) could be moved to junit. actually, `ListSet` probably has the same bug, it was only fixed for ListMap -- can you check? the fix for `ListMap` was in https://github.com/scala/scala/pull/3388.
- method `iterator` first creates a `List` instance, then `reverseIterator` creates the reversed list instance. the following would only create one single list:

```
  def iterator: Iterator[(A,B)] = {
    def reverseList = {
      var self = this
      var res: List[(A, B)] = Nil
      while (!self.isEmpty) {
        res = (self.key, self.value) :: res
        self = self.next
      }
      res
    }
    reverseList.iterator
  }
```

in general, maybe do another pass over the two classes to bring the implementations closer together (rename `ListMap.remove0` to `removeInternal`, replace `ListSet.unchecked_outer` by `next`, etc)

ruippeixotog:
Sure, I'll move t7445.scala code to a junit test and I'll improve the iterator as you suggested (adding two new tests for that). `ListSet` is indeed affected by SI-7445; I'll also fix that.

t3822 is a test that becomes ridiculously slow when the custom builder is taken off (though the Jenkins build seems to have made it past that test)... Do we still want to keep it? [SI-3822](https://issues.scala-lang.org/browse/SI-3822) is a very old issue and it involves an use case that we concluded here that was inadequate for that collection. Regardless of that, all `ListMap`/`ListSet` methods are now tail recursive, so the bug is fixed anyway.

Regarding the Scaladoc and the general revamp I was also thinking of doing that, as it's way easier to fix bugs and add improvement to those collections if their code is as similar as it could be (after all, its internal organization and algorithms associated are exactly the same). However, I believe it would be better if I did that either in a separate pull request or in a separate commit; otherwise, the bugs and improvements present in this PR would become much more difficult to track in the commit history later, as they would be concealed by a lot of method renames and code moves. What do you think? Is there a problem if this PR has multiple commits?

lrytz:
> t3822 is a test that becomes ridiculously slow

Right, I agree we can get rid of it.

> Is there a problem if this PR has multiple commits?

No, this is common practice in the scala repo. I agree that having separate commits for refactorings, scaladoc and bugfixes makes a lot of sense.

> involves an use case that we concluded here that was inadequate for that collection

Can you make sure to include that in the Scaladoc cleanup? All operations are `O(n)` (right?), including the creation of a ListSet/ListMap using companion apply.

ruippeixotog:
I have just added some new commits then. I organized the changes in:
- The original commit, improving the performance of the collections and changing `ListSet` to provide insertion-order iteration;
- Adding tests for both collections regarding some core operations. I found out that I had some bugs in the changes I did in the first commit and this makes sure no regressions happen now nor in the future;
- Making the two collections similar by renaming and reordering methods, improving their Scaladoc and their code style. Regarding this I removed the Scaladoc of most methods, as the text inherited by the superclasses is in general well written. On the other hand, I improved the class and object-level Scaladoc by making explicit the traversal order guarantee and the time complexity of the operations;
- Adding a `SerialVersionUID` annotation that was missing in `ListSet` and including `ListSet` in the serialization stability test in order to ensure that future changes to this collection do not break the serialization format (or at least alert that a break occurred).

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -13,84 +13,74 @@ package immutable
...
-/** $factoryInfo
- *  @define Coll immutable.ListSet
- *  @define coll immutable list set
- *  @since 1
- */
+/**
+  * $factoryInfo
+  *
+  * Note that each element insertion takes O(n) time, which means that creating a list set with
+  * n elements will take O(n^2^) time. This makes the builder suitable only for a small number of
```
lrytz, line 20: extra `^`

src/library/scala/collection/immutable/ListMap.scala
```diff
@@ -6,30 +6,34 @@
...
 object ListMap extends ImmutableMapFactory[ListMap] {
-  /** $mapCanBuildFromInfo */
+
+  /**
+    * $mapCanBuildFromInfo
+    */
   implicit def canBuildFrom[A, B]: CanBuildFrom[Coll, (A, B), ListMap[A, B]] =
     new MapCanBuildFrom[A, B]
+
   def empty[A, B]: ListMap[A, B] = EmptyListMap.asInstanceOf[ListMap[A, B]]
```
lrytz, line 37: could we remove the all the method overrides in the object `EmptyListMap`?

test/junit/scala/collection/immutable/ListMapTest.scala
```diff
@@ -21,6 +21,26 @@ class ListMapTest {
...
+    val m = ListMap(1 -> 1, 2 -> 2, 3 -> 3)
+    assertEquals(ListMap(1 -> 1, 2 -> 2, 3 -> 3, 4 -> 4), m + (4 -> 4))
+    assertEquals(ListMap(1 -> 1, 3 -> 3, 2 -> 4), m + (2 -> 4))
+    assertEquals(ListMap(1 -> 1, 2 -> 2, 3 -> 3), m + (2 -> 2))
+    assertEquals(ListMap(2 -> 2, 3 -> 3), m - 1)
+    assertEquals(ListMap(1 -> 1, 3 -> 3), m - 2)
+    assertEquals(ListMap(1 -> 1, 2 -> 2, 3 -> 3), m - 4)
+  }
+
+  @Test
```
lrytz, line 43: you could squash this commit into the previous one

src/library/scala/collection/immutable/ListSet.scala
```diff
@@ -115,33 +88,34 @@ sealed class ListSet[A] extends AbstractSet[A]
...
+      res
+    }
+    reverseList.iterator
   }
 
   /**
    *  @throws java.util.NoSuchElementException
    */
-  override def head: A = throw new NoSuchElementException("Set has no elements")
+  protected def elem: A = throw new NoSuchElementException("elem of empty set")
```
lrytz, line 96: could you mention in the commit message that this commit also fixes SI-7445 for ListSet?

lrytz:
other than these minor comments, looks good!

src/library/scala/collection/immutable/ListSet.scala
ruippeixotog, line 20: That is actually the superscript notation in Scaladoc: https://wiki.scala-lang.org/display/SW/Syntax#Syntax-Wiki-likeSyntaxinDocumentationSource. I checked the generated documentation and it appears correctly.

lrytz, line 20: ah, in that case you could use the same up here: https://github.com/scala/scala/pull/5103/files/78f1fdd0646e6da5de3f86eaec265d9b5d71b1f4..36a46b1cc2734504c95881c20b2495891c7b9bd7#diff-9f3860217088c43b2414a59d6ce4b2acR20

ruippeixotog:
I have just pushed the modifications you suggested :)

lrytz:
LGTM! thanks for keeping up with the pedantic reviewer :)

ruippeixotog:
No problem at all, they were good suggestions :)

src/library/scala/collection/immutable/ListMap.scala
```diff
@@ -6,223 +6,161 @@
...
+      res
+    }
+    reverseList.iterator
+  }
+
+  protected def key: A = throw new NoSuchElementException("key of empty map")
+  protected def value: B = throw new NoSuchElementException("value of empty map")
+  protected def next: ListMap[A, B] = throw new NoSuchElementException("next of empty map")
+
+  override def stringPrefix = "ListMap"
```
dwijnand, line 106: Btw this changed ListMap's toString between 2.11 and 2.12

PR #5104, opened 2016-04-18
Some potential dead codes in scala compiler

liuyi61ly:
We are doing the project Performance Enhancement of Scala Source Code. This time we find these potential dead codes in Scala compiler.

src\compiler\scala\tools\nsc\backend\Platform.scala It is completely base class definition.

src\compiler\scala\tools\nsc\transform\SampleTransform.scala  It is never called by other part in Scala compiler

src\compiler\scala\tools\reflect\FormatInerpolator.scala  It is never called by other part in Scala compiler

src\compiler\scala\tools\nsc\util\WorkScheduler.scala  It is never called by other part in Scala compiler
src\compiler\scala\tools\nsc\util\Interruptreq.scala  It only have some class call

src\compiler\scala\tools\nsc\util\Spec.scala  It is cmd relevant tool but never called during execution.


liuyi61ly:
cla — @scala-jenkins, i already sign the Scala CLA by clicking on 'Details'

soc:
Hey, I'm not sure what you are trying to do, but the commit doesn't seem to be what you intended!

lrytz:
indeed.. https://github.com/scala/scala/pull/5104/files

Erraach-walid:
*-*hello

vovafuckinkProg:
Hi

PR #2354, opened 2013-04-04
SI-7289 Less strict type application for TypeVar.

adriaanm:
(Proposed alternative for #2294)

When a type constructor variable is applied to the wrong number of arguments,
return a new type variable whose instance is `ErrorType`.

Let's dissect the reported test case.

Define the first implicit:

```
scala> trait Schtroumpf[T]
defined trait Schtroumpf

scala> implicit def schtroumpf[T, U <: Coll[T], Coll[X] <: Traversable[X]]
     |     (implicit minorSchtroumpf: Schtroumpf[T]): Schtroumpf[U] = ???
schtroumpf: [T, U <: Coll[T], Coll[X] <: Traversable[X]](implicit minorSchtroumpf: Schtroumpf[T])Schtroumpf[U]
```

Call it explicitly => kind error during type inference reported.

```
scala> schtroumpf(null): Schtroumpf[Int]
<console>:10: error: inferred kinds of the type arguments (Nothing,Int,Int) do not conform to the expected kinds of the type parameters (type T,type U,type Coll).
Int's type parameters do not match type Coll's expected parameters:
class Int has no type parameters, but type Coll has one
              schtroumpf(null): Schtroumpf[Int]
              ^
<console>:10: error: type mismatch;
 found   : Schtroumpf[U]
 required: Schtroumpf[Int]
              schtroumpf(null): Schtroumpf[Int]
                        ^
```

Add another implicit, and let implicit search weigh them up.

```
scala> implicitly[Schtroumpf[Int]]
<console>:10: error: diverging implicit expansion for type Schtroumpf[Int]
starting with method schtroumpf
              implicitly[Schtroumpf[Int]]
                        ^

scala> implicit val qoo = new Schtroumpf[Int]{}
qoo: Schtroumpf[Int] = $anon$1@c1b9b03

scala> implicitly[Schtroumpf[Int]]
<crash>
```

Implicit search compares the two in-scope implicits in `isStrictlyMoreSpecific`,
which constructs an existential type:

   type ET = Schtroumpf[U] forSome { type T; type U <: Coll[T]; type Coll[_] <: Traversable[_] }

A subsequent subtype check `ET <:< Schtroumpf[Int]` gets to `withTypeVars`, which
replaces the quantified types with type variables, checks conformance of that
substitued underlying type against `Schtroumpf[Int]`, and then tries to solve
the collected type constraints. The type var trace looks like:

```
[...]

PR #5106, opened 2016-04-19
SI-9760 Fix for higher-kinded GADT refinement

milessabin:


test/files/pos/hkgadt.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  def frob[F[_]](foo: Foo[F]): F[Int] =
+    foo match {
+      case Bar() =>
+         List(1)
+    }
+
+  sealed trait Foo1[F]
+  final case class Bar1() extends Foo1[Int]
+  def frob1[A](foo: Foo1[A]) = foo match {
```
larsrh: Are you missing a return type here?

milessabin: Yes. It should be `A`. Will fix.

adriaanm:
FTR, this condition was added as a workaround for https://issues.scala-lang.org/browse/SI-4070. Since its tests are still passing, I guess we've improved existentials enough by now that they can kind of quantify over type constructors...

Could you also play around with some neg test cases? Maybe something with more than one case to exercise lubbing?

milessabin:
There's more information on the [ticket](https://issues.scala-lang.org/browse/SI-9760). At @retronym's suggestion I git-bisected my way to determinng that [this commit of yours](https://github.com/scala/scala/pull/2354/commits/6a61e172d9de912c74af50e49f0057fbe8d4e500) was what made the additional condition unnecessary.

I'll add some more tests.

milessabin:
@adriaanm more tests (multiple cases for pos, and some neg cases) added.

adriaanm:
LGTM

PR #5107, opened 2016-04-19
scala-dev-113: Sync root package words to layout

janekdb:
The package list is on the right.


janekdb:
https://github.com/scala/scala-dev/issues/113

SethTisue:
LGTM

PR #5109, opened 2016-04-20

lrytz:
Implicit conversions are now in package convert as ImplicitConversions,
ImplicitConversionsToScala and ImplicitConversionsToJava.

Deprecated WrapAsJava, WrapAsScala and the values in package object.

Improve documentation.

### DIFF
--- a/src/library/scala/collection/convert/ImplicitConversions.scala
+++ b/src/library/scala/collection/convert/ImplicitConversions.scala
@@ -17,358 +17,76 @@ import JavaConverters._
 
 /** Defines implicit converter methods from Java to Scala collections. */
 trait ToScalaImplicits {
-
-  /**
-   * Implicitly converts a Java `Iterator` to a Scala `Iterator`.
-   *
-   * The returned Scala `Iterator` is backed by the provided Java `Iterator`
-   * and any side-effects of using it via the Scala interface will be visible
-   * via the Java interface and vice versa.
-   *
-   * If the Java `Iterator` was previously obtained from an implicit or
-   * explicit call of `asIterator(scala.collection.Iterator)` then the
-   * original Scala `Iterator` will be returned.
-   *
-   * @param it The `Iterator` to be converted.
-   * @return   A Scala `Iterator` view of the argument.
-   */
+  /** Implicitly converts a Java `Iterator` to a Scala `Iterator`. See [[asScalaIterator]]. */
   implicit def `iterator asScala`[A](it: ju.Iterator[A]): Iterator[A] = asScalaIterator(it)
 
-  /**
-   * Implicitly converts a Java Enumeration to a Scala Iterator.
-   * The returned Scala Iterator is backed by the provided Java
-   * Enumeration and any side-effects of using it via the Scala interface will
-   * be visible via the Java interface and vice versa.
-   *
-   * If the Java Enumeration was previously obtained from an implicit or
-   * explicit call of `enumerationAsScalaIterator(scala.collection.Iterator)`
-   * then the original Scala Iterator will be returned.
-   *
-   * @param i The Enumeration to be converted.
-   * @return A Scala Iterator view of the argument.
-   */
+  /** Implicitly converts a Java `Enumeration` to a Scala `Iterator`. See [[enumerationAsScalaIterator]]. */
   implicit def `enumeration AsScalaIterator`[A](i: ju.Enumeration[A]): Iterator[A] = enumerationAsScalaIterator(i)
 
-  /**
-   * Implicitly converts a Java `Iterable` to a Scala `Iterable`.
-   *
-   * The returned Scala `Iterable` is backed by the provided Java `Iterable`
-   * and any side-effects of using it via the Scala interface will be visible
-   * via the Java interface and vice versa.
-   *
-   * If the Java `Iterable` was previously obtained from an implicit or
-   * explicit call of `iterableAsScalaIterable(scala.collection.Iterable)`
-   * then the original Scala Iterable will be returned.
-   *
-   * @param i The Iterable to be converted.
-   * @return A Scala Iterable view of the argument.
-   */
+  /** Implicitly converts a Java `Iterable` to a Scala `Iterable`. See [[iterableAsScalaIterable]]. */
   implicit def `iterable AsScalaIterable`[A](i: jl.Iterable[A]): Iterable[A] = iterableAsScalaIterable(i)
 
-  /**
-   * Implicitly converts a Java `Collection` to an Scala `Iterable`.
-   *
-   * If the Java `Collection` was previously obtained from an implicit or
-   * explicit call of `collectionAsScalaIterable(scala.collection.SizedIterable)`
-   * then the original Scala `Iterable` will be returned.
-   *
-   * @param i The Collection to be converted.
-   * @return A Scala Iterable view of the argument.
-   */
+  /** Implicitly converts a Java `Collection` to an Scala `Iterable`. See [[collectionAsScalaIterable]]. */
   implicit def `collection AsScalaIterable`[A](i: ju.Collection[A]): Iterable[A] = collectionAsScalaIterable(i)
 
-  /**
-   * Implicitly converts a Java `List` to a Scala mutable `Buffer`.
-   *
-   * The returned Scala `Buffer` is backed by the provided Java `List`
-   * and any side-effects of using it via the Scala interface will
-   * be visible via the Java interface and vice versa.
-   *
-   * If the Java `List` was previously obtained from an implicit or
-   * explicit call of `asScalaBuffer(scala.collection.mutable.Buffer)`
-   * then the original Scala `Buffer` will be returned.
-   *
-   * @param l The `List` to be converted.
-   * @return A Scala mutable `Buffer` view of the argument.
-   */
+  /** Implicitly converts a Java `List` to a Scala mutable `Buffer`. See [[asScalaBuffer]]. */
   implicit def `list asScalaBuffer`[A](l: ju.List[A]): mutable.Buffer[A] = asScalaBuffer(l)
 
-  /**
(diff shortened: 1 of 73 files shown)


lrytz:
Review by @som-snytt 

I squashed all your commits into one.

Compared to the latest commit in #5064 the following has changed:
- Fixed the few test failures
- Eliminated a few remaining references to `JavaConversions`, `wrapAs..`, `decorateAs..`
- Completely deprecated `WrapAsJava` / `WrapAsScala` (instead of just `deprecatedInheritance`). Also deprecated the `WrapAsJava` / `WrapAsScala` objects
- Added `ImplicitConversionsToScala` and `ImplicitConversionsToJava` next to your `ImplicitConversions`. Added a doc-comment explaining that explicit conversions are to be preferred.

lrytz:
There's one problem created by this PR. Now `JavaConverters` provides explicit conversion methods such as

```
scala> val jit: java.util.Iterator[String] = java.util.Collections.emptyList[String]().iterator()
jit: java.util.Iterator[String] = java.util.Collections$EmptyIterator@68b58644

scala> collection.JavaConverters.asScalaIterator(jit)
res2: Iterator[String] = empty iterator
```

The `JavaConverters.asScalaIterator` method did not exist previously.

Unfortunately, the names of these explicit converter methods are exactly the same as the names of the implicit converter methods in the deprecated `JavaConversions`. A problem manifests if a user imports both `JavaConverters._` and `JavaConversions._`:

```
scala> :paste

object T {
  import collection.JavaConverters._
  import collection.JavaConversions._
  val jit: java.util.Iterator[String] = java.util.Collections.emptyList[String]().iterator()
  val sit: Iterator[String] = jit
}

<console>:16: error: type mismatch;
 found   : java.util.Iterator[String]
 required: scala.collection.Iterator[String]
         val sit: Iterator[String] = jit
                                     ^
```

What happens here (`-Xlog-implicits`) is that implicit search finds `JavaConversions.asScalaIterator`, but when it tries to type-check `asScalaIterator(jit)` it encounters an error (ambiguous reference, the method is imported twice) and discards the implicit conversion.

So basically when somebody has both imports, the implicit conversions will stop working after this PR. There will not be a deprecation warning for `JavaConversions` because these are issued later, in refchecks, while the type error occurs during typer.

The alternative would be to rename all implicit conversion methods in `JavaConversions`, which would break code that refers to them explicitly (`JavaConversions.asScalaIterator(jit)`). I think this is actually more common than having both imports. On the other hand, the error message would be more clear in this case (`value asScalaIterator is not a member of JavaConversions`), so maybe it would be better to go with this option.

src/library/scala/collection/convert/ImplicitConversions.scala
```diff
@@ -0,0 +1,407 @@
...
+
+  /**
+   * Implicitly converts a Java `Iterator` to a Scala `Iterator`.
+   *
+   * The returned Scala `Iterator` is backed by the provided Java `Iterator`
+   * and any side-effects of using it via the Scala interface will be visible
+   * via the Java interface and vice versa.
+   *
+   * If the Java `Iterator` was previously obtained from an implicit or
+   * explicit call of `asIterator(scala.collection.Iterator)` then the
```
som-snytt: I didn't update the method names referred to by the implicits here.

test/disabled/presentation/akka/src/akka/actor/ActorRegistry.scala
```diff
@@ -342,7 +342,7 @@ class Index[K <: AnyRef, V <: AnyRef: ArrayTag] {
    * if no matches it returns None
    */
   def findValue(key: K)(f: (V) => Boolean): Option[V] = {
-    import scala.collection.JavaConversions._
+    import scala.collection.convert.wrapAsScala._
```
som-snytt, line 345: I was too lazy to re-update disabled tests.

src/library/scala/collection/package.scala
```diff
@@ -76,13 +76,9 @@ package scala
...
  *
- * === Converting between Java Collections ===
+ * === Converting to and from Java Collections ===
  *
- * The [[scala.collection.JavaConversions]] object provides implicit defs that
- * will allow mostly seamless integration between APIs using Java Collections
- * and the Scala collections library.
- *
- * Alternatively the [[scala.collection.JavaConverters]] object provides a collection
+ * The [[scala.collection.JavaConverters]] object provides a collection
```
som-snytt, line 81: The final wording in `JavaConverters` is improved: `a variety of decorators that enable converting between Scala and Java collections...`.

som-snytt:
Thanks for picking this up. Otherwise, it's kind of thankless. I had another pass to make on the old PR.

I didn't notice that consequence of the naming conflict you point out. I don't have a clever workaround. I would definitely go with not supporting wildcard imports from both. They are presented as alternatives, and no examples show imports from both.

If the issue were serious enough, the "new" explicit methods could be renamed `toSomething`, although that ruins the naming convention. I don't think it's serious enough; but it muddies the release notes.

som-snytt:
Actually, the workaround is to add implicits back to `JavaConversions` by adding them to `WrapAsScala`, etc, under new innocuous names. So this makes the example work again. Adding a method seems like going in the wrong direction, but `Wrap` is deprecated so it's OK. In fact, someone commented somewhere that no one ever introduces API and deprecates it at the same time. Ha!

```
  implicit def `deprecated asScalaIterator`[A](it: ju.Iterator[A]): Iterator[A] =
    asScalaIterator(it)
```

Actually it has to be a lower-priority implicit, to allow implicit selection.

som-snytt:
Yeah, tried it out:

```
// Identical fallback implicits, in case old names are hidden by new members of JavaConverters
@deprecated("Use JavaConverters or consider ToScalaImplicits", since="2.12")
trait WrapAsScala extends LowPriorityWrapAsScala {
  implicit def `deprecated asScalaIterator`[A](it: ju.Iterator[A]): Iterator[A] = asScalaIterator(it)
}

trait LowPriorityWrapAsScala {
```

Making the new method more specific makes it easier to type in.

som-snytt:
Sample `WrapAsScala`

https://github.com/scala/scala/compare/2.12.x...som-snytt:review/pr5064?expand=1

test/disabled/presentation/akka/src/akka/actor/ActorRegistry.scala
lrytz, line 345: i didn't catch all of them either, i think

lrytz:
@som-snytt i pushed two more commits to address your feedback.

lrytz:
generated docs at https://dl.dropboxusercontent.com/u/4881107/library/scala/collection/JavaConverters$.html

som-snytt:
Your solution of renaming the implicits is different from what I proposed. Does it work? It's too late in afternoon here for me to know without trying. I thought importing `f` twice makes it unavailable because of the ambiguity, even if only one is implicit. My proposal was let `f` stop working but add a `g` to the deprecated interface. The `g` has to be more specific than `f`. I only like the idea because it sounds like a crazy solution.

lrytz:
It seems to work, here's the test: https://github.com/scala/scala/pull/5109/commits/75b8875d202c7148bdf36c72b098e977f89931f4#diff-9acf6de82fd57ddd8dc4c907a07bfe72

I liked your idea, but then realized that renaming the implicits and providing non-implicit methods with the old name also works, and it seems simpler.

retronym:
Renaming implicits is itself not source compatible: someone might have imported one specifically, rather than via a wildcard.

lrytz:
ah, true.. i'll give @som-snytt's idea try then.

som-snytt:
Now I see what you had done. I was on minimal sleep and had to get the daughter to Met Opera at the cinema, Sondra Radvanovsky as Q.E. 1. What a performance. If the daughter grows up to be like either Radvanovsky or Elizabeth, I'd be OK with it. Modulo beheading your ex. I didn't actually sleep in Act 1, but I may have rested my eyes.

som-snytt:
I'd still like a motivating use case for `@deprecatedImplicitly` to disable implicit candidacy under `-deprecation`.

lrytz:
> I'd still like a motivating use case for @deprecatedImplicitly to disable implicit candidacy under -deprecation.

Would this mean that `-deprecation` changes semantics? I think that's not a good idea.

lrytz:
I pushed another commit that adds the higher-priority implicits with non-clashing names to JavaConversions.

som-snytt:
Backquotes are totally the boss.

You're right about `-deprecation`. Maybe it's a job for the deprecator plugin that elides deprecated API, in which case a special annotation isn't needed.

The use case for `@deprecatedImplicitly` would be to warn only on implicit usage. But maybe it could function as the explicit priority that people have asked for. It would lower the specificity score. For example, in the current commit, instead of a low priority trait, just annotate the old name and add the new name.

lrytz:
rebased due to merge conflict. @som-snytt lgty?

src/library/scala/collection/package.scala
som-snytt, line 81: A follow-up tweak PR can use lrytz's word `miscellaneous` which is more precise than `variety` (which can mean "one kind" or species).

som-snytt:
LGTM

I see you were cleverly efficient with some more doc updates.

ijuma:
Have we considered the case where Java code uses the methods in `JavaConversions` explicitly because it's more convenient than using `JavaConverters` explicitly? I've seen a few cases like that in existing open-source projects.

som-snytt:
@ijuma [The doc](http://www.scala-lang.org/api/2.12.0-M5/scala/collection/JavaConverters$.html) says "Alternatively, the conversion methods have descriptive names and can be invoked explicitly."

So now they are available on JavaConverters. That wasn't the case previously, only in order to make the implicits work.

lrytz deals with the naming starting at https://github.com/scala/scala/pull/5109#issuecomment-212417789

PR #5110, opened 2016-04-20
Remove the duplicate implem of hash codes for numbers.

sjrd:
Previously, there were two separate implementations of hash
code for boxed number classes:
- One in Statics, used by the codegen of case class methods.
- One in ScalaRunTime + BoxesRunTime, used by everything else.

This commit removes the variant implemented in ScalaRunTime +
BoxesRunTime, and always uses Statics instead. We use Statics
because the one from ScalaRunTime causes an unnecessary module
load.

The entry point ScalaRunTime.hash() is kept, as deprecated,
for bootstrapping reasons.


sjrd:
Review by @lrytz, please.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1013,24 +1013,20 @@ abstract class Erasure extends AddInterfaces
...
-                    val alts    = ScalaRunTimeModule.info.member(nme.hash_).alternatives
-                    def alt1    = alts find (_.info.paramTypes.head =:= qual.tpe)
-                    def alt2    = ScalaRunTimeModule.info.member(nme.hash_) suchThat (_.info.paramTypes.head.typeSymbol == AnyClass)
-                    val newTree = gen.mkRuntimeCall(nme.hash_, qual :: Nil) setSymbol (alt1 getOrElse alt2)
-
-                    global.typer.typed(newTree)
+                  case LongClass                                => staticsCall("longHash")
+                  case FloatClass                               => staticsCall("floatHash")
+                  case DoubleClass                              => staticsCall("doubleHash")
+                  case _                                        => staticsCall("anyHash")
```
lrytz: I'd prefer to have those strings in `StdNames`

sjrd: Ah true. I'll do that, thanks.

sjrd:
Updated.

Note that there is a new commit that first _fixes_ `Statics.doubleHash`. The problem was found in the review of https://github.com/scala-js/scala-js/pull/2346
The first commit contains a test that asserts consistency between `Statics.doubleHash` and `ScalaRunTime.hash` for doubles. That test would fail without the adaptation.

scottcarey:
As noted in https://github.com/scala-js/scala-js/pull/2346

An implementation of Statics.longHash that has no conditionals and is consistent with Statics.intHash would be:

``` scala
val lo = lv.toInt // contains  original bits:  64 (in position 32) and 31 to 1
val hi = ((lv >>> 31) & 0xFFFFFFFF) // contains original bits 63 to 32
val hi2 = hi + (lo >>> 31) // add 1 if a negative number, if 'hi' is all 1's they become 0's
lo ^ hi2
```

or in java

``` java
int lo = (int) lv; // contains  original bits:  64 (in position 32) and 31 to 1
int hi = ((lv >>> 31) & 0xFFFFFFFF); // contains original bits 63 to 32
int hi2 = hi + (lo >>> 31) // add 1 if a negative number, if 'hi' is all 1's they become 0's
return lo ^ hi2
```

The implementation in this PR has a conditional, which should be avoided if possible.

``` java
  int iv = (int)lv;
  if (iv == lv)
    return iv;

  return java.lang.Long.hashCode(lv);
```

But also returns a different value for longs that are larger than an int.   Do we have to retain the exact same hashes, or merely be consistent in double / float ?

sjrd:
As also noted in the Scala.js PR, the algorithm above would be much more expensive to compute in Scala.js, because it uses 2 long operations before we can extract the `hi` part.

Another branch-less algorithm (with yet another result) which would be fast on Scala.js too would be:

``` java
int lo = (int)lv;
lo ^ ((int)(lv >>> 32)) ^ (lo >> 31)
```

Anyway, the goal of this PR is not to change the existing implementation. Only to remove code duplication. Improving the status quo should be discussed separately, IMO.

scottcarey:
The above can simplify doubleHash too:

``` java
  public static int doubleHash(double dv) {
    long lv = (long)dv;
    if (lv == dv)
      return longHash(lv);

    float fv = (float)dv;
    if (fv == dv)
      return java.lang.Float.hashCode(fv);

    return java.lang.Double.hashCode(dv);
  }
```

Because if it matches a long, it will match the int, and we can just calculate the longHash.

scottcarey:
```
Anyway, the goal of this PR is not to change the existing implementation. Only to remove code duplication. Improving the status quo should be discussed separately, IMO.
```

I agree, though my first impression here was that because of the bug fix, the existing implementation is changing (for case classes at least?).

```
int lo = (int)lv;
lo ^ ((int)(lv >>> 32)) ^ (lo >> 31)
```

Drops a bit, using only 63 of 64 for the hash.

Two variations that do not (and only have 1 64 bit op, for javascript) are:

``` java
int lo = (int) lv;
int hi = ((int)(lv >>> 31)) & 0xFFFFFFFF;
lo ^ (hi + (lo >>> 31));
```

``` java
int lo = (int) lv;
int hi = ((int)(lv >>> 31)) & 0xFFFFFFFF;
lo ^ (hi ^ (lo >> 31));
```

scottcarey:
ok, a couple mistakes in my bits above -- casting to int with `(int)` in java or `toInt` in scala simply truncates, taking the lower 32 bits.  No need to do the odd shifting by 31 instead of 32, so the algorithm by @sjrd  is correct, and mine is what drops a bit.

corrected branchless implementation options, both of which are consistent with `java.lang.Integer.hashCode` -- but not `java.lang.Long.hashCode` when the magnitude is larger than an int.  In one case the hash is off by one, and in the other the bits are flipped, for values outside the range of Int.

``` java
int lo = (int) lv;
int hi = (int)(lv >>> 32);
lo ^ (hi + (lo >>> 31));
```

``` java
int lo = (int) lv;
int hi = (int)(lv >>> 32);
lo ^ (hi ^ (lo >> 31));
```

lrytz:
LGTM

PR #5111, opened 2016-04-21
Add examples to Exception object and group members

janekdb:
- Extend main comment with additional examples
- Group methods from a user perspective
- List exceptions special cased by shouldRethrow
- Include overlooked withTry in opt, either notes.

### DIFF
--- a/src/library/scala/util/control/Exception.scala
+++ b/src/library/scala/util/control/Exception.scala
@@ -13,21 +13,136 @@ package control
 import scala.reflect.{ ClassTag, classTag }
 import scala.language.implicitConversions
 
-
 /** Classes representing the components of exception handling.
- *  Each class is independently composable.  Some example usages:
+ *
+ *  Each class is independently composable.
+ *
+ *  This class differs from [[scala.util.Try]] in that it focuses on composing exception handlers rather than
+ *  composing behavior.  All behavior should be composed first and fed to a [[Catch]] object using one of the
+ *  `opt`, `either` or `withTry` methods. Taken together the classes provide a DSL for composing catch and finally
+ *  behaviors.
+ *
+ *  === Examples ===
+ *
+ *  Create a `Catch` which handles specified exceptions.
  *  {{{
  *  import scala.util.control.Exception._
  *  import java.net._
  *
  *  val s = "http://www.scala-lang.org/"
- *  val x1 = catching(classOf[MalformedURLException]) opt new URL(s)
- *  val x2 = catching(classOf[MalformedURLException], classOf[NullPointerException]) either new URL(s)
+ *
+ *  // Some(http://www.scala-lang.org/)
+ *  val x1: Option[URL] = catching(classOf[MalformedURLException]) opt new URL(s)
+ *
+ *  // Right(http://www.scala-lang.org/)
+ *  val x2: Either[Throwable,URL] =
+ *    catching(classOf[MalformedURLException], classOf[NullPointerException]) either new URL(s)
+ *
+ *  // Success(http://www.scala-lang.org/)
+ *  val x3: Try[URL] = catching(classOf[MalformedURLException], classOf[NullPointerException]) withTry new URL(s)
+ *
+ *  val defaultUrl = new URL("http://example.com")
+ *  //  URL(http://example.com) because htt/xx throws MalformedURLException
+ *  val x4: URL = failAsValue(classOf[MalformedURLException])(defaultUrl)(new URL("htt/xx"))
  *  }}}
  *
- *  This class differs from `scala.util.Try` in that it focuses on composing exception handlers rather than
- *  composing behavior.   All behavior should be composed first and fed to a `Catch` object using one of the
- *  `opt` or `either` methods.
+ *  Create a `Catch` which logs exceptions using `handling` and `by`.
+ *  {{{
+ *  def log(t: Throwable): Unit = t.printStackTrace
+ *
+ *  val withThrowableLogging: Catch[Unit] = handling(classOf[MalformedURLException]) by (log)
+ *
+ *  def printUrl(url: String) : Unit = {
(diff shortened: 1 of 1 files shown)


janekdb:
scaladoc-exception-examples-groups-v2

janekdb:
Review by @dickwall, @som-snytt, @SethTisue, @parambirs .

src/library/scala/util/control/Exception.scala
```diff
@@ -13,24 +13,125 @@ package control
...
-
 /** Classes representing the components of exception handling.
- *  Each class is independently composable.  Some example usages:
+ *
+ *  Each class is independently composable.
+ *
+ *  This class differs from [[scala.util.Try]] in that it focuses on composing exception handlers rather than
+ *  composing behavior.  All behavior should be composed first and fed to a [[Catch]] object using one of the
+ *  `opt`, `either` or `withTry` methods. Taken together the classes provide a DSL for composing catch and finally
+ *  behaviours.
```
som-snytt: Two flavo(u)rs of behavio(u)r. Where is spell bot? Does it standardise on Swiss spelling? Standardize?

src/library/scala/util/control/Exception.scala
```diff
@@ -13,24 +13,125 @@ package control
...
+ *  === Examples ===
+ *
+ *  Create a `Catch` which handles specified exceptions.
  *  {{{
  *  import scala.util.control.Exception._
  *  import java.net._
  *
  *  val s = "http://www.scala-lang.org/"
+ *
+ *  // Option[URL]
```
som-snytt: I think it's more natural to show types inline. `val x: Option[URL] = RHS`. or show a REPL transcript if desirable to show values. I don't know that there is a standard for docs yet, let alone executable examples, but I like the idea of REPL transcripts. I've seen them somewhere. I just saw example at JavaConverters will shows explicit types.

src/library/scala/util/control/Exception.scala
```diff
@@ -13,24 +13,125 @@ package control
...
+ *
+ *  @groupname composition-catch-promiscuously Promiscuous Catch behaviours
+ *  @groupprio composition-catch-promiscuously 50
+ *  @groupdesc composition-catch-promiscuously Useful if catching `ControlThrowable` or `InterruptedException` is required.
+ *
+ *  @groupname logic-container Logic Containers
+ *  @groupprio logic-container 60
+ *  @groupdesc logic-container Containers for catch and finally behaviour.
  *
- *  @author Paul Phillips
```
som-snytt, line 32: I don't normally like author tags, but I think this one is justified. This stuff is especially paulpy. It's up to the build not emit the doc tags. And other major contributors do add their names at times, I guess for self-branding purposes.

som-snytt:
Thanks, this is educational. At lunch break, I'm going to work through the examples.

src/library/scala/util/control/Exception.scala
janekdb: I too have read a suggestion that examples should be copy/paste/runnable in the REPL. I started with this approach and I liked it but using an IDEA worksheet was easier in the end. I'll try it again now the examples are stable and see what the output looks like. Failing that moving the types inline is a good idea.

src/library/scala/util/control/Exception.scala
janekdb, line 32: Restored.

src/library/scala/util/control/Exception.scala
janekdb: This is the raw REPL output from evaluating the first example expression by expression,

```
scala> import scala.util.control.Exception._
import scala.util.control.Exception._

scala> import java.net._
import java.net._

scala> val s = "http://www.scala-lang.org/"
s: String = http://www.scala-lang.org/

scala> val x1 = catching(classOf[MalformedURLException]) opt new URL(s)
x1: Option[java.net.URL] = Some(http://www.scala-lang.org/)

scala> val x2 = catching(classOf[MalformedURLException], classOf[NullPointerException]) either new URL(s)
x2: scala.util.Either[Throwable,java.net.URL] = Right(http://www.scala-lang.org/)

scala> val x3 = catching(classOf[MalformedURLException], classOf[NullPointerException]) withTry new URL(s)
x3: scala.util.Try[java.net.URL] = Success(http://www.scala-lang.org/)

scala> val defaultUrl = new URL("http://example.com")
defaultUrl: java.net.URL = http://example.com

scala> val x4 = failAsValue(classOf[MalformedURLException])(defaultUrl)(new URL("htt/xx"))
x4: java.net.URL = http://example.com
```

The `scala>` prompt is clutter that distracts from the example code but the REPL presentation of the expression values is an improvement over the content of the comment style though I still prefer the comments. I like the idea of augmenting the current comments with the full expression value while including the type explicitly on the LHS.

src/library/scala/util/control/Exception.scala
janekdb: Occurrences of each variation,

| Variation | Count |
| --- | --- |
| behavior | 176 |
| behaviour | 51 |

Majority rule => behavior

janekdb:
Review by @som-snytt, @heathermiller.

src/library/scala/util/control/Exception.scala
janekdb: Fixed.

som-snytt: variacioun. http://quod.lib.umich.edu/cgi/m/mec/med-idx?type=id&id=MED50741

lrytz:
LGTM, very nice!

som-snytt:
There's also https://issues.scala-lang.org/browse/SI-9073

I don't know why you don't like birds. Why would you want to kill two with one stone?

janekdb:
Hey, I like Kestrels in any combination!

PR #5116, opened 2016-04-22
More efficient code for deciding if a mixin forwarders is needed

lrytz:
Follow-up for #5085

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Mixin.scala
+++ b/src/compiler/scala/tools/nsc/transform/Mixin.scala
@@ -8,6 +8,7 @@ package transform
 
 import symtab._
 import Flags._
+import scala.annotation.tailrec
 import scala.collection.mutable
 
 abstract class Mixin extends InfoTransform with ast.TreeDSL {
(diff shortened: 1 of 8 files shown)


src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -266,11 +261,23 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
...
-              if (needsForwarder)
+
+              @tailrec
+              def existsCompetingMethod(baseClasses: List[Symbol]): Boolean = baseClasses match {
+                case baseClass :: rest =>
+                  if (baseClass ne mixinClass) {
+                    val m = member.overriddenSymbol(baseClass)
+                    val isCompeting = m.exists && {
+                      !m.owner.isTraitOrInterface ||
+                        (!m.isDeferred && !mixinSuperInterfaces.contains(m.owner))
```
retronym: How did you weigh the choice between this formulation `mixinSuperInterfaces.contains(m.owner)` and my proposal of `mixinClass.isNonBottomSubclass(m.owner)`? The latter amounts to a binary search over an existing data structure (the base type seq).

lrytz: I had a failure at some point because `isNonBottomSubclass` includes parent classes (not only traits/interfaces). I can't remember the details, so I pushed a commit here and trying it out. It looks like class members are handled in the line above (`m.owner.isTraitOrInterface`) so it shouldn't make a difference.

adriaanm:
`JavaUniverseForce` needs an update:

``` diff
--- a
+++ b
@@ -379,2 +379,3 @@
     definitions.ImplicitAmbiguousClass
+    definitions.JUnitTestClass
     definitions.MigrationAnnotationClass
```

```
// Generated Code, validated by run/t6240-universe-code-gen.scala
package scala.reflect
package runtime

trait JavaUniverseForce { self: runtime.JavaUniverse  =>
```

lrytz:
rebased, review by @retronym

retronym:
LGTM

PR #5117, opened 2016-04-22
SI-9516 Fix the behavior of Int shift Long operations.

sjrd:
In any shift operation where the lhs is an Int (or smaller) and
the rhs is a Long, the result kind must be Int, and not Long.
This is important because the lhs must _not_ be promoted to a
Long, as that causes an opcode for long shift to be emitted.
This uses an rhs modulo 64, instead of int shifts which use an
rhs module 32. Instead, the rhs must be downgraded to an Int.

The new behavior is consistent with the same operations in the
Java programming language.

Link to SI-9516: https://issues.scala-lang.org/browse/SI-9516

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/ConstantFolder.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ConstantFolder.scala
@@ -102,13 +102,13 @@ abstract class ConstantFolder {
     case nme.XOR => Constant(x.longValue ^ y.longValue)
     case nme.AND => Constant(x.longValue & y.longValue)
     case nme.LSL if x.tag <= IntTag
-                 => Constant(x.intValue << y.longValue)
+                 => Constant(x.intValue << y.longValue.toInt) // TODO: remove .toInt once starr includes the fix for SI-9516 (2.12.0-M5)
     case nme.LSL => Constant(x.longValue <<  y.longValue)
     case nme.LSR if x.tag <= IntTag
-                 => Constant(x.intValue >>> y.longValue)
+                 => Constant(x.intValue >>> y.longValue.toInt) // TODO: remove .toInt once starr includes the fix for SI-9516 (2.12.0-M5)
     case nme.LSR => Constant(x.longValue >>> y.longValue)
     case nme.ASR if x.tag <= IntTag
-                 => Constant(x.intValue >> y.longValue)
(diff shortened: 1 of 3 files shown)


sjrd:
Review by @lrytz

src/compiler/scala/tools/nsc/typechecker/ConstantFolder.scala
```diff
@@ -102,13 +102,13 @@ abstract class ConstantFolder {
     case nme.XOR => Constant(x.longValue ^ y.longValue)
     case nme.AND => Constant(x.longValue & y.longValue)
     case nme.LSL if x.tag <= IntTag
-                 => Constant(x.intValue << y.longValue)
+                 => Constant(x.intValue << y.longValue.toInt) // .toInt required to bootstrap despite SI-9516
```
lrytz: maybe change this comment, something like `TODO: remove .toInt once starr is includes the fix for SI-9516 (2.12.0-M5)`

sjrd: OK Done.

lrytz:
LGTM!

som-snytt:
I don't think I was paying attention in the related fix. My comment here is that the test could be junit.

sjrd:
@som-snytt As I explained on https://github.com/scala/scala-dev/issues/136, If I make it a JUnit test, it won't succeed if run from sbt. In other words, it would _break_ `sbt test` until the next starr is released.

som-snytt:
@sjrd Thanks for the explanation! Most build issues are boring, but this is interesting --.

id-ilych:
I don't want to seem rude but I wonder how this breaking change wasn't mentioned in [2.12 release notes](http://www.scala-lang.org/news/2.12.0/)?

Does complete list of behavior changes (i.e. same code compiles without a warning but gives different result) exist somewhere? Without it it seems impossible to upgrade 2.11 -> 2.12 on existing project w/o 100% tests coverage (including 3rd-party libraries).

Thanks in advance for the answer.

lrytz:
I added the `release-notes` label, you're right that this PR should have it. You can check closed PRs with that label. However, you'd want to narrow down the selection to 2.12 PRs, I assume that's why we have the (kind of redundant - PRs also have a milestone assigned, but there are many 2.12 milestones..) `2.12` label. So your list would be this: https://github.com/scala/scala/pulls?q=label%3Arelease-notes+label%3A2.12+is%3Aclosed

PR #5119, opened 2016-04-23
[Dates fixing]

FelAl:
First part of fixed files.

@Ichoran
@adriaanm

PR #5120, opened 2016-04-23
[Dates fixing]

FelAl:
Second part of fixed files.

@Ichoran
@adriaanm


janekdb:
I suggest this message `[Fix Copyright Dates]` as a standard message to use when this is done.

The benefits are,
-  More specific than 'Dates`
- Uses imperative voice

See
https://github.com/scala/scala#user-content-git-hygiene
https://github.com/erlang/otp/wiki/Writing-good-commit-messages

> Write the summary line and description of what you have done in the imperative mode, that is as if you were commanding someone. Start the line with "Fix", "Add", "Change" instead of "Fixed", "Added", "Changed".

FelAl:
Thank you for comment.
Done.

FelAl:
Is there something wrong with this PR?

Or maybe I should to change something?
And there also https://github.com/scala/scala/pull/5119 .

Ichoran:
This PR looks fine, but what about the PR with the tool (alone)?  I would tend to want that one merged first, followed by the two that actually contain the changed files.

FelAl:
Ok

lrytz:
agree, it would be good to have the tool in a separate commit, first.

PR #5122, opened 2016-04-25
SI-9121 test case (fixed in new optimizer), SI-9179 test case

lrytz:


src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
```diff
@@ -76,7 +76,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
     val origOwner = sym.originalOwner
     // phase travel necessary: after flatten, the name includes the name of outer classes.
     // if some outer name contains $anon, a non-anon class is considered anon.
-    if (delambdafyInline() && sym.rawowner.isAnonymousFunction) {
+    if (delambdafyInline() && exitingPickler(sym.rawowner.isAnonymousFunction)) {
```
adriaanm, line 79: Could you add a note on this change to the commit message?

lrytz, line 79: done

lrytz:
review by @szeiger

lrytz:
rebased, as there were merge conflicts. ping @szeiger

test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala
```diff
@@ -1505,4 +1505,25 @@ class InlinerTest extends ClearAfterClass {
...
+        |  implicit class RichFunction1[T1, R](val f: String) extends AnyVal {
+        |    def toRx: Any = ""
+        |  }
+        |}
+        |import p1.Implicits._
+        |class C {
+        |  def t(): Unit = new ScalaObservable("").scMap("")
+        |}
+      """.stripMargin
+    val c :: _ = compile(code)
```
szeiger: Is a single source unit sufficient to reproduce the issue? I thought it only showed up on separate compilation.

lrytz: ah, good point!

lrytz:
should be good now

szeiger:
LGTM

PR #4133, opened 2014-11-14
SI-8976 MutableList.tail.iterator.size is length

som-snytt:
This is fixed by capping the length of iterator.

That is OK because mutating the list by adding to it during
iteration is not allowed.

A test is added that will no doubt be superseded by the QCC tests.
(Quasi-Comprehensive Collections.)

Review by @Ichoran with screws tightened by @retronym

PR #5124, opened 2016-04-25
SI-9522 release key reference when deleting from OpenHashMap

performantdata:
review by @axel22

Apparently this also solves SI-9520 for 2.12.x.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,6 +66,7 @@ val scalaXmlDep = withoutScalaLang("org.scala-lang.modules" %% "scala-xml" % ver
 val partestDep = withoutScalaLang("org.scala-lang.modules" %% "scala-partest" % versionNumber("partest"))
 val junitDep = "junit" % "junit" % "4.11"
 val junitIntefaceDep = "com.novocode" % "junit-interface" % "0.11" % "test"
+val jolDep = "org.openjdk.jol" % "jol-core" % "0.5"
 val asmDep = "org.scala-lang.modules" % "scala-asm" % versionProps("scala-asm.version")
 val jlineDep = "jline" % "jline" % versionProps("jline.version")
 val antDep = "org.apache.ant" % "ant" % "1.9.4"
@@ -544,7 +545,7 @@ lazy val junit = project.in(file("test") / "junit")
   .settings(disablePublishing: _*)
   .settings(
     fork in Test := true,
-    libraryDependencies ++= Seq(junitDep, junitIntefaceDep),
+    libraryDependencies ++= Seq(junitDep, junitIntefaceDep, jolDep),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-v"),
     unmanagedSourceDirectories in Test := List(baseDirectory.value)
   )
--- a/build.xml
+++ b/build.xml
@@ -277,6 +277,7 @@ TODO:
       <property name="junit.version" value="4.12"/>
       <artifact:dependencies pathId="junit.classpath" filesetId="junit.fileset">
         <dependency groupId="junit" artifactId="junit" version="${junit.version}"/>
+        <dependency groupId="org.openjdk.jol" artifactId="jol-core" version="0.5"/>
       </artifact:dependencies>
       <copy-deps project="junit"/>
 
--- a/src/library/scala/collection/mutable/OpenHashMap.scala
+++ b/src/library/scala/collection/mutable/OpenHashMap.scala
@@ -21,10 +21,16 @@ object OpenHashMap {
   def apply[K, V](elems : (K, V)*) = new OpenHashMap[K, V] ++= elems
   def empty[K, V] = new OpenHashMap[K, V]
 
-  final private class OpenEntry[Key, Value](val key: Key,
-                                            val hash: Int,
+  /** A hash table entry.
+    * 
+    * The entry is occupied if and only if its `value` is a `Some`;
+    * deleted if and only if its `value` is `None`.
+    * If its `key` is not the default value of type `Key`, the entry is occupied.
+    * If the entry is occupied, `hash` contains the hash value of `key`.
+    */
+  final private class OpenEntry[Key, Value](var key: Key,
+                                            var hash: Int,
                                             var value: Option[Value])
-                extends HashEntry[Key, OpenEntry[Key, Value]]
 
   private[mutable] def nextPositivePowerOfTwo(i : Int) = 1 << (32 - Integer.numberOfLeadingZeros(i - 1))
 }
@@ -64,7 +70,14 @@ extends AbstractMap[Key, Value]
   private[this] val actualInitialSize = OpenHashMap.nextPositivePowerOfTwo(initialSize)
 
   private var mask = actualInitialSize - 1
-  private var table : Array[Entry] = new Array[Entry](actualInitialSize)
+
+  /** The hash table.
+    * 
+    * The table's entries are initialized to `null`, indication of an empty slot.
+    * A slot is either deleted or occupied if and only if the entry is non-`null`.
+    */
+  private[this] var table = new Array[Entry](actualInitialSize)
+
   private var _size = 0
   private var deleted = 0
 
@@ -91,42 +104,43 @@ extends AbstractMap[Key, Value]
     table = new Array[Entry](newSize)
     mask = newSize - 1
     oldTable.foreach( entry =>
-      if (entry != null && entry.value != None) addEntry(entry))
+      if (entry != null && entry.value != None)
+        table(findIndex(entry.key, entry.hash)) = entry )
     deleted = 0
   }
 
   /** Return the index of the first slot in the hash table (in probe order)
-    * that either is empty, or is or was last occupied by the given key.
-    */
-  private[this] def findIndex(key: Key) : Int = findIndex(key, hashOf(key))
-
-  /** Return the index of the first slot in the hash table (in probe order)
-    * that either is empty, or is or was last occupied by the given key.
-    * 
-    * This method is an optimization for when the hash value is in hand.
+    * that is, in order of preference, either occupied by the given key, deleted, or empty.
     * 
     * @param hash hash value for `key`
     */
   private[this] def findIndex(key: Key, hash: Int): Int = {
     var j = hash
-
     var index = hash & mask
     var perturb = index
-    while(table(index) != null &&
-          !(table(index).hash == hash &&
-            table(index).key == key)){
+
+    /** Index of the first slot containing a deleted entry, or -1 if none found yet. */
+    var firstDeletedIndex = -1
+
+    var entry = table(index)
+    while (entry != null) {
+      if (entry.hash == hash && entry.key == key && entry.value != None)
+        return index
+
+      if (firstDeletedIndex == -1 && entry.value == None)
+        firstDeletedIndex = index
+
       j = 5 * j + 1 + perturb
       perturb >>= 5
       index = j & mask
+      entry = table(index)
     }
-    index
+
+    if (firstDeletedIndex == -1) index else firstDeletedIndex
   }
 
-  private[this] def addEntry(entry: Entry) =
-    if (entry != null) table(findIndex(entry.key, entry.hash)) = entry
-
   override def update(key: Key, value: Value) {
-    put(key, hashOf(key), value)
+    put(key, value)
   }
 
   @deprecatedOverriding("+= should not be overridden in order to maintain consistency with put.", "2.11.0")
@@ -150,6 +164,8 @@ extends AbstractMap[Key, Value]
     } else {
       val res = entry.value
       if (entry.value == None) {
+        entry.key = key
+        entry.hash = hash
         size += 1
         deleted -= 1
         modCount += 1
@@ -159,13 +175,22 @@ extends AbstractMap[Key, Value]
     }
   }
 
+  /** Delete the hash table slot contained in the given entry. */
+  @inline
+  private[this] def deleteSlot(entry: Entry) = {
+    entry.key = null.asInstanceOf[Key]
+    entry.hash = 0
+    entry.value = None
+
+    size -= 1
+    deleted += 1
+  }
+
   override def remove(key : Key): Option[Value] = {
-    val index = findIndex(key)
-    if (table(index) != null && table(index).value != None){
-      val res = table(index).value
-      table(index).value = None
-      size -= 1
-      deleted += 1
+    val entry = table(findIndex(key, hashOf(key)))
+    if (entry != null && entry.value != None) {
+      val res = entry.value
+      deleteSlot(entry)
       res
     } else None
   }
@@ -249,7 +274,7 @@ extends AbstractMap[Key, Value]
   }
 
   override def retain(f : (Key, Value) => Boolean) = {
-    foreachUndeletedEntry(entry => if (!f(entry.key, entry.value.get)) {entry.value = None; size -= 1; deleted += 1} )
+    foreachUndeletedEntry(entry => if (!f(entry.key, entry.value.get)) deleteSlot(entry))
     this
   }
 
--- a/test/junit/scala/collection/mutable/OpenHashMapTest.scala
+++ b/test/junit/scala/collection/mutable/OpenHashMapTest.scala
@@ -4,6 +4,10 @@ import org.junit.Test
 import org.junit.Assert._
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
+import org.openjdk.jol.info.GraphLayout
+import org.openjdk.jol.info.GraphWalker
+import org.openjdk.jol.info.GraphVisitor
+import org.openjdk.jol.info.GraphPathRecord
 
 /** Tests for [[OpenHashMap]]. */
 @RunWith(classOf[JUnit4])
@@ -28,7 +32,13 @@ class OpenHashMapTest {
     val fieldMirror = mirror.reflect(m).reflectField(termSym)
 		*/
     // Use Java reflection instead for now.
-    val field = m.getClass.getDeclaredField("deleted")
+    val field =
+      try {  // Name may or not be mangled, depending on what the compiler authors are doing.
+        m.getClass.getDeclaredField("scala$collection$mutable$OpenHashMap$$deleted")
+      } catch {
+        case _: NoSuchFieldException =>
+          m.getClass.getDeclaredField("deleted")
+      }
     field.setAccessible(true)
 
     m.put(0, 0)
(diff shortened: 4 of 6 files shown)


src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -91,42 +104,43 @@ extends AbstractMap[Key, Value]
     table = new Array[Entry](newSize)
     mask = newSize - 1
     oldTable.foreach( entry =>
```
axel22, line 106: Not related to this PR, but it might be a good idea to replace this with a `while` eventually.

src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -64,7 +70,14 @@ extends AbstractMap[Key, Value]
   private[this] val actualInitialSize = OpenHashMap.nextPositivePowerOfTwo(initialSize)
 
   private var mask = actualInitialSize - 1
-  private var table : Array[Entry] = new Array[Entry](actualInitialSize)
+
+  /** The hash table.
+    * 
+    * The table's entries are initialized to `null`, indication of an empty slot.
+    * A slot is either deleted or occupied if and only if the entry is non-`null`.
```
axel22, line 77: Maybe better: "A slot is either deleted or occupied. It is occupied if and only if the entry is non-`null` and entry's key is non-`null`."

src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -91,42 +104,43 @@ extends AbstractMap[Key, Value]
     table = new Array[Entry](newSize)
     mask = newSize - 1
     oldTable.foreach( entry =>
-      if (entry != null && entry.value != None) addEntry(entry))
+      if (entry != null && entry.value != None)
```
axel22, line 107: Wouldn't it be a good idea to now check if the `entry.key` is `null`, and skip entries whose keys are `null`. 

Alternatively, if it makes sense to reuse those objects, one could just insert them anywhere.

src/library/scala/collection/mutable/OpenHashMap.scala
performantdata, line 77: > "A slot is either deleted or occupied."

That's not true. A slot can also be empty.

> "It is occupied if and only if [...] entry's keys is non-`null`."

I try to define the distinction between the deleted and occupied states close to the data that distinguishes them, i.e., in the `OpenEntry` docs. Also, it's not clear to me whether this class intends to support `null` keys.

src/library/scala/collection/mutable/OpenHashMap.scala
performantdata, line 106: Sadly. Though I haven't looked at the disassembled code with 2.12.

This call would also benefit from a streamlined `findIndex()`. Ditto in `remove()`.

src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -91,42 +104,43 @@ extends AbstractMap[Key, Value]
...
     var perturb = index
-    while(table(index) != null &&
-          !(table(index).hash == hash &&
-            table(index).key == key)){
+
+    /** Index of the first slot containing a deleted entry, or -1 if none found yet. */
+    var firstDeletedIndex = -1
+
+    var entry = table(index)
+    while (entry != null) {
```
axel22, line 126: In `FlatHashMap`, the strategy was to, after deletion, move the clashing keys into the hole when possible. There, the collision resolution scheme was linear, so it was easy to find out how much a key can be moved.

Here I'm not sure that's possible, so this seem like an ok solution.

src/library/scala/collection/mutable/OpenHashMap.scala
performantdata, line 107: Same question about whether a `null` key is valid.

axel22:
You should add a test case that stresses the respective scenario - remove many elements from a full hash map, and then reinsert the same keys, or different keys with the same hashes.

In addition, you should create a performance benchmark that tests the speed of lookups and inserts after a lot of removes. Potentially long chains of `null`-key entries could hurt performance.

src/library/scala/collection/mutable/OpenHashMap.scala
axel22, line 77: I see, I missed the distinction between and empty and deleted.

And that's a good point - if it is not specified, and the current implementation allows it, then `null` keys should still be allowed.

src/library/scala/collection/mutable/OpenHashMap.scala
axel22, line 107: I would say yes, in which case it might still make sense to get rid of some entries that have both `null` keys and `None` values.

src/library/scala/collection/mutable/OpenHashMap.scala
axel22, line 106: @lrytz might be able to comment on that.

performantdata:
Oh right, forgot the test code. Actually, what needs testing for the SI-9522 bug is whether the memory leak is gone. That could be tested by adding an entry, removing it, and seeing how much memory is used by the `OpenHashMap` with JOL.

> remove many elements from a full hash map, and then reinsert the same keys, or different keys with the same hashes

What's a "full hash map"? :smiley: I guess you mean one that hasn't had elements removed? I'm not sure what good this would do, what it tests. Reinserting the same keys in the old code reused the key's entry; in the new code, it reuses the first deleted slot (if any), irrespective of the key. Using different keys with the same hashes is equivalent to using different keys with different hashes; the hash check is merely a performance enhancement. Seems that either case would be best tested with a benchmark, not a unit test case.

If you want to help with the benchmarking, #5061 is waiting for review.

> Potentially long chains of `null`-key entries could hurt performance.

I don't see what that means in this context. It's well known that deleted entries do hurt performance in open addressing. But this PR should reduce the number of deleted entries, by reusing them more readily. So I'd like to see the performance numbers, too. I've seen [awful performance for `put()` with `remove()`](https://issues.scala-lang.org/browse/SI-9732).

src/library/scala/collection/mutable/OpenHashMap.scala
performantdata, line 107: But I do eliminate entries that have `value == None`, which, of course, eliminates the ones that also have `key == null.asInstanceOf[Key]`.

src/library/scala/collection/mutable/OpenHashMap.scala
lrytz, line 106: So I was going to take a quick look this morning, turns out it took a bit more time to figure out what's going on :) Here's my example:

``` scala
class C {
  val table: Array[String] = null
  def t(): Unit = {
    table.foreach(addEntry)
  }
  def addEntry(s: String): Unit = ()
}
```

Basically the optimizer is supposed to inline foreach (the implementation is [inherited from `IndexedSeqOptimized`](https://github.com/scala/scala/blob/93f209dd65d5c05fc2cb61916a850940499c9261/src/library/scala/collection/IndexedSeqOptimized.scala#L28)), and the closure should be eliminated. But looking at the bytecode I saw something completely different, basically this:

```
        Function1 foreach_f = s -> {
            this.C$$$anonfun$1(s);
            return BoxedUnit.UNIT;
        };
        new ArrayOps.ofRef(Predef..MODULE$._refArrayOps((Object[])this.table())).iterator().foreach(foreach_f);
```

What the..? Turns out the inliner picked the wrong implementation and inserted the overridden [code from `IterableLike`](https://github.com/scala/scala/blob/93f209dd65d5c05fc2cb61916a850940499c9261/src/library/scala/collection/IterableLike.scala#L70). Here's the ticket: https://github.com/scala/scala-dev/issues/140.

But basically the 2.12.0 optimizer should be able to handle `Array.foreach` well. `Array.map` is a different story and will require much more sophisticated heuristics and stack allocation. I'll continue working on those for later 2.12 minor releases, but they won't be in 2.12.0.

src/library/scala/collection/mutable/OpenHashMap.scala
axel22, line 107: Yep, disregard that comment, was late :)

axel22:
By "full" I mean, the one that had previously a lot of elements inserted. You want to create test that ensures that your changed code gets heavily run. Or identify an existing test that does that, if there is one.

axel22:
> I don't see what that means in this context.

What you write below that is what it means - we should have performance numbers ;)

src/library/scala/collection/mutable/OpenHashMap.scala
axel22, line 106: Thanks a lot for taking a look, Lukas!

lrytz, line 106: Actually, 2.12 won't be able to fully optimize `Array.foreach`: even after inlining the implementation and eliminating the closure, every value is boxed and unboxed before it's passed to the consumer. Logged here: https://github.com/scala/scala-dev/issues/27#issuecomment-216119276

lrytz, line 106: So yes, I'd also recommend a while loop here (for now..)

src/library/scala/collection/mutable/OpenHashMap.scala
lrytz, line 77: > it's not clear to me whether this class intends to support null keys

the situation for `null` seems to be not clearly specified. for example, in https://issues.scala-lang.org/browse/SI-2935, martin says:

> we can't give a clear guideline. I would rather leave the spec of maps underspecified. Some maps will accept nulls but you should not rely on it for all maps

since `OpenHashMap` currently does support `null` keys, we cannot just change this behavior. is there a good reason to change it?

lrytz:
@performantdata what are your thoughts on @axel22's idea to fill up the gap of a `remove` operation directly with the next value for the same hash, and so on until no gap is left?
- `remove` would be slower (bubble-up the gap)
- subsequent `get` operations would be faster (no gaps)

i think this would even allow us to get rid of the `Option` in `OpenEntry` and just have `var value: Value`, which sounds like a big plus.

src/library/scala/collection/mutable/OpenHashMap.scala
performantdata, line 77: I don't see one. But I'm not changing it, either.

performantdata:
@lrytz:

> what are your thoughts on [Alex]'s idea to fill up the gap of a remove operation directly with the next value for the same hash, and so on until no gap is left?

You mean where he writes:

> In `FlatHashMap`, the strategy was to, after deletion, move the clashing keys into the hole when possible. There, the collision resolution scheme was linear, so it was easy to find out how much a key can be moved.

...? I'm not even sure that that's a safe algorithm (moving a key into the hole), but I can't tell from that description. In any case, as I wrote:

> Reinserting the same keys in the old code reused the key's entry; in the new code, it reuses the first deleted slot (if any), irrespective of the key.

You can see this in the way `findIndex` uses `firstDeletedIndex`. This is what you're asking for, no? I think the new code will slightly slow down all the operations that use `findIndex`.

> i think this would even allow us to get rid of the `Option` in `OpenEntry`

#5074 (as you saw). I wanted to push that through 2.11 first, because we can.

Incidentally, you might find the JMH test code there interesting because it doesn't optimize away the `Some` coming back from `getOrElse`. (I think. Haven't looked at it in awhile. Bug me there if that doesn't make sense.)

lrytz:
> You can see this in the way findIndex uses firstDeletedIndex. This is what you're asking for, no?

No, what i have in mind is the following. Assume values `a`, `b`, `c` and `d` have the same hash. Assume that we enter these values in that order and they are stored at indices `1`, `4` and `7`, `8`. Now if we remove `b` from the map, the current implementation marks the slot `4` as deleted by setting `entry.value = None`. The proposal is to extend the `remove` operation to move `c` to index `4`, and then subsequently `d` to `7`, and finally setting `table(8) = null`. This way there would never be deleted values in the array, so there's no need for using an `Option`.

performantdata:
It won't work. The problem is that the slots that are used by this one probe sequence are also used by every other probe sequence. When you set `table(8) = null`, you prematurely terminate all of those sequences that would otherwise pass through slot 8.

lrytz:
ok got it, thanks! so then i don't have anything to add to what Alex mentioned above: it would be good to have some numbers for the case where the change in this PR can cause a slowdown. And also replacing the `Array.foreach` by a while loop.

performantdata:
> Oh right, forgot the test code. Actually, what needs testing for the SI-9522 bug is whether the memory leak is gone. That could be tested by adding an entry, removing it, and seeing how much memory is used by the `OpenHashMap` with JOL.

Added this (and rebased).

performantdata:
I attached my usual performance graphs to the JIRA ticket.
@axel22, back to you for review.

src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -160,10 +176,12 @@ extends AbstractMap[Key, Value]
...
 
   override def remove(key : Key): Option[Value] = {
-    val index = findIndex(key)
-    if (table(index) != null && table(index).value != None){
-      val res = table(index).value
-      table(index).value = None
+    val entry = table(findIndex(key, hashOf(key)))
+    if (entry != null && entry.value != None) {
+      val res = entry.value
+      entry.key = null.asInstanceOf[Key]
```
performantdata: I need to release the `key` in `retain()` as well.

test/junit/scala/collection/mutable/OpenHashMapTest.scala
```diff
@@ -39,4 +43,36 @@ class OpenHashMapTest {
...
+    }
+
+    val m = OpenHashMap.empty[MyClass, Int]
+    val obj = new MyClass
+    assertEquals("Found a key instance in the map before adding one!?", 0, counter.countInstances(m))
+    m.put(obj, 0)
+    assertEquals("There should be only one key instance in the map.", 1, counter.countInstances(m))
+    m.remove(obj)
+    assertEquals("There should be no key instance in the map.", 0, counter.countInstances(m))
+  }
```
axel22, line 97: Test looks good, but I recommend that you add a similar test in which some larger number of keys (e.g. 500) is inserted and then removed. The current test case does not cover the situation in which there are key collisions.

performantdata:
> You want to create test that ensures that your changed code gets heavily run.

I don't necessarily. I want a test that exercises a case with an expected outcome. If that somehow requires running the code "heavily", then so be it.

> The current test case does not cover the situation in which there are key collisions.

What is a "key collision"? It could mean adding two identical keys, but I think you mean adding two keys with the same hash: a key hash collision. The former is worth considering, and for that I can write:

```
m.put(obj, 0)
assertEquals("There should be only one key instance in the map.", 1,...
m.put(obj, 1)
assertEquals("There should still be only one key instance in the map.", 1,...
```

For the latter, I should be able to override `MyClass.hashCode()` to return a constant, then `put` two different instances. But I don't see why adding 500 would make a difference. Perhaps you're thinking to cause a hash collision by sheer quantity, but forcing it to happen is easier.

axel22:
By key collision I mean keys whose hash-codes collide. You can do either by adding and removing a lot of elements, or by forcing it to happen, your choice.

performantdata:
Amended and rebased.

performantdata:
> Adding a large number of keys with the same hash isn't even safe to do, because the original author's probe algorithm is flawed.

Addressing this [here](https://issues.scala-lang.org/browse/SI-9789).

axel22:
LGTM

build.xml
```diff
@@ -277,6 +277,7 @@ TODO:
       <property name="junit.version" value="4.12"/>
       <artifact:dependencies pathId="junit.classpath" filesetId="junit.fileset">
         <dependency groupId="junit" artifactId="junit" version="${junit.version}"/>
+        <dependency groupId="org.openjdk.jol" artifactId="jol-core" version="0.5"/>
```
lrytz, line 280: you'll probably need to add this to `build.sbt` too - did you try running `sbt test`? 

also cc @adriaanm for adding a (test) dependency to scala.

test/junit/scala/collection/mutable/OpenHashMapTest.scala
```diff
@@ -39,4 +49,50 @@ class OpenHashMapTest {
...
+        val walker = new GraphWalker(obj)
+        walker.addVisitor(this)
+        walker.walk
+        instanceCount
+      }
+
+      override def visit(record: GraphPathRecord) {
+        if (record.klass() == classOf[MyClass])  instanceCount += 1
+      }
+    }
```
lrytz, line 76: I was going to suggest to put a utility for this in the `scala.tools.testing` package (in test/junit), then I found the existing method [`assertNotReachable` in `scala.tools.testing.AssertUtil`](https://github.com/scala/scala/blob/95f57600a7498f269d32743788c89d4d65f93700/test/junit/scala/tools/testing/AssertUtil.scala#L68). Its internal method `refs` collects all objects reachable from a certain root, so you could factor this out and use it (or a new utility) in this test. It would be really nice if we could avoid the new dependency on org.openjdk.jol.

performantdata, line 76: @retronym why did you also want JOL in the classpath?

build.xml
adriaanm, line 280: without looking too deeply, I'm okay with another test dependency (since it's non-scala and thus we avoid a cyclic dependency/bootstrapping complication), but we should update the sbt build as well if we end up adding it

adriaanm:
Adding the WIP label until we sort out the test dependency thingy, to avoid merging triggered by the reviewed label.

performantdata:
> you'll probably need to add this to `build.sbt` too

done

test/junit/scala/collection/mutable/OpenHashMapTest.scala
retronym, line 76: > why did you also want JOL in the classpath?

During previous profiling work, I spent some time looking for ways to shrink the footprint of heavily used classes in the compiler internals. Having JOL around would let us write tests to avoid accidentally increasing the size of something that had been performance tuned.

As shown here, it is also handy for writing tests about the object graph, retention, etc. Replacing the home brew version `assertNotReachable` with a dependency from Oracle makes sense to me.

`assertNotReachable` was my suggestion in https://github.com/scala/scala/pull/4133#issuecomment-65720629 when discussing how to write a test case for a similar bug.

Here are the [JOL samples](http://hg.openjdk.java.net/code-tools/jol/file/tip/jol-samples/src/main/java/org/openjdk/jol/samples/) that show what else it can do. A lot of it is interesting mainly as a way to understand how JVM abstractions are implemented in HotSpot.

lrytz, line 76: OK, sounds good.

build.sbt
```diff
@@ -544,7 +544,7 @@ lazy val junit = project.in(file("test") / "junit")
   .settings(disablePublishing: _*)
   .settings(
     fork in Test := true,
-    libraryDependencies ++= Seq(junitDep, junitIntefaceDep),
+    libraryDependencies ++= Seq(junitDep, junitIntefaceDep, "org.openjdk.jol" % "jol-core" % "0.5"),
```
lrytz: can you extract this into a local at https://github.com/scala/scala/blob/2.12.x/build.sbt#L63-L72 to keep the dependencies together?

lrytz:
@performantdata did you check if `sbt test` passes locally? (we don't currently run that on CI, test are just executed through ant)

performantdata:
> did you check if `sbt test` passes locally?

yes

build.sbt
performantdata: yeah, sure

lrytz:
thanks!

performantdata:
Thanks Alex, Lukas!

PR #5125, opened 2016-04-25
SI-9752 never ignore blank lines when parsing code blocks

gourlaysama:
The default behavior when parsing the content of a tag text (like after
`@example`) was to ignore empty lines. That's fine, except when we are
in the middle of a code block, where preserving formatting matters.


gourlaysama:
review @VladUreche @janekdb

lrytz:
LGTM

PR #5126, opened 2016-04-26
WIP: no more type slack for function literal [ci: last-only]

adriaanm:
see scala/scala-dev#138


build.sbt
```diff
@@ -128,6 +128,7 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   // we always assume that Java classes are standalone and do not have any dependency
   // on Scala classes
   compileOrder := CompileOrder.JavaThenScala,
+  incOptions := incOptions.value.withNameHashing(false).withAntStyle(true), // until we have class-based name hashing
```
retronym: It think this is a poor default for new contributors, instead it should be something we enable via `local.sbt`. IIRC, I couldn't figure out an easy way to modify that setting across all projects via `local.sbt`, but with a little persistence and help it should be doable.

adriaanm: Agreed, I should've specified this commit was not meant for the PR, but I need it locally now that I no longer have `ant` on my machine :)

lrytz: you could put it in a different file (`local.sbt`) and add that filename to your global `~/.gitignore`. also for me, it only worked with `in ThisBuild`, so the full line would be

```
(incOptions in ThisBuild) := (incOptions in ThisBuild).value.withNameHashing(false).withAntStyle(true)
```

adriaanm:
One new failure, hmmm.. The community build didn't sprout any new failures, though: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/381/consoleFull

SethTisue:
should we close this just to get it out of the queue?

PR #5127, opened 2016-04-26
BitSet{1,2} conversion

ruslansennov:
In current implementation when we set high word (elems1) of BitSet2 to zero, the result is BitSet2 again. I believe it is leading to excessive memory usage and result should be BitSet1. Private helper method createSmall(a: Long, b: Long) introduced.
I'm not sure, but current tests should be enough

### DIFF
--- a/src/library/scala/collection/immutable/BitSet.scala
+++ b/src/library/scala/collection/immutable/BitSet.scala
@@ -68,6 +68,8 @@ object BitSet extends BitSetFactory[BitSet] {
   /** The empty bitset */
   val empty: BitSet = new BitSet1(0L)
 
+  private def createSmall(a: Long, b: Long): BitSet = if (b == 0L) new BitSet1(a) else new BitSet2(a, b)
+
   /** A builder that takes advantage of mutable BitSets. */
   def newBuilder: Builder[Int, BitSet] = new Builder[Int, BitSet] {
     private[this] val b = new mutable.BitSet
@@ -84,7 +86,7 @@ object BitSet extends BitSetFactory[BitSet] {
     val len = elems.length
     if (len == 0) empty
     else if (len == 1) new BitSet1(elems(0))
-    else if (len == 2) new BitSet2(elems(0), elems(1))
+    else if (len == 2) createSmall(elems(0), elems(1))
     else {
       val a = new Array[Long](len)
       Array.copy(elems, 0, a, 0, len)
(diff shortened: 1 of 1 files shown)


SethTisue:
can you improve the commit message to explain what this is about?

should there be test coverage?

ruslansennov:
done

lrytz:
the commit message (https://github.com/scala/scala/pull/5127/commits/f3fb45be6afba23e53cfb46ffdabc1613b35c8e3) still looks empty.

there are a few more places that could be covered by this guard (`fromBitMask`, `fromBitMaskNoCopy`, `BitSet2.tail`). how about adding a private helper method to object BitSet?

``` scala
private def createSmall(a: Long, b: Long): BitSet = if (b == 0L) new BitSet1(a) else new BitSet2(a, b)
```

ruslansennov:
@lrytz thanks, done

src/library/scala/collection/immutable/BitSet.scala
```diff
@@ -137,7 +139,7 @@ object BitSet extends BitSetFactory[BitSet] {
         if (elems1 == 0L) throw new NoSuchElementException("Empty BitSet")
         new BitSet2(elems0, elems1 - java.lang.Long.lowestOneBit(elems1))
```
lrytz: i think this one should use `createSmall`, as `elems1 - java.lang.Long.lowestOneBit(elems1)` may be `0L`.

src/library/scala/collection/immutable/BitSet.scala
```diff
@@ -137,7 +139,7 @@ object BitSet extends BitSetFactory[BitSet] {
...
       }
-      else new BitSet2(elems0 - java.lang.Long.lowestOneBit(elems0), elems1)
+      else createSmall(elems0 - java.lang.Long.lowestOneBit(elems0), elems1)
```
lrytz: here on the other hand, `elems1` should always be non-zero, so we could keep `new BitSet2`

lrytz:
@ruslansennov could you squash the two commits into one?

ruslansennov:
@lrytz I combined all into the one commit

lrytz:
sorry to bother you once more.. the [commit message](https://github.com/scala/scala/pull/5127/commits) (note that this is different from the main description in the PR) is actually still very short - could you include an explanation of the change?

also, in the PR description, you say "The same for out of bounds of BitSet1." - what do you mean by that?

thanks..!

ruslansennov:
@lrytz OK, now I see that every commit has not only the subject but also the short description. Both commit message and PR description fixed

ruslansennov:
Something wrong with validation. The only difference with the previous commit (which was successful) is commit message

lrytz:
LGTM

fommil:
This causes a regression in ensime because we were using the BitSet representation of a BigInt as a way of encoding it on the wire and this flips the meaning of the byte array. See https://issues.scala-lang.org/browse/SI-10162

PR #5128, opened 2016-04-26
Publish usable Scala distributions to Ivy repositories

szeiger:
- sbt requires a `default` configuration in the Scala distribution but
  doesn’t publish such a configuration to Ivy repositories by default.
  This is not a problem when publishing to a Maven repository because
  Maven doesn’t use the concept of configurations and Ivy creates a
  standard set (including `default`) when resolving artifacts from Maven
  repositories, but it prevents the use of any Scala distribution
  published with `publishLocal`.
  
  The underlying issue is that sbt requires `default` instead of
  `default(compile)`. We work around this limitation by publishing a
  dummy `default` configuration.
- sbt has hardcoded dependencies on the `scala-library` artifact of type
  `jar`. In the tradition of `sbt-osgi` we used type `bundle` when
  publishing via Ivy but this makes the artifacts unusable from sbt. We
  now publish the OSGi bundles directly as type `jar` (which is
  compatible with how they appear in Ivy after resolving from a Maven
  repository).
- We have to be more aggressive about not publishing certain
  subprojects, otherwise `ivy.xml` files could still be published even
  when using `publishArtifacts := false`.
- `removePomDependencies` now also modifies `ivy.xml` in addition to
  the Maven POM so that bogus dependencies do not leak into the Ivy
  descriptors.

Fixes https://github.com/scala/scala-dev/issues/130

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -88,7 +88,7 @@ lazy val publishSettings : Seq[Setting[_]] = Seq(
     val mappings = artifacts.toSeq.map { case (a, f) =>
       val typeSuffix = a.`type` match {
         case "pom" => "-pom.xml"
-        case "bundle" | "jar" => ".jar"
+        case "jar" => ".jar"
         case "doc" => "-docs.jar"
         case tpe => s"-$tpe.${a.extension}"
       }
@@ -103,6 +103,8 @@ lazy val publishSettings : Seq[Setting[_]] = Seq(
     if (file.exists) List(Credentials(file))
     else Nil
   },
+  // Add a "default" Ivy configuration because sbt expects the Scala distribution to have one:
+  ivyConfigurations += Configuration("default", "Default", true, List(Configurations.Runtime), true),
   publishMavenStyle := true
 )
 
@@ -240,8 +242,8 @@ def fixPom(extra: (String, scala.xml.Node)*): Setting[_] = {
   ) ++ extra) }
 }
 
-/** Remove unwanted dependencies from the POM. */
-def removePomDependencies(deps: (String, String)*): Setting[_] = {
+/** Remove unwanted dependencies from the POM and ivy.xml. */
+def removePomDependencies(deps: (String, String)*): Seq[Setting[_]] = Seq(
   pomPostProcess := { n =>
     val n2 = pomPostProcess.value.apply(n)
     import scala.xml._
@@ -256,12 +258,38 @@ def removePomDependencies(deps: (String, String)*): Setting[_] = {
         case n => Seq(n)
       }
     })).transform(Seq(n2)).head
+  },
+  deliverLocal := {
+    import scala.xml._
+    import scala.xml.transform._
+    val f = deliverLocal.value
+    val e = (new RuleTransformer(new RewriteRule {
+      override def transform(node: Node) = node match {
+        case e: Elem if e.label == "dependency" && {
+          val org = e.attribute("org").getOrElse("").toString
+          val name = e.attribute("name").getOrElse("").toString
+          deps.exists { case (g, a) =>
+             org == g && (name == a || name == (a + "_" + scalaBinaryVersion.value))
+          }
+        } => Seq.empty
+        case n => Seq(n)
+      }
+    })).transform(Seq(XML.loadFile(f))).head
+    XML.save(f.getAbsolutePath, e, xmlDecl = true)
+    f
   }
-}
+)
 
 val disableDocs = Seq[Setting[_]](
   sources in (Compile, doc) := Seq.empty,
   publishArtifact in (Compile, packageDoc) := false
+)
+
+val disablePublishing = Seq[Setting[_]](
+  publishArtifact := false,
+  // The above is enough for Maven repos but it doesn't prevent publishing of ivy.xml files
+  publish := {},
+  publishLocal := {}
 )
 
 lazy val setJarLocation: Setting[_] =
@@ -325,10 +353,10 @@ lazy val library = configureAsSubproject(project)
       "/project/name" -> <name>Scala Library</name>,
       "/project/description" -> <description>Standard library for the Scala Programming Language</description>,
       "/project/packaging" -> <packaging>jar</packaging>
-    ),
-    // Remove the dependency on "forkjoin" from the POM because it is included in the JAR:
-    removePomDependencies(("org.scala-lang", "forkjoin"))
+    )
   )
+  // Remove the dependency on "forkjoin" from the POM because it is included in the JAR:
+  .settings(removePomDependencies(("org.scala-lang", "forkjoin")): _*)
   .settings(filterDocSources("*.scala" -- (regexFileFilter(".*/runtime/.*\\$\\.scala") ||
                                            regexFileFilter(".*/runtime/ScalaRunTime\\.scala") ||
                                            regexFileFilter(".*/runtime/StringAdd\\.scala"))): _*)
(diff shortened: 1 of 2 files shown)


milessabin:
Oh boy! I think you've just save me a lot of time! :-)

dwijnand:
Does publishM2 work?

dwijnand:
Oh.. it says so in https://github.com/scala/scala-dev/issues/130 ..

szeiger:
`validate-main` error is due to failing IDE tests (apparently unrelated), the `scala-2.12.x-*` jobs were started in error (this PR targets 2.11, not 2.12), so I think the build can be considered successful.

milessabin:
I've forward ported this to 2.12.x and can confirm that it allows locally published binaries to be used by another project, both with "org.scala-lang" as the `scalaOrganization` and also with an alternative.

Along with the SBT build itself, this represents a huge improvement in the experience for people working on the compiler.

Would you like me to submit a PR for the forward port to 2.12.x?

szeiger:
@milessabin I tested this with 2.12 first before switching to 2.11. I expect it to be mergeable with minimal or no changes, so I think this can go through the regular merge process without requiring a separate PR.

milessabin:
Any ETA on that? I'd love to be able to demo it with a pristine 2.12.x tree on Tuesday :-)

SethTisue:
(the integrate-ide failure would go away if the latest 2.11.x changes were merged, but doing so isn't actually necessary)

build.sbt
```diff
@@ -103,6 +103,7 @@ lazy val publishSettings : Seq[Setting[_]] = Seq(
     if (file.exists) List(Credentials(file))
     else Nil
   },
+  ivyConfigurations += Configuration("default", "Default", true, List(Configurations.Runtime), true),
```
lrytz, line 107: could you add a comment to this line? it's not obvious for an occasional visitor.

lrytz:
i did a diff of `~/.ivy2/local/org.scala-lang/scala-compiler/2.11.9-SNAPSHOT/` before and after. notable changes:
- the `bundles` directory is no longer there - is that expected? (i have no idea what it is for). before it contained three files: `scala-compiler.jar`, `scala-compiler.jar.md5` and `scala-compiler.jar.sha1`.
- the diff in `ivy.xml` is basically because scala-xml re-orders all the attributes.
- `<?xml version="1.0" encoding="UTF-8"?>` is no longer there
- there's a new entry `<conf extends="runtime" description="Default" visibility="public" name="default"/>`
- `removePomDependencies` did its work for `ant` and `scala-asm`
- for various dependencies, the `conf="optional->default(compile)"` attribute changed to `conf="optional-&gt;default(compile)"` - not sure if that's a problem?

could you comment on these?

full diff here: https://gist.github.com/lrytz/66ca48d8d5294b5b4eef560a53a080cc

retronym:
> for various dependencies, the conf="optional->default(compile)" attribute changed to conf="optional-&gt;default(compile)" - not sure if that's a problem?

AFAICT, either variation is valid:

> The safe way is to escape all five characters in attributes, however, the > character needn't be escaped in attributes:

http://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

szeiger:
>    the bundles directory is no longer there - is that expected?

Yes, all JARs are now published as type `jar`, not as type `bundle`.

>    <?xml version="1.0" encoding="UTF-8"?> is no longer there

Could be a problem. There must be a way to have scala-xml generate it again.

>    there's a new entry `<conf extends="runtime" description="Default" visibility="public" name="default"/>`

That's the new default configuration.

> the `conf="optional->default(compile)"` attribute changed to `conf="optional-&gt;default(compile)"` - not sure if that's a problem?

Should be OK. `&gt;` is a predefined entity. For some reason scala-xml likes to emit it, even though there's never a reason for that in XML.

szeiger:
Updated

retronym:
https://scala-ci.typesafe.com/job/scala-2.11.x-validate-publish-core/2566/console

```
++ /usr/local/bin/sbt -sbt-version 0.13.11 'setupPublishCore https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/' publish
[info] Loading global plugins from /home/jenkins/.sbt/0.13/plugins
[info] Loading project definition from /home/jenkins/workspace/scala-2.11.x-validate-publish-core/project
[info] *** Welcome to the sbt build definition for Scala! ***
[info] This build definition has an EXPERIMENTAL status. If you are not
[info] interested in testing or working on the build itself, please use
[info] the Ant build definition for now. Check README.md for more information.
[warn] The `-` command is deprecated in favor of `onFailure` and will be removed in 0.14.0
[error] Expected letter
[error] Expected symbol
[error] Expected '!'
[error] Expected '+'
[error] Expected '++'
[error] Expected 'debug'
[error] Expected 'info'
[error] Expected 'warn'
[error] Expected 'error'
[error] Expected ';'
[error] Expected end of input.
[error] Expected '--'
[error] Expected 'show'
[error] Expected 'all'
[error] Expected '*'
[error] Expected '{'
[error] Expected project ID
[error] Expected configuration
[error] Expected key
[error] Expected '-'
[error] 0.13.11
[error] ^
[info] library/*:sbtVersion
[info]  0.13.11
[info] scaladoc/*:sbtVersion
[info]  0.13.11
[info] scalap/*:sbtVersion
[info]  0.13.11
[info] compiler/*:sbtVersion
[info]  0.13.11
[info] repl-jline/*:sbtVersion
[info]  0.13.11
[info] repl-jline-embedded/*:sbtVersion
[info]  0.13.11
[info] reflect/*:sbtVersion
[info]  0.13.11
[info] actors/*:sbtVersion
[info]  0.13.11
[info] partest-extras/*:sbtVersion
[info]  0.13.11
[info] interactive/*:sbtVersion
[info]  0.13.11
[info] repl/*:sbtVersion
[info]  0.13.11
[info] junit/*:sbtVersion
[info]  0.13.11
[info] library-all/*:sbtVersion
[info]  0.13.11
[info] scala-dist/*:sbtVersion
[info]  0.13.11
[info] forkjoin/*:sbtVersion
[info]  0.13.11
[info] root/*:sbtVersion
[info]  0.13.11
++ mv buildcharacter.properties jenkins.properties
Archiving artifacts
Finished: SUCCESS
```

Seems like `/usr/local/bin/sbt` isn't the version of sbt-extras that we expect.

retronym:
And amusingly, the `-sbt-version 0.13.11 publish` is taken to mean:

```
try {
  0.13.11;
  publish
} catch {
  case _ => sbtVersion
}
```

which means we get a exit code of 0 from the errant SBT invocation!

retronym:
LGTM.

I've been using this locally for a few days (forward ported to 2.12.x as https://github.com/retronym/scala/commit/fef591d714576352a67b08162d47225dd69a1ad6). Seems to work really well.

Been using:

```
(
  export V=d49b02c-SNAPSHOT
  sbt 'set VersionUtil.baseVersionSuffix in Global := "'$V'"' "set publishArtifact in (Compile, packageDoc) in ThisBuild := false" publishLocal && 
  sbt -Dstarr.version=2.12.0-$V 'set every scalaBinaryVersion := "2.12.0-M4"' dist/mkQuick shell
)
```

retronym:
Maybe we scripts or commands to assist people to do this sort of "mini-bootstrap" (as opposed to the full bootstrap in `scripts/jobs/integrate/bootstrap` which includes the modules).

This could:
- `publishLocal`, without docs, with the version number derived from the current git hash, as I did above.
- Overwrite `starr.version` in versions.properties and add `set every scalaBinaryVersion := "2.12.0-M4` in `local.sbt`, so that the next invocation of `sbt` would use the new STARR.

SethTisue:
merging the latest changes from 2.11.x onto this branch should make the integrate-ide failure go away

retronym:
/nothingtoseehere

(overididing IDE failures)

PR #5130, opened 2016-04-27
SD-129 Mark JFunctionN as serializable

retronym:
Before this change, if these were used as the target type of a lambda
in Java source code, the lambda would not be serializable. This is
somewhat suprising when contrasted with the way that Scala lambdas
work in Scala source.

Since we copied these classes over from scala-java8-compat, that
project has opted to add Serializable as a parent to solve this issue.
This commit brings our copy of these interfaces into line with that
change.

Review by @szeiger

Fixes scala/scala-dev#129

### DIFF
--- a/src/library/scala/runtime/java8/JFunction0.java
+++ b/src/library/scala/runtime/java8/JFunction0.java
@@ -6,7 +6,7 @@
 package scala.runtime.java8;
 
 @FunctionalInterface
-public interface JFunction0<R> extends scala.Function0<R> {
+public interface JFunction0<R> extends scala.Function0<R>, java.io.Serializable {
     default void $init$() {
     };
     default void apply$mcV$sp() {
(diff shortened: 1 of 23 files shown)


lrytz:
i'd like to repeat my question on https://github.com/scala/scala-dev/issues/129: what would be the disadvantage in making `scala.FunctionN` serializable instead?

also, aren't the `JFunctionN` interfaces going away now that `scala.FunctionN` are compiled to SAM interfaces with default methods?

retronym:
Marking `FunctionN` serializable would change the serializability of user code that extends this trait. Concievably someone could have logic that checked for the marker, or handled a `NotSerializableException` with some fallback. We would change the behaviour of such code.

I guess we could release note this, but it is a bit of a risk.

szeiger:
LGTM (as in: I don't want to discount the possibility of removing the JFunctions altogether and making all Functions Serializable but for now we should at least make JFunctions serializable again)

PR #5131, opened 2016-04-27
Generate static forwarders for object members in companion interface

retronym:
We used to disable generation of static forwarders when a object had a
trait as a companion, as one could not add methods with bodies to an
interface in JVM 6.

The JVM lifted this restriction to support default methods in interfaces,
so we can lift the restriction on static forwarders, too.

Fixes https://github.com/scala/scala-dev/issues/59

Review by @lrytz 

/cc @ktoso


ktoso:
This is awesome, thanks a lot @retronym (from the entire team btw)! :-)

lrytz:
the commit message talks about `GenASM`, which might confuse some future-us.

src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
```diff
@@ -164,7 +164,7 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
 
       } else {
 
-        val skipStaticForwarders = (claszSymbol.isInterface || settings.noForwarders)
+        val skipStaticForwarders = settings.noForwarders
```
lrytz: inline this boolean?

retronym:
@lrytz sorry about the delay, I've just addressed your comments.

lrytz:
LGTM!

PR #5132, opened 2016-04-27
Avoid function chaining with reverse method

acdenhartog:
`pq.reverse.reverse.reverse.reverse.reverse` could possibly occur in a loop.

(no similar cases found searching rest of the standard library)

### DIFF
--- a/src/library/scala/collection/mutable/PriorityQueue.scala
+++ b/src/library/scala/collection/mutable/PriorityQueue.scala
@@ -200,9 +200,7 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
    *  @return   A reversed priority queue.
    */
   def reverse = {
-    val revq = new PriorityQueue[A]()(new scala.math.Ordering[A] {
-      def compare(x: A, y: A) = ord.compare(y, x)
-    })
+    val revq = new PriorityQueue[A]()(ord.reverse)
     for (i <- 1 until resarr.length) revq += resarr(i)
     revq
   }
--- a/test/junit/scala/collection/mutable/PriorityQueueTest.scala
+++ b/test/junit/scala/collection/mutable/PriorityQueueTest.scala
@@ -14,6 +14,12 @@ class PriorityQueueTest {
   priorityQueue.enqueue(elements :_*)
 
   @Test
+  def orderingReverseReverse() {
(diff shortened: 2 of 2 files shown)


retronym:
Good catch!

LGTM

I also couldn't find any analagous problems, using:

```
 git grep -E 'def (\w+)\((\w+): \w+, (\w+): \w+\) =.*\1\(\3'
src/library/scala/collection/mutable/PriorityQueue.scala:      def compare(x: A, y: A) = ord.compare(y, x)
src/library/scala/math/Ordering.scala:    def compare(x: T, y: T) = outer.compare(y, x)
src/library/scala/math/Ordering.scala:      override def compare(x: Float, y: Float) = outer.compare(y, x)
src/library/scala/math/Ordering.scala:      override def compare(x: Double, y: Double) = outer.compare(y, x)
src/library/scala/math/Ordering.scala:    def compare(x: BigInt, y: BigInt) = x.compare(y)
src/library/scala/math/Ordering.scala:    def compare(x: BigDecimal, y: BigDecimal) = x.compare(y)
src/library/scala/math/PartialOrdering.scala:    def lteq(x: T, y: T) = outer.lteq(y, x)
src/library/scala/math/PartialOrdering.scala:    def tryCompare(x: T, y: T) = outer.tryCompare(y, x)
test/junit/scala/util/SortingTest.scala:    val zs = { val temp = xs.clone; java.util.Arrays.sort(temp, new java.util.Comparator[N] { def compare(a: N, b: N) = a.compare(b) }); temp }
```

lrytz:
@acdenhartog could you add a test to [PriorityQueueTest](https://github.com/scala/scala/blob/2.12.x/test/junit/scala/collection/mutable/PriorityQueueTest.scala)? Something like this:

```
scala> import collection.mutable._
import collection.mutable._

scala> val q = new PriorityQueue[Int]()
q: scala.collection.mutable.PriorityQueue[Int] = PriorityQueue()

scala> q += 1 += 2 += 3
res0: q.type = PriorityQueue(3, 1, 2)

scala> val qq = q.reverse.reverse
qq: scala.collection.mutable.PriorityQueue[Int] = PriorityQueue(3, 1, 2)

scala> q.ord eq qq.ord
res1: Boolean = true
```

retronym:
LGTM

PR #5133, opened 2016-04-27
SD-110 Emit self-synchronized trait methods with ACC_SYNCHRONIZED

retronym:
Avoids the bloat of spelling out the details in the method body.

If we elect to emit trait methods in static methods with an explicit self
parameter (which is under consideration to avoid headaches with using
invokespecial to methods in non-direct parent interfaces), we would
need to revert this change.

Fixes https://github.com/scala/scala-dev/issues/110

Review by @lrytz


retronym:
D'oh.

> Methods of interfaces may have any of the flags in Table 4.6-A set except
> `ACC_PROTECTED`, `ACC_FINAL`, `ACC_SYNCHRONIZED`, and `ACC_NATIVE`

PR #5134, opened 2016-04-27
SD-142 Avoid noisy log output in backend

retronym:
`withCurrentUnit` is designed to be called once per
compilation unit as it side effects by logging and updating
progress counters.

`GenBCode` was calling it more frequently (once per `ClassDef`.)
This is due to the somewhat convoluted internal architecture
of that phase, which is designed to support paralellism in
the future.

This commit factors out the internal part of `withCompilationUnit`
that modifies `currentUnit`, and calls that instead in the loop
over classes.

After this change:

```
% qscala -Ydebug
...
[running phase jvm on <console>] // only once
```

Review by @lrytz


retronym:
Fixes https://github.com/scala/scala-dev/issues/142

lrytz:
LGTM

PR #5135, opened 2016-04-27
Right-bias Either

soc:
- Add operations like map, flatMap which assume right-bias
- Deprecate {Left,Right}Projection
- Deprecate left and right in favor of swap
- Add contains, toOption, toTry, toSeq and filterOrElse
- toSeq returns collection.immutable.Seq instead of collection.Seq
- Don't add get

There are no incompatible changes.
The only possibility of breakage that exists is when people have added
extension methods named map, flatMap etc. to Either in the past doing
something different than the methods added to Either now.

One detail that moved the scales in favor of deprecating LeftProjection
and RightProjection was the desire to have toSeq return
scala.collection.immutable.Seq instead of scala.collection.Seq
like LeftProjection and RightProjection do.
Therefore keeping LeftProjection and RightProjection would introduce
inconsistency.

filter is called filterOrElse because filtering in a for-comprehension
doesn't work if the method needs an explicit argument.

contains was added as safer alternative to
  if (either.isRight && either.right.get == $something) ...

While adding filter with an implicit zero value is possible, it's
dangerous as it would require that developers add a "naked" implicit
value of type A to their scope and it would close the door to a future
in which the Scala standard library ships with Monoid and filter could
exist with an implicit Monoid parameter.

### DIFF
--- a/src/library/scala/util/Either.scala
+++ b/src/library/scala/util/Either.scala
@@ -214,7 +214,7 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
    *  @param elem    the element to test.
    *  @return `true` if the option has an element that is equal (as determined by `==`) to `elem`, `false` otherwise.
    */
-  final def contains[AA >: A](elem: AA): Boolean = this match {
+  final def contains[BB >: B](elem: BB): Boolean = this match {
     case Right(b) => b == elem
     case Left(_)  => false
   }
--- a/src/library/scala/util/Either.scala
+++ b/src/library/scala/util/Either.scala
@@ -32,10 +32,19 @@ package util
  *        Left(in)
  *  }
  *
- *  println( result match {
- *    case Right(x) => "You passed me the Int: " + x + ", which I will increment. " + x + " + 1 = " + (x+1)
- *    case Left(x) => "You passed me the String: " + x
+ *  println(result match {
+ *    case Right(x) => s"You passed me the Int: $x, which I will increment. $x + 1 = ${x+1}"
+ *    case Left(x)  => s"You passed me the String: $x"
  *  })
+ *  }}}
+ *
+ *  Either is right-biased, which means that Right is assumed to be the default case to
+ *  operate on. If it is Left, operations like map, flatMap, ... return the Left value
+ *  unchanged:
+ *
+ *  {{{
+ *  Right(12).map(_ * 2) // Right(24)
+ *  Left(23).map(_ * 2)  // Left(23)
  *  }}}
  *
  *  A ''projection'' can be used to selectively operate on a value of type Either,
@@ -70,11 +79,13 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
   /**
    * Projects this `Either` as a `Left`.
    */
+  @deprecated("use swap instead", "2.12.0")
   def left = Either.LeftProjection(this)
 
   /**
    * Projects this `Either` as a `Right`.
    */
+  @deprecated("Either is now right-biased", "2.12.0")
   def right = Either.RightProjection(this)
 
   /**
@@ -83,8 +94,8 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
    * @example {{{
    * val result: Either[Exception, Value] = possiblyFailingOperation()
    * log(result.fold(
-   *   ex => "Operation failed with " + ex,
-   *   v => "Operation produced value: " + v
+   *   ex => s"Operation failed with $ex",
+   *   v  => s"Operation produced value: $v"
    * ))
    * }}}
    *
@@ -92,9 +103,9 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
    * @param fb the function to apply if this is a `Right`
    * @return the results of applying the function
    */
-  def fold[X](fa: A => X, fb: B => X) = this match {
-    case Left(a) => fa(a)
+  def fold[C](fa: A => C, fb: B => C): C = this match {
     case Right(b) => fb(b)
+    case Left(a)  => fa(a)
   }
 
   /**
@@ -105,8 +116,8 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
    * val r: Either[Int, String] = l.swap // Result: Right("left")
    * }}}
    */
-  def swap = this match {
-    case Left(a) => Right(a)
+  def swap: Either[B, A] = this match {
+    case Left(a)  => Right(a)
     case Right(b) => Left(b)
   }
 
@@ -130,8 +141,9 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
    * This method, and `joinLeft`, are analogous to `Option#flatten`
    */
   def joinRight[A1 >: A, B1 >: B, C](implicit ev: B1 <:< Either[A1, C]): Either[A1, C] = this match {
-    case Left(a)  => Left(a)
     case Right(b) => b
+    case Left(a)  => this.asInstanceOf[Either[A1, C]]
+
   }
 
   /**
@@ -155,7 +167,155 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
    */
   def joinLeft[A1 >: A, B1 >: B, C](implicit ev: A1 <:< Either[C, B1]): Either[C, B1] = this match {
     case Left(a)  => a
-    case Right(b) => Right(b)
+    case Right(b) => this.asInstanceOf[Either[C, B1]]
+  }
+
+  /**
+   * Executes the given side-effecting function if this is a `Right`.
+   *
+   * {{{
+   * Right(12).foreach(x => println(x)) // prints "12"
+   * Left(12).foreach(x => println(x))  // doesn't print
+   * }}}
+   * @param f The side-effecting function to execute.
+   */
+  def foreach[U](f: B => U): Unit = this match {
+    case Right(b) => f(b)
+    case Left(_)  =>
+  }
+
+  /**
+   * Returns the value from this `Right` or the given argument if this is a `Left`.
+   *
+   * {{{
+   * Right(12).getOrElse(17) // 12
+   * Left(12).getOrElse(17)  // 17
+   * }}}
+   */
+  def getOrElse[BB >: B](or: => BB): BB = this match {
+    case Right(b) => b
+    case Left(_)  => or
+  }
+
+  /** Returns `true` if this is a `Right` and its value is equal to `elem` (as determined by `==`),
+   *  returns `false` otherwise.
+   *
+   *  {{{
+   *  // Returns true because value of Right is "something" which equals "something".
+   *  Right("something") contains "something"
+   *
+   *  // Returns false because value of Right is "something" which does not equal "anything".
+   *  Right("something") contains "anything"
+   *
+   *  // Returns false because there is no value for Right.
+   *  Left("something") contains "something"
+   *  }}}
+   *
+   *  @param elem    the element to test.
+   *  @return `true` if the option has an element that is equal (as determined by `==`) to `elem`, `false` otherwise.
+   */
+  final def contains[AA >: A](elem: AA): Boolean = this match {
+    case Right(b) => b == elem
+    case Left(_)  => false
+  }
+
+  /**
+   * Returns `true` if `Left` or returns the result of the application of
+   * the given function to the `Right` value.
+   *
+   * {{{
+   * Right(12).forall(_ > 10) // true
+   * Right(7).forall(_ > 10)  // false
+   * Left(12).forall(_ > 10)  // true
+   * }}}
+   */
+  def forall(f: B => Boolean): Boolean = this match {
+    case Right(b) => f(b)
+    case Left(_)  => true
+  }
+
+  /**
+   * Returns `false` if `Left` or returns the result of the application of
+   * the given function to the `Right` value.
+   *
+   * {{{
+   * Right(12).exists(_ > 10) // true
+   * Right(7).exists(_ > 10)  // false
+   * Left(12).exists(_ > 10)  // false
+   * }}}
+   */
+  def exists(p: B => Boolean): Boolean = this match {
+    case Right(b) => p(b)
+    case Left(_)  => false
+  }
+
+  /**
+   * Binds the given function across `Right`.
+   *
+   * @param f The function to bind across `Right`.
+   */
+  def flatMap[AA >: A, Y](f: B => Either[AA, Y]): Either[AA, Y] = this match {
+    case Right(b) => f(b)
+    case Left(a)  => this.asInstanceOf[Either[AA, Y]]
+  }
+
+  /**
+   * The given function is applied if this is a `Right`.
+   *
+   * {{{
+   * Right(12).map(x => "flower") // Result: Right("flower")
+   * Left(12).map(x => "flower")  // Result: Left(12)
+   * }}}
+   */
+  def map[Y](f: B => Y): Either[A, Y] = this match {
+    case Right(b) => Right(f(b))
+    case Left(a)  => this.asInstanceOf[Either[A, Y]]
+  }
+
+  /** Returns `Right` with the existing value of `Right` if this is a `Right` and the given predicate `p` holds for the right value,
+   *  returns `Left(zero)` if this is a `Right` and the given predicate `p` does not hold for the right value,
+   *  returns `Left` with the existing value of `Left` if this is a `Left`.
+   *
+   * {{{
+   * Right(12).filterOrElse(_ > 10, -1) // Right(12)
+   * Right(7).filterOrElse(_ > 10, -1)  // Left(-1)
+   * Left(12).filterOrElse(_ > 10, -1)  // Left(12)
+   * }}}
+   */
+  def filterOrElse[AA >: A](p: B => Boolean, zero: => AA): Either[AA, B] = this match {
+    case Right(b) => if (p(b)) this else Left(zero)
+    case Left(a)  => this
+  }
+
+  /** Returns a `Seq` containing the `Right` value if
+   *  it exists or an empty `Seq` if this is a `Left`.
+   *
+   * {{{
+   * Right(12).toSeq // Seq(12)
+   * Left(12).toSeq  // Seq()
+   * }}}
+   */
+  def toSeq: collection.immutable.Seq[B] = this match {
+    case Right(b) => collection.immutable.Seq(b)
+    case Left(_)  => collection.immutable.Seq.empty
+  }
+
+  /** Returns a `Some` containing the `Right` value
+   *  if it exists or a `None` if this is a `Left`.
+   *
+   * {{{
+   * Right(12).toOption // Some(12)
+   * Left(12).toOption  // None
+   * }}}
+   */
+  def toOption: Option[B] = this match {
+    case Right(b) => Some(b)
+    case Left(_)  => None
+  }
+
+  def toTry(implicit ev: A <:< Throwable): Try[B] = this match {
+    case Right(b) => Success(b)
+    case Left(a)  => Failure(a)
   }
 
   /**
@@ -186,7 +346,7 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
  * @version 1.0, 11/10/2008
  */
 final case class Left[+A, +B](a: A) extends Either[A, B] {
-  def isLeft = true
+  def isLeft  = true
   def isRight = false
 }
 
@@ -197,11 +357,25 @@ final case class Left[+A, +B](a: A) extends Either[A, B] {
  * @version 1.0, 11/10/2008
  */
 final case class Right[+A, +B](b: B) extends Either[A, B] {
-  def isLeft = false
+  def isLeft  = false
   def isRight = true
 }
 
 object Either {
+
+  /** If the condition is satisfied, return the given `B` in `Right`,
+   *  otherwise, return the given `A` in `Left`.
+   *
+   * {{{
+   * val userInput: String = ...
+   * Either.cond(
+   *   userInput.forall(_.isDigit) && userInput.size == 10,
+   *   PhoneNumber(userInput),
+   *   "The input (%s) does not look like a phone number".format(userInput)
+   * }}}
+   */
+  def cond[X, Y](test: Boolean, right: => Y, left: => X): Either[X, Y] =
+    if (test) Right(right) else Left(left)
 
   /**
    * Allows use of a `merge` method to extract values from Either instances
@@ -216,8 +390,8 @@ object Either {
    */
   implicit class MergeableEither[A](private val x: Either[A, A]) extends AnyVal {
     def merge: A = x match {
-      case Left(a)  => a
       case Right(a) => a
+      case Left(a)  => a
     }
   }
 
@@ -250,7 +424,7 @@ object Either {
    * }}}
    *
    * {{{
-  * // using Either
+   * // using Either
    * def interactWithDB(x: Query): Either[Exception, Result] =
    *   try {
    *     Right(getResultFromDatabase(x))
@@ -270,6 +444,7 @@ object Either {
    * @author <a href="mailto:research@workingmouse.com">Tony Morris</a>, Workingmouse
    * @version 1.0, 11/10/2008
    */
+  @deprecated("use swap instead", "2.12.0")
   final case class LeftProjection[+A, +B](e: Either[A, B]) {
     /**
      * Returns the value from this `Left` or throws `java.util.NoSuchElementException`
@@ -282,9 +457,9 @@ object Either {
      *
      * @throws java.util.NoSuchElementException if the projection is [[scala.util.Right]]
      */
-    def get = e match {
-      case Left(a) => a
-      case Right(_) =>  throw new NoSuchElementException("Either.left.value on Right")
+    def get: A = e match {
+      case Left(a)  => a
+      case Right(_) => throw new NoSuchElementException("Either.left.get on Right")
     }
 
     /**
@@ -296,14 +471,13 @@ object Either {
      * }}}
      * @param f The side-effecting function to execute.
      */
-    def foreach[U](f: A => U) = e match {
-      case Left(a) => f(a)
-      case Right(_) => {}
+    def foreach[U](f: A => U): Unit = e match {
+      case Left(a)  => f(a)
+      case Right(_) =>
     }
 
     /**
-     * Returns the value from this `Left` or the given argument if this is a
-     * `Right`.
+     * Returns the value from this `Left` or the given argument if this is a `Right`.
      *
      * {{{
      * Left(12).left.getOrElse(17)  // 12
@@ -311,8 +485,8 @@ object Either {
      * }}}
      *
      */
-    def getOrElse[AA >: A](or: => AA) = e match {
-      case Left(a) => a
+    def getOrElse[AA >: A](or: => AA): AA = e match {
+      case Left(a)  => a
       case Right(_) => or
     }
 
@@ -327,8 +501,8 @@ object Either {
      * }}}
      *
      */
-    def forall(@deprecatedName('f) p: A => Boolean) = e match {
-      case Left(a) => p(a)
+    def forall(@deprecatedName('f) p: A => Boolean): Boolean = e match {
+      case Left(a)  => p(a)
       case Right(_) => true
     }
 
@@ -343,8 +517,8 @@ object Either {
      * }}}
      *
      */
-    def exists(@deprecatedName('f) p: A => Boolean) = e match {
-      case Left(a) => p(a)
+    def exists(@deprecatedName('f) p: A => Boolean): Boolean = e match {
+      case Left(a)  => p(a)
       case Right(_) => false
     }
 
@@ -357,9 +531,9 @@ object Either {
      * }}}
      * @param f The function to bind across `Left`.
      */
-    def flatMap[BB >: B, X](f: A => Either[X, BB]) = e match {
-      case Left(a) => f(a)
-      case Right(b) => Right(b)
+    def flatMap[BB >: B, X](f: A => Either[X, BB]): Either[X, BB] = e match {
+      case Left(a)  => f(a)
+      case Right(b) => e.asInstanceOf[Either[X, BB]]
     }
 
     /**
@@ -370,9 +544,9 @@ object Either {
      * Right[Int, Int](12).left.map(_ + 2) // Right(12)
      * }}}
      */
-    def map[X](f: A => X) = e match {
-      case Left(a) => Left(f(a))
-      case Right(b) => Right(b)
+    def map[X](f: A => X): Either[X, B] = e match {
+      case Left(a)  => Left(f(a))
+      case Right(b) => e.asInstanceOf[Either[X, B]]
     }
 
     /**
@@ -386,7 +560,7 @@ object Either {
      * }}}
      */
     def filter[Y](p: A => Boolean): Option[Either[A, Y]] = e match {
-      case Left(a) => if(p(a)) Some(Left(a)) else None
+      case Left(a)  => if(p(a)) Some(Left(a)) else None
       case Right(b) => None
     }
 
@@ -399,8 +573,8 @@ object Either {
      * Right(12).left.toSeq // Seq()
      * }}}
      */
-    def toSeq = e match {
-      case Left(a) => Seq(a)
+    def toSeq: Seq[A] = e match {
+      case Left(a)  => Seq(a)
       case Right(_) => Seq.empty
     }
 
@@ -413,8 +587,8 @@ object Either {
      * Right(12).left.toOption // None
      * }}}
      */
-    def toOption = e match {
-      case Left(a) => Some(a)
+    def toOption: Option[A] = e match {
+      case Left(a)  => Some(a)
       case Right(_) => None
     }
   }
@@ -434,6 +608,7 @@ object Either {
    * @author <a href="mailto:research@workingmouse.com">Tony Morris</a>, Workingmouse
    * @version 1.0, 11/10/2008
    */
+  @deprecated("Either is now right-biased", "2.12.0")
   final case class RightProjection[+A, +B](e: Either[A, B]) {
 
     /**
@@ -447,9 +622,9 @@ object Either {
      *
      * @throws java.util.NoSuchElementException if the projection is `Left`.
      */
-    def get = e match {
-      case Left(_) =>  throw new NoSuchElementException("Either.right.value on Left")
-      case Right(a) => a
+    def get: B = e match {
+      case Right(b) => b
+      case Left(_)  => throw new NoSuchElementException("Either.right.get on Left")
     }
 
     /**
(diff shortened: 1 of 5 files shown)


soc:
Discuss!

Some notes:
- `filter` is called `filterOrElse` because filtering in a for-comprehension doesn't work if the method needs an explicit argument, and I don't want to close the door to a future in which the Scala standard library ships with `Monoid` and `filter` could exist with an implicit `Monoid` parameter.
- There is a lot of code doing no-ops like `case Left(a) => Left(a)` replacing it with `this` would require a cast. Not sure what's the right decision between efficiency and code beauty is ...

smarter:
> There is a lot of code doing no-ops like case Left(a) => Left(a) replacing it with this would require a cast.

Can't you do something like `this match { case left @ Left(_) => left; ... }`

soc:
@smarter the problem is the other case: If `Right` as any kind of transformation, the type of `Left` doesn't fit anymore (`Left` and `Right` both carry both type params).

If you have e. g. `map` with `R => S` then the result is `Either[L, S]`, but `Left` is still `Either[L, R]`.

Sciss:
I think the problem is if you want to introduce right-bias, you should also redefine `Left` as `Either[A, Nothing]`. The following package works as @smarter suggested:

``` scala
sealed trait Either[+A, +B]

final case class Left[+A](a: A) extends Either[A, Nothing]

final case class Right[+B](b: B) extends Either[Nothing, B]

implicit class EitherOps[A, B](in: Either[A, B]) {
  def flatMap[AA >: A, Y](f: B => Either[AA, Y]): Either[AA, Y] = in match {
     case l @ Left (_) => l
     case     Right(b) => f(b)
  }

  def map[Y](f: B => Y): Either[A, Y] = in match {
    case l @ Left (_) => l
    case     Right(b) => Right(f(b))
  }
}
```

soc:
@Sciss I agree with you, but this would break every single code base where someone wrote `Left[L, R]` or `Right[L, R]`.

Sciss:
You could also safely use `asInstanceOf[Left[AA, Y]]` - it's ugly as hell, but scala-collections is already full of it...

Sciss:
Why not deprecate `Either` in 2.12, add a correct implementation e.g. `Or`, and then in 2.14 (?) we drop the problematic `Either`?

soc:
@Sciss Might be possible. I just want to avoid making library author's lives who want to support 2.10-2.1x harder than necessary.

EECOLOR:
I might have missed it, but it seems it's missing the `withFilter`. This is required if you want to extract things in a for comprehension:

```
for {
  (a, b) <- ...
} yield ...
```

Oh wait, this is probably problematic... what to do when `A => Boolean` returns `false`...

I wished Scala made the distinction between extracting tuples or case classes and guards or pattern matching.

soc:
@EECOLOR `withFilter` has the same problem as `filter`.

src/library/scala/util/Either.scala
```diff
@@ -159,6 +159,129 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
   }
 
   /**
+   * Executes the given side-effecting function if this is a `Right`.
+   *
+   * {{{
+   * Right(12).right.foreach(x => println(x)) // prints "12"
+   * Left(12).right.foreach(x => println(x))  // doesn't print
```
xuwei-k:
```
Right(12).foreach(x => println(x)) // prints "12"
Left(12).foreach(x => println(x))  // doesn't print
```

lrytz:
Related: https://groups.google.com/forum/#!topic/scala-internals/WNBFv4Ngof8/discussion

soc:
Updated!

Things to discuss:
- [x] Add `get` to `Either`?
  _Most opinions: "no"_
- [x] Added `toTry`. Good/Bad? (Should it special-case fatal exceptions?)
  _Opinions somewhere between "don't care" and "sure, why not?"_
- [ ] Should `RightProjection` be deprecated?
  _Opinions generally in favor_
- [ ] Should `LeftProjection` be deprecated? (In favor of `swap`)
  _Opinions generally in favor, but less so than for deprecating `RightProjection`_
- [x] More precise return type for `Either#toSeq`? (`collection.immutable.Seq` instead of `collection.Seq`) _Opinions varied, usually along the lines of "would be nice if the inconsistency with `LeftProjection` and `RightProjection` doesn't become to apparent" -> Reason for deprecating `LeftProjection` and `RightProjection`?_
- [x] Should we switch all `case`s to check `Right` first?
  _Mostly, I flipped cases for `Either` and `RightProjection`, keeping `LeftProjection` as-is._

japgolly:
Here's a tiny nit-pick, when using right-biased either the most frequent path is generally the success path (i.e. the `Right` path). Therefore it would be slightly better performance to switch most of the cases from

``` scala
case Left (_) => _
case Right(_) => _
```

to

``` scala
case Right(_) => _
case Left (_) => _
```

because IIUC Scala will check each case sequentially. One of those things you'd never notice would can add up after a large number of calls.

soc:
@japgolly I flipped the checks in the `Either` and `Right` types, left `Left` as is. I think this should be the optimal configuration to have the first check succeed.

som-snytt:
@japgolly b/c "In the interest of efficiency the evaluation of a pattern matching expression may try patterns in some other order than textual sequence." I'd be interested in knowing whether the compiler does the right thing when there are just two choices.

soc:
Added ScalaCheck tests.

src/library/scala/util/Either.scala
```diff
@@ -155,7 +165,133 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
...
+   * Right(12).map(x => "flower") // Result: Right("flower")
+   * Left(12).map(x => "flower")  // Result: Left(12)
+   * }}}
+   */
+  def map[Y](f: B => Y): Either[A, Y] = this match {
+    case Right(b) => Right(f(b))
+    case Left(a)  => this.asInstanceOf[Either[A, Y]]
+  }
+
+  /** Returns `None` if this is a `Left` or if the
```
szeiger: "Returns this?"

src/library/scala/util/Either.scala
```diff
@@ -155,7 +165,133 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
...
+   *  given predicate `p` does not hold for the right value,
+   *  otherwise, returns a `Right`.
+   *
+   * {{{
+   * Right(12).filterOrElse(_ > 10, -1) // Right(12)
+   * Right(7).filterOrElse(_ > 10, -1)  // Right(-1)
+   * Left(12).filterOrElse(_ > 10, -1)  // Right(-1)
+   * }}}
+   */
+  def filterOrElse[BB >: B](p: B => Boolean, zero: => BB): Either[A, BB] = this match {
```
szeiger: What's the motivation for adding this method? Neither RightProjection, Option nor Try currently have it.

szeiger:
Is a pattern match the best solution for implementing all these methods? It should compile to an instanceOf check but I would expect polymorphic dispatch to be faster still.

szeiger:
My take on the discussion points:
- Deprecate `RightProjection` (not needed anymore)
- Deprecate `LeftProjection` (in favor of `swap`)
- Make `toSeq` return an `immutable.Seq`. toSeq methods in the collections library have varying return types. Generally immutable collections return an immutable.Seq (and declare it as such). `Either` is immutable so it would be consistent with the other implementations.

soc:
@szeiger I think @Ichoran is the performance expert ... what do you think, @Ichoran?

src/library/scala/util/Either.scala
soc: People seem to want `filter` (as observed from third-party libraries), but as we lack a Monoid type in the standard library, there is no way to make it work in for-comprehensions without introducing a dangerous `implicit zero: BB` parameter).

Picking a different name allows this method to exist without interfering with for-comprehension and doesn't prevent us from adding a for-comprehension-compatible `filter` in the future if the std lib ever gets `Monoid`.

If we want this method, we should discuss whether `=> B` shoud be `B` instead, and whether we should move the parameter to a different param list...

src/library/scala/util/Either.scala
soc: Is this different from https://github.com/scala/scala/pull/5135#issuecomment-215204827?

szeiger: I was referring to "Returns `None`" in the doc comment. `None` is an `Option`. This method returns an `Either`.

src/library/scala/util/Either.scala
szeiger: I see. I assumed the reason why people seem to want `filter` is only to enable for comprehensions.

src/library/scala/util/Either.scala
soc: Ah, I see, sorry. Looks like I fixed this in a later change.

Ichoran:
@soc - You can't tell in any particular case whether pattern match or instanceof will be faster, but for two cases it's usually equivalent once the JIT compiler gets its hands on it (both are compiled down to fast single-branch code; you don't do generic polymorphic dispatch which involves walking a pointer).  If it really matters, you have to benchmark it both ways.

sjrd:
If it doesn't matter for the JVM, I can tell you that the virtual call is faster than the instanceof tests in Scala.js.

soc:
@szeiger Ok, fixed the things you mentioned.
- "Right-bias Either" is the same, just with an implementation bug in filterOrElse and the documentation fixed.
- "[squash] Add contains" It's basically the same as `Option#contains`.
- "Deprecate Either's left, right, {Left,Right}Projection" The deprecations you suggested.

soc:
@sjrd Should I rewrite it? I'm a bit concerned as the implementation gets longer and harder to follow.

sjrd:
No, keep it like that. Actually I realized that most methods are higher-order, so polymorphic dispatch for those is bad.

The best situation is that of `Option`, which has only `isEmpty` and `get` as polymorphic, and then all other methods are implemented in terms of `isEmpty` and `get`, without additional polymorphic dispatch. This is really good because the higher-order methods can be inlined and the closures eliminated, and only the two core methods remain polymorphic. For `Either`, a similar implementation would be awkward, I guess. So it's probably best to keep the current implementation.

soc:
```
*** Error in `/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java': free(): invalid pointer: 0x00000001003931a8 ***
```

WTH Jenkins?

szeiger:
LGTM

adriaanm:
This needs a commit message. It should summarize this PR, discuss design goals, motivate the change,...

adriaanm:
Adding WIP label until that's taken care of. Thanks!

adriaanm:
Thanks for being patient with me here as I play catch up, could someone convince me this not going to break existing code? Did we run the community build against this PR? Was this discussed on scala-user? (I saw the scala-internals discussion, but I fear the audience there is a bit, erm, biased :-))

soc:
Will add a decent commit message! Thanks!

soc:
Done.
Discussion: https://groups.google.com/forum/#!topic/scala-user/dRFRYy5hO3Y

Ichoran:
@soc @adriaanm - The deprecation of projections affects about one line per two thousand in my personal code bases, so it definitely creates some busywork.  Would be good to count new deprecation messages in the community build.  It shouldn't break much code which isn't already right- (or left-!) biasing Either because Either's getting new methods, not changing old ones.

SethTisue:
> This needs a commit message. It should summarize this PR, discuss design goals, motivate the change

agree. basically it needs a SLIP, not for bureaucratic reasons, but because changes like this need to be justified and agreed upon. I suggest putting it off to 2.13

soc:
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000000000010, pid=7211, tid=140609267898112
#
# JRE version: OpenJDK Runtime Environment (8.0_45-b14) (build 1.8.0_45-internal-b14)
# Java VM: OpenJDK 64-Bit Server VM (25.45-b02 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  0x0000000000000010
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/jenkins/workspace/scala-2.12.x-validate-test/hs_err_pid7211.log
[thread 140609270003456 also had an error]

[error occurred during error reporting , id 0xb]
```

soc:
Alright, closing. Anybody enthusiastic about doing a SLIP can feel free to pick this up in the future.
(I think everybody who was interested in this subject has already commented during the last SIP process on Either.)

adriaanm:
I'm still hopeful we can focus this PR to a scope that doesn't require a SLIP.

@soc, does it have any breaking changes beyond deprecation?

SethTisue:
> does it have any breaking changes beyond deprecation

I'm suggesting that the answers to questions like these be collected in a single document interested parties can review, without everyone having to individually review years of convoluted back and forth on github comment threads and mailing list threads. I'm not married to whether the word "SLIP" is attached to such a document, I would just like to see a design document justifying the change.

ghost:
> see a design document justifying the change.

Right: SLIP documents under [the slip repository](https://github.com/scala/slip/tree/master/text), eg [slip 22](https://github.com/scala/slip/blob/master/text/0022-async.md), do make sense for "big" changes.

For small changes, a document isn't required. 

But I agree with @SethTisue that for PR's like this (medium), a document is useful, also to record the _final_ outcome, not the journey. But to maintain a document independent to the PR is a PITA.

So why not have them in the PR itself, added to  say https://github.com/scala/scala/docs/PR/PR-5135-right-bias-either.md ?

adriaanm:
We can draft the document in the PR with the Q&A I started. Then collect the notes in the commit message, or potentially a separate doc. Overall, I prefer the commit message since we can't misplace it :-)

Ichoran:
If we simply _add_ the methods that give Either a right bias without _removing_ those that were used to dispatch unbiasedly, I expect the impact would be a lot less.

But, you say, that just propagates confusion and error and multiple ways to do the same thing (which is bad)!  Yeah, it does.  You could do it the "old" way and not know, or have to deal with people writing both.

So I dunno where @soc stands on this one regarding good-is-the-enemy-of-the-great.  I expect that adding a bias isn't really SLIP-worthy, whereas essentially deprecating the entire usable part of the Either interface (as opposed to it being a bare ADT) kinda is.

soc:
@adriaanm @Ichoran There are no incompatible changes. The only possibility of breakage that exists is when people have added extension methods named `map`, `flatMap` etc. to `Either` in the past doing something different than the methods added to `Either` now.

One detail that moved the scales in favor of deprecating `LeftProjection` and `RightProjection` was the desire to have `toSeq` return `scala.collection.immutable.Seq` instead of `scala.collection.Seq` like `LeftProjection` and `RightProjection` do. Therefore keeping `LeftProjection` and `RightProjection` would introduce inconsistency.

adriaanm:
Excellent! I fully expect it will LGTeveryone once the changes are documented at the top of the PR and in the commit message.

Ichoran:
Do deprecation warnings count as an "incompatible change"?  Some people like -Xfatal-warnings.

``` scala
Welcome to Scala 2.12.0-20160526-205352-c4b1362 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_72-ea).

scala> val e: Either[String, Int] = Right(1)
e: Either[String,Int] = Right(1)

scala> e.right.get
res0: Int = 1
```

Also, it's probably a good idea to have tests that demonstrate that behavior is identical between `right` and the `Either` as is.  (I'm assuming that'll get the `left` case also.)  Otherwise the deprecation message isn't solved as easily as just deleting or changing a key word.

adriaanm:
I'd say they are fine for a major release. (Ignoring deprecations on a new major release should be an error 😏 )

soc:
@Ichoran The biased ScalaCheck tests were copied from the RightProjection ones, so this should ensure that they behave equivalently.

soc:
Btw, is there a reason why scala-lang.org/api/nightly/ hasn't been updated in a while?

soc:
Small remaining notes:
- `filterOrElse` takes a by-name parameter similar to `getOrElse`. Should/can `filter` have the same signature, or would a difference be acceptable in this case?
- Should `toTry` handle fatal exceptions differently when converting to `Try`? (See the `NonFatal` pattern here...)
- The return type of `toSeq` seems to be all over the place in the collection library, but I think going forward we would rather prefer all of them to be `scala.collection.immutable.Seq` instead of `scala.collection.Seq`, right?

lrytz:
@soc for nightly api docs, check https://github.com/scala/scala#nightly-builds - where did you get the stale link? it should be fixed.

soc:
@lrytz Thanks, didn't know that. It was just muscle memory ... api/current --> api/nightly.

Ichoran:
I'm not sure what you mean about the signature of `filter`.  What are you proposing?

I don't see any problem leaving fatal exceptions that aren't already being thrown inside a `Try`.  Its job is to catch them, and it can't _catch_ the fatal ones safely.  But if they're already just sitting there in a data structure, it's fine.

`immutable.Seq` is preferred, yes!

lrytz:
While the deprecations of `left` / `right` and the projection classes are not directly breaking code, they still require people to change it eventually. What would be the drawback of keeping the projections around?

soc:
@lrytz Inconsistency of use, inconsistency of return types (as mentioned).

@Ichoran `filter` usually doesn't have a by-name parameter, but `filterOrElse` has analogous to `getOrElse`. I'm not sure `filter` can or should have that by-name parameter ... it might cause confusing behaviour when it gets rewritten in for-comprehensions where it's hard to inspect for developers.

Ichoran:
@lrytz - Additionally, that `right` sounds like an unsafe method to get the value on the right (if present).  That it instead gives you a weird transient class is a bit, well, weird.  Making that bit of weirdness go away (since it is unneeded) would be a long-term positive.

soc:
@Ichoran +1.

lrytz:
ok, sounds right. the current projection classes are basically biased eithers, but also less efficient because creating one is an allocation.

patrick-premont:
A possibility would be to deprecate `.right` to avoid confusion, but not `.left` since its rewrite is tricky: `a.left.flatMap(f)` becomes `a.swap.flatMap(f(_).swap).swap`.

If users could silence particular deprecations the impact of deprecations would be reduced. But without that, and with the possibility to treat warnings as errors, it may be best to be conservative about deprecations.

Ichoran:
Good point!  That is a less trivial rewrite.  We might want to deprecate `left` anyway, but add a `leftView` or somesuch (just to fix the naming).

szeiger:
We just discussed this PR in the Scala team. While I think the implementation is good my recommendation was not to hurry this because there doesn't seem to be a broad consensus on whether it is the right thing to do. We're wrapping up M5 in a few days so the options as we see them now are to postpone this to 2.13 (possibly via a SLIP?) or get a non-contentious minimum version in.

For example, the cleanup and possible performance improvements due to optimizing branching in existing methods for right-bias would be a useful subset.

The trade-offs of the whole thing, as far as I understand, are:
- Most use cases are either non-biased (no change required) or right-biased (usually gets easier and faster)
- Left-biased use cases become harder
- Use of for-comprehensions can become harder (no `filter` anymore)

sjrd:
> Use of for-comprehensions can become harder (no filter anymore)

Hum ... there never was a `filter` to begin with, so that's not becoming harder, does it?

> Left-biased use cases become harder

To counter this, what about _not_ deprecating `.left` and `.right`? So basically provide a purely additive PR.

This would also be good to simplify maintaining a codebase cross-compiling with 2.11 and 2.12. If we deprecate `.right` in 2.12, how am I supposed to write deprecation-free cross-compiling code? With an additive-only PR, if I need to cross-compile, I can still use `.left` and `.right` warning-free. If I only target 2.12, I can start taking advantage of the niceties of right-bias.

szeiger:
> Hum ... there never was a `filter` to begin with, so that's not becoming harder, does it?

`LeftProjection` and `RightProjection` have `filter`, the biased `Either` does not (for obvious reasons).

soc:
@szeiger Left/RighProjection`s filter doesn't make any sense, and it's not what people ask for when they want a filter method. I'd prefer getting rid of that mess as fast as possible.

> for obvious reasons

What are those reasons?

As @Ichoran proposed, I'd rather have a working LeftBiasedEither instead.

EECOLOR:
A filter method is currently required for destructuring. While I understand
why, I wish it was different from pattern matching (instead it being a
feature of products).

While not completely safe, I tend to add a no-op filter to custom monadic
types just for destructuring.

Ichoran:
For the record, I don't recall actually advocating a left-biased either, and if I did, my current self disagrees with my past self.  (I may have raised it as a possible solution to certain issues.  But it raises, others, like "Aaaah! Why are there two of them!", which is even worse than the current `Either` that looks okay until you need a projection.)

soc:
@Ichoran 

> For the record, I don't recall actually advocating a left-biased either
> 
> > We might want to deprecate left anyway, but add a leftView or somesuch (just to fix the naming).

Ichoran:
@soc - Oh, that's what you mean.  Yes, I think that's a sensible thing to do (at least in the long run).  The point is that it is _not_ a left-biased either, but basically `LeftProjection`, which is it's own weird viewy thing on top of `Either`.  It can probably be a value class, and it would resolutely push the user back rightward to avoid the "Aaah!  Two of them!" problem.

szeiger:
> @szeiger Left/RighProjection`s filter doesn't make any sense, and it's not what people ask for when they want a filter method. I'd prefer getting rid of that mess as fast as possible.

The most compelling reason for `filter` (or rather: `withFilter`) is to enable for-comprehensions. Downgrading an `Either` projection to an `Option` is not ideal but at least it does that.

On second thought, if you're content with an `Option`, you may as well start with `toOption` in the first place.

> > for obvious reasons
> 
> What are those reasons?

You can't create an empty value for the left-hand side. The best you could do is throw an exception, which could still be useful for `withFilter`. If all you care about is for-comprehensions, throwing a `MatchError` for a non-matching `withFilter` call may be acceptable. It's consistent with a destructuring assignment outside of a for-comprehension.

Ichoran:
I strongly discourage exploding on "safe" calls like `filter` and `withFilter`.  If you can't always return a sensible value, the compiler should fail to compile them.  Much of the point of these things is safe handling; throwing exceptions subverts that in a most unexpected way.

adriaanm:
As I said, I would like to get a conservative consensus-driven subset of this PR into 2.12.0. To give us enough time for a careful discussion, it looks like RC1 will be a better target for this?

adriaanm:
M5 has to be out before the end of next week, which means we have to get our PR queue flushed in the next few working days.

swaldman:
So, I'm certain I'm in a very small minority. I prefer a [different solution](https://github.com/swaldman/leftright), and use that already in my own libraries. For me, I think a symmetrical disjunction is worth keeping when Scala makes it so easy to layer a bias on top, and I think that getting standard filter behavior and full monadic citizenship should be important aspects of any fix to `Either`.

I'm wondering, in a selfish way, will this break my code? In practice, I conventionally use a right bias as the success case, and the bias I superimpose implements the usual methods in the usual ways. So, hopefully things would mostly still work. But I do rely upon standard filter behavior (as my bias allows me to nominate a token of the Left type to represent empty). And I use my biased Eithers in rich for comprehensions that might break with a nonstandard filter/withFilter. This would not quite be a nonbreaking change for me.

I feel like if I'm making any contribution at all, it's just a contribution to stasis, and that's no fun. You guys have done and are doing good work. Including this wouldn't be the end of the world to me by any stretch. But fwiw, I'm not entirely enthusiastic.

soc:
> The most compelling reason for filter (or rather: withFilter) is to enable for-comprehensions.
> You can't create an empty value for the left-hand side.

@szeiger Isn't all of this addressed in the commit message already?

szeiger:
@soc Sure. Which made me wonder why you'd ask in the first place.

swaldman:
Oh. I missed that on the top-level Either the PR implements only a nonstandard `filterOrElse(...)` rather than `filter` or `withFilter` methods. So this should not break code (like mine) that implicitly enriches Either with `withFilter`. So that's good. Sorry about that!

I'm not sure I understand the motivation for adding a right bias if it won't fix the broken-in-for-comprehension behavior of the projections. But I don't think this breaks anything (of mine).

(It would break left-biasing, which the libraries I use allow. But I don't do that in practice.)

soc:
I investigated the option of adding a left-biased Either projection/view/class over the last few weeks.

Based on that experience, I propose that we don't do it for now. Adding a left-biased Either would more or less double Either's footprint in terms of classes, methods and code, and add concerns about keeping the APIs in sync. (We would probably have to consider adding some `EitherLike` trait to prevent the classes from drifting apart over time etc.)

If there is some sign of overwhelming demand we can always revisit this decision later.

Is everyone happy/content/okayish with moving ahead with what he have already?

soc:
@szeiger @adriaanm Ping! :-)

szeiger:
We just discussed this in the team meeting again. Let's do it for RC1, including the deprecations. If this causes too many issues we can always undeprecate during the RC cycle or even later.

xuwei-k:
still using deprecated methods in Either scaladoc.

https://github.com/scala/scala/blob/09d689b8/src/library/scala/util/Either.scala#L50-L69

```
 *  A ''projection'' can be used to selectively operate on a value of type Either,
 *  depending on whether it is of type Left or Right. For example, to transform an
 *  Either using a function, in the case where it's a Left, one can first apply
 *  the `left` projection and invoke `map` on that projected Either. If a `right`
 *  projection is applied to that Left, the original Left is returned, unmodified.
 *
 *  {{{
 *  val l: Either[String, Int] = Left("flower")
 *  val r: Either[String, Int] = Right(12)
 *  l.left.map(_.size): Either[Int, Int] // Left(6)
 *  r.left.map(_.size): Either[Int, Int] // Right(12)
 *  l.right.map(_.toDouble): Either[String, Double] // Left("flower")
 *  r.right.map(_.toDouble): Either[String, Double] // Right(12.0)
 *  }}}
 *
 *  Like with other types which define a `map` method, the same can be achieved
 *  using a for-comprehension:
 *  {{{
 *  for (s <- l.left) yield s.size // Left(6)
 *  }}}
```

soc:
Ouch. I'll fix the Scaladoc. Sorry for this oversight.

soc:
It's too late now, but did you intend to merge this without (me) squashing it?

szeiger:
Yes, it was only two commits (with one being a small bug fix). It's been sitting around for long enough that I didn't expect any new issues to be discovered until it's merged and more people take notice.

notxcain:
I think it worth adding very useful `def valueOr[BB >: B](f: A => BB): BB`.

soc:
@notxcain Seems like `getOrElse` is a subset of `valueOr` that ignores the value of left... Would be nice to generalize, but I'm concerned about naming (`valueOr` is not in line with the common `get...` naming) and future evolution (if adapting `getOrElse`). (In the sense of "if we had some kind of interface standardizing `getOrElse` across types, could such a method be expressed?"

notxcain:
@soc not sure about standardisation (do we even need this interface inheritance?), but this operator is used in Cats (`Xor` and `XorT`) and scalaz (`\/`) and it's very handy.

soc:
@notxcain `getOrLeft`, `getOrElseLeft`? WDYT?

notxcain:
@soc `getOrLeft` is nice, but `valueOr` is much better, I don't like `get` prefixes in Scala :) We need some feedback from community.

mdedetrich:
I like the sound of `getOrLeft`, its at least consistent with how `Option` works

soc:
@notxcain Not "standardization" with a big S, what I had in mind was "if we ever introduce an internal trait to prevent signatures drifting apart, would this interact nicely with this extension?" (like what I would have done if we decided to add a left-biased Either: I would had introduced something like `Eitherlike` to make sure that the signatures of `Either` and `LeftEither` couldn't drift apart).

soc:
@mdedetrich I think `getOrLeft` would be ok, especially because the naming is Either-specific. (So we don't consume a possible name that could be a generalization of `getOrElse` later on...)

SethTisue:
for those interested, there are some afterthoughts on the design, and on the process by which this PR got formulated and merged, at https://github.com/scala/slip/pull/20

SethTisue:
good discussion at https://github.com/typelevel/cats/issues/1192 about uptake in Cats and other libraries

ronanM:
Why there is no `leftMap(A => C) : Either[C, B]` like in previous version of `Xor` in `Cats` ?

``` scala
def leftMap[C](f: A => C): Either[C, B] = this match {
    case Left(a)      => Left(f(a))
    case r @ Right(_) => r
}
```

IMHO `myEither.leftMap(f)` is much simpler than `myEither.swap.map(f).swap`

sjrd:
Quoting myself from https://github.com/scala/scala/pull/5135#issuecomment-222946753

> > Left-biased use cases become harder
> 
> To counter this, what about _not_ deprecating `.left` and `.right`? So basically provide a purely additive PR.
> 
> This would also be good to simplify maintaining a codebase cross-compiling with 2.11 and 2.12. If we deprecate `.right` in 2.12, how am I supposed to write deprecation-free cross-compiling code? With an additive-only PR, if I need to cross-compile, I can still use `.left` and `.right` warning-free. If I only target 2.12, I can start taking advantage of the niceties of right-bias.

Now I can show concretely what I meant by this problem. The first commit of https://github.com/scala-js/scala-js/pull/2590 illustrates the issue (the commit is https://github.com/scala-js/scala-js/pull/2590/commits/491dee16cf02cf5ffdc6fc3e5778c6f332db8c2d, but this link might become stale). Because 1) right-biased methods do not exist yet in 2.11 (nor 2.10) and 2) `.right` is deprecated in 2.12, there is no way to write cross-version warning-free code between 2.11 and 2.12. I have to use pattern matching instead, which in this case is much less readable and more verbose.

mpilquist:
FWIW, I dealt with this in FS2 under -Xfatal-warnings, -deprecation, and -Ywarn-unused-imports by creating an enrichment on `Either` that adds the 2.12 methods, implementing those in terms of pattern matching, and putting it in the main fs2 package object (to avoid an unused import warning).

https://github.com/functional-streams-for-scala/fs2/blob/series/0.9/core/shared/src/main/scala/fs2/fs2.scala#L25-L40

sjrd:
@mpilquist Yes, I think we're going to do something similar in Scala.js. It seems a bit cleaner than the current verbose code.

dwijnand:
@mpilquist you might want to consider a internal or impl package object (with an empty stub for 2.11) instead unless you're happy for those extension methods to leak into the user-land code via `fs2._`.

mpilquist:
@dwijnand I think doing so will trigger an unused import warning on 2.12 though. The extension class is marked `private[fs2]` which should accomplish the same, no? (Ignoring non-Scala clients)

dwijnand:
Hmm, not sure.

SethTisue:
re: warnings, have y'all tried https://github.com/ghik/silencer ?

sjrd:
@SethTisue No, I haven't. In theory, it looks interesting. In practice, Scala.js cannot afford to have any Scala dependency besides Scala itself. Otherwise, we wouldn't be able to publish for new Scala versions as quickly as needed. Note that most of the ecosystem transitively depends on Scala.js being published at this point, so we need to stay very close to the root of the dependency graph.

SethTisue:
I wonder if we should delay the deprecation of `.left` and `.right` until 2.13, to facilitate warning-free 2.11/2.12 cross-building. in my work on the Scala 2.12 community build lately I'm seeing huge amounts of these deprecation warnings. and it's not clear (to me anyway) how satisfying the suggested replacement for `.left` (namely) (`.swap`) is in practice. and if you are cross-building, there is no suitable substitute for `.right`

community: how are y'all feeling about this? now that 2.12.0-RC1 has been out for a few weeks

supporting 2.11/2.12 cross-building was one of our core promises for 2.12

in addition to the comments on this in the last 13 days, see also previous discussion back in June, beginning with https://github.com/scala/scala/pull/5135#issuecomment-222946753

Ichoran:
I would rather not deprecate it immediately.  The benefit of right-biasing need not come with such a heavy penalty of changing existing code and living with deprecation warnings.  Yes, we'll be tied to .right and .left for longer, but the cost is relatively low if they are things that people don't want to use, and if people do want to use them, why are we taking them away so fast?

mpilquist:
+1 for delaying the deprecation until 2.13. This also has the benefit of giving us some time to consider what other changes (if any) we want to make to `Either`.

ghost:
My 2 cents: Other than two, reported scalac issues, `Either` changes are the only remaining issues and as noted ^^^ are a killer for cross version compiling, unless we all write some compat code such as [pasted by @mpilquist ](https://github.com/functional-streams-for-scala/fs2/blob/series/0.9/core/shared/src/main/scala/fs2/fs2.scala#L25-L40).

Is it not possible to have a standard "2.12 fwd compat" module for 2.10 and 2.11? And/or a backwards compat module for 2.12 so existing code works out-of-the-box? Or in other words, keep the change but have tools and strategies in place for easy migration.

milessabin:
It's a shame, but I'm also in favour of postponing the deprecation. I'd like to see support for early migration to the new form though.

dwijnand:
@mpilquist 

> @dwijnand I think doing so will trigger an unused import warning on 2.12 though. The extension class is marked private[fs2] which should accomplish the same, no? (Ignoring non-Scala clients)

Yep, seems to do the trick, nice one (I can stop creating impl package objects for these then) :+1:

``` scala
scala> import scala.util._
import scala.util._

scala> val e: Either[String, Int] = Right(1)
e: scala.util.Either[String,Int] = Right(1)

scala> e.map(_ + 1)
<console>:16: error: value map is not a member of scala.util.Either[String,Int]
       e.map(_ + 1)
         ^

scala> import fs2._
import fs2._

scala> e.map(_ + 1)
<console>:19: error: value map is not a member of scala.util.Either[String,Int]
       e.map(_ + 1)
         ^
```

SethTisue:
we discussed this at a Scala team meeting as well. there seems to be a consensus that the deprecation should be delayed. I'll make a PR for 2.12.0-RC2

ghost:
@SethTisue Thanks for the heads up. 

Is there a _rough_ ETA for 2.12.0-RC2? And related to that, what would the advice be for a project like cats that has made some unpublished changes around this, but is yet to publish as 2.12.0-RC1? ie should it hold off on a 2.12.0-RC1 release or....?

adriaanm:
We plan to merge the last few RC2 PRs this week, and release next week,
assuming no new blocker bugs are discovered before the end of the week.

ghost:
:+1:

SethTisue:
> Is there a rough ETA for 2.12.0-RC2?

We're hoping to freeze it in the next couple days, so then there will be a nightly which is a release-candidate-candidate. If all goes well, RC2 could conceivably hit Maven Central next week sometime. Keep an eye on https://github.com/scala/scala/milestones/2.12.0-RC2

> And related to that, what would the advice be for a project like cats that has made some unpublished changes around this, but is yet to publish as 2.12.0-RC1? ie should it hold off on a 2.12.0-RC1 release or....?

Unless there's a great deal of extra hassle involved, I would still strongly recommend everyone try to publish for 2.12.0-RC1 — _especially_ a popular dependency like Cats. The best way to avoid having to do an RC3 (heaven forbid) is for everyone to be doing as much testing/publishing as possible now.

PR #5136, opened 2016-04-27
fix the Eclipse classpaths for 2.12.0-M4

performantdata:
Replace `_M4` with `-M4`.

review by @lrytz


lrytz:
LGTM, thanks for that!

PR #5137, opened 2016-04-28
Add an empty sentinel value to avoid `Future[Unit]`

backuitist:
Taken from `akka.Done` and enriched with an example of a value discarding pitfall.

### DIFF
--- a/src/library/scala/util/Try.scala
+++ b/src/library/scala/util/Try.scala
@@ -266,3 +266,7 @@ final case class Success[+T](value: T) extends Try[T] {
   override def fold[U](fa: Throwable => U, fb: T => U): U =
     try { fb(value) } catch { case NonFatal(e) => fa(e) }
 }
+
+object Success {
+  val empty: Success = Success(())
+}


retronym:
Review by @viktorklang

src/library/scala/concurrent/Done.scala
```diff
@@ -0,0 +1,36 @@
...
+ *    //
+ *    // compiles just fine: `map[T]` is infered to be `map[Unit]`, then `Future { println("two"); Done }` which
+ *    // is of type `Future[Done]` is discarded to satisfy `map[Unit]`
+ *    def compute: Future[Unit] = Future[Done] { println("one"); Done }.map( _ => Future { println("two"); Done } )
+ *
+ *    // does not compile!
+ *    def compute: Future[Done] = Future[Done] { println("one"); Done }.map( _ => Future { println("two"); Done } )
+ * }}}
+ */
+sealed abstract class Done
```
viktorklang: I think this ought to make sure that no illegitimate subclasses can be created (think javac)

so a constructor body which says: require(this.getClass.getName == "scala.concurrent.Done$")
seems like a good thing.

src/library/scala/concurrent/Done.scala
```diff
@@ -0,0 +1,36 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+package scala.concurrent
```
viktorklang: Seems a bit weird to put this in concurrent, wouldn't it be more fitting in util?
(I understand that "Done" is rather indicative of progress)

I could be convinced either way.

viktorklang:
There is a lack of tests here.

Also, I think we need to have a discussion about how this is to be introduced into the API without breaking code. See the Future companion object for instance.

viktorklang:
Ping @rkuhn

src/library/scala/concurrent/Done.scala
sjrd: Such a test could break in Scala.js if the user decides not to preserve class names at run-time. I suggest instead:

``` scala
require(this.getClass == new Array[Done.type](0).getClass.getComponentType)
```

viktorklang: @sjrd Nice catch!

rkuhn:
Yep, I’ve seen this; the question is also how to migrate Akka (if at all—probably 3.0 material).

/cc @patriknw

patriknw:
I have also seen this ;-)

One of the original purposes with `Done` and `NotUsed` was to have something that felt natural in Java APIs (`BoxedUnit` felt like internal Scala). Moving it to Scala library can work but it would be even better to have it in the JDK.

yes, we will not be able to change Akka until 3.0.

backuitist:
@patriknw If this ends up in the JDK won't we lose the niceties of Scala when used in Scala code?

src/library/scala/concurrent/Done.scala
backuitist: @viktorklang this feels a little weird IMO. I haven't this anywhere else in the scala library. Also, why isn't it in the current Akka version?

src/library/scala/concurrent/Done.scala
backuitist: > I could be convinced either way.

Me too :)

src/library/scala/concurrent/Done.scala
```diff
@@ -0,0 +1,36 @@
...
+ *    def compute: Future[Done] = Future[Done] { println("one"); Done }.map( _ => Future { println("two"); Done } )
+ * }}}
+ */
+sealed abstract class Done
+
+case object Done extends Done {
+  /**
+   * Java API: the singleton instance
+   */
+  def getInstance(): Done = this
```
sjrd: Speaking of weird ... I don't think there is any precedent for a `getInstance()` method in the Scala lib either. If `Nil` never received a `getInstance()` for Java, why should `Done`?

backuitist:
@viktorklang what do you mean by the Future companion object? Do you mean that this ~~class should find its way in the scala library~~ PR should also update the library with `Done` where possible (replace usages of `Unit`)? What kind of tests do you think of?

src/library/scala/concurrent/Done.scala
backuitist: It's been imported from Akka, but I could argue that most of the constructs from the scala library would never end up being used in Java code (in particular its immutable list)... except maybe for Done?
Then having a getInstance method doesn't hurt my eyes as much as making up for the lack of scalac options (WRT producing a singleton) through reflection.

patriknw:
> If this ends up in the JDK won't we lose the niceties of Scala when used in Scala code?

@backuitist the one that is in Akka is specially designed to be usable from both Scala and Java. That is important to make it useful in APIs for both Java and Scala.

backuitist:
@patriknw OK, I understand that you want to be Java-friendly, but how could it remain Scala-friendly if it was in the JDK?

patriknw:
ah, I see what you mean, like the companion object

src/library/scala/concurrent/Done.scala
viktorklang: @backuitist Do you know of another case in the stdlib where the need for a sealed abstract class instead of a plain case object exists?

(The reason why it isn't like I suggested in Akka is because I didn't review that piece of code)

src/library/scala/concurrent/Done.scala
viktorklang: If it's for java, perhaps make it protected[Done] or something similar, so it is hidden from Scala but visible for Java :)

src/library/scala/concurrent/Done.scala
backuitist: @viktorklang No, I agree this isn't idiomatic scala code, and as @sjrd pointed out, nor is `getInstance`.

I suggested the inclusion of `Done` within the scala lib because it seemed useful beyond Akka's boundaries. That is, in libraries that may not have an Akka dependency. As it stands now, those libs will become annoying to integrate with Akka code that is making use of `Done`. Think `futureOfUnit.map(_ => Done)`.

Now this is all about setting the limit for these hacks so that it gets accepted by both Akka and Scala. IMO we're fine with the `sealed abstract class` but not with the `require(reflectionHack)`.
Maybe someone who's more deeply involved with the scala lib could say something about this?

src/library/scala/concurrent/Done.scala
```diff
@@ -0,0 +1,36 @@
...
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+package scala.concurrent
+
+/**
+ * Typically used together with `Future` to signal completion when
+ * there is no actual value completed. More clearly signals intent
+ * than `Unit` and is available both from Scala and Java (which `Unit` is not).
```
soc: Maybe we should just make `Unit` available from Java instead?

viktorklang: @soc That would not solve the problem with Unit-inference (value discarding) of Unit tho :/

src/library/scala/concurrent/Done.scala
szeiger: +1. If it's only for Java interop, it should be hidden from Scala.

szeiger:
Since this was assigned to me, here's my take (after getting some input from the Akka team):

It looks to me like the only motivation for a separate value for this purpose is Java interop. How much should we care about that when it comes to the use of Scala Futures in the Scala standard library? Purely from a Scala point of view, I think the same meaning could be expressed by `Success[Unit]`.

Which brings me to my second point, the name. Is `Done` really what it means? I'd say that "being done" is conferred by having a completed Future, not by the value that's inside it. The meaning of the value is rather "success without a result", i.e. `Unit` or `Success[Unit]`.

All of `Done`, `Unit`, `Success.type`, `Success[Unit]` seem to be in use in Akka applications at the moment. I see the motivation for adding `scala.concurrent.Done` but It'd prefer leveraging an existing solution like `Success.type` or `Success[Unit]` and making it easier to use from Java.

src/library/scala/concurrent/Done.scala
backuitist: Ok, how can you do that? I don't think `protected[Done]` is going to work does it?

backuitist:
@szeiger what about that value-discarding issue with `Unit`?
Isn't it better to have:
`Future[Done] { /* ... */ Done }` than `Future[Success[Unit]] { /* ... */ Success(()) }`

backuitist:
To me the major annoyance is that Akka is pushing for `Done` and because `Done` isn't part of the scala library, at some point you'll have to concile `Future[Done]` and `Future[Success[Unit]]` or `Future[Success.type]` through ugly `.map(_ => Success(()))` or `.map(_ => Done)`.

viktorklang:
`Unit` from a Java/JVM PoV means everyone'll have to deal with BoxedUnit.
Success.type is not ideal there either, since that looks pretty bad as well. ($MODULE$?)
Success(()) is even worse since now you have a Success[Success[Unit]] where you either had to allocate a new instance for each usage, or have to reinvent the same machinery as Done.

szeiger:
Using `Success[Unit]` from Java is not satisfactory at the moment, which apparently prompted the addition of `akka.Done`. But if we make concessions to Java in the stdlib with a `Done.getInstance` method, we might as well add a method like `Success.empty: Success[Unit]`. Whatever we change here will require everyone to switch and in the medium term it will turn n solutions into n+1.

viktorklang:
I'm OK with defering this. The real problem is the lacking usability of Unit/BoxedUnit from Java paired with the value discarding issue in Scala.

szeiger:
@viktorklang I agree about `Unit`. It is less type-safe than the alternatives and I don't see a reasonable way of changing it (i.e. disallowing automatic conversions to `Unit`), which is why I suggested alternative types that could take the role of an empty sentinel value. But in the end what we need is a consensus from the Akka team and Akka community for a new solution (and a migration path from the currently used ones), otherwise there is no point in adding yet another "standard" to the standard library.

backuitist:
@szeiger I've renamed the PR and updated the commit

rkuhn:
An empty sentinel value that has nothing to do with `Unit` is indeed the only fix to value-discarding problems, and unfortunately `Success[Unit]` is also a monadic structure with `Unit` value inside and thus suffers from these very same problems.

I’m not advocating for an addition of the `Done` type to scala-library, as that would add yet another such value to the ecosystem—`akka.Done` cannot go away anytime soon I assume. Is there sufficient need?

backuitist:
@rkuhn I fail to see how `Success[Unit]` suffers the same problems as `Unit`?

viktorklang:
@backuitist I assume @rkuhn means something along the lines of someTry.map(x => y) <-- value discard if expected type is Success[Unit]

viktorklang:
Also, I think `Success.empty` is a misnomer. The corresponding entity in Future is called `Future.unit`

backuitist:
@viktorklang but we're talking about a replacement for `Done`, I don't understand why you would operate on `Success.empty` more than on `Done`?
About the name, I think `empty` is fine as `Success` is a container.

viktorklang:
@backuitist But it is not an empty container, it is a container which observably has the value of Unit in it.
And since it is a container with Unit in it, it has collection-like operations which sadly expose the same issues which using Unit directly does in terms of the return value expected (Success[Unit])

szeiger:
Since we don't have an agreement from the Akka guys on how it should be done or even that a short-term switch to a new stdlib feature would be possible or desirable, I'm going to close this. Maybe something to reconsider for Akka 3?

PR #5138, opened 2016-04-28
Warn, drop @native annotation in traits

retronym:
This avoids a potential VerifyError.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
@@ -552,7 +552,15 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
         return
       }
 
-      val isNative         = methSymbol.hasAnnotation(definitions.NativeAttr)
+      val isNative =
+        (methSymbol.hasAnnotation(definitions.NativeAttr)) && {
+          if (methSymbol.owner.isTrait) {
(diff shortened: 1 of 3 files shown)


retronym:
Review by @lrytz

src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
```diff
@@ -552,7 +552,15 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
         return
       }
 
-      val isNative         = methSymbol.hasAnnotation(definitions.NativeAttr)
+      val isNative =
+        (methSymbol.hasAnnotation(definitions.NativeAttr)) && {
+          if (methSymbol.owner.isTrait) {
+            reporter.warning(methSymbol.pos, "Native methods are not supprted in traits, ignoring")
```
sjrd, line 558: Typo: supported.

sjrd:
What if it is "abstract-but-concrete". IIRC, the following method is considered concrete from refchecks' point of view:

``` scala
trait T {
  @native def foo: Int
}
```

Ignoring `@native` without error would emit it as abstract but never implemented in the bytecode. Is this intended?

lrytz:
good piont, sebastien. i'd move the check to an early phase (refchecks?) and make it an error.

sjrd:
rechecks sounds like the right place to test this, yes.

adriaanm:
I'll adopt this one.

PR #5140, opened 2016-04-28
SD-140 inline the correct default method

lrytz:
When inheriting multiple default methods, select the correct one to
inline. Implements method resolution according to the JVM spec.


lrytz:
review by @retronym

retronym:
LGTM

PR #5144, opened 2016-05-02
SI-5183, SI-7088 tests

benhutchison:


test/files/pos/t5183.scala
```diff
@@ -0,0 +1,7 @@
+trait Day
+
+object Test {
+  def foo(t: Int with Day) = t == t
+}
+
+class DayOps(val i: Int with Day) extends AnyVal
```
lrytz: could you include the other / related examples that are mentioned in the various comments in the ticket? (case classes with 1 and 2 fields, the varargs issue from david's gist).

lrytz:
you can also put everything into a single commit, that matches better the granularity we are used to :)

test/files/pos/t5183.scala
```diff
@@ -0,0 +1,33 @@
...
+
+case class Test1(d: Int with Day)
+case class Test2(d1: Int with Day, d2: Int with Day)
+
+
+class User
+class Checkin
+  
+type Tagged[U] = { type Tag = U }
```
lrytz: scala doesn't allow top-level type aliases; this needs to be nested in some object

lrytz:
ping @benhutchison

lrytz:
LGTM

PR #5145, opened 2016-05-02
Dates fixing tool

FelAl:
I want my tool to run during "sbt test" command. Is it ok for CI?
It will be analyzeDates, and it'll show wrong dates without any automatic fixing.

Or maybe I should run the tool in a more appropriate way?

Thank you for your attention.

### DIFF
--- a/tools/CopyrightDateTool.scala
+++ b/tools/CopyrightDateTool.scala
@@ -1,0 +1,174 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2016, LAMP/EPFL                  **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+import sys.process._
+
+object CopyrightDateTool {
+  val usage = """
+
+    This tool compares a copyright date in files with the date of last change in git.
+    If the message of last change contains "[Sync copyright dates]" it will be ignored and
+    we will take next commit.
+
+    analyze - just shows if anything is ok
+    update - file fixing
+
+    From root folder:
+    Usage: scala tools/CopyrightDateTool.scala analyze
+           scala tools/CopyrightDateTool.scala update
+
+  """
+  trait DateComparison {
+    def fileLastUpdateYear: String
+    def gitLastUpdateYear: String
+    def fileCreationYear: String
+    def gitCreationYear: String
+
+    def isUpdateDateWrong(): Boolean = {
+      fileLastUpdateYear.toInt != gitLastUpdateYear.toInt
+    }
+
+    def isUpdateDateWrongStr(): Boolean = {
+      fileLastUpdateYear != gitLastUpdateYear
+    }
+  }
+
+  case class FileDates(pathToFile: String,
+                      fileLastUpdateYear: String,
+                      gitLastUpdateYear: String,
+                      fileCreationYear: String,
+                      gitCreationYear: String) extends DateComparison
+
+  case class GitInfo(message: String, date: String)
+
+  def readFilesList(): Array[String] = {
+    val git_files = Seq("/bin/sh", "-c", 
+      "git grep --name-only --fixed-strings" + 
+      " '|    (c) 20' -- '*.scala'").!!
+    val files_array = git_files.split("\n")
+    files_array
+  }
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)("UTF-8")
+    val lines = try source.mkString finally source.close()
+    lines
+  }
+
+  def getGitCreationYear(pathToFile: String): String = {
+    val buf_dates = s"git log --format=%ci $pathToFile" !!
+    val dates = buf_dates.split("\\n")
+    val year_column = 0
+    val res = dates.last.split("-")(year_column)
+    res
+  }
+
+  def getGitLastUpdateYear(pathToFile: String): String = {
+    val buf_dates = s"git log -3 --format=%cd $pathToFile" !!
+    val buf_messages = s"git log -3 $pathToFile" !!
+    val first_empty_elem = 1
+    val messages = buf_messages.split("commit").drop(first_empty_elem)
+    val dates = buf_dates.split("\\n")
+    val gits_info = (messages, dates).zipped.map{ case(a,b) => GitInfo(a,b)}
+    val res = gits_info.filterNot(_.message.contains("[Sync copyright dates]"))
+    val year_column = 4
+    val first = 0
+    res(first).date.split(" ")(year_column)
+  }
+
+  def getYear(pathToFile: String, lower_bound: Int, upper_bound: Int): String = {
+    val file = readFile(pathToFile)
+    val third_line = 2
+    val start_index = file.split("\\n")(third_line).indexOf("(c) 20")
+    val year = file.split("\\n")(third_line)
+                   .substring(start_index + lower_bound,
+                              start_index + upper_bound)
+    year
+  }
+
+  def getFileCreationYear(pathToFile: String): String = {
+    val lb = "(c) ".length()
+    val ub = "(c) 20XX".length()
+    val fileCreationYear = getYear(pathToFile, lb, ub)
+    fileCreationYear
+  }
+
+  def getFileLastUpdateYear(pathToFile: String): String = {
+    val lb = "(c) 20XX-".length()
+    val ub = "(c) 20XX-XXXX".length()
+    val fileLastUpdateYear = getYear(pathToFile, lb, ub)
+    fileLastUpdateYear
+  }
+
+  def analyzeDates(): Array[FileDates] = {
+    val files_array = readFilesList()
+    val data = files_array map {
+      case file_path =>
+      val fileLastUpdateYear = getFileLastUpdateYear(file_path)
+      val gitLastUpdateYear = getGitLastUpdateYear(file_path)
+      val fileCreationYear = getFileCreationYear(file_path)
+      val gitCreationYear = getGitCreationYear(file_path)
+      FileDates(file_path,
+                fileLastUpdateYear,
+                gitLastUpdateYear,
+                fileCreationYear,
+                gitCreationYear)
+     }
+    data
+  }
+
+  def fixWrongDates(): Boolean = {
+    var check = false
+    val files = analyzeDates()
+    files.foreach { file =>
+      val file_file = readFile(file.pathToFile)
+      var new_file: Option[String] = None
+      if (file.fileLastUpdateYear.forall(_.isDigit)) {
+        if (file.isUpdateDateWrong()) {
+          new_file =Some(file_file.
+                       replace(
+                        s"${file.fileCreationYear}-${file.fileLastUpdateYear}",
+                        s"${file.fileCreationYear}-${file.gitLastUpdateYear}"
+                       ))
+        }
+      }
+      if(new_file.isDefined) {
+        scala.tools.nsc.io.File(file.pathToFile)("UTF-8").writeAll(new_file.get)
+        check = true 
+      }
+    }
+    check
+  }
+
+  def resultAnalyzeDates(): Boolean = {
+    val data = analyzeDates
+    val (error_files, ok_files) = data.partition(x =>
+      x.isUpdateDateWrongStr() &&
+      x.fileLastUpdateYear.forall(_.isDigit))
+    if (error_files.nonEmpty)
+    {
+      println("""
+        Format
+        FileDates(file_path,file update year,git update year, file creation year, git creation year)
+        """)
+      error_files.foreach { file => println(file) }
+      println("Oops...We have some error dates")
+    }
+    else println ("Everything is ok")
+    error_files.nonEmpty
+  }
+
+  def main(args: Array[String]) {
+    val argList = args.toList
+    argList match {
+      case "analyze" :: Nil => resultAnalyzeDates()
+      case "update" :: Nil => fixWrongDates()
+      case _ => println(usage)
+    }
+  }
+}


Ichoran:
I'm not sure it's worth the electricity to make sure the copyrights are exactly right instead of off by maybe a year.  Copyright lasts a long time.  I think we can do this manually once or a few times a year.

FelAl:
I'm not sure it's worth the electricity(с)

I think so too. Execution takes about 8-10 minutes.

So, I am going to fix any suggestions(scala style, etc)

FelAl:
@janekdb 
This is a commit with a tool only.

adriaanm:
Great! Let's get this tool merged first and then submit a PR to 2.11.x to correct the dates there. Then, we can merge 2.11.x into 2.12.x and see how it goes. Finally, let's run it again to produce a PR that targets 2.12.x. If you agree, this will mean closing the PRs that are currently targeting for M5 (#5119, #5120). Thanks for your patience & persistence on this!

FelAl:
0) Now, I am waiting for reviewing. After review, I'll fix errors.
1) Next, someone will merge this PR to 2.11.x
2) After merge, I will produce 2 PR's with correct dates into 2.11.x
3) Next, I'll be waiting for merging those 2 PR into 2.11.x and merging 2.11.x into 2.12.x
4) Next, I'll run this tool on 2.12.x and create a new PR(the last) if there are any wrong dates.

And that's it.
Is my plan correct?

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)("UTF-8")
+    val lines = try source.mkString finally source.close()
+    lines
+  }
+
+  def ifFileWasInGit(pathToFile: String): Boolean = {
+    val res_dates = (s"git log $pathToFile" !!)
+    if(res_dates == "") {
```
janekdb: The if expression is equivalent to `res_dates.nonEmpty`. Possibly clearer than the if expression.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+      false
+    } else {
+      true
+    }
+  }
+
+  def getGitCreationYear(pathToFile: String): String = {
+    val buf_dates = (s"git log   --format=%cd $pathToFile" !!)
```
janekdb: `ci` could be better choice to extract the year from the commit date,

```
$ git log --format=%cd|head -4
Mon May 2 15:51:37 2016 +0200
Mon May 2 15:36:09 2016 +0200
Mon May 2 12:20:53 2016 +0200
Mon May 2 10:09:09 2016 +0200

$ git log --format=%ci|head -4
2016-05-02 15:51:37 +0200
2016-05-02 15:36:09 +0200
2016-05-02 12:20:53 +0200
2016-05-02 10:09:09 +0200

```

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+    val res = dates.last.split(" ")(4)
+    res
+  }
+
+  def getGitLastUpdateYear(pathToFile: String): String = {
+    val buf_dates = (s"git log -3  --format=%cd $pathToFile" !!)
+    val buf_messages = (s"git log -3  $pathToFile" !!)
+    val messages = buf_messages.split("commit").distinct.drop(1)
+    val dates = buf_dates.split("\\n").distinct
+    val gits_info = (messages, dates).zipped.map{ case(a,b) => GitInfo.tupled((a,b))}
```
janekdb: Just `GitInfo(a, b)`?

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+    val res = gits_info.filter(!_.message.contains("[Dates fixing]"))
```
janekdb: Use a more targeted tag: `[Sync copyright dates]`? This at least mentioned the type of date being fixed.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+    res(0).date.split(" ")(4)
+  }
+
+  def getYear(pathToFile: String, lower_bound: Int, upper_bound: Int): String = {
+    val file = readFile(pathToFile)
+    val start_index = file.indexOf("(c) 20")
```
janekdb: Suggest applying a limit to how far into the file the substring can be search for. Since the copyright header has a known format I'd be inclined to look in line 3 only,

```
/*                     __                                               *\
**     ________ ___   / /  ___     Scala API                            **
**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
** /____/\___/_/ |_/____/_/ | |                                         **
**                          |/                                          **
\*                                                                      */
```

You might be interesting in backlogging other types of copyright header or discussing a move to a uniform header.

```
/* NSC -- new Scala compiler
 * Copyright 2007-2013 LAMP/EPFL
 * @author  Manohar Jonnalagedda
 */
```

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+    val year = file.substring(start_index + lower_bound, start_index + upper_bound)
+    year
+  }
+
+  def getActualCreationYear(pathToFile: String): String = {
+    val lb = "(c) ".length()
```
janekdb: Do the same thing with the "junit" string earlier. i.e. no magic number `4`.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+  def analyzeDates(): Array[FileDates] = {
+    val files_array = readFilesList()
+    var data = Array[FileDates]()
+    for(file_path <- files_array) {
+      if(readFile(file_path).contains("|    (c) 20") && ifFileWasInGit(file_path)) {
+        val actualLastUpdateYear = getActualLastUpdateYear(file_path)
+        val gitLastUpdateYear = getGitLastUpdateYear(file_path)
+        val actualCreationYear = getActualCreationYear(file_path)
+        val gitCreationYear = getGitCreationYear(file_path)
+        data = data :+ FileDates.tupled((file_path,
```
janekdb: `tupled` isn't required.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+    }
+    data
+  }
+
+  def fixWrongDates(): Boolean = {
+    var check = false
+    val files = analyzeDates()
+    files.foreach { file =>
+      val actual_file = readFile(file.pathToFile)
+      var new_file = ""
```
janekdb: Using `""` as a rogue value is fine but you could consider using `None`/`Some` instead.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+          check = true
+        }
+      }
+    }
+    if (check) println("Oops...We have some error dates")
+    else println ("Everything is ok")
+    check
+  }
+
+  def main(args: Array[String]) {
```
janekdb: What error code does the script exit with for `resultAnalyzeDates` in these two scenarios,
- No out of sync dates
- Out of sync dates

Just checking the status code is useful from a scripting point of view.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
+import sys.process._
+
+object CheckDates {
```
janekdb: `CopyrightDateTool` is an improved name because,
- It is more specific
- It requires no explanation as to it's purpose

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+  val usage = """
+
+    This tool compares a copyright date in files with the date of last change in git.
+    If the message of last change match regex "\\[*.\\]" (e.g. '[Tech message]')  it will be ignored and
```
janekdb: This regex is looser than required.

```
scala> val techPat = "\\[*.\\]".r
techPat: scala.util.matching.Regex = \[*.\]

scala> techPat findFirstIn "aa [Sync copyright dates] bb"
res0: Option[String] = Some(s])

scala> techPat findFirstIn "aa dates] bb"
res1: Option[String] = Some(s])
```

Drop the over generalisation and look for a fixed string. Here is my recurring candidate: `[Sync copyright dates]`.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
+import sys.process._
+
+object CheckDates {
+  val usage = """
+
+    This tool compares a copyright date in files with the date of last change in git.
+    If the message of last change match regex "\\[*.\\]" (e.g. '[Tech message]')  it will be ignored and 
+    we will take next commit.
+
+    analyzeDates - just shows if anything is ok
```
janekdb: We know from the tool name `CopyrightDateTool` that we are dealing with dates. Less is more?

```
  analyze - list copyright dates that are out of sync with version history
  update - fix incorrect copyright dates
```

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+      actualLastUpdateYear != gitLastUpdateYear
+    }
+
+    def ifOneYear(): Boolean = {
+      gitCreationYear != gitLastUpdateYear
+    }
+  }
+
+  case class FileDates(pathToFile: String,
+                      actualLastUpdateYear: String,
```
janekdb: The actual value is the value in the Git history. Use "file" instead of "actual".
- Rename `actualLastUpdateYear` to `fileLastUpdateYear`
- Rename `actualCreationYear` to `fileCreationYear`

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+                      gitCreationYear: String) extends DateComparison
+
+  case class GitInfo(message: String, date: String)
+
+  def findRootFolder(): String = {
+    val wd = (s"pwd") !!
+    var root_folder = wd
+
+    if (wd.split("/").contains("junit")) {
+       root_folder =  wd.split("/").dropRight(4).mkString("/")
```
janekdb: Avoid magic number `4`.

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  case class FileDates(pathToFile: String,
+                      actualLastUpdateYear: String,
+                      gitLastUpdateYear: String,
+                      actualCreationYear: String,
+                      gitCreationYear: String) extends DateComparison
+
+  case class GitInfo(message: String, date: String)
+
+  def findRootFolder(): String = {
```
janekdb: If there assumptions about where this command should be run from add a guide to the `usage` string. Otherwise consider adding a command option to avoid the root to be specified.

`CopyrightDateTool ~/workspace/scala analyze`

tools/CheckDates.scala
```diff
@@ -0,0 +1,184 @@
...
+    val buf_dates = (s"git log -3  --format=%cd $pathToFile" !!)
+    val buf_messages = (s"git log -3  $pathToFile" !!)
+    val messages = buf_messages.split("commit").distinct.drop(1)
+    val dates = buf_dates.split("\\n").distinct
+    val gits_info = (messages, dates).zipped.map{ case(a,b) => GitInfo.tupled((a,b))}
+    val res = gits_info.filter(!_.message.contains("[Dates fixing]"))
+    res(0).date.split(" ")(4)
+  }
+
+  def getYear(pathToFile: String, lower_bound: Int, upper_bound: Int): String = {
```
janekdb: Would this code that deals with offsets and upper and lower bounds be simplified by using this pattern as a regular expression?

```
\(c\) 20\d\d
```

It would mean this code could be replace with an `Option.isDefined` call

```
if (file.actualLastUpdateYear.forall(_.isDigit)) {
```

tools/CheckDates.scala
FelAl: backlogging other types of copyright header or discussing a move to a uniform header.(c)

They are excluded by `if(readFile(file_path).contains("|    (c) 20") && ifFileWasInGit(file_path)) {`

I can analyze/fix another type of copyrigths in an additional PR.

tools/CheckDates.scala
FelAl: If we have found wrong date = true returns
all is ok and we don't find nothing = false returns.

tools/CheckDates.scala
FelAl: In most cases
/ __/ __// _ | / /  / _ |    (c) 2002-2013, LAMP/EPFL

But in some cases
/ __/ __// _ | / /  / _ |    (c) 2016, LAMP/EPFL
And with upper and lower bounds I cut precisely 4 symbols and analyze them.(_.isDigit or not)

In my opinion, using of addition regex is "overengineering".

tools/CopyrightDateTool.scala
```diff
@@ -67,17 +56,17 @@ object CheckDates {
 
   def ifFileWasInGit(pathToFile: String): Boolean = {
     val res_dates = (s"git log $pathToFile" !!)
-    if(res_dates == "") {
-      false
-    } else {
+    if(res_dates.nonEmpty) {
       true
```
janekdb: More like this,

```
def ifFileWasInGit(pathToFile: String): Boolean = {
  val res_dates = (s"git log $pathToFile" !!)
  res_dates.nonEmpty
}
```

tools/CopyrightDateTool.scala
```diff
@@ -135,20 +129,20 @@ object CheckDates {
...
-                       )
+                        s"${file.fileCreationYear}-${file.fileLastUpdateYear}",
+                        s"${file.fileCreationYear}-${file.gitLastUpdateYear}"
+                       ))
         }
       }
-      if(new_file != "") {
-        scala.tools.nsc.io.File(file.pathToFile)("UTF-8").writeAll(new_file)
-        check = true
+      if(new_file != None) {
```
janekdb: `if(new_file.isDefined) {`

tools/CopyrightDateTool.scala
FelAl: Oh))) You are right)))

FelAl:
@adriaanm, could you answer my question?

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
+import sys.process._
+
+object CheckDates {
```
janekdb: Align object name to file name: `CopyrightDateTool`

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+           scala tools/CheckDates.scala update
+
+  """
+  trait DateComparison {
+    def fileLastUpdateYear: String
+    def gitLastUpdateYear: String
+    def fileCreationYear: String
+    def gitCreationYear: String
+
+    def ifUpdateDateWrong(): Boolean = {
```
janekdb: `isUpdateDateWrong`

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+      fileLastUpdateYear.toInt != gitLastUpdateYear.toInt
+    }
+
+    def ifUpdateDateWrongStr(): Boolean = {
```
janekdb: `isUpdateDateWrongStr`

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+      fileLastUpdateYear != gitLastUpdateYear
+    }
+
+    def ifOneYear(): Boolean = {
```
janekdb: `isAtLeastOneYearOld`

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)("UTF-8")
+    val lines = try source.mkString finally source.close()
+    lines
+  }
+
+  def ifFileWasInGit(pathToFile: String): Boolean = {
+    val res_dates = (s"git log $pathToFile" !!)
+    res_dates.nonEmpty
```
janekdb: LGTM

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+    val ub = "(c) 20XX-XXXX".length()
+    val fileLastUpdateYear = getYear(pathToFile, lb, ub)
+    fileLastUpdateYear
+  }
+
+  def analyzeDates(): Array[FileDates] = {
+    val files_array = readFilesList()
+    var data = Array[FileDates]()
+    for(file_path <- files_array) {
+      if(readFile(file_path).contains("|    (c) 20") && ifFileWasInGit(file_path)) {
```
janekdb: This could be simplified by starting with a list of Scala files that are in the repo instead of starting with a list of Scala files that are in the working directory and then skipping the files that are not in the repo.

This gives a list of *.scala files that are known to be in the repo,

`git ls-files|grep '\.scala$'`

This will avoid thousands of calls to `git log $pathToFile`. If you make this change please post the before and after times for `analyse` to run.

In fact we can go further and delegate the content matching to `git` as well,

From `git help grep`,

```
  git grep 'time_t' -- '*.[ch]'
       Looks for time_t in all tracked .c and .h files in the working directory and its subdirectories.
```

Try this as see if it gives the same list of Scala files,

```
git grep --name-only "|    (c) 20" -- '*.scala'
```

FelAl: "find . -name _.scala ! -path ./tools/_  -type f" files array length

```
=============Length!!!!!!
6745
=========Length!!!!!!
```

Git ls Files array length

```
=============Length!!!!!!
6765
=========Length!!!!!!
```

exclude .check extension(`*.scala.check` breaks tool during `readFile`)

```
=============Length!!!!!!
6761
=========Length!!!!!!
```

`time scala tools/CopyrightDateTool.scala analyze`

time execution with `ifFileWasInGit`

```
real    9m19.294s
user    9m0.976s
sys 0m17.852s
```

time execution without `ifFileWasInGit`

```
real    5m15.929s
user    5m5.572s
sys 0m11.052s
```

af@fv:~/mkdev/scala$ time scala tools/CopyrightDateTool.scala update
warning: there were 5 feature warnings; re-run with -feature for details
one warning found

```
real    5m8.499s
user    4m58.980s
sys 0m10.596s
```

FelAl: `git ls-files|grep '\.scala$'`

This actually didn't work. A string was empty(I'd tried different combinations). So I did workaround.

FelAl:
What can I improve in the tool also?

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+  case class FileDates(pathToFile: String,
+                      fileLastUpdateYear: String,
+                      gitLastUpdateYear: String,
+                      fileCreationYear: String,
+                      gitCreationYear: String) extends DateComparison
+
+  case class GitInfo(message: String, date: String)
+
+  def readFilesList(): Array[String] = {
+    val git_files = (s"git ls-files") !!
```
janekdb: In fact we can go further and delegate the content matching to `git` as well,

From `git help grep`,

```
  git grep 'time_t' -- '*.[ch]'
       Looks for time_t in all tracked .c and .h files in the working directory and its subdirectories.
```

This will result in a list of files that are known to the repo, include the target string and end in .scala.

```
git grep --name-only --fixed-strings "|    (c) 20" -- '*.scala'
```

For me this results in 602 files. Here is the output followed by a cross checking using `fgrep` and `find`,

```
$ git grep --name-only --fixed-strings "|    (c) 20" -- '*.scala'|sort|cat -n|tail
593 src/reflect/scala/reflect/io/File.scala
594 src/scaladoc/scala/tools/ant/Scaladoc.scala
595 test/disabled/pos/spec-List.scala
596 test/files/pos/spec-Function1.scala
597 test/instrumented/library/scala/runtime/ScalaRunTime.scala
598 test/junit/scala/util/control/ExceptionTest.scala
599 test/scaladoc/resources/doc-root/AnyRef.scala
600 test/scaladoc/resources/doc-root/Any.scala
601 test/scaladoc/resources/doc-root/Nothing.scala
602 test/scaladoc/resources/doc-root/Null.scala

$ fgrep -l  '|    (c) 20' $(find -name '*.scala')|sort|cat -n|tail
593 ./src/reflect/scala/reflect/io/File.scala
594 ./src/scaladoc/scala/tools/ant/Scaladoc.scala
595 ./test/disabled/pos/spec-List.scala
596 ./test/files/pos/spec-Function1.scala
597 ./test/instrumented/library/scala/runtime/ScalaRunTime.scala
598 ./test/junit/scala/util/control/ExceptionTest.scala
599 ./test/scaladoc/resources/doc-root/AnyRef.scala
600 ./test/scaladoc/resources/doc-root/Any.scala
601 ./test/scaladoc/resources/doc-root/Nothing.scala
602 ./test/scaladoc/resources/doc-root/Null.scala
```

By starting with a list of files known to pass this test,

```
if(readFile(file_path).contains("|    (c) 20")) {
```

the number of files processed will be reduced by 10x so the execution time should drop down to ~ 30 seconds.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+        check = true 
+      }
+    }
+    check
+  }
+
+  def resultAnalyzeDates(): Boolean = {
+    val data = analyzeDates
+    var check = false
+    data.foreach {file =>
```
janekdb: This is where `partition` can be used. partition `data` into two lists,
- A list of files with the wrong date. Call that A.
- The other files.

If A is non empty
- print "Oops..."
- print all paths in A

If A is empty print "No incorrect copyright dates found"

Return A.nonEmpty

Take a look at the docs: http://www.scala-lang.org/api/2.11.7/index.html#scala.Array@partition(p:A=>Boolean):(Repr,Repr)

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+    files_array
+  }
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)("UTF-8")
+    val lines = try source.mkString finally source.close()
+    lines
+  }
+
+  def ifFileWasInGit(pathToFile: String): Boolean = {
```
janekdb: This method is no longer used. It can be safely deleted.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,174 @@
...
+    }
+    data
+  }
+
+  def fixWrongDates(): Boolean = {
+    var check = false
+    val files = analyzeDates()
+    files.foreach { file =>
+      val file_file = readFile(file.pathToFile)
+      var new_file = None: Option[String]
```
janekdb: This would work as well and is less unusual: `var new_file: Option[String] = None`

tools/CopyrightDateTool.scala
FelAl: "the number of files processed will be reduced by 10x so the execution time should drop down to ~ 30 seconds."

It hasn't reduced anything. ".contains" was not the heaviest operation.
The most time-consuming action is `"git ..." !!` I think.
Execution time is still 5 min and has reduced very slightly.

FelAl:
All suggestions have fixed. What's next? Last style fix(check addtitional spaces, etc.) and squash commits?

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+        check = true 
+      }
+    }
+    check
+  }
+
+  def resultAnalyzeDates(): Boolean = {
+    val data = analyzeDates
+    var check = false
+    val (error_files, ok_files) = data.partition(x => (x.isUpdateDateWrongStr() && (x.fileLastUpdateYear.forall(_.isDigit))))
```
janekdb: Some parentheses can be removed from the predicate.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+        scala.tools.nsc.io.File(file.pathToFile)("UTF-8").writeAll(new_file.get)
+        check = true 
+      }
+    }
+    check
+  }
+
+  def resultAnalyzeDates(): Boolean = {
+    val data = analyzeDates
+    var check = false
```
janekdb: Drop the use of `check`. All the required information about the existence of errors is available from `error_files`.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+      println("============")
+      println("""
+        Format
+        FileDates(file_path,file update year,git update year, file creation year, git creation year)
+        """)
+      error_files.foreach { file => println(file) }
+      println("Oops...We have some error dates")
+    }
+    else println ("Everything is ok")
+    check
```
janekdb: Replace with  `error_files.nonEmpty`

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  def resultAnalyzeDates(): Boolean = {
+    val data = analyzeDates
+    var check = false
+    val (error_files, ok_files) = data.partition(x => (x.isUpdateDateWrongStr() && (x.fileLastUpdateYear.forall(_.isDigit))))
+    if (error_files.nonEmpty) 
+    {
+      check = true
```
janekdb: Remove `check`

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+  case class FileDates(pathToFile: String,
+                      fileLastUpdateYear: String,
+                      gitLastUpdateYear: String,
+                      fileCreationYear: String,
+                      gitCreationYear: String) extends DateComparison
+
+  case class GitInfo(message: String, date: String)
+
+  def readFilesList(): Array[String] = {
+    val git_files = Seq("/bin/sh", "-c",
```
janekdb, line 50: What is the difference in the running time?

FelAl, line 50: https://github.com/scala/scala/pull/5145#discussion-diff-63394600R42

I answered already there. There is no huge difference. I chose `Seq()` because it works exactly like regular shell(no `#|` etc.)

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+                      gitLastUpdateYear: String,
+                      fileCreationYear: String,
+                      gitCreationYear: String) extends DateComparison
+
+  case class GitInfo(message: String, date: String)
+
+  def readFilesList(): Array[String] = {
+    val git_files = Seq("/bin/sh", "-c", 
+      "git grep --name-only --fixed-strings" + 
+      " '|    (c) 20' -- '*.scala'|grep -v CopyrightDateTool").!!
```
som-snytt: Today I learned

```
amarki@amarki-462836:~/projects/scala$ git grep -E -e '(\(c\)|Copyright) 20[0-9]{2}-(20[0-9]{2})' --and --not -e EPFL -- '*.scala' | more
src/library/scala/collection/mutable/History.scala:**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/tPFL             **
```

I see you're checking only the banner here; it took me a while to understand the leading VBAR.

There are other badly formed or deviant notices, too, from third parties like the classic Contributor.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+                              start_index + upper_bound)
+    year
+  }
+
+  def getFileCreationYear(pathToFile: String): String = {
+    val lb = "(c) ".length()
+    val ub = "(c) 20XX".length()
+    val fileCreationYear = getYear(pathToFile, lb, ub)
+    fileCreationYear
+  }
```
som-snytt, line 99: Instead of reading the file twice, and resplitting the file, you could just use a regex on the output of git grep. (Which gives path, start year and optionally end year.)

Also beware that some files have multiple lines that appear to match, like scaladoc sources that emit copyrights.

Footnote if you were reading the file, you'd `file.getLines.take(10)` to ensure you're only looking at headers.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+    val files = analyzeDates()
+    files.foreach { file =>
+      val file_file = readFile(file.pathToFile)
+      var new_file: Option[String] = None
+      if (file.fileLastUpdateYear.forall(_.isDigit)) {
+        if (file.isUpdateDateWrong()) {
+          new_file =Some(file_file.
+                       replace(
+                        s"${file.fileCreationYear}-${file.fileLastUpdateYear}",
+                        s"${file.fileCreationYear}-${file.gitLastUpdateYear}"
```
som-snytt, line 136: This replacement would be more flexible with the regex match, as well as safer. For example, you would replace the string constants in the scaladoc sources by doing global replace on the file contents. And you could replace "(c) 2009" with "Copyright 2009-2016" or however it ought to be normalized.

som-snytt:
Footnote, I wonder about the rules for (c) Contributor. Do you update that date, or add a supplementary notice for changes that are (c) Us?

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+        """)
+      error_files.foreach { file => println(file) }
+      println("Oops...We have some error dates")
+    }
+    else println ("Everything is ok")
+    check
+  }
+
+  def main(args: Array[String]) {
+    if (args.length != 1) println(usage)
```
janekdb: Remove this line. The `case _` match on line 167 handles this condition.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+    if (args.length == 1) {
```
janekdb: Remove this if check. The match handles all scenarios.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+    val lb = "(c) 20XX-".length()
+    val ub = "(c) 20XX-XXXX".length()
+    val fileLastUpdateYear = getYear(pathToFile, lb, ub)
+    fileLastUpdateYear
+  }
+
+  def analyzeDates(): Array[FileDates] = {
+    val files_array = readFilesList()
+    var data = Array[FileDates]()
+    for(file_path <- files_array) {
```
janekdb: map over `files_array` instead of using a for loop

```
val data = files_array map {
  case file_path =>
    val ...
    val ...
    val ...
    val ...
    FileDates(file_path, ...)
 }
```

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)("UTF-8")
+    val lines = try source.mkString finally source.close()
+    lines
+  }
+
+  def getGitCreationYear(pathToFile: String): String = {
+    val buf_dates = (s"git log   --format=%ci $pathToFile" !!)
+    val dates = buf_dates.split("\\n").distinct
```
janekdb: Remove `distinct` and you can remove the assumption `distinct` preserves the sort order.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  def readFile(pathToFile: String): String = {
+    val source = scala.io.Source.fromFile(pathToFile)("UTF-8")
+    val lines = try source.mkString finally source.close()
+    lines
+  }
+
+  def getGitCreationYear(pathToFile: String): String = {
+    val buf_dates = (s"git log   --format=%ci $pathToFile" !!)
```
janekdb: Please review this post: http://stackoverflow.com/questions/11533199/find-commit-where-file-was-added

Can this be done with less Scala and more Git?

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  def getGitLastUpdateYear(pathToFile: String): String = {
+    val buf_dates = (s"git log -3  --format=%cd $pathToFile" !!)
+    val buf_messages = (s"git log -3  $pathToFile" !!)
+    val first_empty_elem = 1
+    val messages = buf_messages.split("commit").distinct.drop(first_empty_elem)
+    val dates = buf_dates.split("\\n").distinct
+    val gits_info = (messages, dates).zipped.map{ case(a,b) => GitInfo(a,b)}
+    val res = gits_info.filter(!_.message.contains("[Sync copyright dates]"))
```
janekdb: Use `filterNot` and invert the sense of the predicate.

tools/CopyrightDateTool.scala
FelAl: `scala tools/CopyrightDateTool.scala update`
`scala tools/CopyrightDateTool.scala analyze`

`case _` match
`scala tools/CopyrightDateTool.scala updat` or
`scala tools/CopyrightDateTool.scala analdfaaf`

FelAl:
Footnote, I wonder about the rules for (c) Contributor. Do you update that date, or add a supplementary notice for changes that are (c) Us?
@som-snytt 

I don't understand what you are talking about. What rules?

tools/CopyrightDateTool.scala
```diff
@@ -4,7 +4,7 @@ object CopyrightDateTool {
   val usage = """
 
     This tool compares a copyright date in files with the date of last change in git.
-    If the message of last change contains "[Sync copyright dates]" it will be ignored and 
+    If the message of last change contains "[Sync copyright dates]" it will be ignored and
```
janekdb, line 15: lgtm

tools/CopyrightDateTool.scala
```diff
@@ -53,16 +53,16 @@ object CopyrightDateTool {
   }
 
   def getGitCreationYear(pathToFile: String): String = {
-    val buf_dates = (s"git log   --format=%ci $pathToFile" !!)
+    val buf_dates = s"git log   --format=%ci $pathToFile" !!
```
janekdb: lgtm

tools/CopyrightDateTool.scala
```diff
@@ -53,16 +53,16 @@ object CopyrightDateTool {
...
     val dates = buf_dates.split("\\n").distinct
     val year_column = 0
     val res = dates.last.split("-")(year_column)
     res
   }
 
   def getGitLastUpdateYear(pathToFile: String): String = {
-    val buf_dates = (s"git log -3  --format=%cd $pathToFile" !!)
-    val buf_messages = (s"git log -3  $pathToFile" !!)
+    val buf_dates = s"git log -3  --format=%cd $pathToFile" !!
```
janekdb: lgtm

tools/CopyrightDateTool.scala
```diff
@@ -99,19 +99,18 @@ object CopyrightDateTool {
 
   def analyzeDates(): Array[FileDates] = {
     val files_array = readFilesList()
-    var data = Array[FileDates]()
-    for(file_path <- files_array) {
+    val data = files_array map {
```
janekdb, line 110: lgtm

tools/CopyrightDateTool.scala
```diff
@@ -140,12 +139,11 @@ object CopyrightDateTool {
 
   def resultAnalyzeDates(): Boolean = {
     val data = analyzeDates
-    var check = false
-    val (error_files, ok_files) = data.partition(x => (x.isUpdateDateWrongStr() && (x.fileLastUpdateYear.forall(_.isDigit))))
-    if (error_files.nonEmpty) 
+    val (error_files, ok_files) = data.partition(x =>
```
janekdb, line 150: lgtm

tools/CopyrightDateTool.scala
```diff
@@ -154,18 +152,16 @@ object CopyrightDateTool {
       println("Oops...We have some error dates")
     }
     else println ("Everything is ok")
-    check
+    error_files.nonEmpty
```
janekdb, line 163: lgtm

tools/CopyrightDateTool.scala
```diff
@@ -154,18 +152,16 @@ object CopyrightDateTool {
...
   }
 
   def main(args: Array[String]) {
     if (args.length != 1) println(usage)
```
janekdb: If line 159 is left in then the usage details will be printed twice when the number of args is <> 1

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,167 @@
...
+        """)
+      error_files.foreach { file => println(file) }
+      println("Oops...We have some error dates")
+    }
+    else println ("Everything is ok")
+    error_files.nonEmpty
+  }
+
+  def main(args: Array[String]) {
+    if (args.length != 1) println(usage)
```
janekdb: If this line was removed it would prevent the usage message from being printed twice when the number of args is <> 1.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,167 @@
...
+      val argList = args.toList
```
janekdb: Fix indenting.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,167 @@
...
+  def getFileLastUpdateYear(pathToFile: String): String = {
+    val lb = "(c) 20XX-".length()
+    val ub = "(c) 20XX-XXXX".length()
+    val fileLastUpdateYear = getYear(pathToFile, lb, ub)
+    fileLastUpdateYear
+  }
+
+  def analyzeDates(): Array[FileDates] = {
+    val files_array = readFilesList()
+    val data = files_array map {
```
janekdb, line 110: lgtm

tools/CopyrightDateTool.scala
FelAl: We didn't discuss other badly formed notices here, but we are going to discuss unification of all notices(where it can be possible).

som-snytt:
My previous question was just: if the current copyright is `(c) 2005 Some Contributer` and then someone makes significant changes, instead of updating the date, do you add a clause like, `(c) 2016 Lightbend` or perhaps `Modifications (c) 2016 Lightbend` and so on? There are pockets of code in that boat. I don't mean the automated tool must handle that; but hopefully it won't facilitate making updates we don't intend.

janekdb:
@FelAl There's clearly scope for additional improvements in a separate PR. Perhaps you can extract the various additional activities that surfaced in this review and start a new PR that tackles these. Some that come to mind,
1. Ensure the tool does not make unintended edits (see @som-snytt 's scenario above)
2. Surveying the full set of copyright headers in use and proposing a single format
3. Expand a single year into a range as appropriate (see A)

A. _replace "(c) 2009" with "Copyright 2009-2016" or however it ought to be normalized_

Once you address the last set of review comments I will checkout locally and try the tool.

adriaanm:
It was a just question, @som-snytt -- the tool should be focussed on keeping the dates in synch with git. Potentially, it could also ensure the copyright message is uniform across the codebase. 

When I was asking about doing this on the CI, I was thinking it could be like a code formatter, where only the changed files are checked for conformance to the copyright header format and contents (date range only).

FelAl:
1) I've fixed last review comments
2) I've made two PRs with dates only
3) I've closed two old PRs(with dates)

tools/CopyrightDateTool.scala
FelAl: I don't see a huge difference. Now I'm splitting via "\n", and in this example(Stackoverflow) I must split via " " (just space). It doesn't reduce an amount of code.

FelAl:
1) I've fixed last review comments
2) I've made two PRs with dates only
3) I've closed two old PRs(with dates)

tools/CopyrightDateTool.scala
FelAl: Maybe I can do it in bash directly(year extraction), but it'll look a little ugly.

tools/CopyrightDateTool.scala
FelAl, line 99: Now It's one line in one file. Scaladoc sources, I think It's not `*.scala` files. I am checking `*.scala` only. I can reflect on different file extensions in the future.(new PRs etc.)

som-snytt, line 99: The scaladoc class in question is scala/tools/nsc/doc/html/page/Entity.scala which has code which could spuriously look like the copyright of the file itself:

```
      {
        if (Set("epfl", "EPFL").contains(tpl.universe.settings.docfooter.value))
          <div id="footer">Scala programming documentation. Copyright (c) 2003-2016 <a href="http://www.epfl.ch" target="_top">EPFL</a>, with contributions from <a href="http://www.lightbend.com" target="_top">Lightbend</a>.</div>
        else
          <div id="footer"> { tpl.universe.settings.docfooter.value } </div>
      }
```

som-snytt:
Ironically, and perhaps intentionally, the source file itself has no copyright notice.

tools/CopyrightDateTool.scala
FelAl, line 99: `Copyright (c) 2003-2016` - It's not a target copyright. It won't change.

FelAl:
Janek asked me if I would place copyrights, but I thought it was a joke. Do I have to place it now?

som-snytt:
That's the thing about ironic humor, you never know how to take it. Or rather, there's never just one way to take it.

som-snytt:
Naughty @retronym hasn't updated the (c) on Typers since the transition from the paulp era in late 2012. Heather did the massive "update ALL the (c)" at that time; but of course github won't display so many deltas, so arguably those huge commits obscure git history. Probably the tool has to operate either as a formatter or as a commit-time check to be useful.

FelAl:
`of course github won't display so many deltas`
If I understand you right, we can see any deltas in git.

`Probably the tool has to operate either as a formatter or as a commit-time check to be useful.`

```
    analyze - just shows if anything is ok
     update - file fixing

     From root folder:
     Usage: scala tools/CheckDates.scala analyze
            scala tools/CheckDates.scala update
```

analyze - commit-time check
update - a formatting method

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,166 @@
...
+        """)
+      error_files.foreach { file => println(file) }
+      println("Oops...We have some error dates")
+    }
+    else println ("Everything is ok")
+    error_files.nonEmpty
+  }
+
+  def main(args: Array[String]) {
+    val argList = args.toList
```
janekdb, line 167: LGTM

No other comments at this point. I'd like to checkout this out locally and try it before asking for the commits to be squashed down to one.

som-snytt:
Here's a version that leverages my suggestions. I found that running one log command and processing it is much faster; but `git log --follow` offers interesting results.

https://github.com/scala/scala/pull/5201

Besides the issue of formatting or normalizing, I discovered other issues around correctness, like what year do you choose on --follow; it's an interesting problem.

retronym:
This tool seems general/useful enough to contribute to sbt-header (it would solve: https://github.com/sbt/sbt-header/issues/35). There will be a bit more work to fit into the existing assumptions of that tool, but we'd be able to depend on the resulting plugin from our SBT build.

@som-snytt / @FelAl Would either of you be interested in making that happen?

FelAl:
@som-snytt / @FelAl Would either of you be interested in making that happen?

Yes, I am very interested in it and can do it.

som-snytt:
I see it's been on the radar for a while. https://github.com/sbt/sbt-header/issues/23

I can lend @FelAl a hand as needed.

som-snytt:
Their issue 31 is about which java to support. I wanted to assume java 8 and scala 2.12; a standalone script has greater freedom.

My other feedback is that it was fun doing a quick script, self-contained. I plan to do more, day-to-day. But not sure whether they will be ammonite scripts.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,166 @@
+import sys.process._
```
janekdb, line 9: Add copyright header.

tools/CopyrightDateTool.scala
janekdb: I do not understand why `CopyrighDateTool` is special cased and ignored.

tools/CopyrightDateTool.scala
```diff
@@ -0,0 +1,166 @@
...
+
+    This tool compares a copyright date in files with the date of last change in git.
+    If the message of last change contains "[Sync copyright dates]" it will be ignored and
+    we will take next commit.
+
+    analyze - just shows if anything is ok
+    update - file fixing
+
+    From root folder:
+    Usage: scala tools/CheckDates.scala analyze
```
janekdb: The class is now `CopyrightDateTool`,

janekdb:
@FelAl - It would be worth opening a later PR to improve the output of `analyze` by making it tabular. This is how it currently looks, correct but jagged,

```
21:38 $ ~/lib/scala-2.11.8/bin/scala tools/CopyrightDateTool.scala analyze
warning: there were three feature warnings; re-run with -feature for details
one warning found

        Format
        FileDates(file_path,file update year,git update year, file creation year, git creation year)

FileDates(src/actors/scala/actors/Actor.scala,2013,2014,2005,2006)
FileDates(src/actors/scala/actors/ActorCanReply.scala,2013,2012,2005,2010)
FileDates(src/actors/scala/actors/ActorProxy.scala,2013,2012,2005,2006)
FileDates(src/actors/scala/actors/ActorTask.scala,2013,2012,2005,2009)
FileDates(src/actors/scala/actors/Combinators.scala,2013,2012,2005,2010)
...
```

Again in a subsequent PR take a look at the runtime. Consider adopting some of the optimisations from #5201 . Failing that add some `println` to provide feedback every 20 files. For reference purpose here is my execution time on a VirtualBox Ubuntu 14.4 VM.

```
21:47 $ time ~/lib/scala-2.11.8/bin/scala tools/CopyrightDateTool.scala update
warning: there were three feature warnings; re-run with -feature for details
one warning found

real    5m14.192s
user    4m43.104s
sys 0m24.588s
```

janekdb:
Running the tool locally yields some changes that need explaining.

```
21:59 $ git diff -U0|grep -E -- '-\*\*'|sort|uniq -c
     96 -**    / __/ __// _ | / /  / _ |    (c) 2002-2013, LAMP/EPFL             **
    148 -**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
      2 -**    / __/ __// _ | / /  / _ |    (c) 2003-2015, LAMP/EPFL             **
     29 -**    / __/ __// _ | / /  / _ |    (c) 2005-2013, LAMP/EPFL             **
     22 -**    / __/ __// _ | / /  / _ |    (c) 2006-2013, LAMP/EPFL             **
      1 -**    / __/ __// _ | / /  / _ |    (c) 2006-2015, LAMP/EPFL             **
      5 -**    / __/ __// _ | / /  / _ |    (c) 2007-2013, LAMP/EPFL             **
      1 -**    / __/ __// _ | / /  / _ |    (c) 2008-2013, LAMP/EPFL             **
      5 -**    / __/ __// _ | / /  / _ |    (c) 2010-2013, LAMP/EPFL             **
```

```
21:59 $ git diff -U0|grep -E -- '\+\*\*'|sort|uniq -c
     37 +**    / __/ __// _ | / /  / _ |    (c) 2002-2012, LAMP/EPFL             **
      9 +**    / __/ __// _ | / /  / _ |    (c) 2002-2014, LAMP/EPFL             **
     21 +**    / __/ __// _ | / /  / _ |    (c) 2002-2015, LAMP/EPFL             **
     29 +**    / __/ __// _ | / /  / _ |    (c) 2002-2016, LAMP/EPFL             **
     10 +**    / __/ __// _ | / /  / _ |    (c) 2003-2012, LAMP/EPFL             **
     52 +**    / __/ __// _ | / /  / _ |    (c) 2003-2014, LAMP/EPFL             **
     77 +**    / __/ __// _ | / /  / _ |    (c) 2003-2015, LAMP/EPFL             **
     11 +**    / __/ __// _ | / /  / _ |    (c) 2003-2016, LAMP/EPFL             **
     19 +**    / __/ __// _ | / /  / _ |    (c) 2005-2012, LAMP/EPFL             **
      3 +**    / __/ __// _ | / /  / _ |    (c) 2005-2014, LAMP/EPFL             **
      7 +**    / __/ __// _ | / /  / _ |    (c) 2005-2015, LAMP/EPFL             **
      8 +**    / __/ __// _ | / /  / _ |    (c) 2006-2012, LAMP/EPFL             **
      4 +**    / __/ __// _ | / /  / _ |    (c) 2006-2014, LAMP/EPFL             **
     10 +**    / __/ __// _ | / /  / _ |    (c) 2006-2015, LAMP/EPFL             **
      1 +**    / __/ __// _ | / /  / _ |    (c) 2006-2016, LAMP/EPFL             **
      1 +**    / __/ __// _ | / /  / _ |    (c) 2007-2014, LAMP/EPFL             **
      3 +**    / __/ __// _ | / /  / _ |    (c) 2007-2015, LAMP/EPFL             **
      1 +**    / __/ __// _ | / /  / _ |    (c) 2007-2016, LAMP/EPFL             **
      1 +**    / __/ __// _ | / /  / _ |    (c) 2008-2015, LAMP/EPFL             **
      2 +**    / __/ __// _ | / /  / _ |    (c) 2010-2012, LAMP/EPFL             **
      2 +**    / __/ __// _ | / /  / _ |    (c) 2010-2014, LAMP/EPFL             **
      1 +**    / __/ __// _ | / /  / _ |    (c) 2010-2015, LAMP/EPFL             **
```

If I've managed to get my regexes correct the first block is all the lines that were updated, and the second the updated lines. Since 2012 appears in the updated files but is less than every end date in  the original files we need to analyse a particular case to be understand how we got there.

This is the first case in the diffs,

```
diff --git a/src/actors/scala/actors/ActorCanReply.scala b/src/actors/scala/actors/ActorCanReply.scala
index 07191ec..a11b758 100644
--- a/src/actors/scala/actors/ActorCanReply.scala
+++ b/src/actors/scala/actors/ActorCanReply.scala
@@ -1,6 +1,6 @@
 /*                     __                                               *\
 **     ________ ___   / /  ___     Scala API                            **
-**    / __/ __// _ | / /  / _ |    (c) 2005-2013, LAMP/EPFL             **
+**    / __/ __// _ | / /  / _ |    (c) 2005-2012, LAMP/EPFL             **
 **  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
 ** /____/\___/_/ |_/____/_/ | |                                         **
 **                          |/                                          **
```

This is time travelling commit. Note the year is `2012`.

```
commit 807dbe557a47b6944a7d352c0316bcd78733f473
Author: Heather Miller <heather.miller@epfl.ch>
Date:   Fri Nov 2 18:12:57 2012 +0100

    Brings all copyrights (in comments) up-to-date, from 2011/12 to 2013

```

The diff from that commit,

```
diff --git a/src/actors/scala/actors/ActorCanReply.scala b/src/actors/scala/actors/ActorCanReply.scala
index d92fb18..07191ec 100644
--- a/src/actors/scala/actors/ActorCanReply.scala
+++ b/src/actors/scala/actors/ActorCanReply.scala
@@ -1,6 +1,6 @@
 /*                     __                                               *\
 **     ________ ___   / /  ___     Scala API                            **
-**    / __/ __// _ | / /  / _ |    (c) 2005-2011, LAMP/EPFL             **
+**    / __/ __// _ | / /  / _ |    (c) 2005-2013, LAMP/EPFL             **
 **  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
 ** /____/\___/_/ |_/____/_/ | |                                         **
 **                          |/                                          **
```

So in conclusion the tool has correctly moved the end year back from 2013 to 2012.

Spot checks on three advanced years all pass.

janekdb:
@FelAl - There are two changes it would be good to see and then it will be looking good to me,
1. Add copyright header
2. Correct class name in `usage` string

som-snytt:
@janekdb My tool outputs `2013~>2012` for an end date that was in the future. The (really) interesting case was `AbstractPartialFunction` (see my commit comment) where `(c) 2011` was updated to `(c) 2013` by the Great Blast of 2013.

My experience was that analyzing why a file bears a date is extremely labor-intensive. My first assumption was that the date on `AbstractPartialFunction` was intentionally updated to show that it was rewritten during the `PartialFunction` refactor. But I was mistaken. So, in a word, the tool is useless if you have to review the numbers.

Similarly, it's obvious that `App` contains copyrighted material from `Application` (and git knows this); its copyright was incorrectly updated by hand. But git thinks `AbstractPartialFunction` contains earlier material; it does, but only the standard file header; so if the tool does `--follow` past the file creation, it has to exclude headers when deciding if the copyright should be earlier.

So Jason is right that this is better as an external tool. Let someone else iron out these kinks, rather than take on the maintenance burden here.

I do think the tool has to be higher quality. For instance, Jason's wip of sbt-header includes the unicode `(c)` in the regex.

Another possible feature: if A is cut/pasted to B, it should carry its copyright years forward and also any third-party copyrights (which should not be updated after the move).

Finally, I wonder if sbt-header is the best host as opposed to scalariform or scalafmt. Ideally, the processing is done only once; but maybe sbt-header is really a precursor step (maybe incurring the cost of git history) to pure formatting.

adriaanm:
I agree this would make for a great first step towards tooling for code formatting in scala/scala. Really exciting! Since this tooling will see heavy usage with how many PRs we get, I concur that the tool should be very robust and high quality. I really appreciate the development and the reviewing work that has gone into this already.

However, let's not over-engineer the copyright header year updater 😄  It's great to have them be accurate reflections of git history, but if the update introduces spurious changes, this advantage is lost. Also, from a legal perspective, while we should do our best to maintain correct copyright dates,  the crucial part is really the license file and the copyright notice itself.

For those reasons, it does sound like this is bigger than a scala/scala pull request, and collaboration may be more fruitful in a separate repository.

tools/CopyrightDateTool.scala
FelAl: A legacy from an old version. Now it is deleted.

FelAl:
So, squash commits?

janekdb:
Yes, squash commits.

som-snytt:
I'm not sure @janekdb is on the same page, as to whether to continue this PR or submit a PR to `sbt-header` or similar formatter.

FelAl:
I think I'll submit a new PR to `sbt-header`. @janekdb what do you think?

janekdb:
@adriaanm Could you clarify your view on the future of this PR?

My view is the tool as-is would improve on the current accuracy of the date ranges so should be considered for merge. Is there a view this could be a net loss in terms of accuracy if merged and used in the future? I may have missed that so a couple of examples of the problems caused would be useful.

The tools has bags of room for improvement and I agree that a smarter tool is a project in itself and deserves a different repo.

adriaanm:
I would prefer this work continues in an sbt plugin where it can be useful to all Scala projects, not just here in scala/scala. Now we're finally switching to an sbt build, it would be easy to add the sbt-header plugin, as suggested by @FelAl (thanks!).

janekdb:
@FelAl - The next step for this code on the way to life as an sbt plugin is to close this PR.

FelAl:
Ok

PR #4586, opened 2015-06-27
Improved message for missing argument list

som-snytt:
Clarifies the language and rules for eta-expansion.

A missing argument in a list, as opposed to a missing
argument list, results in a different message.

The comical expansion in parens does not attempt to show
what was already applied, but succeeds in showing at a
glance the shape of the method in question.

```
scala> def m(i: Int, j: Int)(x: Int) = ???
m: (i: Int, j: Int)(x: Int)Nothing

scala> m
<console>:12: error: missing argument list for method m
Unapplied methods are only converted to functions when a function type is expected.
You can make this conversion explicit by writing `m _` or `m(_,_)(_)` instead of `m`.
       m
       ^
```

The original submission #4555 was due to sschaef and the wording due
to adriaanm, with a minor tweak.

Review by @adriaanm

PR #5147, opened 2016-05-05
SI-8667 Improve too-many-args message

som-snytt:
Use removeNames to help diagnose the application.
Supplement the error message with how many extra
args and any other residual assignments that the
user might have thought was a properly named arg.

### DIFF
--- a/test/files/neg/protected-constructors.check
+++ b/test/files/neg/protected-constructors.check
@@ -1,6 +1,6 @@
 protected-constructors.scala:17: error: no arguments allowed for nullary constructor Foo1: ()dingus.Foo1
     val foo1 = new Foo1("abc")
-               ^
+                        ^
 protected-constructors.scala:18: error: constructor Foo2 in class Foo2 cannot be accessed in object P
  Access to protected constructor Foo2 not permitted because
  enclosing object P in package hungus is not a subclass of
@@ -19,4 +19,7 @@ protected-constructors.scala:15: error: class Foo3 in object Ding cannot be acce
  object Ding in package dingus where target is defined
     class Bar3 extends Ding.Foo3("abc")
                             ^
-four errors found
+protected-constructors.scala:15: error: no arguments allowed for nullary constructor Object: ()Object
+    class Bar3 extends Ding.Foo3("abc")
+                                 ^
+5 errors found
--- a/test/files/neg/protected-constructors.check
+++ b/test/files/neg/protected-constructors.check
@@ -1,4 +1,4 @@
-protected-constructors.scala:17: error: too many arguments for constructor Foo1: ()dingus.Foo1
+protected-constructors.scala:17: error: no arguments allowed for nullary constructor Foo1: ()dingus.Foo1
     val foo1 = new Foo1("abc")
                ^
 protected-constructors.scala:18: error: constructor Foo2 in class Foo2 cannot be accessed in object P
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -538,8 +538,33 @@ trait ContextErrors {
       def NamedAndDefaultArgumentsNotSupportedForMacros(tree: Tree, fun: Tree) =
         NormalTypeError(tree, "macro applications do not support named and/or default arguments")
 
-      def TooManyArgsNamesDefaultsError(tree: Tree, fun: Tree) =
-        NormalTypeError(tree, "too many arguments for "+treeSymTypeMsg(fun))
+      def TooManyArgsNamesDefaultsError(tree: Tree, fun: Tree, expected: Int, supplied: Int, unknowns: List[Name]) = {
+        val msg = {
+          val badappl = {
+            val excess = supplied - expected
+            val target = treeSymTypeMsg(fun)
+
+            if (expected == 0) s"no arguments allowed for nullary $target"
+            else if (excess < 3 && expected <= 5) s"too many arguments ($supplied) for $target"
+            else if (expected > 10) s"$supplied arguments but expected $expected for $target"
+            else {
+              val oneOf =
+                if (excess == 1) "one more argument"
+                else if (excess > 0) s"$excess more arguments"
+                else "too many arguments"
+              s"$oneOf than can be applied to $target"
+            }
+          }
+          val suppl = 
+            unknowns.size match {
+              case 0 => ""
+              case 1 => s"\nNote that '${unknowns.head}' is not a parameter name of the invoked method."
+              case _ => unknowns.mkString("\nNote that '", "', '", "' are not parameter names of the invoked method.")
+            }
+          s"${badappl}${suppl}"
+        }
+        NormalTypeError(tree, msg)
+      }
 
       // can it still happen? see test case neg/overloaded-unapply.scala
       def OverloadedUnapplyError(tree: Tree) =
@@ -551,7 +576,7 @@ trait ContextErrors {
       def MultipleVarargError(tree: Tree) =
         NormalTypeError(tree, "when using named arguments, the vararg parameter has to be specified exactly once")
 
-      def ModuleUsingCompanionClassDefaultArgsErrror(tree: Tree) =
+      def ModuleUsingCompanionClassDefaultArgsError(tree: Tree) =
         NormalTypeError(tree, "module extending its companion class cannot use default constructor arguments")
 
       def NotEnoughArgsError(tree: Tree, fun: Tree, missing: List[Symbol]) = {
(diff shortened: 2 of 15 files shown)


janekdb:
This is an improvement but `25 too many arguments for method` could be regarded as the the more convoluted precursor to `25 more arguments than expected for method`.

Compare,

| A | B |
| --- | --- |
| one too many arguments for method | one more argument than expected for method |
| 2 too many arguments for method | 2 more arguments than expected for method |
| 11 too many arguments for method | 11 more arguments than expected for method |
| 25 too many arguments for method | 25 more arguments than expected for method |

som-snytt:
I think "expected" isn't the right word, because of defaults and tupling.

`25 more arguments than can be applied to method f: (i: Int)Int`

janekdb:
That formulation looks good.

som-snytt:
I'll miss the colloquial terseness of "too many args", and why did it even spell out "arguments", but thanks for the nudge.

som-snytt:
I didn't squash, in case someone prefers this terser version:

```
t8667.scala:13: error: too many args (4) for constructor C: (a: Int, b: Int)C ; c, d are not parameters

//reduces counting but demands mental subtraction
t8667.scala:13: error: too many args (4 supplied but requires 2) for constructor C: (a: Int, b: Int)C ; c, d are not parameters

// instead of now
t8667.scala:13: error: 2 more arguments than can be applied to constructor C: (a: Int, b: Int)C ; c, d are not parameters
```

som-snytt:
I squashed with an update for gradual messages that depend on the arg count. Off by one for a short param list is different from off by 5 for a 22-arity generated function.

Also made the "not a param name" message the same as for missing args.

retronym:
Here's the current Javac version for comparison's sake:

```
% cat sandbox/Test.java  && javac sandbox/Test.java
class Test {
    void foo(int a) {};
    void bar() {
        foo(1, 2, 3);
    }
}
sandbox/Test.java:4: error: method foo in class Test cannot be applied to given types;
        foo(1, 2, 3);
        ^
  required: int
  found: int,int,int
  reason: actual and formal argument lists differ in length
1 error
```

retronym:
An alternative (or additional ) idea: should be focus the error message at the first superfluous argument, rather than at the invoked method.

som-snytt:
I'd forgotten that missing-args was fiddled with a year ago: https://github.com/scala/scala/pull/4586

som-snytt:
Note to self, despite ambiguity of assignment, call out the erroneous `k=1`:

```
scala> def f(i: Int, j: Int) = i + j
f: (i: Int, j: Int)Int

scala> f(1,2,3)
<console>:13: error: too many arguments (3) for method f: (i: Int, j: Int)Int
       f(1,2,3)
             ^

scala> f(k = 1,i=2,j=3)
<console>:13: error: parameter 'i' is already specified at parameter position 1
Note that 'k' is not a parameter name of the invoked method.
       f(k = 1,i=2,j=3)
                ^
<console>:13: error: too many arguments (3) for method f: (i: Int, j: Int)Int
Note that 'k' is not a parameter name of the invoked method.
       f(k = 1,i=2,j=3)
                    ^
```

som-snytt:
My first job had me writing up the appendix of compiler error codes:

http://www.aztecmuseum.ca/docs/Aztec_C86_v3_2d_for_CPM86_Manual.pdf

I like that I wrote out "zero through seven", but other locutions are painfully awkward.

1: bad digit in octal constant
The only numerals permitted in the base 8 (octal) counting system are zero through
 seven.  In order to distinguish between octal, hexadecimal, and decimal constants, octal constants are preceded by a zero.  Any number beginning with a zero must not contain a digit greater than seven.  Octal constants look like this: 01, 027, 003.  Hexadecimal constants begin with 0x (e.g., 0x1, 0xAA0, 0xFFF).
2: string space exhausted
The compiler maintains an internal table of the strings appearing in the source code.  Since this table has a finite size, it may overflow during compilation and cause this error code.  The table default size is about one or two thousand characters depending on the operating system.  The size can be changed using the compiler option -Z.  Through simple guesswork, it is possible to arrive at a table size sufficient for compiling your program.
3: unterminated string
All strings must begin and end with double quotes (“).   This message indicates that a double quote has remained unpaired.
4: internal error
This error message should not occur.  It is a check on the internal workings of the compiler and is not known to be caused by any particular piece of code.  However, if this error code appears, please bring it to the attention of MANX.  It could be a bug in the compiler.  The release documentation enclosed with the product contains further information.
5:  illegal type for function
The type of a function refers to the type of the value which it returns.  Functions return an 
int by default unless they are declared otherwise.  However, functions are not allowed to return aggregates (arrays or structures).  An attempt to write a function such as  struct sam func() will generate this error code.  The legal function types are  char,  int, float, double,unsigned, long, void and a pointer to any type (including structures).
6:  error in argument declaration
The declaration list for the formal parameters of a function stands immediately before the left brace of the function body, as shown below.  Undeclared arguments default to int,  though it is usually better practice to declare everything. Naturally, this declaration list may be empty, whether or not the function takes any arguments at all.


11: data type too complex
This message is best explained by example:
  char *******foo;
The form of this declaration implies six pointers to pointers.   The seventh asterisk indicates a pointer to a char.   The compiler is unable to keep track of so many “levels”.  Removing just one of the asterisks will cure the error; all that is being declared in any case is a single two byte pointer. However it is to be hoped that such a construct will never be needed.

lrytz:
looks good to me - @janekdb @retronym can you take a look if you also like the new messages? https://github.com/scala/scala/pull/5147/files#diff-159fe691bd78013ef009e40722925f13

test/files/neg/protected-constructors.check
```diff
@@ -1,4 +1,4 @@
-protected-constructors.scala:17: error: too many arguments for constructor Foo1: ()dingus.Foo1
+protected-constructors.scala:17: error: no arguments allowed for nullary constructor Foo1: ()dingus.Foo1
```
janekdb, line 1: +1

janekdb:
Exceeds my expectations so have a √LGTM leaving the deciding vote to @retronym.

som-snytt:
I haven't pushed retronym's idea to push the caret to the bad arg.

som-snytt:
Rebased. Moves the caret across the goal line.

Also noted illustrious history https://issues.scala-lang.org/browse/SI-3818

test/files/neg/protected-constructors.check
```diff
@@ -19,4 +19,7 @@ protected-constructors.scala:15: error: class Foo3 in object Ding cannot be acce
  object Ding in package dingus where target is defined
     class Bar3 extends Ding.Foo3("abc")
                             ^
-four errors found
+protected-constructors.scala:15: error: no arguments allowed for nullary constructor Object: ()Object
+    class Bar3 extends Ding.Foo3("abc")
+                                 ^
+5 errors found
```
som-snytt, line 25: What's the fifth error?

som-snytt, line 25: Did it start showing up because the position changed?

som-snytt, line 25: Yes, in fact under `-Ydebug` it was:

```
test/files/neg/protected-constructors.scala:15: error: [ suppressed ] too many arguments for primary constructor Object: ()lang.this.Object
    class Bar3 extends Ding.Foo3("abc")
                            ^
```

som-snytt:
@retronym I moved the caret and a suppressed meaningless error is emitted. Is that a deal breaker? I didn't see a quick way to detect "I'd be suppressed at this other position."  Lo and behold, a decoupled interface with Reporter is actually inconvenient if I can't interrogate it. Or add API, `error(pos, msg)(suppressedByErrorAt = otherpos)`. Maybe the semantics is, "my underlying cause is at a different pos, at which suppression is computed".

adriaanm:
The Reporter API is certainly open to evolution (to simplify the process, we could start with an internal subtype that has richer methods than the API used by sbt etc).

som-snytt:
Positions can be shifted, as in scripts that begin at an offset into the source file.

Analogously, you'd like to say, `error(pos withShiftedFocus displayFocus, "too bad")`. Then the error-suppressing reporter can check both the shifted focus and the underlying `pos`.

adriaanm:
LGTM

som-snytt:
I'll follow up in future about the spurious error. I didn't see where the inaccessible super ctor results in using nullary ctor. You'd hope the error wouldn't cascade. But I also like that last idea, where an error position (for display) has an underlying unshifted position (for semantics).

retronym:
> But I also like that last idea, where an error position (for display) has an underlying unshifted position (for semantics).

You could model something like that with the current API by issuing another error at the old position with a special message (e.g. empty string) that we could take as meaning it shouldn't be displayed.

PR #5149, opened 2016-05-09
Fix compilation errors in demo code in 02-identifiers-names-and-scope…

jacobsallan:
Problem: Example code at the end of Chapter 2 does not compile.

Solution: Remove the TODO at line 37.  The spec and compiler passed all 10 tests thrown at it.  The example is the problem.

Part of the problem with the code is that the spec reader has to know what a compilation unit is and to infer that the definitions for X in packages P and Q belong in a different file from the rest of the example.  Text is added to the specification to clue the reader in.

The function abc is added at line 18 because the compiler complains  “error: Unit does not take parameters”.  A main method was added at the end to check the compilation and run time behavior.  To make the function reachable, an object D was placed at line 15.

### DIFF
--- a/spec/02-identifiers-names-and-scopes.md
+++ b/spec/02-identifiers-names-and-scopes.md
@@ -34,8 +34,6 @@ in some inner scope _shadows_ bindings of lower precedence in the
 same scope as well as bindings of the same or lower precedence in outer
 scopes.
 
-<!-- TODO: either the example, the spec, or the compiler is wrong
-
 Note that shadowing is only a partial order. In a situation like
 
 ```scala
@@ -48,7 +46,6 @@ val x = 1
 
 neither binding of `x` shadows the other. Consequently, the
 reference to `x` in the last line of the block above would be ambiguous.
--->
 
 A reference to an unqualified (type- or term-) identifier $x$ is bound
 by the unique binding, which
@@ -84,7 +81,9 @@ package Q {
 ```
 
 The following program illustrates different kinds of bindings and
-precedences between them.
+precedences between them.  It is assumed that the definitions for the
+objects names named `X` in packages `P` and `Q` are made in a different
+file (a different compilation unit) than that of the program.
 
 ```scala
 package P {                  // `X' bound by package clause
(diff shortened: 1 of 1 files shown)


som-snytt:
The example you uncommented also needs to be fixed. It used to have a semicolon. I'd suggest using `locally` here and in the large example. I think adding `main` is unnecessary, since you just paste into the REPL and evaluate `P.A`, `P.A.B` etc. Since `B` and `C` aren't used for imports, they could also be converted to just `locally`. Then just `P.A` causes evaluation. `Q.X.y` should be something more visible than empty string.

```
val x = 1
locally {
  import P.X.x
  x
}
```

spec/02-identifiers-names-and-scopes.md
```diff
@@ -84,7 +81,9 @@ package Q {
 ```
 
 The following program illustrates different kinds of bindings and
-precedences between them.
+precedences between them.  It is assumed that the definitions for the
+objects names named `X` in packages `P` and `Q` are made in a different
+file (a different compilation unit) than that of the program.
```
som-snytt, line 86: Grammar: `names`. I would add the comment to the line about 

```
Assume the following two definitions of objects named `X` in packages `P` and `Q` in a separate compilation unit.
```

som-snytt:
Renamed to `Y` by contrast to `X`:

```
scala> :pa -raw

package P {                  // `X' bound by package clause
import Console._             // `println' bound by wildcard import
object Y {
  println("L4: "+X)          // `X' refers to `P.X' here
  locally {
    import Q._               // `X' bound by wildcard import
    println("L7: "+X)        // `X' refers to `Q.X' here
    import X._               // `x' and `y' bound by wildcard import
    println("L8: "+x)        // `x' refers to `Q.X.x' here
    locally {
      val x = 3              // `x' bound by local definition
      println("L12: "+x)     // `x' refers to constant `3' here
      locally {
        import Q.X._         // `x' and `y' bound by wildcard import
//      println("L14: "+x)   // reference to `x' is ambiguous here
        import X.y           // `y' bound by explicit import
        println("L16: "+y)   // `y' refers to `Q.X.y' here
        locally {
          val x = "abc"      // `x' bound by local definition
          import P.X._       // `x' and `y' bound by wildcard import
//        println("L19: "+y) // reference to `y' is ambiguous here
          println("L20: "+x) // `x' refers to string "abc" here
}}}}}}

scala> P.Y
L4: P.X$@290dbf45
L7: Q.X$@145eaa29
L8: true
L12: 3
L16: 
L20: abc
res4: P.Y.type = P.Y$@3bb9a3ff
```

som-snytt:
Since this is tedious, I went ahead and https://github.com/scala/scala/pull/5151

That includes the clarifying language for SI-2458 which is the same section.

Edit: by "this is tedious", I meant that the work to tweak the language is tedious; sorry I didn't start with your commit, I was too lazy to fetch it.

lrytz:
closing this in favor of #5151

PR #5150, opened 2016-05-09
SI-9397 Add "_root_" to "GenUtils.scalaFactoryCall" to avoid the scala package name conflits

zsxwing:
When a user imports some package ending with `scala`, the macro expansion of TypeTag may not work because it uses `scala.collection.immutable.List` but it's overrided (see the example in [SI-9397](https://issues.scala-lang.org/browse/SI-9397)). This patch adds the `_root_` prefix to avoid the scala package name conflits. Of cause, after this fix, importing a package ending with `_root_` has the same issue. However, people rarely do that.


zsxwing:
@xeno-by could you take a look, please?

lrytz:
LGTM

PR #3626, opened 2014-03-13
SI-2458 Explicit import should shadow def in other comp. unit

retronym:
Fixes the implementation of point 4. from SLS chapter 2:

> Definitions made available by a package clause not in the
> compilation unit where the definition occurs have lowest precedence.

We had code to deal with this:

```
if (defSym.exists && impSym.exists) {
  if (isPackageOwnedInDifferentUnit(defSym))
    defSym = NoSymbol
```

But, in the `pos/t2458a`, no `impSym` was around, due to filtering
in `depthOk`. The depth of the imported symbol was the same as the
depth of the defined symbol, which didn't pass:

```
imp.depth > symbolDepth
```

`depthOk` already had special handling for Java sources: if the
import is non-wildcard and depths are equal, retain the import.

This commit does the same thing for Scala sources, using the
`isPackageOwnedInDifferentUnit` as the criteria. It also makes
the check for the "imported ... is permanently hidden" warning
to avoid this spurious warning:

```
test/files/pos/t2458a/A.scala:3: warning: imported `BitSet' is permanently hidden by definition of class BitSet in package p
import scala.collection.BitSet
                    ^
```

I have retained compatibiltiy with the bug under -Xsource:2.10

Review by @adriaanm. I'l trigger a dbuild of this one to check for fallout.

PR #5151, opened 2016-05-09
SI-2458 Clarify spec for package syms

som-snytt:
Package definitions are priority 4.

Update the big example to be cut/pastable.

### DIFF
--- a/spec/02-identifiers-names-and-scopes.md
+++ b/spec/02-identifiers-names-and-scopes.md
@@ -17,12 +17,12 @@ which are collectively called _bindings_.
 Bindings of different kinds have a precedence defined on them:
 
 1. Definitions and declarations that are local, inherited, or made
-   available by a package clause in the same compilation unit where the
-   definition occurs have highest precedence.
+   available by a package clause and also defined in the same compilation unit
+   as the reference, have highest precedence.
 1. Explicit imports have next highest precedence.
 1. Wildcard imports  have next highest precedence.
-1. Definitions made available by a package clause not in the
-   compilation unit where the definition occurs have lowest precedence.
+1. Definitions made available by a package clause, but not also defined in the
+   same compilation unit as the reference, have lowest precedence.
 
 There are two different name spaces, one for [types](03-types.html#types)
 and one for [terms](06-expressions.html#expressions). The same name may designate a
@@ -34,21 +34,17 @@ in some inner scope _shadows_ bindings of lower precedence in the
 same scope as well as bindings of the same or lower precedence in outer
 scopes.
 
-<!-- TODO: either the example, the spec, or the compiler is wrong
-
-Note that shadowing is only a partial order. In a situation like
+Note that shadowing is only a partial order. In the following example,
+neither binding of `x` shadows the other. Consequently, the
+reference to `x` in the last line of the block is ambiguous.
 
 ```scala
 val x = 1
-{
-  import p.x
+locally {
+  import p.X.x
   x
 }
 ```
-
-neither binding of `x` shadows the other. Consequently, the
-reference to `x` in the last line of the block above would be ambiguous.
--->
 
 A reference to an unqualified (type- or term-) identifier $x$ is bound
 by the unique binding, which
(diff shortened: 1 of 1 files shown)


lrytz:
can you remind me of the current state of the discussions related to SI-2458?
- is there consensus that the current spec is the way it should be?
- is the current implementation in scala still buggy? it seems https://github.com/scala/scala/pull/3626 was never merged.
- does dotty implement the current spec (https://github.com/scala/scala-dist/pull/96#issuecomment-34149582)?

spec/02-identifiers-names-and-scopes.md
```diff
@@ -69,43 +65,59 @@ the member of the type $T$ of $e$ which has the name $x$ in the same
...
 ```
 
 The following program illustrates different kinds of bindings and
 precedences between them.
 
 ```scala
-package P {                  // `X' bound by package clause
-import Console._             // `println' bound by wildcard import
-object A {
-  println("L4: "+X)          // `X' refers to `P.X' here
```
lrytz: lower-case `p` in comment

spec/02-identifiers-names-and-scopes.md
```diff
@@ -69,43 +65,59 @@ the member of the type $T$ of $e$ which has the name $x$ in the same
...
-  object B {
-    import Q._               // `X' bound by wildcard import
-    println("L7: "+X)        // `X' refers to `Q.X' here
```
lrytz: lower-case `q` in comment, also below

spec/02-identifiers-names-and-scopes.md
```diff
@@ -69,43 +65,59 @@ the member of the type $T$ of $e$ which has the name $x$ in the same
...
-        import X.y           // `y' bound by explicit import
-        println("L16: "+y)   // `y' refers to `Q.X.y' here
-        { val x = "abc"      // `x' bound by local definition
-          import P.X._       // `x' and `y' bound by wildcard import
-//        println("L19: "+y) // reference to `y' is ambiguous here
-          println("L20: "+x) // `x' refers to string "abc" here
+package p {                   // `X' bound by package clause
+import Console._              // `println' bound by wildcard import
+object Y {
+  println(s"L4: $X")          // `X' refers to `P.X' here
```
lrytz: lower-case `p` in comment, similar for `q` below

lrytz:
There's also an ambiguity in the numbered list of precedences above https://github.com/scala/scala/pull/5151/files#diff-07166d892e8021d88f7e2925b693f30bR19 - in points 1 and 4 "where the definition occurs" should be replaced by "where the reference occurs". This would fit well into this PR :)

som-snytt:
Retronym's branch is still an easy rebase, but requires a tweak. 

https://github.com/som-snytt/scala/tree/ticket/2458

Yes, there is consensus. Martin said the impl is "morally" wrong, and then Jason's doubts were assuaged. I haven't tried dotty.

spec/02-identifiers-names-and-scopes.md
som-snytt: thx, you're an excellent reviewer. Ditto for the other PRs.

som-snytt:
Truthfully, I have to take a weekend retreat to consider the phrase, "made available by a package clause [not] in the compilation unit where the definition occurs."

The gist is that another compilation unit elects to define symbols and they become visible to me because I have some package statements `package p1 { package p2.p3 { ... } }` but that shouldn't screw up my imports.

The package clause might be my `package p1`, which makes `p1._` visible. It is my package clause that does that. If the definition of stuff in p1 is in a different comp unit, then low priority. But if I define some `p1.X` in my comp unit, then high priority. So it means to say definition and not reference.

I'm distracted by day job, so I may be totally upside-down, of course.

som-snytt:
BTW, the canonical example would be: I'm defining collections, with lots of naming collisions, and I'm writing a Map that needs java.util.List for buckets so I import it, and I want that List and not someone's List from the package in which I happen to dwell. This is a deep question about modularity.

lrytz:
> it means to say definition and not reference

you're most likely right. i need more time to understand what the spec is trying to tell me..

lrytz:
i give up. last item is: "Definitions made available by a package clause not in the compilation unit where the definition occurs". what does "the definition" refer to?

does it mean "Definitions in a different compilation unit than the current, made available through a package clause"?

som-snytt:
:) The ambiguity for me, I think I realized, was that "not in the c.u." doesn't modify "package clause" but "definitions". b/c obviously only my package clauses make things visible to me.

The package clause is always in the current c.u. The definition might be in the same c.u. (ours) or in a different c.u. (somebody else's).

I bet there's a simple way to reword that so it's not mind-bending. There may be no reason to make it sound harder than it is.

I see I have to add my sympathetic smiley emoji to your comment, and not in-line.

lrytz:
so you agree with my re-wording above? at least to me it looks more clear.

som-snytt:
Yes. I didn't have a chance to revisit yet.

som-snytt:
I used your word "reference" as most clear. I also added a clarifying line near the bottom, since this is not obvious behavior, and the other clarifying line is also about non-obvious behavior.

lrytz:
LGTM!

som-snytt:
The actual fix from retronym's branch underway at https://github.com/scala/scala/pull/5339

PR #5153, opened 2016-05-10
SI-5463 Check .jars before using them

petermz:
This patch makes compiler ignore broken JAR files on classpath, the same way nonexistent JARs are ignored. When an exception occurs while processing a JAR, we treat it as containing no packages and no classes.

### DIFF
--- a/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
+++ b/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
@@ -6,6 +6,7 @@ package scala.tools.nsc.classpath
 import java.net.URL
 import scala.annotation.tailrec
 import scala.collection.mutable.ArrayBuffer
+import scala.reflect.internal.FatalError
 import scala.reflect.io.AbstractFile
 import scala.tools.nsc.util.ClassPath
 import scala.tools.nsc.util.ClassRepresentation
@@ -72,7 +73,16 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
     getDistinctEntries(_.sources(inPackage))
 
   override private[nsc] def list(inPackage: String): ClassPathEntries = {
-    val (packages, classesAndSources) = aggregates.map(_.list(inPackage)).unzip
+    val (packages, classesAndSources) = aggregates.map { cp =>
+      try {
+        cp.list(inPackage)
+      } catch {
+        case ex: java.io.IOException =>
+          val e = new FatalError(ex.getMessage)
+          e.initCause(ex)
+          throw e
+      }
+    }.unzip
     val distinctPackages = packages.flatten.distinct
     val distinctClassesAndSources = mergeClassesAndSources(classesAndSources: _*)
     ClassPathEntries(distinctPackages, distinctClassesAndSources)
--- a/test/files/run/t5463.scala
+++ b/test/files/run/t5463.scala
@@ -1,0 +1,21 @@
+import scala.reflect.internal.FatalError
+import scala.tools.partest.DirectTest
+
+object Test extends DirectTest {
+
+  def code = "class A"
+
+  override def show(): Unit = {
+    // Create a broken JAR file and put it on compiler classpath
+    val jarpath = testOutput.path + "/notajar.jar"
+    scala.reflect.io.File(jarpath).writeAll("This isn't really a JAR file")
+
+    val classpath = List(sys.props("partest.lib"), jarpath, testOutput.path) mkString sys.props("path.separator")
+    try {
+      compileString(newCompiler("-cp", classpath, "-d", testOutput.path))(code)
+      throw new Error("Compilation should have failed");
+    } catch {
+      case ex: FatalError => // this is expected
+    }
+  }
+}


src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
```diff
@@ -72,7 +72,13 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
     getDistinctEntries(_.sources(inPackage))
 
   override private[nsc] def list(inPackage: String): ClassPathEntries = {
-    val (packages, classesAndSources) = aggregates.map(_.list(inPackage)).unzip
+    val (packages, classesAndSources) = aggregates.map { cp =>
+      try {
+        cp.list(inPackage)
+      } catch {
+        case _: java.io.IOException => ClassPathEntries(Seq(), Seq())
```
mpociecha: We should log the error in that case, I think.

petermz: My idea was to handle broken JARs the same way nonexistent JARs are handled. E.g. when there's a typo in the name of a JAR file the compiler silently ignores it -- I thought it made sense to do the same for broken JARs

mpociecha: Ok. Let's see what others will say. :)

petermz:
review @mpociecha @lrytz 
Michal, Lukas -- let me add you as reviewers as most recent committers to this area. Let me know if I should add someone else

janekdb:
As a user my preference would be to fail hard when a broken jar is found. There is a possibility this silent corrective behaviour will successfully mask the broken jar problem but result in a different symptom of the same problem further down the chain.

If I'm using `scalac` then I'm on a development environment so I can fix a broken jar and would want to know about it as early as possible.

Can you outline a production scenario where skipping a broken jar would be the correct course of action?

lrytz:
I agree with @janekdb, I think there should be a fatal error when a broken jar is on the classpath. Currently the error message for an invalid `A.jar` looks like this:

```
error: error while loading <root>, Error accessing A.jar
error: scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.
    at scala.reflect.internal.MissingRequirementError$.signal(MissingRequirementError.scala:17)
    at scala.reflect.internal.MissingRequirementError$.notFound(MissingRequirementError.scala:18)
    at scala.reflect.internal.Mirrors$RootsBase.scala$reflect$internal$Mirrors$RootsBase$$$anonfun$4(Mirrors.scala:54)
```

SethTisue:
I think I agree with Janek and Lukas. but @petermz, I'm wondering, how does javac handle this situation?

petermz:
@SethTisue 

```
> javac -jar notajar.jar J.java 
Error: Invalid or corrupt jarfile notajar.jar
```

petermz:
I've updated the patch to make it a fatal error when a broken jar is encountered. The output now is

```
error: scala.reflect.internal.FatalError: Error accessing notajar.jar
    at scala.tools.nsc.classpath.AggregateClassPath.scala$tools$nsc$classpath$AggregateClassPath$$$anonfun$7(AggregateClassPath.scala:80)
    at scala.collection.TraversableLike.scala$collection$TraversableLike$$$anonfun$3(TraversableLike.scala:234)
    at scala.collection.Iterator.foreach(Iterator.scala:932)
    ...
```

It's only marginally better than what we have now, but at least it doesn't mention `object java.lang.Object in compiler mirror` anymore

test/files/run/t5463.scala
```diff
@@ -0,0 +1,20 @@
...
+  def code = "class A"
+
+  override def show(): Unit = {
+    // Create a broken JAR file and put it on compiler classpath
+    val jarpath = testOutput.path + "/notajar.jar"
+    scala.reflect.io.File(jarpath).writeAll("This isn't really a JAR file")
+
+    val classpath = List(sys.props("partest.lib"), jarpath, testOutput.path) mkString sys.props("path.separator")
+    try {
+      compileString(newCompiler("-cp", classpath, "-d", testOutput.path))(code)
```
lrytz, line 15: could you add `throw new Error()` here, as to test that the error is actually thrown?

petermz:
Test updated to ensure that the error is actually thrown.

test/files/run/t5463.scala
petermz, line 15: Sure. Patch updated

lrytz:
LGTM

src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
```diff
@@ -72,7 +73,13 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
     getDistinctEntries(_.sources(inPackage))
 
   override private[nsc] def list(inPackage: String): ClassPathEntries = {
-    val (packages, classesAndSources) = aggregates.map(_.list(inPackage)).unzip
+    val (packages, classesAndSources) = aggregates.map { cp =>
+      try {
+        cp.list(inPackage)
+      } catch {
+        case ex: java.io.IOException => throw new FatalError(ex.getMessage)
```
retronym: This discards potentially important information (the stack trace and class of `ex`).

lrytz: right.. how about `val e = new FatalError(ex.getMessage); e.initCause(ex)`

petermz:
Ok, fixed the fatal error to carry the original exception

mpociecha:
Thanks.

PR #5157, opened 2016-05-12
Lambda impl methods static and more stably named

retronym:
The body of lambdas is compiled into a synthetic method
in the enclosing class. Previously, this method was a public
virtual method named `fully$qualified$Class$$anonfun$n`.
For lambdas that didn't capture a `this` reference, a static
method was used.

This commit changes two aspects.

Firstly, all lambda impl methods are now emitted static.
An extra parameter is added to those that require a this
reference.

This is an improvement as it:
- allows, shorter, more readable names for the lambda impl method
- avoids pollution of the vtable of the class. Note that javac uses
  private instance methods, rather than public static methods. If
  we followed its lead, we would be unable to support important use
  cases in our inliner

Secondly, the name of the enclosing method has been included in
the name of the lambda impl method to improve debuggability and
to improve serialization compatibility. The serialization improvement
comes from the way that fresh names for the impl methods are
allocated: adding or removing lambdas in methods not named "foo" won't
change the numbering of the `anonfun$foo$n` impl methods from methods
named "foo". This is in line with user expectations about anonymous
class and lambda serialization stability. Brian Goetz has described
this tricky area well in:

  http://cr.openjdk.java.net/~briangoetz/eg-attachments/lambda-serialization.html

This commit doesn't go as far a Javac, we don't use the hash of the
lambda type info, param names, etc to map to a lambda impl method name.
As such, we are more prone to the type-1 and -2 failures described there.
However, our Scala 2.11.8 has similar characteristics, so we aren't going
backwards.

Special case in the naming: Use `"new"` rather than `"<init>"` for constructor enclosed
lambdas, as javac does.

Review by @lrytz

I am still looking at one of your earlier review comments:
- [x] don't we need a recursive call to nonLocalEnclosingMember here? i.e., `if (sym.isLocalToBlock) nonLocalEnclosingMember(sym.originalOwner) else sym`.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/TreeGen.scala
+++ b/src/compiler/scala/tools/nsc/ast/TreeGen.scala
@@ -344,7 +344,7 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     val newSym = maybeClone(orig.symbol)
     newSym.setFlag(STATIC)
     // Add an explicit self parameter
-    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF)
+    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF).setFlag(ARTIFACT)
     newSym.updateInfo(newSym.info match {
       case mt @ MethodType(params, res) => copyMethodType(mt, selfParamSym :: params, res)
     })
--- a/src/compiler/scala/tools/nsc/ast/TreeGen.scala
+++ b/src/compiler/scala/tools/nsc/ast/TreeGen.scala
@@ -283,7 +283,7 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     }
     val ownerName = nonLocalEnclosingMember(fun.symbol.originalOwner).name match {
       case nme.CONSTRUCTOR => nme.NEWkw // do as javac does for the suffix, prefer "new" to "$lessinit$greater$1"
-      case x => x
+      case x => x.dropLocal
     }
     val newName = nme.ANON_FUN_NAME.append(nme.NAME_JOIN_STRING).append(ownerName)
     mkMethodForFunctionBody(localTyper)(owner, fun, newName)(additionalFlags = ARTIFACT)
--- a/src/compiler/scala/tools/nsc/ast/TreeGen.scala
+++ b/src/compiler/scala/tools/nsc/ast/TreeGen.scala
@@ -238,7 +238,8 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
    *  (outside the synchronized block).
    *
    *  The idiom works only if the condition is using a volatile field.
-   *  @see http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
+    *
+    *  @see http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
    */
   def mkSynchronizedCheck(clazz: Symbol, cond: Tree, syncBody: List[Tree], stats: List[Tree]): Tree =
     mkSynchronizedCheck(mkAttributedThis(clazz), cond, syncBody, stats)
@@ -274,8 +275,19 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
   }
 
   // used to create the lifted method that holds a function's body
-  def mkLiftedFunctionBodyMethod(localTyper: analyzer.Typer)(owner: Symbol, fun: Function) =
-    mkMethodForFunctionBody(localTyper)(owner, fun, nme.ANON_FUN_NAME)(additionalFlags = ARTIFACT)
+  def mkLiftedFunctionBodyMethod(localTyper: global.analyzer.Typer)(owner: global.Symbol, fun: global.Function) = {
+    def nonLocalEnclosingMember(sym: Symbol): Symbol = {
+      if (sym.isLocalDummy) sym.enclClass.primaryConstructor
+      else if (sym.isLocalToBlock) nonLocalEnclosingMember(sym.originalOwner)
+      else sym
+    }
+    val ownerName = nonLocalEnclosingMember(fun.symbol.originalOwner).name match {
+      case nme.CONSTRUCTOR => nme.NEWkw // do as javac does for the suffix, prefer "new" to "$lessinit$greater$1"
+      case x => x
+    }
+    val newName = nme.ANON_FUN_NAME.append(nme.NAME_JOIN_STRING).append(ownerName)
+    mkMethodForFunctionBody(localTyper)(owner, fun, newName)(additionalFlags = ARTIFACT)
+  }
 
 
   /**
@@ -308,6 +320,38 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     val moveToMethod = new ChangeOwnerTraverser(fun.symbol, methSym)
 
     newDefDef(methSym, moveToMethod(useMethodParams(fun.body)))(tpt = TypeTree(resTp))
+  }
+
+  /**
+    * Create a new `DefDef` based on `orig` with an explicit self parameter.
+    *
+    * Details:
+    *   - Must by run after erasure
+    *   - If `maybeClone` is the identity function, this runs "in place"
+    *     and mutates the symbol of `orig`. `orig` should be discarded
+    *   - Symbol owners and returns are substituted, as are parameter symbols
+    *   - Recursive calls are not rewritten. This is correct if we assume
+    *     that we either:
+    *       - are in "in-place" mode, but can guarantee that no recursive calls exists
+    *       - are associating the RHS with a cloned symbol, but intend for the original
+    *         method to remain and for recursive calls to target it.
+    */
+  final def mkStatic(orig: DefDef, maybeClone: Symbol => Symbol): DefDef = {
+    assert(phase.erasedTypes, phase)
+    assert(!orig.symbol.hasFlag(SYNCHRONIZED), orig.symbol.defString)
+    val origSym = orig.symbol
+    val origParams = orig.symbol.info.params
+    val newSym = maybeClone(orig.symbol)
+    newSym.setFlag(STATIC)
+    // Add an explicit self parameter
+    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF)
+    newSym.updateInfo(newSym.info match {
+      case mt @ MethodType(params, res) => copyMethodType(mt, selfParamSym :: params, res)
+    })
+    val selfParam = ValDef(selfParamSym)
+    val rhs = orig.rhs.substituteThis(newSym.owner, atPos(newSym.pos)(gen.mkAttributedIdent(selfParamSym)))
+      .substituteSymbols(origParams, newSym.info.params.drop(1)).changeOwner(origSym -> newSym)
+    treeCopy.DefDef(orig, orig.mods, orig.name, orig.tparams, (selfParam :: orig.vparamss.head) :: Nil, orig.tpt, rhs).setSymbol(newSym)
   }
 
   // TODO: the rewrite to AbstractFunction is superfluous once we compile FunctionN to a SAM type (aka functional interface)
--- a/src/compiler/scala/tools/nsc/transform/UnCurry.scala
+++ b/src/compiler/scala/tools/nsc/transform/UnCurry.scala
@@ -213,6 +213,7 @@ abstract class UnCurry extends InfoTransform
         // Expand the function body into an anonymous class
         gen.expandFunction(localTyper)(fun, inConstructorFlag)
       } else {
+        val mustExpand = mustExpandFunction(fun)
         // method definition with the same arguments, return type, and body as the original lambda
         val liftedMethod = gen.mkLiftedFunctionBodyMethod(localTyper)(fun.symbol.owner, fun)
 
@@ -221,11 +222,18 @@ abstract class UnCurry extends InfoTransform
           gen.mkForwarder(gen.mkAttributedRef(liftedMethod.symbol), (fun.vparams map (_.symbol)) :: Nil)
         ))
 
+        if (!mustExpand) {
+          liftedMethod.symbol.updateAttachment(DelambdafyTarget)
+          liftedMethod.updateAttachment(DelambdafyTarget)
+        }
+
         val typedNewFun = localTyper.typedPos(fun.pos)(Block(liftedMethod, super.transform(newFun)))
-        if (mustExpandFunction(fun)) {
+        if (mustExpand) {
           val Block(stats, expr : Function) = typedNewFun
           treeCopy.Block(typedNewFun, stats, gen.expandFunction(localTyper)(expr, inConstructorFlag))
-        } else typedNewFun
+        } else {
+          typedNewFun
+        }
       }
 
     def transformArgs(pos: Position, fun: Symbol, args: List[Tree], formals: List[Type]) = {
@@ -341,13 +349,18 @@ abstract class UnCurry extends InfoTransform
 
     private def isSelfSynchronized(ddef: DefDef) = ddef.rhs match {
       case Apply(fn @ TypeApply(Select(sel, _), _), _) =>
-        fn.symbol == Object_synchronized && sel.symbol == ddef.symbol.enclClass && !ddef.symbol.enclClass.isTrait
+        fn.symbol == Object_synchronized && sel.symbol == ddef.symbol.enclClass && !ddef.symbol.enclClass.isTrait &&
+          !ddef.symbol.isDelambdafyTarget /* these become static later, unsuitable for ACC_SYNCHRONIZED */
       case _ => false
     }
 
     /** If an eligible method is entirely wrapped in a call to synchronized
      *  locked on the same instance, remove the synchronized scaffolding and
      *  mark the method symbol SYNCHRONIZED for bytecode generation.
+     *
+     *  Delambdafy targets are deemed ineligible as the Delambdafy phase will
+     *  replace `this.synchronized` with `$this.synchronzed` now that it emits
+     *  all lambda impl methods as static.
      */
     private def translateSynchronized(tree: Tree) = tree match {
       case dd @ DefDef(_, _, _, _, _, Apply(fn, body :: Nil)) if isSelfSynchronized(dd) =>
--- a/src/compiler/scala/tools/nsc/typechecker/Duplicators.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Duplicators.scala
@@ -229,7 +229,12 @@ abstract class Duplicators extends Analyzer {
 
         case ddef @ DefDef(_, _, _, _, tpt, rhs) =>
           ddef.tpt modifyType fixType
-          super.typed(ddef.clearType(), mode, pt)
+          val result = super.typed(ddef.clearType(), mode, pt)
+          // TODO this is a hack, we really need a cleaner way to transport symbol attachments to duplicated methods
+          // bodies in specialized subclasses.
+          if (ddef.hasAttachment[DelambdafyTarget.type])
+            result.symbol.updateAttachment(DelambdafyTarget)
+          result
 
         case fun: Function =>
           debuglog("Clearing the type and retyping Function: " + fun)
--- a/test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala
+++ b/test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala
@@ -945,11 +945,11 @@ class InlinerTest extends BytecodeTesting {
     val t1 = getMethod(c, "t1")
     assertNoIndy(t1)
     // the indy call is inlined into t, and the closure elimination rewrites the closure invocation to the body method
-    assertInvoke(t1, "C", "C$$$anonfun$2")
+    assertInvoke(t1, "C", "$anonfun$m$2")
 
     val t2 = getMethod(c, "t2")
     assertNoIndy(t2)
-    assertInvoke(t2, "M$", "M$$$anonfun$1")
+    assertInvoke(t2, "M$", "$anonfun$m$1")
   }
 
   @Test
@@ -1033,7 +1033,7 @@ class InlinerTest extends BytecodeTesting {
       """.stripMargin
 
     val List(c) = compile(code)
-    assertInvoke(getMethod(c, "t1"), "C", "C$$$anonfun$1")
+    assertInvoke(getMethod(c, "t1"), "C", "$anonfun$t1$1")
     assertInvoke(getMethod(c, "t2"), "C", "a")
     assertInvoke(getMethod(c, "t3"), "C", "b")
     assertNoInvoke(getMethod(c, "t4"))
@@ -1097,8 +1097,8 @@ class InlinerTest extends BytecodeTesting {
       """.stripMargin
 
     val List(c) = compile(code)
-    assertInvoke(getMethod(c, "t1"), "C", "C$$$anonfun$1")
-    assertInvoke(getMethod(c, "t2"), "C", "C$$$anonfun$2")
+    assertInvoke(getMethod(c, "t1"), "C", "$anonfun$t1$1")
+    assertInvoke(getMethod(c, "t2"), "C", "$anonfun$t2$1")
     assertInvoke(getMethod(c, "t3"), "scala/Function1", "apply$mcII$sp")
     assertInvoke(getMethod(c, "t4"), "scala/Function1", "apply$mcII$sp")
     assertInvoke(getMethod(c, "t5"), "C", "h")
@@ -1273,39 +1273,39 @@ class InlinerTest extends BytecodeTesting {
       """.stripMargin
     val List(c, _, _) = compile(code)
 
-    assertSameSummary(getMethod(c, "t1"), List(BIPUSH, "C$$$anonfun$1", IRETURN))
-    assertSameSummary(getMethod(c, "t1a"), List(LCONST_1, "C$$$anonfun$2", IRETURN))
-    assertSameSummary(getMethod(c, "t2"), List(ICONST_1, ICONST_2, "C$$$anonfun$3",IRETURN))
+    assertSameSummary(getMethod(c, "t1"), List(BIPUSH, "$anonfun$t1$1", IRETURN))
+    assertSameSummary(getMethod(c, "t1a"), List(LCONST_1, "$anonfun$t1a$1", IRETURN))
+    assertSameSummary(getMethod(c, "t2"), List(ICONST_1, ICONST_2, "$anonfun$t2$1",IRETURN))
 
     // val a = new ValKl(n); new ValKl(anonfun(a.x)).x
     // value class instantiation-extraction should be optimized by boxing elim
     assertSameSummary(getMethod(c, "t3"), List(
       NEW, DUP, ICONST_1, "<init>", ASTORE,
       NEW, DUP, ALOAD, "x",
-      "C$$$anonfun$4",
+      "$anonfun$t3$1",
       "<init>",
       "x", IRETURN))
 
-    assertSameSummary(getMethod(c, "t4"), List(BIPUSH, "C$$$anonfun$5", "boxToInteger", ARETURN))
-    assertSameSummary(getMethod(c, "t4a"), List(ICONST_1, LDC, "C$$$anonfun$6", LRETURN))
-    assertSameSummary(getMethod(c, "t5"), List(BIPUSH, ICONST_3, "C$$$anonfun$7", "boxToInteger", ARETURN))
-    assertSameSummary(getMethod(c, "t5a"), List(BIPUSH, BIPUSH, I2B, "C$$$anonfun$8", IRETURN))
-    assertSameSummary(getMethod(c, "t6"), List(BIPUSH, "C$$$anonfun$9", RETURN))
-    assertSameSummary(getMethod(c, "t7"), List(ICONST_1, "C$$$anonfun$10", RETURN))
-    assertSameSummary(getMethod(c, "t8"), List(ICONST_1, LDC, "C$$$anonfun$11", LRETURN))
-    assertSameSummary(getMethod(c, "t9"), List(ICONST_1, "boxToInteger", "C$$$anonfun$12", RETURN))
+    assertSameSummary(getMethod(c, "t4"), List(BIPUSH, "$anonfun$t4$1", "boxToInteger", ARETURN))
+    assertSameSummary(getMethod(c, "t4a"), List(ICONST_1, LDC, "$anonfun$t4a$1", LRETURN))
+    assertSameSummary(getMethod(c, "t5"), List(BIPUSH, ICONST_3, "$anonfun$t5$1", "boxToInteger", ARETURN))
+    assertSameSummary(getMethod(c, "t5a"), List(BIPUSH, BIPUSH, I2B, "$anonfun$t5a$1", IRETURN))
+    assertSameSummary(getMethod(c, "t6"), List(BIPUSH, "$anonfun$t6$1", RETURN))
+    assertSameSummary(getMethod(c, "t7"), List(ICONST_1, "$anonfun$t7$1", RETURN))
+    assertSameSummary(getMethod(c, "t8"), List(ICONST_1, LDC, "$anonfun$t8$1", LRETURN))
+    assertSameSummary(getMethod(c, "t9"), List(ICONST_1, "boxToInteger", "$anonfun$t9$1", RETURN))
 
     // t9a inlines Range.foreach, which is quite a bit of code, so just testing the core
-    assertInvoke(getMethod(c, "t9a"), "C", "C$$$anonfun$13")
+    assertInvoke(getMethod(c, "t9a"), "C", "$anonfun$t9a$1")
     assertInvoke(getMethod(c, "t9a"), "scala/runtime/BoxesRunTime", "boxToInteger")
 
     assertSameSummary(getMethod(c, "t10"), List(
       ICONST_1, ISTORE,
       ALOAD, ILOAD,
-      "C$$$anonfun$14", RETURN))
+      "$anonfun$t10$1", RETURN))
 
     // t10a inlines Range.foreach
-    assertInvoke(getMethod(c, "t10a"), "C", "C$$$anonfun$15")
+    assertInvoke(getMethod(c, "t10a"), "C", "$anonfun$t10a$1")
     assertDoesNotInvoke(getMethod(c, "t10a"), "boxToInteger")
   }
 
@@ -1330,8 +1330,8 @@ class InlinerTest extends BytecodeTesting {
       """.stripMargin
     val List(c) = compile(code)
     assertSameCode(getMethod(c, "t1"), List(Op(ICONST_0), Op(ICONST_1), Op(IADD), Op(IRETURN)))
-    assertEquals(getInstructions(c, "t2") collect { case i: Invoke => i.owner +"."+ i.name }, List(
-      "scala/runtime/IntRef.create", "C.C$$$anonfun$1"))
+    assertEquals(getMethod(c, "t2").instructions collect { case i: Invoke => i.owner +"."+ i.name }, List(
+      "scala/runtime/IntRef.create", "C.$anonfun$t2$1"))
   }
 
   @Test
@@ -1449,9 +1449,9 @@ class InlinerTest extends BytecodeTesting {
 
     // box-unbox will clean it up
     assertSameSummary(getMethod(c, "t"), List(
-      ALOAD, "C$$$anonfun$1", IFEQ /*A*/,
-      "C$$$anonfun$2", IRETURN,
-      -1 /*A*/, "C$$$anonfun$3", IRETURN))
+      ALOAD, "$anonfun$t$1", IFEQ /*A*/,
+      "$anonfun$t$2", IRETURN,
+      -1 /*A*/, "$anonfun$t$3", IRETURN))
   }
 
   @Test
@@ -1501,7 +1501,7 @@ class InlinerTest extends BytecodeTesting {
     val List(c) = compile(code)
     val t = getMethod(c, "t")
     assertNoIndy(t)
-    assertInvoke(t, "C", "C$$$anonfun$1")
+    assertInvoke(t, "C", "$anonfun$t$1")
   }
 
   @Test
--- a/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
@@ -121,7 +121,7 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
   def getBoxedUnit: FieldInsnNode = new FieldInsnNode(GETSTATIC, srBoxedUnitRef.internalName, "UNIT", srBoxedUnitRef.descriptor)
 
-  private val anonfunAdaptedName = """.*\$anonfun\$\d+\$adapted""".r
+  private val anonfunAdaptedName = """.*\$anonfun\$.*\$\d+\$adapted""".r
   def hasAdaptedImplMethod(closureInit: ClosureInstantiation): Boolean = {
     isBuiltinFunctionType(Type.getReturnType(closureInit.lambdaMetaFactoryCall.indy.desc).getInternalName) &&
     anonfunAdaptedName.pattern.matcher(closureInit.lambdaMetaFactoryCall.implMethod.getName).matches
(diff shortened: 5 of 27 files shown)


src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +321,40 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
+    *   - If `maybeClone` is the identity function, this runs "in place"
+    *     and mutates the symbol of `orig`. `orig` should be discarded
+    *   - Symbol owners and returns are substituted
+    *   - Recursive calls are not rewritten. This is correct if we assume
+    *     that we either:
+    *       - are in "in-place" mode, but can guarantee that no recursive calls exists
+    *       - are associating the RHS with a cloned symbol, but intend for the original
+    *         method to remain and for recursive calls to target it.
+    */
+  final def mkStatic(orig: DefDef, maybeClone: Symbol => Symbol): DefDef = {
```
retronym, line 339: The flexibility of `maybeClone` is to support a forthcoming PR that will reuse this method to emit static methods in traits.

retronym:
A closer look at the failure in the optimizer test: https://gist.github.com/a50a921b70944811bdd55268f1942c6b

retronym:
Needed to update the regex in `BackendUtils.anonfunAdaptedName`, in addition to the method names in the expected bytecode summaries. Still getting an extra null check.

test/junit/scala/issues/OptimizedBytecodeTest.scala
```diff
@@ -56,9 +55,9 @@ class OptimizedBytecodeTest extends ClearAfterClass {
     val List(c) = compileClasses(compiler)(code)
 
     assertSameSummary(getSingleMethod(c, "t"), List(
-      LDC, ASTORE, ALOAD /*0*/, ALOAD /*1*/, "C$$$anonfun$1", IRETURN))
-    assertSameSummary(getSingleMethod(c, "C$$$anonfun$1"), List(LDC, "C$$$anonfun$2", IRETURN))
-    assertSameSummary(getSingleMethod(c, "C$$$anonfun$2"), List(-1 /*A*/, GOTO /*A*/))
+      LDC, ASTORE, ALOAD /*0*/, ALOAD /*1*/, "$anonfun$t$1", IRETURN))
+    assertSameSummary(getSingleMethod(c, "$anonfun$t$1"), List(ALOAD, IFNONNULL /*8*/, ACONST_NULL, ATHROW, -1 /*8*/, LDC, "$anonfun$t$2", IRETURN))
```
retronym: We could avoid the null check by teaching the backend that the `$this` parameter is non-nullable.

retronym:
I've had to disable this for lambdas that don't use `invokedynamic`, but instead still need to use an anonymous class. The code that implements needs some more polish / documentation, which I'll add before we merge.

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -269,13 +269,25 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     if (!sam.exists) mkMethodForFunctionBody(localTyper)(owner, fun, nme.apply)()
     else {
       val samMethType = fun.tpe memberInfo sam
-      mkMethodForFunctionBody(localTyper)(owner, fun, sam.name.toTermName)(methParamProtos = samMethType.params, resTp = samMethType.resultType)
+      val res = mkMethodForFunctionBody(localTyper)(owner, fun, sam.name.toTermName)(methParamProtos = samMethType.params, resTp = samMethType.resultType)
+      res
```
lrytz: - [x] debugging artifact?

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +322,41 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     newDefDef(methSym, moveToMethod(useMethodParams(fun.body)))(tpt = TypeTree(resTp))
   }
 
+  /**
+    * Create a new `DefDef` based on `orig` with an explicit self parameter.
+    *
+    * Details:
+    *   - Must by runs after erasure
```
lrytz: `by runs` -> `be run`?

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +322,41 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
+    val newSym = maybeClone(orig.symbol)
+    newSym.setFlag(STATIC)
+    // Add an explicit self parameter
+    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF)
+    newSym.updateInfo(newSym.info match {
+      case mt @ MethodType(params, res) => copyMethodType(mt, selfParamSym :: params, res)
+    })
+    val selfParam = ValDef(selfParamSym)
+    val oldSyms = origParams
+    val newSyms = newSym.info.params.drop(1)
```
lrytz: maybe call it `newParams`

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +322,41 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
+    val origParams = orig.symbol.info.params
+    val newSym = maybeClone(orig.symbol)
+    newSym.setFlag(STATIC)
+    // Add an explicit self parameter
+    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF)
+    newSym.updateInfo(newSym.info match {
+      case mt @ MethodType(params, res) => copyMethodType(mt, selfParamSym :: params, res)
+    })
+    val selfParam = ValDef(selfParamSym)
+    val oldSyms = origParams
```
lrytz: `oldSyms` is just an alias for `origParams`, not needed

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +322,41 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
   }
 
+  /**
+    * Create a new `DefDef` based on `orig` with an explicit self parameter.
+    *
+    * Details:
+    *   - Must by runs after erasure
+    *   - If `maybeClone` is the identity function, this runs "in place"
+    *     and mutates the symbol of `orig`. `orig` should be discarded
+    *   - Symbol owners and returns are substituted
```
lrytz: `returns` -> `parameters`

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +322,41 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
+    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF)
+    newSym.updateInfo(newSym.info match {
+      case mt @ MethodType(params, res) => copyMethodType(mt, selfParamSym :: params, res)
+    })
+    val selfParam = ValDef(selfParamSym)
+    val oldSyms = origParams
+    val newSyms = newSym.info.params.drop(1)
+
+    val rhs = orig.rhs.substituteThis(newSym.owner, atPos(newSym.pos)(gen.mkAttributedIdent(selfParamSym)))
+      .substituteSymbols(oldSyms, newSyms).changeOwner(origSym -> newSym)
```
lrytz: here we are mutating the `rhs` of the original `DefDef` (`substituteSymbols`, `changeOwner`). IUC the the original `DefDef` should remain usable.

src/compiler/scala/tools/nsc/transform/UnCurry.scala
```diff
@@ -348,9 +352,13 @@ abstract class UnCurry extends InfoTransform
...
      *  locked on the same instance, remove the synchronized scaffolding and
      *  mark the method symbol SYNCHRONIZED for bytecode generation.
+     *
+     *  Delambdafy targets are deemed ineligible as the Delambdafy phase will
+     *  replace `this.synchronized` with `$this.synchronzed` now that it emits
+     *  all lambda impl methods as static.
      */
     private def translateSynchronized(tree: Tree) = tree match {
-      case dd @ DefDef(_, _, _, _, _, Apply(fn, body :: Nil)) if isSelfSynchronized(dd) =>
+      case dd @ DefDef(_, _, _, _, _, Apply(fn, body :: Nil)) if isSelfSynchronized(dd) && dd.symbol.isDelambdafyFunction =>
```
lrytz: missing a `!` ? and maybe you could add the condition into `isSelfSynchronized`, which already performs some additional checks.

test/junit/scala/issues/OptimizedBytecodeTest.scala
lrytz: here's a wip: https://github.com/lrytz/scala/commit/bb9d0b8f0815026f6c01f7cf7bd4379d906d22c4

test/junit/scala/tools/nsc/backend/jvm/analysis/ProdConsAnalyzerTest.scala
```diff
@@ -181,7 +181,7 @@ class ProdConsAnalyzerTest extends ClearAfterClass {
     val List(dup2)  = findInstr(m, "DUP2")
     val List(add)   = findInstr(m, "IADD")
     val List(swap)  = findInstr(m, "SWAP")
-    val List(store) = findInstr(m, "ISTORE")
+    val store = getInstr(m, "ISTORE")
```
lrytz: there's a few more just above that could be replaced

test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala
```diff
@@ -1501,16 +1499,48 @@ class InlinerTest extends ClearAfterClass {
...
+        |}
+      """.stripMargin
+    val c = compile(code).find(_.name == "C").get
+    val t = getSingleMethod(c, "t")
+    assertNoIndy(t)
+    assertInvoke(t, "C", "$anonfun$t$1")
+  }
+
+  @Test
+  @Ignore // TODO figure out exactly why this fails
```
lrytz: did it fail on CI or only locally? if you don't go through a full bootstrap (no locker) the failure is expected

lrytz: (STARR / 2.12.0-M4 inlines the wrong method due to https://github.com/scala/scala/pull/5140)

lrytz:
the test failure (`tailcalls` for synchronized method) might be due to https://github.com/scala/scala/pull/5157#discussion-diff-63159265

test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala
retronym: It failed locally in the SBT build.

test/junit/scala/tools/nsc/backend/jvm/analysis/ProdConsAnalyzerTest.scala
retronym: Ha, the extra spaces after the `=` fooled my regex :)

test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala
lrytz: so it should be fine on the ci

retronym: Right. But shouldn't we eventually restrict ourselves to writing JUnit tests that work under the constraints of the SBT build?

lrytz: yes, it will prevent confusions like in this case.. it will work once we have a new starr though :)

src/compiler/scala/tools/nsc/ast/TreeGen.scala
retronym: Actually `changeOwner` patches up `return`-s, so it should be `owners, returns, and param symbols`

retronym:
Rebased to resolve conflicts with @lrytz's recent refactorings to the JUnit test utilities.

src/compiler/scala/tools/nsc/ast/TreeGen.scala
lrytz: ping here

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -269,13 +269,25 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
     }
   }
 
   // used to create the lifted method that holds a function's body
-  def mkLiftedFunctionBodyMethod(localTyper: analyzer.Typer)(owner: Symbol, fun: Function) =
-    mkMethodForFunctionBody(localTyper)(owner, fun, nme.ANON_FUN_NAME)(additionalFlags = ARTIFACT)
+  def mkLiftedFunctionBodyMethod(localTyper: analyzer.Typer)(owner: Symbol, fun: Function, newNamePrefix: TermName) = {
+    def nonLocalEnclosingMember(sym: Symbol) = {
+      if (sym.isLocalDummy) sym.enclClass.primaryConstructor else {
+        if (sym.isLocalToBlock) sym.originalOwner else sym
```
lrytz: - [x] i still think we need a recursive call to `nonLocalEnclosingMember` here

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -310,6 +322,38 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
+    val origParams = orig.symbol.info.params
+    val newSym = maybeClone(orig.symbol)
+    newSym.setFlag(STATIC)
+    // Add an explicit self parameter
+    val selfParamSym = newSym.newSyntheticValueParam(newSym.owner.typeConstructor, nme.SELF)
+    newSym.updateInfo(newSym.info match {
+      case mt @ MethodType(params, res) => copyMethodType(mt, selfParamSym :: params, res)
+    })
+    val selfParam = ValDef(selfParamSym)
+    val rhs = orig.rhs.substituteThis(newSym.owner, atPos(newSym.pos)(gen.mkAttributedIdent(selfParamSym)))
```
lrytz, line 352: all `this` references will be replaced by the same tree (same object reference), as the `tree` parameter of `substituteThis` is not by-name: https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/Trees.scala#L184-L185

I guess this is an oversight, because the `ThisSubstituter` class does take a by-name param (https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/Trees.scala#L1541).

src/partest-extras/scala/tools/partest/ASMConverters.scala
```diff
@@ -2,7 +2,7 @@ package scala.tools.partest
 
 import scala.collection.JavaConverters._
 import scala.tools.asm
-import asm.{tree => t}
+import asm.{Opcodes, tree => t}
```
lrytz: no-op

src/compiler/scala/tools/nsc/transform/UnCurry.scala
```diff
@@ -213,19 +213,23 @@ abstract class UnCurry extends InfoTransform
         // Expand the function body into an anonymous class
         gen.expandFunction(localTyper)(fun, inConstructorFlag)
       } else {
+        val mustExpand = mustExpandFunction(fun)
+        val newNamePrefix = if (mustExpand) nme.DELAMBDAFY_LAMBDA_CLASS_NAME else nme.ANON_FUN_NAME
```
lrytz: so if we expand to an anon class the method is called `$lambda`, if we use LFM it's named `$anonfun` - why that distinction?

lrytz:
did you have a chance to look at https://github.com/lrytz/scala/commit/bb9d0b8f0815026f6c01f7cf7bd4379d906d22c4? fixes nullness analysis for the `$this` parameter.

src/compiler/scala/tools/nsc/transform/UnCurry.scala
retronym: I need to document this. I needed to exclude non LMF functions/SAMs from the static-rewrite in `Delambdafy`, which runs for methods that satisfy `isDelambdafyTarget`.

Ideally we'd be able to rewrite them as well, but it gets more involved as we need to rewrite the anonymous class that delegates to this method in a non-trivial way.
- [x] Maybe a tree or symbol attachment would be a cleaner way to implement `isDelambdafyTarget`.

retronym:
> did you have a chance to look at lrytz@bb9d0b8? fixes nullness analysis for the $this parameter.
- [x] Yep, that looked-GTM. I'll include it here.

src/compiler/scala/tools/nsc/ast/TreeGen.scala
retronym, line 352: The lack of by-name seems to date back to https://github.com/scala/scala/commit/a03c34c140709e95bb6ab6a8984747b83d630eea#diff-092465703161e9bcdd19dd4d0ca51221R131

Agree it appears to be an oversight.

The tree sharing is probably benign after erasure (we won't clear the type then retype it in one occurence that leave the wrong result for another context). 
- [x] But I'll modify the API in any case.

retronym:
@lrytz Ready for re-review

lrytz:
LGTM!

retronym:
@lrytz My change to use symbol attachments to identify methods that are delambdafy targets interacted poorly with specialization. I've pushed a slightly modified version with a hack in `Duplicators` to make it work.

src/compiler/scala/tools/nsc/typechecker/Duplicators.scala
```diff
@@ -229,7 +229,12 @@ abstract class Duplicators extends Analyzer {
 
         case ddef @ DefDef(_, _, _, _, tpt, rhs) =>
           ddef.tpt modifyType fixType
-          super.typed(ddef.clearType(), mode, pt)
+          val result = super.typed(ddef.clearType(), mode, pt)
+          // TODO this is a hack, we really need a cleaner way to transport symbol attachments to duplicated methods
+          // bodies in specialized subclasses.
+          if (ddef.hasAttachment[DelambdafyTarget.type])
+            result.symbol.updateAttachment(DelambdafyTarget)
+          result
```
lrytz, line 237: i see the problem.. i don't really understand how duplicators work. first it seems to set `tree.symbol = NoSymbol` and then call `super.typed` - where is the new symbol created and assigned?

adriaanm, line 237: I believe that's done in calls to the `retyped` method.

lrytz, line 237: i don't really see how that works yet..

adriaanm, line 237: It's (probably) a little too clever: the invalidSyms map tracks the old symbol to the new one by mapping to the tree, whose symbol field will be mutated later to the new symbol...

adriaanm, line 237:
```
  private val invalidSyms: mutable.Map[Symbol, Tree] = perRunCaches.newMap[Symbol, Tree]()
...
   /** Return the new symbol corresponding to `sym`. */
    private def updateSym(sym: Symbol): Symbol =
      if (invalidSyms.isDefinedAt(sym))
        invalidSyms(sym).symbol
      else
        sym
....
          case vdef @ ValDef(mods, name, _, rhs) if mods.hasFlag(Flags.LAZY) =>
            debuglog("ValDef " + name + " sym.info: " + vdef.symbol.info)
            invalidSyms(vdef.symbol) = vdef
            val newowner = owner orElse context.owner
            val newsym = vdef.symbol.cloneSymbol(newowner)
            newsym.setInfo(fixType(vdef.symbol.info))
            vdef.symbol = newsym
...
```

lrytz, line 237: ok, so we go through `cloneSymbol` which seems to also clone attachments (https://github.com/scala/scala/blob/25165b7ad9c4aafdddb4e871b195a91014a91b72/src/reflect/scala/reflect/internal/Symbols.scala#L1951) - shouldn't cloned sym already have the `DelambdafyTarget` attachment?

adriaanm, line 237: I pasted the `ValDef` case because the `invalidSyms` connection is clearer there, sorry for the confusion. For `DefDef`s the symbol is thrown away and the type checker will create a new one, which explains why the attachment goes walkabout.

retronym, line 237: Yep, tree attachments make it through, but I need a symbol attachment to support the one place (in `LambdaLift`) where we need to know if the method we're invoking is a delambdafy target. :sigh:

retronym, line 237: ... although, I could get rid of this with:

``` diff
diff --git a/src/compiler/scala/tools/nsc/transform/Delambdafy.scala b/src/compiler/scala/tools/nsc/transform/Delambdafy.scala
index 65449b8..c3ba4f4 100644
--- a/src/compiler/scala/tools/nsc/transform/Delambdafy.scala
+++ b/src/compiler/scala/tools/nsc/transform/Delambdafy.scala
@@ -258,7 +258,7 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
         super.transform(transformFunction(fun))
       case Template(_, _, _) =>
         def pretransform(tree: Tree): Tree = tree match {
-          case dd: DefDef if dd.symbol.isDelambdafyTarget =>
+          case dd: DefDef if dd.hasAttachment[DelambdafyTarget.type] =>
             if (!dd.symbol.hasFlag(STATIC) && methodReferencesThis(dd.symbol)) {
               gen.mkStatic(dd, dd.symbol.name, sym => sym)
             } else {
@@ -338,7 +338,7 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
     private var currentMethod: Symbol = NoSymbol

     override def traverse(tree: Tree) = tree match {
-      case DefDef(_, _, _, _, _, _) if tree.symbol.isDelambdafyTarget =>
+      case DefDef(_, _, _, _, _, _) if tree.hasAttachment[DelambdafyTarget.type] =>
         // we don't expect defs within defs. At this phase trees should be very flat
         if (currentMethod.exists) devWarning("Found a def within a def at a phase where defs are expected to be flattened out.")
         currentMethod = tree.symbol
diff --git a/src/compiler/scala/tools/nsc/transform/LambdaLift.scala b/src/compiler/scala/tools/nsc/transform/LambdaLift.scala
index 074acc1..0f40d29 100644
--- a/src/compiler/scala/tools/nsc/transform/LambdaLift.scala
+++ b/src/compiler/scala/tools/nsc/transform/LambdaLift.scala
@@ -523,10 +523,7 @@ abstract class LambdaLift extends InfoTransform {
   } // class LambdaLifter

   private def addFree[A](sym: Symbol, free: List[A], original: List[A]): List[A] = {
-    val prependFree = (
-         !sym.isConstructor // this condition is redundant for now. It will be needed if we remove the second condition in 2.12.x
-      && (settings.Ydelambdafy.value == "method" && sym.isDelambdafyTarget) // SI-8359 Makes the lambda body a viable as the target MethodHandle for a call to LambdaMetafactory
-    )
+    val prependFree = !sym.isConstructor
     if (prependFree) free ::: original
     else             original ::: free
   }
diff --git a/src/compiler/scala/tools/nsc/transform/UnCurry.scala b/src/compiler/scala/tools/nsc/transform/UnCurry.scala
index 374e843..11380b7 100644
--- a/src/compiler/scala/tools/nsc/transform/UnCurry.scala
+++ b/src/compiler/scala/tools/nsc/transform/UnCurry.scala
@@ -222,10 +222,8 @@ abstract class UnCurry extends InfoTransform
           gen.mkForwarder(gen.mkAttributedRef(liftedMethod.symbol), (fun.vparams map (_.symbol)) :: Nil)
         ))

-        if (!mustExpand) {
-          liftedMethod.symbol.updateAttachment(DelambdafyTarget)
+        if (!mustExpand)
           liftedMethod.updateAttachment(DelambdafyTarget)
-        }

         val typedNewFun = localTyper.typedPos(fun.pos)(Block(liftedMethod, super.transform(newFun)))
         if (mustExpand) {
@@ -350,7 +348,7 @@ abstract class UnCurry extends InfoTransform
     private def isSelfSynchronized(ddef: DefDef) = ddef.rhs match {
       case Apply(fn @ TypeApply(Select(sel, _), _), _) =>
         fn.symbol == Object_synchronized && sel.symbol == ddef.symbol.enclClass && !ddef.symbol.enclClass.isTrait &&
-          !ddef.symbol.isDelambdafyTarget /* these become static later, unsuitable for ACC_SYNCHRONIZED */
+          !ddef.hasAttachment[DelambdafyTarget.type] /* these become static later, unsuitable for ACC_SYNCHRONIZED */
       case _ => false
     }

diff --git a/src/reflect/scala/reflect/internal/Symbols.scala b/src/reflect/scala/reflect/internal/Symbols.scala
index 62a33ad..edffe73 100644
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -807,7 +807,6 @@ trait Symbols extends api.Symbols { self: SymbolTable =>

     final def isAnonymousFunction = isSynthetic && (name containsName tpnme.ANON_FUN_NAME)
     final def isDelambdafyFunction = isSynthetic && (name containsName tpnme.DELAMBDAFY_LAMBDA_CLASS_NAME)
-    final def isDelambdafyTarget  = isArtifact && isMethod && hasAttachment[DelambdafyTarget.type]
     final def isDefinedInPackage  = effectiveOwner.isPackageClass
     final def needsFlatClasses    = phase.flatClasses && rawowner != NoSymbol && !rawowner.isPackageClass


```

lrytz, line 237: > For `DefDefs` the symbol is thrown away and the type checker will create a new one

OK, sorry I'm not giving up yet :) where is that in the typer?

retronym, line 237: If we compile:

```
class C[@specialized A] {
  def foo(a: A) = { def bar = 42; a }
}

```

We get to:

https://github.com/scala/scala/blob/92affd3e76d1a3019f1c35ac20af32360d914641/src/compiler/scala/tools/nsc/typechecker/Duplicators.scala#L167-L167

For:

```
tree: def bar(): Int = 42
```

From a call stack:

```
"main@1" prio=5 tid=0x1 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator.invalidate(Duplicators.scala:167)
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator.scala$tools$nsc$typechecker$Duplicators$BodyDuplicator$$$anonfun$18(Duplicators.scala:181)
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator.scala$tools$nsc$typechecker$Duplicators$BodyDuplicator$$$anonfun$18$adapted(Duplicators.scala:181)
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator$$Lambda$991.478209179.apply(Unknown Source:-1)
      at scala.collection.immutable.List.foreach(List.scala:376)
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator.invalidateAll(Duplicators.scala:181)
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator.typed(Duplicators.scala:220)
      at scala.tools.nsc.typechecker.Typers$Typer.transformedOrTyped(Typers.scala:5602)
      at scala.tools.nsc.typechecker.Typers$Typer.typedDefDef(Typers.scala:2243)
      at scala.tools.nsc.typechecker.Typers$Typer.typedMemberDef$1(Typers.scala:5305)
      at scala.tools.nsc.typechecker.Typers$Typer.typed1(Typers.scala:5356)
      at scala.tools.nsc.typechecker.Typers$Typer.runTyper$1(Typers.scala:5393)
      at scala.tools.nsc.typechecker.Typers$Typer.typedInternal(Typers.scala:5423)
      at scala.tools.nsc.typechecker.Typers$Typer.body$2(Typers.scala:5367)
      at scala.tools.nsc.typechecker.Typers$Typer.typed(Typers.scala:5371)
      at scala.tools.nsc.typechecker.Duplicators$BodyDuplicator.typed(Duplicators.scala:232)
      at scala.tools.nsc.typechecker.Typers$Typer.typed(Typers.scala:5448)
      at scala.tools.nsc.typechecker.Duplicators.retyped(Duplicators.scala:36)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationDuplicator.super$retyped(SpecializeTypes.scala:1331)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationDuplicator.scala$tools$nsc$transform$SpecializeTypes$SpecializationDuplicator$$$anonfun$154(SpecializeTypes.scala:1331)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationDuplicator$$Lambda$963.1043639521.apply(Unknown Source:-1)
      at scala.reflect.internal.SymbolTable.enteringPhase(SymbolTable.scala:235)
      at scala.tools.nsc.Global.enteringSpecialize(Global.scala:967)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationDuplicator.retyped(SpecializeTypes.scala:1331)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.duplicateBody(SpecializeTypes.scala:1791)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$196(SpecializeTypes.scala:1641)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$950.1527007086.apply(Unknown Source:-1)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.reportError(SpecializeTypes.scala:1445)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transformDefDef$1(SpecializeTypes.scala:1641)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform1(SpecializeTypes.scala:1713)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$164(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$816.778720569.apply(Unknown Source:-1)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.reportError(SpecializeTypes.scala:1445)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1396)
      at scala.reflect.api.Trees$Transformer.scala$reflect$api$Trees$Transformer$$$anonfun$6(Trees.scala:2559)
      at scala.reflect.api.Trees$Transformer$$Lambda$703.2052435819.apply(Unknown Source:-1)
      at scala.collection.immutable.List.loop$1(List.scala:171)
      at scala.collection.immutable.List.mapConserve(List.scala:187)
      at scala.reflect.api.Trees$Transformer.transformTrees(Trees.scala:2559)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$190(SpecializeTypes.scala:1624)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$911.1692073709.apply(Unknown Source:-1)
      at scala.reflect.api.Trees$Transformer.atOwner(Trees.scala:2600)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:30)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:25)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transformTemplate$1(SpecializeTypes.scala:1624)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform1(SpecializeTypes.scala:1626)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$164(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$816.778720569.apply(Unknown Source:-1)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.reportError(SpecializeTypes.scala:1445)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1396)
      at scala.reflect.api.Trees$Transformer.transformTemplate(Trees.scala:2563)
      at scala.reflect.internal.Trees.scala$reflect$internal$Trees$$$anonfun$37(Trees.scala:1412)
      at scala.reflect.internal.Trees$$Lambda$699.2089589253.apply(Unknown Source:-1)
      at scala.reflect.api.Trees$Transformer.atOwner(Trees.scala:2600)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:30)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:25)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:17)
      at scala.reflect.internal.Trees.itransform(Trees.scala:1411)
      at scala.reflect.internal.SymbolTable.itransform(SymbolTable.scala:16)
      at scala.reflect.internal.SymbolTable.itransform(SymbolTable.scala:16)
      at scala.reflect.api.Trees$Transformer.transform(Trees.scala:2555)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.transform(TypingTransformers.scala:44)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform1(SpecializeTypes.scala:1737)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$164(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$816.778720569.apply(Unknown Source:-1)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.reportError(SpecializeTypes.scala:1445)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1396)
      at scala.reflect.api.Trees$Transformer.scala$reflect$api$Trees$Transformer$$$anonfun$13(Trees.scala:2589)
      at scala.reflect.api.Trees$Transformer$$Lambda$698.1611370719.apply(Unknown Source:-1)
      at scala.collection.immutable.List.loop$1(List.scala:171)
      at scala.collection.immutable.List.mapConserve(List.scala:187)
      at scala.reflect.api.Trees$Transformer.transformStats(Trees.scala:2587)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$186(SpecializeTypes.scala:1610)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$818.41765385.apply(Unknown Source:-1)
      at scala.reflect.api.Trees$Transformer.atOwner(Trees.scala:2600)
      at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:30)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform1(SpecializeTypes.scala:1608)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.scala$tools$nsc$transform$SpecializeTypes$$anon$2$$$anonfun$164(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2$$Lambda$816.778720569.apply(Unknown Source:-1)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.reportError(SpecializeTypes.scala:1445)
      at scala.tools.nsc.transform.SpecializeTypes$$anon$2.transform(SpecializeTypes.scala:1453)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationTransformer.scala$tools$nsc$transform$SpecializeTypes$SpecializationTransformer$$$anonfun$269(SpecializeTypes.scala:1983)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationTransformer$$Lambda$815.1415630650.apply(Unknown Source:-1)
      at scala.reflect.internal.SymbolTable.enteringPhase(SymbolTable.scala:235)
      at scala.reflect.internal.SymbolTable.exitingPhase(SymbolTable.scala:256)
      at scala.tools.nsc.Global.exitingSpecialize(Global.scala:957)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationTransformer.transform(SpecializeTypes.scala:1983)
      at scala.tools.nsc.transform.SpecializeTypes$SpecializationTransformer.transform(SpecializeTypes.scala:1979)
      at scala.tools.nsc.ast.Trees$Transformer.transformUnit(Trees.scala:140)
      at scala.tools.nsc.transform.Transform$Phase.apply(Transform.scala:30)
      at scala.tools.nsc.Global$GlobalPhase.scala$tools$nsc$Global$GlobalPhase$$$anonfun$11(Global.scala:402)
      at scala.tools.nsc.Global$GlobalPhase$$Lambda$387.888655833.apply$mcV$sp(Unknown Source:-1)
      at scala.tools.nsc.Global$GlobalPhase.withCurrentUnitNoLog(Global.scala:395)
      at scala.tools.nsc.Global$GlobalPhase.withCurrentUnit(Global.scala:386)
      at scala.tools.nsc.Global$GlobalPhase.applyPhase(Global.scala:402)
      at scala.tools.nsc.Global$GlobalPhase.scala$tools$nsc$Global$GlobalPhase$$$anonfun$10(Global.scala:366)
      at scala.tools.nsc.Global$GlobalPhase.scala$tools$nsc$Global$GlobalPhase$$$anonfun$10$adapted(Global.scala:366)
      at scala.tools.nsc.Global$GlobalPhase$$Lambda$386.980138431.apply(Unknown Source:-1)
      at scala.collection.Iterator.foreach(Iterator.scala:932)
      at scala.collection.AbstractIterator.foreach(Iterator.scala:1409)
      at scala.tools.nsc.Global$GlobalPhase.run(Global.scala:366)
      at scala.tools.nsc.Global$Run.compileUnitsInternal(Global.scala:1396)
      at scala.tools.nsc.Global$Run.compileUnits(Global.scala:1381)
      at scala.tools.nsc.Global$Run.compileSources(Global.scala:1376)
      at scala.tools.nsc.Global$Run.compile(Global.scala:1470)
      at scala.tools.nsc.Driver.doCompile(Driver.scala:35)
      at scala.tools.nsc.MainClass.doCompile(Main.scala:24)
      at scala.tools.nsc.Driver.process(Driver.scala:55)
      at scala.tools.nsc.Driver.main(Driver.scala:68)
      at scala.tools.nsc.Main.main(Main.scala:-1)

```

Note that the delambdafy target method that is still method-local at the specialize phase.

adriaanm, line 237: in `invalidate`:

```
          case DefDef(_, name, tparams, vparamss, _, rhs) =>
            // invalidate parameters
            invalidateAll(tparams ::: vparamss.flatten)
            tree.symbol = NoSymbol // <-- so long, symbol!
```

lrytz, line 237: i see where the symbol is removed from the tree, but still wonder where the new one is created. anyway, i'm waisting everyone's time - i think i'd prefer the tree-based approach you mention above (https://github.com/scala/scala/pull/5157#discussion_r65637086), but i don't want to hold things back unnecessarily.

retronym, line 237:
```
    def typedBlock(block0: Block, mode: Mode, pt: Type): Block = {
      val syntheticPrivates = new ListBuffer[Symbol]
      try {
        namer.enterSyms(block0.stats)
```

lrytz, line 237: thanks, got it now! i didn't connect the dots that those were local methods.

adriaanm, line 237: No problem! It's good to dig deep. I think we'd also create new symbols for non-local methods in `typedTemplate`:

```
enterSyms(context.outer.make(templ, clazz, clazz.info.decls), body1)
```

(it eventually gets to `enterDefDef` in Namers)

retronym:
I think this one and #5099 are ready to roll. @adriaanm WDYT?

adriaanm:
:shipit:!

PR #5158, opened 2016-05-12
Reduce boilerplate in compiler JUnit tests

retronym:
Many JUnit tests share a compiler instance between all
test cases in a class to reduce overhead.

This commit refactors the mechanism to reduce the boilerplate.

In the new scheme:
- Using the `@ClassRule` hook in JUnit, we create a per-class
  map for each test class.
- Per-class values are registered from the test class itself
  by calling `cached("someKey", () => mkExpensiveThing)`
- At the end of the test, the entries in this map are `close()`-ed
  (if they implement `Closable`), and are released for
  garbage collection.)

Review by @lrytz

### DIFF
--- a/test/junit/scala/tools/testing/ClearAfterClass.java
+++ b/test/junit/scala/tools/testing/ClearAfterClass.java
@@ -1,20 +1,53 @@
 package scala.tools.testing;
 
-import org.junit.AfterClass;
+import org.junit.ClassRule;
+import org.junit.rules.TestRule;
+import org.junit.runners.model.Statement;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * Extend this class to use JUnit's @AfterClass. This annotation only works on static methods,
+ * Extend this class to use JUnit's @ClassRule. This annotation only works on static methods,
  * which cannot be written in Scala.
  *
  * Example: {@link scala.tools.nsc.backend.jvm.opt.InlinerTest}
  */
 public class ClearAfterClass {
-    public static interface Clearable {
-        void clear();
+    private static Map<Class<?>, Map<String, Object>> cache = new ConcurrentHashMap<>();
+
+    @ClassRule
+    public static TestRule clearClassCache() {
+        return (statement, desc) -> new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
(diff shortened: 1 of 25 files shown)


test/junit/scala/tools/testing/ClearAfterClass.java
```diff
@@ -1,20 +1,53 @@
...
  *
  * Example: {@link scala.tools.nsc.backend.jvm.opt.InlinerTest}
  */
 public class ClearAfterClass {
-    public static interface Clearable {
-        void clear();
+    private static Map<Class<?>, Map<String, Object>> cache = new ConcurrentHashMap<>();
+
+    @ClassRule
+    public static TestRule clearClassCache() {
```
lrytz, line 22: How about using [`ExternalResource`](https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/rules/ExternalResource.java#L37) instead of the bare `TestRule`? It would be a bit more easy to understand..

test/junit/scala/tools/testing/ClearAfterClass.java
```diff
@@ -1,20 +1,53 @@
...
+        if (o instanceof Closeable) {
+            try {
+                ((Closeable) o).close();
+            } catch (IOException e) {
+                // ignore
+            }
+        }
+    }
+
+    public <T> T cached(String key, scala.Function0<T> t) {
```
lrytz, line 48: we could add a Scala class between `ClearAfterClass` and the test calsses, then `t` could be a by-name param – if you think that's an improvement.

test/junit/scala/tools/testing/ClearAfterClass.java
retronym, line 22: We need access to the `Class` of the test case. `ExternalResource` doesn't expose that to the `before`/`after` methods.

test/junit/scala/tools/testing/ClearAfterClass.java
retronym, line 48: I considered this but didn't think it was worth it. Also considered more magic to autogenerate the string key from the line number of the enclosing val name, but thought better of it.

test/junit/scala/tools/testing/ClearAfterClass.java
lrytz, line 22: i see

lrytz:
LGTM!

PR #5159, opened 2016-05-12
Add check to scala REPL package to improve feedback about implicits

DieBauer:
When the repl is started with additional compiler flags that prevent implicits
being in scope (like -Yno-predef) the default message of implicits in scope
using :implicits stays the same. 
This additional check will return the same message if predef is indeed in
scope. And an adjusted message if Predef is not in scope.

``` console
scala -Yno-predef
scala> :implicits
No implicits have been imported other than those in Predef.
```

In the new situation this will be catched and return a more correct answer about in scope implicits:

``` console
scala -Yno-predef
scala> :implicits
No implicits have been imported.
```

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/package.scala
+++ b/src/repl/scala/tools/nsc/interpreter/package.scala
@@ -88,9 +88,6 @@ package object interpreter extends ReplConfig with ReplStrings {
           }
       }
 
-      if (filtered.isEmpty)
-        return "No implicits have been imported other than those in Predef."
-
       filtered foreach {
         case (source, syms) =>
           p("/* " + syms.size + " implicit members imported from " + source.fullName + " */")
@@ -126,7 +123,14 @@ package object interpreter extends ReplConfig with ReplStrings {
           }
           p("")
       }
-      ""
+      
+      if (filtered.nonEmpty) 
+        "" // side-effects above
+      else if (global.settings.nopredef || global.settings.noimports) 
+        "No implicits have been imported."
+      else
+        "No implicits have been imported other than those in Predef." 
+
     }
 
     def kindCommandInternal(expr: String, verbose: Boolean): Unit = {
--- a/test/files/run/repl-implicits-nopredef.scala
+++ b/test/files/run/repl-implicits-nopredef.scala
@@ -1,0 +1,10 @@
+import scala.tools.partest.ReplTest
+import scala.tools.nsc.Settings
+
+object Test extends ReplTest {
+  override def transformSettings(settings: Settings): Settings = {
+    settings.nopredef.value = true
+    settings
+  }  
+  def code = ":implicits"
+}
--- a/test/files/run/repl-implicits-nopredef.check
+++ b/test/files/run/repl-implicits-nopredef.check
@@ -1,0 +1,5 @@
+
+scala> :implicits
+No implicits have been imported.
+
+scala> :quit
--- a/test/files/run/repl-implicits.check
+++ b/test/files/run/repl-implicits.check
@@ -1,0 +1,5 @@
+
+scala> :implicits
+No implicits have been imported other than those in Predef.
+
+scala> :quit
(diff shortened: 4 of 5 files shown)


src/repl/scala/tools/nsc/interpreter/package.scala
```diff
@@ -88,8 +88,12 @@ package object interpreter extends ReplConfig with ReplStrings {
           }
       }
 
-      if (filtered.isEmpty)
+      if (filtered.isEmpty) {
+        if (global.settings.nopredef || global.settings.noimports) {
+          return "No implicits have been imported."
+        }
         return "No implicits have been imported other than those in Predef."
+      }
```
som-snytt: You could avoid the returns by moving the if to result position.

test/files/run/repl-implicits-nopredef.scala
```diff
@@ -0,0 +1,10 @@
+import scala.tools.partest.ReplTest
+import scala.tools.nsc.Settings
+
+object Test extends ReplTest {
+  override def transformSettings(settings: Settings): Settings = {
+    settings.nopredef.value = true
+    settings
+  }  
+  def code = ":implicits\n"
```
som-snytt: The newline should be unnecessary. ILoop.runForTranscript adds one for good measure after stripping yours: `val input = new BufferedReader(new StringReader(code.trim + "\n"))`.

DieBauer: That's true. Fixed.

src/repl/scala/tools/nsc/interpreter/package.scala
DieBauer: Is it desirable to get rid of the return statements? 
Right now you know the 'termination' of the method if these conditions are met.

(I know the foreach won't do anything if the conditions are met as well, but could make it harder to read through the end of the method).

DieBauer:
Should we rewrite the return into a case match?

``` scala
val replString = filtered match {
  case Nil if (global.settings.nopredef || global.settings.noimports) =>
     "No implicits have been imported."
  case Nil => "No implicits have been imported other than those in Predef."
  case _ => ""
```

src/repl/scala/tools/nsc/interpreter/package.scala
som-snytt: Always desirable to get rid of returns. The code is confusing anyway b/c side-effects. I don't know why it doesn't return a string to output? I don't want to know. If you really think baby-step of if-else at bottom is too confusing, put the foreach behind a `def printStuff()` so it's `if (stuff.isEmpty) "str" else { printStuff(); "" }` or similar.

som-snytt:
The `isEmpty` is clear. Don't make me wonder if `filtered` is a `List`.

src/repl/scala/tools/nsc/interpreter/package.scala
lrytz: Agree, this makes the side effects more clear.

lrytz:
can you also squash all commits into one?

src/repl/scala/tools/nsc/interpreter/package.scala
```diff
@@ -130,7 +123,18 @@ package object interpreter extends ReplConfig with ReplStrings {
...
+        if (global.settings.nopredef || global.settings.noimports) {
+          "No implicits have been imported."
+        }
+        else {
+          "No implicits have been imported other than those in Predef." 
+        }
+      }
+      else {
+        ""
+      }
```
lrytz: i'd avoid some of the curly braces and newlines

```
if (filtered.nonEmpty)
  "" // side-effects above
else if (global.settings.nopredef || global.settings.noimports)
  ".."
else
  ".."
```

DieBauer: You're right, I wasn't totally awake ;). Will adjust and squash.

lrytz:
LGTM!

PR #5160, opened 2016-05-12
SI-9773 linesWithSeparators is like split

som-snytt:
StringLike.linesWithSeparators is like split but
preserves the separators.

Change the doc and implementation to align.

In particular, an empty string should yield one element.

### DIFF
--- a/src/library/scala/collection/immutable/StringLike.scala
+++ b/src/library/scala/collection/immutable/StringLike.scala
@@ -100,19 +100,23 @@ self =>
   /** Return all lines in this string in an iterator, including trailing
    *  line end characters.
    *
-   *  The number of strings returned is one greater than the number of line
-   *  end characters in this string. For an empty string, a single empty
-   *  line is returned. A line end character is one of
-   *  - `LF` - line feed   (`0x0A` hex)
-   *  - `FF` - form feed   (`0x0C` hex)
+   *  This method is analogous to `s.split(EOL).toIterator`,
+   *  except that any existing line endings are preserved in the result strings.
+   *
+   *  The empty string yields an iterator with one element, the empty string.
+   *
+   *  A line end character is one of
+   *  - `LF` - line feed   (`0x0A`)
+   *  - `FF` - form feed   (`0x0C`)
    */
   def linesWithSeparators: Iterator[String] = new AbstractIterator[String] {
     val str = self.toString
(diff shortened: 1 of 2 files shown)


som-snytt:
Two presentation compiler tests have an extra empty line of output. Actually, I can't tell at a glance how the tests work or what produces output. Probably it's just `--update-check`?

```
  [partest] !! 28 - presentation/t8941b                       [output differs]
  [partest] ok 29 - presentation/scope-completion-1         
  [partest] ok 30 - presentation/scope-completion-3         
  [partest] ok 31 - presentation/partial-fun                
  [partest] ok 32 - presentation/hyperlinks                 
  [partest] ok 33 - presentation/ide-t1000976               
  [partest] !! 34 - presentation/forgotten-ask.scala          [output differs]
```

som-snytt:
This is where I really wish partest were reunited with the mother ship.

When the log file is scrubbed of anomalies like path strings, `mapInPlace` relies on `fileLines` returning `Nil` for an empty file. Note that `fileContents` returns an empty string for a missing file.

https://github.com/scala/scala-partest/blob/v1.0.14/src/main/scala/scala/tools/partest/package.scala#L59

https://github.com/scala/scala-partest/blob/v1.0.14/src/main/scala/scala/tools/partest/package.scala#L81

After this change, partest trivially normalizes an empty log file to a single newline.

For the record, this took a while to track down. I'd looked at the "diff" logic and figured there was an append somewhere, because by coincidence earlier today the pattern surfaced:

https://github.com/scala/scala/pull/5159#discussion_r63045883

```
val input = new BufferedReader(new StringReader(code.trim + "\n"))
```

But actually I was suspicious of `sys.process` and poked in there, too, and got some kind of greasy soot under my fingernails that just won't come out. Finally had to breakpoint the new line of code.

som-snytt:
In case it is useful,

https://github.com/scala/scala-partest/pull/59

I was able to publish-local, with bumped versions (1.0.15 actually), and run sbt-partest. That was worth learning about. What advances in technology since I've been asleep!

som-snytt:
This change requires the partest fix, of course.

lrytz:
probably it's safer not to change `.lines`, so i vote for #5161.

som-snytt:
The current behavior is not counter-intuitive.

PR #5162, opened 2016-05-14
SI-9361 fixed assert allowing display of improved error message.

milessabin:
Bug fixed during Typelevel hackday at Hack the Tower :-)


lrytz:
LGTM

PR #5164, opened 2016-05-15
Improve Future documentation (+ minor code cleanups)

viktorklang:
Primarily improves the documentation but also makes some minor adjustments to implementation, without semantical difference.

Also adds a "missing" test case to ensure that it is possible to add numerous callbacks in a chain and still have the correct number of invocations. (this is an important test esp. if the representation of the callbacks for the standard implementation changes down the line)

### DIFF
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -20,11 +20,14 @@ import scala.collection.generic.CanBuildFrom
 import scala.reflect.ClassTag
 
 
-/** The trait that represents futures.
+/** A `Future` represents a value which may or may not *currently* be available,
+ *  but will be available at some point, or an exception if that value could not be made available.
  *
- *  Asynchronous computations that yield futures are created with the `Future.apply` call:
+ *  Asynchronous computations that yield futures are created with the `Future.apply` call and are computed using a supplied `ExecutionContext`,
+ * which can be backed by a Thread pool.
  *
  *  {{{
+ *  import ExecutionContext.Implicits.global
  *  val s = "Hello"
  *  val f: Future[String] = Future {
  *    s + " future!"
@@ -88,6 +91,7 @@ import scala.reflect.ClassTag
  * thread. That is, the implementation may run multiple callbacks
  * in a batch within a single `execute()` and it may run
  * `execute()` either immediately or asynchronously.
+ * Completion of the Future must *happen-before* the invocation of the callback.
  */
 trait Future[+T] extends Awaitable[T] {
   import Future.{ InternalCallbackExecutor => internalExecutor }
@@ -101,9 +105,13 @@ trait Future[+T] extends Awaitable[T] {
    *  If the future has already been completed with a value,
    *  this will either be applied immediately or be scheduled asynchronously.
    *
+   *  Note that the returned value of `pf` will be discarded.
+   *
    *  $swallowsExceptions
    *  $multipleCallbacks
    *  $callbackInContext
+   *
+   * @group Callbacks
    */
   @deprecated("use `foreach` or `onComplete` instead (keep in mind that they take total rather than partial functions)", "2.12")
   def onSuccess[U](pf: PartialFunction[T, U])(implicit executor: ExecutionContext): Unit = onComplete {
@@ -122,9 +130,13 @@ trait Future[+T] extends Awaitable[T] {
    *
    *  Will not be called in case that the future is completed with a value.
    *
+   *  Note that the returned value of `pf` will be discarded.
+   *
    *  $swallowsExceptions
    *  $multipleCallbacks
    *  $callbackInContext
+   *
+   * @group Callbacks
    */
   @deprecated("use `onComplete` or `failed.foreach` instead (keep in mind that they take total rather than partial functions)", "2.12")
   def onFailure[U](@deprecatedName('callback) pf: PartialFunction[Throwable, U])(implicit executor: ExecutionContext): Unit = onComplete {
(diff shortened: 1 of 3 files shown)


src/library/scala/concurrent/Future.scala
```diff
@@ -101,9 +107,13 @@ trait Future[+T] extends Awaitable[T] {
    *  If the future has already been completed with a value,
    *  this will either be applied immediately or be scheduled asynchronously.
    *
+   *  Note that the returned value of `pf` will be discarded.
+   *
    *  $swallowsExceptions
    *  $multipleCallbacks
    *  $callbackInContext
+   *
+   * @group Callbacks
```
janekdb, line 114: Use @groupprio in the main comment to pick an order for these groups?

viktorklang, line 114: Not sure there is a prio really. What does it pick by default? Lexicographical sort?

janekdb, line 114: Looks like lexicographical so with `Transformations` not sandwiched by `Polling` and `Callbacks` the default ordering is fine.

scaladoc-futures

Is `andThen` more of a transformation than a callback?

viktorklang, line 114: @janekdb `andThen` does not modify the result of the returned Future, it merely injects an effect between `this`-Future and `that`-Future (where this is the Future it is invoked upon and that is the Future which is returned by the invokation)

src/library/scala/concurrent/Future.scala
```diff
@@ -618,7 +664,7 @@ object Future {
    */
   def sequence[A, M[X] <: TraversableOnce[X]](in: M[Future[A]])(implicit cbf: CanBuildFrom[M[Future[A]], A, M[A]], executor: ExecutionContext): Future[M[A]] = {
     in.foldLeft(successful(cbf(in))) {
-      (fr, fa) => for (r <- fr; a <- fa) yield (r += a)
```
retronym, line 621: Just trying to deconstruct this change, I'm always a bit wary of the "... and just changed a little detail in the implementation" parts of doc commits 😄 

If I read things correctly, this change does three things:
- makes the code more concise
- makes it more efficient for `failed` and Future.never`
- uses `InternalCallbackExecutor`, for the `flatMap` (the outer generator in the for comprehension) while still using `executor` for the `map` (the inner generator).
- uses `InternalCallbackExecutor` for the explicitly written `map` that remains in this method.

@viktorklang Is this list correct? What's the rationale? What changes for the user?

viktorklang, line 621: > Just trying to deconstruct this change, I'm always a bit wary of the "... and just changed a little detail in the implementation" parts of doc commits

Now that, @retronym, is a sign of good health :)

You're right on those accounts, and also: it shows what `zipWith` is good for (the code which was replaced is an alternate, slower encoding).

(Also, I'd argue, that it makes the code easier to reason about since you don't have to desugar the FC in your head :))

src/library/scala/concurrent/Future.scala
```diff
@@ -618,7 +664,7 @@ object Future {
...
+      (fr, fa) => fr.zipWith(fa)(_ += _)
```
viktorklang, line 665: @retronym Speaking of this line btw, will the compiler hoist the allocation since it is a pure function? (I'm going to assume: no, in which case it perhaps ought to be manually hoisted) (I have no access to the bytecode on this machine)

src/library/scala/concurrent/Future.scala
```diff
@@ -20,11 +20,16 @@ import scala.collection.generic.CanBuildFrom
 import scala.reflect.ClassTag
 
 
-/** The trait that represents futures.
+/** 
  *
- *  Asynchronous computations that yield futures are created with the `Future.apply` call:
+ *  A `Future` represents a value which may or may not *currently* be available,
```
ktoso: This scaladoc is inconsistent (in style) with the others, fix this or the others?

src/library/scala/concurrent/Future.scala
```diff
@@ -791,10 +837,9 @@ object Future {
    * @return          the `Future` of the `TraversableOnce` of results
    */
   def traverse[A, B, M[X] <: TraversableOnce[X]](in: M[A])(fn: A => Future[B])(implicit cbf: CanBuildFrom[M[A], B, M[B]], executor: ExecutionContext): Future[M[B]] =
-    in.foldLeft(successful(cbf(in))) { (fr, a) =>
-      val fb = fn(a)
-      for (r <- fr; b <- fb) yield (r += b)
-    }.map(_.result())
+    in.foldLeft(successful(cbf(in))) {
+      (fr, a) => fr.zipWith(fn(a))(_ += _)
+    }.map(_.result())(InternalCallbackExecutor)
```
ktoso, line 840: Good. Nice to see Future is using this trick more and more internally 👍 
Also: reads much better now.

ktoso:
LGTM (not that I have LGTMing powers on this repo though ;-)), read through and pondered the changes for a while – seems to me worth the inclusion.

src/library/scala/concurrent/Future.scala
retronym, line 665: Yep, since 2.12.0-M4 we hoist non-capturing lambda (actually `LambdaMetaFactory` does it for us)

```
scala> val List(f1, f2) = (0 to 1).map(_ => (x: Any) => x).toList
f1: Any => Any = $$Lambda$1329/1454366618@64fe9da7
f2: Any => Any = $$Lambda$1329/1454366618@64fe9da7

scala> f1 eq f2
res0: Boolean = true
```

ktoso, line 665: Awesome news! :-)

retronym:
LGTM, too.

src/library/scala/concurrent/Future.scala
viktorklang, line 665: @retronym That's music to my ears. I guess I could remove some of the manual hoistings in this file then :)

src/library/scala/concurrent/Future.scala
viktorklang: ?

ktoso: Most of this file is using the 

```
/** First first
 *   Second
```

style, this one is a bit off

janekdb: I'd suggest starting to adopt the style here: http://docs.scala-lang.org/style/scaladoc.html

```
/** Hi there
  * 
  * Line 1
  * Line 2
  */
```

This has the advantage of surviving the auto formatting attentions of IDEA 16 code format and requires less column-edit-mode-fu to maintain.

viktorklang: Aye, will address.

src/library/scala/concurrent/Future.scala
```diff
@@ -20,11 +20,14 @@ import scala.collection.generic.CanBuildFrom
 import scala.reflect.ClassTag
 
 
-/** The trait that represents futures.
+/** A `Future` represents a value which may or may not *currently* be available,
+ *  but will be available at some point, or an exception if that value could not be made available.
  *
- *  Asynchronous computations that yield futures are created with the `Future.apply` call:
+ *  Asynchronous computations that yield futures are created with the `Future.apply` call and are computed using a supplied `ExecutionContext`,
+ * which can be backed by a Thread pool.:
```
soc: Either `.` or `:`?

viktorklang: Nice catch :)

PR #5165, opened 2016-05-16
SI-9767 String->integer/float conversions now trim and throw consiste…

intronic:


Ichoran:
I don't think we can consider this without a benchmark showing that the trim doesn't inordinately affect performance.  I'm especially wary of the implications for memory usage if someone tries to attack a system by passing a big non-numeric string.  This automatically doubles memory usage if they use leading or trailing spaces, which might not be a disaster, but would make it more effective.

intronic:
Thanks for your comments. A few questions (pls excuse, first time contributor), or suggestion welcome. 
- can you suggest an acceptable scala benchmark tool/procedure so I can do this?
- also note, a mitigating factor for the memory, attack, and performance issue concerns: this is already the case for .toDouble/.toFloat - Java already does the .trim() as the first line of code. This PR just extends it to the integer conversions.
- alternatively I could just update the documentation, as per java doc for toInt/toDouble, which explains the difference in behaviour

Cheers.

intronic:
Thinking about the memory, attack, and performance issue some more:
- all the conversions currently (eg, including .toInt) are already vulnerable to attack with strings with an unlimited number of leading zeros.
- Note that longest string a Long/Int/Double should need (without space or zero padding) string is only around 23 chars, so some sort of guard on 'excessive' length strings could be applied to fix up the existing problem?
- How to take this thread offline to work it out?

Thanks for your time...

retronym:
My vote would be to document the inconsistency but to leave the status quo. Increasingly, we are trying to avoid small deviations from the conventions of the underlying Java standard library (even ones that appear in isolation to be a small win), especially around numerics.

One example of a case where we went against this convention, and now regret, is offering `scala.Double.{MinPositiveValue, MinValue}` vs `java.lang.Double.MIN_VALUE`). I don't think your change as anywhere near as troublesome, but just wanted to give an example to show why I'm erring on the side of caution in this area.

intronic:
Per suggestion, I changed PR to just better document the existing behaviour, no change to code, and updated test cases to reflect behaviour.

Not sure whether I should have closed the previous PR and made a new one, add to it like I've done.

src/library/scala/collection/immutable/StringLike.scala
```diff
@@ -288,27 +288,35 @@ self =>
...
    */
   def toInt: Int         = java.lang.Integer.parseInt(toString)
   /**
+   * Parse string as a Long (string must contain only decimal digits and optional leading '-').
    * @throws java.lang.NumberFormatException  - If the string does not contain a parsable long.
    */
   def toLong: Long       = java.lang.Long.parseLong(toString)
   /**
-   * @throws java.lang.NumberFormatException - If the string does not contain a parsable float.
+    * Parse string as a Double (surrounding whitespace removed with trim).
```
Ichoran: It's a bit of a nitpick, but I think this reads better as `(surrounding whitespace is removed with a trim)`.  Complete sentences FTW?

Also, I think all the instances of `string` in the description of the method are redundant.  Just `Parse as a Long` is enough--it's clear from context that it's talking about the string.  (You need the word `string` in the description of the exceptions.  Those should stay as-is.)

src/library/scala/collection/immutable/StringLike.scala
```diff
@@ -288,27 +288,35 @@ self =>
...
+    * @throws java.lang.NumberFormatException - If the string does not contain a parsable float.
+    * @throws java.lang.NullPointerException - If the string is null.
    */
   def toFloat: Float     = java.lang.Float.parseFloat(toString)
   /**
-   * @throws java.lang.NumberFormatException - If the string does not contain a parsable double.
+    * Parse string as a Float (surrounding whitespace removed with trim).
+    * @throws java.lang.NumberFormatException - If the string does not contain a parsable double.
```
retronym: The `-` should be removed after the exception.

src/library/scala/collection/immutable/StringLike.scala
retronym: Backticks around Double to render it as monospace. Same applies to `Long`, etc. above.

intronic:
OK fixed sentences/string/- in throws/backticks.
Rebased to latest 2.12.x

retronym:
@intronic Could you please squash this into a single commit?

```
git reset 2.12.x; git add -A; git commit
```

retronym:
@intronic I've squashed the commits through the GitHub UI and merged the result: https://github.com/scala/scala/commit/608387529b790c2a0fc87b4cf5b5324d68d02648

PR #5167, opened 2016-05-16
SI 9766 - allow ++ on empty ConcatIterator

martijnhoekstra:
review by @Ichoran , @szeiger 

This is the simplest way to fix this, but I'm not sure it's the cleanest way. Letting `++` do more work by checking if `current eq null` and calling `advance()` if so may be nicer.

### DIFF
--- a/test/junit/scala/collection/IteratorTest.scala
+++ b/test/junit/scala/collection/IteratorTest.scala
@@ -192,4 +192,22 @@ class IteratorTest {
     assertSameElements(exp, res)
     assertEquals(8, counter) // was 14
   }
+
+  // SI-9766
+  @Test def exhaustedConcatIteratorConcat: Unit = {
+    def consume[A](i: Iterator[A]) = {
+      while(i.hasNext) i.next()
+    }
+    val joiniter = Iterator.empty ++ Seq(1, 2, 3)
+    assertTrue(joiniter.hasNext)
+    consume(joiniter)
+    val concatiter = joiniter ++ Seq(4, 5, 6)
+    assertTrue(concatiter.hasNext)
+    consume(concatiter)
+    assertFalse(concatiter.hasNext)
+    val concatFromEmpty = concatiter ++ Seq(7, 8, 9)
+    assertTrue(concatFromEmpty.hasNext)
+    consume(concatFromEmpty)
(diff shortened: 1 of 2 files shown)


som-snytt:
The cleanest and simplest fix would be to backport the shiny new ConcatIterator.

https://github.com/scala/scala/commit/f7019903bf9c3f88f996253bb88bbd4264b28be6#diff-0f291618b38e49124a043fadd6fcd618

unless there is some subtle reason it couldn't go into 2.11.x.

test/junit/scala/collection/IteratorTest.scala
```diff
@@ -192,4 +192,26 @@ class IteratorTest {
     assertSameElements(exp, res)
     assertEquals(8, counter) // was 14
   }
+  
+  // SI-9766
+  @Test def exhaustedConcatIteratorConcat: Unit = {
+    var i: Iterator[Int] = Iterator.empty
```
som-snytt: This is the ticket example, but mutables are as hard to read in unit tests as in real code.

martijnhoekstra:
@som-snytt If that can be backported, that's much nicer. I assumed it couldn't, because I assumed that if it could, f701990 would have targeted 2.11.x rather than 2.12.x.
especially since https://github.com/scala/scala/commit/ecee7b75328ae1f856b5fa832ebad7fc9e42f64a was committed less than a week before the new ConcatIterator. Hopefully wrong assumptions.

som-snytt:
I guess binary compatibility is the problem.

martijnhoekstra:
The more I think about it, letting ++ do the "work", which isn't much, is much, much nicer, as it doesn't break the invariant that if current eq null, the iterator is exhausted.

Alternative implementation in https://github.com/martijnhoekstra/scala/commit/c25d36e0882db715adc933832570362a7f18d1dd

som-snytt:
I think your alternative intuition is definitely correct, and for the reason you give.

However, constructing the iterator immediately sacrifices laziness. The added collection is by-name, so neither the traversable nor the iterator is created until invoked.

Instead, you could `new JoinIterator(Iterator.empty, that)` for that case.

martijnhoekstra:
Ah, I see. I changed that.

lrytz:
LGTM!

PR #5169, opened 2016-05-17
SI-4625 Recognize App in script

som-snytt:
Cheap name test: if the script object extends "App",
take it for a main-bearing parent.

Note that if `-Xscript` is not `Main`, the default,
then the source is taken as a snippet and there is
no attempt to locate an existing `main` method.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -380,11 +380,15 @@ self =>
           case DefDef(_, nme.main, Nil, List(_), _, _)  => true
           case _                                        => false
         }
+        def isApp(t: Tree) = t match {
+          case Template(ps, _, _) => ps.exists { case Ident(x) if x.decoded == "App" => true ; case _ => false }
+          case _ => false
+        }
         /* For now we require there only be one top level object. */
         var seenModule = false
         val newStmts = stmts collect {
           case t @ Import(_, _) => t
-          case md @ ModuleDef(mods, name, template) if !seenModule && (md exists isMainMethod) =>
+          case md @ ModuleDef(mods, name, template) if !seenModule && (isApp(template) || md.exists(isMainMethod)) =>
             seenModule = true
             /* This slightly hacky situation arises because we have no way to communicate
              * back to the scriptrunner what the name of the program is.  Even if we were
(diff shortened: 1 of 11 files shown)


src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -380,11 +380,15 @@ self =>
           case DefDef(_, nme.main, Nil, List(_), _, _)  => true
           case _                                        => false
         }
+        def isApp(t: Tree) = t match {
+          case Template(ps, _, _) => ps.exists { case Ident(x) if x.decoded == "App" => true ; case _ => false }
```
retronym: `x == tpnme.App` would be more idiomatic. That form avoids needless string creation.

retronym: or even `case Ident(tpnme.App) =>`

retronym:
LGTM, nitpick aside.

retronym:
Scope creep request: Would be great if we could detect Appy modules in files with more than one definition.

```
% cat sandbox/test.scala && scala sandbox/test.scala
object Test {  def main(args: Array[String]) { println("!!") }}
!!

% cat sandbox/test.scala && scala sandbox/test.scala
object O
object Test {  def main(args: Array[String]) { println("!!") }}

%
```

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
som-snytt: Thx, I grepped for it but missed it, trying to beat the evening rush home. Unless you mean add it to names.

retronym: Yep, I mean to add it.

som-snytt:
I considered a REPLish strategy, put the entire text in a package object.

Edit: didn't do that here. But it now allows

```
trait X { def x = 42 }

object Main extends X with App {
  println(x)
}
```

som-snytt:
I thought the scope creep was the guy with binoculars who peers in people's windows.

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -407,54 +407,58 @@ self =>
...
         else {
           if (seenModule)
             warning(disallowed.pos.point, "Script has a main object but statement is disallowed")
-          None
+          EmptyTree
         }
       }
 
-      if (mainModuleName == newTermName(ScriptRunner.defaultScriptMain))
-        searchForMain() foreach { return _ }
```
lrytz, line 406: wow, never seen that pattern before..!

lrytz:
LGTM, too

som-snytt:
I just expected this to work, but I guess it hasn't been merged forward yet...

PR #4117, opened 2014-11-09
SI-8449 Fix spurious error with Java raw type over a Scala class

retronym:
When the Scala typechecker typechecks signatures in Java sources,
it must enable use of a type constructor without applied type
arguments (a raw type).

However, the check in `adaptType::properTypeRequired` was too
restricive: it only allows raw types given a) we are typechecking a
Java source file and b) the type constructor itself is Java defined.

The second check does not make sense; it is perfectly legal to define
a Java raw type for over a Scala defined type constructor. This commit
removes it.

The bug was noticed in the Scaladoc context, as it explores the
signatures of all Java sources so it can document them. But the
enclosed test demonstrates the underlying bug under normal
compilation.

Review by @adriaanm

PR #5172, opened 2016-05-17
Move t8449 to correct place

lrytz:
Follow-up for https://github.com/scala/scala/pull/4117


lrytz:
```
...
  [partest] ok 1027 - pos/t8449                               
...
```

retronym:
LGTM

PR #5173, opened 2016-05-17
Group Console and AnsiColor entities and add usage examples

janekdb:
- Group AnsiColor and Console members
- Add Console IO redefinition examples
- Add Console control code examples
- Add color swatches to AnsiColor
- Add AnsiColor control code examples
- Link to StdIn for reading
- Fix link syntax for java.util.Formatter
- Fix withErr method example
### Reviewing

The matching Scaladoc is available here to make reviewing this easier,
- http://janekdb.github.io/scala/PR-5173/scala/Console$.html
- http://janekdb.github.io/scala/PR-5173/scala/io/AnsiColor.html

Happy to take reviews from anyone with the inclination!
### Comment Format

`AnsiColor` follows the commenting style here: http://docs.scala-lang.org/style/scaladoc.html whereas `Console` sticks to the older style because this had a greater presence in `Console`.
### IO Redefinition versus IO Redirection

I wasn't convinced the mechanism for substituting alternative streams in `Console` qualified as redirection but I'm open to persuasion.

### DIFF
--- a/src/library/scala/Console.scala
+++ b/src/library/scala/Console.scala
@@ -1,6 +1,6 @@
 /*                     __                                               *\
 **     ________ ___   / /  ___     Scala API                            **
-**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
+**    / __/ __// _ | / /  / _ |    (c) 2003-2016, LAMP/EPFL             **
 **  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
 ** /____/\___/_/ |_/____/_/ | |                                         **
 **                          |/                                          **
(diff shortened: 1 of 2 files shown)


janekdb:
Low bit depth previews. Please use links above for full versions.

scaladoc-console-pr-5173-v1
scaladoc-ansicolor-pr-5173-v1

janekdb:
These previews are out of date. Use the links above.

scaladoc-console-pr-5173-v3

scaladoc-ansicolor-pr-5173-v2

src/library/scala/io/AnsiColor.scala
```diff
@@ -1,52 +1,164 @@
...
+  * @groupprio color-white 180
+  */
 trait AnsiColor {
-  /** Foreground color for ANSI black */
+  /** Foreground color for ANSI black
+    * @group color-black
+    **/
   final val BLACK      = "\u001b[30m"
-  /** Foreground color for ANSI red */
+  /** Foreground color for ANSI red
```
janekdb, line 76: Alignment. This needs fixing here and in all other cases.

janekdb:
Error was `*** Error in `/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java': free(): invalid pointer: 0x00000001003939a8 ***`

src/library/scala/io/AnsiColor.scala
```diff
@@ -1,52 +1,163 @@
 package scala
 package io
 
+/** ANSI escape codes providing control over text formatting and color on supporting text terminals.
+  *
+  * ==ANSI Style and Control Codes==
+  *
+  * This group of escape codes provides control over text styling. For example turn on reverse video with bold and
```
som-snytt: "For example, to turn..."

src/library/scala/io/AnsiColor.scala
```diff
@@ -1,52 +1,163 @@
...
+  * then turn off all styling,
+  *
+  * {{{
+  * import io.AnsiColor._
+  *
+  * object ColorDemo extends App {
+  *
+  *   println(s"${REVERSED}${BOLD}Hello 1979!${RESET}")
```
som-snytt: I would have said "Party like it's 1979" but that's just me.

som-snytt:
Is there any sentiment for aliases `Red` for `RED` and `Blink` for `BLINK`?  I renamed on import the couple of times I used it. I meant to contribute aliases with deprecations a long time ago. All caps really does feel like 1979.

Edit: if there is sentiment, I'll contribute the edit. But now would be the time.

src/library/scala/Console.scala
```diff
@@ -12,12 +12,115 @@ import java.io.{ BufferedReader, InputStream, InputStreamReader, OutputStream, P
...
+ *      def isPrime(): Unit = {
+ *
+ *        val candidate = io.StdIn.readInt().ensuring(_ > 1)
+ *
+ *        val prime = (2 to candidate - 1).forall(candidate % _ != 0)
+ *
+ *        if (prime)
+ *          Console.println(s"${RESET}${GREEN}yes${RESET}")
+ *        else
+ *          Console.err.println(s"${RESET}${YELLOW_B}${RED}${UNDERLINED}NO!${RESET}")
```
soc, line 46: Can't we drop the curly braces?

janekdb, line 46: Sure. I'll remove as many as possible.

janekdb:
Changes made and generated docs push to GitHub pages. Preview not updated.

src/library/scala/Console.scala
janekdb, line 46: @soc - Now with minimal curly braces.

scaladoc-pr-5173-without-curlies

Initial version with Curlies<sup>MAX</sup>

scaladoc-pr-5173-with-curlies

I prefer the initial version because it's easier to pick out the codes. WDYT?

janekdb:
@som-snytt I'd got no particular appetite for `Red` and `Blink` but very happy to see these extensions added if you want to. We can drop `RED` etc out of the groups and mark as deprecated.

som-snytt:
Since naming is hard, I don't have a good way to improve `RED_B`. Cleanest might be a separate name space, such as

```
object background { final val Red = RED_B }
```

or

```
object Red { final val background = RED_B ; override final val toString = RED }
```

or a `background` macro extension method that automates `Red.background` for the color constants.

It's harder to pick up the constants from `object Red`, where you'd want your `s"${Red}Hi, $name${Reset}"` or similar to be constant folded, whether by `Red + "Hi"` or constant-aware `f"${Red}Hi"`.

Although I don't like that `RED_B` is all caps and has a suffix that isn't instantly mnemonic, I don't want to create needless work, especially since the new color bars are so pretty.

src/library/scala/Console.scala
```diff
@@ -12,12 +12,115 @@ import java.io.{ BufferedReader, InputStream, InputStreamReader, OutputStream, P
...
+ *    object PrimeTest {
+ *
+ *      def isPrime(): Unit = {
+ *
+ *        val candidate = io.StdIn.readInt().ensuring(_ > 1)
+ *
+ *        val prime = (2 to candidate - 1).forall(candidate % _ != 0)
+ *
+ *        if (prime)
+ *          Console.println(s"$RESET${GREEN}yes$RESET")
```
som-snytt: I happen to find the uniform braces easier to read. Another option:

```
def greenly(s: String) = s"$RESET$GREEN$s$RESET"
println(greenly("yes"))
```

janekdb: @soc - Having tried out brace minimisation I'm keen to restore the fully braced strings. I feel this makes the embedded constants stand out more which is useful in this context. What do you say?

lrytz: +1 for fully curly braced strings here

lrytz:
LGTM other than the curlies

janekdb:
@lrytz - Fully curly strings restored.
@soc - I can see that less curly braces would be preferable if you have become accustomed to this more streamlined format. I think for tutorial purposes the fully curlies have an advantage when scanning the code.

All changes done. Please re-review.

lrytz:
LGTM!

PR #5174, opened 2016-05-18
SI-8756 Fix generic signature for refinement of primitive

retronym:
Java generic signature generation was making the wrong
assumption about how refinement types should erase to
Java generics.

This commit passes through the current value of `primitiveOk`,
rather than forcing it to `true`.

This flag is true when generating the signature for `f2`,
but false in `i2` (as we are in a type argument position).


retronym:
Review by @lrytz or @adriaanm

lrytz:
LGTM

PR #5175, opened 2016-05-18
SI-9656 Distinguish Numeric with step type

som-snytt:
Shorthand `toString` for `Range` and `NumericRange`.

```
scala> 1 to 10 by 2
res0: scala.collection.immutable.Range = inexact Range 1 to 10 by 2

scala> 1 until 1 by 2
res1: scala.collection.immutable.Range = empty Range 1 until 1 by 2

scala> .1 to 1.0 by .1
<console>:1: error: ';' expected but double literal found.
res1.1 to 1.0 by .1
    ^

scala> 0.1 to 1.0 by .1
res2: scala.collection.immutable.NumericRange[Double] = NumericRange 0.1 to 1.0 by 0.1

scala> Range.Double(.1, 1.0, .1)
res3: scala.collection.immutable.NumericRange[Double] = NumericRange 0.1 until 1.0 by 0.1 (using NumericRange 0.1 until 1.0 by 0.1 of BigDecimal)

scala> 0.0 to 1.0
res4: Range.Partial[Double,scala.collection.immutable.NumericRange[Double]] = Range requires step

scala> 0 to 1
res5: scala.collection.immutable.Range.Inclusive = Range 0 to 1
```

No parens because `Range(0 to 1)` looks like a factory but just gives a confusing error message.

### DIFF
--- a/src/library/scala/collection/immutable/Range.scala
+++ b/src/library/scala/collection/immutable/Range.scala
@@ -396,22 +396,20 @@ extends scala.collection.AbstractSeq[Int]
     case _ =>
       super.equals(other)
   }
-  /** Note: hashCode can't be overridden without breaking Seq's
-   *  equals contract.
-   */
 
-  override def toString() = {
-    val endStr =
-      if (numRangeElements > Range.MAX_PRINT || (!isEmpty && numRangeElements < 0)) ", ... )" else ")"
-    take(Range.MAX_PRINT).mkString("Range(", ", ", endStr)
+  /* Note: hashCode can't be overridden without breaking Seq's equals contract. */
+
+  override def toString = {
+    val preposition = if (isInclusive) "to" else "until"
+    val stepped = if (step == 1) "" else s" by $step"
+    val prefix = if (isEmpty) "empty " else if (!isExact) "inexact " else ""
+    s"${prefix}Range $start $preposition $end$stepped"
   }
 }
 
 /** A companion object for the `Range` class.
  */
 object Range {
-  private[immutable] val MAX_PRINT = 512  // some arbitrary value
-
   /** Counts the number of range elements.
    *  @pre  step != 0
    *  If the size of the range exceeds Int.MaxValue, the
@@ -514,6 +512,7 @@ object Range {
   // we offer a partially constructed object.
   class Partial[T, U](f: T => U) {
     def by(x: T): U = f(x)
+    override def toString = "Range requires step"
   }
 
   // Illustrating genericity with Int Range, which should have the same behavior
--- a/src/library/scala/collection/immutable/NumericRange.scala
+++ b/src/library/scala/collection/immutable/NumericRange.scala
@@ -161,6 +161,12 @@ extends AbstractSeq[T] with IndexedSeq[T] with Serializable {
       override def isEmpty = underlyingRange.isEmpty
       override def apply(idx: Int): A = fm(underlyingRange(idx))
       override def containsTyped(el: A) = underlyingRange exists (x => fm(x) == el)
+
+      override def toString = {
+        def simpleOf(x: Any): String = x.getClass.getName.split("\\.").last
+        val stepped = simpleOf(underlyingRange.step)
+        s"${super.toString} (using $underlyingRange of $stepped)"
+      }
     }
   }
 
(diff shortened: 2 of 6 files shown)


src/library/scala/collection/immutable/Range.scala
```diff
@@ -400,18 +400,17 @@ extends scala.collection.AbstractSeq[Int]
...
 
-  override def toString() = {
-    val endStr =
-      if (numRangeElements > Range.MAX_PRINT || (!isEmpty && numRangeElements < 0)) ", ... )" else ")"
-    take(Range.MAX_PRINT).mkString("Range(", ", ", endStr)
+  override def toString = {
+    val preposition = if (isInclusive) "to" else "until"
+    f"$start $preposition $end${
+      if (step == 1) "" else f" by $step"
+    }"
```
sjrd: Is there any reason to use `f` instead of `s`? `f` will pull in a lot of code in Scala.js code bases that do not otherwise use `String.format`.

Also, I'm not convinced the nested interpolator is more readable than:

``` scala
s"$start $preposition $end" + (if (step == 1) "" else s" by $step")
```

som-snytt: Just a finger twitch. At first I wrote `f"$start%d"`, I think. But it's interesting to hear the JS consequences. f-interpolator needs to special case if it's just appending strings. It already detects constant results.

For some reason, I really want the multiline interpolated expression to be readable, but that hasn't actually happened yet.

ps) created an issue for f"". https://issues.scala-lang.org/browse/SI-9784

som-snytt:
Observes @sjrd observations, and adds whether a range is empty or inexact.

Squashed to recognize @stevorobs3 initiative.

som-snytt:
Also fixes the bogus `hashCode` doc. You know your doc mechanism is severely broken if that can happen. Hint: the double-star convention is actually broken.

som-snytt:
Thanks @adriaanm . I just sent the error to my colleague who was dealing with groovy exceptions in gradle last week. I said, "It's easy to make fun of things when they don't work, so that's what I'll do."

adriaanm:
Pleasure! This time we squirreled away too many snapshots. You never know when they might come in handy, I guess.

szeiger:
I like this version. It gives a nice readable output like #5001 while preserving the distinction between `Range` and `NumericRange`. LGTM.

soc:
I'm a bit concerned as this change makes `Range`'s `toString` completely different and inconsistent compared to all other collection classes. I know many people who write a collection in the REPL and expect that the `toString` gives them a reasonable hint about the class involved.

Wouldn't it be possible to wrap the output in `Range()`?

som-snytt:
How about `empty|inexact NumericRange 0 days until 0 days by 1 second of FiniteDuration`. I'm open to better word for `inexact`, to mean the step doesn't cover the range exactly. It can't be both empty and inexact.

soc:
`NumericRange[FiniteDuration](0 days until 0 days by 1 second) <empty>`?

som-snytt:
The idea was just to distinguish what the step is.

```
scala> 0.0 until 2.0 by 0.1
res0: scala.collection.immutable.NumericRange[Double] = NumericRange 0.0 until 2.0 by 0.1 of Double

scala> 0.0 until 2.0 by 1
res1: scala.collection.immutable.NumericRange[Double] = NumericRange 0.0 until 2.0

scala> List(1,2,3)
res2: List[Int] = List(1, 2, 3)
```

This could be nicer:

```
scala> 0.0 until 0.0
res3: Range.Partial[Double,scala.collection.immutable.NumericRange[Double]] = scala.collection.immutable.Range$Partial@498c535d
```

som-snytt:
Actually, I'm not convinced by "step of". It would be better if the `mapRange` instance overrode `toString` to indicate what it is really doing, namely, (precise) stepping using the underlying range.

Also, strings are typically ambiguous:

```
scala> List(BigDecimal("0.1"))
res10: List[scala.math.BigDecimal] = List(0.1)

scala> List(0.1)
res11: List[Double] = List(0.1)
```

som-snytt:
So I didn't even realize that you don't normally get precise stepping goodness. Possibly the comments about it not being complete mean you must opt in:

```
scala> 0.0 to 1.0 by 0.1
res0: scala.collection.immutable.NumericRange[Double] = NumericRange 0.0 to 1.0 by 0.1

scala> .toList
res1: List[Double] = List(0.0, 0.1, 0.2, 0.30000000000000004, 0.4, 0.5, 0.6, 0.7, 0.7999999999999999, 0.8999999999999999, 0.9999999999999999)

scala> Range.Double(0.0, 1.0, 0.1)
res2: scala.collection.immutable.NumericRange[Double] = NumericRange 0.0 until 1.0 by 0.1 (NumericRange 0.0 until 1.0 by 0.1 of BigDecimal)

scala> .toList
res3: List[Double] = List(0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9)
```

Maybe this is an improvement:

```
scala> 0.0 to 1.0
res4: Range.Partial[Double,scala.collection.immutable.NumericRange[Double]] = scala.collection.immutable.Range$Partial@112c2930 Range requires step
```

som-snytt:
Edited the first comment.

szeiger:
Marking as on-hold for now in case you want to make further changes. I think the current state is acceptable but the proposed changes would also be fine with me.

som-snytt:
I believe I'm done. As in done done.

szeiger:
LGTM

PR #5176, opened 2016-05-18
SI-9671, SI-7397 fix null.asInstanceOf[Int] when pt erases to Object

lrytz:
Erasure first replaces null.asInstanceOf[Int] by unbox(null). If the
expected type erases to object, erasure then introduces a box operation,
yielding box(unbox(null)). Note that this value is a box of zero, not
null.

Erasure has an optimization to replace box(unbox(x)) in case x is
of primitive type. 60f1b4b extended this to the case when x is null,
which is incorrect in general. The reason was to prevent creating a
primitive box to be stored in the unused generic field when creating
an instance of a specialized class. A special case ensures that this
optimization is still performed.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Erasure.scala
+++ b/src/compiler/scala/tools/nsc/transform/Erasure.scala
@@ -606,10 +606,8 @@ abstract class Erasure extends AddInterfaces
           // !!! Make pending/run/t5866b.scala work. The fix might be here and/or in unbox1.
           if (isPrimitiveValueType(targ.tpe) || isErasedValueType(targ.tpe)) {
             val noNullCheckNeeded = targ.tpe match {
-              case ErasedValueType(_, underlying) =>
-                isPrimitiveValueClass(underlying.typeSymbol)
-              case _ =>
-                true
+              case ErasedValueType(_, underlying) => isPrimitiveValueType(underlying)
+              case _ => true
             }
             if (noNullCheckNeeded) unbox(qual1, targ.tpe)
             else {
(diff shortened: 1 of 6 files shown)


lrytz:
added another commit for SI-9066

lrytz:
review by @retronym 

there's more that could be fixed.. SI-602, SI-1448, SI-8097, SI-7396, SI-4572

lrytz:
also, i guess the following is as specified, as the cast is not to a primitive type:

```
scala> def nl[T]: T = null.asInstanceOf[T]
nl: [T]=> T

scala> def s(a: Any) = "" + a
s: (a: Any)String

scala> s(nl[Int])
res0: String = null
```

retronym:
LGTM!

soc:
@lrytz Does this change the behavior of SI-602?

lrytz:
@soc no, the test case for SI-602 (checking the current state) is unchanged, see https://github.com/scala/scala/pull/5176/files#diff-1c42a32543c0a81512cbf6bde5766c38L58

soc:
Ok, thanks! While I think the behavior is extremely scary, changing it would probably have subtly broken some code out there ...

som-snytt:
Maybe someone else will also wonder if the Unit case should be the unit value:

```scala
scala 2.13.0-M4> def f[A] = null.asInstanceOf[A]
f: [A]=> A

scala 2.13.0-M4> println(f)
null

scala 2.13.0-M4> println(f: Int)
0

scala 2.13.0-M4> println(f: Unit)
null
```

PR #5141, opened 2016-04-29
Introducing: the fields phase [ci: last-only]

adriaanm:
Part of the plan to simplify the mixin phase by making each phase that introduces new trait members responsible for mixing in the corresponding members into subclasses of these traits. (uncurry runs before fields, so it can expand SAM traits with vals without further work. Lambdalift, on the other hand introduces new fields when local traits capture -- mixins still has to deal with this, as well as lazy vals and early init vals)

The [bytecode diff](https://gist.github.com/adriaanm/3f2de4349209755532c5d127f3322306) is progress:
- correct bridges & signatures for trait setters (fields runs before erasure)
- implement final vals of constant type as getters in trait (no fields/overridden accessors in subclasses)
- other (non-constant) accessors for vals are abstract in traits
- no spurious outer class attributes

Fix scala/scala-dev#118
# TODO
- [x] Needs a benign update for test/files/run/t8549.scala, which was lost in the rebase. All green otherwise!
- [x] SI-9840 method cannot override immutable value:

```
abstract class Node(name: String) {
  var value: Option[Boolean] = None
}

case class Group(name: String, nodes: Node*) extends Node(name) {
  override def value = ???
  override def value_=(b: Option[Boolean]) = ???
}
```
- [x] revert accidental deletion of "// Phase travel necessary. For example, nullary methods (getter of an abstract val) get an"
## community build

latest: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/448/consoleFull:
- [x] failing tests in scala-js
- [x] scala-js has sprouted a new dependency (mixin plugin) -- its test suite passed for me locally, though
- [x] twitter-util exploits a weirdness with private[this] val/vars:

```
trait T {
  private[this] var threads = Set[String]()
  protected def threads(): Array[String] = ???
}
```
- [x] sounds like https://github.com/scala/scala/pull/5294/commits/cbffad9db29702368aed0902a9f5377f4b9d740d

```
[twitter-util] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/twitter-util-3ecf6d03ea63cea0a4fd70085fc0a7dee036c0ca/util-core/src/main/scala/com/twitter/concurrent/Scheduler.scala:229: values cannot be volatile
[twitter-util] [error]   @volatile private[this] var _threads = Set[Thread]() // XXX: unused?
[twitter-util] [error]                               ^
[twitter-util] [error] two errors found
```

PR #5177, opened 2016-05-19
Emit trait method bodies in statics

retronym:
In 2.12.0-M4 / #5003, we changed the trait encoding to do away with trait implementation classes, in favour of leaving trait methods in place as interface default methods.

However, we've since discovered that we can no longer precisely target a particular statically resolved method with `invokespecial` due to impedance mismatch between JVM and Scala semantics.

This pull request emits a pair of methods for each concrete trait method. The default method remains, with the same signature as before. However, it now forwards to a same-named static method (with an explicit self param) that contains the method code.

This encoding allows us to precisely target a particular method with `invokestatic`.

This is binary incompatible, and will require bootstrapping of the modules.

(More details to follow)


retronym:
Module bootstrap progressing well: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/424/console

=> I need to update some tests now.

```
Testsuite: scala.issues.BytecodeTest
Tests run: 10, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 5.357 sec
------------- Standard Output ---------------
Label(0)
LineNumber(4, Label(0))
VarOp(ALOAD, 1)
VarOp(ALOAD, 1)
Field(GETFIELD, a/B, t, I)
Op(ICONST_1)
Op(IADD)
Field(PUTFIELD, a/B, t, I)
Op(RETURN)
Label(9)
------------- ---------------- ---------------

Testcase: invocationReceiversProtected took 2.053 sec
Testcase: invocationReceivers took 0.213 sec
    FAILED
assertion failed: 
Expected: List(VarOp(ALOAD, 0), Invoke(INVOKESPECIAL, T, clone, ()Ljava/lang/Object;, false), Op(ARETURN))
Actual  : List(VarOp(ALOAD, 0), Invoke(INVOKESTATIC, T, clone, (LT;)Ljava/lang/Object;, false), Op(ARETURN))
junit.framework.AssertionFailedError: assertion failed: 
Expected: List(VarOp(ALOAD, 0), Invoke(INVOKESPECIAL, T, clone, ()Ljava/lang/Object;, false), Op(ARETURN))
Actual  : List(VarOp(ALOAD, 0), Invoke(INVOKESTATIC, T, clone, (LT;)Ljava/lang/Object;, false), Op(ARETURN))
    at scala.tools.nsc.backend.jvm.CodeGenTools$.assertSameCode(CodeGenTools.scala:169)
    at scala.tools.nsc.backend.jvm.CodeGenTools$.assertSameCode(CodeGenTools.scala:167)
    at scala.issues.BytecodeTest.invocationReceivers(BytecodeTest.scala:350)

Testcase: traitMethodForwarders took 0.412 sec
    FAILED
assertion failed: 
Expected: List(VarOp(ALOAD, 0), Invoke(INVOKESPECIAL, T3, f, ()I, false), Op(IRETURN))
Actual  : List(VarOp(ALOAD, 0), Invoke(INVOKESTATIC, T3, f, (LT3;)I, false), Op(IRETURN))
junit.framework.AssertionFailedError: assertion failed: 
Expected: List(VarOp(ALOAD, 0), Invoke(INVOKESPECIAL, T3, f, ()I, false), Op(IRETURN))
Actual  : List(VarOp(ALOAD, 0), Invoke(INVOKESTATIC, T3, f, (LT3;)I, false), Op(IRETURN))
    at scala.tools.nsc.backend.jvm.CodeGenTools$.assertSameCode(CodeGenTools.scala:169)
    at scala.tools.nsc.backend.jvm.CodeGenTools$.assertSameCode(CodeGenTools.scala:167)
    at scala.issues.BytecodeTest$forwarderTestUtils$.checkForwarder(BytecodeTest.scala:224)
    at scala.issues.BytecodeTest.traitMethodForwarders(BytecodeTest.scala:285)

Testcase: bytecodeForBranches took 0.643 sec
Testcase: noTraitMethodForwardersForOverloads took 0.087 sec
Testcase: t6288bJumpPosition took 0.328 sec
Testcase: t8731 took 0.123 sec
Testcase: t8926 took 0.078 sec
Testcase: traitMethodForwardersForJavaDefaultMethods took 0.125 sec
    FAILED
assertion failed: 
Expected: List(VarOp(ALOAD, 0), Invoke(INVOKESPECIAL, T2, f, ()I, false), Op(IRETURN))
Actual  : List(VarOp(ALOAD, 0), Invoke(INVOKESTATIC, T2, f, (LT2;)I, false), Op(IRETURN))
junit.framework.AssertionFailedError: assertion failed: 
Expected: List(VarOp(ALOAD, 0), Invoke(INVOKESPECIAL, T2, f, ()I, false), Op(IRETURN))
Actual  : List(VarOp(ALOAD, 0), Invoke(INVOKESTATIC, T2, f, (LT2;)I, false), Op(IRETURN))
    at scala.tools.nsc.backend.jvm.CodeGenTools$.assertSameCode(CodeGenTools.scala:169)
    at scala.tools.nsc.backend.jvm.CodeGenTools$.assertSameCode(CodeGenTools.scala:167)
    at scala.issues.BytecodeTest$forwarderTestUtils$.checkForwarder(BytecodeTest.scala:224)
    at scala.issues.BytecodeTest.traitMethodForwardersForJavaDefaultMethods(BytecodeTest.scala:343)

Testcase: specialInvocationReceivers took 0.056 sec
```

retronym:
Another swing at the bootstrap: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/426/

retronym:
@lrytz  I'm a bit stuck trying to teach the inliner about the new setup. I'm working on `InlinerTest.inlineArrayForeachEquivalent`. Now, the `invokevirtual` is inlined, but only inlined the forwarder (an `invokestatic`), and it doesn't enable elimination of the closure. 

I found the `post` inline request feature of the inliner, which seems to do the right thing. See the last commit for a hacky demo of this: `InlinerTest.inlineArrayForeachEquivalent` works after that change.
- Is that the right place?
- What's the right way to express `isTraitVirtualForwarder` ? Is that the sort of thing we need to add the the `InlineInfo` attribute, or could we figure it out from the bytecode + existing metadata available in the backend?

lrytz:
i'll take a look today

lrytz:
@retronym i think your approach works fine, here's a WIP that identifies trait forwarders invoking static impl meethods well enough (i'd say): https://github.com/lrytz/scala/commit/9f50d64ff127f9b67f902d2b1305916e5516a93d (needs cleaning up)

lrytz:
https://github.com/lrytz/scala/commit/9f50d64ff127f9b67f902d2b1305916e5516a93d is the latest WIP

retronym:
@lrytz thanks for that. I've cleaned it up by:
- making it check that we're calling a Scala trait, not just any interface default method
- applying it to the other heuristic strategies
- refactoring the code a little bit.

I'll continue now with the job of adapting the JUnit tests for the optimizer, and then seeing how partest goes.

src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala
```diff
@@ -123,6 +123,8 @@ object BytecodeUtils {
 
   def isReference(t: Type) = t.getSort == Type.OBJECT || t.getSort == Type.ARRAY
 
+  def isScala(classNode: ClassNode) = classNode.attrs != null && classNode.attrs.asScala.exists(a => a.`type` == BTypes.ScalaAttributeName || a.`type` == BTypes.ScalaSigAttributeName)
```
lrytz: i just was a bit confused why this works, because the scala pickle is stored in the `ScalaSignature` _annotation_ nowadays (no longer in the `ScalaSig` _attribute_), but it turns out we still put the attribute. do you know if we have a test for that? https://github.com/scala/scala/blob/2e40aa5871859b9ab6bdc03d56a81746fa90b1e5/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala#L1119-L1121

retronym: The switch to annotation was made in ef1577a9c5a9f2905e9b89b1508f97127d941bd7. I don't know whether leaving the attribute was intentional or an oversight.

retronym:
@lrytz I found a different approach to fixing trait inlining that seems to work better, I changed the backend to use the inline metadata of the forward method for the static method, too.

This has got all the JUnit tests working again. I'm running a bootstrap build again to run the partest suite, too: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/433/console

lrytz:
I found a problem:

```
➜  sandbox git:(pr5177) ✗ cat Test.scala
trait T {
  def f: Int = 0
  def f(t: T): Int = 1
}

object Test {
    new T { }
  }
}
➜  sandbox git:(pr5177) ✗ qsc Test.scala
➜  sandbox git:(pr5177) ✗ qs Test
java.lang.ClassFormatError: Duplicate method name&signature in class file T
```

lrytz:
@retronym for your fix to the inliner metadata, I think it's a bit of a hack; we should start by adding an entry for the newly-generated static method to the `ClassBType`'s `methodInfos`. I'll take a look.

lrytz:
to expand: what happens in your fix is that the static method is inlined into the forwarder. the only reason this didn't happen before is because the `MethodInlineInfo` is missing. this way, inlining a call of the default method yields the actual implementation, enabling closure elimination.

We did rely on the same mechanics before, when we still had trait impl classes.

lrytz:
@retronym I implemented the change to represent the static methods in the ScalaInlineInfo: https://github.com/retronym/scala/compare/topic/trait-statics...lrytz:pr5177

There's first commit is necessary to fix a cross-talk between tests, the test introduced in the last commit would cause some other tests to fail otherwise.

The second commit is a similar fix, but not related to this PR.

lrytz:
@retronym the implementation of `isTraitMethodRequiringStaticImpl(sym: Symbol)` looks a bit too brittle: https://github.com/retronym/scala/commit/e3535b4ab54e4abdfb192bdc66b48fa79f088c25#diff-7a3aa9d1c9acd44da15dc536d4760e5aR53

The problem is that when building the `InlineInfo`, we don't have the method's `DefDef` at hand, so we cannot check `dd.rhs.isEmpty`. At least we can check that the two are consistent during code gen (https://github.com/retronym/scala/commit/e3535b4ab54e4abdfb192bdc66b48fa79f088c25#diff-492148adc97fa1758c101777df204a20R493), but it's not a beauty..

retronym:
@lrytz I'm going to try to move things back to `mixin`. To start with, I'm trying to add a symbol attachment with the `isTraitMethodRequiringStaticImpl` for the backend to pickup. The next step would be to create the trees for the static and forwarder pair at that point, too.

sjrd:
> The next step would be to create the trees for the static and forwarder pair at that point, too.

Oh. That would be a pity for Scala.js. the JS backend is very happy with the current trees. It doesn't need those static methods.

lrytz:
> I'm going to try to move things back to mixin

@retronym how about just adding `$init` to the static method names?

retronym:
@lrytz Agree that we shouldn't blindly inline the static method into the forwarder; we only want to inline it into a caller of the forwarder.

```
trait T {
  @inline final def foo = 42
}

class C extends T {
  def baz = foo
}
```

Currently:

```
public class C implements T {
  public int baz();
    Code:
       0: bipush        42
       2: ireturn

  public C();
    Code:
       0: aload_0
       1: invokespecial #18                 // Method java/lang/Object."<init>":()V
       4: aload_0
       5: invokestatic  #22                 // Method T.$init$:(LT;)V
       8: return
}
Compiled from "test.scala"
public interface T {
  public static int foo(T);
    Code:
       0: bipush        42
       2: ireturn

  public int foo();
    Code:
       0: bipush        42
       2: ireturn

  public static void $init$(T);
    Code:
       0: return

  public void $init$();
    Code:
       0: aload_0
       1: invokestatic  #19                 // Method $init$:(LT;)V
       4: return
}
```

Hopefully we're not too far away from a solution.

lrytz:
oh, i guess the question is more about whether to emit the method or not. if that's easier to decide in mixin, then an attachment makes sense. for splitting up, i think the current approach works fine, but doing it in mixing would also work for me. we could have a flag to disable it for scala.js.

actually, (one of?) the reasons we are changing this is to allow super calls to target the static method directly. when are those super calls translated, only in the backend?

lrytz:
for the inlining question, if we don't want to duplicate the static body in the forwarder that means we'd have to go with the previous solution of using the `post` request.

sjrd:
> we could have a flag to disable it for scala.js.

No, that's not option if it produces invalid bytecode. We still need valid class files to be emitted in Scala.js runs, for macros.

retronym:
I'm happy to leave it in the backend, but was prepared to move it if demanded by review, as what I'm doing is pretty non-standard.

lrytz:
i'm happy to have it in the backend

retronym:
@lrytz I'm trying to move back towards using `postRequests`. I've managed to avoid inlining into the forwarder (by effectively using a `@noinline` within the forwarding call). I haven't written the test case to show this yet, I tested it manually.

A few things elude me. I had to add `addIfMissing` to load the bytecode for the forwarder during the `inlineRequest` call to get `InlinerSeparateCompilation` tests working, not sure if this is a the right approach.

I'm currently having trouble getting the joint compilation tests working (e.g. `InlinerTest#inlineFromTraits`). I think the problem is related to the ordering of inline requests. I've pushed my WIP in case you have a chance to help out again!

lrytz:
will take a look today!

retronym:
Here's the inline and post inline requests that are being sorted:

image

They are all considered to be leaves, which isn't what we want.

lrytz:
@retronym here's a patch that fixes the ordering of inline requests. JUnit tests pass now. I anticipated the change of implementation method names in the tests. https://github.com/retronym/scala/compare/topic/trait-statics...lrytz:pr5177

retronym:
@lrytz Thanks! I've applied your change, although I have left the static method name unchanged. I'd like to avoid mangling this as it shows up in stack traces. JUnit tests are all passing. I've got a few partest failures now, but they seem to all be a simple matter of updating expectations and fixing a bug in the new implementation restriction.

https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/441/console

I also have changed the impl restriction to run during `Mixin` and to suggest to extend, for example, `Iterable[String]` rather than `java/lang/Iterable`.

I've also removed the virtual `$init` methods.

lrytz:
@retronym if you leave the static method names unchanged, how can we work around the problem described here https://github.com/scala/scala/pull/5177#issuecomment-221500061 ?

retronym:
@lrytz Oh, I'd missed that comment. Good catch. Indeed we'll either need to mangle the static name (everywhere, uniformly), or levy a new implementation restriction (in the same spirit as the existing "methods erase to the same sig" error.)

ijuma:
Good catch even if annoying that we have to go back to having the implementation in static methods due to the somewhat uncommon super calls. I guess this means that there abstract classes still have some advantages, right?

retronym:
Just pushed a commit to add a trailing `$` to the static impl method name. Testing this (through a module boostrap) in [scala-2.12.x-integrate-bootstrap/447](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/447/console).

retronym:
@ijuma My intuition is that the indirection from the default method to the static impl method won't incur any performance penalty. Will need to micro benchmark this though.

Even before this PR, I identifed an advantage to abstract classes.

Compare `jbaseInterface_16` and `baseClass_16` in https://github.com/retronym/scala-jmh-suite/commit/3b4594c8a4be5a33a2d9cd5025ffe9979d336736. HotSpot is likely to get better at this in future releases, some optimizations were buggy and disabled when default methods were used, and are yet to be reenabled.

See also: 
- https://github.com/scala/scala/pull/5085#issuecomment-213334155 
- https://bugs.openjdk.java.net/browse/JDK-8036580

retronym:
Removed the WIP label and running CI on the commits. "Emit trait method bodies in statics" and beyond will fail due to the binary incompatibility with partest. I have triggered a bootstrapping [build for the final commit](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/452/console) that will run the tests with a bootstrapped partest.

src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -1001,8 +1001,13 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
           val parents1 = currentOwner.info.parents map (t => TypeTree(t) setPos tree.pos)
           // mark fields which can be nulled afterward
           lazyValNullables = nullableFields(templ) withDefaultValue Set()
+          val bodyEmptyAccessors = if (!sym.enclClass.isTrait) body else body mapConserve {
+            case dd: DefDef if dd.symbol.isAccessor && !dd.symbol.isLazy =>
+              deriveDefDef(dd)(_ => EmptyTree)
+            case tree => tree
+          }
```
lrytz, line 1017: maybe this won't be necessary after the fields PR is in? would be good to add a comment so we can check that once both changes are in.

adriaanm, line 1017: I'm pretty sure that's fixed in my PR. Maybe we could get that one merged first, or rebase yours on top of it -- it's ready (also, I'm flying to Old Country on Wednesday night). Well, the last commit needs a bit of polish, but I expect the first commits to stay as-is.

src/compiler/scala/tools/nsc/backend/jvm/opt/InlinerHeuristics.scala
```diff
@@ -91,7 +92,27 @@ class InlinerHeuristics[BT <: BTypes](val bTypes: BT) {
...
-      case None => Right(InlineRequest(callsite, Nil, reason))
+      case None =>
+        val callee = callsite.callee.get
+        val postInlineRequest: List[InlineRequest] = callee.calleeDeclarationClass.isInterface match {
+          case Right(true) =>
+            // Treat the pair of trait interface method and static method as one for the purposes of inlining:
+            // if we inline invokeinterface, invoke the invokestatic, too.
+            val calls = callee.callee.instructions.iterator().asScala.filter(BytecodeUtils.isCall).take(2).toList
+            calls match {
+              case List(x: MethodInsnNode) if x.getOpcode == Opcodes.INVOKESTATIC && x.name == (callee.callee.name + "$") =>
```
adriaanm, line 104: Reuse `traitImplMethodName`?

src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -1037,4 +1053,14 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
       finally localTyper = saved
     }
   }
+
+  private def isTraitMethodRequiringStaticImpl(dd: DefDef): Boolean = {
+    val sym = dd.symbol
+    sym != null && sym.owner.isTrait && !sym.isDeferred &&
+      (sym.isLazy || !sym.isAccessor) && !sym.isSuperAccessor && !sym.isModule && // these are not deferred, but defDef.rhs.isEmpty
```
adriaanm: After the fields PR, we'll have SYNTHESIZE_IMPL_IN_SUBCLASS for accessors to denote their implementation is deferred to a subclass, but it's compiler-generated, so we should consider the method as concrete for name resolution etc. Super accessors don't get that flag yet, but maybe we should so that we can streamline this logic further (it also comes up in other spots).

lrytz: i'm about to simplify this, it was originally much simpler (rhs.nonempty), but then i changed it to inspect the symbol only  as we needed the same test in a spot where we don't have the tree. but that's no longer the case, now that we have an attachment for that.

src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -143,6 +143,7 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
       // implementation class, as it's a clone that was made after erasure, and thus it does not
       // know its info at the beginning of erasure anymore.
       val sym = mixinMember cloneSymbol clazz
+      sym.removeAttachment[NeedStaticImpl.type]
```
lrytz: it's not clear to me why we need this line, so i think it deserves a comment

lrytz: is it this? even though we're in preTransform, the mixin member (of a different compilation unit) might already have been transformed by mixin's postTransform and therefore have the attachment. we remove it from the forwarder.

lrytz:
@retronym i did a few cleanups, here's my branch: https://github.com/retronym/scala/compare/topic/trait-statics...lrytz:pr5177?expand=1

src/compiler/scala/tools/nsc/backend/jvm/opt/InlinerHeuristics.scala
lrytz, line 104: that method works on names, we only have a string here - i think it's fine to keep it.

adriaanm, line 104: A method would help keep things in synch and makes it easier to grep  for those not familiar. Overload it on string?

lrytz:
the test failure in the latest commit (`HtmlDocletTest`) is due to the binary incompatibility with modules, i can reproduce it locally:

```
java.lang.NoSuchMethodError: scala.collection.immutable.Traversable.$init$()V
    at scala.xml.NodeSeq.<init>(NodeSeq.scala:45)
    at scala.xml.Node.<init>(Node.scala:36)
```

src/compiler/scala/tools/nsc/backend/jvm/opt/InlinerHeuristics.scala
lrytz, line 104: sounds good

lrytz:
@retronym i just pushed another commit to my branch that fixes the `VerifyError`, it's an interaction between https://github.com/scala/scala/pull/5099 and the change here. see https://github.com/retronym/scala/compare/topic/trait-statics...lrytz:pr5177?expand=1

retronym:
@lrytz @adriaanm Thanks for the detailed comments + code! I'll incorporate this round of review in the PR later today.

retronym:
Latest bootstrap build attempt: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/455/

lrytz:
i think a rebase on 2.12 would be good, as bc868b25a049cd4878b95d28d7ba81aef98a16ed and 6136415d32e1f7690f8654077a0859bcbd21594b are already in there (with different hashes)

retronym:
will do

lrytz:
or even a rebase on top of 65326c85dedd4e18685b0c783d67937d10d68a1d with is part of the fields PR #5141. then the first commit in this PR could probably be skipped.

src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
```diff
@@ -390,8 +390,12 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
...
 
       def load(locSym: Symbol) {
-        val Local(tk, _, idx, _) = slots(locSym)
-        bc.load(idx, tk)
+        slots.get(locSym) match {
+          case Some(Local(tk, _, idx, _)) =>
+            bc.load(idx, tk)
+          case None =>
+            ???
+        }
```
lrytz: revert this?

src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -344,6 +348,11 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
               reporter.error(clazz.pos, "Member %s of mixin %s is missing a concrete super implementation.".format(
                 mixinMember.alias, mixinClass))
             case alias1 =>
+              val temp = alias1
```
lrytz: i guess this local was for debugging purposes?

retronym: - [x] oops, will inline.

src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
retronym: - [x] will do. this was needed to debug a failure in test/files/pos/SI-7638.scala, which I fixed by centralizing the removal of the symbol attachment from the cloned symbol of mixed in methods.

retronym:
> or even a rebase on top of 65326c8 with is part of the fields PR #5141. then the first commit in this PR could probably be skipped.
- [x] I'll probably rebase on 2.12.x to let us merge these PRs in either order. I will address the TODO you added to revert that one commit later on.

adriaanm:
Since my PR has already had a first review pass (thanks, @lrytz!), is almost ready to go (the last commit needs to be discussed -- I could also submit it as a separate PR) and does not depend on this PR, I believe we should merge it first.

lrytz:
bootstrap passed! now we just need to figure out where to publish the starr - one could argue for maven central (instead of some private repo) to ensure the starr stays around and the build remains reproducible in the future.

adriaanm:
Great! Let's get that STARR published (I agree maven is the best location). (EDIT: I also experimented with publishing internally, but we'll go the same route as last time -- via maven)

adriaanm:
(I've pushed a commit to #5141 to give the starr a spin)

lrytz:
I tried to see how the community build performs here, but couldn't get it running.

The problem is that we fix scala-xml and scala-partest to 2.12.0-M4, which is not binary compatible. The result is crashes when using scaladoc:

https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/416/consoleFull

```
[sbt-testng-interface] java.lang.NoSuchMethodError: scala.Product.$init$()V
[sbt-testng-interface]  at scala.util.parsing.json.JSONArray.<init>(Parser.scala:103)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.page.IndexScript.$anonfun$packages$1(IndexScript.scala:54)
[sbt-testng-interface]  at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
[sbt-testng-interface]  at scala.collection.immutable.Map$Map3.foreach(Map.scala:176)
[sbt-testng-interface]  at scala.collection.TraversableLike.map$(TraversableLike.scala:234)
[sbt-testng-interface]  at scala.collection.TraversableLike.map(TraversableLike.scala:227)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.page.IndexScript.<init>(IndexScript.scala:29)
```

I tried a few hacks, including
- comment-out `extra.commands: "set scalaVersion := \"2.12.0-M4\""`, but then it just seems to use what ever scala version is specified in the module's build file (2.12.0-M2).
- manually set `extra.commands: "set scalaVersion := \"2.12.0-dbuildxe6b32a79ee3b9b29b840ae9d4e9a0bc70fddbe92\""`, but that didn't work, that scala version could not be resolved while trying to build the modules

Note that https://github.com/scala/scala/pull/4851 would fix the problem for scaladoc and parser-combinators, but we might still have it for projects using xml or parser-combinators in other ways.

retronym:
Following the "bootstrap plan" we used in #5003
- [ ] Rebased and dropped commits already present in 2.12.x
- [ ] [bootstrap build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/459/console) of the last commit with parameters to publish to a staging repository on Sonatype. 
- [ ] [Community build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/425/console) (for expediency, based on modules built on [2.12.0-213c6d3-nightly](https://scala-ci.typesafe.com/artifactory/scala-release-temp/org/scala-lang/scala-compiler/2.12.0-213c6d3-nightly/), yesterdays version of this PR. Community build config in [retronym/community-builds:topic/trait-statics](https://github.com/scala/community-builds/compare/2.12.x...retronym:topic/trait-statics?expand=1)

lrytz:
bootstrap build on latest rev https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/467/console

adriaanm:
Community build using those bootstrapped modules, and fully bootstrapped scala by dbuild (can't skip locker due to binary incompat): https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/432/console

retronym:
The process of restarring will be a tiny bit easier than last time after #5229 is merged.

adriaanm:
Moving to the rebased PR

PR #5180, opened 2016-05-20
Clean up JUnit tests, make sure they pass without bootstrap [ci: last-only]

lrytz:
A few JUnit tests were failing when run in IntelliJ or sbt because they are compiled with the starr compiler in this case.


lrytz:
review by @retronym 
- for 72a59d9, the change is [here](https://github.com/scala/scala/pull/5180/commits/72a59d932db6f16defd14bd729e0f6ec894c7e1b#diff-4c3cc825efa40f380d97d0e02290447d). the rest is fixing up.
- for 5d5a6aa, the change is [here](https://github.com/scala/scala/pull/5180/commits/5d5a6aa154ab533dca4a0e589a7ae67c3c037d61#diff-4c3cc825efa40f380d97d0e02290447d). there's a new class `Compiler` that holds a `Global` and provides some methods to compile strings. the rest is fixing up.
- d9ce4dc is small, similar to the above
- ba510ab makes methods in BytecodeTesting more systematic, see [here](https://github.com/scala/scala/pull/5180/commits/ba510abcdcf176c06ba93c8e9dc6398015877f5e#diff-4c3cc825efa40f380d97d0e02290447d). the rest is fixing up.
- 0e7964a and 46d523b are small
- 2537027 creates new test packages `scala.lang.annotations`, `scala.lang.primitives` to hold JUnit tests for language features. The test methods in the existing `scala.issues.RunTest` and `scala.issues.BytecodeTest` are distributed into new tests in these packages. also some other tests are moved.
- 64fdae8 turns some tests into run tests to make them pass without bootstrap. I think it's safe / good to revert them to ordinary tests once there's a suitable starr. in a future compiler change would cause a regression in such a test it would go unnoticed locally, but the CI would always catch it as it performs a full bootstrap.
- e26835c renames a test to make sure it runs in sbt.

test/junit/scala/tools/testing/BytecodeTesting.scala
```diff
@@ -2,30 +2,99 @@ package scala.tools.testing
...
+      assert(false, "The compiler issued non-allowed warnings or errors:\n" + msg)
+    }
+  }
+
+  def compile(scalaCode: String, javaCode: List[(String, String)] = Nil, allowMessage: StoreReporter#Info => Boolean = _ => false): List[(String, Array[Byte])] = {
+    val run = newRun
+    run.compileSources(makeSourceFile(scalaCode, "unitTestSource.scala") :: javaCode.map(p => makeSourceFile(p._1, p._2)))
+    checkReport(allowMessage)
+    getGeneratedClassfiles(global.settings.outputDirs.getSingleOutput.get)
+  }
```
retronym, line 52: A future refactoring idea: encapsulate the returned value in a `CompileResult` type, and generalize that to let us look at some/all of trees, ASM, bytecode.

I ended up with a similar datastructure in https://github.com/retronym/scalac-survival-guide/blob/master/src/main/scala/guide/package.scala#L89-L95

retronym:
LGTM, merci vilmal for the reviewers guide!

PR #5181, opened 2016-05-21
SI-9776 Fix type of PriorityQueue.newBuilder and improve performance

chrisokasaki:
review by @Ichoran

Mainly fixes issues from SI-9776:
- Fixes the return type of mutable.PriorityQueue.newBuilder to be Builder[A, PriorityQueue[A]] rather than PriorityQueue[A].
- Improves performance of bulk inserts from O(N log N) to O(N), using the heapify algorithm instead of one-by-one inserts.  This primarily affects the builder, ++=, and reverse.  (These changes indirectly benefit the many, many other methods that use the builder or ++=.)

Also fixes issue from SI-9757:
- Fixes space leak in dequeue.  (However, there is still a similar space leak in clear.)

Non SI improvements:
- Slightly improves performance of clone.
- Improves scaladoc description of reverse, which was already out of date from an earlier commit.

Informal timings of the speedups using scalameter:
- Creating a priority queue of 1 million increasing numbers: speedup from 3098ms to 151ms!
- Creating a priority queue of 1 million random numbers: speedup from 338ms to 133ms.
- Reversing a priority queue of 1 million random numbers: speedup from 590ms to 219ms.

### DIFF
--- a/src/library/scala/collection/mutable/PriorityQueue.scala
+++ b/src/library/scala/collection/mutable/PriorityQueue.scala
@@ -66,7 +66,7 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
     def p_swap(a: Int, b: Int) = super.swap(a, b)
   }
 
-  protected[this] override def newBuilder = new PriorityQueue[A]
+  protected[this] override def newBuilder = PriorityQueue.newBuilder[A]
 
   private val resarr = new ResizableArrayAccess[A]
 
@@ -89,14 +89,15 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
     }
   }
 
-  protected def fixDown(as: Array[AnyRef], m: Int, n: Int): Unit = {
+  protected def fixDown(as: Array[AnyRef], m: Int, n: Int): Boolean = {
+    // returns true if any swaps were done (used in heapify)
     var k: Int = m
     while (n >= 2 * k) {
       var j = 2 * k
       if (j < n && toA(as(j)) < toA(as(j + 1)))
         j += 1
       if (toA(as(k)) >= toA(as(j)))
-        return
+        return k != m
       else {
         val h = as(k)
         as(k) = as(j)
@@ -104,6 +105,7 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
         k = j
       }
     }
+    k != m
   }
 
   /** Inserts a single element into the priority queue.
@@ -117,6 +119,66 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
     fixUp(resarr.p_array, resarr.p_size0)
     resarr.p_size0 += 1
     this
+  }
+
+  override def ++=(xs: TraversableOnce[A]): this.type = {
+    val from = resarr.p_size0
+    for (x <- xs) unsafeAdd(x)
+    heapify(from)
+    this
+  }
+
+  private def unsafeAdd(elem: A): Unit = {
+    // like += but skips fixUp, which breaks the ordering invariant
+    // a series of unsafeAdds MUST be followed by heapify
+    resarr.p_ensureSize(resarr.p_size0 + 1)
+    resarr.p_array(resarr.p_size0) = elem.asInstanceOf[AnyRef]
+    resarr.p_size0 += 1
+  }
+
+  private def heapify(from: Int): Unit = {
+    // elements at indices 1..from-1 were already in heap order before any adds
+    // elements at indices from..n are newly added, their order must be fixed
+    val n = length
+
+    if (from <= 2) {
+      // no pre-existing order to maintain, do the textbook heapify algorithm
+      for (i <- n/2 to 1 by -1) fixDown(resarr.p_array, i, n)
+    }
+    else if (n - from < 4) {
+      // for very small adds, doing the simplest fix is faster
+      for (i <- from to n) fixUp(resarr.p_array, i)
+    }
+    else {
+      var min = from/2 // tracks the minimum element in the queue
+      val queue = scala.collection.mutable.Queue[Int](min)
+
+      // do fixDown on the parents of all the new elements
+      // except the parent of the first new element, which is in the queue
+      // (that parent is treated specially because it might be the root)
+      for (i <- n/2 until min by -1) {
+        if (fixDown(resarr.p_array, i, n)) {
+          // there was a swap, so also need to fixDown i's parent
+          val parent = i/2
+          if (parent < min) { // make sure same parent isn't added twice
+            min = parent
+            queue += parent
+          }
+        }
+      }
+
+      while (queue.nonEmpty) {
+        val i = queue.dequeue()
+        if (fixDown(resarr.p_array, i, n)) {
+          val parent = i/2
+          if (parent < min && parent > 0) {
+            // the "parent > 0" is to avoid adding the parent of the root
+            min = parent
+            queue += parent
+          }
+        }
+      }
+    }
   }
 
   /** Adds all elements provided by a `TraversableOnce` object
@@ -142,9 +204,11 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
   def dequeue(): A =
     if (resarr.p_size0 > 1) {
       resarr.p_size0 = resarr.p_size0 - 1
-      resarr.p_swap(1, resarr.p_size0)
+      val result = resarr.p_array(1)
+      resarr.p_array(1) = resarr.p_array(resarr.p_size0)
+      resarr.p_array(resarr.p_size0) = null // erase reference from array
       fixDown(resarr.p_array, 1, resarr.p_size0 - 1)
-      toA(resarr.p_array(resarr.p_size0))
+      toA(result)
     } else
       throw new NoSuchElementException("no element to remove from heap")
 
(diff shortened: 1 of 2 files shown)


src/library/scala/collection/mutable/PriorityQueue.scala
```diff
@@ -89,21 +89,23 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
...
-        return
+        return k != m
       else {
         val h = as(k)
         as(k) = as(j)
         as(j) = h
         k = j
       }
     }
+    return k != m
```
Ichoran: This should just be `k != m` since it's the last expression.

src/library/scala/collection/mutable/PriorityQueue.scala
```diff
@@ -142,9 +204,11 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
   def dequeue(): A =
     if (resarr.p_size0 > 1) {
       resarr.p_size0 = resarr.p_size0 - 1
-      resarr.p_swap(1, resarr.p_size0)
+      val elementToReturn = resarr.p_array(1)
```
Ichoran: I think `result` is a more common name for temporary storage for a return value.

src/library/scala/collection/mutable/PriorityQueue.scala
```diff
@@ -254,12 +320,24 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
    *
    *  @return  a priority queue with the same elements.
    */
-  override def clone(): PriorityQueue[A] = new PriorityQueue[A] ++= this.iterator
+  override def clone(): PriorityQueue[A] = {
+    val pq = new PriorityQueue[A]
+    for (i <- 1 to this.length) pq.unsafeAdd(resarr(i))
```
Ichoran: This is way less bad than the previous version, but why not just copy the underlying array?  For instance, dynamically expanding the array rather than just setting it to the correct size results in twice the allocation needed and 2.4ish the number of copies (depending on the distribution of the lengths).

src/library/scala/collection/mutable/PriorityQueue.scala
```diff
@@ -254,12 +320,24 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
...
 }
 
 
 object PriorityQueue extends OrderedTraversableFactory[PriorityQueue] {
-  def newBuilder[A](implicit ord: Ordering[A]) = new PriorityQueue[A]
+  def newBuilder[A](implicit ord: Ordering[A]): Builder[A, PriorityQueue[A]] = {
+    new Builder[A, PriorityQueue[A]] {
+      val pq = new PriorityQueue[A]
+      def +=(elem: A): this.type = { pq.unsafeAdd(elem); this }
+      def result(): PriorityQueue[A] = { pq.heapify(1); pq }
```
Ichoran, line 346: Can you decide whether this builder should be reusable or not (see the new ReusableBuilder trait)?  As written, the builder can be used once efficiently, but can still be added to and regenerated a little less efficiently (because heapify is called from 1, not only on the new elements).  Since most builder-use is single shot, this is probably the better way to go (as opposed to maintaining a list of number of elements stored since the last call to result).  In any case, this seems reusable, so maybe you can tag it as such.

src/library/scala/collection/mutable/PriorityQueue.scala
```diff
@@ -186,25 +250,27 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
...
-   *  assuming the original ordering abides its contract.
+  /** Returns the reverse of this priority queue. The new priority queue has
+   *  the same elements as the original, but the opposite ordering.
    *
-   *  Note that the order of the elements will be reversed unless the
-   *  `compare` method returns 0. In this case, such elements
-   *  will be subsequent, but their corresponding subinterval may be inappropriately
-   *  reversed. However, due to the compare-equals contract, they will also be equal.
+   *  For example, the element with the highest priority in `pq` has the lowest
+   *  priority in `pq.reverse`, and vice versa.
```
Ichoran, line 257: I mostly prefer the new wording, but it doesn't say anything about what happens with ties.  Maybe just "Elements with equal priority will come off in arbitrary order (not necessarily the reverse of the original arbitrary order)," unless the new approach does guarantee exactly opposite order.  (I think not, in general?)

src/library/scala/collection/mutable/PriorityQueue.scala
```diff
@@ -254,12 +320,24 @@ sealed class PriorityQueue[A](implicit val ord: Ordering[A])
    *
    *  @return  a priority queue with the same elements.
    */
-  override def clone(): PriorityQueue[A] = new PriorityQueue[A] ++= this.iterator
+  override def clone(): PriorityQueue[A] = {
+    val pq = new PriorityQueue[A]
+    for (i <- 1 to this.length) pq.unsafeAdd(resarr(i))
+    pq
```
Ichoran, line 336: Maybe you want to mention, if you leave it this way, that you don't need to call `heapify` because the heap is already in order!  (Otherwise it's a bit jarring: "wait, didn't you say never to return after unsafeAdd without calling heapify?")

test/files/scalacheck/MutablePriorityQueue.scala
```diff
@@ -0,0 +1,102 @@
...
+  property("reverse then +=") = forAll { (x: E, list: List[E]) =>
+    val pq = PriorityQueue(list : _*).reverse += x
+    checkInvariant(pq)(implicitly[Ordering[E]].reverse) &&
+    pq.dequeueAll == (x +: list).sorted
+  }
+
+  property("clone") = forAll { list: List[E] =>
+    val pq = PriorityQueue(list : _*)
+    val c = pq.clone()
+    pq != c &&
```
Ichoran: Do you mean `(pq ne c)` here?

Ichoran:
Looks good; just a few minor comments.  Thanks!  This should speed PriorityQueue up a lot.

src/library/scala/collection/mutable/PriorityQueue.scala
chrisokasaki, line 346: I think you're suggesting a behavior where you can add 1,2,3 to the builder, call result to get a priority queue with elements 1,2,3, then add 4,5,6 to the builder, and call result again to get a priority queue with elements 1,2,3,4,5,6.  But the priority queues are mutable, and I don't think you would want adding 4 to the builder to retroactively also add 4 to the first priority queue.  Or maybe I'm misunderstanding what you are saying.

Ichoran, line 346: No, you understood me, I just wasn't saying anything sensible.  My mistake!  The builders for immutable collections are often reusable, but you're right, it doesn't make sense for this one to be.

chrisokasaki:
I made all of those changes (except the reusable builder).  The speedup to clone was huge, nearly a factor of 10!  I was also able to use some of those same ideas in reverse, but there the improvement is much more modest because it still needs to heapify.

Ichoran:
Glad that made a big difference!  There's still the `pq != c` vs. `pq ne c` issue in the test, though (my last comment).  You want reference equality (the latter), right?

chrisokasaki:
Sigh.  Yes, I made the change but then accidentally left the test file out of the commit.

test/files/scalacheck/MutablePriorityQueue.scala
```diff
@@ -0,0 +1,102 @@
+import scala.collection.mutable.PriorityQueue
+import org.scalacheck._
+import Prop._
+import Arbitrary._
+
+object PQTest extends Properties("PriorityQueue") {
```
Ichoran: I think this has to be `object Test` to get run properly.  I'm not very familiar with ScalaCheck, but the other tests invariably use just `object Test` as the name.

Ichoran:
I think ScalaCheck isn't running the test because it wants `object Test` (not `PQTest`).  Also, if you could squash all commits (no value in keeping them separate; the history is cleaner with a single commit--though of course you can make your own branch that preserves them), the PR should be set to merge.

chrisokasaki:
Thanks, I couldn't make sense of the error reports.  I'll change the name and squash the commits.

chrisokasaki:
FYI, changing the name (PQTest to Test) and squashing the commits in one go is giving me conflicts, so I'm going try doing that in separate steps, which means that there's going to be at least one meaningless intermediate update before I can get it right.  Just ignore it. :-)

chrisokasaki:
Ok, here's the deal.  This is the first time I've ever used git (I normally use hg).  And like every git newbie, not only have I dorked things up far beyond my ability to fix them, but in my ignorant flailing around trying to fix things, I've also put things into an inconsistent state where these 6 commits CAN'T be squashed (or at least not by me).

The good news is that it's only 2 files.

From my end, the easiest thing to do would be to kill this pull request, and make a new one from scratch.  The main drawback is that it would lose the record of the conversation above.

Unless you have a better suggestion, that's my leaning.

retronym:
@chrisokasaki Here's the recipe to update this pull request with a single commit: 

```
% git reset 2.12.x
% git add -A
% git commit
% git push -f
```

I've submitted a pull request with the squashed version that we could merge instead, if you prefer:
https://github.com/scala/scala/pull/5184

You are still attributed as the author of the squashed commit.

chrisokasaki:
Awesome, that worked!  Thanks!

Ichoran:
LGTM

chrisokasaki:
Thanks for the help @Ichoran

PR #5183, opened 2016-05-22
SI-9789 use quadratic probing in OpenHashMap

performantdata:


retronym:
> creating the possibility that an insertion may loop forever even though a free slot exists.

Could you add a ScalaCheck based test to demonstrate this bug? You could use this as a starting point: https://github.com/scala/scala/blob/2.12.x/test/files/scalacheck/treeset.scala

performantdata:
> You could use this as a starting point: https://github.com/scala/scala/blob/2.12.x/test/files/scalacheck/treeset.scala

I don't see it's relevance, other than as a ScalaCheck example. I'd need to generate objects with the same hash code, and I don't see that here. What is the advantage of (my learning) ScalaCheck, rather than [faking it in JUnit](https://github.com/performantdata/scala/blob/c2e71e398b747aa61a3ca21cbc4adad9060b7995/test/junit/scala/collection/mutable/OpenHashMapTest.scala#L58-L60)? Something about `Arbitrary`?

retronym:
Yes, I wan't sure if you were familiar with the layout of our Scalacheck tests, so it was just intended as an example.

JUnit and Scalacheck are equally capable of testing a collection a test datatype that has a particular hashcode implementation.

A JUnit test for the minimal case that shows the bug would indeed be the right place to start.

I assumed from the content of this PR that you hadn't identified that yet, and that Scalacheck would be a way to find the problem.

A Scalacheck test of `OpenHashMap` would have have the added benefit to convince us that we aren't introducing a regression, or that it doesn't have other bugs. We could take another `Map` implementation as a reference, `java.util.Map`, and check that during a randomly generated sequence of operations, the datastructures are equal.

But, before you do down that route, note that we have another testing framework with a similar goal for collections that @Ichoran wrote: https://github.com/scala/scala-collections-laws . We run that manually, it isn't currently part of our patch validation.

Perhaps @Ichoran will be able to update that (or guide you how to update it) to demonstrate the bug in `OpenHashMap`. That might be a better use of time than a specific Scalacheck test for `OpenHashMap`, because the new test case would be then run on all map instances.

https://github.com/scala/scala-collections-laws/blob/master/laws/src/main/resources/single-line.tests#L375-L396
https://github.com/scala/scala-collections-laws/blob/master/laws/src/main/resources/replacements.tests#L732-L812 (Add variants that use a key type that is pathologically bad, and one that collides often.)

performantdata:
> We could take another Map implementation as a reference [...] and check the for a randomly generated sequence of operations, the datastructures are equal.

There's already [`SetMapConsistencyTest`](https://github.com/scala/scala/blob/2.12.x/test/junit/scala/collection/SetMapConsistencyTest.scala).

retronym:
Oh cool, didn't know about that one. If you can update that to demonstrate the bug, job done!

retronym:
Invoking Greenspun: Any sufficiently large test suite has two ad hoc, informally-specified, bug-ridden, slow implementations of half of Scalacheck. 😄

performantdata:
I don't see a good, safe way to test the bug's presence, since it would cause an infinite loop. Is it sufficient to use JUnit's `@Test(timeout = ...)`? That may leave the thread running until JUnit shuts down. Of course, we should never see the bug once this commit is in.

Ichoran:
@performantdata  - You test equality, so you can have a mutable class that starts throwing exceptions after a certain number of tests have been reached.

retronym:
@performantdata Please include the test with this commit. It is enough to mention here that you test drove things, and the test case hung before the patch to the implementation.

performantdata:
I'm no longer calling this a bug, in practice. See the [JIRA ticket](https://issues.scala-lang.org/browse/SI-9789). Need to improve the benchmark instead.

retronym:
Based on http://www.drmaciver.com/2008/08/new-collections-in-scala-272/, I think the original algorithm comes from Python.

There are some detailed notes about this in : https://github.com/python/cpython/blob/92ba6865557482f201f3f5db6f41ccef0a3427ee/Objects/dictobject.c#L88-L179

Taking a step back: how widely used is `OpenHashMap`? I hadn't heard of it until you started fixing it. Is it a worthy member of the standard library, or something that might be a deprecation candidate?

performantdata:
> how widely used is `OpenHashMap`?

How would anyone know? I think it'd be valuable to have an open addressing hash map that fits into the collections framework, unlike Debox, which seems to be the most popular solution.

performantdata:
> I think the original algorithm comes from Python.

[The justification given there](http://svn.python.org/projects/python/trunk/Objects/dictobject.c), though, is that the probe sequence is like a random sequence, and the initial hash value needn't be uniform. That's backwards from a normal implementation. One normally wants to randomize the slot at which the sequence starts (i.e., the hash value), then probe in a nearby region to take advantage of memory caching. The former utilizes the more complex computation, the latter is just an addition (not a multiplication).

retronym:
> How would anyone know

I believe @Ichoran has a dataset to give one answer to that question: he trawled all the JARs in Maven that depend on the standard library to see what classes and methods are used. Obviously this doesn't inform us about closed source usage.

A GitHub search is a little tricky because of a same-named collection in Spark: https://github.com/search?l=scala&p=6&q=OpenHashMap&type=Code&utf8=%E2%9C%93

Ichoran:
It was only lightly used (I don't remember exactly how much, but far less than regular HashMap), probably because of problems with it.  `AnyRefMap` and `LongMap` also use an open scheme, but without many of the problems, and with better performance because they don't have to wade through all the alternative forms of equals that an `Any` might have to.

I'm not convinced `OpenHashMap` should be deprecated, but if it were I don't think it'd be a huge loss.  I'm also not convinced it's worth spending a great deal of time to improve it either.

performantdata:
I uploaded the performance results to the [JIRA ticket](https://issues.scala-lang.org/browse/SI-9789). I needed to write a test for `AnyRef` keys, since an `Int`'s hash is simply the `Int` itself, so keys were being inserted mostly without collisions. `AnyRef`'s hash is a pseuodrandom number. You can see up to about a 15% improvement in both `get` and `put` for `AnyRef`s.

performantdata:
review by @Ichoran? The commit mostly consists of changes to the benchmarks, including the ones you suggested in #5061.

Ichoran:
LGTM

PR #5185, opened 2016-05-24
Fix Scaladoc link syntax on java.util.Formatter references

janekdb:


janekdb:
This fixes the format of the links. It does not tackle the configuration of `-doc-external-doc` to actually produce links to https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html

From what I can see `-doc-external-doc` requires a review to make it practical to use. It appears to select the base URL for the external documentation based on the name of the jar the classes being linked to is loaded from.

For example from the tests,

```
 "-doc-external-doc build/pack/bin/../lib/scala-library.jar#http://www.scala-lang.org/api/current/"
```

Certainly something to backlog for a separate PR.

VladUreche:
LGTM

VladUreche:
and yes, please do have a look at `-doc-external-doc` and see if it can be made more practical!

PR #5189, opened 2016-05-25
Rename -Yopt to -opt, -Yopt-warnings to -opt-warnings

lrytz:
Keep -Yopt-inline-heuristics and -Yopt-trace unchanged

### DIFF
--- a/build-ant-macros.xml
+++ b/build-ant-macros.xml
@@ -6,7 +6,7 @@
     <attribute name="name"/>
     <sequential>
       <antcall target="@{name}">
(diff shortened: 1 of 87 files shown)


retronym:
LGTM

gourlaysama:
It's just a thought, but I think this change is important enough for the "release-note" label. WDYT?

And maybe also mention that `-optimize` is now a deprecated alias for `-opt:l:classpath`.

lrytz:
sounds good, we'll talk about the optimizer in the release notes for sure

PR #5152, opened 2016-05-10
Eliminate major sources of daily noise in SBT build.

retronym:
Eliminate major sources of daily noise in SBT build.
- Intercept incorrect "binary conflict" warning issued by SBT.
  Fixes https://github.com/scala/scala-dev/issues/100
- Bump to a new version of pantsbuild/jarjar to fix an
  incompatibility with Java 8 parameter names in class
  files, which we run into on the 2.12.x branch. See:
  https://github.com/pantsbuild/jarjar/pull/19
- Disable info level logging for dependency resolve/download.

Review by @szeiger or @lrytz

PR #5190, opened 2016-05-25
Use sbt for PR validation [ci: last-only]

szeiger:
Picking up the work on #4933


szeiger:
Ah, we still need `pull-binary-libs` call for partest to fetch some JARs. The larger jsoup JAR could probably be resolved from Maven Central and the bunch of small JARs added directly to the git repo.

But this doesn't explain the test failure. The `pull-binary-libs` call is still there and you can clearly see from the log that it resolves the files, but none of them end up on the scalac classpath for partest. I was unable to reproduce this issue locally.

retronym:
@szeiger Related discussion about pull-binary-libs: https://issues.scala-lang.org/browse/SI-8379

retronym:
>  bunch of small JARs added directly to the git repo.

We could also rewrite the tests to generate the necessary bytecode programmatically. We can use ASMifier to reify the ASM code that would create a given class, and then whittle that down the miminum needed to provoke the original bug.

Of course, we might just be able to include Java source files and use the separate compilation test case feature of partest `JavaSource_1.java` / `ScalaClient_2.scala`. Some of the JARs might exists because we couldn't bank on a recent enough Java compiler being available, perhaps when we supported, but did not require, Java 1.5.

szeiger:
It looks like we have our first successful PR validation build on sbt. Remaining issues:
- [x] Run scaladoc tests from sbt
- [x] Port OSGi tests to sbt
- [x] Put the build artifacts in the same location as the ant build or change the Jenkins configuration:

```
Archiving artifacts
```

The following are non-issues (at least for this PR) IMHO:
- Respect `scalac.args.optimise` - The PR validation build is always with optimization. If we need other builds without, they get different setup commands (in `project/ScriptCommands.scala`) where optimization can be left off.
- Do not require `pull-binary-libs` for PR validation (or anything else) - We should do this but it's not a requirement for switching from ant to sbt, so not in scope for this PR.

ghost:
re scaladoc. We have issues generating scaladoc from sbt in cats. I know the issue/solution but have not PR'd anything- ref chat https://gitter.im/sbt/sbt?at=574b4bde10f0fed86f49499c

In case you hit any issues before I PR anything, please shout.

szeiger:
@InTheNow Scaladoc generation has been working in the Scala build for months. The remaining issue here is about running tests on the scaladoc tool (using partest from sbt)

szeiger:
Marking "Put the build artifacts in the same location as the ant build or change the Jenkins configuration" as solved. Some build artifacts were produced by the ant build and then archived by Jenkins because they contain additional details about build errors. The sbt build logs everything to the console instead, so these files should not be necessary anymore. The warning can be safely ignored and the Jenkins configuration changed after this PR has been merged.

szeiger:
Ready for review.

build.sbt
```diff
@@ -326,6 +321,11 @@ def regexFileFilter(s: String): FileFilter = new FileFilter {
   def accept(f: File) = pat.matcher(f.getAbsolutePath.replace('\\', '/')).matches()
 }
 
+// This project provides the STARR scalaInstance for bootstrapping
+lazy val bootstrap = (project in file("target/bootstrap")).settings(
```
adriaanm, line 335: neat!

adriaanm:
I would really like to reduce the additional noise in the output. Maybe suppress [info]-level output? I know it's annoying to fix, but if we postpone it I'm afraid we'll just end up living with it.

retronym:
I just tried to use the native SBT import in IntelliJ to see if the changes in this PR had been enough to get it working. (It wasn't working before because of our custom `scalaInstance`.)

 It does successfully complete, but ends up with the wrong directory mappings:

image

Not a blocker for this PR, just a pity.

retronym:
> It does successfully complete, but ends up with the wrong directory mappings:

Oh, that's a side effect of a change I made previously:

```
  // When we fork subprocesses, use the base directory as the working directory.
  // This enables `sbt> partest test/files/run/t1.scala` or `sbt> scalac sandbox/test.scala`
  baseDirectory in Compile := (baseDirectory in ThisBuild).value,
  baseDirectory in Test := (baseDirectory in ThisBuild).value,
```

IntelliJ [looks up `compile:baseDirectory`](https://github.com/JetBrains/sbt-structure/blob/a8075f0ed59b036a5e55782a30bffabad5397766/extractor/src/main/scala/org/jetbrains/sbt/extractors/ProjectExtractor.scala#L112) rather than `baseDirectory` to use as the content root of the module.

szeiger:
@adriaanm What noise are you referring to specifically? The OSGi tests are noisy but I didn't bother trying to change that (not sure that I could - I only managed to turn some of that noise off in Slick's OSGi tests) because you don't normally run them manually and the console output replaces the JUnit XML logs from the ant build.

adriaanm:
Most lines that start with `[info]`. For example, the ones about `resolving ...`. The JUnit tests also output more than necessary.

build.sbt
lrytz, line 335: i wonder about the various ways to create `File`s that we see in `build.sbt`:
- `file("target/...")`
- `file(".") / "target" / ...`
- `(baseDirectory in ThisBuild).value / "target" / ...`
- `buildDirectory.value / "quick" / ...`

does it matter, should we unify / clean up a bit?

build.sbt
```diff
@@ -616,18 +667,17 @@ lazy val test = project
...
-      new sbt.TestDefinition(
-        "partest",
-        // marker fingerprint since there are no test classes
-        // to be discovered by sbt:
-        new sbt.testing.AnnotatedFingerprint {
-          def isModule = true
-          def annotationName = "partest"
-        }, true, Array())
-     )
+    testOptions in IntegrationTest += Tests.Argument("-Dpartest.scalac_opts=" + (scalacOptions in Compile).value.mkString(" ")),
```
lrytz, line 692: it looks like space separation might not work if there are multiple scalac options

```
-Dpartest.scalac_opts=-optimize -deprecation
```

maybe surround the options by single quotes? just guessing..

build.sbt
```diff
@@ -701,16 +751,18 @@ lazy val scalaDist = Project("scala-dist", file(".") / "target" / "scala-dist-di
...
   .settings(generateBuildCharacterFileSettings: _*)
   .settings(
     publish := {},
     publishLocal := {},
     commands ++= ScriptCommands.all,
-    Quiet.silenceIvyUpdateInfoLogging
-)
+    Quiet.silenceIvyUpdateInfoLogging,
+    antStyle := false,
+    incOptions := incOptions.value.withNameHashing(!antStyle.value).withAntStyle(antStyle.value)
```
lrytz, line 791: I used to have a separate `local.sbt` (which I keep in my global `~/.gitignore`) with the following line:

```
(incOptions in ThisBuild) := (incOptions in ThisBuild).value.withNameHashing(false).withAntStyle(true)
```

This gives an error now:

```
> compile
...
[error] (library/*:incOptions) java.lang.AssertionError: assertion failed: Name hashing and Ant-style cannot be enabled at the same time.
...
```

I tried putting `antStyle := true` in my `local.sbt`, but that doesn't work (`local.sbt:1: error: not found: value antStyle`). What is the correct way to enable it? I wouldn't like to type `set antStyle := true` each time I start sbt.

build.sbt
szeiger, line 335: We should be able to get rid of many of those when we move to a more standard sbt layout after ant is gone. (The first 3 all do the same, the last one is different)

build.sbt
szeiger, line 692: No, there's no parsing of quoted arguments happening anywhere. This is a single `Argument` for partest which does a `.split(' ')` on the RHS of the `-D` def to get the scalac args. There is currently no way use spaces inside of scalac args but that's a limitation of partest (and is also present when launching it from ant or the command line).

lrytz: if you tell me we're fine for now then that's fine with me :)

lrytz: oh, wrong spot for this comment.

build.sbt
lrytz, line 335: here: if you tell me we're fine for now then that's fine with me :)

build.sbt
lrytz, line 692: OK, thanks!

build.sbt
dwijnand, line 791: One way is by re-defining `antStyle` in `local.sbt`, sbt will merge them if they have the same type.

build.sbt
```diff
@@ -616,18 +667,17 @@ lazy val test = project
     fork in IntegrationTest := true,
     javaOptions in IntegrationTest += "-Xmx1G",
     testFrameworks += new TestFramework("scala.tools.partest.sbt.Framework"),
-    testOptions in IntegrationTest += Tests.Setup( () => root.base.getAbsolutePath + "/pull-binary-libs.sh" ! ),
```
lrytz, line 622: so when I run

```
sbt
> partest test/files/pos/cycle-jsoup.scala
```

on a fresh checkout (or after `git clean -dfx`), I get:

```
...
[info] Packaging /Users/luc/scala/scala4/build/pack/lib/scala-partest-javaagent.jar ...
[info] Done packaging.
...
Resolving [346d3dff4088839d6b4d163efa2892124039d216/test/files/lib/jsoup-1.3.1.jar]
...
Partest version:
Compiler under test: $baseDir/compiler
Scala version is:    Scala compiler version 2.12.0-20160606-115156-79e7334 -- Copyright 2002-2016, LAMP/EPFL
Scalac options are:
Compilation Path:    /Users/luc/scala/scala4/target/test/it-classes:$baseDir/test:$baseDir/compiler:$baseDir/library:$baseDir/reflect:$baseDir/interactive:$baseDir/repl-jline-embedded:$baseDir/repl-jline:$baseDir/repl:$baseDir/scalap:$baseDir/partest-extras:/Users/luc/scala/scala4/build/pack/lib/scala-partest-javaagent.jar:$baseDir/scaladoc:/Users/luc/scala/scala4/test/lib:/Users/luc/.ivy2/cache/org.apache.ant/ant/jars/ant-1.9.4.jar:/Users/luc/.ivy2/cache/org.apache.ant/ant-launcher/jars/ant-launcher-1.9.4.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-asm/bundles/scala-asm-5.0.4-scala-3.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-xml_2.12.0-M4/bundles/scala-xml_2.12.0-M4-1.0.5.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-parser-combinators_2.12.0-M4/bundles/scala-parser-combinators_2.12.0-M4-1.0.4.jar:/Users/luc/.ivy2/cache/jline/jline/jars/jline-2.14.1.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-partest_2.12.0-M4/jars/scala-partest_2.12.0-M4-1.0.14.jar:/Users/luc/.ivy2/cache/com.googlecode.java-diff-utils/diffutils/jars/diffutils-1.3.0.jar:/Users/luc/.ivy2/cache/org.scala-sbt/test-interface/jars/test-interface-1.0.jar:/Users/luc/.ivy2/cache/org.scalacheck/scalacheck_2.12.0-M4/jars/scalacheck_2.12.0-M4-1.11.6.jar:/Users/luc/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-agent-0.13.11.jar:/Users/luc/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-interface-1.0.jar
Java binaries in:    /Library/Java/JavaVirtualMachines/jdk1.8.0_77.jdk/Contents/Home/jre/bin
Java runtime is:     Java HotSpot(TM) 64-Bit Server VM (build 25.77-b03, mixed mode)
Java options are:    -Xmx1024M -Xms64M -XX:MaxPermSize=128M
baseDir:             /Users/luc/scala/scala4/build/quick/classes
sourceDir:           /Users/luc/scala/scala4/test/files

Selected 1 tests drawn from specified tests

# starting 1 test in pos
!! 1 - pos/cycle-jsoup.scala                     [compilation failed]
```

with

```
cycle-jsoup.scala:3: error: object jsoup is not a member of package org
    org.jsoup.Jsoup.parse(null: java.net.URL, 3000)
        ^
one error found
```

If I then run the same command `partest test/files/pos/cycle-jsoup.scala` again in the same sbt shell process (no restart), the test passes:

```
> partest test/files/pos/cycle-jsoup.scala --show-diff
[info] Packaging /Users/luc/scala/scala4/build/pack/lib/scala-partest-javaagent.jar ...
[info] Done packaging.
Partest version:
Compiler under test: $baseDir/compiler
Scala version is:    Scala compiler version 2.12.0-20160606-115156-79e7334 -- Copyright 2002-2016, LAMP/EPFL
Scalac options are:
Compilation Path:    /Users/luc/scala/scala4/target/test/it-classes:$baseDir/test:$baseDir/compiler:$baseDir/library:$baseDir/reflect:$baseDir/interactive:$baseDir/repl-jline-embedded:$baseDir/repl-jline:$baseDir/repl:$baseDir/scalap:$baseDir/partest-extras:/Users/luc/scala/scala4/build/pack/lib/scala-partest-javaagent.jar:$baseDir/scaladoc:$sourceDir/lib/annotations.jar:$sourceDir/lib/enums.jar:$sourceDir/lib/genericNest.jar:$sourceDir/lib/jsoup-1.3.1.jar:$sourceDir/lib/macro210.jar:$sourceDir/lib/methvsfield.jar:$sourceDir/lib/nest.jar:/Users/luc/scala/scala4/test/lib:/Users/luc/.ivy2/cache/org.apache.ant/ant/jars/ant-1.9.4.jar:/Users/luc/.ivy2/cache/org.apache.ant/ant-launcher/jars/ant-launcher-1.9.4.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-asm/bundles/scala-asm-5.0.4-scala-3.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-xml_2.12.0-M4/bundles/scala-xml_2.12.0-M4-1.0.5.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-parser-combinators_2.12.0-M4/bundles/scala-parser-combinators_2.12.0-M4-1.0.4.jar:/Users/luc/.ivy2/cache/jline/jline/jars/jline-2.14.1.jar:/Users/luc/.ivy2/cache/org.scala-lang.modules/scala-partest_2.12.0-M4/jars/scala-partest_2.12.0-M4-1.0.14.jar:/Users/luc/.ivy2/cache/com.googlecode.java-diff-utils/diffutils/jars/diffutils-1.3.0.jar:/Users/luc/.ivy2/cache/org.scala-sbt/test-interface/jars/test-interface-1.0.jar:/Users/luc/.ivy2/cache/org.scalacheck/scalacheck_2.12.0-M4/jars/scalacheck_2.12.0-M4-1.11.6.jar:/Users/luc/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-agent-0.13.11.jar:/Users/luc/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-interface-1.0.jar
Java binaries in:    /Library/Java/JavaVirtualMachines/jdk1.8.0_77.jdk/Contents/Home/jre/bin
Java runtime is:     Java HotSpot(TM) 64-Bit Server VM (build 25.77-b03, mixed mode)
Java options are:    -Xmx1024M -Xms64M -XX:MaxPermSize=128M
baseDir:             /Users/luc/scala/scala4/build/quick/classes
sourceDir:           /Users/luc/scala/scala4/test/files

Selected 1 tests drawn from specified tests

# starting 1 test in pos
ok 1 - pos/cycle-jsoup.scala
```

note that `jsoup-1.3.1.jar` is only on the "Compilation Path" in the second invocation.

project/MiMa.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  lazy val mimaOldVersion =
+    settingKey[String]("Scala version number to run MiMa against")
+
+  lazy val mima =
+    taskKey[Unit]("run Migration Manager to detect binary incompatibilities")
+
+  lazy val settings =
+    Seq(
+      mimaOldVersion := VersionUtil.versionProps("scala.binary.version") + ".0",
```
lrytz: not sure if using `scala.binary.version` is correct here, see its doc comment and its value in the 2.12.x branch: https://github.com/scala/scala/blob/2.12.x/versions.properties#L16-L22

the code here will only work after `.0` is released. also note that we'd like to enable mima once M5 is out and set M5 (or maybe later RC1) as the base version. in ant we used `2.11.0` directly, no?

related, what is the scenario to enable / disable mima during the milestone cycle? just don't invoke it from the CI bash scripts?

build.sbt
lrytz: oh, my mistake, i just did that on current 2.12.x, let me try again what's the state after this PR!

lrytz: wrong spot again, sorry..!!!!

build.sbt
lrytz, line 622: here: oh, my mistake, i just did that on current 2.12.x, let me try again what's the state after this PR!

lrytz, line 622: yeah, so as expected: after this PR, `partest test/files/pos/cycle-jsoup.scala` fails until `pull-binary-libs.sh` is run manually. i think we should find a way to fix this, it's too confusing for newcomers.

test/scaladoc/run/t8557.scala
```diff
@@ -15,10 +17,22 @@ object Test extends ScaladocModelTest {
...
+  def scalaURL = "http://www.scala-lang.org/api/current/"
+
   // a non-canonical path to scala-library.jar should still work
-  // this is a bit fragile (depends on the current directory being the root of the repo ;
-  // ant & partest seem to do that properly)
-  def scaladocSettings = "-doc-external-doc build/pack/bin/../lib/scala-library.jar#http://www.scala-lang.org/api/current/"
+  override def scaladocSettings = {
+    val samplePath = getClass.getClassLoader.getResource("scala/Function1.class").getPath.replace('\\', '/')
+    val scalaLibPath = if(samplePath.contains("!")) { // in scala-library.jar
+    val scalaLibUri = samplePath.split("!")(0)
```
lrytz: nit: indent this line

build.sbt
lrytz, line 791: that works, thanks!

build.sbt
szeiger, line 622: This is expected. You still need to run `pull-binary-libs.sh` initially. Note that even with the now removed line it used to fail on the first run because doing it in `Test.Setup` is too late (the classpath has already been built before the JARs are downloaded). Only subsequent runs would succeed.

But I don't really want to call it from the sbt build at all because that would require a Unix environment. Since most of the binaries are already resolved through Ivy, I hope that we can get of it entirely. JSoup is on Maven Central, the remaining binary files are about 10 kB in total. It would be nice to generate them on demand but the size is so low that we could just check them into git. Another option is to download them with Scala code in the sbt build.

lrytz, line 622: OK, agree. could we maybe add an assertion somewhere in the sbt startup, require that `test/files/lib/annotations.jar` exists?

project/MiMa.scala
szeiger: I assumed that MiMa would only be enabled after a .0 release. We can add a separate base version setting to the sbt build. It's also hardcoded in build.xml. The ant build reacts to a `test.bc.skip` setting but I don't know if it's actually used. I would change it in the CI script or build.sbt as required (especially if the base version is also set there; these two settings belong together)

lrytz: agreed

test/scaladoc/run/t8557.scala
szeiger: Another victim of IntelliJ's broken formatter (doesn't indent the line after a `{` when followed by a line comment)

build.sbt
szeiger, line 791: I've asked the sbt guys for help with this because I couldn't make it work. If there's no cleaner solution, I could move the definition into an object (which can be referenced from all `.sbt` files) and add an alias in `build.sbt` to make it available with an unqualified name on the command line.

dwijnand, line 791: An alternative is to define a local AutoPlugin just for the purpose of defining auto-imported keys: https://github.com/scala-js/scala-js/blob/af7c61d43bc0f9395fa6aec2ca2aecfcb0883636/project/Build.scala#L42-L50

build.sbt
szeiger, line 622: We could check for it in `Test.Setup` and log a warning if the JARs are missing. I don't think the whole build should fail. Only a handful of partest test cases still need `pull-binary-libs`. Everything else is resolved through Ivy in the sbt build.

lrytz:
```
Caused by: sbt.ForkMain$ForkError: java.lang.RuntimeException: 'instrumented.jar' not found in '/home/jenkins/workspace/scala-2.11.x-validate-test/test/files/speclib'.
```

retronym:
> Most lines that start with [info]. For example, the ones about resolving .... The JUnit tests also output more than necessary.

@adriaanm I turned that off in https://github.com/scala/scala/pull/5152/files#diff-991b9473f196f6364820f0c3652eb3faR32 . Perhaps you are on a branch doesn't have that change it its hiatory yet?

retronym:
> Most lines that start with [info]. For example, the ones about resolving .... The JUnit tests also output more than necessary.

@adriaanm I turned that off in https://github.com/scala/scala/pull/5152/files#diff-991b9473f196f6364820f0c3652eb3faR32 . Perhaps you are on a branch doesn't have that change it its hiatory yet?

szeiger:
_sigh_ The reference to speclib/intrumented.jar is hardcoded in partest itself

szeiger:
@retronym I've just pushed another commit with a workaround for your scalacheck bootstrapping problem (provided that it passes CI and doesn't cause new problems).

adriaanm:
> Perhaps you are on a branch doesn't have that change it its hiatory yet?

I was looking at the output of this PR, so I guess this that'll take care of itself when we merge.

szeiger:
No, it's in the history and I see the code. It may be scoped too narrowly to catch everything. I'll investigate. (BTW, noise has already been considerably reduced by running everything from a single sbt invocation; now you only get the resolver spam once at the beginning)

szeiger:
Everything should be fixed now.

adriaanm:
Fix all the things!! :party:

szeiger:
Squashed, rebased and switched to partest 1.0.15

project/MiMa.scala
```diff
@@ -0,0 +1,94 @@
+// It would be nice to use sbt-mima-plugin here, but the plugin is missing
+// at least two features we need:
+// * ability to run MiMa twice, swapping `curr` and `prev`, to detect
+//   both forwards and backwards incompatibilities
```
dwijnand: I think this isn't true as of 0.1.9: https://github.com/typesafehub/migration-manager/commit/2844ffa48b6d2255aa64bd687703aec21dadd55e

project/MiMa.scala
```diff
@@ -0,0 +1,94 @@
...
+// * ability to pass a filter file
```
dwijnand: I think this is tracked in https://github.com/typesafehub/migration-manager/issues/102. Perhaps it would be useful to add that as a link in the comment.

szeiger:
Offending jarlister tests disabled, updated to a new partest 1.0.16, MiMa comments added.

lrytz:
LGTM! 🎈

PR #5191, opened 2016-05-25
SI-9382 Privatize enhanced x in Tuple2Zipped.Ops

som-snytt:
Probably there should be an Abide rule to avoid leaking
the "underlying value" of a value class. The spec or SIP defines
"underlying type" but doesn't mention the underlying value.

The argument for concealing the member is that it is redundant
and makes autocompletion results harder to read. Also, possibly
an additional implicit might want to add a member so-named.


adriaanm:
LGTM

som-snytt:
@adriaanm Maybe you could sponsor me for a merit badge.

I noticed `(xs, ys).zipped.unzip` is non-trivial, so `(xs, ys).zipped.colls` is useful, even if the name is a bit casual.

Edit: alternatively, you could just get really annoyed.

lrytz:
> the name is a bit casual

so is the type

```
scala> (List(1,2,3), List(4,5,6)).zipped.colls
res2: (scala.collection.TraversableLike[Int,List[Int]], scala.collection.IterableLike[Int,List[Int]]) = (List(1, 2, 3),List(4, 5, 6))
```

PR #5192, opened 2016-05-25
Fully qualify types in REPL generated code

dwijnand:
This is so the Scala REPL can work with `-Yno-imports` and `-Yno-predef`.

The partest run case is a bit of a shotgun approach to seeing if I can get it to fall over in any other way, stealing code from existing repl test cases.  If there's a better approach, please let me know.

I'm happy to iterate.

Review by @retronym.

### DIFF
--- a/test/files/run/repl-no-imports-no-predef-classbased.scala
+++ b/test/files/run/repl-no-imports-no-predef-classbased.scala
@@ -1,0 +1,19 @@
+object Test extends scala.tools.partest.ReplTest {
+
+  override def transformSettings(settings: scala.tools.nsc.Settings) = {
+    settings.noimports.value = true
+    settings.nopredef.value = true
+    settings.Yreplclassbased.value = true
+    settings
+  }
+
+  def code = """
+case class K(s: java.lang.String)
+class C { implicit val k: K = K("OK?"); override def toString = "C(" + k.toString + ")" }
+val c = new C
+import c.k
+scala.Predef.implicitly[K]
+val k = 42
+k                 // was K(OK?)
+"""
+}
--- a/test/files/run/repl-no-imports-no-predef.scala
+++ b/test/files/run/repl-no-imports-no-predef.scala
@@ -1,0 +1,108 @@
+object Test extends scala.tools.partest.ReplTest {
+
+  override def transformSettings(settings: scala.tools.nsc.Settings) = {
+    settings.noimports.value = true
+    settings.nopredef.value = true
+    settings
+  }
+
+  // replace indylambda function names by <function0>
+  override def normalize(s: String) = """\$\$Lambda.*""".r.replaceAllIn(s, "<function0>")
+
+  def code = """
+1
+1.0
+()
+"abc"
+(1, 2)
+
+{ import scala.Predef.ArrowAssoc; 1 -> 2 }
+{ import scala.Predef.ArrowAssoc; 1 → 2 }
+1 -> 2
+1 → 2
+
+val answer = 42
+{ import scala.StringContext; s"answer: $answer" }
+s"answer: $answer"
+
+"abc" + true
+
+{ import scala.Predef.any2stringadd; true + "abc" }
+true + "abc"
+
+var x = 10
+var y = 11
+x = 12
+y = 13
+
+2 ; 3
+{ 2 ; 3 }
+5 ; 10 ; case object Cow ; 20 ; class Moo { override def toString = "Moooooo" } ; 30 ; def
+bippy = {
+  1 +
+  2 +
+  3 } ; bippy+88+11
+
+object Bovine { var x: scala.List[_] = null } ; case class Ruminant(x: scala.Int) ; bippy * bippy * bippy
+Bovine.x = scala.List(Ruminant(5), Cow, new Moo)
+Bovine.x
+
+(2)
+(2 + 2)
+((2 + 2))
+  ((2 + 2))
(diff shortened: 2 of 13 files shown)


dwijnand:
Also please let me know if there are any more places that need changing for `-Yno-imports` and `-Yno-predef`.

Btw, it's very handy that @scala-jenkins automatically adds pull-requests to the right milestone!

retronym:
Did you try `:power` mode?

retronym:
```
def preambleHeader = "class %s extends Serializable { "
```

(under `-Yrepl-classbased`)

retronym:
```
eval("new javax.script.CompiledScript { def eval(context: javax.script.ScriptContext): Object = { " + script + " }.asInstanceOf[Object]; def getEngine: javax.script.ScriptEngine = engine }")
```

retronym:
```
"scala.runtime.ScalaRunTime.replStringOf(%s, %s)".format(x, maxlen)
```

test/files/run/repl-no-imports-no-predef.scala
```diff
@@ -0,0 +1,112 @@
...
+    settings.noimports.value = true
+    settings.nopredef.value = true
+    settings
+  }
+
+  // replace indylambda function names by <function0>
+  override def eval() = {
+    val lines = super.eval
+    val r = """\$\$Lambda.*""".r
+    lines.map(l => r.replaceAllIn(l, "<function0>"))
```
som-snytt: You can override `normalize` https://github.com/scala/scala/blob/2.12.x/src/partest-extras/scala/tools/partest/ReplTest.scala#L24

som-snytt:
That's great. I noticed that REPL was dissed on gitter for not supporting `-Yno-nothing`, so I was going to do it, too. I just verified that fixing `_root_.java.io.Serializable` makes `-Yrepl-class-based` work.

I think retronym's comment on CompiledScript is now at

https://github.com/scala/scala/blob/2.12.x/src/repl/scala/tools/nsc/interpreter/Scripted.scala#L45

som-snytt:
Boy, that was kind of annoying, but:

https://github.com/som-snytt/scala/commit/faf6059845712bc6f7d2381c8fc67b2fd31cdafd

Probably there should be a neg test to show that it's actually not importing.

dwijnand:
Cherry-picked @som-snytt's work in `Scripted`, updated to test for `:power`, `-Yrepl-classbased` and switched from `ReplTest#normalise`.

@retronym 
`"scala.runtime.ScalaRunTime.replStringOf(%s, %s)".format(x, maxlen)` I think is fine, the string part is fully qualified, and the `.format` part isn't running under `-Yno-predef` so the string gets augmented.

I can't figure out where

``` scala
eval("new javax.script.CompiledScript { def eval(context: javax.script.ScriptContext): Object = { " + script + " }.asInstanceOf[Object]; def getEngine: javax.script.ScriptEngine = engine }")
```

comes from, how to trigger it, or even better how to test for it.

Thank you for the review.

retronym:
>  I think is fine, the string part is fully qualified

I wan't able to trigger a failure with it, but what I had in mind was `"_root_.scala.runtime...."`.

dwijnand:
Done.

retronym:
The change is looking comprehensive now.

I used IntelliJ to search in literal strings for suspicious patterns:

image

retronym:
`_root_` is an extra safeguard if the generated code is nested within a context that has unknown enclosing imports or defintions, that might shadow `java` or `scala`. It would be good to use this consistently IMO.

dwijnand:
Updated.

som-snytt:
Yes, I had `_root_`ed but when the unit test still failed, I backed it out, thinking I must have mistyped something; it turned out to be a missing `scala.Any` in the init code for Scripted, and the failure was ignored, because how could it not compile? Too late at night to re-`_root_`, sorry. Feel free to squash my commit.

dwijnand:
Thanks @som-snytt, squashed.

adriaanm:
🌲 root stuff! 🌲

dwijnand:
oh, it's green!

som-snytt:
I was rooting for you.

retronym:
LGTM. Thanks and welcome, @dwijnand!

test/files/run/repl-no-imports-no-predef-classbased.scala
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  def code = """
+case class K(s: java.lang.String)
+class C { implicit val k: K = K("OK?"); override def toString = "C(" + k.toString + ")" }
+val c = new C
+import c.k
+scala.Predef.implicitly[K]
+val k = 42
+k                 // was K(OK?)
```
som-snytt, line 17: I don't know what this is trying to test. There's the bug with `-Yrepl-class-based` that imports of implicits aren't handled correctly, but that has nothing to do with `-Yno-imports`. https://issues.scala-lang.org/browse/SI-9799

dwijnand, line 17: As I tried to convey in the pull request description I took a bit of a shotgun approach to trying to find edge cases to fix, stealing code from existing repl test cases.

som-snytt, line 17: OK, thanks. The test fails a couple of ways on #5220 which I'm trying to put to bed before a flight later...

PR #5193, opened 2016-05-27
SI-9794 Error advice uses decoded method name

som-snytt:
So much work went into polishing this error message,
it's worth buffing the method name when it's an operator.
The message now says `+` instead of `$plus`.


adriaanm:
LGTM

PR #5194, opened 2016-05-27
Topic/repl wrappers

som-snytt:
Rebased and inched forward. Checking the checks.


adriaanm:
Lets assign this to RC1 to give you a little more time to reach the finish line.

som-snytt:
Probably the ball is already over the goal line, I just have to fall on it. (As in American football.)

Is it OK to squash to one commit? I'll make some time by the weekend.

adriaanm:
I'm currently at the stage where I'm training on mapping names of teams to the kind of sport they sports -- sports metaphors is next semester. [google intervenes] carry the ball! and yes, feel free to squash it

som-snytt:
Here is the squashed rebase. I haven't amended the message yet.

Now that the company is no longer Typesafe, there must be exploitable sports puns around the "safety", https://en.wikipedia.org/wiki/List_of_safety_records  The actor Fred Dryer ("Hunter") is the only NFL player to have two safeties in a game.  https://en.wikipedia.org/wiki/Safety_%28gridiron_football_score%29

The difference between a safety and a touchback is "impetus".  I think "impetus" could become an important concept in software development, perhaps more crucial than "velocity". Impetus means who got the ball to cross the goal line of the offensive team, who would prefer to cross the goal line at the other end of the field.

test/files/run/repl-classbased.check
```diff
@@ -6,18 +6,26 @@ scala> class C { implicit val k: K = K("OK?"); override def toString = s"C($k)"
 defined class C
 
 scala> val c = new C
-c: C = C(K(OK?))
+c: INSTANCE.C = C(K(OK?))
```
som-snytt, line 9: This was not intended. You saw where it's handled?

adriaanm, line 9: I intended very little with my changes beyond the impetus to get this lacrosse the line. (Which was a big intention in its own right.)

som-snytt, line 9: That note was intended for Som, I think.

som-snytt:
Fails a scaladoc scalacheck test! I think I've read this test before:

```
    // these things can be tricky
    // this test _will_ fail again some time in the future.
```

I mean this particular test.

src/repl/scala/tools/nsc/interpreter/IMain.scala
```diff
@@ -900,10 +896,7 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends
                           |}
                           |""".stripMargin
 
-      import nme.{ INTERPRETER_IMPORT_WRAPPER => iw }
-
-      /** Adds a val that instantiates the wrapping class. */
-      def postwrap = s"}\nval $iw = new $iw\n"
+      def postwrap = s"}\n"
```
dwijnand: this makes it match `ObjectBasedWrapper`, so perhaps you want to just define it once in `Wrapper`

som-snytt: Thanks for pointing that out @dwijnand . The Wrapper also doesn't need to be passed into the Imports generator. There is no more import wrapping and therefore no need for postwrap at all.

som-snytt:
History is borked. Now at https://github.com/scala/scala/pull/5220

PR #5197, opened 2016-05-27
SI-9104 Autodetect raw pastage

som-snytt:
If `-raw` is not supplied explicitly to REPL `:paste`,
see if the code text starts with `package` keyword or
else see if it parses to a named package (to cope with
leading commentary). In that case, take it as raw.

This was a paulp idea from a while ago.

I couldn't get the existentials to align, trying to match on `intp.parse.Success(_)` from `ILoop`; quite an unnecessary detour trying to make that work.


som-snytt:
Build went off the rails.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -758,14 +758,18 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
       if (!errless) echo("There were compilation errors!")
     }
     if (code.nonEmpty) {
-      if (raw) compileCode() else interpretCode()
+      val compiling = raw || {
+        code match { case paste.leadingPackage() => true case _ => false }
+      } || intp.parse.packaged(code)
```
som-snytt: You could detect possible leading comment and only parse in that case.

som-snytt:
I was aware of the interesting mechanism to convert Unit to `LoopCommands.Result`. It took me just a second, when changing a def from Unit to Boolean, to see what broke. But there is a deep lesson about value discarding that I'll unpack later. Another lesson about having too many things named `Result`. Also, "result" < "re-saltare", rebound, bounce back. I bet there's a lesson in there, too, about the nature of a result.

som-snytt:
What if there were a hook for what to do with value discards? You could error or warn at compile time, but also log at runtime, something like exception handling. Someone wanted a mechanism to say, if this result isn't used, it's an error.

adriaanm:
Could you rebase & squash?

som-snytt:
OK, now I'm contributing only half as many commits.

I'd forgotten about that recent conflicting change.

adriaanm:
But just as many locs!

som-snytt:
Too many loci makes you loco. Seems to work only after an import.

```
scala> :pa

package p { }

<console>:1: error: illegal start of definition
package p { }
^

scala> import concurrent._
import concurrent._

scala> :pa

package p { }

scala> :quit
```

PR #5198, opened 2016-05-28
Privatize the deprecated members of `immutable.Range`.

sjrd:
The implementation of these obscure members of `Range` are uselessly complicated for the purposes of `Range` itself. Making them private will allow to relax their semantics to the specific needs of `Range`, making them simpler, together with the initialization code of `Range`.

`terminalElement` becomes dead code and is removed.

Subsequently, we relax the initialization of `lastElement` not to care about the empty case, since it is only used in code paths where the range is non-empty.

About compatibility with 2.11: this is not the first breaking change of deprecated parts of `Range` in 2.12. `Range` was sealed in https://github.com/scala/scala/commit/cb1a4524d2b34605232afa083dd43f0b7d39b7a7, according to a deprecated inheritance that was added in 2.11. The present PR is arguably a less damaging break than that.

### DIFF
--- a/src/library/scala/collection/immutable/Range.scala
+++ b/src/library/scala/collection/immutable/Range.scala
@@ -81,8 +81,7 @@ extends scala.collection.AbstractSeq[Int]
     || (start == end && !isInclusive)
   )
 
-  @deprecated("this method will be made private, use `length` instead", "2.11.0")
-  final val numRangeElements: Int = {
+  private val numRangeElements: Int = {
     if (step == 0) throw new IllegalArgumentException("step cannot be 0.")
     else if (isEmpty) 0
     else {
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @Ichoran. Also /cc @adriaanm regarding the 2.11/2.12 compatibility and the possibility of removing deprecated members.

See also the mailing list discussion https://groups.google.com/forum/#!topic/scala-internals/OTfJuM747c0

sjrd:
Hum ... JVM segaults in one run test in each commit (not the same both times).

Ichoran:
LGTM

lrytz:
conflit de liaison à cause de https://github.com/scala/scala/pull/5076..

sjrd:
Rebased.

lrytz:
/rebuild -- `*** Error in '/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java': free(): invalid pointer: 0x0000000100395c28 ***`

PR #5199, opened 2016-05-28
Make removing forkjoin dependent on the organization key

dwijnand:
Such that changing the organization (eg. to "org.typelevel") does the right thing.

I'm happy to iterate if there's an alternative fix.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -220,7 +220,7 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   // directly to stdout
   outputStrategy in run := Some(StdoutOutput),
   Quiet.silenceScalaBinaryVersionWarning
-)
+) ++ removePomDependencies
 
 /** Extra post-processing for the published POM files. These are needed to create POMs that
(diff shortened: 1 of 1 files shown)


milessabin:
:+1:

retronym:
Perhaps we ought to refactor things to make this a little easier to understand.

How about something like:

```
val pomDependencyExclusions = settingKey[Seq[GroupArtifactID]]

pomDependencyExclusions += organization.value % "forkjoin"

pomPostProcess := {
  // don't really need to run the old value of `pomPostProcess` anymore as we know this is the only customization

 val x = pomDependencyExclusions.value
 ...
}
```

Note that in the 2.12.x branch, we use the same POM filtering code:

```
pomDependencyExclusions ++= List("org.apache.ant" % "ant", "org.scala-lang.modules" % "scala-asm")
```

dwijnand:
@retronym I like that idea, pushed. Sadly I can't use `GroupArtifactID` because its properties are all `private[sbt]`..

dwijnand:
Rebased, now that the pr validation changes have been merged.

szeiger:
LGTM

PR #5200, opened 2016-05-30
Don't pass -opt to starr / locker build in the bootstrap script

lrytz:
Renaming -Yopt to -opt revealed that we're passing the flag when
building the locker (and optionally the starr) version. This is not
necessary: when building the next stage with the flag enabled, the same
optimizations are performed no matter if the current stage was built
with the flag or not.


lrytz:
review by @adriaanm, should fix the bootstrap build (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/440/artifact/logs/builds/*view*/). Started a build here https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/442

retronym:
LGTM

PR #5201, opened 2016-05-31
Topic/copyrighter

som-snytt:
yacc, yet another copyright comparator.

The tool shows current (start, end) and recommended replacements, which it applies on `--update`. The squiggly arrow warns that the end date was previously advanced past the last commit.

```
src/compiler/scala/tools/nsc/ast/NodePrinters.scala: (2005, 2013->2015)
src/compiler/scala/tools/nsc/ast/parser/BracePair.scala: (2005, 2013~>2012)
```

With `--follow`, it shows that `App.scala` should bear the earlier date for `Application.scala`, from which it is derived.

```
< src/library/scala/App.scala: (2010, 2013->2014)

---
> src/library/scala/App.scala: (2010->2004, 2013->2014)
```

But `runtime/AbstractPartialFunction` wants to go too far, probably because most of the initial commit was just the copyright banner itself:

```
< src/library/scala/runtime/AbstractPartialFunction.scala: (2013->2011, ?->2015)

---
> src/library/scala/runtime/AbstractPartialFunction.scala: (2013->2005, ?->2015)
```

Notice that its date was incorrectly updated from 2011 to 2013 in a huge "update copyrights" commit.

Cherry-picked the `App` commits which haven't landed from 2.11 yet.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -379,11 +379,15 @@ self =>
           case DefDef(_, nme.main, Nil, List(_), _, _)  => true
           case _                                        => false
         }
+        def isApp(t: Tree) = t match {
+          case Template(ps, _, _) => ps.exists { case Ident(x) if x.decoded == "App" => true ; case _ => false }
+          case _ => false
+        }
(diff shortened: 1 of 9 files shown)


som-snytt:
Path-dependent handling, like excluding `test/**`? And multiformats.

PR #5202, opened 2016-05-31
Keep line numbers when inlining from the same compilation unit

lrytz:
So far, line numbers were kept only when inlining from the same class.
We can also keep them when inlining from a different class defined in
the same compilation unit.

Longer-term we should support JSR-45, see SI-7518 and scala-dev#3.

### DIFF
--- a/src/reflect/scala/reflect/internal/Flags.scala
+++ b/src/reflect/scala/reflect/internal/Flags.scala
@@ -234,13 +234,8 @@ class Flags extends ModifierFlags {
    */
   final val AllFlags = -1L
 
-  /** These flags can be set when class or module symbol is first created.
-   */
-  final val TopLevelCreationFlags =
-    MODULE | PACKAGE | FINAL | JAVA
-
   // TODO - there's no call to slap four flags onto every package.
-  final val PackageFlags = TopLevelCreationFlags
+  final val PackageFlags = MODULE | PACKAGE | FINAL | JAVA
 
   // FINAL not included here due to possibility of object overriding.
   // In fact, FINAL should not be attached regardless.  We should be able
@@ -300,7 +295,7 @@ class Flags extends ModifierFlags {
   final val ConstrFlags = JAVA
 
   /** Module flags inherited by their module-class */
-  final val ModuleToClassFlags = AccessFlags | TopLevelCreationFlags | CASE | SYNTHETIC
+  final val ModuleToClassFlags = AccessFlags | PackageFlags | CASE | SYNTHETIC
 
   /** These flags are not pickled */
   final val FlagsNotPickled = IS_ERROR | OVERLOADED | LIFTED | TRANS_FLAG | LOCKED | TRIEDCOOKING
--- a/src/reflect/scala/reflect/internal/Flags.scala
+++ b/src/reflect/scala/reflect/internal/Flags.scala
@@ -235,7 +235,6 @@ class Flags extends ModifierFlags {
   final val AllFlags = -1L
 
   /** These flags can be set when class or module symbol is first created.
-   *  They are the only flags to survive a call to resetFlags().
    */
   final val TopLevelCreationFlags =
     MODULE | PACKAGE | FINAL | JAVA
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -725,7 +725,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
     def setFlag(mask: Long): this.type   = { _rawflags |= mask ; this }
     def resetFlag(mask: Long): this.type = { _rawflags &= ~mask ; this }
-    def resetFlags() { rawflags &= TopLevelCreationFlags }
+    def resetFlags() { rawflags = 0 }
 
     /** Default implementation calls the generic string function, which
      *  will print overloaded flags as <flag1/flag2/flag3>.  Subclasses
--- a/src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala
@@ -77,15 +77,16 @@ abstract class GenBCode extends BCodeSyncAndTry {
 
     /* ---------------- q2 ---------------- */
 
-    case class Item2(arrivalPos:   Int,
-                     mirror:       asm.tree.ClassNode,
-                     plain:        asm.tree.ClassNode,
-                     bean:         asm.tree.ClassNode,
-                     outFolder:    scala.tools.nsc.io.AbstractFile) {
+    case class Item2(arrivalPos:     Int,
+                     mirror:         asm.tree.ClassNode,
+                     plain:          asm.tree.ClassNode,
+                     bean:           asm.tree.ClassNode,
+                     sourceFilePath: String,
+                     outFolder:      scala.tools.nsc.io.AbstractFile) {
       def isPoison = { arrivalPos == Int.MaxValue }
     }
 
-    private val poison2 = Item2(Int.MaxValue, null, null, null, null)
+    private val poison2 = Item2(Int.MaxValue, null, null, null, null, null)
     private val q2 = new _root_.java.util.LinkedList[Item2]
 
     /* ---------------- q3 ---------------- */
@@ -205,6 +206,7 @@ abstract class GenBCode extends BCodeSyncAndTry {
         val item2 =
           Item2(arrivalPos,
                 mirrorC, plainC, beanC,
+                cunit.source.file.canonicalPath,
                 outF)
 
         q2 add item2 // at the very end of this method so that no Worker2 thread starts mutating before we're done.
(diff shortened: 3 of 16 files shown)


retronym:
```
Testcase: keepLineNumbersPerCompilationUnit took 0.006 sec
    Caused an ERROR
List() (of class scala.collection.immutable.Nil$)
scala.MatchError: List() (of class scala.collection.immutable.Nil$)
    at scala.tools.nsc.backend.jvm.opt.InlinerTest.keepLineNumbersPerCompilationUnit(InlinerTest.scala:1560)
    at scala.tools.testing.ClearAfterClass$1.evaluate(ClearAfterClass.java:29)

```

lrytz:
yeah, there's some cross-talk between tests: it fails when running all of `InlinerTest`, but passes individually..

lrytz:
ah, it turns out i need https://github.com/scala/scala/pull/5177/commits/e82e457504f730f5aaeb9e7e917467b36311369b, which is part of your pr :) i'll cherry-pick it here.

retronym:
I guess we'd have to push this further to support the case where we transitively inline code from the current compilation unit via a method in another?

```
object A {
  @inline def a1 = {
    println("A.scala:3")
  }
  def a2 = B.b1
}
```

```
object B {
  @inline def b1 = A.a1
}
```

With this PR:

```
  public void b1();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=2, locals=1, args_size=1
         0: getstatic     #18                 // Field A$.MODULE$:LA$;
         3: ifnonnull     8
         6: aconst_null
         7: athrow
         8: getstatic     #23                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
        11: ldc           #25                 // String A.scala:3
        13: invokevirtual #29                 // Method scala/Predef$.println:(Ljava/lang/Object;)V
        16: return
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      17     0  this   LB$;

  public void a2();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=2, locals=1, args_size=1
         0: getstatic     #31                 // Field B$.MODULE$:LB$;
         3: ifnonnull     8
         6: aconst_null
         7: athrow
         8: getstatic     #33                 // Field MODULE$:LA$;
        11: ifnonnull     16
        14: aconst_null
        15: athrow
        16: getstatic     #18                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
        19: ldc           #20                 // String A.scala:3
        21: invokevirtual #24                 // Method scala/Predef$.println:(Ljava/lang/Object;)V
        24: return
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      25     0  this   LA$;
      LineNumberTable:
        line 5: 0
```

I tried to come up with a less contrived example involving lambdas, but wasn't able to. We'd need something that inlines the code in the lambda impl method through a pattern of code like:

```
class A { B.b(() => {
  "line 2"
}})
```

```
object B { @inine def b(f: () => Any) = f(()
```

lrytz:
right, good observation. this is not supported. for the common case involving lambdas i think we're fine though: the inlined body of the higher-oder method won't have line numbers, but it will invoke the lambda body method which has line numbers again (whether it's inlined or not).

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -725,7 +725,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
     def setFlag(mask: Long): this.type   = { _rawflags |= mask ; this }
     def resetFlag(mask: Long): this.type = { _rawflags &= ~mask ; this }
-    def resetFlags() { rawflags &= TopLevelCreationFlags }
+    def resetFlags() { rawflags = 0 }
```
retronym, line 728: This will invalidate the comment in:

```
  /** These flags can be set when class or module symbol is first created.
   *  They are the only flags to survive a call to resetFlags().
   */
  final val TopLevelCreationFlags =
    MODULE | PACKAGE | FINAL | JAVA

  // TODO - there's no call to slap four flags onto every package.
  final val PackageFlags = TopLevelCreationFlags
```

lrytz, line 728: right. cc @adriaanm - what do you think about this change? note the commit message, which says:

```
clear all flags when resetting a symbol
this change is a bit scary because it changes code that's not been
changed in 11 years
7fa7c93#diff-d5789e5ae5061197d782d08324b260dbL214
```

adriaanm, line 728: Could you add some motivation for the new behavior? It does seem pretty weird to preserve these flags... I could see something going between final and not-final, for example. Maybe we could strip the `TopLevelCreationFlags` to its essence?

adriaanm, line 728: All this flags business really makes me wish we tracked flags in the TypeHistory, just like a symbol's info.

lrytz, line 728: the motivation was indeed `final`, see the test case: https://github.com/scala/scala/pull/5202/commits/fbf549b615786f5d749de5022b0fefdc1e2ab435#diff-80a0746bc22d5376fd194dfb8985330d

i think the non-hacky solution would be to remove `TopLevelCreationFlags`, its uses are more like "oh, this seems to be a well-fitting collection of flags for my use-case"

retronym, line 728: It is worth considering that this code is used not only in resident compilation, but also in single-run compilation when entering a `ClassDef` that redefines a class symbol on the compilation classpath.  See `enterClassSymbol` / `updatePosFlags`.

In any case, I'm struggling to understand the motivation for `TopLevelCreationFlags`. 

The `JAVA` flag is also one that should not be persisted: if the current compilation run redefines a class on the classpath that used to be Java compiled, we should forget the `JAVA` flag.

Maybe there is something more nuanced about `PACKAGE` / `MODULE`.

lrytz, line 728: i checked in dotty, there's [`PackageCreationFlags`](https://github.com/lampepfl/dotty/blob/845b98186047f38013e2f6aa35508e974eedafb7/src/dotty/tools/dotc/core/Flags.scala#L504) which is the same as scala's `PackageFlags`, which are in fact `TopLevelCreationFlags`. but dotty doesn't have something like `TopLevelCreationFlags`. dotty actually seems to create a new symbol instead of updating the existing one (https://github.com/lampepfl/dotty/blob/845b98186047f38013e2f6aa35508e974eedafb7/src/dotty/tools/dotc/typer/Namer.scala#L258-L279).

retronym:
This is in need of a rebase.

src/reflect/scala/reflect/internal/Symbols.scala
retronym, line 728: I'm happy to go forward with the total reset of the flags

retronym:
Rebased in https://github.com/scala/scala/pull/5217

lrytz:
I pushed the commits you re-based here and added another one to remove `TopLevelCreationFlags`

retronym:
LGTM

adriaanm:
Looks like the removal of `TopLevelCreationFlags` broke the compiler.

src/reflect/scala/reflect/internal/Flags.scala
```diff
@@ -300,7 +295,7 @@ class Flags extends ModifierFlags {
   final val ConstrFlags = JAVA
 
   /** Module flags inherited by their module-class */
-  final val ModuleToClassFlags = AccessFlags | TopLevelCreationFlags | CASE | SYNTHETIC
+  final val ModuleToClassFlags = AccessFlags | MODULE | FINAL | JAVA | CASE | SYNTHETIC
```
adriaanm: `PACKAGE` objects broken by this?

lrytz: i thought about this.. not well enough..

Xuxue1:
today I am trying import scala code into idea  but i hava a bug  that is the Exception 

---

java.lang.StringIndexOutOfBoundsException: String index out of range: 7
        at java.lang.String.substring(Unknown Source)
        at VersionUtil$$anonfun$versionPropertiesImpl$1.apply(VersionUtil.scala:80)
        at VersionUtil$$anonfun$versionPropertiesImpl$1.apply(VersionUtil.scala:60)
        at sbt.EvaluateSettings$MixedNode.evaluate0(INode.scala:175)
        at sbt.EvaluateSettings$INode.evaluate(INode.scala:135)
        at sbt.EvaluateSettings$$anonfun$sbt$EvaluateSettings$$submitEvaluate$1.apply$mcV$sp(INode.scala:69)
        at sbt.EvaluateSettings.sbt$EvaluateSettings$$run0(INode.scala:78)
        at sbt.EvaluateSettings$$anon$3.run(INode.scala:74)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.lang.Thread.run(Unknown Source)

---

then  I modify VersionUtil.scala at 79 line  "val sha = executeTool("get-scala-commit-sha").substring(0, 7)“    into  ”val sha = executeTool("get-scala-commit-sha")“   then import  SUCCESS  . that is  a  bug?

retronym:
@Xuxue1 That does seem like a bug in the build. What is the connection to this pull request though?

In case you haven't seen it, here's our documentation on [IDE setup](https://github.com/scala/scala/#ide-setup)

PR #5205, opened 2016-06-01
Small optimizations around use of Scopes.

adriaanm:
`Scope`'s `filter` is implemented using `toList`, so may as well start with `toList`ourselves.

Also fused some `filter`/`foreach` combos.


retronym:
LGTM

PR #5206, opened 2016-06-02
Fields pt2 [ci: last-only]

adriaanm:
#5141, the sequel


adriaanm:
[Bytecode diff](https://gist.github.com/adriaanm/9e9e6d2b69ead0b20ef6fd1c47d3ed88) shows two changes:
- more methods are final: ✅ 
- captured object refs are made not-private: ❓ it means we use invokeinterface instead of invokespecial.. maybe better to leave them private

In fact, generally speaking we should leave `private` on more private trait methods, as long as they don't need to be implemented in classes.

src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
```diff
@@ -271,7 +272,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
           //
           // However, due to https://github.com/scala/scala-dev/issues/126, this currently does not
           // work, the abstract accessor for O will be marked effectivelyFinal.
```
lrytz, line 274: should remove that last part of the comment, since it fixes https://github.com/scala/scala-dev/issues/126 (right?)

adriaanm, line 274: Yep. I think I fixed it twice actually :-) once by removing lateDeferred and once by tweaking isEffectivelyFinal

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -83,7 +83,7 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
   private def accessorImplementedInSubclass(accessor: Symbol) =
     (accessor hasFlag SYNTHESIZE_IMPL_IN_SUBCLASS) && (accessor hasFlag (ACCESSOR | MODULE))
 
-  private def concreteOrSynthImpl(sym: Symbol): Boolean = !(sym hasFlag DEFERRED) || (sym hasFlag SYNTHESIZE_IMPL_IN_SUBCLASS)
+  @inline final def notDeferredOrSynthImpl(sym: Symbol): Boolean = !(sym hasFlag DEFERRED) || (sym hasFlag SYNTHESIZE_IMPL_IN_SUBCLASS)
```
lrytz, line 88: what prompted you to put `@inline` here? i think the vm usually handles small first-order methods well.

adriaanm, line 88: because I added another layer [below in Mixin](https://github.com/scala/scala/pull/5206/files/c498000413303c837fa8c66865a47471b4fdb087..fa95dea53e3f36854c132a2f0cea5c6d12d8b3a3#diff-8b2a191e6261b31a9a6462e4a01d3b9cR110) -- which itself was motivated by minimal diff

test/junit/scala/tools/nsc/backend/jvm/opt/ScalaInlineInfoTest.scala
```diff
@@ -75,7 +75,7 @@ class ScalaInlineInfoTest extends BytecodeTesting {
       false, // final class
       None, // not a sam
       Map(
-        ("O()LT$O$;",                                                 MethodInlineInfo(false,false,false)),
+        ("O()LT$O$;",                                                 MethodInlineInfo(false ,false,false)),
```
adriaanm, line 78: this shows scala-dev#216 was already fixed (should also fix the whitespace error)

lrytz, line 78: right, i saw this in the other PR

lrytz:
this also looks good, i'll take another pass after squashing / rebasing

adriaanm:
Cool! I'll merge this into the main fields PR

PR #5207, opened 2016-06-02
Fix comparisons involving NaN

lrytz:
Floating point comparisons involving NaN should always return false,
except for !=. Fixes a regression introduced by #4963.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/ScalaPrimitives.scala
+++ b/src/compiler/scala/tools/nsc/backend/ScalaPrimitives.scala
@@ -61,8 +61,8 @@ abstract class ScalaPrimitives {
   final val NE = 43                            // x != y
   final val LT = 44                            // x < y
   final val LE = 45                            // x <= y
-  final val GE = 46                            // x > y
-  final val GT = 47                            // x >= y
+  final val GT = 46                            // x > y
+  final val GE = 47                            // x >= y
 
   // Boolean unary operations
   final val ZNOT = 50                          // !x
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/backend/ScalaPrimitives.scala
```diff
@@ -61,8 +61,8 @@ abstract class ScalaPrimitives {
   final val NE = 43                            // x != y
   final val LT = 44                            // x < y
   final val LE = 45                            // x <= y
-  final val GE = 46                            // x > y
-  final val GT = 47                            // x >= y
+  final val GT = 46                            // x > y
+  final val GE = 47                            // x >= y
```
lrytz, line 65: this part is not a bugfix, it just cleans up the inconsistency between the code and the comment on the right.

lrytz:
review by @retronym

retronym:
LGTM

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1110,22 +1110,19 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
     }
 
     /* Emit code to compare the two top-most stack values using the 'op' operator. */
-    private def genCJUMP(success: asm.Label, failure: asm.Label, op: TestOp, tk: BType, targetIfNoJump: asm.Label) {
-      if (targetIfNoJump == success) genCJUMP(failure, success, op.negate, tk, targetIfNoJump)
+    private def genCJUMP(success: asm.Label, failure: asm.Label, op: TestOp, tk: BType, targetIfNoJump: asm.Label, negated: Boolean = false) {
+      if (targetIfNoJump == success) genCJUMP(failure, success, op.negate, tk, targetIfNoJump, negated = !negated)
```
liufengyun, line 1114: @lrytz I encountered the same problem in Dotty. 

I have a hacky fix which passed the CI without introducing `negated` --- I'm afraid some corner cases might be not handled. Could you please advise? Thanks.

https://github.com/lampepfl/dotty/pull/6722

PR #5208, opened 2016-06-03
[SI-9800] Implement variants of 'par' method that allow user to set number of threads or thread pool in-line

erikerlandson:


erikerlandson:
I wrote this PR against 2.11.x, but as far as I can tell, the change should work against 2.10.x, 2.11.x and 2.12.x.  I think that would be good, not sure what the workflow would be.

retronym:
We don't change API in minor releases due to our policy of maintaining forward- and backward binary compatibility, so this would need to be discussed in the context of a pull request to 2.12.x.

erikerlandson:
@retronym, thx -- it will be easy to re-submit against 2.12 series

PR #5209, opened 2016-06-03
Prohibit @native method in trait

adriaanm:
On the JVM, a @native interface method results in a VerifyError.

Other platforms could decide to be more permissive, but it seems
like allowing them in classes is enough.

Follow up for #5138


retronym:
LGTM, thanks

lrytz:
```
fail - jvm/natives.scala  [compilation failed]% scalac natives.scala
```

adriaanm:
Oh I see -- will fix.

PR #5210, opened 2016-06-03
SI-9799 Keep ImportHandlers that imports implicit symbols for -Yrepl-class-based

zsxwing:
Since we don't know whether the implicit imports are necessary, they should be always kept.


ScrapCodes:
No this change is not going to help us. Removing that line will lead to importing much more than necessary for declaring the class. 

It will lead to surfacing of old fixed bugs like: SPARK-1199. You can try running ReplSuite in spark for scala 2.10 by making your change as above.

zsxwing:
Closed as this will introduce other issues.

PR #5213, opened 2016-06-03
SI-9737 [no-merge] Backport stringOf ParIterable

som-snytt:
Cherry-picked c5f3d3f286ee5c26c8ddcf10f6878058e8f7e040

Edited comment: in stringOf, let GenIterable subsume
both Iterable and ParIterable.

This change is required for Scala.js compatibility as it does not
support parallel collections.

Conflicts:

```
src/library/scala/runtime/ScalaRunTime.scala
```

### DIFF
--- a/src/library/scala/runtime/ScalaRunTime.scala
+++ b/src/library/scala/runtime/ScalaRunTime.scala
@@ -9,7 +9,7 @@
 package scala
 package runtime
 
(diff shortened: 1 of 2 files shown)


som-snytt:
I normally prefer dotless infix, but I was tempted to re-dot the changed line and neighboring lines, so now I'm swinging more as Odersky recommends. Which kind of makes me feel like an old fogey, but I have to be honest with myself.

retronym:
LGTM

PR #5218, opened 2016-06-07
SI-9806 Fix incorrect codegen with optimizer, constants, try/catch

retronym:
The constant optimizer phase performs abstract interpretation of the
icode representation of the progam in order to eliminate dead
code.

For each basic block, the possible and impossible states of each local
variable is computed for both a normal and an exceptional exit.

A bug in this code incorrectly tracked state for exception exits.
This appears to have been an oversight: the new state was computed
at each instruction, but it was discarded rather than folded through
the intepreter.


lrytz:
LGTM

dragos:
Maybe this could be merged now?

PR #5221, opened 2016-06-09
SI-7594 Test for line numbers

som-snytt:
Line numbers in value class in DelayedInit progressed in 2.12.

Nothing to do with other line numbers work.

### DIFF
--- a/test/junit/scala/tools/nsc/backend/jvm/LineTest.scala
+++ b/test/junit/scala/tools/nsc/backend/jvm/LineTest.scala
@@ -1,0 +1,39 @@
+
+package scala.tools.nsc.backend.jvm
+
+import org.junit.Assert._
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+@RunWith(classOf[JUnit4])
+class LineTest {
+  import java.lang.{ StackTraceElement => JSTE }
+  object StackTraceElement {
+    def unapply(ste: JSTE): Option[(String, String, String, Int)] =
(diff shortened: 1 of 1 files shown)


test/junit/scala/tools/nsc/backend/jvm/LineTest.scala
```diff
@@ -0,0 +1,39 @@
...
+}
+
+object SampleMain extends App {
+  implicit class Averrable(val b: Boolean) extends AnyVal {
+    def yes = assert(b)
+    def no = assert(!b)
+  }
+
+  false.yes
+}
```
retronym, line 39: In JUnit tests, this sort of code should be provided as string arguments to a programatic `compile` call, rather than alongside the test itself. The test themselves are compiled with STARR, not quick, so you would not catch a regression.

som-snytt, line 39: Ha. I was thinking it was kind of a junky (imprecise) test, but not for that reason. I switched it to junit to save compiling.

retronym, line 39: See `keepLineNumbersPerCompilationUnit` for an example of how to test this with ASM, rather than inspection of a runtime stack trace.

som-snytt, line 39: Thanks. I saw `BytecodeTest` when I looked under junit, and my eyes glazed over. That's why I stuck with the quick/dirty test. But I'm not unwilling to be edified.

PR #5220, opened 2016-06-09
SI-6623 Avoid $iw wrappers in REPL

som-snytt:
Use a conventional import to bump context depth in REPL templates.

Code is still ordinarily wrapped in a `$read` object.

This is a step toward 6623-like transparency.

`retronym` takes the blame for this innovation.
`adriaanm` collaborated in its commission.
`somsnytt` batted clean-up.

Notice how the handles line up.

PR #5222, opened 2016-06-09
Introduce ReplPrinter as a hook for alternatives to replStringOf

dwijnand:
Requires making the imports available during the eval phase of the REPL.

Review by @retronym.

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/ReplPrinter.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ReplPrinter.scala
@@ -8,7 +8,7 @@ trait ReplPrinter[-A] extends Any {
 }
 
 object ReplPrinter {
-  implicit def default[A]: ReplPrinter[A] = new ReplPrinter[A] {
-    def print(x: A, maxElements: Int) = ScalaRunTime.replStringOf(x, maxElements)
+  implicit object defaultReplPrinter extends ReplPrinter[Any] {
+    def print(x: Any, maxElements: Int) = ScalaRunTime.replStringOf(x, maxElements)
   }
 }
--- a/src/repl/scala/tools/nsc/interpreter/ReplPrinter.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ReplPrinter.scala
@@ -1,0 +1,14 @@
+package scala.tools.nsc
+package interpreter
+
+import scala.runtime.ScalaRunTime
+
+trait ReplPrinter[-A] extends Any {
+  def print(x: A, maxElements: Int): String
+}
+
+object ReplPrinter {
+  implicit def default[A]: ReplPrinter[A] = new ReplPrinter[A] {
+    def print(x: A, maxElements: Int) = ScalaRunTime.replStringOf(x, maxElements)
+  }
+}
--- a/test/files/run/repl-out-dir.check
+++ b/test/files/run/repl-out-dir.check
@@ -11,6 +11,8 @@ repl-out-dir-run.obj
         $eval$.class
         $eval.class
     $line2
+        $eval$$iw$$iw$.class
+        $eval$$iw$.class
         $eval$.class
         $eval.class
         $read$$iw$$iw$.class
@@ -18,6 +20,8 @@ repl-out-dir-run.obj
         $read$.class
         $read.class
     $line3
+        $eval$$iw$$iw$.class
+        $eval$$iw$.class
         $eval$.class
         $eval.class
         $read$$iw$$iw$.class
@@ -27,6 +31,8 @@ repl-out-dir-run.obj
         $read$.class
         $read.class
     $line4
+        $eval$$iw$$iw$.class
+        $eval$$iw$.class
         $eval$.class
         $eval.class
         $read$$iw$$iw$.class
@@ -34,6 +40,8 @@ repl-out-dir-run.obj
         $read$.class
         $read.class
     $line5
+        $eval$$iw$$iw$.class
+        $eval$$iw$.class
         $eval$.class
         $eval.class
         $read$$iw$$iw$.class
--- a/test/files/run/repl-printer-classbased.check
+++ b/test/files/run/repl-printer-classbased.check
@@ -1,0 +1,41 @@
+
+scala> :paste < EOF
+// Entering paste mode (EOF to finish)
+
+trait Show[-A] extends Any { def show(x: A): String }
+object Show {
+  implicit def shouty[Any]: Show[Any] =
+    new Show[Any] { def show(x: Any) = if (x == null) "" else x.toString + "!" }
+}
+EOF
+
+// Exiting paste mode, now interpreting.
+
+defined trait Show
+defined object Show
+
+scala> 
+
+scala> implicitly[Show[Int]] show 23
+res0: String = 23!
+
+scala> 
+
+scala> import scala.tools.nsc.interpreter.ReplPrinter
+import scala.tools.nsc.interpreter.ReplPrinter
+
+scala> implicit def showReplPrinter[A](implicit z: Show[A]): ReplPrinter[A] = new ReplPrinter[A] {
+  def print(x: A, maxElements: Int): String = {
+    val s = z show x
+    val nl = if (s contains "\n") "\n" else ""
+    nl + s + "\n"
+  }
+}
+showReplPrinter: [A](implicit z: Show[A])scala.tools.nsc.interpreter.ReplPrinter[A]
+
+scala> 
+
+scala> 23
+res1: Int = 23!
+
+scala> :quit
--- a/test/files/run/repl-printer-classbased.check
+++ b/test/files/run/repl-printer-classbased.check
@@ -36,6 +36,6 @@ showReplPrinter: [A](implicit z: Show[A])scala.tools.nsc.interpreter.ReplPrinter
 scala> 
 
 scala> 23
-res1: Int = 23!
+res1: Int = 23
 
 scala> :quit
--- a/test/files/run/repl-printer-classbased.check
+++ b/test/files/run/repl-printer-classbased.check
@@ -35,6 +35,8 @@ showReplPrinter: [A](implicit z: Show[A])scala.tools.nsc.interpreter.ReplPrinter
 
 scala> 
 
+scala> // the lack of ! after 23 is expected because we don't enable custom string under -Yrepl-class-based
+
 scala> 23
 res1: Int = 23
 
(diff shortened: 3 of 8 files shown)


dwijnand:
It would be lovely if this could make M5 :-)

Also i expect the unwrapping of $print will start failing once the repl wrappers are gone (#5220).

dwijnand:
Ah didn't realise this had failed the tests. Updated `test/files/run/repl-out-dir.check`.

Also, review by @som-snytt?

som-snytt:
The one everyone asks for is to quote strings. It might be nice to include a standard one for easy import from a suitable namespace, like `ReplPrinter.Quoted.quoter` so `import ReplPrinter.Quoted._`.

This isn't nuanced:

```
scala> "hi"
res0: String = hi

scala> import scala.tools.nsc.interpreter.ReplPrinter
import scala.tools.nsc.interpreter.ReplPrinter

scala> :pa

implicit val `quoted strings`: ReplPrinter[String] =
  new ReplPrinter[String] {
    def print(s: String, max: Int) =
      if (s.contains('"')) s"${"\""*3}$s${"\""*3}"
      else s""""$s""""
  }

quoted strings: scala.tools.nsc.interpreter.ReplPrinter[String] = $anon$1@6ea246af

scala> res0
res1: String = "hi"
```

You quickly run into implicit management by name:

```
scala> implicit val `quoted strings`: ReplPrinter[String] = new ReplPrinter[String] {      
     | def print(s: String, max: Int) = s.toUpperCase
     | }
quoted strings: scala.tools.nsc.interpreter.ReplPrinter[String] = $anon$1@fd4459b

scala> res0
res2: String = HI

scala> implicit val upperly: ReplPrinter[String] = new ReplPrinter[String] {
     | def print(s: String, max: Int) = s.toUpperCase }
upperly: scala.tools.nsc.interpreter.ReplPrinter[String] = $anon$1@6200b644

scala> res0
<console>:19: error: ambiguous implicit values:
 both value quoted strings of type => scala.tools.nsc.interpreter.ReplPrinter[String]
 and value upperly of type => scala.tools.nsc.interpreter.ReplPrinter[String]
 match expected type scala.tools.nsc.interpreter.ReplPrinter[String]
 + "res3: String = " + _root_.scala.tools.nsc.interpreter.replStringOf(res3, 1000)
                                                                      ^
```

This is where smart imports is useful, either better automation or manually quarantine imports. The other ticket for spark has similar problem.

My only concern is how the object nesting in `$eval` affects spark, who introduced `-Yrepl-class-based`. Especially if all the imports are in play? That's what the other ticket is about, this PR https://github.com/scala/scala/pull/5210 for SI-9799.

Really, what's needed is a general way to specify the templating. For example, another requested feature is to emit `val res1: String = "hi"`so that the output line is code.

Since the concerns are out-of-scope, maybe just disable the change under the spark option; unless it's shown not to be a problem.

dwijnand:
@som-snytt I see. I've revert the behaviour under `-Yrepl-class-based`. WDYT?

dwijnand:
Bump?

som-snytt:
More ideas https://issues.scala-lang.org/browse/SI-9829 such as convert controls to escapes in color, CR -> \r or \u000d in red?

src/repl/scala/tools/nsc/interpreter/ReplPrinter.scala
```diff
@@ -0,0 +1,14 @@
...
+package interpreter
+
+import scala.runtime.ScalaRunTime
+
+trait ReplPrinter[-A] extends Any {
+  def print(x: A, maxElements: Int): String
+}
+
+object ReplPrinter {
+  implicit def default[A]: ReplPrinter[A] = new ReplPrinter[A] {
```
adriaanm: Could we have one instance, like so:

```
 implicit object defaultReplPrinter extends ReplPrinter[Any] {
  def print(x: Any, maxElements: Int) = ScalaRunTime.replStringOf(x, maxElements)
}
```

?

dwijnand: Yep, good eye. Done.

test/files/run/repl-printer-classbased.check
```diff
@@ -0,0 +1,41 @@
...
+    val nl = if (s contains "\n") "\n" else ""
+    nl + s + "\n"
+  }
+}
+showReplPrinter: [A](implicit z: Show[A])scala.tools.nsc.interpreter.ReplPrinter[A]
+
+scala> 
+
+scala> 23
+res1: Int = 23
```
adriaanm, line 41: I was expecting a "!" at the end here -- how come there's not?

adriaanm, line 41: Oh, this is the class-based one.

adriaanm, line 41: perhaps add a comment to the test case that says we don't enable custom string of under this mode

som-snytt, line 41: Everyone expects it to end with a bang.

som-snytt:
I just returned from remote regions; I'll give this another look and try it out.

test/files/run/repl-printer-classbased.check
dwijnand, line 41: I wasn't sure how to leave the comment, so I've opted for in the transcript.

dwijnand, line 41: This comment didn't fold because that line it's on didn't change, so, for clarity, I've added a commit that adds the comment.

dwijnand:
Anything left here? This is a REPL-only change. Can it go in?

som-snytt:
The templating part may conflict with #5220, for which I have a test to fix still. I'll try to nail that down this weekend, and try this on top of it.

test/files/run/repl-out-dir.check
```diff
@@ -11,13 +11,17 @@ repl-out-dir-run.obj
         $eval$.class
         $eval.class
     $line2
+        $eval$$iw$$iw$.class
+        $eval$$iw$.class
```
lrytz, line 15: does this mean every repl line generates two additional classfiles? is there a drawback to this? (performance?)

lrytz:
I think it would be weird to have such a mechanism work only in one of `object-based` / `class-based`, there does not seem to be any inherent relation between the two concerns.

test/files/run/repl-out-dir.check
dwijnand, line 15: Yeah it does. I'm not sure what drawbacks there might be, but I avoided it for Spark-reasons.

dwijnand:
Which two concerns do you mean? I removed it for class-based because @som-snytt was concerned about the performance impact for Spark.

lrytz:
i just think it doesn't make sense for a feature like the one you're implementing to be dependent on some internal representation within the repl.

dwijnand:
Yeah it's not ideal, but it's a compromise I personally was fine with.

How do we go about proving that this doesn't affect Spark?

Alternatively how do you feel about an opt-out option, for if it proves to be a problem?

som-snytt:
Also, sorry my weekend went to understanding how imports work. I still intend to help with the concerns.

adriaanm:
Could we do a bit of a cost/benefit analysis of this and #5220, as well as a risk assessment of causing a regression in the repl, which would lengthen the RC cycle? I'm tempted to push this to 2.12.1 so that we can consider this together with #5220 and go a bit deeper in dealing with some templating technical debt, rather than taking on more.

dwijnand:
I don't like the idea of "shackling" together two pull requests which are doing distinct thing, even if they're in the same area of the codebase.

Why are you saying that this adds technical debt?

adriaanm:
I don't think it's unreasonable to consider these PRs together, as they touch very similar parts of the REPL. Having to disable this for the class-based templates seems to imply we're going to have to come back to this and make it work there too, no?

adriaanm:
I also would like to see a more technical response to @lrytz's concern. Everything is a compromise, but what are we conceding and what for? How would we solve this if we had more time? The REPL encoding is already very complicated, so any change is taking on technical debt for a nicer way of implementing the REPL.

som-snytt:
Sorry, I meant to devote time to this, but the day job got busy. Time flies.  Also, it slipped my mind.  Possibly I was hoping it was fixed in ammonite.

adriaanm:
For the reasons stated above, and with this not being a release blocker, I'm re-assigning this to 2.12.1. It would be great to have a concerted effort in cleaning up the REPL encoding once 2.12.0 has shipped.

dwijnand:
I would be happy help clean up the REPL (but I would need some guidance as I have no prior experience of REPL implementations to compare this one against). I'm fine with this going in 2.12.1 rather than 2.12.0, but I would just like it one way or another.

Without doing a cost/benefit analysis of this and #5220, or a risk assessment, would you consider this feature being opt-in via a compiler flag? Or is that a no-go?

adriaanm:
Thanks for understanding -- we're happy to help with this once 2.12.0 is out!

som-snytt:
I've been using spark for day job; you know spark shell doesn't even inherit the improved welcome message in 2.11.8? I sense that there's something important to do for repl re-use and customizability for 2.12, but not sure if I'll find time to realize it.

FTR, there's an episode of "Inspector Morse" which turns on the OED recommendation of "-ize". (A note writer uses "-ise" inappropriately and misspells "desperate" like "separate".) (It was a suicide note, putatively, and this was a few years before the internet.) In hindsight, I would have used stronger language about "-optimise."

http://blog.oxforddictionaries.com/2011/03/ize-or-ise/
http://www.oxforddictionaries.com/words/ize-ise-or-yse

som-snytt:
Because there is no ticket?, linking to https://issues.scala-lang.org/browse/SI-9829 about pretty printing against impossible odds.

adriaanm:
> would you consider this feature being opt-in via a compiler flag? Or is that a no-go?

sorry I missed this the first time -- I would prefer putting this under a flag, as I'm concerned this could slow down the REPL due to additional wrapping

lrytz:
I still think we should come up with a solution that is also fine for `-Yrepl-class-based` / spark. IUC we'd also like to switch to this encoding by default at some point, and then the feature would go away.

dwijnand:
If I put it behind a feature flag, the feature would be enabled under class-based too, IMO.

dwijnand:
Clearly there's more work to do here for this to be still in the 2.12.1 queue.

But thanks for the recent feedback.

dwijnand:
I don't have the stamina to take this to completion if it requires first delivering the "avoid $iw wrappers" change in the REPL.

som-snytt:
There's pressure both to de-wrapperize and avoid object wrappers entirely and also do customizable printing. To which I'd add customizable wrapping. This won't go unused.

PR #5224, opened 2016-06-11
Add -Yno-string-plus-any to opt out of String.+(Any)

dwijnand:
Review by @adriaanm.

### DIFF
--- a/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
@@ -202,6 +202,7 @@ trait ScalaSettings extends AbsScalaSettings
   val inferByName     = BooleanSetting    ("-Yinfer-by-name", "Allow inference of by-name types. This is a temporary option to ease transition. See SI-7899.").withDeprecationMessage(removalIn212)
   val YdisableFlatCpCaching  = BooleanSetting    ("-YdisableFlatCpCaching", "Do not cache flat classpath representation of classpath elements from jars across compiler instances.")
   val YpartialUnification = BooleanSetting ("-Ypartial-unification", "Enable partial unification in type constructor inference")
+  val YnoStringPlusAny = BooleanSetting   ("-Yno-string-plus-any", "Compile without synthetic String.+(Any) method")
 
   val exposeEmptyPackage = BooleanSetting ("-Yexpose-empty-package", "Internal only: expose the empty package.").internalOnly()
   val Ydelambdafy        = ChoiceSetting  ("-Ydelambdafy", "strategy", "Strategy used for translating lambdas into JVM code.", List("inline", "method"), "method")
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -1107,7 +1107,12 @@ trait Definitions extends api.StandardDefinitions {
     lazy val Object_synchronized = newPolyMethod(1, ObjectClass, nme.synchronized_, FINAL)(tps =>
       (Some(List(tps.head.typeConstructor)), tps.head.typeConstructor)
     )
-    lazy val String_+ = enterNewMethod(StringClass, nme.raw.PLUS, AnyTpe :: Nil, StringTpe, FINAL)
+    lazy val String_+ = {
+      if (settings.YnoStringPlusAny) {
+        newMethod(NoSymbol, nme.raw.PLUS, AnyTpe :: Nil, StringTpe, FINAL)
+      } else
+        enterNewMethod(StringClass, nme.raw.PLUS, AnyTpe :: Nil, StringTpe, FINAL)
+    }
 
     def Object_getClass  = getMemberMethod(ObjectClass, nme.getClass_)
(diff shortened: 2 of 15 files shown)


julienrf:
I think you should also disable `Any.+(String)` (ie `any2stringadd`). Maybe the flag could synthesize another implicit, which would lead to ambiguity?

dwijnand:
`Any.+(String)` can already be disabled via either the Predef import trick, or just opting out of Predef. I don't think creating implicit ambiguity is the right way to opt out of an implicit.

cvogt:
Yes! +1K.

odersky:
I think this is great in principle, but I hate the proliferation of -Y options. I propose that now we are at it, we do it the right way: Integrate in language imports. For instance, introduce a language import

```
language.future.noStringPlus
```

If we can agree on something like this we get the assurance that things will continue to work in future releases (which might not honor the same -Y options).

erikvanoosten:
You can already exclude the method on a per file basis with a smart `import` statement. I think the rationale of this PR is to exclude this on an entire codebase. How would one do this with an import statement?

odersky:
@erikvanoosten:

You can compile with `-language:future.noStringPlus`

Language imports are mirrored as compiler flags.

dwijnand:
@erikvanoosten You can't exclude the method on a per file basis.

This PR is about `String.+(Any): String`, I think you're confusing that with `Any.+(String): String)`, aka `any2stringadd` which @julienrf mentioned above.

@odersky Sure. I'll just need a hand figuring out how to check in Typer (see gitter).

erikvanoosten:
@odersky didn't know that option, awesome!
@dwijnand You're right. Thanks.

dwijnand:
@odersky I've switched to `scala.language.future.noStringPlus` (pushed).

One thing that could improve is the error message in `neg/no-string-plus-2`, but I've pushed this as is for now, for review.

I've changed things such that the test cases I've created work, but I'm happy to iterate on the implementation details.

dwijnand:
Ah.. looks like my use of `immediate = true` has broken existing language import usage, when bootstrapping... :-(

dwijnand:
I've switched this branch back to the compiler flag implementation - it seems the way language imports are implemented is by a delayed side-effect of sending a warning or an error to the reporter, which doesn't work for me as I need to know immediately if to opt-out of the synthetic `String.+(Any): String` such that a user-land implicit conversion can be sought (as demonstrated in the test cases).

I've kept my attempt at implementing it as a language import at https://github.com/scala/scala/compare/2.12.x...dwijnand:string-plus-any-language-import

som-snytt:
I was looking at the `language` setting the other way around:

If `noStringPlus` is set, don't enter `PLUS` method on `String` in `Definitions`. 

If `noStringPlus` is set to false by `-language:-future.noStringPlus`, then also don't enter `PLUS`, but if the typecheck fails because no implicit conversion supplies a plus method, then do it magically as a last resort, unless the language feature has been introduced implicitly.

If `noStringPlus` is unset (in the default state), then you get the status quo, which is that `String` has plus. If there's an implicit value of `noStringPlus` and you use it, you get a warning (rather than somehow stripping `String` of plus in this context).

I thought it was an error that the feature check does implicit search before the option check (which is cheaper); but maybe that's so that intentionally ambiguous implicits can error out first.

I made this change, but `Definitions` in the reflect lib can't see regular compiler `ScalaSettings`. I got it turned off, but I'll experiment more tomorrow and send a PR to your branch for you to see. Or, I might not have time to push it very far. Maybe it's obvious to someone that this approach isn't tenable, etc.

dwijnand:
I'd be very interested in your patch, but I'm not sure if it's a good idea to treat language features differently according to whether you use the compiler flag or the import..

som-snytt:
Trying it out over there https://github.com/scala/scala/pull/5230 

FWIW, after trying it, I'm not sure the ergonomics of tri-state `noStringPlus` is worth it. Since I defined an internal `YnoStringPlus` flag, I considered using that to signal status quo (if disabled). I also considered that if transition is necessary, maybe use `-Xfuture`. A last gasp would be just that disabling `-future.noStringPos` gives the status quo.

adriaanm:
I've assigned this to the 2.12.1 milestone, see the [2.12.0-RC1 milestone description](https://github.com/scala/scala/milestone/47) for the reasoning. Short version: we must focus on getting 2.12.0 out ASAP :-) EDIT: moved it back because it would require adding a class, which we could not do in 2.12.1

som-snytt:
My follow-on improvement is now at https://github.com/scala/scala/pull/5235

That already feels like a long time ago. Yet it felt like the future.

adriaanm:
There can only be one, as they said in some mythical past -- could you gents (gently) settle which PR it shall be?

dwijnand:
I'm happy for @som-snytt to take the bridle here.

som-snytt:
It's not a subtle difference -- I used Dale's original commit for a language control, and the impl is quite different -- so I think it's a matter of fiat by SLIPpers.

som-snytt:
Is that like giving me enough rope to hang myself?

dwijnand:
Just looks like you have more of a handle of how to do this as a language import, which looks (semi) like a requirement.

Either way the compiler flag approach will always be available as a fall back.

PR #5226, opened 2016-06-12
If Range is sealed, it makes sense to have Range.Inclusive final.

Arneball:


som-snytt:
It also makes sense for `Inclusive` to disappear and let `isInclusive` be a class parameter, or param to a factory method and take `Inclusive` private.

cf `ParRange.apply`.

lrytz:
Agree that `Inclusive` could go, but also we that can also happen in some other PR.

SethTisue:
the logic "if A is sealed, and B extends A, then B might as well be final" logic here isn't actually sound logic. `sealed` means that the parent only has a fixed set of direct children. it doesn't mean the children can't be extended in turn

for example, over at https://github.com/scala/community-builds/issues/146, akka-http was anonymous-subclassing `Range.Inclusive` in order to override `toString`. there's no contradiction with `sealed`

note that I am not suggesting we revert this change, which I think is fine by a "hey, stay out of our implementation internals, and never subclass collections anyway" sort of argument, rather than the `sealed` reasoning

SethTisue:
though hmm maybe we should just revert it for RC2, since it could interfere with 2.11/2.12 cross building — we now have one example where it actually did

Ichoran:
Touching Range is a minefield.  Almost anything you do can raise ugly corner cases.  I would say if someone is extending Range.Internal for 2.11, either they are doing it wrong or the next person to touch that code will be doing it wrong.  I don't recommend reverting the change lightly.

som-snytt:
I just commented on the linked that the Range.toString was upgraded but maybe not backported? I guess that doesn't help people not running on 2.11.9. Offer a special "Range.toString upgrade pack"! Through subscription, of course.

sjrd:
The logic in this case was correct, because there is no `Range.Exclusive`. `Range` itself is the exclusive version. If you can't extend the exclusive version, or makes sense that you can't extend the inclusive version either.

The 2.11/2.12 cross-building argument is once again interpreted the wrong way. The goal of the extra compatibility was _not_ to more easily upgrade from 2.11 to 2.12. The goal was to make it easier to maintain a cross-compiling codebase. This change doesn't make that harder, as I can simply follow the 2.12 version, and it will also work with 2.11. One change that really made this harder was the deprecated of `Either` projections.

SethTisue:
> The goal of the extra compatibility was not to more easily upgrade from 2.11 to 2.12. The goal was to make it easier to maintain a cross-compiling codebase 

Totally agree, and your reminder is well taken.

I don't think I'd lost sight of that here, but I'll try not to get defensive about it :-)

SethTisue:
> maybe we should just revert it for RC2

I withdraw the thought/suggestion.

PR #5227, opened 2016-06-14
Topic/sip 23 redux [ci: last-only]

milessabin:
This is a WIP PR for discussion. Original work due to @adriaanm on [this branch](https://github.com/adriaanm/scala/tree/sip23) .. see his notes on the PR [here](https://github.com/adriaanm/scala/pull/21).

I've put the feature behind `-Yliteral-types` (also included in `-Xexperimental`).

The current status is that with the flag selectively enabled all tests pass.

With the flag turned on globally there are some expected failures (listed [here](https://github.com/milessabin/scala/blob/topic/sip-23-redux/test-failures.txt)) which are all largely unproblematic. The main issue is the change in types reported in type errors ... eg. for an argument type mismatch an error like `found "abc" expected Int` is reported instead of `found String expected Int`. It might be desirable to widen these types for error reporting.

Other than a lot of tieing up of loose ends and general tidying (see @adriaanm's TODO list at the top of the PR linked to above) the remaining work is the addition of a `ValueOf` type class and corresponding `valueOf` operator and more tests.

I would also like to explore merging the new `LiteralType` with `ConstantType` ... the introduction of the former will break existing macros which already inspect `ConstantType` and other things being equal it would be better to preserve source compatibility with libraries which would otherwise benefit from this change in their current state (eg. shapeless, refined, and things which depend on them). This would have the added advantage of not introducing new types into the public reflection API, which on balance I think would be a good thing.


adriaanm:
Added [ci: last-only] to avoid DOSing poor jenkins 😁

milessabin:
@adriaanm eep! Sorry!

Is there a list of useful Jenkins incantations somewhere so that I can avoid this in future?

adriaanm:
No worries -- scabot's menu is here: https://github.com/scala/scabot#pr-title-modifiers

fthomas:
I think the spec about singleton types needs to be updated. It currently reads:

```
A singleton type is of the form $p.$type, where $p$ is a path pointing to a value expected to
conform to scala.AnyRef. The type denotes the set of values consisting of null and the value
denoted by $p$ (i.e., the value $v$ for which v eq p).
```

But on this branch, the following now works:

``` scala
scala> def foo(x: Int): Option[x.type] = Some(x)
foo: (x: Int)Option[x.type]

scala> foo(4)
res13: Option[Int(4)] = Some(4)
```

although `x.type` does not conform to `scala.AnyRef`.

adriaanm:
We could drop LiteralType by having ConstantTypes only be inlined if they have the `@inline` annotation (automatically added by the type checker).

adriaanm:
See also my wip on constant folding at the type level instead of rewriting trees during typer. https://github.com/adriaanm/scala/tree/constantfold-in-type

milessabin:
@adriaanm I'm quite keen on your improved semantics for `asInstanceOf[<singleton type>]`.

Could we work around the unsound global cast in the compiler sources (and similar things elsewhere) by replacing, `hack.asInstanceOf[global.type]` with `Some(hack).asInstanceOf[Some[global.type]].get` or similar?

adriaanm:
We're likely not the only codebase that would be broken by this change (stricter semantics for `asInstanceOf[p.type]`), so we'll have to deprecate this before we can migrate to other semantics. See also https://issues.scala-lang.org/browse/SI-1503

som-snytt:
It's just semantics.

milessabin:
Fair enough.

Is it worth keeping the new semantics, but behind a (the same?) flag? Or deprecate the semantics, but only introduce the new ones when the deprecation cycle is complete?

Also, does "deprecate" imply "introduce a warning" in this context?

adriaanm:
I think it's enough to put the new semantics under a flag and warn users
about the changed semantics once they opt in. First, we need a broader
discussion and agree that an equality test is the correct semantics for a
singleton type test, though.

milessabin:
I'm going to close this for now and submit a fresh, hopefully close to final, PR shortly.

PR #5228, opened 2016-06-14
SI-9817 immutable queue `forall` and `exists` implementations

dpogretskiy:
https://issues.scala-lang.org/browse/SI-9817

@axel22

These are ones that i'm particularly interested in, since they create some GC pressure in my application and not as fast as they could be.
Not sure if it's safe to do so, since ordering is broken in case of side effecting predicates.

### DIFF
--- a/src/library/scala/collection/immutable/Queue.scala
+++ b/src/library/scala/collection/immutable/Queue.scala
@@ -84,6 +84,14 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
     else if (in.nonEmpty) new Queue(Nil, in.reverse.tail)
     else throw new NoSuchElementException("tail on empty queue")
 
+  /* This is made to avoid inefficient implementation of iterator. */
+  override def forall(p: A => Boolean): Boolean =
+    in.forall(p) && out.forall(p)
+
+  /* This is made to avoid inefficient implementation of iterator. */
+  override def exists(p: A => Boolean): Boolean =
+    in.exists(p) || out.exists(p)
+
   /** Returns the length of the queue.
    */
   override def length = in.length + out.length


src/library/scala/collection/immutable/Queue.scala
```diff
@@ -84,6 +84,11 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
     else if (in.nonEmpty) new Queue(Nil, in.reverse.tail)
     else throw new NoSuchElementException("tail on empty queue")
 
+  override /*IterableLike*/ def forall(p: A => Boolean): Boolean =
+    in.forall(p) && out.forall(p)
+  override /*IterableLike*/ def exists(p: A => Boolean): Boolean =
```
axel22: Please add an empty line between methods.

src/library/scala/collection/immutable/Queue.scala
```diff
@@ -84,6 +84,11 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
     else if (in.nonEmpty) new Queue(Nil, in.reverse.tail)
     else throw new NoSuchElementException("tail on empty queue")
 
+  override /*IterableLike*/ def forall(p: A => Boolean): Boolean =
```
axel22: It might generally be a good idea to add comments like this, but since we did not consistently do this in the past, I suggest to remove it.

axel22:
Thanks for your contribution!

Can you show an example where the previous implementation creates GC pressure and affects performance?

maxcom:
> Can you show an example where the previous implementation creates GC pressure and affects performance?

We have a wait list queue defined like this: `Queue[(ActorRef, Deadline)]`. Here is few examples where of forall/exists functions usage:

1) This code removes dead actors from queue:

```
    case Terminated(actor) if waitList.exists(_._1 == actor) ⇒
      waitList = waitList.filterNot(_._1 == actor)
```

2) Add actor to queue if it is not there already:

```
           if (waitList.forall(_._1!=sender())) {
              log.debug(s"No free helpers, put ${sender()} to wait list")
              waitList = waitList enqueue sender() -> deadline
              context.watch(sender())
            }
```

We detected this cases by profiling tool in our code base.

dpogretskiy:
@axel22
I added new line and removed comments.

Review please.

axel22:
Ok, since there are no tests, can you add a comment before each of these methods (not doc-comment) saying that they are added to avoid the inefficiency with the `iterator` implementation of the `Queue`.

After that, LGTM.

dpogretskiy:
Added comments.

Review please.

src/library/scala/collection/immutable/Queue.scala
```diff
@@ -84,6 +84,14 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
     else if (in.nonEmpty) new Queue(Nil, in.reverse.tail)
     else throw new NoSuchElementException("tail on empty queue")
 
+  /*This is made to avoid inefficient implementation of iterator*/
```
axel22: Please delimit `/*` and `This` with a space, and change to `iterator. */` (dot and a space).
We should have lint tests for this in the future, but for now we just ensure it during review.

axel22:
LGTM

dpogretskiy:
@axel22 sorry for delay, comments are properly formatted now.

SethTisue:
thanks @dpogretskiy, nice catch

PR #5229, opened 2016-06-14
Temporarily insource Scalacheck

retronym:
This is a temporary measure until we release Scala
2.12.0. It means we are able to release milestones,
and RCs of Scala without needing a public release of
Scalacheck.

After we release Scala 2.12.0, we'll revert to a binary
dependency on the standard Scalacheck.

This uncovered an existing bug in our SBT build related to
STARR being favoured on the classpath when performing test
discovery. This has be worked around by customizing the
`testLoader` setting.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -214,6 +214,19 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   Quiet.silenceIvyUpdateInfoLogging
 )
 
+// Two modifications are needed from the stock SBT configuration in order to exclude STARR
+// from the classloader that performs test discovery.
+//   - We make `isManagedVersion` hold by providing an explicit Scala version, in order to go into the desired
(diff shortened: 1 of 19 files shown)


retronym:
Review by @szeiger

szeiger:
Should we merge https://github.com/scala/scala/pull/5190 first (it only needs a final approval, everything is ready) and port it to 2.12.x? This would avoid conflicts in build.sbt and make CI run with sbt (since the ant build is not being updated here). Is sbt sufficient anyway? `bootstrap` still uses ant at the moment.

retronym:
@szeiger Yes, let's merge the SBT changes first, then merge to 2.12.x, then rebase this PR. I'll also add Ant changes.

adriaanm:
SBT changes in #5190 are now in 2.12.x via #5242

PR #5230, opened 2016-06-15
Review/string plus any language import [ci:last-only]

som-snytt:


som-snytt:
Neglected a check file, but not sure why `qual.symbol == null` when predef and preimports turned off in test/files/run/repl-no-imports-no-predef.scala

som-snytt:
https://github.com/scala/scala/pull/5235

PR #5231, opened 2016-06-15
SI-9585 hide auto-implicit conversions from scaladoc

gourlaysama:
This hides implicits conversions (and potential members obtained through
it) that converts a type into itself, because these conversions are
usually non-sensical. They are not completely removed, just placed behind
`-doc-implicits-show-all`, like other implicits deemed probably useless.

---

Consider the scaladoc for the following class:

```
object Box {
  implicit def anyToBox[T](t: T): Box[T] = new Box(t)
}
class Box[T](val t: T)
```

When looking for implicits members to add to class `Box`, it finds the
implicit conversion `anyToBox`, and applies it to itself to have an
implicit conversion to `Box[Box[T]]`, which brings a useless implicit
member `t: Box[T]`.

This commit makes scaladoc ignore any conversion from a type to itself
(even if type parameters differ) by default.

Using the (very useful) `tools/scaladoc-diff` script, I found that this
change removes the following conversion from the library doc:

```
Ensuring[A] to Ensuring[Ensuring[A]]
anytostringadd[A] to any2stringadd[anytostringadd[A]]
ArrowAssoc[A] to ArrowAssoc[ArrowAssoc[A]]
=:=[From,To] to =:=[From,To]
SearchImpl[A,Repr] to SearchImpl[A,SearchImpl[A,Repr]]
CollectionsHaveToParArray[C, T]  to CollectionsHaveToParArray[CollectionsHaveToParArray[C, T], T]
Ordered[A] to Ordered[Ordered[A]]
StringFormat[A] to StringFormat[StringFormat[A]]
```

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/model/ModelFactoryImplicitSupport.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/model/ModelFactoryImplicitSupport.scala
@@ -171,6 +171,20 @@ trait ModelFactoryImplicitSupport {
         return Nil
       }
 
+      if (!settings.docImplicitsShowAll && viewSimplifiedType.resultType.typeSymbol == sym) {
+        // If, when looking at views for a class A, we find one that returns A as well
+        // (possibly with different type parameters), we ignore it.
+        // It usually is a way to build a "whatever" into an A, but we already have an A, as in:
+        // {{{
+        //    object Box {
+        //      implicit def anyToBox[T](t: T): Box[T] = new Box(t)
+        //    }
+        //    class Box[T](val t: T)
+        // }}}
+        // We don't want the implicit conversion from Box[T] to Box[Box[T]] to appear.
+        return Nil
+      }
+
       // type the view application so we get the exact type of the result (not the formal type)
       val viewTree = result.tree.setType(viewSimplifiedType)
       val appliedTree = new ApplyImplicitView(viewTree, List(Ident("<argument>") setType viewTree.tpe.paramTypes.head))
--- a/test/scaladoc/run/t9585.check
+++ b/test/scaladoc/run/t9585.check
@@ -1,0 +1,6 @@
+warning: there was one feature warning; re-run with -feature for details
+any2stringadd[Box[T]]
+StringFormat[Box[T]]
+Ensuring[Box[T]]
+ArrowAssoc[Box[T]]
+Done.
--- a/test/scaladoc/run/t9585.scala
+++ b/test/scaladoc/run/t9585.scala
@@ -1,0 +1,25 @@
+import scala.tools.nsc.doc.model._
+import scala.tools.partest.ScaladocModelTest
+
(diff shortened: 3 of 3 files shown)


gourlaysama:
review by @janekdb @VladUreche

janekdb:
Do you think this warrants a note in the trait comment?

```
/**
 * This trait finds implicit conversions for a class in the default scope and creates scaladoc entries for each of them.
 *
 * Let's take this as an example:
 * {{{
 *    object Test {
 *      class A
 *
 *      class B {
 *        def foo = 1
 *      }
 *
 *      class C extends B {
 *        def bar = 2
 *        class implicit
 *      }
 *
 *      D def conv(a: A) = new C
 *    }
 * }}}
 *
 * Overview:
 * - scaladoc-ing the above classes, `A` will get two more methods: foo and bar, over its default methods
 * - the nested classes (specifically `D` above), abstract types, type aliases and constructor members are not added to
 * `A` (see makeMember0 in ModelFactory, last 3 cases)
 * - the members added by implicit conversion are always listed under the implicit conversion, not under the class they
 * actually come from (`foo` will be listed as coming from the implicit conversion to `C` instead of `B`) - see
 * `definitionName` in MemberImpl
 *
 * Internals:
 * TODO: Give an overview here
 */
```

Off topic, but does the trait comment code example read correctly? What is `class implicit` and where is `D` defined?

src/scaladoc/scala/tools/nsc/doc/model/ModelFactoryImplicitSupport.scala
```diff
@@ -171,6 +171,20 @@ trait ModelFactoryImplicitSupport {
         return Nil
       }
 
+      if (!settings.docImplicitsShowAll && viewSimplifiedType.resultType.typeSymbol == sym) {
```
VladUreche, line 174: Another pattern is adding covariance or contravariance for an invariant type parameter:

```
  case class C[T](t: T)
  implicit addCovariance[X, Y >: X](c: C[X]): C[Y] = c.asInstanceOf[C[Y]]
```

Do you also want to remove this case?

VladUreche:
@gourlaysama LGTM, except for a comment in the code, which you may want to either handle differently or document in the code.

@janekdb That's my fault for the missing example, sorry :( 
We can probably remove it altogether and come up with a simple example.

lrytz:
@gourlaysama ping for Vlad's comment (https://github.com/scala/scala/pull/5231#discussion_r68503231) - I'd say it makes sense that your patch also excludes this conversion, do you agree?

szeiger:
To expand on my "thumbs up": I think this PR is a change in the right direction so I plan to merge it either way before the end of the week but the suggested improvements would still make sense, possibly in a separate PR.

VladUreche:
If Antoine is not available, I can make a quick PR to address this last remaining issue.

szeiger:
@VladUreche Sounds good. I'll merge this one.

PR #5232, opened 2016-06-15
Escape triple quotes when printing multiline strings

andyscott:
When printing multiline literals containing triple quotes, the triple quotes are currently left intact and invalid code is emitted.

This addresses this issue by using string concatenation to insert escaped triple quotes.

Originally discovered in scala-exercises/scala-exercises#350.

### DIFF
--- a/src/reflect/scala/reflect/internal/Printers.scala
+++ b/src/reflect/scala/reflect/internal/Printers.scala
@@ -1055,7 +1055,16 @@ trait Printers extends api.Printers { self: SymbolTable =>
               val splitValue = x.stringValue.split(s"$LF").toList
               val multilineStringValue = if (x.stringValue.endsWith(s"$LF")) splitValue :+ "" else splitValue
               val trQuotes = "\"\"\""
-              print(trQuotes); printSeq(multilineStringValue) { print(_) } { print(LF) }; print(trQuotes)
+
+              def escapeSegment(value: String) {
+                printSeq(value.split(trQuotes, -1).toList) { print(_) } {
+                  print(trQuotes)
+                  print(" + \"\\\"\\\"\\\"\" + ")
+                  print(trQuotes)
+                }
+              }
+
+              print(trQuotes); printSeq(multilineStringValue)(escapeSegment) { print(LF) }; print(trQuotes)
             case _ =>
               // processing Float constants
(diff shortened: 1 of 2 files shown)


som-snytt:
I can't tell by looking at the code or the test, but I understand you to mean that this is wrong:

```
scala> showCode(Literal(Constant("hello\n\"\"\"world")))
res0: String =
"""hello
"""world"""
```

I'm also not sure what the original code intends by checking for embedded triple-quote and newline. The `size > 1` test is extraneous.

I'm guessing your result wasn't the pretty part of pretty-printing.

```
scala> showCode(Literal(Constant("hello\n\"\"\"world")))
res0: String =
"""hello
""" + "\"\"\"" + """world"""
```

Actually, the code originated [in sprinter](https://github.com/VladimirNik/sprinter/blob/master/src/main/scala/scala/sprinter/printers/PrettyPrinters.scala#L652) where the test for triples is inverted.

It looks like it had a flag to do multiline output, and that's what the special case is for. So the minimal fix is to make it check `!s.contains(tq)`.

You can also normalize tq by escaping just the third quote:

```
scala> val tq = """["]{3}""".r
tq: scala.util.matching.Regex = ["]{3}

scala> def f(s: String) = tq.replaceAllIn(s, """""\\"""")
f: (s: String)String

scala> f("""""\"hello""\", world""")
res0: String = ""\"hello""\", world
```

Also the multiline splitting should just use `s.lines`.

```
scala> showCode(Literal(Constant(""""hello"
     | world""")))
res4: String =
""""hello"
world"""

scala> showCode(Literal(Constant("hello\nworld")))
res5: String =
"""hello
world"""
```

andyscott:
I wasn't very sure of the surrounding logic in the printers, so I tried to keep my changes very isolated. (Also I had a brain fart and couldn't figure out how to cleanly escape triple quotes.)

Your solution looks much more reasonable!

som-snytt:
Thanks for finding the problem and locating it. Took me a while to get the context, so I'll follow up with a little refactor. I just noticed printers will emit octal constants, which are deprecated; code, like garden soil, likes to be turned over periodically.

andyscott:
Sure thing! Let me know if there is any way to help.

PR #5233, opened 2016-06-15
Generalize `Future.traverse` and `Future.sequence`

markus1189:
Allow `CanBuildFrom` to determine the resulting collection type, instead
of forcing it to be the same as the input.  This allows to do a
`Future.traverse` on a `Map[A,B]` and using `collection.breakOut` get
back a Future[Map[A,B]] again.

### DIFF
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -496,7 +496,7 @@ object Future {
   /** Simple version of `Future.traverse`. Transforms a `TraversableOnce[Future[A]]` into a `Future[TraversableOnce[A]]`.
    *  Useful for reducing many `Future`s into a single `Future`.
    */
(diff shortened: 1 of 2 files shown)


axel22:
Looks like it should work.
Can't say if there are any cases where it breaks type inference, though.

szeiger:
It looks safe to me. It should always be possible to infer the 3rd type parameter. My only concern with this change is that we're planning to move away from `CanBuildFrom` in the 2.13 collections, which means removing `collection.breakOut` as one of the few breaking changes.

markus1189:
@szeiger Thanks for picking this up. I am also fine with just closing this if it will be obsolete soon.

szeiger:
I suggest putting it on hold for a few months until we have a better idea of what the impact of the 2.13 collection changes will be on this PR.

SethTisue:
if anyone wants to pick this up, please (unless you're reading this after 2.13.0-M4 is out) take it up over in https://github.com/scala/collection-strawman

PR #5235, opened 2016-06-17
language.future.{noStringPlus,syntax}

som-snytt:
Something like that.

This has `-language:-future.noStringPlus` enable local implicit, where `String.+` is not installed but is tacked on when typecheck fails, on the conservative theory that extra implicit searches for string concat would hurt compilation time for status quo. But that theory is not tested, so maybe there's no reason to be conservative.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -726,13 +726,14 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       }
     }
 
-    /** Check whether feature given by `featureTrait` is enabled.
+    /** Check whether feature identified by `featureTrait` is enabled.
      *  If it is not, issue an error or a warning depending on whether the feature is required.
      *  @param  construct  A string expression that is substituted for "#" in the feature description string
      *  @param  immediate  When set, feature check is run immediately, otherwise it is run
      *                     at the end of the typechecking run for the enclosing unit. This
      *                     is done to avoid potential cyclic reference errors by implicits
      *                     that are forced too early.
+     *  @param  isAnOptInFeature  Indicates if the feature is opt-in or not
      *  @return if feature check is run immediately: true if feature is enabled, false otherwise
      *          if feature check is delayed or suppressed because we are past typer: true
      */
@@ -745,13 +746,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         def action(): Boolean = {
           def hasImport = inferImplicitByType(featureTrait.tpe, context).isSuccess
           def hasOption = settings.language contains featureName
-          val OK = hasImport || hasOption
-          if (!OK) {
+          val featureIsEnabled = hasImport || hasOption
+          if (!featureIsEnabled) {
             val Some(AnnotationInfo(_, List(Literal(Constant(featureDesc: String)), Literal(Constant(required: Boolean))), _)) =
               featureTrait getAnnotation LanguageFeatureAnnot
             context.featureWarning(pos, featureName, featureDesc, featureTrait, construct, required)
           }
-          OK
+          featureIsEnabled
         }
         if (immediate) {
           action()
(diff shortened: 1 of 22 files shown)


soc:
I'm sympathetic to the idea, but I think this piecemeal introduction of language options and settings does more harm than good:
- With this going in, we would have two vastly different ways of "doing the same thing" from a dev perspective depending on where exactly the `+` comes from. I think that's rather confusing.
- Is there any reasonable benefit derived from the ability to disable one `+` but not the other?
- I'd rather have one option that gets rid of all String-related `+` methods, combined with appropriate migration tools.
- (What's the reason this is named `noStringPlus` instead of `noStringAdd`? Is this some kind of distinction between `+` on `String` and `+` on `Any`?)

som-snytt:
@soc I think you're saying, why are there two mechanisms for `String.+` and the `Predef.any2stringadd` conversion? Besides historical reasons?

An implicit `Predef.string2anyadd` would be a no-op currently and unify the mechanisms under `future.noStringPlus`. Probably the folks who use these things know what they're doing and remain uncowed. Probably they are `-language:future._ -Yno-predef` anyway.

However, I haven't looked at whether the backend gets to do something efficient with `String.$concat` if there's an intervening conversion.

The operators that haven't been mentioned are the primitive `+(String)`. `(c: Char) + ""`.

It's called `noStringPlus` because the explorer who gets there first gets to name it. Also, `+` is `$plus`.

som-snytt:
It's a Sad, Sad, Sad, Sad World.

http://stackoverflow.com/questions/37948824/type-mismatch-when-using-operator

dwijnand:
`String.$concat` becomes `StringBuilder` calls in the backend: https://github.com/scala/scala/blob/3cbe3e89264ad63c9b86ea4962da553a24bfb128/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala#L218

Also I don't mind if it's called add or plus, I just want it to not compile :)

som-snytt:
@dwijnand My concern about how it's encoded was that an implicit wrapper might not be completely eliminated. Also, I learned on Scala-JS it's just plus. https://issues.scala-lang.org/browse/SI-9784

dwijnand:
Interesting.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -717,17 +717,18 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
 
-    /** Check whether feature given by `featureTrait` is enabled.
+    /** Check whether feature identified by `featureTrait` is enabled.
      *  If it is not, issue an error or a warning depending on whether the feature is required.
      *  @param  construct  A string expression that is substituted for "#" in the feature description string
      *  @param  immediate  When set, feature check is run immediately, otherwise it is run
      *                     at the end of the typechecking run for the enclosing unit. This
      *                     is done to avoid potential cyclic reference errors by implicits
      *                     that are forced too early.
+     *  @param  isAnOptInFeature  Indicates if the feature is opt-in or not
```
dwijnand, line 736: this isn't what the parameter is called anymore

dwijnand:
@som-snytt feel free to fold my commit if you'd like

som-snytt:
@dwijnand no way, it's your fault this ball got rolling! Besides, it already has a green check.

adriaanm:
I'm sorry that we were not able to give this the review/discussion attention it deserved, but I unfortunately don't think we can include this in 2.12.0. It's an exciting addition to the feature flags, but the design space is big enough to warrant some more exploration. As a compromise, I propose we do this exploration under a -Y flag in 2.12.x and, once baked, SIP it and expose it as a feature flag in 2.13.

adriaanm:
As I said, I'm open to a 2.12.x PR that introduces a -Y flag to experiment with this, but ultimately the full solution will require a SIP and a 2.13 target.

som-snytt:
OK, thanks, a lot of old girlfriends said that and I wish I'd listened when my wife said it would need a sip and a different target. I'll ask @dwijnand and the other haters of plus co-optation if this is worth pushing forward.

dwijnand:
I don't have the hate or bandwidth necessary atm.

Looked for someone on twitter https://twitter.com/dwijnand/status/834556490866389002

nafg:
Instead of removing `String#+`, how about moving it to an implicit in Predef, that seems like it enables a more natural way of disabling it.

PR #5236, opened 2016-06-17
[nomerge] Avoid triple-quoting triple quotes

som-snytt:
The boolean test for triples was inadvertently flipped.

Adds test for pretty printed multiline strings


lrytz:
LGTM

PR #5238, opened 2016-06-20
SI-5712 Support dependent argument types for constructor parameter lists

hamishdickson:
Issue details can be found at https://issues.scala-lang.org/browse/SI-5712

Implementation as per @adriaanm 's comment in the ticket. From reading the comments and the surrounding code, this looks like the correct solution to me (unless I've misunderstood something)

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -182,10 +182,7 @@ trait Namers extends MethodSynthesis {
       sym
     }
     def namerOf(sym: Symbol): Namer = {
-      val usePrimary = sym.isTerm && (
-           (sym.isParamAccessor)
-        || (sym.isParameter && sym.owner.isPrimaryConstructor)
-      )
+      val usePrimary = sym.isTerm && (sym.isParameter && sym.owner.isPrimaryConstructor)
 
       if (usePrimary) createPrimaryConstructorParameterNamer
       else innerNamer
--- a/test/files/pos/t5712.scala
+++ b/test/files/pos/t5712.scala
@@ -1,0 +1,9 @@
+class Global {
+  class Typer[T]
+}
+
+class Reifier[T](global: Global)(typer: global.Typer[T])
+
+object Test {
+  def mkReifier[T](global: Global)(typer: global.Typer[T]) = typer
+}


hamishdickson:
I would guess this is for @adriaanm or @retronym to review

hamishdickson:
Interesting - I have a couple of failures. I'll have a look at these and update my PR - sorry about that!

adriaanm:
No worries! Thanks for picking this up. Let us know how we can help out, though we have to prioritize getting 2.12.0-M5 and RC1 out at this point.

adriaanm:
I've conservatively assigned this to 2.12.1, though we can discuss after M5 if this can go into RC1. From personal experience, changes to Namers usually take a while to get right, so I expect we'll have to defer until after 2.12.0, and possibly put this under a flag.

hamishdickson:
Great - thank you! 😄 

I completely understand - I'm in no special rush to get this into M5, so that works for me

retronym:
Here's a regression:

```
scala> class C(s1: s2.type) { val s2: Int }
<console>:11: error: type mismatch;
 found   : C.this.s2.type (with underlying type Int)
 required: AnyRef
       class C(s1: s2.type) { val s2: Int }
                   ^
```

Oddly, the incorrect error is only issued with `s2` has an invalid type for a prefix (`Int` in this case.) If it has `String` as a type, we get the expected error message:

```
scala> class C(s1: s2.type) { val s2: AnyRef }
<console>:11: error: not found: value s2
       class C(s1: s2.type) { val s2: AnyRef }
                   ^
```

retronym:
Here's what seems to be a bug in the current implementation. Understanding this might help make the new impl correct.

```
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_91).

scala> { val s1 = "a"; class C(s1: String)(s2: s1.type); new C("b")("a") }
<console>:12: error: type mismatch;
 found   : String("a")
 required: String("b")
       { val s1 = "a"; class C(s1: String)(s2: s1.type); new C("b")("a") }
                                                                    ^


scala> { class A; val a1 = new A; val a2 = new A; class C(a1: A)(a2: a1.type); new C(a2)(a2) }
res12: AnyRef = C$1@6cf0e0ba

scala> { class A; val a1 = new A; val a2 = new A; class C(a1: A)(a2: a1.type); new C(a1)(a1) }
res13: AnyRef = C$1@36ebc363

scala> { class A; val a1 = new A; val a2 = new A; class C(a1: A)(a2: a1.type); new C(a2)(a1) }
<console>:12: error: type mismatch;
 found   : a1.type (with underlying type A)
 required: a2.type
       { class A; val a1 = new A; val a2 = new A; class C(a1: A)(a2: a1.type); new C(a2)(a1) }
                                                                                         ^
```

Seems like dependent constructor types are accidentally allowed if there is a val of the same name and type in scope.

hamishdickson:
Cool - thank you @retronym I'll start looking around that and see what comes out of it!

adriaanm:
See also https://issues.scala-lang.org/browse/SI-4663, btw.

dwijnand:
@retronym what part of those last 4 examples is unexpected? Looks all expected to me.

paulp:
> Seems like dependent constructor types are accidentally allowed if there is a val of the same name and type in scope.

@retronym That reminds me of [SI-7730](https://issues.scala-lang.org/browse/SI-7730), specifically this.

``` scala

scala> class A { def f(x: Any) = x match { case x: (T forSome { type T }) => x } }
<console>:11: error: not found: type T
       class A { def f(x: Any) = x match { case x: (T forSome { type T }) => x } }
                                                    ^

scala> class A { type T ; def f(x: Any) = x match { case x: (T forSome { type T }) => x } }
<console>:11: warning: abstract type pattern A.this.T[] is unchecked since it is eliminated by erasure
       class A { type T ; def f(x: Any) = x match { case x: (T forSome { type T }) => x } }
                                                               ^
defined class A
```

paulp:
@dwijnand In code like `class C(a1: A)(a2: a1.type)` the a1 in `a1.type` should refer to the a1 in the first parameter list. But it only compiles when there happens to be another a1 in scope. Yet if there IS another a1 in scope, it is compiled as if that a1 did refer to the one in the first parameter list.

dwijnand:
Err.... Oh.

paulp:
I've heard good things about Err-Oh types.

paulp:
I saw this sitting here and since I've run into this bug every couple weeks for several years I thought "what, can it be so simple?" So I took a look at Namers.

No, it is not so simple, not by a long ways.

I would go so far as to say that the extent to which multiple constructor parameter lists works in scalac is accidental. There's no way you're going to fix Namers for dependent constructor parameters with a small tweak to a predicate. It's missing the required logic, and much of the logic which IS there is doing ad hoc jiggery-pokery which will have to be unraveled.

hamishdickson:
oh yeah, basically agree 100% @paulp

> I would go so far as to say that the extent to which multiple constructor parameter lists works in scalac is accidental.

It's good you say that because I've tried (multiple times) to refactor Namers into something where I can work out how they currently work and have largely failed

@adriaanm @retronym given that any PR that might resolve SI-5238 is unlikely to  be along this approach, do you want me to close this PR?

hamishdickson:
Oops - didn't actually mean to close this PR (yet)

retronym:
i'd suggest to start by tackling the smaller plugin of fixing the bug that "Seems like dependent constructor types are accidentally allowed if there is a val of the same name and type in scope.". 

That should help get to grips with the relevant parts of namer/typer that will need to be cajoled into the bigger change proposed here.

I'm happy to help with that after the dust settles from the 2.12.0 release.

som-snytt:
There's a ticket where @lrytz says something about how these contexts are actually and erroneously nested, but I can't find it because thanks Jira.

Actor [talking into personal device]: Hi, Jira! Where is that ticket where Lukas said scopes for param lists is totally f--d up?
Jira [speaking as Siri]: Sorry, Lukas doesn't seem to use that sort of language.

SethTisue:
anyone interested in resuming work on this...?

milessabin:
@SethTisue Yes, I am.

adriaanm:

PR #5239, opened 2016-06-21
Doc: capitalize doesn't handle non-BMP characters

martijnhoekstra:
After a brief discussion on Gitter with @Ichoran, documenting the behaviour is preferable over changing it.

### DIFF
--- a/src/library/scala/collection/immutable/StringLike.scala
+++ b/src/library/scala/collection/immutable/StringLike.scala
@@ -137,6 +137,7 @@ self =>
 
   /** Returns this string with first character converted to upper case.
    * If the first character of the string is capitalized, it is returned unchanged.
+   * This method does not convert characters outside the Basic Multilingual Plane (BMP).
    */
   def capitalize: String =
     if (toString == null) null


soc:
If it doesn't do what it advertises, I'd rather just deprecate it.
Without taking a locale argument, capitalization is not well-defined anyway.

Ichoran:
It does what is advertised for almost everyone in almost all cases.  That
there are a couple of corner cases it doesn't catch is not reason to take
the functionality away from everyone.   (Not in a world where you can
randomly index into the middle of a multi-byte character in a string and
get nonsense.) The missing cases should be documented, though.

soc:
I'm not sure how you come to this conclusion. "Capitalization" without specifying a locale is not a well-defined operation.

I disagree on your broken-window stance here. Other things being broken does not excuse having additional, unrelated things being broken.

The implementation fails on basic it's-2016-and-this-shouldn't-be-broken tests.

martijnhoekstra:
Documenting current behaviour seems like a reasonable first step, that neither requires nor hinders further discussion on whether capitalizing without further context is a reasonable thing to support in the stdlib

adriaanm:
LGTM

PR #5240, opened 2016-06-21
SI-9336 Enable paste detect in jline

som-snytt:
When the next char is available immediately after a tab,
the tab is taken raw instead of invoking completion.


adriaanm:
LGTM

PR #5243, opened 2016-06-23
SI-8739 Typecheck args on failed apply

som-snytt:
This allows for completion of incomplete args.

The target method must not be overloaded.

Previously, arity checks would short-circuit.

Work done by retronym and wedens as linked on ticket.

### DIFF
--- a/test/files/presentation/t8739/src/Completions.scala
+++ b/test/files/presentation/t8739/src/Completions.scala
@@ -1,0 +1,9 @@
+
+package wedens
+
+object t8739 {
+  implicit class OptionW[T](opt: Option[T]) {
+    def cata[A](some: T => A, none: A) = opt.map(some) getOrElse none
+  }
+  Option("").cata(x => x./*!*/)
+}
--- a/src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala
@@ -72,7 +72,7 @@ trait SyntheticMethods extends ast.TreeDSL {
 
     val synthesizer = new ClassMethodSynthesis(
       clazz0,
-      newTyper( if (reporter.hasErrors) context makeSilent false else context )
+      newTyper( if (reporter.hasErrors) context.makeSilent(reportAmbiguousErrors = false) else context )
     )
     import synthesizer._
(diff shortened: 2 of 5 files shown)


som-snytt:
Review by @retronym

adriaanm:
I would suggest we move this to 2.12.x, but I'll let @retronym take a look first.

lrytz (approved):
I think this is a safe and very valuable addition for 2.11.x.

test/files/presentation/t8739/src/Completions.scala
```diff
@@ -0,0 +1,9 @@
+
+package wedens
+
+object t8739 {
+  implicit class OptionW[T](opt: Option[T]) {
+    def cata[A](some: T => A, none: A) = opt.map(some) getOrElse none
+  }
+  Option("").cata(x => x./*!*/)
```
retronym, line 8: The checkfile is a bit fragile because it lists all the members and extension members of `String`. Could you modify the test to use something more focussed, e.g. `class You { def completeMe: Any }`

retronym:
There is a risk with this sort of change that in a pathological case we could we could do a lot more work, e.g. through interplay with "expected" typechecking failures in implicit search. No example springs to mind at the moment, though.

I'm not sure that swallowing errors during argument typechecking is the right thing to do.

For example, we don't report the error for `symbol not found: doesNotExist` in:

```scala
package wedens

object t8739 {
  implicit class OptionW[T](opt: Option[T]) {
    def cata[A](some: T => A, none: A) = opt.map(some) getOrElse none
  }
  Option("").cata(x => x.doesNotExist)
}
```

IntelliJ's presentation compiler highlights that as an error.

retronym:
Here's a look at the [difference in post-typer ASTs](https://gist.github.com/retronym/3088fac3c3fb6d039b1e653b4420bce6) for a slightly extended version of that test case.

som-snytt:
You know you're in trouble when Jason starts making gists.

I'll move to 2.12.x since it's ancillary. I'll give it a review over the holiday, but probably more like Lunar New Year. Maybe Summer Solstice, which is a full day or two shy of a year?

Retronym: "I think this is a dangerous and mostly marginally useful change for the last chance we have to break things in 2.11.x."

PR #5244, opened 2016-06-23
Temporarily insource Scalacheck

retronym:
This is a temporary measure until we release Scala
2.12.0. It means we are able to release milestones,
and RCs of Scala without needing a public release of
Scalacheck. While we've never had to wait very long
for these in the past (Thanks, Rickard!) we'd like
to spare the maintainer some work betwen now and 2.12.0.

After we release Scala 2.12.0, we'll revert to a binary
dependency on the standard Scalacheck.

I've had to remove scalacheck as a SBT test framework
in our build. We don't use it directly as such (instead,
it is used indirectly through `partest --scalacheck`),
and it's test discovery (which we expect to return nothing)
fails after re-STARR-ing due to an unsolved problem with
SBT's testLoader including either STARR or sbt-launch.jar
on the classpath used to discover and spawn tests.

For the record, I tried the following to no avail:

```
// Two modifications are needed from the stock SBT configuration in order to exclude STARR
// from the classloader that performs test discovery.
//   - We make `isManagedVersion` hold by providing an explicit Scala version, in order to go into the desired
//     branch in `createTestLoader`
//   - We remove STARR from the classloader of the scala instance
def fixTestLoader = testLoader := {
  val s = scalaInstance.value
  val scalaInstance1 =
    new ScalaInstance(s.version, appConfiguration.value.provider.scalaProvider.loader(), s.libraryJar, s.compilerJar, s.extraJars, Some(s.actualVersion))
  assert(scalaInstance1.isManagedVersion)
  TestFramework.createTestLoader(Attributed.data(fullClasspath.value), scalaInstance1, IO.createUniqueDirectory(taskTemporaryDirectory.value))
}

```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -60,7 +60,6 @@ val scalaParserCombinatorsDep    = scalaDep("org.scala-lang.modules", "scala-par
 val scalaSwingDep                = scalaDep("org.scala-lang.modules", "scala-swing")
 val scalaXmlDep                  = scalaDep("org.scala-lang.modules", "scala-xml")
 val partestDep                   = scalaDep("org.scala-lang.modules", "scala-partest",              versionProp = "partest")
-val scalacheckDep                = scalaDep("org.scalacheck",         "scalacheck",                 scope = "it")
 
 // Non-Scala dependencies:
 val junitDep          = "junit"                  % "junit"           % "4.11"
@@ -562,6 +561,7 @@ lazy val junit = project.in(file("test") / "junit")
     fork in Test := true,
     libraryDependencies ++= Seq(junitDep, junitInterfaceDep, jolDep),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-v"),
+    testFrameworks -= new TestFramework("org.scalacheck.ScalaCheckFramework"),
     unmanagedSourceDirectories in Test := List(baseDirectory.value)
   )
 
@@ -642,7 +642,7 @@ lazy val test = project
   .settings(disablePublishing: _*)
   .settings(Defaults.itSettings: _*)
   .settings(
-    libraryDependencies ++= Seq(asmDep, partestDep, scalaXmlDep, scalacheckDep),
+    libraryDependencies ++= Seq(asmDep, partestDep, scalaXmlDep),
     libraryDependencies ++= {
       // Resolve the JARs for all test/files/lib/*.jar.desired.sha1 files through Ivy
       val baseDir = (baseDirectory in ThisBuild).value
@@ -659,6 +659,7 @@ lazy val test = project
     fork in IntegrationTest := true,
     javaOptions in IntegrationTest += "-Xmx2G",
     testFrameworks += new TestFramework("scala.tools.partest.sbt.Framework"),
+    testFrameworks -= new TestFramework("org.scalacheck.ScalaCheckFramework"),
     testOptions in IntegrationTest += Tests.Argument("-Dpartest.java_opts=-Xmx1024M -Xms64M -XX:MaxPermSize=128M"),
     testOptions in IntegrationTest += Tests.Argument("-Dpartest.scalac_opts=" + (scalacOptions in Compile).value.mkString(" ")),
     testOptions in IntegrationTest += Tests.Setup { () =>
--- a/test/files/scalacheck/parallel-collections/ParallelSeqCheck.scala
+++ b/test/files/scalacheck/parallel-collections/ParallelSeqCheck.scala
@@ -24,6 +24,7 @@ abstract class ParallelSeqCheck[T](collName: String) extends ParallelIterableChe
   def fromSeq(s: Seq[T]): CollType
 
   override def instances(vals: Seq[Gen[T]]): Gen[Seq[T]] = oneOf(
+    Gen.const(ofSize(vals, 1)),
     sized(
       sz =>
       ofSize(vals, sz)
@@ -74,7 +75,7 @@ abstract class ParallelSeqCheck[T](collName: String) extends ParallelIterableChe
     coll.patch(updateStart, coll, howMany)
   }
 
-  property("segmentLengths must be equal") = forAll(collectionPairsWithLengths) { case (s, coll, len) =>
+  property("segmentLengths must be equal") = forAllNoShrink(collectionPairsWithLengths) { case (s, coll, len) =>
     (for ((pred, ind) <- segmentLengthPredicates.zipWithIndex) yield {
       val slen = s.segmentLength(pred, if (len < 0) 0 else len)
       val clen = coll.segmentLength(pred, len)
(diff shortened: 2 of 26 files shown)


build.sbt
```diff
@@ -550,6 +549,7 @@ lazy val junit = project.in(file("test") / "junit")
     fork in Test := true,
     libraryDependencies ++= Seq(junitDep, junitIntefaceDep, jolDep),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-v"),
+    testFrameworks -= new TestFramework("org.scalacheck.ScalaCheckFramework"),
```
dwijnand, line 564: `-=`! 🎉

adriaanm:
LGTM

retronym:
Locally, I'm seeing a problen in `nan-ordering-test`, which amounts to this unexplained variation

```
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_91).

scala> def test(f1: Float, f2: Float) = (implicitly[Numeric[Float]].lt(f1, f2), f1 < f2)
test: (f1: Float, f2: Float)(Boolean, Boolean)

scala> test(Float.NaN, 0f)
res0: (Boolean, Boolean) = (false,false)
```

vs

```
Welcome to Scala 2.12.0-20160624-065100-be7ba72 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_91).

scala> def test(f1: Float, f2: Float) = (implicitly[Numeric[Float]].lt(f1, f2), f1 < f2)
test: (f1: Float, f2: Float)(Boolean, Boolean)

scala> test(Float.NaN, 0f)
res1: (Boolean, Boolean) = (true,false)
```

2.12.0-M4 has yet another idea:

```
scala> test(Float.NaN, 0f)
res2: (Boolean, Boolean) = (true,true)
```

adriaanm:
Are we compiling scalacheck with the right compiler?

retronym:
It is compiled with STARR in the SBT build. Not 100% sure about the Ant build.

retronym:
Test failure here:

```
! Parallel collections.ParVector[Int].padTos must be equal: Falsified after 1 passed tests.
> Labels of failing property: 
smaller
> ARG_0: (List(),ParVector(0),0)
> ARG_0_ORIGINAL: (WrappedArray(0),ParVector(0),1)
```

adriaanm:
I think this mix of compilers is the problem. It seems to me the proposal to go with a private published scalacheck from the bootstrap build for one commit is the easier solution.

adriaanm:
> It is compiled with STARR in the SBT build. Not 100% sure about the Ant build.

So could the test failures be explained by the fact that scalacheck is compiled with STARR, while the tests are compiled with quick? The failures could be a result of different super call semantics, right?

retronym:
This PR is not based on the trait-statics PR. I don't think there is a material difference between STARR and quick in super calls. But anything is possible...

adriaanm:
Right, but we have changed other things about the encoding, like #5085

retronym:
I believe that the scalacheck tests might been neutered since the change to remove unnecessary forwarders. They seem to rely on mixin in a `main` method from `trait Props` into each `object Test`. We had a similar problem with JUnit test cases which we added a special case to the compiler to support.

retronym:
#5207 is probably related, too.

sjrd:
> We had a similar problem with JUnit test cases which we added a special case to the compiler to support.

By "support" you mean display a warning? It's a good step, but I wouldn't call it "support".

We do provide actual support in a compiler plugin, though: https://github.com/scala-js/scala-2.12-junit-mixin-plugin
Still needs documentation, but basically it plugs in after mixin to add the necessary bridges for JUnit 4. And it's totally platform-agnostic, so it also works in Scala.js projects.

retronym:
RIght, I think the `NaN` problem makes sense: the implementation in `Ordering[Float]` will be incorrect until we recompile it with a new STARR that includes @lrytz fix for the `NaN` regression. But we compile the `nan-ordering` test case itself with quick. We could move that test to `disabled` and make a ticket to re-enable it after M5.

I just rebased this on top the latest 2.12.x, which includes the switch to SBT based PR validation.  That change does not skip locker, so at least the NaN problem should be gone.

retronym:
@sjrd We may yet need to add the forwarders back in as we're facing a non-neglible performance hit without them. 😦

retronym:
The reported failure in the `padTo` scalacheck is misleading. Scalacheck by default attempt to shrink the falsification. However, in that test, this independentally shrinks the collections in the random `(Seq[T], Seq[T], Int)` datapoint, which breaks the assumption that the sequences contain the same elements (as constructed in `collectionPairsWithLengths`. 

We could avoid the problem by using `forAllNoShrink`, rather than `forall`, or instead performing the collection copy (`fromSeq`) within the test itself.

To troubleshoot further, I'm using `forAllNoShrink` locally. Maybe the underlying problem is random test data is too large in some cases and is hitting an OOME.

retronym:
The root problem (that is being masked by the incorrect shrinking) is that this throws an unsupported operation error in 2.11 and 2.12

```
collection.parallel.immutable.ParVector(-3,-2,-1).padTo(4, 1)
```

```
java.lang.UnsupportedOperationException
    at scala.collection.parallel.immutable.Repetition.seq(package.scala:23)
    at scala.collection.parallel.immutable.Repetition.seq(package.scala:19)
    at scala.collection.parallel.ParSeqLike$class.patch(ParSeqLike.scala:222)
    at scala.collection.parallel.immutable.ParVector.patch(ParVector.scala:38)
    at scala.collection.parallel.ParSeqLike$class.padTo(ParSeqLike.scala:256)
    at scala.collection.parallel.immutable.ParVector.padTo(ParVector.scala:38
```

I'm not sure why that wasn't being caught by the Scalacheck tests before.

adriaanm:
Wow this just gets weirder and weirder. Maybe we've never been testing what we thought we were testing? 😳  Sounds like I should've been more keen on insourcing when you suggested it.

retronym:
One confounding factor is that I failed to insource exactly the same version of Scalacheck as we'd been using before. I'm going to rebase to include a commit to upgrade first and see what that tells me. The actual fix for the `padTo` test is easy enough, btw. But hopefully the step by step approach will help solve our epistemological crisis.

retronym:
Okay, the last piece of the puzzle was the fact that I had stubbed out the Scalacheck option parser and just used the defaults, which meant our custom option of `-minSuccessfulTests 5` was no longer added, and a more exhaustive search for the bugs found the "corner" case of a collection with 1-256 elements.

Details of my fix for this in the new commit comment

retronym:
@som-snytt DYR whether your addition of `-minSuccessfulTests 5` in https://github.com/scala/scala/commit/bf4466982854ffa8be57068752ea31daf7776eee#diff-5d9130d7e2cd71725aeac13615844703R3 was replicating existing behaviour or was a change?

som-snytt:
@retronym I haven't clicked that diff yet, but it would have to be "replicate existing".

Edit: Yes, the commented-out `main` shows the original test configuration.

adriaanm:
I'm running the ant build locally to verify it passes.

test/files/scalacheck/parallel-collections/ParallelSeqCheck.scala
```diff
@@ -248,29 +249,34 @@ abstract class ParallelSeqCheck[T](collName: String) extends ParallelIterableChe
...
-    val sdoub = s.padTo(len * 2, someValue)
-    val cdoub = coll.padTo(len * 2, someValue)
-    if (sdoub != cdoub) {
-      println("from: " + s)
-      println("and: " + coll)
-      println(sdoub)
-      println(cdoub)
+  property("padTos must be equal") = forAllNoShrink(collectionPairsWithLengths) { case (s, coll, len) =>
+//    assert(s.size == coll.size, (s.size, coll.size))
+    println("=" + coll.size + "/" + coll.getClass)
```
adriaanm: Using `ant` this gets to the console -- intentional?

retronym: Nope, that's stray debugging. Removed.

adriaanm:
LGTM!

milessabin:
I'm seeing a problem using a distribution built with `dist/mkPack` in conjunction with `scalaHome` in SBT. SBT has some fiddly logic for deciding which class loader to load classes for running tests, and the addition of ScalaCheck class files to `build/pack/lib/scala-partest-extras.jar` trips it up. The result is that tests compile, but when run we get a `ClassNotFoundException` for classes which would normally be found in SBT's `test-interface.jar`.

It would be possible to tweak the SBT class loading logic to accommodate this, but before doing that I'd like to be sure that we really do need `dist/mkPack` to be building and packaging `scala-partest-extras.jar`.

milessabin:
See the discussion [here](https://gitter.im/sbt/sbt-dev?at=578e470a4786ff0a48ae78b4) for more detail of how this manifests itself in SBT.

PR #5245, opened 2016-06-24
SI-4986 Allow trailing commas before newlines

dwijnand:
.. with updated specification.

I've taken a pragmatic approach of adding trailing comma support to:
- tuples
- argument and parameter groups, including for implicits, for functions, methods and constructors
- import selectors

I'm happy to add more, but I wanted to start conservative.

If need be, I'm happy to draft a SIP (or SLIP) for this.

Review by @adriaanm.

[edit]

SIP: https://github.com/scala/scala.github.com/pull/533

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
@@ -386,6 +386,14 @@ trait Scanners extends ScannersCommon {
           next copyFrom this
           this copyFrom prev
         }
+      } else if (token == COMMA) {
+        // SIP-27 Trailing Comma (multi-line only) support
+        // If a comma is followed by a new line & then a closing paren, bracket or brace
+        // then it is a trailing comma and is ignored
+        val saved = new ScannerData {} copyFrom this
+        fetchToken()
+        if (token != RPAREN && token != RBRACKET && token != RBRACE || !afterLineEnd())
+          this copyFrom saved
       }
 
 //      print("["+this+"]")
--- a/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
@@ -392,8 +392,11 @@ trait Scanners extends ScannersCommon {
         // then it is a trailing comma and is ignored
         val saved = new ScannerData {} copyFrom this
         fetchToken()
-        if (token != RPAREN && token != RBRACKET && token != RBRACE || !afterLineEnd())
-          this copyFrom saved
+        if (afterLineEnd() && (token == RPAREN || token == RBRACKET || token == RBRACE)) {
+          /* skip the trailing comma */
+        } else if (token == EOF) { // e.g. when the REPL is parsing "val List(x, y, _*,"
+          /* skip the trailing comma */
+        } else this copyFrom saved
       }
 
 //      print("["+this+"]")
--- a/test/files/neg/trailing-commas.check
+++ b/test/files/neg/trailing-commas.check
@@ -1,0 +1,130 @@
+trailing-commas.scala:10: error: illegal start of simple expression
+trait ArgumentExprs1 { f(23, "bar", )(Ev0, Ev1) }
+                                    ^
+trailing-commas.scala:10: error: ')' expected but '}' found.
+trait ArgumentExprs1 { f(23, "bar", )(Ev0, Ev1) }
+                                                ^
+trailing-commas.scala:11: error: illegal start of simple expression
+trait ArgumentExprs2 { f(23, "bar")(Ev0, Ev1, ) }
+                                              ^
+trailing-commas.scala:11: error: ')' expected but '}' found.
+trait ArgumentExprs2 { f(23, "bar")(Ev0, Ev1, ) }
+                                                ^
+trailing-commas.scala:12: error: illegal start of simple expression
+trait ArgumentExprs3 { new C(23, "bar", )(Ev0, Ev1) }
+                                        ^
+trailing-commas.scala:12: error: ')' expected but '}' found.
+trait ArgumentExprs3 { new C(23, "bar", )(Ev0, Ev1) }
+                                                    ^
+trailing-commas.scala:13: error: illegal start of simple expression
+trait ArgumentExprs4 { new C(23, "bar")(Ev0, Ev1, ) }
+                                                  ^
+trailing-commas.scala:13: error: ')' expected but '}' found.
+trait ArgumentExprs4 { new C(23, "bar")(Ev0, Ev1, ) }
+                                                    ^
+trailing-commas.scala:15: error: identifier expected but ')' found.
+trait Params1 { def f(foo: Int, bar: String, )(implicit ev0: Ev0, ev1: Ev1, ) = 1 }
+                                             ^
+trailing-commas.scala:15: error: ':' expected but '}' found.
+trait Params1 { def f(foo: Int, bar: String, )(implicit ev0: Ev0, ev1: Ev1, ) = 1 }
+                                                                                  ^
+trailing-commas.scala:16: error: identifier expected but ')' found.
+trait Params2 { def f(foo: Int, bar: String, )(implicit ev0: Ev0, ev1: Ev1, ) = 1 }
+                                             ^
+trailing-commas.scala:16: error: ':' expected but '}' found.
+trait Params2 { def f(foo: Int, bar: String, )(implicit ev0: Ev0, ev1: Ev1, ) = 1 }
+                                                                                  ^
+trailing-commas.scala:17: error: identifier expected but ')' found.
+trait ClassParams1 { final class C(foo: Int, bar: String, )(implicit ev0: Ev0, ev1: Ev1) }
+                                                          ^
+trailing-commas.scala:17: error: ':' expected but '}' found.
+trait ClassParams1 { final class C(foo: Int, bar: String, )(implicit ev0: Ev0, ev1: Ev1) }
+                                                                                         ^
+trailing-commas.scala:18: error: identifier expected but ')' found.
+trait ClassParams2 { final class C(foo: Int, bar: String)(implicit ev0: Ev0, ev1: Ev1, ) }
+                                                                                       ^
+trailing-commas.scala:18: error: ':' expected but '}' found.
+trait ClassParams2 { final class C(foo: Int, bar: String)(implicit ev0: Ev0, ev1: Ev1, ) }
+                                                                                         ^
+trailing-commas.scala:20: error: illegal start of simple expression
+trait SimpleExpr { (23, "bar", ) }
+                               ^
+trailing-commas.scala:20: error: ')' expected but '}' found.
+trait SimpleExpr { (23, "bar", ) }
+                                 ^
+trailing-commas.scala:22: error: identifier expected but ']' found.
+trait TypeArgs { def f: C[Int, String, ] }
+                                       ^
+trailing-commas.scala:22: error: ']' expected but '}' found.
+trait TypeArgs { def f: C[Int, String, ] }
+                                         ^
+trailing-commas.scala:23: error: identifier expected but ']' found.
+trait TypeParamClause { type C[A, B, ] }
+                                     ^
+trailing-commas.scala:23: error: ']' expected but '}' found.
+trait TypeParamClause { type C[A, B, ] }
+                                       ^
+trailing-commas.scala:24: error: identifier expected but ']' found.
+trait FunTypeParamClause { def f[A, B, ] }
+                                       ^
+trailing-commas.scala:24: error: ']' expected but '}' found.
+trait FunTypeParamClause { def f[A, B, ] }
+                                         ^
+trailing-commas.scala:26: error: identifier expected but ')' found.
+trait SimpleType { def f: (Int, String, ) }
+                                        ^
+trailing-commas.scala:26: error: ')' expected but '}' found.
+trait SimpleType { def f: (Int, String, ) }
+                                          ^
+trailing-commas.scala:27: error: identifier expected but ')' found.
+trait FunctionArgTypes { def f: (Int, String, ) => Boolean }
+                                              ^
+trailing-commas.scala:27: error: ')' expected but '}' found.
+trait FunctionArgTypes { def f: (Int, String, ) => Boolean }
+                                                           ^
+trailing-commas.scala:29: error: illegal start of simple pattern
+trait SimplePattern { val (foo, bar, ) = null: Any }
+                                     ^
+trailing-commas.scala:31: error: identifier expected but '}' found.
+trait ImportSelectors { import foo.{ Ev0, Ev1, } }
+                                               ^
+trailing-commas.scala:33: error: identifier expected but '}' found.
+trait Import { import foo.Ev0, foo.Ev1, }
+                                        ^
+trailing-commas.scala:35: error: illegal start of simple pattern
+trait ValDcl { val foo, bar, = 23 }
+                             ^
+trailing-commas.scala:35: error: '=' expected but '}' found.
+trait ValDcl { val foo, bar, = 23 }
+                                  ^
+trailing-commas.scala:36: error: illegal start of simple pattern
+trait VarDcl { var foo, bar, = 23 }
+                             ^
+trailing-commas.scala:36: error: '=' expected but '}' found.
+trait VarDcl { var foo, bar, = 23 }
+                                  ^
+trailing-commas.scala:37: error: illegal start of simple pattern
+trait VarDef { var foo, bar, = _ }
+                             ^
+trailing-commas.scala:37: error: '=' expected but '}' found.
+trait VarDef { var foo, bar, = _ }
+                                 ^
+trailing-commas.scala:38: error: illegal start of simple pattern
+trait PatDef { val Foo(foo), Bar(bar), = bippy }
+                                       ^
+trailing-commas.scala:38: error: '=' expected but '}' found.
+trait PatDef { val Foo(foo), Bar(bar), = bippy }
+                                               ^
+trailing-commas.scala:45: error: illegal start of simple expression
+trait SimpleExpr2 { (23, ) }
+                         ^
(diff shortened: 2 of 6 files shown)


spec/04-basic-declarations-and-definitions.md
```diff
@@ -586,8 +586,8 @@ Def                ::=  ‘def’ FunDef
 FunDef             ::=  FunSig [‘:’ Type] ‘=’ Expr
 FunSig             ::=  id [FunTypeParamClause] ParamClauses
 FunTypeParamClause ::=  ‘[’ TypeParam {‘,’ TypeParam} ‘]’
-ParamClauses       ::=  {ParamClause} [[nl] ‘(’ ‘implicit’ Params ‘)’]
-ParamClause        ::=  [nl] ‘(’ [Params] ‘)’}
+ParamClauses       ::=  {ParamClause} [[nl] ‘(’ ‘implicit’ Params [‘,’] ‘)’]
+ParamClause        ::=  [nl] ‘(’ [Params] [‘,’] ‘)’}
```
densh: If you put trailing comma in the definition of `Params`, you wouldn't need to duplicate it everywhere. 

```
Params             ::=  Param {‘,’ Param} [‘,‘]
```

spec/05-classes-and-objects.md
```diff
@@ -672,8 +672,8 @@ TmplDef           ::=  `class' ClassDef
 ClassDef          ::=  id [TypeParamClause] {Annotation}
                        [AccessModifier] ClassParamClauses ClassTemplateOpt
 ClassParamClauses ::=  {ClassParamClause}
-                       [[nl] `(' implicit ClassParams `)']
-ClassParamClause  ::=  [nl] `(' [ClassParams] ')'
+                       [[nl] `(' implicit ClassParams [‘,’] `)']
+ClassParamClause  ::=  [nl] `(' [ClassParams] [‘,’] ')'
```
densh: The same thing here.
```
ClassParams       ::=  ClassParam {`,' ClassParam} [‘,‘]
```

mpociecha:
Was it discussed somewhere? Personally I don't like this change [edit: as much as others do]. I don't think benefits are significant and it can be error-prone e.g. in case of overloaded methods (I mean the typical users).

densh:
For consistency patterns, types and type parameters should all support trailing commas if those exist for expressions and method parameters. (See `SimplePattern`, `SimpleType` and `TypeParamClause` in the syntax summary.)

densh:
Here is the full list of rules that contain comma-separated parts:
- [ ] `ValDcl`, `VarDcl`, `VarDef` via `ids`
- [ ] `Type` via `FunctionArgTypes`
- [ ] `SimpleType`, `TypeArgs` via `Types`
- [ ] `Expr`, `ResultExpr` via `Bindings`
- [x] `SimpleExpr1`, `ArgumentExprs` via `Exprs`
- [ ] `SimplePattern` via `Patterns`
- [ ] `TypeParamClause`, `FunTypeParamClause`
- [x] `ParamClause`, `ParamClauses` via `Params`
- [x] `ClassParamClause`, `ClassParamClauses` via `ClassParams`
- [ ] `ImportExp`
- [x] `ImportSelector`
- [ ] `PatDef`

sjrd:
As much as I would welcome this change, I agree with @mpociecha that this needs some kind of open discussion. It should actually be a SIP, at least insofar as having a document explaining the change and why it is important.

gabrieljones:
How about make commas optional the same way semicolons are optional.

cvogt:
I wish we had this forever. Makes it easier to move code around and comment out lines. +1

som-snytt:
SI-8739 is about autocompletion of `f(g, x)` at `f(x => x. ,`, where the comma is currently needed because of arity check for applicability. I wonder if there may be an edge case in function args.

dwijnand:
@mpociecha

> Was it discussed somewhere? Personally I don't like this change. I don't think benefits are significant and it can be error-prone e.g. in case of overloaded methods (I mean the typical users).

and @sjrd

> As much as I would welcome this change, I agree with @mpociecha that this needs some kind of open discussion. It should actually be a SIP, at least insofar as having a document explaining the change and why it is important.

For context, Scala used to support trailing commas in the past. Then they got removed without discussion. Since then, over the years, there have been a number of short conversations about it, on scala-internals, scala-language and more recently on Gitter and Reddit, but nothing extensive or complete, and typically mixed up with discussing other language or library changes in Scala. Additionally, I'm sure there have also been discussions around this language feature in other languages.

I knew that that there was a high probability that a SIP would be necessary, my approach was to provide a (hopefully) complete-looking PR for early feedback and start an open discussion here.

But **I will draft up a SIP for this**, such that it can be fully discussed and cleanly documented for the future.

Also, @mpociecha, please could you provide an example where this is error-prone, e.g. in case of overloaded methods?

dwijnand:
@densh

> For consistency patterns, types and type parameters should all support trailing commas if those exist for expressions and method parameters. (See SimplePattern, SimpleType and TypeParamClause in the syntax summary.)

I would be perfectly happy to add ubiquitous, consistent support to trailing commas in every case, but I chose instead this subset as it's the minimum that I feel much much more strongly about, and think would have a greater chance of success (ie. inclusion).

But I will make sure to add this as to the SIP as a potential alternative.

dwijnand:
@gabrieljones

> How about make commas optional the same way semicolons are optional.

Personally I find that a very interesting idea, but it's much more of an aggressive change then my little thing here.

dwijnand:
@som-snytt

> SI-8739 is about autocompletion of f(g, x) at f(x => x. ,, where the comma is currently needed because of arity check for applicability. I wonder if there may be an edge case in function args.

Thanks for the link, I'll circle back to that after working on the SIP.

mpociecha:
@dwijnand Thanks for the explanation. Now "return" in the title makes sense.

Honestly I didn't mean anything more than:

```
scala> :paste

def foo(bar: Int) = bar
def foo(bar: Int, baz: Int) = bar + baz
foo(1,)

foo: (bar: Int)Int <and> (bar: Int, baz: Int)Int
foo: (bar: Int)Int <and> (bar: Int, baz: Int)Int
res0: Int = 1
```

The question is whether someone forgot to add the second param and the compiler didn't report this due to the new behaviour, or they came up with the idea of adding commas at the end in case they would like to add another param in the future. ;) Yes, I know that I'm exaggerating, but there are people who write

```
def foo = {
  something
}
```

instead of

```
def foo = something
```

"in case they'd want to add something else to such a method in the future". Hence, nothing will surprise me. ;)

dwijnand:
@mpociecha Yep, I see what you mean. But I'm not sure I would classify that as "error-prone" - that line of thinking would mean that all overloading should be removed: you might've written `bar(1)` when you meant to write `bar(1, 2)`.

mpociecha:
Yes, that's why I admitted that I'm exaggerating.

som-snytt:
The completion example is not an exaggeration. But the effects might be subtle. (The proposed fix for that example says the method can't be overloaded, FWIW.)

adriaanm:
Happy to discuss here, but we can't schedule this for 2.12.0, so I'm assigning it to the WIP milestone. Also, yes, this will need to be SIPped.

dwijnand:
I've drafted and pull requested the SIP for this: https://github.com/scala/scala.github.com/pull/533.

xeno-by:
I thought that you guys should know that we've reviewed Dale's proposal during our last SIP meeting and decided that it sounds interesting. In the coming weeks/months, we'll be working with Dale to comprehensively review his SIP. Feel free to join the discussion at https://github.com/scala/scala.github.com/pull/533.

milessabin:
A backport of this to 2.11.x for 2.11.9 would be useful as an aid to migration.

dwijnand:
The SIP for this ([SIP-27](http://docs.scala-lang.org/sips/pending/trailing-commas.html)) was accepted in the November SIP committee meeting.

I've updated the implementation to reflect the final design decision.

I've also written a pos and a neg test to test at least one instance of trailing commas in each of the relevant elements of the syntax grammar.

Could anyone spare some time to review this and give my their feedback please?

Also, could this have it's `on-hold` label removed and moved into the `2.12.2` milestone please?

som-snytt:
I think this is your latest tweak, inspired by your comment about `_*`:
```
$ ./build/pack/bin/scala
Welcome to Scala 2.12.2-20161210-162902-143b068 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_111).

scala> val vs = List(1,2,
     | 3,
     | )
vs: List[Int] = List(1, 2, 3)

scala> val vs = List(1,2,_*,
     | )
<console>:11: error: missing parameter type for expanded function ((x$1: <error>) => x$1.$times)
       val vs = List(1,2,_*,
                         ^
<console>:11: warning: postfix operator * should be enabled
by making the implicit value scala.language.postfixOps visible.
This can be achieved by adding the import clause 'import scala.language.postfixOps'
or by setting the compiler option -language:postfixOps.
See the Scaladoc for value scala.language.postfixOps for a discussion
why the feature should be explicitly enabled.
       val vs = List(1,2,_*,
                          ^

scala> apm@mara:~/clones/review/dwijnand$ scala

scala> val vs = List(1,2,_*,
     | )
<console>:2: error: illegal start of simple expression
)
^
```

dwijnand:
Are you pointing out a regression?

I think this is the comparison to make and the output is the same:

```scala

scala> val vs = List(1, 2, 3, _*)
<console>:11: error: missing parameter type for expanded function ((x$1: <error>) => x$1.$times)
       val vs = List(1, 2, 3, _*)
                              ^
<console>:11: warning: postfix operator * should be enabled
by making the implicit value scala.language.postfixOps visible.
This can be achieved by adding the import clause 'import scala.language.postfixOps'
or by setting the compiler option -language:postfixOps.
See the Scaladoc for value scala.language.postfixOps for a discussion
why the feature should be explicitly enabled.
       val vs = List(1, 2, 3, _*)
                               ^
```

```scala
Welcome to Scala 2.12.2-20161211-000404-7966edd (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> val vs = List(1, 2, 3, _*,
     | )
<console>:11: error: missing parameter type for expanded function ((x$1: <error>) => x$1.$times)
       val vs = List(1, 2, 3, _*,
                              ^
<console>:11: warning: postfix operator * should be enabled
by making the implicit value scala.language.postfixOps visible.
This can be achieved by adding the import clause 'import scala.language.postfixOps'
or by setting the compiler option -language:postfixOps.
See the Scaladoc for value scala.language.postfixOps for a discussion
why the feature should be explicitly enabled.
       val vs = List(1, 2, 3, _*,
                               ^
```

som-snytt:
Sorry, I just had a minute to try it yesterday and that was the first thing I tried. I mean after building it. In a pattern, `pattern3` does `checkWildStar`. I'll make some time today.

lrytz:
Nice that the implementation is so small!

This PR should come with an update to the spec. I think the spec changes in the SIP are still incomplete:
  - Trailing commas are only allowed if followed by a `nl`. I don't actually see how the current spec allows `nl`s between parameters / arguments, is it incomplete?
  - `SimpleExpr1 ::= ... | ‘(’ [Exprs] ‘)’ | ...` also supports trailing commas
  -  [edit] Similarly, `SimpleType`, `ImportSelectors`, `SimplePattern`

lrytz:
a few more cases for the tests: https://gist.github.com/lrytz/e10b166ffbed2e47348a7ef8cd072fd9

dwijnand:
Yep, I will be updating the SIP document as well as updating the relevant parts of the spec (in this PR).

I was hoping for some implementation and/or testing review/feedback.
On advice from Adriaan I'll try and convert these partests to parser unit tests.

lrytz:
> convert these partests to parser unit tests.

For the neg tests, it would be valuable to keep the partest version to check the error messages.

som-snytt:
This works, though I'm not sure TC in args in pat valdef (should be same as case) is in the tests:
```
  val List(x, y, _*,
  ) = 42 :: 17 :: Nil
```
but in REPL
```
scala> val List(x,y,
     | ) = 42 :: 17 :: Nil
x: Int = 42
y: Int = 17

scala> val List(x,y,_*,
<console>:1: error: bad simple pattern: bad use of _* (a sequence pattern must be the last pattern)
val List(x,y,_*,
               ^
```
Is that because of how REPL does a probing parse to see if the input is complete?

som-snytt:
I remember raising the issue that TC could confuse completion; but maybe here, REPL just has to supply the newline when parsing? Or:

```
scala> import reflect.runtime._,universe._,scala.tools.reflect._
import reflect.runtime._
import universe._
import scala.tools.reflect._

scala> val tb = currentMirror.mkToolBox()
tb: scala.tools.reflect.ToolBox[reflect.runtime.universe.type] = scala.tools.reflect.ToolBoxFactory$ToolBoxImpl@2b0061b7

scala> tb.parse("val List(x, y, z @ _*,\n")
scala.tools.reflect.ToolBoxError: reflective compilation has failed:

bad simple pattern: bad use of _* (a sequence pattern must be the last pattern)
illegal start of simple pattern
  at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl$ToolBoxGlobal.throwIfErrors(ToolBoxFactory.scala:320)
  at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl$ToolBoxGlobal.parse(ToolBoxFactory.scala:295)
  at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl.$anonfun$parse$1(ToolBoxFactory.scala:421)
  at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl$withCompilerApi$.liftedTree2$1(ToolBoxFactory.scala:359)
  at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl$withCompilerApi$.apply(ToolBoxFactory.scala:359)
  at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl.parse(ToolBoxFactory.scala:418)
  ... 31 elided

scala> tb.parse("val List(x, y, z @ _*,\n) = 42 :: 17 :: Nil")
res1: tb.u.Tree =
{
  <synthetic> <artifact> private[this] val x$3 = {
    <synthetic> <artifact> val x$2 = 42;
    {
  <synthetic> <artifact> val x$1 = 17;
  Nil.$colon$colon(x$1)
}.$colon$colon(x$2)
  }: @scala.unchecked match {
    case List((x @ _), (y @ _), (z @ (_)*)) => scala.Tuple3(x, y, z)
  };
  val x = x$3._1;
  val y = x$3._2;
  val z = x$3._3;
  ()
}

```
The wildstar is misprinted unprettily, as the extra parens are not allowed:
```
scala> val List(x, y, (_)*) = 42 :: 17 :: Nil
<console>:1: error: bad simple pattern: use _* to match a sequence
val List(x, y, (_)*) = 42 :: 17 :: Nil
                   ^
```
But what a great error message.

som-snytt:
Maybe just this tweak:
```
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -1974,6 +1974,7 @@ self =>
         def isCloseDelim = in.token match {
           case RBRACE => isXML
           case RPAREN => !isXML
+          case COMMA  => true
           case _      => false
         }
```
for
```
scala> val List(x, y, _*,
     | )
     | =
     | 17 :: 42 :: Nil
x: Int = 17
y: Int = 42
```

dwijnand:
Thanks Som, I'll give that a shot, and make sure we test that case with a REPL test.

dwijnand:
@som-snytt Thank you for that test case. Unfortunately your suggested tweak broke test/files/neg/t1878.scala and test/files/neg/t5702-neg-bad-and-wild.scala. I've come up with an alternative fix: adding EOF as a way to ignore trailing commas. I hope that is acceptable.

@lrytz Thank you for those interesting cases. I've adapted and added them to the pos tests I had.

@adriaanm Can this lose its "on-hold" tag and be moved into the 2.12.2 queue please? I'd like to do as much as I can on my side to try and get this in in time (and with sufficient time for satisfactory review). Thank you.

som-snytt:
For the record, one of my resolutions this new year is to say fewer unhelpful things. Thanks for the return of anything glorious.

dwijnand:
Any chance this can get into the 2.12.2 queue, please?

adriaanm:
🎉 

Thank you, @dwijnand for your hard work, and everyone involved for your feedback and your patience! Feel free to let your commas trail henceforth.

adriaanm (approved):

adriaanm:
PS: one more thing :trollface: this will need a spec update ;-)

som-snytt:
Probably "organize imports" will eliminate trailing commas. Teams with a mix of IDE users and real programmers will have to tread carefully. It might be the last straw for groups who worked through tabs vs spaces.

dwijnand:
Thanks @adriaanm for the review and merge!

TODO:
- [ ] PR an update to the SIP document with the final implementation details
- [ ] PR an update the spec
- [ ] PR intellij-scala
- [ ] PR scalariform
- [ ] PR scala.meta
- [ ] PR scalaparse (lihaoyi/fastparse)

adriaanm:
TODO list moved to https://github.com/scala/scala-dev/issues/299

milessabin:
FYI: this has been [merged](https://github.com/typelevel/scala/blob/typelevel-readme/notes/2.12.1.md#trailing-commas-pull5245-dwijnand) in Typelevel Scala 2.12.1.

xeno-by:
Hi everyone! Can we have this pull request backported to 2.11.10 please?

Judging by the history of 2.10.x, I think that people are going to use 2.11.x for quite a while. Therefore, I believe that trailing comma support in 2.11.x will be quite important for cross-compilation reasons.

gabrieljones:
How hard would it be to backport to 2.10 while we're at it?

adriaanm:
Sorry, no, we're not going to backport this to 2.10 or 2.11. (As always, it's not about the cost of doing the backport, which is quite low compared to the ongoing cost of such changes. For example: the impact on the tooling eco-system, user experience, required support,... Also, note that this has nothing to do with our commercial side, since we'd happily generate more paying support tickets there ;-))

milessabin:
It'll be in Typelevel Scala 2.11.10.

PR #5246, opened 2016-06-24
SI-4826 Retain javadoc comments in scaladoc [ci: last-only]

jodersky:
Double-star comments in java sources currently do not show up in scaladoc output, although their related classes and members do. This is an inconvenience for multi-language projects (talking about Spark specifically).

The changes in this PR retain documentation comments in java sources and include them in scaladoc.
To achieve this, the following parts of the compiler are modified:
- java lexer and parser: overridable methods are added to allow scaladoc to consume java doc comments. These changes were greatly inspired by the existing way scaladoc hooks into the scala parser.
- namer and typer: ignore certain checks to validate ASTs generated by the java parser. The main reason for these changes is that scaladoc requires symbols to be assigned to java members, and the easiest way to achieve that is by using the existing namer and typer. If this solution is too hacky, an alternative would be to add an extra traversal (or even phase?) to java ASTs that does symbol assignment solely for java.

I know that I'm quite late to the 2.12 party, however since these changes are quite marginal and mostly affect scaladoc, I based this PR on the 2.12.x branch.

cc @VladUreche


src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -1654,7 +1654,7 @@ trait Namers extends MethodSynthesis {
         if (sym hasAnnotation NativeAttr)
           sym resetFlag DEFERRED
         else {
-          if (!symbolAllowsDeferred && ownerRequiresConcrete) fail(AbstractVar)
+          if (!symbolAllowsDeferred && ownerRequiresConcrete && !sym.isJava) fail(AbstractVar)
```
jodersky: I'm not sure about this, please see my comments in the PR description regarding the namer and typer

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4876,6 +4876,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           case LookupInaccessible(sym, msg) => issue(AccessError(tree, sym, context, msg))
           case LookupNotFound               =>
             inEmptyPackage orElse lookupInRoot(name) match {
+              // ignore wildcard bodies inserted by the Java parser
+              case NoSymbol if context.unit.isJava && name == termNames.WILDCARD => tree
```
jodersky: I'm not sure about this, please see my comments in the PR description regarding the namer and typer

VladUreche:
Glad to see such a great addition to Scaladoc! Thank you @jodersky!

The failure in the CI system is genuine:

```
!! 33 - run/javacomments.scala                    [compilation failed]
##### Log file '/home/jenkins/workspace/scala-2.12.x-validate-test/test/scaladoc/run/javacomments-run.log' from failed test #####

javacomments.scala:23: error: value contains is not a member of scala.tools.nsc.doc.base.comment.Inline
    assert(method.comment.get.short.contains(Tag))
                                    ^
one error found
```

(https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/2404/console)

I think `extractCommentText` may help there, since it turns the comment into a string so you can check the marker presence: https://github.com/scala/scala/blob/2.12.x/src/scaladoc/scala/tools/partest/ScaladocModelTest.scala#L169-L182

For the two comments, I'm surprised you need to change the namer and the typer. Could you minimize the cases that required the new code? 

For the merge: @adriaanm, are you okay with merging later in the the release pipeline?

jodersky:
thanks for the tip, I'll fix the tests. (turns out that my local tests actually didn't run. `sbt partest --srcpath scaladoc --grep javacomments` should do the trick)

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jodersky: This kind of bugs me, however java interfaces fail without it.

jodersky:
Regarding the other changes, the namer modification was used to support java interfaces: it seems that they are translated into a scala trait with companion object, however the abstract method is copied to the object as well without a dummy definition.

The typer changes were necessary to make type-checking pass on any ASTs originating from java source. The java parser replaces method bodies and other parameter values with wildcards (_).
In my opinion, strictly type-checking java is not necessary however (as I found out the hard way) the type checker actually forces symbol assignment that is otherwise done lazily by the namer. If there is an easy way to resolve symbols without the type checker, I'm all for using that instead! @adriaanm do you have any suggestions?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
VladUreche: [getting back to this now, please forgive the delay]
What I don't understand is why adding doc comments changes things. Normally, Java code goes through namer and typer when mixed Scala-Java projects are used. Therefore, retaining the doc comments should not change anything at all.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
VladUreche: Same here.

VladUreche: What failure mode do you get?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jodersky: [no worries, I guess everyone was extremely busy with the latest milestone release]

I totally agree with your reasoning; that's also why I was quite puzzled as to why the java doc comments weren't showing up after my initial changes to the parser. Nevertheless, after some more searching, I found a method `def createJavadoc: Boolean = false` in `Global`, which appears to control if the typer is run on java.
`Global#GlobalPhase`:

``` scala
def cancelled(unit: CompilationUnit) = {
  // run the typer only if in `createJavadoc` mode
  val maxJavaPhase = if (createJavadoc) currentRun.typerPhase.id else currentRun.namerPhase.id
  reporter.cancelled || unit.isJava && this.id > maxJavaPhase
}
```

It was added by Donna Malayeri in 2010, however I was unable to find any use of it in the current code base. My thinking was that `createJavadoc` must have been lost in refactoring over time (I did not do an extensive search with git though, since scalac has changed quite a bit since then and I'm only just getting familiar with the code base). Overriding it in `ScaladocGlobal` enters the typer and causes the errors that require the changes to namer and typer.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
jodersky: I get the compilation errors like these 

``` scala
JavaClazz.java:20: not found: value _
    public void foo() {}
                ^
```

They appear at all places where the java parser inserts a `_` as placeholder body, for instance in method and field bodies and parameters.

Although the error message doesn't show the underscores, they can be seen by printing the ast right after parsing.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jodersky: Ok, so after some more testing I can observe that this specific change is only required to fix errors that occur with java _interfaces_.

The typer changes are still required for all java constructs though.

MechCoder:
I've been bitten by this before. I'm really looking forward to this feature getting merged into scala.

jodersky:
@adriaanm, do you have any pointers on what I could do with the namer and typer?

adriaanm:
Sorry for the delay! Slowly emerging from M5 and #5141. Will take a look today^W this week :-) -- got sucked into some weird stuff over at my fields pr.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: I wonder why `!sym.owner.isClass` for a java definition in a class?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
adriaanm: Can we just skip type checking the body all together for java definitions since we know it'll always be `_`?

jodersky: they are also inserted as parameter values, i.e. `void foo(int x)` gets parsed to `def foo(x: Int = _): Unit`.

The change in the typer was the best place I could find. However this is also the first time I'm working with the compiler's internals so maybe there is a better, more generic place to add the wildcard-ignore.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jodersky: I'm not sure I understand what you mean.

When omitting this change, I get an

```
only classes can have declared but undefined members
        int abstractCommentedMethod);
```

error. However, the error only occurs when a method in an interface has a doc comment. Without a comment it compile just fine. I'll have to revisit that.

SethTisue:
this is https://issues.scala-lang.org/browse/SI-4826

src/compiler/scala/tools/nsc/typechecker/Typers.scala
adriaanm: Weird, the `ValDef` for a param should have an `EmptyTree` for its rhs.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: I would expected `ownerRequiresConcrete` to be `false` because the methods owner is a class but not a module class (the owner of members in objects) nor an anonymous class.

jodersky:
updated the title, thanks Seth

src/compiler/scala/tools/nsc/typechecker/Typers.scala
jodersky: the java parser (`compiler/scala/tools/nsc/javac/JavaParser.scala`) does the following substitutions that result in wildcards:

```
method body -> blankExpr
formalParam -> varDecl -> blankExpr
blankExpr -> Ident(WILDCARD)
```

As a quick test I replaced the wildcard with an `EmptyTree` and, behold, the previous compilation errors in the type checker and namer go away. Could it be so simple? I assume the wildcard placeholder was there for a reason. I'll run some tests and update the PR if they are successful.

jodersky:
@VladUreche , @adriaanm , what do you think of the latest commit? By changing the parsed java AST to contain `EmptyTrees` instead of `Ident(WILDCARD)`, the most intrusive changes into the namer and typer can be avoided.
Let me know if this is a better solution and I'll squash this PR for review

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -111,7 +111,7 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
     def arrayOf(tpt: Tree) =
       AppliedTypeTree(scalaDot(tpnme.Array), List(tpt))
 
-    def blankExpr = Ident(nme.WILDCARD)
+    def blankExpr = EmptyTree
```
jodersky, line 114: this avoids the previous changes in namer and typer

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2248,9 +2248,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         }
 
       if (meth.isClassConstructor && !isPastTyper && !meth.owner.isSubClass(AnyValClass)) {
-        // At this point in AnyVal there is no supercall, which will blow up
-        // in computeParamAliases; there's nothing to be computed for Anyval anyway.
-        if (meth.isPrimaryConstructor)
+        // There are no supercalls for AnyVal or constructors from Java sources, which
+        // would blow up in computeParamAliases; there's nothing to be computed for them
+        // anyway.
+        if (meth.isPrimaryConstructor && !meth.isJava)
```
jodersky: only change required in typer, but it makes sense IMO

VladUreche:
LGTM, thanks for all the work Jakob! I'll let @SethTisue or @adriaanm merge, since they take care of the release process.

jodersky:
thanks for the review, Vlad. I squashed and reworded the commits.

adriaanm:
Good stuff, thanks @jodersky!

MechCoder:
Thanks for the amazing work @jodersky !! Cannot wait to test this feature in scala 2.12

adriaanm:
Looks like there's an issue [running this on akka with the community build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/456/consoleFull):

```
[akka] java.lang.NullPointerException
[akka]  at scala.tools.nsc.doc.ScaladocSyntaxAnalyzer$ScaladocJavaUnitScanner.skipBlockComment(ScaladocAnalyzer.scala:236)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.skipComment(JavaScanners.scala:598)
[akka]  at scala.tools.nsc.doc.ScaladocSyntaxAnalyzer$ScaladocJavaUnitScanner.skipComment(ScaladocAnalyzer.scala:244)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.fetchToken(JavaScanners.scala:475)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.nextToken(JavaScanners.scala:249)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.init(JavaScanners.scala:864)
```

jodersky:
oh no! :disappointed:  Let me check it out.

adriaanm:
How about something like?

``` diff
--- i/src/scaladoc/scala/tools/nsc/doc/ScaladocAnalyzer.scala
+++ w/src/scaladoc/scala/tools/nsc/doc/ScaladocAnalyzer.scala
@@ -216,7 +216,7 @@ abstract class ScaladocSyntaxAnalyzer[G <: Global](val global: G) extends Syntax

   class ScaladocJavaUnitScanner(unit: CompilationUnit) extends JavaUnitScanner(unit) {

-    private val docBuffer: StringBuilder = new StringBuilder
+    private var docBuffer: StringBuilder = _
     private var inDocComment = false
     private var docStart: Int = 0
     private var lastDoc: DocComment = null
@@ -224,6 +224,11 @@ abstract class ScaladocSyntaxAnalyzer[G <: Global](val global: G) extends Syntax
     // get last doc comment
     def flushDoc(): DocComment = try lastDoc finally lastDoc = null

+    override def init(): Unit = {
+      docBuffer = new StringBuilder
+      super.init()
+    }
+

```

given the stacktrace

```
[akka] java.lang.NullPointerException
[akka]  at scala.tools.nsc.doc.ScaladocSyntaxAnalyzer$ScaladocJavaUnitScanner.skipBlockComment(ScaladocAnalyzer.scala:236)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.skipComment(JavaScanners.scala:598)
[akka]  at scala.tools.nsc.doc.ScaladocSyntaxAnalyzer$ScaladocJavaUnitScanner.skipComment(ScaladocAnalyzer.scala:244)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.fetchToken(JavaScanners.scala:475)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.nextToken(JavaScanners.scala:249)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaScanner.init(JavaScanners.scala:864)
[akka]  at scala.tools.nsc.javac.JavaScanners$JavaUnitScanner.<init>(JavaScanners.scala:870)
```

that should hopefully do it

jodersky:
Hey @adriaanm  , I was just going to propose the same thing ;)
The other solution I was exploring is to use early initialization 

``` scala
 class ScaladocJavaUnitScanner(unit: CompilationUnit) extends {
    private val docBuffer: StringBuilder = new StringBuilder
  } with JavaUnitScanner(unit)
```

but I think that's more confusing than just using a var

adriaanm:
Agreed! Plus, early vals are on their way out ;-)

jodersky:
Do you have any info on the top of your head on why a val does not work here though?
I know that there are lots of `newAnalyzer()` etc factory methods that are overridden in different use cases, however I would assume that everything is initialized before parsing actually starts

Btw, should I open a new PR for the changes?

jodersky:
@adriaanm , have you tested this? I'm getting typer assertion errors now when compiling Akka. It could be however that I have a messed up build (I'm spoofing my locally published scala SNAPSHOT as M5 to workaround dependencies in Akka)

adriaanm:
I have not tested it yet, so I can't confirm type checking issues. The
community build I have in the queue should tell us more by tomorrow
morning.

The val behavior is weird but as expected. Initialization begins with
statements in the subclass body (among which init()), before calling the
super constructor, which would initialize the val in this case.

jodersky:
> The community build I have in the queue should tell us more by tomorrow morning.

Does that mean the fix you proposed is already in? Let me know where I can find the community build so I can help debug in case the issue is not solved

adriaanm:
I have a temporary commit in my fields-lazies branch on adriaanm/scala to
check it out (https://github.com/adriaanm/scala/commit/e4babfb51aef7faac64b13d9768f449b2203bda0). On phone right now. The build is queued on
scala-ci.typesafe.com (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/460/console -- it's testing my new PR, so failures could indicate many things :-))

adriaanm:
Which assert failures are you seeing?

jodersky:
I get 

```
ava.lang.AssertionError: assertion failed: 
  <static> private[actor] val cellOffset: Long = _
     while compiling: /home/jodersky/projects/akka/akka-actor/src/main/java/akka/actor/AbstractActorRef.java
        during phase: typer
     library version: version 2.12.0-20160720-172442-6b99dfe
    compiler version: version 2.12.0-20160720-172442-6b99dfe
```

```
last tree to typer: TypeTree(class Long)
       tree position: line 10 of /home/jodersky/projects/akka/akka-actor/src/main/java/akka/actor/AbstractActorRef.java
            tree tpe: Long
              symbol: (final abstract) class Long in package scala
   symbol definition: final abstract class Long extends AnyVal (a ClassSymbol)
      symbol package: scala
       symbol owners: class Long
           call site: object AbstractActorRef in package actor in package actor

     7 
     8 final class AbstractActorRef {
     9     final static long cellOffset;
    10     final static long lookupOffset;
    11 
    12     static {
    13         try {
    at scala.tools.nsc.Global.assert(Global.scala:229)
    at scala.tools.nsc.typechecker.MethodSynthesis$MethodSynth$BaseGetter.validate(MethodSynthesis.scala:382)
```

But if you look at the date of the compiler versions being used, I'm pretty sure my environment is messed up (I'm mixing dependencies with M5 and my snapshot build on top of latest 2.1.2.x). Since you already have a community build scheduled, it's probably easiest to await the results before trying to get akka and all its dependencies to work with a "pure" snapshot build.

adriaanm:
Weird, it looks like it's trying to synthesize a getter for a java-defined val? That's not right because we should not get here for java units: https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/typechecker/Namers.scala#L648-L654

adriaanm:
But yes, let's see what dbuild says tomorrow.

adriaanm:
Seems like some more checks need to be suppressed for Java-defined units/members. Didn't see any crashes though :-)

```
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:30: not found: type R
[akka] [error]     public R apply(I i) throws Exception;
[akka] [error]            ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:30: not found: type I
[akka] [error]     public R apply(I i) throws Exception;
[akka] [error]                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:30: only classes can have declared but undefined members
[akka] [error]     public R apply(I i) throws Exception;
[akka] [error]              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:48: not found: type R
[akka] [error]     public R apply(I1 i1, I2 i2) throws Exception;
[akka] [error]            ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:48: not found: type I1
[akka] [error]     public R apply(I1 i1, I2 i2) throws Exception;
[akka] [error]                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:48: not found: type I2
[akka] [error]     public R apply(I1 i1, I2 i2) throws Exception;
[akka] [error]                           ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:48: only classes can have declared but undefined members
[akka] [error]     public R apply(I1 i1, I2 i2) throws Exception;
[akka] [error]              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:63: not found: type T
[akka] [error]     public boolean defined(T t);
[akka] [error]                            ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:63: only classes can have declared but undefined members
[akka] [error]     public boolean defined(T t);
[akka] [error]                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:80: not found: type T
[akka] [error]     public boolean defined(T t, U u);
[akka] [error]                            ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:80: not found: type U
[akka] [error]     public boolean defined(T t, U u);
[akka] [error]                                 ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:80: only classes can have declared but undefined members
[akka] [error]     public boolean defined(T t, U u);
[akka] [error]                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:94: not found: type I
[akka] [error]     public void apply(I i) throws Exception;
[akka] [error]                       ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:94: only classes can have declared but undefined members
[akka] [error]     public void apply(I i) throws Exception;
[akka] [error]                 ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:110: not found: type I1
[akka] [error]     public void apply(I1 i1, I2 i2) throws Exception;
[akka] [error]                       ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:110: not found: type I2
[akka] [error]     public void apply(I1 i1, I2 i2) throws Exception;
[akka] [error]                              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:110: only classes can have declared but undefined members
[akka] [error]     public void apply(I1 i1, I2 i2) throws Exception;
[akka] [error]                 ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:128: not found: type I1
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3) throws Exception;
[akka] [error]                       ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:128: not found: type I2
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3) throws Exception;
[akka] [error]                              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:128: not found: type I3
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3) throws Exception;
[akka] [error]                                     ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:128: only classes can have declared but undefined members
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3) throws Exception;
[akka] [error]                 ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:148: not found: type I1
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3, I4 i4) throws Exception;
[akka] [error]                       ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:148: not found: type I2
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3, I4 i4) throws Exception;
[akka] [error]                              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:148: not found: type I3
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3, I4 i4) throws Exception;
[akka] [error]                                     ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:148: not found: type I4
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3, I4 i4) throws Exception;
[akka] [error]                                            ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:148: only classes can have declared but undefined members
[akka] [error]     public void apply(I1 i1, I2 i2, I3 i3, I4 i4) throws Exception;
[akka] [error]                 ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:158: only classes can have declared but undefined members
[akka] [error]     public void apply() throws Exception;
[akka] [error]                 ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/FI.java:171: only classes can have declared but undefined members
[akka] [error]     public boolean defined(Object o);
[akka] [error]                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/PFBuilder.java:20: constructor PFBuilder is defined twice;
[akka] [error]   the conflicting constructor PFBuilder was defined at line 15:20
[akka] [error]   public PFBuilder() {
[akka] [error]                   ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/java/akka/japi/pf/UnitPFBuilder.java:23: constructor UnitPFBuilder is defined twice;
[akka] [error]   the conflicting constructor UnitPFBuilder was defined at line 18:20
[akka] [error]   public UnitPFBuilder() {
[akka] [error]                       ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/scala/akka/actor/AbstractFSM.scala:115: akka.japi.pf.FI.UnitApply2[S,S] does not take parameters
[akka] [error]     onTransition(transitionHandler(_: S, _: S))
[akka] [error]                                   ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/scala/akka/japi/pf/CaseStatements.scala:16: value defined is not a member of akka.japi.pf.FI.Predicate
[akka] [error]   override def isDefinedAt(o: F) = predicate.defined(o)
[akka] [error]                                              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/scala/akka/japi/pf/CaseStatements.scala:18: value apply is not a member of akka.japi.pf.FI.Apply[P,T]
[akka] [error]   override def apply(o: F) = apply.apply(o.asInstanceOf[P])
[akka] [error]                                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/scala/akka/japi/pf/CaseStatements.scala:24: value defined is not a member of akka.japi.pf.FI.Predicate
[akka] [error]   override def isDefinedAt(o: F) = predicate.defined(o)
[akka] [error]                                              ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/scala/akka/japi/pf/CaseStatements.scala:26: value apply is not a member of akka.japi.pf.FI.UnitApply[P]
[akka] [error]   override def apply(o: F) = apply.apply(o.asInstanceOf[P])
[akka] [error]                                    ^
[akka] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-401ab93b08c69081e4e00a27fd34cbb1d7360b40/akka-actor/src/main/scala/akka/routing/ConsistentHashing.scala:160: Cannot find LogSource for Class[_ <: akka.routing.ConsistentHashingRoutingLogic] please see ScalaDoc for LogSource for how to obtain or construct one.
[akka] [error]   private lazy val log = Logging(system, getClass)
[akka] [error]                                 ^
[akka] [info] No documentation generated with unsuccessful compiler run
```

jodersky:
Great, at least no more NPEs.
I experimented a bit more and figured it also crashes on documented fields (although that's probably quite a rare occurrence in the java world). I already have a fix available for that on my branch and will address the above errors next. I'll also update the test suite to something more complete.

How should I contribute the work? I can open a new WIP pr until everything is ready.

adriaanm:
(Apologies for delay. Retreated into the High Sierra wilderness for a long
weekend ;-). A WIP pr would be great. I can help with running community
build and provide pointers where helpful.)

PR #5249, opened 2016-06-28
Override `.slice` in ArrayOps to use arraycopy.

dimatkach:
This makes it ~10x faster when copying large chunks arround.
My benchmark:

  def bm(duration: Long)(f: => Unit): Int = {
     val end = System.currentTimeMillis + duration
     var count = 0
     while(System.currentTimeMillis < end) {
       f
       count += 1
     }
     count
  }

  def measure(seconds: Int)(f: => Unit) = (1 to seconds).map { _ => bm(1000)(f) }.sum / seconds

  val array = scala.util.Random.alphanumeric.take(1000).toArray

  measure(20) { array.slice(100, 500) }
  // ~5 million
  measure(20) { scala.collection.WrappedArray(array).slice(100, 500) }
  // ~300K

### DIFF
--- a/src/library/scala/collection/mutable/ArrayOps.scala
+++ b/src/library/scala/collection/mutable/ArrayOps.scala
@@ -41,6 +41,15 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
     if (l > 0) Array.copy(repr, 0, xs, start, l)
   }
 
+  override def slice(from: Int, until: Int): Array[T] = {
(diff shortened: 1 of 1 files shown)


dimatkach:
Ok, I must be doing something wrong here - CI failed once, so, I fixed it, and added a commit to the branch, but now it won't run, because it says that it "found earlier commit marked failed". Can't squash it either for some reason ("origin" is always "ahead"). I guess, I'll just close it, and create a new one instead.

PR #5250, opened 2016-06-28
Override `.slice` in ArrayOps to use arraycopy.

dimatkach:
This makes it ~10x faster when copying large chunks arround.
My benchmark:

```
def bm(duration: Long)(f: => Unit): Int = {
   val end = System.currentTimeMillis + duration
   var count = 0
   while(System.currentTimeMillis < end) {
     f
     count += 1
   }
  count
}

def measure(seconds: Int)(f: => Unit) = (1 to seconds).map { _ => bm(1000)(f) }.sum / seconds

val array = scala.util.Random.alphanumeric.take(1000).toArray

measure(20) { array.slice(100, 500) }
// ~5 million
measure(20) { scala.collection.WrappedArray(array).slice(100, 500) }
// ~300K
```

### DIFF
--- a/src/library/scala/collection/mutable/ArrayOps.scala
+++ b/src/library/scala/collection/mutable/ArrayOps.scala
@@ -44,9 +44,9 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
   override def slice(from: Int, until: Int): Array[T] = {
      val lo = math.max(from, 0)
      val hi = math.min(math.max(until, 0), repr.length)
-     val size = math.max(hi-lo, 0)
+     val size = math.max(hi - lo, 0)
      val result = java.lang.reflect.Array.newInstance(elementClass, size)
-     if(size > 0) {
+     if (size > 0) {
       Array.copy(repr, lo, result, 0, size)
      }
      result.asInstanceOf[Array[T]]
--- a/src/library/scala/collection/mutable/ArrayOps.scala
+++ b/src/library/scala/collection/mutable/ArrayOps.scala
@@ -41,6 +41,17 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
     if (l > 0) Array.copy(repr, 0, xs, start, l)
   }
 
+  override def slice(from: Int, until: Int): Array[T] = {
+     val lo = math.max(from, 0)
+     val hi = math.min(math.max(until, 0), repr.length)
+     val size = math.max(hi-lo, 0)
+     val result = java.lang.reflect.Array.newInstance(elementClass, size)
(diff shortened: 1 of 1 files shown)


dimatkach:
@Ichoran @axel22 would you guys care to take a look at this?

src/library/scala/collection/mutable/ArrayOps.scala
```diff
@@ -41,6 +41,17 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
     if (l > 0) Array.copy(repr, 0, xs, start, l)
   }
 
+  override def slice(from: Int, until: Int): Array[T] = {
+     val lo = math.max(from, 0)
+     val hi = math.min(math.max(until, 0), repr.length)
+     val size = math.max(hi-lo, 0)
```
axel22: We usually delimit operators like `-` with whitespaces.

src/library/scala/collection/mutable/ArrayOps.scala
```diff
@@ -41,6 +41,17 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
...
+     val result = java.lang.reflect.Array.newInstance(elementClass, size)
+     if(size > 0) {
```
axel22: Convention is to put space after `if`.

axel22:
Thanks for your contribution! LGTM on my side.

Please verify if there is a test for `slice` on arrays, and add it if there isn't.

lrytz:
@dimatkach for test coverage, there is at least https://github.com/scala/scala/blob/2.12.x/test/files/run/slices.scala. If you'd like to add to it, that test could be converted to a JUnit test (see https://github.com/scala/scala/tree/2.12.x/test/junit/scala/collection).

dimatkach:
@lrytz  I am not sure I understand - why/where do I need to "add" it, if it's already there? :) 
(not trying to be difficult, I am just new, and don't know the right way). 
I know that it does get run by the CI, because I got several failures on this change, and had to fix them before it passed.

lrytz:
Oh, i was just pointing out the existing test, but you already know about it then. My question is if you think it is comprehensive enough. If you know of more corner cases that should be tested, you could add them to the test, and if you work on the test you could consider converting it to JUnit. JUnit tests run more efficiently.

Also, it would be nice to address Alex's code style remarks.

For setting up your local dev env, see https://github.com/scala/scala#get-ready-to-contribute.

dimatkach:
@lrytz @axel22  I have fixed the style issues you pointed out. Re. testing, slices.scala seems to be thorough enough, I could not think of any other cases that would be useful.

axel22:
LGTM

PR #5251, opened 2016-06-28
Emit trait method bodies in statics [rebase of #5177]

adriaanm:
Rebased #5177 on 2.12.x after merging #5229.

Original description:
In 2.12.0-M4 / #5003, we changed the trait encoding to do away with trait implementation classes, in favour of leaving trait methods in place as interface default methods.

However, we've since discovered that we can no longer precisely target a particular statically resolved method with `invokespecial` due to impedance mismatch between JVM and Scala semantics.

This pull request emits a pair of methods for each concrete trait method. The default method remains, with the same signature as before. However, it now forwards to a same-named static method (with an explicit self param) that contains the method code.

This encoding allows us to precisely target a particular method with `invokestatic`.

This is binary incompatible, and will require bootstrapping of the modules.

(More details to follow)


adriaanm:
- [x] Bootstrap: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/487/console
   repo_ref   9901daf
   SCALA_VER_SUFFIX   -M4-9901daf
   STARR_REF  9901daf
   PARTEST_VER    1.0.17
   publishToSonatype  yes
- [x] [Community build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/) once 2.12.0-M4-7d51b3f is on maven (scala_ref=scala/scala.git#7d51b3f, repo_user =adriaanm,repo_ref=2.12.x-pr5251). Community build config in [adriaanm/community-builds:2.12.x-pr5251](https://github.com/scala/community-builds/compare/2.12.x...adriaanm:2.12.x-pr5251?expand=1)
- [x] update starr to 2.12.0-M4-9901daf, partest to 1.0.17

adriaanm:
Retrying bootstrap without building scalacheck and a [new partest](https://github.com/scala/scala-partest/compare/v1.0.17) (1.0.17 -- not tagged yet) that doesn't depend on scalacheck: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/487/console

adriaanm:
bootstrap success :)

adriaanm:
Running community build with sonatype-staging added to the dbuild virtual repo: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/436/console

adriaanm:
Drad, it failed with that same 

```
[sbt-testng-interface] java.lang.NoSuchMethodError: scala.Product.$init$(Lscala/Product;)V
[sbt-testng-interface]  at scala.util.parsing.json.JSONArray.<init>(Parser.scala:103)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.page.IndexScript.scala$tools$nsc$doc$html$page$IndexScript$$$anonfun$2(IndexScript.scala:54)
[sbt-testng-interface]  at scala.collection.TraversableLike.scala$collection$TraversableLike$$$anonfun$3(TraversableLike.scala:234)
[sbt-testng-interface]  at scala.collection.immutable.Map$Map3.foreach(Map.scala:176)
[sbt-testng-interface]  at scala.collection.TraversableLike.map(TraversableLike.scala:234)
[sbt-testng-interface]  at scala.collection.AbstractTraversable.map(Traversable.scala:104)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.page.IndexScript.<init>(IndexScript.scala:29)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.page.IndexScript$.apply(IndexScript.scala:153)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.HtmlFactory.generate(HtmlFactory.scala:104)
[sbt-testng-interface]  at scala.tools.nsc.doc.html.Doclet.generateImpl(Doclet.scala:23)
[sbt-testng-interface]  at scala.tools.nsc.doc.doclet.Generator.generate(Generator.scala:23)
[sbt-testng-interface]  at scala.tools.nsc.doc.DocFactory.generate$1(DocFactory.scala:128)
[sbt-testng-interface]  at scala.tools.nsc.doc.DocFactory.document(DocFactory.scala:131)
[sbt-testng-interface]  at xsbt.Runner.run(ScaladocInterface.scala:26)

```

retronym:
> Drad, it failed with that same

Trying [again](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/437/console) without `-Dskip.locker` (IIRC that was what fixed this last time).

adriaanm:
You recall correctly. I forgot -- time to fast track that effort to resolve Scala from maven.

retronym:
LGTM

adriaanm:
Ok, so next step is to push a commit that sets starr.version to 2.12.0-M4-9901daf and run the release build (in the same vein as 41d5b9f)

adriaanm:
TODO: https://github.com/scala/scala-partest/issues/67

adriaanm:
```
[info] >>> The dbuild result is------------: SUCCESS (project rebuilt ok)
```

lrytz:
🎈

smarter:
> In 2.12.0-M4 / #5003, we changed the trait encoding to do away with trait implementation classes, in favour of leaving trait methods in place as interface default methods.
However, we've since discovered that we can no longer precisely target a particular statically resolved method with invokespecial due to impedance mismatch between JVM and Scala semantics.

Dotty doesn't use the scheme implemented in the PR and I'm having trouble finding a test case where this makes a difference (I couldn't find anything like that in the tests added in this PR), could you give me an example @retronym  ?

smarter:
I'm still confused (please help!), but my best guess, is that this PR is about code like this:
```scala
trait A { def foo = "A" }
trait B extends A { override def foo = "B" }
class C extends A with B {
  def bar = super[A].foo
}
object Test extends C {
    assert(bar == "A")
  }
}
```

This works in both Scala 2 and Dotty, but the mechanism is very different, in Scala 2 we get:
```scala
  public java.lang.String foo();
    descriptor: ()Ljava/lang/String;
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokestatic  #16                 // InterfaceMethod B.foo$:(LB;)Ljava/lang/String;
         4: areturn
```

and in Dotty we get:
```scala
  public java.lang.String foo();
    descriptor: ()Ljava/lang/String;
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokespecial #19                 // InterfaceMethod B.foo:()Ljava/lang/String;
         4: areturn
```

But the PR description says:
> However, we've since discovered that we can no longer precisely target a particular statically resolved method with `invokespecial` due to impedance mismatch between JVM and Scala semantics.

So how come the Dotty version works ? The only other relevant difference I see from javap is that in Scala 2 we get:
```scala
public class C implements B
```
But in Dotty we get:
```scala
public class C implements A,B
```

So if declaring all the interfaces you extend from is enough to make invokespecial work as expected, why isn't Scala 2 doing that ? Is there some more complicated scenario where this falls apart ? This PR contains commit https://github.com/scala/scala/pull/5251/commits/91b066aac5edf53ca18603f8486eb255514b3118 whose commit message seems to be advocating for not pruning any parent, so I'm even more confused now. Ping @retronym ?

retronym:
@smarter IIRC, (redundantly) declaring all interfaces at each level triggered the default method related code in the JVM's classloading to be slow, which impacts startup time.

Given `trait A1 <: trait Ai <: trait An`, it also results in an `O(N^2)` cost in classfile size.

We considered doing it conditionally when a super call was actually present, but opted not to. I'd need to dig up the PR discussion to remember why.

retronym:
Related:

https://github.com/scala/scala-dev/issues/228
https://github.com/scala/scala/pull/5085#issuecomment-213365299

https://github.com/scala/scala-dev/issues/98#issuecomment-194602923

smarter:
Thanks for the pointers this is really helpful! I ended up writing a novel-sized issue to organize my thoughts here: https://github.com/lampepfl/dotty/issues/5928

PR #5254, opened 2016-06-28
SI-9834 Improve error on failed op=

som-snytt:
If rewriting `x += y` fails to typecheck, emit error messages
for both the original tree and the assignment.

If rewrite is not attempted because `x` is a val, then say so.


som-snytt:
Consider `s"Can't assign to $qual2"`. Similarly for `s"Missing update method."`.

sbromberger:
Thank you for this PR!

SethTisue:
Is this really high value and low risk enough to target 2.11.x rather than 2.12.x?

som-snytt:
There's a chance it creates even whackier error messages, but they are only error messages.

PR #5256, opened 2016-07-04
SI-9515 closure elimination also for non-Scala-Function SAM types

lrytz:
Also logged in as SD-162

The optimizer had conservative checks in place to perform closure
elimination only for Scala Function types. We can eliminate IndyLambda
instructions for any functional interface. LambdaMetaFactory only
constructs lambda objects for interface types, which don't have any
side-effects on construction - they don't have a constructor.


src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
```diff
@@ -430,7 +430,7 @@ class CallGraph[BT <: BTypes](val btypes: BT) {
     def unapply(insn: AbstractInsnNode): Option[(InvokeDynamicInsnNode, Type, Handle, Type)] = insn match {
       case indy: InvokeDynamicInsnNode if indy.bsm == metafactoryHandle || indy.bsm == altMetafactoryHandle =>
         indy.bsmArgs match {
-          case Array(samMethodType: Type, implMethod: Handle, instantiatedMethodType: Type, xs@_*) => // xs binding because IntelliJ gets confused about _@_*
+          case Array(samMethodType: Type, implMethod: Handle, instantiatedMethodType: Type, _@_*) =>
```
adriaanm, line 433: 🐌

adriaanm:
LGTM

SethTisue:
very cool

PR #4421, opened 2015-03-31
SI-9019  TraversableLike stringPrefix broken for inner classes

Ichoran:
Added a simple heuristic to detect inner class names: it starts with a capital.  This allows most inner class collections to automatically get their name instead of some other random thing, but avoids being confused by specialization.

Test added to verify behavior.

PR #5258, opened 2016-07-05
SI-9019  TraversableLike stringPrefix broken for inner classes

szeiger:
Resurrecting https://github.com/scala/scala/pull/4421 for 2.12.x

JIRA: https://issues.scala-lang.org/browse/SI-9019

### DIFF
--- a/src/library/scala/collection/TraversableLike.scala
+++ b/src/library/scala/collection/TraversableLike.scala
@@ -606,12 +606,21 @@ trait TraversableLike[+A, +Repr] extends Any
    *           simple name of the collection class $coll.
    */
   def stringPrefix : String = {
-    var string = repr.getClass.getName
-    val idx1 = string.lastIndexOf('.' : Int)
-    if (idx1 != -1) string = string.substring(idx1 + 1)
-    val idx2 = string.indexOf('$')
-    if (idx2 != -1) string = string.substring(0, idx2)
-    string
+    val fqn = repr.getClass.getName
+    val cls = {
+      val idx1 = fqn.lastIndexOf('.' : Int)
+      if (idx1 != -1) fqn.substring(idx1 + 1) else fqn
+    }
+    val parts = cls.split('$')
+    val last = parts.length - 1
+    parts.zipWithIndex.foldLeft("") { case (z, (s, i)) =>
+      if (s.isEmpty) z
+      else if (i != last && s.forall(java.lang.Character.isDigit)) "" // drop prefix in method-local classes
+      else if (i == 0 || java.lang.Character.isUpperCase(s.charAt(0))) {
+        if (z.isEmpty) s else z + '.' + s
+      }
+      else z
+    }
   }
 
   /** Creates a non-strict view of this $coll.
--- a/test/files/run/xMigration.check
+++ b/test/files/run/xMigration.check
@@ -1,11 +1,11 @@
 
 scala> Map(1 -> "eis").values    // no warn
-res0: Iterable[String] = MapLike(eis)
+res0: Iterable[String] = MapLike.DefaultValuesIterable(eis)
 
 scala> :setting -Xmigration:none
 
 scala> Map(1 -> "eis").values    // no warn
-res1: Iterable[String] = MapLike(eis)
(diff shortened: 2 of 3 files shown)


szeiger:
After seeing the test failure in `run/xMigration.scala`, I'm not convinced that the object name is "obviously wrong" as the description of SI-9019 asserts. I suspect this may be very much the intended behavior. The current behavior allows you to put implementation classes of a collection type into the companion object and have the stringPrefix automatically hide the concrete implementation class name. With this change you now get the much less useful implementation class names instead of the desired collection names.

szeiger:
Actually, scratch that. `MapLike.DefaultValuesIterable` is not even a subclass of `MapLike`. This may well be an intended feature but in the failing test case it's a bug.

szeiger:
Updated with an improved version that creates qualified names in these cases.

src/library/scala/collection/TraversableLike.scala
```diff
@@ -606,12 +606,21 @@ trait TraversableLike[+A, +Repr] extends Any
...
   def stringPrefix : String = {
-    var string = repr.getClass.getName
-    val idx1 = string.lastIndexOf('.' : Int)
-    if (idx1 != -1) string = string.substring(idx1 + 1)
-    val idx2 = string.indexOf('$')
-    if (idx2 != -1) string = string.substring(0, idx2)
-    string
+    val fqn = repr.getClass.getName
+    val cls = {
+      val idx1 = fqn.lastIndexOf('.' : Int)
```
SethTisue, line 611: nitpick (carried over from the old code): the `: Int` is doing nothing here

but anyway, doesn't this part boil down to `val cls = fqn.split('.').last`?

SethTisue:
LGTM, but @szeiger should (since the first commit fails validation) squash before merging (and address my trivial style nitpicks if he feels like it)

adriaanm:
(Note that GitHub gives you the option to squash on merge these days.)

szeiger:
Oops, didn't notice that the style nitpicks were new. An opportunity for a new PR :)

sjrd:
FTR, this PR brought a 66 KB increase in code size in Scala.js for any codebase that uses collections, but does not otherwise use `String.split()` nor the Unicode methods of `java.lang.Character` (most of it being due to `String.split()`). After Closure + gzip, those 66 KB are still 6 KB, which represents a **20% increase** on our code size "benchmark".

See https://github.com/scala-js/scala-js/issues/2591

I am not sure yet what is the best approach to fixing this issue, at this point (and whether it should be within Scala.js or in scala/scala). I'll think about it more tomorrow; but I thought I'd let you guys know.

szeiger:
That's interesting. `String.split(Char)` should be trivial to implement. Is the problem that the default implementation brings in dependencies on other large methods / classes? If that's the case, it could make sense for Scala.js to have its own implementation.

Using the Unicode-aware `Character` methods is conceptionally the right thing to do (of course, it would be even more correct if the code didn't fail on surrogate pairs) but I doubt that it has any practical relevance here. We could restrict this to checks for the appropriate ASCII ranges.

sjrd:
See #5400 for a fix and some more discussion. It turns out using Unicode-aware was not the best thing to do in this specific case.

For `String.split(Char)`, it is not _trivial_ to implement (the spec of the JDK's `split()` has extremely obscure corner cases), but it sure can be done. We _could_ have a Scala.js-specific implementation, but this means maintaining a separate copy of `StringLike.scala` in Scala.js. We do have that kind of "overrides" [here](https://github.com/scala-js/scala-js/tree/master/scalalib) when we don't have any other option, but prefer to have as little as possible of those, since they increase maintenance costs considerably.

szeiger:
@sjrd What we call here is not Java's `String.split` but Scala's `StringOps.split`. This looks like a good candidate for optimization. It currently escapes the `Char` as a regexp and then lets Java do the job. A custom implementation would probably be a lot faster (and Scala.js-friendly).

sjrd:
I understand that. I also think we could and should improve `StringLike.split(Char)`. But this is not enough for the issue I was facing. I'm happy to submit a follow-up PR with a better version of it, though.

PR #5259, opened 2016-07-05
SI-6881 Detect reference equality when comparing streams

szeiger:
`==` already covers this case. We override `equals` in `Stream` to do
the same when `equals` is called directly. This takes care of identical
streams.

To support short-circuiting equality checks on elements prepended to
identical streams we also override `sameElements` in `Cons` to treat
the case where both sides are `Cons` separately.

Tests in StreamTest.test_reference_equality.

Review by @Ichoran 

JIRA: https://issues.scala-lang.org/browse/SI-6881

### DIFF
--- a/src/library/scala/collection/immutable/Stream.scala
+++ b/src/library/scala/collection/immutable/Stream.scala
@@ -1029,6 +1029,8 @@ sealed abstract class Stream[+A] extends AbstractSeq[A]
    */
   override def stringPrefix = "Stream"
 
+  override def equals(that: Any): Boolean =
+    if (this eq that.asInstanceOf[AnyRef]) true else super.equals(that)
 }
 
 /** A specialized, extra-lazy implementation of a stream iterator, so it can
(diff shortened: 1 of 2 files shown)


szeiger:
Fixed stupid mistake...

SethTisue:
LGTM

I assume the use of a `@tailrec` helper is primarily about avoiding stack overflow, rather than primarily about saving a few CPU cycles here and there. you might add a comment to that effect before merging?

it took me a while to convince myself that a stack-safe `sameElements` actually needs to be this complicated, but I convinced myself. I think it could be improved sliiiiiightly by doing the `eq` check at the outer level of `sameElements` rather than inside the inner match. then in the case where I call `sameElements` directly (not from `==`) and `this` and `that` are the same cons cell, you avoid an unnecessary comparison of the head values. but whatevs, it's a really tiny nitpick, feel free to change it or merge as-is, as you see fit

szeiger:
Doing the `eq` check at the outer level of `sameElements` is not sufficient because there is no recursion on `sameElements` anymore. Tails are checked with `consEq` instead, so we need to perform the `eq` check there.

SethTisue:
> Doing the eq check at the outer level of sameElements is not sufficient

oops, I meant to say at the outer level of `consEq`. anyway, nbd

szeiger:
AFAIR I didnt't do that to avoid unnecessary comparisons when `sameElements` is called from `equals` or `==`.

PR #5260, opened 2016-07-05
SI-9068 Deprecate scala.collection.mutable.Stack

szeiger:
JIRA: [SI-9068](https://issues.scala-lang.org/browse/SI-9068)

### DIFF
--- a/src/library/scala/collection/mutable/Stack.scala
+++ b/src/library/scala/collection/mutable/Stack.scala
@@ -54,6 +54,7 @@ object Stack extends SeqFactory[Stack] {
  *  @define mayNotTerminateInf
  *  @define willNotTerminateInf
  */
+@deprecated("Stack is an inelegant and potentially poorly-performing wrapper around List. Use a List assigned to a var instead.", "2.12.0")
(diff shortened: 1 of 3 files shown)


sjrd:
Can we fix it instead? E.g., using an `Array[AnyRef]` + `size` inside?

chris-martin:
Jira link: [SI-9068](https://issues.scala-lang.org/browse/SI-9068)

szeiger:
Updated `serialization-new.check` to accomodate the new deprecation warning. I was unable to produce the other test failures locally. Let's see what happens on the next build...

lrytz:
> Can we fix it instead

not in 2.12, this should be part of a collections rework

lrytz:
i agree though that it's not great to deprecate it without an alternative (other than a var over a list), it does make code harder to read at times. there are also a few uses within the compiler.

SethTisue:
fyi @xuwei-k who tweeted a bad mutable.Stack bug recently

szeiger:
rebased

SethTisue:
LGTM

PR #5261, opened 2016-07-05
SI-9827 MatchIterator advances itself

som-snytt:
To avoid caveats about calling `next` (or `hasNext`) before
using `MatchData` methods on `MatchIterator`, just do it
internally as necessary.

Note `MatchIterator` behavior in the docs.

### DIFF
--- a/src/library/scala/util/matching/Regex.scala
+++ b/src/library/scala/util/matching/Regex.scala
@@ -11,21 +11,14 @@
  * with the main goal of pulling out information from those matches, or replacing
  * them with something else.
  *
- * There are four classes and three objects, with most of them being members of
- * Regex companion object. [[scala.util.matching.Regex]] is the class users instantiate
- * to do regular expression matching.
+ * [[scala.util.matching.Regex]] is the class users instantiate to do regular expression matching.
  *
- * The remaining classes and objects in the package are used in the following way:
- *
- * * The companion object to [[scala.util.matching.Regex]] just contains the other members.
+ * The companion object to [[scala.util.matching.Regex]] contains supporting members:
  * * [[scala.util.matching.Regex.Match]] makes more information about a match available.
- * * [[scala.util.matching.Regex.MatchIterator]] is used to iterate over multiple matches.
+ * * [[scala.util.matching.Regex.MatchIterator]] is used to iterate over matched strings.
  * * [[scala.util.matching.Regex.MatchData]] is just a base trait for the above classes.
  * * [[scala.util.matching.Regex.Groups]] extracts group from a [[scala.util.matching.Regex.Match]]
  *   without recomputing the match.
- * * [[scala.util.matching.Regex.Match]] converts a [[scala.util.matching.Regex.Match]]
- *   into a [[java.lang.String]].
- *
  */
 package scala.util.matching
 
@@ -35,6 +28,7 @@ import java.util.regex.{ Pattern, Matcher }
 /** A regular expression is used to determine whether a string matches a pattern
  *  and, if it does, to extract or transform the parts that match.
  *
+ *  === Usage ===
  *  This class delegates to the [[java.util.regex]] package of the Java Platform.
  *  See the documentation for [[java.util.regex.Pattern]] for details about
  *  the regular expression syntax for pattern strings.
@@ -53,6 +47,7 @@ import java.util.regex.{ Pattern, Matcher }
  *  Since escapes are not processed in multi-line string literals, using triple quotes
  *  avoids having to escape the backslash character, so that `"\\d"` can be written `"""\d"""`.
  *
+ *  === Extraction ===
  *  To extract the capturing groups when a `Regex` is matched, use it as
  *  an extractor in a pattern match:
  *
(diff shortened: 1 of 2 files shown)


som-snytt:
This has the virtue of allowing the one-liner `num.findAllIn("1 2 3").group(0)` instead of `num.findFirstMatchIn("1 2 3").map(_.matched).get`.

Side-effecting `hasNext` is weird.

```
scala> val num = """(\d+)""".r
num: scala.util.matching.Regex = (\d+)

scala> val ns = num.findAllIn("1 2 3")
ns: scala.util.matching.Regex.MatchIterator = non-empty iterator

scala> ns.hasNext
res0: Boolean = true

scala> ns.start
res1: Int = 0

scala> ns.next
res2: String = 1

scala> ns.start
res3: Int = 0

scala> ns.hasNext
res4: Boolean = true

scala> ns.start
res5: Int = 2

scala> ns.next
res6: String = 2

scala> ns.next
res7: String = 3

scala> ns.next
java.util.NoSuchElementException
  at scala.util.matching.Regex$MatchIterator.next(Regex.scala:783)
  at scala.util.matching.Regex$MatchIterator.next(Regex.scala:785)
  ... 30 elided

scala> ns.start
java.lang.IllegalStateException
  at scala.util.matching.Regex$MatchIterator.ensure(Regex.scala:799)
  at scala.util.matching.Regex$MatchIterator.start(Regex.scala:803)
  ... 30 elided
```

som-snytt:
Review by @Ichoran 
For its state machine homage.

I'm still not 100% persuaded by the `MatchIterator` API, but `Iterator[String]` is clearly useful, and for imperative looping, `MatchData` is helpful. So this PR just removes the big stumbling block. 

I also normalized the doc examples to dotted syntax.

src/library/scala/util/matching/Regex.scala
```diff
@@ -92,48 +87,68 @@ import java.util.regex.{ Pattern, Matcher }
...
- *  But `findAllIn` returns a special iterator of strings that can be queried for the `MatchData`
- *  of the last match:
+ *  To iterate over the matched strings, use `findAllIn`, which returns a special iterator
+ *  that can be queried for the `MatchData` of the last match:
  *
  *  {{{
- *  val mi = date findAllIn dates
- *  val oldies = mi filter (_ => (mi group 1).toInt < 1960) map (s => s"$s: An oldie but goodie.")
+ *  val mi = date.findAllIn(dates)
+ *  while (mi.hasNext) {
```
szeiger, line 116: Why an imperative version?

src/library/scala/util/matching/Regex.scala
```diff
@@ -92,48 +87,68 @@ import java.util.regex.{ Pattern, Matcher }
...
  *
  *  Note that `findAllIn` finds matches that don't overlap. (See [[findAllIn]] for more examples.)
  *
  *  {{{
  *  val num = """(\d+)""".r
- *  val all = (num findAllIn "123").toList  // List("123"), not List("123", "23", "3")
+ *  val all = num.findAllIn("123").toList  // List("123"), not List("123", "23", "3")
+ *  }}}
+ *
+ *  Also, the "current match" of a `MatchIterator` may be advanced by either `hasNext` or `next`.
```
szeiger, line 128: There's the core problem IMHO. `hasNext` should not affect the current state. But I can't think of a migration path towards an API where this is no longer the case, other than deprecating and replacing `MatchInterator` and the methods that produce it.

src/library/scala/util/matching/Regex.scala
```diff
@@ -743,49 +758,76 @@ object Regex {
...
    */
   class MatchIterator(val source: CharSequence, val regex: Regex, val groupNames: Seq[String])
   extends AbstractIterator[String] with Iterator[String] with MatchData { self =>
 
     protected[Regex] val matcher = regex.pattern.matcher(source)
-    private var nextSeen = false
 
-    /** Is there another match? */
+    // 0 = not yet matched, 1 = matched, 2 = advanced to match, 3 = no more matches
+    private var nextSeen = 0
```
szeiger: Can we make this `private[this]` (and `ensure`, too)?

src/library/scala/util/matching/Regex.scala
som-snytt, line 116: I think that's the appropriate use case for the API. The previous example is deceptive in filter.

src/library/scala/util/matching/Regex.scala
som-snytt, line 128: Yes, my initial expectation was to deprecate. But the API is not terribly harmful, modulo the behavior that is fixed. But, could also deprecate for removal next time. I'm agnostic on that point.

src/library/scala/util/matching/Regex.scala
som-snytt: yep.

som-snytt:
Made privates more privater, and also squashed.

Also, recalled the other reason for doing the easy fix, rather than rejigger the API (if that's the right thing to do): internally, MatchIterator is the underlying iterator, and Iterator[Match] is atop it. So conceivably, MatchIterator avoids allocations of Match if you don't want them.

szeiger:
LGTM

PR #3961, opened 2014-09-03
SI-8786 disable part of thest that's failing the jdk8 build

lrytz:
As noted in SI-8786, this test currently fails the [jdk 8 build](https://scala-webapps.epfl.ch/jenkins/view/2.11.x/job/scala-nightly-auxjvm-2.11.x/jdk=jdk8,label=auxjvm/122/console).

PR #5262, opened 2016-07-06
SI-8786 fix generic signature for @varargs forwarder methods

lrytz:
When generating a varargs forwarder for

```
def foo[T](a: T*)
```

the parameter type of the forwarder needs to be Array[Object]. If we
gnerate Array[T] in UnCurry, that would be erased to plain Object, and
the method would not be a valid varargs.

Unfortunately, setting the parameter type to Array[Object] lead to
an invalid generic signature - the generic signature should reflect the
real signature.

This change adds an attachment to the parameter symbol in the varargs
forwarder method and special-cases signature generation.

Also cleanes up the code to produce the varargs forwarder. For example,
type parameter and parameter symbols in the forwarder's method type were
not clones, but the same symbols from the original method were re-used.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/UnCurry.scala
+++ b/src/compiler/scala/tools/nsc/transform/UnCurry.scala
@@ -744,72 +744,88 @@ abstract class UnCurry extends InfoTransform
       if (!dd.symbol.hasAnnotation(VarargsClass) || !enteringUncurry(mexists(dd.symbol.paramss)(sym => definitions.isRepeatedParamType(sym.tpe))))
         return flatdd
 
-      def toArrayType(tp: Type): Type = {
-        val arg = elementType(SeqClass, tp)
-        // to prevent generation of an `Object` parameter from `Array[T]` parameter later
-        // as this would crash the Java compiler which expects an `Object[]` array for varargs
-        //   e.g.        def foo[T](a: Int, b: T*)
-        //   becomes     def foo[T](a: Int, b: Array[Object])
-        //   instead of  def foo[T](a: Int, b: Array[T]) ===> def foo[T](a: Int, b: Object)
-        arrayType(
-          if (arg.typeSymbol.isTypeParameterOrSkolem) ObjectTpe
-          else arg
-        )
-      }
+      val forwSym = currentClass.newMethod(dd.name.toTermName, dd.pos, VARARGS | SYNTHETIC | flatdd.symbol.flags)
 
-      val theTyper   = typer.atOwner(dd, currentClass)
-      val flatparams = flatdd.symbol.paramss.head
       val isRepeated = enteringUncurry(dd.symbol.info.paramss.flatten.map(sym => definitions.isRepeatedParamType(sym.tpe)))
 
-      // create the type
-      val forwformals = map2(flatparams, isRepeated) {
-        case (p, true) => toArrayType(p.tpe)
-        case (p, false)=> p.tpe
-      }
-      val forwresult = dd.symbol.tpe_*.finalResultType
-      val forwformsyms = map2(forwformals, flatparams)((tp, oldparam) =>
-        currentClass.newValueParameter(oldparam.name.toTermName, oldparam.pos).setInfo(tp)
-      )
-      def mono = MethodType(forwformsyms, forwresult)
-      val forwtype = dd.symbol.tpe match {
-        case MethodType(_, _) => mono
-        case PolyType(tps, _) => PolyType(tps, mono)
-      }
+      val oldPs = flatdd.symbol.paramss.head
 
-      // create the symbol
-      val forwsym = currentClass.newMethod(dd.name.toTermName, dd.pos, VARARGS | SYNTHETIC | flatdd.symbol.flags) setInfo forwtype
-      def forwParams = forwsym.info.paramss.flatten
+      // see comment in method toArrayType below
+      val arrayTypesMappedToObject = mutable.Map.empty[Symbol, Type]
 
-      // create the tree
-      val forwtree = theTyper.typedPos(dd.pos) {
-        val locals = map3(forwParams, flatparams, isRepeated) {
-          case (_, fp, false)       => null
-          case (argsym, fp, true)   =>
-            Block(Nil,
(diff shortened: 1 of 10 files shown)


adriaanm:
(Reverts #3961)

src/compiler/scala/tools/nsc/transform/UnCurry.scala
```diff
@@ -744,72 +744,88 @@ abstract class UnCurry extends InfoTransform
...
+          val elem = if (arg.typeSymbol.isTypeParameterOrSkolem && !(arg <:< AnyRefTpe)) {
+            // To prevent generation of an `Object` parameter from `Array[T]` parameter later
+            // as this would crash the Java compiler which expects an `Object[]` array for varargs
+            //   e.g.        def foo[T](a: Int, b: T*)
+            //   becomes     def foo[T](a: Int, b: Array[Object])
+            //   instead of  def foo[T](a: Int, b: Array[T]) ===> def foo[T](a: Int, b: Object)
+            //
+            // In order for the forwarder method to type check we need to insert a cast
+            //   def foo'[T'](a: Int, b: Array[Object]) = foo[T'](a, wrapRefArray(b).asInstanceOf[Seq[T']])
+            // The target type for that cast is stored in the `arrayTypesMappedToObject` map.
```
adriaanm: Small nit: the map does not store the exact target type, but the element type `T` of the type `Seq[T]` that we'll cast to. Maybe we should store `seqType(originalArg)` instead?

adriaanm: ... although the status quo is more symmetric with `TypeParamVarargsAttachment`'s argument

src/compiler/scala/tools/nsc/transform/UnCurry.scala
```diff
@@ -744,72 +744,88 @@ abstract class UnCurry extends InfoTransform
...
+            val wrap = gen.mkWrapArray(Ident(param), parTp)
+            arrayTypesMappedToObject.get(param) match {
+              case Some(tp) => gen.mkCast(wrap, seqType(tp))
+              case _ => wrap
+            }
+          }
+        })
+        val end = if (forwSym.isConstructor) List(UNIT) else Nil
+
+        DefDef(forwSym, BLOCK(Apply(gen.mkAttributedRef(flatdd.symbol), seqArgs) :: end : _*))
```
adriaanm: Simplify to the following?

```
val forwCall = Apply(gen.mkAttributedRef(flatdd.symbol), seqArgs)
DefDef(forwSym, if (forwSym.isConstructor) Block(List(forwCall), Literal(Constant(()))) else forwCall)
```

adriaanm: This pattern of block with singleton stat + unit expr is pretty common. At some point we should introduce a factory method for it (I'm not saying in this PR -- some day)

test/files/jvm/t8786-sig.scala
```diff
@@ -0,0 +1,112 @@
...
+  check(sig("m4", sq)   , "public int A.m4(scala.collection.Seq)")
+  check(sig("m5", sq)   , "public java.lang.String A.m5(scala.collection.Seq)")
+  check(sig("m6", sq)   , "public java.lang.String A.m6(scala.collection.Seq)")
+  check(sig("m7", sq)   , "public int A.m7(scala.collection.Seq)")
+  check(sig("m8", sq)   , "public java.lang.Object A.m8(scala.collection.Seq)")
+
+  check(genSig("m1", sq), "public <T> T A.m1(scala.collection.Seq<T>)")
+  check(genSig("m2", sq), "public <T> T A.m2(scala.collection.Seq<T>)")
+  check(genSig("m3", sq), "public <T> T A.m3(scala.collection.Seq<T>)")
+  check(genSig("m4", sq), "public <T> T A.m4(scala.collection.Seq<T>)") // wrong signature for T <: Int, SI-9846
```
adriaanm: Thanks for noting this. It would be doubleplus clear if it read something like:

> // TODO: this is testing the (incorrect) status quo -- the signature should be `public int A.m7(scala.collection.Seq<java.lang.Object>)` (see SI-9846)

 (assuming that is indeed the correct signature :-))

adriaanm: Could we just drop invalid generic signatures like this later in the pipeline? Probably not a big deal, though -- who writes code like that!?!? 😎

adriaanm:
Nice cleanups for tricky stuff! A bit of polish and then LGTM

src/compiler/scala/tools/nsc/transform/UnCurry.scala
lrytz: will update the comment

src/compiler/scala/tools/nsc/transform/UnCurry.scala
lrytz: will do, thanks

test/files/jvm/t8786-sig.scala
lrytz: ok, updated the comment

adriaanm:
LGTM

adriaanm:
Java bug?
```
public interface Foo {
  public <T> void tempMethod(T... ia) throws Exception;
}
```

```
$ javac -version
javac 1.6.0_65
```

```
➜  /tmp ~/scala/scala-2.11.8/bin/scala
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.6.0_65).

scala> classOf[Foo].getDeclaredMethod("tempMethod", classOf[Array[Object]]).toGenericString
res1: String = public abstract <T> void Foo.tempMethod(T[]) throws java.lang.Exception

➜  /tmp java_home 1.8               
➜  /tmp scala
Welcome to Scala 2.12.0 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> classOf[Foo].getDeclaredMethod("tempMethod", classOf[Array[Object]]).toGenericString
res0: String = public abstract <T> void Foo.tempMethod(T...) throws java.lang.Exception

```

PR #5263, opened 2016-07-06
SI-5294 SI-6161 Hard graft in asSeenFrom, refinements, and existentials [ci: last-only]

retronym:
Fixing SI-6161 / SI-5954 seemed like it would be straightforward, after I had pinpointed a logical error in `asSeenFrom`: a type seen from a prefix with type arguments can have more base classes than `tp.typeSymbol.baseClasses`. `asSeenFrom` was only considering the latter when matching the prefix.

However, the fix uncovered a number of latent bugs related to existential and refinement types, which I needed to fix first.
- Make the "refined-type-as-a-lazy-base-type" idiom more distinguished by calling out the construction and deconstruction of such types with calls to `[is]IntersectionTypeForLazyBaseTypeElement`. Also changed: the type symbol shared by all parents as the type symbol for the `RefinedType` avoiding the one special case in `BaseTypeSeq#typeSymbol`
- In LUB/GLB-s,`mergePrefixAndArgs` needs to be aware of this idiom. I also found it was unreliable (different answers based on the order of the list of types) with existentials.
- Finally, back the the original bug fix, we can [update `asSeenFrom`](https://github.com/scala/scala/pull/5263/files?w=1#diff-fd1e83c55a8c7d6bec7b5290d7fb5998R613) to compile [t6161b.scala](https://github.com/retronym/scala/blob/5d59cd6e6dc2ffaf92a5244a999d90948a5adc56/test/files/pos/t6161b.scala), without triggering regressions elsewhere. 

I've checked `t6161b.scala` in dotc. It compiles that already, presumably because it is [using `pre.baseTypeRef(thiscls).exists` in `toPrefix`](https://github.com/lampepfl/dotty/blob/9da40cb84d613bcea3a0e0890b2e53129fe60bc6/src/dotty/tools/dotc/core/TypeOps.scala#L67).

Our `matchesPrefixAndArgs` still contains more special cases that I would like, related to the inconsistency of the base type sequence of `ThisType` (noted as [SD#166](https://github.com/scala/scala-dev/issues/166)), the fact that the base type sequence of `TypeVar` does not contain the bounds of the originating abstract type, and a tricky issue with aliases+refinement types.


retronym:
Rework of the rework in #5041.

retronym:
Community build: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/453/

retronym:
Review by @adriaanm.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -142,6 +142,7 @@ class Inliner[BT <: BTypes](val btypes: BT) {
           elided += r
         else
           result += r
+        ()
```
retronym, line 145: The improvements to LUBs now compute a sharper LUB for this expression, but that fails in bound checks. Adding `()` to avoid the LUB getting `refchecked`.

src/reflect/scala/reflect/internal/BaseTypeSeqs.scala
```diff
@@ -56,42 +56,44 @@ trait BaseTypeSeqs {
...
-     *  no evaluation needed.
-     */
-    def typeSymbol(i: Int): Symbol = {
-      elems(i) match {
-        case RefinedType(v :: vs, _) => v.typeSymbol
-        case tp => tp.typeSymbol
-      }
-    }
+    /** The type symbol of the type at i'th position in this sequence */
+    def typeSymbol(i: Int): Symbol = elems(i).typeSymbol
```
retronym, line 96: This special case was removed in favour of setting the type symbol of the `RefinedType` to the common type symbol of the parents on construction in `intersectionTypeForLazyBaseType`

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -3598,6 +3602,14 @@ trait Types
     case tp :: Nil  => tp
     case _          => refinedType(tps, commonOwner(tps))
   }
+  def intersectionTypeForLazyBaseType(tps: List[Type]) = tps match {
+    case tp :: Nil  => tp
+    case _          => RefinedType(tps, newScope, tps.head.typeSymbolDirect)
+  }
+  def isIntersectionTypeForLazyBaseType(tp: RefinedType) = tp.parents match {
+    case head :: _ => tp.typeSymbolDirect eq head.typeSymbolDirect
```
retronym, line 3610: I'm also using the fact that we're using the underlying type symbol as the type symbol of the `RefinedType` as the way to distinguish them here.

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4411,83 +4423,93 @@ trait Types
...
+    val tps = tps0.flatMap {
+      case rt: RefinedType if isIntersectionTypeForLazyBaseType(rt) => rt.parents
+      case ExistentialType(qs, underlying) =>
+        tparams ++= qs
+        underlying match {
+          case rt: RefinedType if isIntersectionTypeForLazyBaseType(rt) => rt.parents
+          case tp => tp :: Nil
+        }
+      case tp => tp :: Nil
+    }
```
retronym: Unwrap all the existentials, IntersectionTypeForLazyBaseType-s, and existentials-over-IntersectionTypeForLazyBaseType.

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4411,83 +4423,93 @@ trait Types
...
-              else {
-                if (tparam.variance == variance) lub(as, depth.decr)
-                else if (tparam.variance == variance.flip) glb(as, depth.decr)
+          else transposeSafe(argss) match {
+            case None =>
+              // transpose freaked out because of irregular argss
+              // catching just in case (shouldn't happen, but also doesn't cost us)
+              // [JZ] It happens: see SI-5683.
+              debuglog(s"transposed irregular matrix!? tps=$tps argss=$argss")
+              NoType
```
retronym, line 4488: Prior to this PR, lub-from-hell-2.scala compiled because we hit this case, skipped unification at `scala.collection.mutable.Clonable`, and went on to unify at `scala.Cloneable`. 

Why did we get here?

Above, in:

``` scala
val argss = tps map (_.normalize.typeArgs) // symbol equality (of the tp in tps) was checked using typeSymbol, which normalizes, so should normalize before retrieving arguments
```

We got the refinement class symbol of the `RefinedType` encoding the lazy BTS element.

src/reflect/scala/reflect/internal/Types.scala
retronym: I've since noticed duplication with `stripExistentialsAndTypeVars`, used elsewhere in this file for the same purpose.

retronym:
``` diff
% diff /home/jenkins/workspace/scala-2.12.x-validate-test/test/files/neg/lub-from-hell-2-neg.log /home/jenkins/workspace/scala-2.12.x-validate-test/test/files/neg/lub-from-hell-2.check
@@ -1,7 +1,7 @@
 lub-from-hell-2.scala:3: error: type arguments [Any,Iterable[Any] with Int => Any with scala.collection.generic.Subtractable[Any,Iterable[Any] with Int => Any with scala.collection.generic.Subtractable[Any,Iterable[Any] with Int => Any]{def seq: Iterable[Any] with Int => Any}]{def seq: Iterable[Any] with Int => Any{def seq: Iterable[Any] with Int => Any}}] do not conform to trait Subtractable's type parameter bounds [A,+Repr <: scala.collection.generic.Subtractable[A,Repr]]
   def foo(a: Boolean, b: collection.mutable.Set[Any], c: collection.mutable.ListBuffer[Any]) = if (a) b else c
       ^
-lub-from-hell-2.scala:4: error: type arguments [Any,scala.collection.mutable.Iterable[Any] with scala.collection.mutable.Cloneable[scala.collection.mutable.Iterable[Any] with scala.collection.mutable.Cloneable[scala.collection.mutable.Iterable[Any] with Cloneable with Int => Any] with Int => Any{def seq: scala.collection.mutable.Iterable[Any] with Cloneable with Int => Any}] with scala.collection.generic.Growable[Any] with Int => Any with scala.collection.generic.Subtractable[Any,Iterable[Any] with Int => Any with scala.collection.generic.Subtractable[Any,Iterable[Any] with Int => Any]{def seq: Iterable[Any] with Int => Any}] with scala.collection.generic.Shrinkable[Any] with scala.collection.script.Scriptable[Any]] do not conform to trait Subtractable's type parameter bounds [A,+Repr <: scala.collection.generic.Subtractable[A,Repr]]
+lub-from-hell-2.scala:4: error: type arguments [Any,scala.collection.mutable.Iterable[Any] with scala.collection.mutable.Cloneable[scala.collection.mutable.Iterable[Any] with scala.collection.mutable.Cloneable[scala.collection.mutable.Iterable[Any] with Cloneable with Int => Any] with Int => Any{def seq: scala.collection.mutable.Iterable[Any] with Cloneable with Int => Any}] with scala.collection.generic.Growable[Any] with Int => Any with scala.collection.generic.Shrinkable[Any] with scala.collection.generic.Subtractable[Any,Iterable[Any] with Int => Any with scala.collection.generic.Subtractable[Any,Iterable[Any] with Int => Any]{def seq: Iterable[Any] with Int => Any}] with scala.collection.script.Scriptable[Any]] do not conform to trait Subtractable's type parameter bounds [A,+Repr <: scala.collection.generic.Subtractable[A,Repr]]
   def bar(a: Boolean, b: scala.collection.mutable.SetLike[Any,scala.collection.mutable.Set[Any]], c: scala.collection.mutable.Buffer[Any]) = if (a) b else c
       ^
 two errors found
```

retronym:
- [ ] add diagnostic to the change in `ContainsCollector`  to see if we run into the changed behaviour during partest or bootstrap
- [x] refactor `normalize` / `normalizeImpl` into a standalone implementation method to better capture the current semantics: (**DEFERRED** to https://github.com/scala/scala-dev/issues/200)
- [x] decide whether to fix (?) `RefinementTypeRef#normalize` (`pre.memberInfo(sym)` or `pre.memberInfo(sym).normalize`)
- [x] add comment to `ContainsCollector` to explain the subtle change there.
- [x] Before this change, `type Inty[X] = Int; typeOf[Inty[String]].contains(symbolOf[String])` was false. This looks like a bug (right?): understand why this is and fix it.
- [x] Also, should `typeOf[Inty[String]].contains(IntClass)`. Should it? What breaks without that?
- [x] Consolidate changes with `stripExistentialsAndTypeVars`?
- [x] Make the status of the lub-from-hell tests clearer (they capture the status quo rather than the ideal case)

retronym:
/rebuild

Some failures were due to JVM crashes due to lack of memory on the build server.

adriaanm:
Alas, this will need a rebase.

retronym:
Rebased

retronym:
I'm analysing the effect of the change to `contains`. Here are cases compiling the library in which the old and new implementations differ.

```
⚡ qscalac -J-Xmx2G -d /tmp src/library/**/*.scala
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: scala.collection.GenTraversable[_$1] with scala.collection.generic.GenericTraversableTemplate[_$1,CC].contains(type _$1) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: Traversable[_$2] with scala.collection.generic.GenericClassTagTraversableTemplate[_$2,CC].contains(type _$2) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: Traversable[_$2] with scala.collection.generic.GenericOrderedTraversableTemplate[_$2,CC].contains(type _$2) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: scala.collection.parallel.ParIterable[_$1] with scala.collection.generic.GenericParTemplate[_$1,CC].contains(type _$1) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: scala.collection.parallel.ParMap[_$2,_$3] with scala.collection.parallel.ParMapLike[_$2, _$3, CC[_$2,_$3], _].contains(type _$2) = false / true
contains changed: scala.collection.parallel.ParMap[_$2,_$3] with scala.collection.parallel.ParMapLike[_$2, _$3, CC[_$2,_$3], _].contains(type _$3) = false / true
contains changed: scala.collection.parallel.ParMap[_$2,_$3] with scala.collection.parallel.ParMapLike[_$2, _$3, CC[_$2,_$3], _].contains(type _$2) = false / true
contains changed: scala.collection.parallel.ParMap[_$2,_$3] with scala.collection.parallel.ParMapLike[_$2, _$3, CC[_$2,_$3], _].contains(type _$3) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: CC.contains(type CC) = false / true
contains changed: scala.collection.parallel.ParSet[_$1] with scala.collection.parallel.ParSetLike[_$1, CC[_$1], _] with scala.collection.generic.GenericParTemplate[_$1,CC].contains(type _$1) = false / true
```

These are happening during finitary class graph checking (the constraints we place on inheritance to make subtyping decidable.)

My gut feeling is that the new behaviour is what we want `TypeRef(_, CC, Nil).contains(CC)` right? I'm seeing if the old behaviour can be exploited to a latent bug.

Here's a small look at the surprising old behaviour of `contains`:

```
scala> class C[M[_]]
defined class C

scala> val M = symbolOf[C[Any]].typeParams.head; val M_tpeHK = M.tpeHK; M_tpeHK.contains(M)
M: $r.intp.global.Symbol = type M
M_tpeHK: $r.intp.global.Type = M
res8: Boolean = false
```

retronym:
Giving the community build a spin shortly in [scala-2.12.x-integrate-community-build/480](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/480/)

adriaanm:
(I hit rebuild with corrected scala_ref)

retronym:
The community build failed with errors in Scaladoc that have since been fixed in 2.12.x. I have rebased and [restarted that build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/482/console).

adriaanm:
Looks like the error message is correct, and scalaz was accidentally relying on old buggy behavior?

```
effect/src/main/scala/scalaz/effect/MonadIO.scala:49: ambiguous implicit values:
  both method eitherTMonadError in class EitherTInstances3 
    of type [F[_], E](implicit F0: scalaz.Monad[F])scalaz.MonadError[[α, β]scalaz.EitherT[F,α,β],E]
  and method eitherTMonad in class EitherTInstances1 
    of type [F[_], L](implicit F0: scalaz.Monad[F])scalaz.Monad[[α]scalaz.EitherT[F,L,α]]
  match expected type scalaz.Monad[[α]scalaz.EitherT[F,E,α]]
   implicit def eitherTMonadIO[F[_]: MonadIO, E] = fromLiftIO[({type λ[α]=EitherT[F, E, α]})#λ]
                                                             ^
```

retronym:
Standalone repro:

``` scala
trait EitherT[F[_], L, A]
trait Monad[F[_]]
trait MonadError[G[_, _], S] extends Monad[({type l[A] = G[S, A]})#l]

object Test {
  trait Partial[F[_], L] {
    type EitherFL[A] = EitherT[F, L, A]  
  }
  trait Partial1[F[_]] {
    type EitherF[L, A] = EitherT[F, L, A]
  }

  trait EitherTInstances3 {
    implicit def eitherTMonadError[F[_], E]: MonadError[Partial1[F]#EitherF,E] = ???
  }
  trait EitherTInstances1 extends EitherTInstances3 {    
    implicit def eitherTMonad[F[_], L]: Monad[Partial[F, L]#EitherFL] = ???
  }
  object Implicits extends EitherTInstances1
  import Implicits._
  def foo[F[_]](implicit F: Monad[F]) = ()

  foo[Partial[Option, Int]#EitherFL](eitherTMonad)
  foo[Partial[Option, Int]#EitherFL](eitherTMonadError)
  foo[Partial[Option, Int]#EitherFL] // ambiguous under #5263
}
```

retronym:
This is a progression. Here's a minimization:

```
trait M1[A]; trait M2[A] extends M1[A]

trait N1[A[_]]; trait N2[A[_]] extends N1[A]

object Test {
  type D = DummyImplicit
  def m[A](implicit d: D): M1[A] = ???
  def m[A](implicit d1: D, d2: D): M2[A] = ???
  m // not ambiguous, M1[_] <:< M2[_] (type params of the methods are existentially abstracted in `isAsSpecificValueType`)

  def n[A[_]](implicit d: D): N1[A] = ???
  def n[A[_]](implicit d1: D, d2: D): N2[A] = ???
  n // was not ambiguous in 2.11:
    // existentialAbtraction(A :: Nil, N2[A]) did not notice that `N2[A]` contained
    // `A` (because it was eta expanded to [_]A[_] before matching for `case TypeRef(_, ASymbol, _)`)
}

```

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz, line 145: this looks undesirable - can you give a minimized example?

retronym, line 145: See the comments in the `lub-from-hell*.scala` tests. We actually had similar cases that failed before this change, this one "accidentally" worked thanks to a bug.

retronym, line 145: Here's a minimal example of the pattern of an unlubbable pair of types:

``` scala
scala> trait T[A, +Repr <: T[A, Repr]]; class U extends T[Int, U]; class V extends T[Char, V]
defined trait T
defined class U
defined class V

scala> trait T[A, +Repr <: T[A, Repr]]; class U extends T[Int, U]; class V extends T[Char, V]; val x = List(new U, new V)
<console>:14: error: type arguments [?,T[_ >: Char with Int <: AnyVal, Object]] do not conform to trait T's type parameter bounds [A,+Repr <: T[A,Repr]]
       trait T[A, +Repr <: T[A, Repr]]; class U extends T[Int, U]; class V extends T[Char, V]; val x = List(new U, new V)
                                                                                                   ^
<console>:14: error: type arguments [?,T[_ >: Char with Int <: AnyVal, Object]] do not conform to trait T's type parameter bounds [A,+Repr <: T[A,Repr]]
       trait T[A, +Repr <: T[A, Repr]]; class U extends T[Int, U]; class V extends T[Char, V]; val x = List(new U, new V)
                                                                                                       ^
```

retronym, line 145: I think the actual LUB should look more like:

```
T[_1, UV] forSome {
   type _1 >: Int with Char <: AnyVal
   type UV >: U with V <: T[_1, UV]
}
```

`````` scala

scala> val x = List[T[_1, UV] forSome { type _1 >: Int with Char <: AnyVal; type UV >: U with V <: T[_1, UV]}](new U, new V)
x: List[T[_1,UV] forSome { type _1 >: Int with Char <: AnyVal; type UV >: U with V <: T[_1,UV] }] = List(U@319aa9ee, V@1d806de6)```
``````

File: src/reflect/scala/reflect/internal/tpe/TypeMaps.scala [hunk 6]
```diff
@@ -610,11 +610,26 @@ private[internal] trait TypeMaps {
...
+    private def matchesPrefixAndClass(pre: Type, clazz: Symbol)(candidate: Symbol) = (clazz == candidate) && {
+      val pre1 = pre match {
+        case tv: TypeVar =>
+          // Needed with existentials in prefixes, e.g. test/files/pos/typevar-in-prefix.scala
+          // Perhaps the base type sequence of a type var should include it bounds?
+          tv.origin
+        case _ => pre
+      }
+      // widen needed (at least) because of https://github.com/scala/scala-dev/issues/166
+      (clazz == candidate) && (
```
lrytz: redundant check (already checked above)

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -87,9 +90,24 @@ trait JFileDirectoryLookup[FileEntryType <: ClassRepresentation] extends Directo
...
+      case Some(f) => dir.listFiles(mkFileFilter(f))
+      case None => dir.listFiles()
+    }
+
+    // Sort by file name for stable order of directory .class entries in package scope.
+    // This gives stable results ordering of base type sequences for unrelated classes
+    // with the same base type depth.
+    //
+    // Notably, this will stably infer`Product with Serializable`
+    // as the type of `ase class C(); case class D(); List(C(), D()).head`, rather than the opposite order.
```
vendethiel: `case class C`

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
```diff
@@ -610,11 +610,26 @@ private[internal] trait TypeMaps {
...
-    // Since pre may be something like ThisType(A) where trait A { self: B => },
-    // we have to test the typeSymbol of the widened type, not pre.typeSymbol, or
-    // B will not be considered.
-    private def matchesPrefixAndClass(pre: Type, clazz: Symbol)(candidate: Symbol) =
-      (clazz == candidate) && (pre.widen.typeSymbol isSubClass clazz)
+    private def matchesPrefixAndClass(pre: Type, clazz: Symbol)(candidate: Symbol) = (clazz == candidate) && {
+      val pre1 = pre match {
+        case tv: TypeVar =>
+          // Needed with existentials in prefixes, e.g. test/files/pos/typevar-in-prefix.scala
+          // Perhaps the base type sequence of a type var should include it bounds?
```
vendethiel: minor typo: "its bounds"

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz, line 145: OK, thanks - I hope it's not too common in user code that the bugfix unshadows an invalid LUB.

retronym:
Here's a miminization of the Slick failure in the community build:

``` scala
trait OuterA[Inner_] { type Inner = Inner_ }
trait OuterB         { type Inner }

abstract class Test1 {
  val o: OuterA[_]
  def foo(a: Any): o.Inner
  val etaExpanded: Any => o.Inner = foo // used to be disallowed, now disallowed (found: Any => Any, required: ...)
}

abstract class Test2 {
  def test(o: OuterA[_]) {
    def foo(a: Any): o.Inner = ???
    val etaExpanded: Any => o.Inner = foo // used to be allowed, now disallowed (found: Any => Any, required ...)
  }
}

abstract class Test3 {
  def test(o: OuterB) {
    def foo(a: Any): o.Inner = ???
    val etaExpanded: Any => o.Inner = foo // used to be disallowed, now disallowed (method with dependent type (a: Any)o.Inner cannot be converted to function value)
  }
}
```

retronym:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/488/console

adriaanm:
Looks like there's another instance of the implicit problem in specs2 3.6:

```
common/src/main/scala/org/specs2/control/package.scala:44: value >> is not a member of org.specs2.control.package.Action[Unit]
   warn(message) >> ActionT.fail[IO, Logs, Logger, A](failureMessage)
                 ^
common/src/main/scala/org/specs2/control/package.scala:55: value >> is not a member of org.specs2.control.package.Action[Unit]
   log(t.getMessage) >>
                     ^
common/src/main/scala/org/specs2/io/FileSystem.scala:47: value void is not a member of org.specs2.control.Action[Boolean]
   action.andFinally(deleteFile(path).void)
                                      ^
common/src/main/scala/org/specs2/io/FileSystem.scala:51: value void is not a member of org.specs2.control.ActionT[scalaz.effect.IO,org.specs2.control.Logs,org.specs2.control.Logger,Boolean]
   Actions.safe(path.toFile.mkdirs).void
                                    ^
common/src/main/scala/org/specs2/io/FileSystem.scala:123: value >> is not a member of org.specs2.control.Action[Unit]
   mkdirs(dest) >>
                ^
common/src/main/scala/org/specs2/io/FileSystem.scala:128: Implicit not found: scalaz.Unapply[scalaz.Applicative, org.specs2.control.Action[Unit]]. Unable to unapply type `org.specs2.control.Action[Unit]` into a type constructor of kind `M[_]` that is classified by the type class `scalaz.Applicative`. Check that the type class is defined by compiling `implicitly[scalaz.Applicative[type constructor]]` and review the implicits in object Unapply, which only cover common type 'shapes.'
       directories.toList.map(dir => copyDir(dir, dest / dir.name)).sequenceU.void
                                                                    ^
common/src/main/scala/org/specs2/io/FileSystem.scala:137: value >> is not a member of org.specs2.control.Action[Unit]
   mkdirs(dest) >>
                ^
common/src/main/scala/org/specs2/io/FileSystem.scala:142: value void is not a member of org.specs2.control.ActionT[scalaz.effect.IO,org.specs2.control.Logs,org.specs2.control.Logger,java.nio.file.Path]
[specs2_36] [error] possible cause: maybe a semicolon is missing before `value void'?
   }.void
     ^
common/src/main/scala/org/specs2/io/FileSystem.scala:146: value >> is not a member of org.specs2.control.Action[Unit]
   mkdirs(filePath.dir) >>
                        ^
common/src/main/scala/org/specs2/io/FileSystem.scala:151: value void is not a member of org.specs2.control.ActionT[scalaz.effect.IO,org.specs2.control.Logs,org.specs2.control.Logger,Boolean]
   Actions.safe(file.toFile.delete).void
                                    ^
common/src/main/scala/org/specs2/io/FileSystem.scala:155: Implicit not found: scalaz.Unapply[scalaz.Applicative, org.specs2.control.Action[Unit]]. Unable to unapply type `org.specs2.control.Action[Unit]` into a type constructor of kind `M[_]` that is classified by the type class `scalaz.Applicative`. Check that the type class is defined by compiling `implicitly[scalaz.Applicative[type constructor]]` and review the implicits in object Unapply, which only cover common type 'shapes.'
   listFilePaths(dir).flatMap(_.map(delete).toList.sequenceU.void) >>
                                                   ^
common/src/main/scala/org/specs2/text/SourceFile.scala:30: value >> is not a member of org.specs2.control.package.Action[Unit]
   log("  found classes: "+found.mkString(", "), verbose && found.nonEmpty) >>
                                                                            ^
```

adriaanm:
I [rebased](https://github.com/scala/scala/compare/2.12.x...adriaanm:review/5041) this to move the commits that I understood to be problematic ("Backwards compabiilty with normalizing contains" and "Remove normalization from Type#contains") to the end, and am running a [community build](https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/498/console) on HEAD^^ of that branch ([Improve RefinementTypeRef#normalize](https://github.com/adriaanm/scala/commit/e9e8618))

adriaanm:
Green!

retronym:
@adriaanm Thanks! I've dropped those commits and left a breadcrumb in https://github.com/scala/scala-dev/issues/207 to pick up the investigation again.

adriaanm:
LGTM

retronym:
BTW, credit to @paulp who, in #5041, uncovered some important details that had previously prevented me from getting 893acc18.

PR #5264, opened 2016-07-06
SI-8561 named subclasses for known Manifest / ClassTag instances

lrytz:
This helps keeping ClassTag serialization stable under accidental
changes (like changing the order of definitions, which would change the
name of the anonymous classes).


lrytz:
@retronym i restored parts of the serialization test that were disabled in https://github.com/scala/scala/commit/dc9effeb3af789d9e4a6b8e126ac73ac718270fa#diff-f902cea1239346244c50f9198db286f5

adriaanm:
LGTM

PR #5265, opened 2016-07-06
SI-6947 Better type parameter names for Map classes

szeiger:
Type parameter names are currently assigned pretty much alphabetically
without any meaning. This change renames all key parameters in Map
classes from `A` to `K` and all value parameters from `B` to `V` to
make them more meaningful. Derived names are renamed accordingly (e.g.
`V1` instead of `B1` for an upper bound on `V`, `W` instead of `C` for
a new value type).

As a side-effect this solves the documentation problem in SI-6947. Due
to using `B` both as a type parameter for `foldLeft[B]` in
`GenTraversableOnce[A]` and in `Map[A, B]` which extends
`GenTraversableOnce[(A, B)]`, the signature of `Map.foldLeft` was
rendered in scaladoc as

```
def foldLeft[B](z: B)(op: (B, (A, B)) ⇒ B): B 
```

Now you get an unambiguous version:

```
def foldLeft[B](z: B)(op: (B, (K, V)) ⇒ B): B
```

### DIFF
--- a/src/library/scala/collection/GenMap.scala
+++ b/src/library/scala/collection/GenMap.scala
@@ -18,18 +18,18 @@ import generic._
  *  @author Aleksandar Prokopec
  *  @since 2.9
  */
-trait GenMap[A, +B]
-extends GenMapLike[A, B, GenMap[A, B]]
-   with GenIterable[(A, B)]
+trait GenMap[K, +V]
+extends GenMapLike[K, V, GenMap[K, V]]
+   with GenIterable[(K, V)]
 {
-  def seq: Map[A, B]
+  def seq: Map[K, V]
 
-  def updated [B1 >: B](key: A, value: B1): GenMap[A, B1]
+  def updated [V1 >: V](key: K, value: V1): GenMap[K, V1]
 }
 
 object GenMap extends GenMapFactory[GenMap] {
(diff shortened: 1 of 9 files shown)


soc:
LGTM assuming that this doesn't break existing code including subclasses.

szeiger:
Failing test updated. Subclasses should not be affected. The changes are only to type parameters (which are not visible outside of their lexical scope), not to type members. In fact, I did not update the implementation subclasses in the standard library (because they do not contribute new API or documentation) or the parallel Map classes (which already used K/V instead of A/B).

soc:
Would be nice to update subclasses, too, so that things are consistent in the code...

szeiger:
@soc I didn't want to touch too much unnecessary code because this "small" change is so pervasive that it makes it pretty much impossible to merge other changes between both versions.

kmizu:
If better type parameter names are required for readability, why are `[Key ,Value]` not used instead of `[K, V]` ?

szeiger:
I hope that K/V is less controversial because it doesn't break with the tradition of single-character type parameter names.

adriaanm:
We can discuss longer names, but I would venture to say it's uncontroversial that `K` conveys meaning closer to what you'd expect than `A`, so let's start here.

PR #5267, opened 2016-07-07
Deprecate @remote

lrytz:
Also a few minor cleanups in the backend.


soc:
What is the issue with `@remote`?

lrytz:
i was assuming nobody is using that - i haven't discussed this with anyone, i might be completely wrong :)

it's the kind of annotations-compiler-magic that we're moving away from for a while (`@cloneable`, `@serializable`)

soc:
Ok, agree. I'm slightly concerned overall with the introduction of new dependencies on Java types, but I can see that this is just following the direction into which we are going already.

adriaanm:
+1 on the deprecation -- I very much doubt anyone was using this

adriaanm:
LGTM

PR #5268, opened 2016-07-07
Remove eager warning when unneeded transitive dependencies are absent from the classpath

johnynek:
This addresses: https://issues.scala-lang.org/browse/SI-9673

This issue becomes acute for people using https://github.com/bazelbuild/bazel and warnings as errors. In bazel, the compile and runtime classpaths are distinct, and a goal is to minimize the compile classpath to the absolute minimum that is needed to compile the target.

According to @retronym (I seem to recall some chat/message I can't find now) this warning is safe to ignore and if it becomes a real issue, there will be an error (which I have also seen in cases where the class is actually required).


johnynek:
Note, huge props to @non who paired with me on this!

johnynek:
I signed the CLA, but somehow it is not showing. Any way to refresh?

johnynek:
actually, I'm not sure why we would ever want this warning. It seems like we should either error if we need the class or stay quiet. Is there a reason we shouldn't just remove the warning branch here?

retronym:
What does javac do in the same situation?

johnynek:
@retronym javac is silent:

```
// A.java
class A {

}
// B.java
class B {
  public int work(A a) {
    return 0;
  }
}
// C.java
class C {
  public void work(B b) { }
}
// C.scala
class C {
  def work(b: B): Unit = ()
}
```

now run:

``` sh
#!/bin/sh

javac A.java
javac -cp . B.java
rm A.class
javac -cp . C.java

rm C.class
scalac -cp . C.scala
```

and note the only output is from the last command:

```
warning: Class A not found - continuing with a stub.
one warning found
```

som-snytt:
Other options to consider are to emit under `-Xdev` or maybe `-Ydebug`?

retronym:
I think I'm convinced. Let's disable this by warning, but enable it under any of `-Xdev` or `-verbose` (`if (settings.verbose || settings.developer) warn`)

(`settings.developer` is true if either `-Xdev` or `-Ydebug` is set)

johnynek:
@retronym here is your suggestion. Did I get it right? Seems to compile for me.

retronym:
Could you leave `!settings.isScaladoc` to be on the safe side in case someone is running with `scaladoc -verbose` for some reason?

retronym:
@johnynek Assuming the last commit addresses the test failures (if not, `sbt "partest --update-check"` ought to do update any remaining `.check` files), could you please squash to a single commit?

johnynek:
Not really sure what failed here. Looks like some build related issues.

johnynek:
@retronym what does the combined failure mean? Is this ready?

retronym:
We require all commits in the pull request to pass CI, not just the last. GitHub doesn't directly support this so we add an extra status to the last commit to track whether or not all previous commits have green ticks.

The combined error points to the fact that 9d1489 did not pass validation.

The [failure](https://scala-ci.typesafe.com/job/scala-2.12.x-validate-publish-core/2875/console) there looks like a build infrastructure problem:

```
++ /usr/local/bin/sbt -sbt-version 0.13.11 --warn 'setupPublishCore https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/' generateBuildCharacterPropertiesFile
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=512M; support was removed in 8.0
[warn] /home/jenkins/workspace/scala-2.12.x-validate-publish-core@2/src/build/genprod.scala:302: Selecting value MAX_ARITY from object genprod, which extends scala.DelayedInit, is likely to yield an uninitialized value
[warn]     if (i <= MAX_ARITY / 2) str(mdefs)
[warn]              ^
[warn] one warning found
[warn] The `-` command is deprecated in favor of `onFailure` and will be removed in 0.14.0
[error] Expected letter
[error] Expected symbol
```

I've seen that before but can't put my finger on the root cause. It's something like we've got the version of the SBT launcher script. I'll investigate, apologies for the friction.

retronym:
Trying again in #5297

PR #2666, opened 2013-06-20
SI-7265 General test for spec version

som-snytt:
```
The test for isJavaAtLeast uses the specification.version.

The method argument must have the form "major.minor".

The scaladoc is updated to reflect the new reality and a
test is added under junit.

Note that this implementation aims to be a simple
compromise between the functional and imperative camps,
that is, to be free of both closures and while loops.
And to elicit no cruft like regexes and wrappers for strings.
No doubt even more could be done in this department, but we
don't wish to spoil the fun on codegolf.stackexchange.com.
However, we might decide to sponsor a new site:
codereviewpingpong.com

For 2.10.x, javaSpecVersion is provided as a private member.
The active test is under `run` and the `junit` test must
bide its time in `pending`.

For 2.11, the private members can be public and the app test
replaced with the unit test.
```

Review by @soc and @adriaanm

PR #2580, opened 2013-05-22
SI-7265 Buy some time to fix s.u.Properties.isJavaAtLeast

soc:

PR #5270, opened 2016-07-08
SI-9750 scala.util.Properties.isJavaAtLeast will not work with JDK9

golem131:
- Support JEP-223
- Tests update

### DIFF
--- a/src/library/scala/util/Properties.scala
+++ b/src/library/scala/util/Properties.scala
@@ -13,6 +13,28 @@ package util
 import java.io.{ IOException, PrintWriter }
 import java.util.jar.Attributes.{ Name => AttributeName }
 
+sealed abstract class JavaSpecVersion extends Ordered[JavaSpecVersion] {
+  import scala.math.Ordered.orderingToOrdered
+    def compare(that: JavaSpecVersion): Int = (this, that) match {
+      case (x: JavaSpecVersion9AndBeyond, y: JavaSpecVersionBefore9) => 1
+      case (x: JavaSpecVersionBefore9, y: JavaSpecVersion9AndBeyond) => -1
+      case (x: JavaSpecVersion9AndBeyond, y: JavaSpecVersion9AndBeyond) => x.major - y.major
+      case (x: JavaSpecVersionBefore9, y: JavaSpecVersionBefore9) => (x.major, x.minor) compare (y.major, y.minor)
+    }
+}
+
+case class JavaSpecVersionBefore9(major: Int, minor: Int) extends JavaSpecVersion
+case class JavaSpecVersion9AndBeyond(major: Int) extends JavaSpecVersion
+
+
+object JavaSpecVersionBefore9 {
+  val pattern = "1.([5-8])"r
+}
+
+object JavaSpecVersion9AndBeyond {
+  val pattern = "(9|1[0-9])"r
+}
+
 /** Loads `library.properties` from the jar. */
 object Properties extends PropertiesTrait {
   protected def propCategory    = "library"
@@ -181,14 +203,19 @@ private[scala] trait PropertiesTrait {
    * }}}
    */
   def isJavaAtLeast(version: String): Boolean = {
-    def parts(x: String) = {
-      val i = x.indexOf('.')
-      if (i < 0) throw new NumberFormatException("Not a version: " + x)
-      (x.substring(0, i), x.substring(i+1, x.length))
+    val parsedJavaSpecVersion: JavaSpecVersion = parseVersion(javaSpecVersion)
+    val parsedVersion: JavaSpecVersion = parseVersion(version)
+    parsedJavaSpecVersion >= parsedVersion
+
+  }
+
+  def parseVersion(version: String): JavaSpecVersion = version match {
+    case JavaSpecVersion9AndBeyond.pattern(versionVal) => versionVal.toInt match {
+      case x if x >= 9 => JavaSpecVersion9AndBeyond(x)
+      case _ => throw new NumberFormatException("Not a version: " + version)
     }
-    val (v, _v) = parts(version)
-    val (s, _s) = parts(javaSpecVersion)
-    s.toInt >= v.toInt && _s.toInt >= _v.toInt
+    case JavaSpecVersionBefore9.pattern(versionVal) => JavaSpecVersionBefore9(1, versionVal.toInt)
+    case _ => throw new NumberFormatException("Not a version: " + version)
   }
 
   // provide a main method so version info can be obtained by running this
--- a/test/junit/scala/util/SpecVersionTest.scala
+++ b/test/junit/scala/util/SpecVersionTest.scala
@@ -21,6 +21,26 @@ class SpecVersionTest {
     override lazy val scalaProps = new java.util.Properties
   }
 
+  val sut9 = new PropertiesTrait {
+    override def javaSpecVersion = "9"
+
+    override protected def pickJarBasedOn: Class[_] = ???
+    override protected def propCategory: String = "test"
+
+    // override because of vals like releaseVersion
+    override lazy val scalaProps = new java.util.Properties
+  }
+
+  @Test
+  def comparesJDK9Correctly(): Unit = {
+    assert(!(sut isJavaAtLeast "9"))
+    assert(sut9 isJavaAtLeast "1.5")
+    assert(sut9 isJavaAtLeast "1.6")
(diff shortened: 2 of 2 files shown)


golem131:
I have signed the Scala CLA, but check not changed (

som-snytt:
That's a lot of infrastructure just to compare a number. Also, it leaks details.

Last time, @soc objected to regex, with the idea that this check should be absolutely minimal (and not require dependencies).

I'd suggest keeping the current code and just strip the optional "1." prefix, perhaps complaining if the suffix is not in "0-8".

src/library/scala/util/Properties.scala
```diff
@@ -13,6 +13,28 @@ package util
...
+      case (x: JavaSpecVersionBefore9, y: JavaSpecVersionBefore9) => (x.major, x.minor) compare (y.major, y.minor)
+    }
+}
+
+case class JavaSpecVersionBefore9(major: Int, minor: Int) extends JavaSpecVersion
+case class JavaSpecVersion9AndBeyond(major: Int) extends JavaSpecVersion
+
+
+object JavaSpecVersionBefore9 {
+  val pattern = "1.([5-8])"r
```
som-snytt, line 31: postfix is warnable.

src/library/scala/util/Properties.scala
```diff
@@ -13,6 +13,28 @@ package util
...
+  import scala.math.Ordered.orderingToOrdered
+    def compare(that: JavaSpecVersion): Int = (this, that) match {
+      case (x: JavaSpecVersion9AndBeyond, y: JavaSpecVersionBefore9) => 1
+      case (x: JavaSpecVersionBefore9, y: JavaSpecVersion9AndBeyond) => -1
+      case (x: JavaSpecVersion9AndBeyond, y: JavaSpecVersion9AndBeyond) => x.major - y.major
+      case (x: JavaSpecVersionBefore9, y: JavaSpecVersionBefore9) => (x.major, x.minor) compare (y.major, y.minor)
+    }
+}
+
+case class JavaSpecVersionBefore9(major: Int, minor: Int) extends JavaSpecVersion
```
Ichoran, line 26: The major version here is always 1, so this can be replaced by a single `Int`.

src/library/scala/util/Properties.scala
```diff
@@ -13,6 +13,28 @@ package util
 import java.io.{ IOException, PrintWriter }
 import java.util.jar.Attributes.{ Name => AttributeName }
 
+sealed abstract class JavaSpecVersion extends Ordered[JavaSpecVersion] {
```
Ichoran, line 16: Too much machinery to compare two integers.  That's all that's really happening.

src/library/scala/util/Properties.scala
```diff
@@ -181,14 +203,19 @@ private[scala] trait PropertiesTrait {
    * }}}
    */
   def isJavaAtLeast(version: String): Boolean = {
-    def parts(x: String) = {
-      val i = x.indexOf('.')
-      if (i < 0) throw new NumberFormatException("Not a version: " + x)
-      (x.substring(0, i), x.substring(i+1, x.length))
```
som-snytt, line 187: e.g. just ignore first part if present and equals 1 and second part < 9. Maybe rename `parts` to `specOf`...

Ichoran:
Can't this whole thing just be

``` scala
def javaVersion(s: String) = s.split("\\.")(if (s.startsWith("1.")) 1 else 0).toInt
```

There admittedly isn't much error checking here, but do we need any more logic than that?

src/library/scala/util/Properties.scala
som-snytt, line 16: If you put it that way, the only thing anyone ever does is compare bits. :8ball:

som-snytt:
I was going to say `s.stripPrefix("1.")` if no checking. But, yeah. I think `split` is the regex one, right? I remember paulp said that on the old PR.

src/library/scala/util/Properties.scala
Ichoran, line 16: Well, yes, but the combination of the bits is what is interesting.  In this case, both sets of bits represent integers, and it's normal integer comparison that is wanted.

Ichoran:
Are we playing the no-regex but yes-StringOps-extensions game?  Okay.

``` scala
def javaVersion(s: String) = s.stripPrefix("1.").takeWhile(_ != '.').toInt
```

Can use `(if (s startsWith "1.") s substring 2 else s)` in place of `stripPrefix` if you want to stay with Java methods only.  (Why are we playing this game again?)

som-snytt:
I noticed a second ago that in fact, last time, we said no extensions either, so, jinx.

I think @soc plays with other platforms where he has to boot in 16KB of ram.

golem131:
@Ichoran the problem of your recommended decision is not in Java System API, but in user input.
For example, what would happen in your version, when programer send to minimal version "10.1" or "11.1" or ".1" ?

Ichoran:
@golem131 - `10.1` returns 10.  `11.1` returns 11.  That's as desired, no?  `.1` throws an exception.  It's a `NumberFormatException` instead of one that explains what you did wrong, hence my comment about "not much error checking".  If someone entered `1.9` it would give `9`, but I'm not sure that's a bug instead of a feature.  What would they mean other than Java 9 (unless it's just a typo, in which case 1.7 typoed as 1.8 is just as bad but can't be caught).

golem131:
@Ichoran - Ok, ok. So, what decision? We try to create fastest version, simplest version, correct version (more secure checks, more custom exceptions etc.), or we say, that "version string" is not just sequence of numbers, and may be more complex ("9-ea+73" or "9.1.2+62" for example)?

Forgot to say, that we have sbt-nsa sbt plugin (https://github.com/mentat-labs/sbt-nsa) with type safe versions and compiler options. Do you think this plugin have wrong basic idea?

Ichoran:
For libraries that want to extract all the information available in a version string you have different requirements than a routine that merely needs to tell "is this 9 or is this 8".

Realistically, it should handle all the examples in http://openjdk.java.net/jeps/223 which means it can't stop on `.` alone, but making a decent effort to get the major version number should be enough.

For instance, getting the major version number with this code would, I think, be enough, though it could be rewritten with while loops and such if we wanted to make it easier for Proguard and the like to chop out as much as possible:

``` scala
// Java version string old format is 1.x.stuff
// Java version string new format is x[.-+u]stuff
// This handles both (and admits mixtures of the two)
def javaVersion(s: String): Int =
  try { s.stripPrefix("1.").takeWhile(c => !(".-+u" contains c.toString)).toInt }
  catch { case nfe: NumberFormatException => 
    throw new NumberFormatException("Could not locate major Java version number in " + s)
  }
```

Then all the rest of the logic can be done with just an `Int`.

golem131:
@Ichoran ok, no problem, i'll fix my PR after weekend.

Ichoran:
For reference, a while loop version would look something like this:

``` scala
def javaVersion(s: String): Int =
  try {
    val i = if (s startsWith "1.") 2 else 0
    var j = i
    while (j < s.length && !(".-+u" contains s.substring(j, j+1))) j += 1
    s.substring(i, j).toInt
  }
  catch { case nfe: NumberFormatException =>
    throw new NumberFormatException("Could not locate major Java version number in " + s)
  }
```

som-snytt:
The previous long conversation:

https://github.com/scala/scala/pull/2580

https://github.com/scala/scala/pull/2666

In which the paulpian knot is cut by defining the method as a spec version test, 1.1, 1.8, 9, 10.

IIUC, this conversation is suggesting to return to impl version tests? That wouldn't be my choice.

golem131:
@som-snytt then what's your choice?

som-snytt:
I would stick with spec version tests. Fix the Scaladoc if it says different. That's all it used to do:

```
def isJavaAtLeast(version: String) = {
  val okVersions = version match {
    case "1.5"    => List("1.5", "1.6", "1.7")
    case "1.6"    => List("1.6", "1.7")
    case "1.7"    => List("1.7")
    case _        => Nil
  } 
  okVersions exists (javaVersion startsWith _)
}
```

It ought to have been, `okVersions exists (javaSpecVersion == _)`.

Moral equivalent, `javaSpecVersion.stripPrefix("1.").toInt >= version.stripPrefix("1.").toInt`.

Also, I just re-remembered that `s.split("\\.")` doesn't use regex although it's required in general.

SethTisue:
superseded by #5276

PR #5225, opened 2016-06-12
Group math package functions

janekdb:
Group math package functions

Groups
- Mathematical Constants
- Minimum and Maximum
- Rounding
- Exponential and Logarithmic
- Trigonometric
- Angular Measurement Conversion
- Hyperbolic
- Absolute Values
- Signs
- Root Extraction
- Polar Coordindates
- Unit of Least Precision

Other changes,
- Dropped use of `double` for `Double` in all cases
- Grouped some methods in the source
- Extended notes about exception to method forwarding
- Minor method documentation enhancements

IEEERemainder is in the Rounding group since it is related to rounding and did not justify a new group.
## Review Snapshot

The Scaladoc is published here for review purposes: http://janekdb.github.io/scala/PR-5225/scala/math/index.html

PR #5271, opened 2016-07-08
SI-9813 Add Java 1.6. and 1.8 java.lang.Math methods into scala.math

golem131:
Add Java 1.6. and 1.8 java.lang.Math methods into scala.math.package

### DIFF
--- a/src/library/scala/math/package.scala
+++ b/src/library/scala/math/package.scala
@@ -199,4 +199,96 @@ package object math {
 
   /** Returns the size of an ulp of the given `Float` value. */
   def ulp(x: Float): Float = java.lang.Math.ulp(x)
+
+  /** Returns the sum of its arguments, throwing an exception if the result overflows an `Int`. */
+  def addExact(x: Int, y: Int): Int = java.lang.Math.addExact(x, y)
+
+  /** Returns the sum of its arguments, throwing an exception if the result overflows a `Long`. */
+  def addExact(x: Long, y: Long): Long = java.lang.Math.addExact(x, y)
+
+  /** Returns the difference of the arguments, throwing an exception if the result overflows an `Int`. */
+  def subtractExact(x: Int, y: Int): Int = java.lang.Math.subtractExact(x, y)
+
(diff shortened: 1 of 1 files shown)


golem131:
Do we need some tests?

janekdb:
This was merged recently: #5225

Kindly rebase and decide if you want maintain the Scaladoc groupings. If not I'll be happy to extend the groupings following the merge. Would be best to maintain the groupings as we go.

golem131:
I have updated my pull request.
Are this ok?

soc:
I'm a bit concerned about the copying of the ScalaDoc from OpenJDK. Can someone comment on whether this is acceptable?

mpociecha:
Please squash commits.

Ichoran:
I agree with @soc that copying the docs verbatim could be problematic.  Also, some of them would be a lot clearer rewritten (e.g. "scalb" where the exponentiation didn't come through).

golem131:
@soc so, just remove them at all, or rewrite them or what?

adriaanm:
I'm going to have to close this PR without looking at it, due to the risk entailed by copy/pasting docs from a codebase that's governed by a license that's not compatible with ours.

🚨  The prospect of violating another project's license is terrifying to an open source project! Please, _never_ derive docs or code and submit it for inclusion under our license unless the original license allows that. When you sign the CLA, that's one of the things you promise to do. 🚨

janekdb:
@golem131 - I suggest you create a new PR which contains only the forwarding to the Java methods and suitable Scaladoc groupings, similar to the exact group. Leave the methods without descriptions on the basis that the person looking at the methoda will be know what they are looking at because they a familiar with numerical computing or they are able to infer they should read the JavaDocs.

When I was looking at some existing undocumented methods earlier I realised it would be impossible for a non-expert to document the methods with original text so decided documentation omission was less damaging that partial documentation.

golem131:
@janekdb please check https://github.com/scala/scala/pull/5271

SethTisue:
the replacement PR is #5314

PR #5273, opened 2016-07-12
SI-9855 Fix regression in extractor pattern translation

retronym:
In faa5ae6, I changed the pattern matchers code generator to
use stable references (`Ident`-s with the singleton type, rather
than the widened type) to the synthetic vals used to store
intermediate results ("binders").

In the case where the scrutinee matched the unapply parameter
type of some extractor pattern, but the pattern subsequently
failed, this led to an regression.

It turns out that this was due to the way that the type of
the binder was mutated to upcast to the exact type of a subsequent
pattern in `ensureConformsTo`:

   https://github.com/scala/scala/blob/953559988/src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala#L165-L174

This was added in 32c57329a as a workaround for the problem caused
in t6664.scala, when the binder type was `KList with KCons`, and
the code generator wasn't able to find the case field accessors
for `KCons` in the decls.

The change to use stable references meant that this mutation was
now observed in another part of the tree, as opposed to the 2.11.8
situation, where we had used the original, sharper type of the binder
eagerly to assign to the `Ident` that referred to it. This led to
a tree:

```
Assign(Ident(x3), Ident(x1).setType(x1.tpe)
```

Now that we instead refer generate:

```
Assign(Ident(x3), Ident(x1).setType(stableTypeFor(x1))
```

and we don't typecheck this until after the mutation of `x1.symbol.info`,
we can get a type error.

This commit removes this mutation of the binder type altogether, and
instead uses `aligner.wholeType`, which is based on the result type of
the `Apply(TypeTree(MethodType(params, resultType))` that encodes a
typechecked constructor pattern. In `t6624.scala`, this is `KCons`,
the case class that has the extractors as its decls.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala
+++ b/src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala
@@ -125,7 +125,7 @@ trait MatchTranslation {
         // TODO: paramType may contain unbound type params (run/t2800, run/t3530)
         val makers = (
           // Statically conforms to paramType
-          if (this ensureConformsTo paramType) treeMaker(binder, false, pos) :: Nil
+          if (tpe <:< paramType) treeMaker(binder, false, pos) :: Nil
           else typeTest :: extraction :: Nil
         )
         step(makers: _*)(extractor.subBoundTrees: _*)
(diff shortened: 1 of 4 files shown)


retronym:
Review by @adriaanm

adriaanm:
LGTM -- so long ensureConformsTo 👋

PR #5276, opened 2016-07-13
SI-9750 scala.util.Properties.isJavaAtLeast works with JDK9

som-snytt:
The utility method compares javaSpecVersion, which has the
form "1.8" previously and "9" going forward.

The method accepts "1.n" for n < 9. More correctly, the string
argument should be a single number.

Supports JEP-223.

Supersedes https://github.com/scala/scala/pull/5270

Review by @golem131

This supposes golem was annoyed by our preciousness.

### DIFF
--- a/src/library/scala/util/Properties.scala
+++ b/src/library/scala/util/Properties.scala
@@ -168,33 +168,59 @@ private[scala] trait PropertiesTrait {
 
   /** Compares the given specification version to the specification version of the platform.
    *
-   * @param version a specification version number (legacy forms acceptable)
-   * @return `true` iff the specification version of the current runtime
-   * is equal to or higher than the version denoted by the given string.
-   * @throws NumberFormatException if the given string is not a version string
+   *  @param version a specification version number (legacy forms acceptable)
+   *  @return `true` if the specification version of the current runtime
+   *    is equal to or higher than the version denoted by the given string.
+   *  @throws NumberFormatException if the given string is not a version string
    *
-   * @example {{{
-   * // In this example, the runtime's Java specification is assumed to be at version 8.
-   * isJavaAtLeast("1.6")            // true
-   * isJavaAtLeast("1.8")            // true
-   * isJavaAtLeast("8")              // true
-   * isJavaAtLeast("9")              // false
-   * isJavaAtLeast("1.9")            // throws
-   * }}}
+   *  @example {{{
+   *  // In this example, the runtime's Java specification is assumed to be at version 8.
+   *  isJavaAtLeast("1.8")            // true
+   *  isJavaAtLeast("8")              // true
+   *  isJavaAtLeast("9")              // false
+   *  isJavaAtLeast("9.1")            // false
+   *  isJavaAtLeast("1.9")            // throws
+   *  }}}
    */
   def isJavaAtLeast(version: String): Boolean = {
-    def versionOf(s: String): Int = s.indexOf('.') match {
-      case 1 if s.charAt(0) == '1' =>
-        val v = s.substring(2).toInt
-        if (v < 9) v else -1
-      case -1 => s.toInt
-      case _  => -1
+    def versionOf(s: String, depth: Int): (Int, String) =
+      s.indexOf('.') match {
+        case 0 =>
+          (-2, s.substring(1))
+        case 1 if depth == 0 && s.charAt(0) == '1' =>
+          val r0 = s.substring(2)
+          val (v, r) = versionOf(r0, 1)
+          val n = if (v > 8 || r0.isEmpty) -2 else v   // accept 1.8, not 1.9 or 1.
+          (n, r)
+        case -1 =>
+          val n = if (!s.isEmpty) s.toInt else if (depth == 0) -2 else 0
+          (n, "")
+        case i  =>
+          val r = s.substring(i + 1)
+          val n = if (depth < 2 && r.isEmpty) -2 else s.substring(0, i).toInt
+          (n, r)
+      }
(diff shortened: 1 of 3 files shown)


src/library/scala/util/Properties.scala
```diff
@@ -168,27 +168,31 @@ private[scala] trait PropertiesTrait {
...
    */
   def isJavaAtLeast(version: String): Boolean = {
-    def parts(x: String) = {
-      val i = x.indexOf('.')
-      if (i < 0) throw new NumberFormatException("Not a version: " + x)
-      (x.substring(0, i), x.substring(i+1, x.length))
+    def versionOf(s: String): Int = s.indexOf('.') match {
+      case 1 if s.charAt(0) == '1' =>
+        val v = versionOf(s.substring(2))
+        if (v < 9) v else -1
```
som-snytt: Probably didn't intend to accept `1.1.8`.

src/library/scala/util/Properties.scala
```diff
@@ -168,27 +168,31 @@ private[scala] trait PropertiesTrait {
...
+      case -1 => s.toInt
+      case _  => -1
     }
-    val (v, _v) = parts(version)
-    val (s, _s) = parts(javaSpecVersion)
-    s.toInt >= v.toInt && _s.toInt >= _v.toInt
+    val v = versionOf(version)
+    if (v < 0) throw new NumberFormatException(s"Not a version: $version")
```
som-snytt: Is that the clearest message? `Not a version: 1.9`. Use `assertThrows` to verify the message in a test.

src/library/scala/util/Properties.scala
```diff
@@ -168,27 +168,31 @@ private[scala] trait PropertiesTrait {
...
+   * isJavaAtLeast("9")              // false
+   * isJavaAtLeast("1.9")            // throws
    * }}}
    */
   def isJavaAtLeast(version: String): Boolean = {
-    def parts(x: String) = {
-      val i = x.indexOf('.')
-      if (i < 0) throw new NumberFormatException("Not a version: " + x)
-      (x.substring(0, i), x.substring(i+1, x.length))
+    def versionOf(s: String): Int = s.indexOf('.') match {
```
varming: [JEP223](http://openjdk.java.net/jeps/223) specify that the system property java.specification.version is a $VNUM and $VNUMs satisfy the regular expression: `^[1-9][0-9]*(((\.0)*\.[1-9][0-9]*)*)*$`. This code doesn't look ready to pick the major version from all strings satisfying that regular expression.

som-snytt: @varming You're right, it does say $VNUM. I was fooled by the table of examples where the spec is $MAJOR? Is the spec $MAJOR in practice? I think it can't hurt to accept leading "1." and then ignore suffix of ".anything". The doc should say the arg must be the $MAJOR, if that is the desired behavior. Certainly previous usage has been just to distinguish major versions of the platform.

som-snytt: Actually, re-reading what they intend by $MINOR, which includes revision of API, it's possible this API should support $MAJOR.$MINOR. I know I care if we're on java 8 or 9, but will I care if it is 9.1 or 9.2? assuming the spec string is 9.1. Alternatively, maybe supply `isJavaMajorSpec` for platforms that don't want to use regex for this test; otherwise just use regex like normal people.

varming: Dear @som-snytt,

I don't know if there will be a time where $minor, $security, and later parts of the $VNUM will be important, but I would expect the code to work with all $VNUMs. In most settings you probably want to support $major, $minor, and $security. Having special functions to check that the major is at least greater than some integer seems like a good idea.

Right now the code handles backwards compatibility with <= JDK8, but it cannot handle JDK10. It would be nice if it handled backwards compatibility separately from general $VNUMs and let backwards compatibility take precedence. It doesn't look like you need full regular expressions to parse a $VNUM, simply take all characters up to the next dot (or the end of the string if there is no more dots) and parse them as an integer, and repeat this until you can decide the version question.

som-snytt:
By coincidence, there's a performance regression on start-up in JDK-9 that has to do with loading regex classes, fixed by checking if the version is just a major number.  [The call for review.](http://mail.openjdk.java.net/pipermail/core-libs-dev/2016-July/042538.html)

After the nuclear apocalypse, the only things left will be cockroaches and programs that do `s.indexOf('.')`.

retronym:
Looks like we can actually remove the only use of this method from our own code. Would you mind doing that too? We know statically that `isJavaAtLeast("1.7")` is true.

```
val suppressable = isJavaAtLeast("1.7")
...
      if (suppressable) {
        import scala.language.reflectiveCalls
        type Suppressing = { def getSuppressed(): Array[Throwable] }
        for (s <- e.asInstanceOf[Suppressing].getSuppressed) print(s, Suppressed, frames, indents + 1)
      }
```

som-snytt:
Sure thing, it gave me something to do while babysitting my second Windows 10 upgrade this week, with a few hours to spare.

I only wish that guy would comment his code. Frankly, I never want to touch his code ever again. I don't see how anyone puts up with it.

Technically, `isJavaAtLeast(Int)` still calls `isJavaAtLeast(String)`.

SethTisue:
shouldn't this target 2.11.x?

som-snytt:
@SethTisue Too risky!

retronym:
LGTM. I think we can get target Java 9 support to the 2.12.1+, and then backport to 2.11.x once some problems are shaken out.

PR #5279, opened 2016-07-14
SD-183 Make refinement classes ineligible as SAMs

retronym:
Only non-refinement class types need apply, which is the same
restriction that we levy on parent types of a class.

```
scala> class C; class D extends C; type CD = C with D; class E extends CD
<console>:11: error: class type required but C with D found
       class C; class D extends C; type CD = C with D; class E extends CD
                                                                       ^

scala> class C; class D extends C; type DC = D with C; class E extends DC
<console>:11: error: class type required but D with C found
       class C; class D extends C; type DC = D with C; class E extends DC
                                                                       ^
```

Prior to this change:

```
scala> trait T { def t(a: Any): Any }; trait U; abstract class C extends T
defined trait T
defined trait U
defined class C
```

For indy-based lambdas:

```
scala> val tu: T with U = x => x
tu: T with U = $$Lambda$1812/317644782@3c3c4a71

scala> tu: U
java.lang.ClassCastException: $$Lambda$1812/317644782 cannot be cast to U
  ... 30 elided
```

For anon class based lambdas:

```

scala> ((x => x): C with U)
<console>:14: error: class type required but C with U found
       ((x => x): C with U)
           ^

scala> implicit def anyToCWithU(a: Any): C with U = new C with U { def t(a: Any) = a }
anyToCWithU: (a: Any)C with U

scala> (((x: Any) => x): C with U) // SAM chosen but fails to typecheck the expansion uncurry
<console>:17: error: class type required but C with U found
       (((x: Any) => x): C with U) // SAM chosen but fails to typecheck the expansion uncurry
                  ^
```

Fixes https://github.com/scala/scala-dev/issues/183

While it is tempting to special case refinement classes with no decls by
flattening their parents into the parents of the lambda. But there are
some subtle issues at play with lineriazation order, as Martin pointed out
when I brought this up before: http://www.scala-lang.org/old/node/6817.html

### DIFF
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -837,9 +837,9 @@ trait Definitions extends api.StandardDefinitions {
      * The class defining the method is a supertype of `tp` that
      * has a public no-arg primary constructor.
      */
-    def samOf(tp: Type): Symbol = if (!doSam) NoSymbol else {
+    def samOf(tp: Type): Symbol = if (!doSam) NoSymbol else if (!isNonRefinementClassType(unwrapToClass(tp))) NoSymbol else {
       // look at erased type because we (only) care about what ends up in bytecode
-      // (e.g., an alias type or intersection type is fine as long as the intersection dominator compiles to an interface)
+      // (e.g., an alias type is fine as long as is compiles to a single-abstract-method)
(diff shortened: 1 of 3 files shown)


retronym:
Review by @adriaanm

adriaanm:
LGTM, thanks!

PR #5280, opened 2016-07-14
SI-8079 Only expand local aliases during variance checks

retronym:
We've been flip-flopping on this one through the years, right
now we issue an two errors for the enclosed test.

After this commit, variance validation only expands aliases that
are `{private,protected}[this]`. The rest need not be expanded,
as we have already variance validated the RHS of the alias.

It also removes a seemingly incorrect check to `isLocalOnly`.

This also means that we can use `@uncheckedVariance` to created
variant type aliases for Java interfaces.

However, if such a type alias is declared private local, it
_will_ be expanded. That shouldn't be a problem, other than
for the fact that we run through an as-seen-from that strips
the `@uV` annotations in the type expansion. This has been
recorded in a pending test.

### DIFF
--- a/src/reflect/scala/reflect/internal/Variances.scala
+++ b/src/reflect/scala/reflect/internal/Variances.scala
@@ -53,8 +53,7 @@ trait Variances {
     // return Bivariant if `sym` is local to a term
     // or is private[this] or protected[this]
     def isLocalOnly(sym: Symbol) = !sym.owner.isClass || (
-         sym.isTerm // ?? shouldn't this be sym.owner.isTerm according to the comments above?
-      && (sym.isLocalToThis || sym.isSuperAccessor) // super accessors are implicitly local #4345
+         (sym.isLocalToThis || sym.isSuperAccessor) // super accessors are implicitly local #4345
       && !escapedLocals(sym)
     )
 
@@ -66,7 +65,7 @@ trait Variances {
        *  Initially the state is covariant, but it might change along the search.
        *
        *  A local alias type is treated as Bivariant;
-       *  this is OK because we always expand aliases for variance checking.
+       *  this is OK because such aliases are expanded for variance checking.
        *  However, for an alias which might be externally visible, we must assume Invariant,
        *  because there may be references to the type parameter that are not checked,
        *  leading to unsoundness (see SI-6566).
@@ -126,7 +125,7 @@ trait Variances {
         tp match {
           case _ if isUncheckedVariance(tp)                    =>
           case _ if resultTypeOnly(tp)                         => this(tp.resultType)
-          case TypeRef(_, sym, _) if sym.isAliasType           => this(tp.normalize)
+          case TypeRef(_, sym, _) if shouldDealias(sym)        => this(tp.normalize)
           case TypeRef(_, sym, _) if !sym.variance.isInvariant => checkVarianceOfSymbol(sym) ; mapOver(tp)
           case RefinedType(_, _)                               => withinRefinement(mapOver(tp))
           case ClassInfoType(parents, _, _)                    => parents foreach this
@@ -137,6 +136,9 @@ trait Variances {
         // cloning during the recursion, it is important to return the input `tp`, rather
         // than the result of the pattern match above, which normalizes types.
         tp
+      }
+      private def shouldDealias(sym: Symbol): Boolean = {
+        sym.isAliasType && isLocalOnly(sym)
       }
       def validateDefinition(base: Symbol) {
         val saved = this.base
--- a/src/reflect/scala/reflect/internal/Variances.scala
+++ b/src/reflect/scala/reflect/internal/Variances.scala
@@ -50,9 +50,8 @@ trait Variances {
          sym.isParameter
       && !(tvar.isTypeParameterOrSkolem && sym.isTypeParameterOrSkolem && tvar.owner == sym.owner)
     )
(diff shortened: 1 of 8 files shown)


src/reflect/scala/reflect/internal/Variances.scala
```diff
@@ -53,8 +53,7 @@ trait Variances {
     // return Bivariant if `sym` is local to a term
     // or is private[this] or protected[this]
     def isLocalOnly(sym: Symbol) = !sym.owner.isClass || (
-         sym.isTerm // ?? shouldn't this be sym.owner.isTerm according to the comments above?
-      && (sym.isLocalToThis || sym.isSuperAccessor) // super accessors are implicitly local #4345
+         (sym.isLocalToThis || sym.isSuperAccessor) // super accessors are implicitly local #4345
```
retronym, line 55: - [x] Maybe rename to `isExemptFromVariance`

src/reflect/scala/reflect/internal/Variances.scala
```diff
@@ -138,6 +137,9 @@ trait Variances {
         // than the result of the pattern match above, which normalizes types.
         tp
       }
+      private def shouldDealias(sym: Symbol): Boolean = {
+        sym.isAliasType && isLocalOnly(sym)
```
retronym: - [x] Justify why we need to dealias for local only symbols:

```
trait C[-I] {
  private[this] type X = C[I]

  def f2(b: X): Unit
}

```

Because we didn't check `C[I]` in the RHS of the local only type alias, we need to expand the alias so we check it in the public signature of `f2`.

test/files/neg/t8079a.scala
```diff
@@ -0,0 +1,4 @@
+trait C[-I] {
+  private[this] type X = C[I]
+  def f2(b: X): Unit
```
retronym, line 3: - [x] Investigate why this doesn't give a variance error (even before this PR) when `X` is `protected[this]`.
  - I can't reproduce this problem (before or after). I've submitted a test for the `protected[this]` case to confirm this.

retronym:
- [x] I need to update a few checkfiles (NOPE, see comment below):

```
partest --update-check \
  /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/neg/names-defaults-neg-ref.scala \
  /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/neg/names-defaults-neg.scala \
  /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/run/collection-conversions.scala \
  /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/run/t1195-new.scala \
  /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/run/t4287inferredMethodTypes.scala
```

retronym:
Dropping the last commit (f8d7762) (that propagated `@uncheckedVariance` through the `AsSeenFrom` `TypeMap`, as it inexplicably changed behaviour of one test (!).

https://gist.github.com/0434e939f889223c20d272ce5342500c

The rest of this PR stands alone as a worthwhile improvement.

retronym:
Review by @adriaanm

retronym:
@adriaanm Needs a (trivial) rebase a la https://github.com/scala/scala/compare/2.12.x...retronym:rebase/5294?expand=1

adriaanm:
LGTM

adriaanm:
Looks like this will need a compensating action over at scalaz:

```
core/src/main/scala/scalaz/Id.scala:16: covariant type Id occurs in invariant position
 in type => scalaz.Traverse1[IdInstances.this.Id] with scalaz.Each[IdInstances.this.Id] with scalaz.Monad[IdInstances.this.Id] with scalaz.Comonad[IdInstances.this.Id] with scalaz.Distributive[IdInstances.this.Id] with scalaz.Zip[IdInstances.this.Id] with scalaz.Unzip[IdInstances.this.Id] with scalaz.Align[IdInstances.this.Id] with scalaz.Cozip[IdInstances.this.Id] of value id
   val id: Traverse1[Id] with Each[Id] with Monad[Id] with Comonad[Id] with Distributive[Id] with Zip[Id] with Unzip[Id] with Align[Id] with Cozip[Id] =
       ^
one error found
```

from https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/510/consoleFull

SethTisue:
> Looks like this will need a compensating action over at scalaz

patched in the community build: https://github.com/scala/community-builds/pull/265
reported upstream: https://github.com/scalaz/scalaz/issues/1249

mielientiev:
Hi @SethTisue @adriaanm 
I've faced with some problems in Scala 2.12 after this fix. 
Here's sample: https://gist.github.com/mielientiev/1873b2aea8f6f2bff81e13b597a7849c

In Scala 2.11 that code compiles. (due to bug in type checker). But now it fails because of type alias is in the contravariant position.
So what is the best solution for fixing this issue in 2.12?
I can add @uncheckedVariance but I'm not sure that's a good idea... Please correct me if I'm wrong

propensive:
Hi @mielientiev,
We can work around this by redefining the type in the last example from,
```
type CoVariant[+T, _ <: Exception] = T
```
to,
```
type CoVariant[+T, _ <: Exception] = T2 forSome { type T2 <: T }
```

mielientiev:
@propensive Yep. That's work.

thank you

PR #5281, opened 2016-07-14
Deprecate and rename Left#a/Right#b to Left#value/Right#value

dwijnand:
Inspired by #5275 and @fthomas.


soc:
Alright, just wanted to do the same. :-)
LGTM, assuming there a no weird serialization issues.

dwijnand:
Review by @szeiger?

PR #5282, opened 2016-07-15
SD-128 fix override checks for default methods

lrytz:
The check for inheriting two conflicting members was wrong for default
methods, leading to a missing error message.

Removes two methods:

isDeferredOrJavaDefault had a single use that is removed in this commit.

isDeferredNotJavaDefault is redundant with `isDeferred`: every default
method has the DEFERRED flag. For symbols originating in the classfile
parser this was the case from day one: abstract interface methods have
the JAVA_ACC_ABSTRACT flag in bytecode, which the classfile parser
translates to DEFERRED. Default methods don't have that classfile flag.
For symbols created by the Java source parser, we don't add the
DEFERRED to default methods anymore since 373db1e.


adriaanm:
I believe the commit message needs `s/every/no/`, so that "every default method has the DEFERRED flag." reads "no default method has the DEFERRED flag."

adriaanm:
Reworded in #5331

PR #5283, opened 2016-07-15
SD-182 compiler option -Xgen-mixin-forwarders

lrytz:
Introduce a compiler option -Xgen-mixin-forwarders to always generate
mixin forwarder methods.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Mixin.scala
+++ b/src/compiler/scala/tools/nsc/transform/Mixin.scala
@@ -246,43 +246,51 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
           case NoSymbol =>
             val isMemberOfClazz = clazz.info.findMember(member.name, 0, 0L, stableOnly = false).alternatives.contains(member)
             if (isMemberOfClazz) {
-              // `member` is a concrete method defined in `mixinClass`, which is a base class of
-              // `clazz`, and the method is not overridden in `clazz`. A forwarder is needed if:
-              //
-              //   - A non-trait base class of `clazz` defines a matching method. Example:
-              //       class C {def f: Int}; trait T extends C {def f = 1}; class D extends T
-              //     Even if C.f is abstract, the forwarder in D is needed, otherwise the JVM would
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -246,43 +246,51 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
...
+                      isCompeting || existsCompetingMethod(rest)
+                    } else existsCompetingMethod(rest)
+
+                  case _ => false
+                }
+
+                if (existsCompetingMethod(clazz.baseClasses))
+                  genForwarder()
+                else if (!settings.nowarnDefaultJunitMethods && JUnitTestClass.exists && member.hasAnnotation(JUnitTestClass))
+                  warning(member.pos, "JUnit tests in traits that are compiled as default methods are not executed by JUnit 4. JUnit 5 will fix this issue.")
```
sjrd, line 292: Maybe this warning could list the possible ways to fix the problem:
- Either use `-Xgen-mixin-forwarders` (with an impact on all classes). In sbt this can easily be applied only to the `Test` configuration with `scalacOptions in Test += "-Xgen-mixin-forwarders"`.
- Or use https://github.com/scala-js/scala-2.12-junit-mixin-plugin

?

lrytz, line 292: or we could just emit the forwarder..

sjrd, line 292: Yes, but in that case the test is not sufficient. `@Test` is not the only annotation for which forwarders are required. You also have `@Before`, `@After`, `@BeforeClass`, `@AfterClass` and `@Ignore`, as mentioned [here](https://github.com/scala-js/scala-2.12-junit-mixin-plugin#how-it-works).

retronym:
LGTM. We need this option around to support our benchmarking efforts for the compiler itself, so I'd like to merge it now and figure out the how to improve the wording of the JUnit warnings later.

PR #5285, opened 2016-07-15
Switch the bootstrap build over to sbt

szeiger:
All of the individual ant builds that occured during `bootstrap` are
replaced by equivalent sbt builds.
- Allow extra dashes in version suffix when using SPLIT
- Clean up ScriptCommands
- Building an extra `locker` for stability testing with ant was not
  necessary but sbt also drops `strap`, so we need to build again
  with `quick` to get the equivalent of `strap`. The script for checking
  stability is invoked directly from the bootstrap script, not from sbt.
- `STARR` and `locker` build output is still logged to `logs/builds`,
  the main build runs log directly to the main console with colored
  output.
- Allow `—show-log` option on partest command line in sbt
- Normalize inferred LUB in `run/t7747-repl.scala`
- Add `normalize` feature from `ReplTest` to `InteractiveTest`
- Normalize inferred LUBs in `presentation/callcc-interpreter`

### DIFF
--- a/test/files/presentation/callcc-interpreter/Runner.scala
+++ b/test/files/presentation/callcc-interpreter/Runner.scala
@@ -1,3 +1,6 @@
 import scala.tools.nsc.interactive.tests._
 
-object Test extends InteractiveTest
+object Test extends InteractiveTest {
+  // Normalize ordering of LUB
+  override def normalize(s: String) = s.replace("Serializable with Product", "Product with Serializable")
+}
--- a/project/PartestUtil.scala
+++ b/project/PartestUtil.scala
@@ -29,7 +29,7 @@ object PartestUtil {
     val knownUnaryOptions = List(
       "--pos", "--neg", "--run", "--jvm", "--res", "--ant", "--scalap", "--specialized",
       "--scalacheck", "--instrumented", "--presentation", "--failed", "--update-check",
-      "--show-diff", "--verbose", "--terse", "--debug", "--version", "--self-test", "--help")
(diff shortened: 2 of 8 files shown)


szeiger:
Bootstrap build of this version: https://scala-ci.typesafe.com/view/scala-2.12.0/job/scala-2.12.0-integrate-bootstrap/33/

test/files/presentation/callcc-interpreter/Runner.scala
```diff
@@ -1,3 +1,6 @@
 import scala.tools.nsc.interactive.tests._
 
-object Test extends InteractiveTest
\ No newline at end of file
+object Test extends InteractiveTest {
+  // Normalize ordering of LUB
+  override def normalize(s: String) = s.replace("Serializable with Product", "Product with Serializable")
```
adriaanm, line 5: Crafty 😎, but we should figure out what's going on here at some point.

adriaanm:
LGTM! Nice cleanups in there!

test/files/presentation/callcc-interpreter/Runner.scala
som-snytt, line 5: I thought it was just me!

retronym, line 5: It turned out to be non-determistic listing order of directories on, e.g. ext4, combined with the way that the `Symbol#id` is used a tie breaker for ordering elements of the base type sequence that have the same base type depth.

Fixed in https://github.com/scala/scala/pull/5263/commits/c7ec84f3b99b587c8dc60b4eda61ab2a939ab68e

PR #5287, opened 2016-07-15
SI-9691 BufferedIterator should expose a headOption

ChristopherDavenport:
This exposes a new API to the BufferedIterator trait. It will return the next element of an iterator as an Option.  The return will be  Some(value) if there is a next value, and None if there is not a next element.

This has been squashed and documentation put in place.  Was previously covered by #5277

### DIFF
--- a/src/library/scala/collection/BufferedIterator.scala
+++ b/src/library/scala/collection/BufferedIterator.scala
@@ -24,5 +24,11 @@ trait BufferedIterator[+A] extends Iterator[A] {
    */
   def head: A
 
+  /** Returns an option of the next element of an iterator without advancing beyond it.
+    * @return  the next element of this iterator if it has a next element
+    *           `None` if it does not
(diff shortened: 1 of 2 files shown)


szeiger:
LGTM. Congratulations to your first merged PR!

pathikrit:
👍

adriaanm:
Btw, this broke slick: https://github.com/slick/slick/blob/master/slick/src/main/scala/slick/util/ReadAheadIterator.scala#L26 (and thus the community build: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/456/consoleFull)

szeiger:
How ironic, but we don't really have any way to avoid this sort of problem. The only solution I can think of that allows cross-building Slick is to make `ReadAheadIterator.headOption` an extension method.

PR #5290, opened 2016-07-19
SD-48 limit the length of inlined local variable names

lrytz:
When inlining local variables, the names are prefixed with the callee
method name. In long chains of inlining, these names can grow
indefinitely. This commits introduces a limit.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala
@@ -324,15 +324,33 @@ object BytecodeUtils {
    * Clone the local variable descriptors of `methodNode` and map their `start` and `end` labels
    * according to the `labelMap`.
    */
-  def cloneLocalVariableNodes(methodNode: MethodNode, labelMap: Map[LabelNode, LabelNode], prefix: String, shift: Int): List[LocalVariableNode] = {
-    methodNode.localVariables.iterator().asScala.map(localVariable => new LocalVariableNode(
-      prefix + localVariable.name,
-      localVariable.desc,
-      localVariable.signature,
-      labelMap(localVariable.start),
-      labelMap(localVariable.end),
-      localVariable.index + shift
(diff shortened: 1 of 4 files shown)


szeiger:
Which problems do the long names cause? Is there a hard limit somewhere or is this just for practical reasons (like seeing them in a debugger)?

src/compiler/scala/tools/nsc/backend/jvm/opt/BytecodeUtils.scala
```diff
@@ -324,15 +324,33 @@ object BytecodeUtils {
...
+          // keep at least 5 characters per method name
+          val maxNumMethNames = BTypes.InlinedLocalVariablePrefixMaxLenght / 5
+          val usedMethNames =
+            if (methNames.length < maxNumMethNames) methNames
+            else {
+              val half = maxNumMethNames / 2
+              methNames.take(half) ++ methNames.takeRight(half)
+            }
+          val charsPerMethod = BTypes.InlinedLocalVariablePrefixMaxLenght / usedMethNames.length
+          usedMethNames.foldLeft("")((res, methName) => res + methName.take(charsPerMethod) + "_") + varName
```
szeiger, line 344: Should this be `charsPerMethod - 1` instead of `charsPerMethod` since you're adding an extra "_"?

lrytz, line 344: I was not trying to be precise about the limit, I expect the name to go a char or two beyond in some cases.

lrytz:
> Is there a hard limit somewhere or is this just for practical reasons (like seeing them in a debugger)?

It's for practical reasons. The name is a [`CONSTANT_Utf8_info`](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.13) in the classfile, which could be very long (the lenght is an `u2`, so i guess the max lenght is `(1 << 16) - 1`).

szeiger:
LGTM

PR #5291, opened 2016-07-19
SD-20 Include static methods in the InlineInfo in mixed compilation

lrytz:
In mixed compilation, the InlineInfo for a Java-defined class is created
using the class symbol (vs in separate compilation, where the info is
created by looking at the classfile and its methods). The scala compiler
puts static java methods into the companion symbol, and we forgot to
include them in the list of methods in the InlineInfo.


adriaanm:
LGTM

PR #5292, opened 2016-07-20
Switch Windows CI build to sbt (w/ some sbt build improvements) [ci: last-only]

szeiger:
Switch Windows CI build to sbt (w/ some sbt build improvements)
- Use sbt in `integrate/windows`: This essentially combines the Unix CI
  jobs `validate/publish-core` and `validate/test`, first publishing a
  local release built with STARR and then building a new version with
  that and running all tests on it.
- Unify repository handling across build scripts: A new function
  `generateRepositoriesConfig` in `common`, based on the existing code
  in `integrate/bootstrap`, writes the `repositories` file for sbt,
  either with or without an extra bootstrap repository for resolving a
  previously built version. It is used in all CI scripts to ensure that
  artifacts are only resolved through the sanctioned proxies and
  upstream repositories.
- The repository URL arguments in `setupPublishCore` and
  `setupValidateTest` are now optional as well. These commands are used
  without a URL from `integrate/windows`, which publishes to `local`
  instead of a temporary remote repository.
- `testAll` is now a task instead of a command. It runs the same
  sequence of sub-tasks as before but does not propagate failures
  immediately. It always runs all subtasks and reports errors at the
  end.
- The `generateBuildCharacterPropertiesFile` task now includes all
  properties from `versions.properties` (whose values have potentially
  been overwritten with `-D` options) in `buildcharacter.properties`.

---

Bootstrap build run of this version at https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/513/

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -769,6 +769,32 @@ lazy val root: Project = (project in file("."))
       GenerateAnyVals.run(dir.getAbsoluteFile)
       state
     },
+    testAll := {
+      val results = ScriptCommands.sequence[Result[Unit]](List(
+        (Keys.test in Test in junit).result,
+        (testOnly in IntegrationTest in testP).toTask(" -- run pos neg jvm").result,
+        (testOnly in IntegrationTest in testP).toTask(" -- res scalap specialized scalacheck").result,
+        (testOnly in IntegrationTest in testP).toTask(" -- instrumented presentation").result,
+        (testOnly in IntegrationTest in testP).toTask(" -- --srcpath scaladoc").result,
+        (Keys.test in Test in osgiTestFelix).result,
+        (Keys.test in Test in osgiTestEclipse).result,
+        (MiMa.mima in library).result,
+        (MiMa.mima in reflect).result,
+        Def.task(()).dependsOn( // Run these in parallel:
+          doc in Compile in library,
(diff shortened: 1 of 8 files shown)


szeiger:
Windows build of this version at https://scala-ci.typesafe.com/view/All/job/scala-2.12.x-integrate-windows/285/

szeiger:
Still causes problems with all subsequent Windows builds failing: https://scala-ci.typesafe.com/view/All/job/scala-2.12.x-integrate-windows/286/console. This will probably require changes to the scripts in jenkins-infra to perform the cleanup in a different way that doesn't fail.

dwijnand:
Cool improvements.

szeiger:
This should be good to merge. It broke the Windows build because `core.longpaths` was not set properly on the Windows worker. It is already part of our setup (https://github.com/scala/scala-jenkins-infra/blob/master/recipes/_worker-config-windows-cygwin.rb#L83-L89) but for some reason didn't work (or maybe hadn't been redeployed since it was introduced). @adriaanm now made the change manually on the worker.

adriaanm:
LGTM! sbt ftw 🎉

PR #5293, opened 2016-07-20
upgrade asm to 5.1

lrytz:


src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
```diff
@@ -258,7 +258,8 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
           coreBTypes.jliMethodTypeRef,
           ArrayBType(ObjectRef)),
         coreBTypes.jliCallSiteRef
-      ).descriptor)
+      ).descriptor,
+      /* itf = */ false)
```
varming: coreBTypes.jliLambdaMetafactoryRef.isInterface.get ?

src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
```diff
@@ -270,7 +271,8 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
           coreBTypes.jliMethodTypeRef
         ),
         coreBTypes.jliCallSiteRef
-      ).descriptor)
+      ).descriptor,
+      /* itf = */ false)
```
varming: coreBTypes.srLambdaDeserialize.isInterface.get ?

src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
```diff
@@ -418,13 +418,13 @@ class CallGraph[BT <: BTypes](val btypes: BT) {
     private val metafactoryHandle = {
       val metafactoryMethodName: String = "metafactory"
       val metafactoryDesc: String       = "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;"
-      new Handle(Opcodes.H_INVOKESTATIC, lambdaMetaFactoryInternalName, metafactoryMethodName, metafactoryDesc)
+      new Handle(Opcodes.H_INVOKESTATIC, lambdaMetaFactoryInternalName, metafactoryMethodName, metafactoryDesc, /* itf = */ false)
```
varming: Is it possible to use CoreBTypes.jliLambdaMetafactoryRef instead of creating a new reference to j.l.i.LambdaMedatafactory. ?

varming:
@lrytz LGTM. I have a similar branch and it works as intended.

src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
lrytz: ok, that looks cleaner

src/compiler/scala/tools/nsc/backend/jvm/opt/CallGraph.scala
lrytz: right.. we can actually get rid of the `metafactoryHandle` definition entirely!

lrytz:
@varming can you take another look?

varming:
LGTM.

PR #5294, opened 2016-07-20
Fields: expand lazy vals during fields, like modules 

adriaanm:
(follow up for #5141 )
# Fields phase fully expands lazy vals and modules

Mixins is relieved of its bitmap emission duties.
## Important changes
- the locking scope of a local lazy val is now tied to the lazy val, following Dotty's encoding
- specialized lazy vals work because the full expansion happens before specialization
## Internal details

Lazy vals remain `ValDef`s (with a method symbol for its accessor) until the fields phase, along with deferred vals and all trait vals. Code that previously only considered `DefDef`s when looking for accessors will need to widen this search to `ValOrDefDef`. Also, `accessed` will not lead to a field in these cases (ultimately, we should delay all field creation until fields.)

Lazy val accessors, fields and bitmaps are now mixed in during the fields phase, which means erasure and specialization get a better look at them. Should we also move mixins before erasure? 

@lrytz did the bulk of the work in moving our encoding of local lazy vals to the scheme used by Dotty, based on a suggestion by @DarkDimius. Locking is now down on the holder of the lazy val's value / init bit, instead of on the enclosing class (which had widened from the closure class to the owner of the method, making this more prone to deadlocks).

All lazy vals and modules use double-checked locking, but modules no longer spin off their initialization to a compute method (it's short enough not to interfere with the inlining of the lazy accessor).
## Notes from the implementation lab journal
- remove lazy accessor logic
  now that we have a single ValDef for lazy vals,
  with the underlying machinery being hidden until the fields phase
  leave a `@deprecated def lazyAccessor` for scala-refactoring
- don't skolemize in purely synthetic getters,
  but _do_ skolemize in lazy accessor during typers
  
  Lazy accessors have arbitrary user code, so have to skolemize.
  We exempt the purely synthetic accessors (`isSyntheticAccessor`)
  for strict vals, and lazy accessors emitted by the fields phase
  to avoid spurious type mismatches due to issues with existentials
  (That bug is tracked as scala/scala-dev#165)
  
  When we're past typer, lazy accessors are synthetic,
  but before they are user-defined to make this hack less hacky,
  we could rework our flag usage to allow for
  requiring both the ACCESSOR and the SYNTHETIC bits
  to identify synthetic accessors and trigger the exemption.
  
  see also scala/scala-dev#165
  
  ok 7 - pos/existentials-harmful.scala
  ok 8 - pos/t2435.scala
  ok 9 - pos/existentials.scala
  
  previous attempt: skolemize type of val inside the private[this] val
  
  because its type is only observed from inside the
  accessor methods (inside the method scope its existentials are skolemized)
- bean accessors have regular method types, not nullary method types
- must re-infer type for param accessor
  some weirdness with scoping of param accessor vals and defs?
- tailcalls detect lazy vals, which are defdefs after fields
- can inline constant lazy val from trait
- don't mix in fields etc for an overridden lazy val
- need try-lift in lazy vals: the assign is not seen in uncurry
  because fields does the transform (see run/t2333.scala)
- ensure field members end up final in bytecode
- implicit class companion method: annot filter in completer
- update check: previous error message was tangled up with unrelated
  field definitions (`var s` and `val s_scope`),
  now it behaves consistently whether those are val/vars or defs
- analyzer plugin check update seems benign, but no way to know...
- error message gen: there is no underlying symbol for a deferred var;  look for missing getter/setter instead
- avoid retypechecking valdefs while duplicating for specialize --  see pos/spec-private
- Scaladoc uniformly looks to field/accessor symbol
- test updates to innerClassAttribute by Lukas
### tests / review follow up
- [x] https://github.com/scala/scala/pull/5141#discussion_r74368912
- [x] scalacheck/quasiquotes  
- [x] presentation/t5708  
- [x] presentation/t8459  
- [x] run/t7767.scala  
- [x] scalacheck/HtmlFactoryTest.scala
- [x] SI-8873 `@volatile` var cannot be defined in case class with confusing error message
- [x] neg/t712.scala
- [x] run/existential-rangepos.scala
- [x] run/idempotency-case-classes.scala
- [x] run/reify_ann3.scala
- [x] run/reify_ann5.scala
- [x] run/repl-colon-type.scala
- [x] run/trait-fields-override-lazy.scala
- [x] run/analyzerPlugins.scala
- [x] neg/abstract-vars.scala  update users of `analyzer.underlyingSymbol` to reflect abstract val/vars don't have a field
- [x] pos/spec-private.scala
- [x] run/t8549.scala  (check bytecode of scala/collection/mutable/DoubleLinkedList.scala)
- [X] jvm/innerClassAttribute -- understand bytecode changes in jvm/innerClassAttribute/Classes_1.scala  https://gist.github.com/adriaanm/833837b204141a130abaaa6a62ed8596 (thanks, Lukas!) 
- [x] -Xcheckinit
- [x] look at bootstrapped bytecode diff
### community build
- [x] bring back lazyAccessor for scala-refactoring
- [x] scalaz (missing sig for implicit val at position after use site)
- [x] scala-js ExportsTest fail (valdefs may also contribute exported methods: deferred val/trait val/lazy val)
- [x] akka (existential in lazy val accessor)
### follow ups after RC1:
- failing test in macro-paradise
- rangepos


adriaanm:
I aborted the test run because the JDK was having trouble under high load.

lrytz:
I checked the `innerClassAttribute` test, here's the fix: https://github.com/lrytz/scala/commit/e685a9a90d34ed2f22a928c60269c413bfa24979. note that the diff I see is smaller than the one you link to in your comment. The only differences are in numbering lifted definitions and anonymous classes.

adriaanm:
Community build: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/474/console

retronym:
I've taken a breadth-first pass over this and haven't found any issues so far. I'll dive in deeper tomorrow.

lrytz:
@adriaanm I started looking at implementing https://github.com/scala/scala-dev/issues/133, but it's not clear to me what's the ultimate goal for the lazy val desugaring. Currently, the fields phase creates a variable and a getter for each lazy val. The lazyvals phase then splits up the getter into a getter and a `lzy$compute` method. Will this remain the same?

Also, there's still some duplication between lazyvals and mixin: both have a method `mkLazyDef`, the two implementations are basically the same. Will this also remain?

For https://github.com/scala/scala-dev/issues/133, the idea is to expand a local `lazy val v = init` into

```
val v$lzy = new LazyRef()
def v = {
  if (!v$lzy.initialized) v$lzy.synchronized {
    if (!v$lzy.initialized) { v$lzy.initialized = true; v$lzy.value = init }
  }
  v$lzy.value
}
```

Should all of this be done in the fields phase? Or during lazyvals?

adriaanm:
Thanks! My goal is to move the remaining lazy desugaring from mixin to lazyvals, so that there's no duplication. I have started poking around to convince myself that `Mixin`'s `mkLazyDef` isn't used anymore. I'm not sure I'll have time to move the logic that emits the init flags from Mixins to Lazyvals, but ultimately that would be the goal as well.

adriaanm:
Now I've paged back in where I left this: mixins handles lazy vals in classes and uses the same init bit logic as checkinit, so moving that to lazyvals could be tricky to do within the time budget for RC1.

lrytz:
yeah, just found out the same :)

lrytz:
so lazyvals only handles local lazy vals? then i think we can get rid of the phase and move the translation for locals to the fields phase. i'm happy to take a look at this, doesn't look too hard (famous last words).

adriaanm:
Ah, that's an interesting approach, hadn't thought of that.

adriaanm:
It looks like it's some strange mixture of the two (EDIT: local lazy vals and modules in classes --> both could be done during the fields phase) -- let's synch on slack how to divide the work here.

lrytz:
wip: https://github.com/lrytz/scala/commit/9f2f990ea478eddf74e265ab8602618b76c1c175

currently crashes in backend, will pick it up tomorrow.

```
warning: an unexpected type representation reached the compiler backend while compiling Test.scala: [T0]()T0. If possible, please file a bug on issues.scala-lang.org.
scala.MatchError: [T0]()T0 (of class scala.reflect.internal.Types$PolyType)
    at scala.tools.nsc.backend.jvm.BTypesFromSymbols.typeToBType(BTypesFromSymbols.scala:212)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.tpeTK(BCodeSkelBuilder.scala:79)
    at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genApply(BCodeBodyBuilder.scala:523)
```

lrytz:
@adriaanm here's the desugaring for local vals: https://github.com/lrytz/scala/commit/26fbaf6bf187a4728a6dd79d4740b4a65e270f5d. Note that this revision is not rebased on the current top of this PR.

I tried rebasing on top of this PR, the result is here: https://github.com/lrytz/scala/commit/3d56b1fa25ca34a7d072c39c7f10eb73af6f171d. But this revision doesn't pass the test I wrote.

There are a few TODOs in the commit, please take a look. A few more open questions:
- Can we remove the lazyvals phase? It seems it's still used for desugaring `object`s into a field, bitmap and lzycompute.
- Should we write the LazyHolders in Java? It would save getter / setter accesses.

How should we proceed?

src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
```diff
@@ -0,0 +1,479 @@
...
+        * If the class contains only a single lazy val then the bitmap is
+        * represented as a Boolean and the condition checking is a simple bool test.
+        *
+        * Private fields used only in this initializer are subsequently set to null.
+        */
+      private def mkLazySlowPathDef(lzyVal: Symbol, rhs: Tree): DefDef = {
+        val (init, retVal) =
+          rhs match {
+            case rhs if isUnitGetter(lzyVal) => (List(rhs), UNIT)
+            case Block(stats, res) => (stats, Select(thisRef, res.symbol))
```
lrytz: i have an example where `res.symbol` returns `null`, causing an NPE. will take a look

lrytz:
here's some more wip: https://github.com/lrytz/scala/commits/pr5294 things look a bit messy in the current stage. for example:

```
class C {
  def getInt: Int    = { println("getInt"); 1 }
  lazy val l = getInt
}
```

gives

```
  class C extends Object {
    final <synthetic> lazy private[this] var l: Int = _;
    def getInt(): Int = {
      scala.Predef.println("getInt");
      1
    };
    private def l$lzycompute(): Int = {
      C.this.synchronized(if (C.this.bitmap$0.unary_!)
        {
          C.this.l = C.this.getInt();
          C.this.bitmap$0 = true;
          ()
        });
      C.this.l
    };
    <stable> <accessor> lazy def l(): Int = if (C.this.bitmap$0.unary_!())
      C.this.l$lzycompute()
    else
      {
        C.this.synchronized[Unit](if (C.this.bitmap$0.unary_!())
          {
            C.this.l = C.this.getInt();
            C.this.bitmap$0 = true;
            ()
          });
        C.this.l
      }
  }
}
```

note how the slow path is duplicated.

adriaanm:
Thanks! I pushed a reworked commit that expands on your wip for class members. Will incorporate the local val expansion after lunch

SethTisue:
validation failed because https://github.com/scala/scala-jenkins-infra/issues/182

adriaanm:
1120e23 is mostly green, except for:
- [x] run/t3569.scala
- [x] run/checked.scala
- [x] run/lazy-leaks.scala
- [x] run/nullable-lazyvals.scala
- [x] run/t720.scala

the problems are nulling of single-use lazy fields (I disabled because it nulled too much) and checkinit (haven't integrated that yet)

adriaanm:
@lrytz, I like the simplification of dropping the compute method for local lazy vals, but are we sure that this is ok from a performance point of view? What about the compute methods for member lazy vals?

adriaanm:
AFAICT, one more thing before we can decommission the lazyvals phase:
- [x] add double-checked locking for module instantiation during fields

adriaanm:
- [x] run/t6853.scala  run/t6937.scala  run/t7096.scala  run/t7185.scala  run/t7200.scala  run/t7319.scala 
- [x] run/t3569.scala: error: assignment to non variable
- [x] run/t3670.scala, pos/t2910.scala, pos/t3670.scala:  warning: an unexpected type representation reached the compiler backend: ()Int.
- [x] run/delambdafy_t6028.scala, run/t6028.scala: update check
- [x] run/t6240-universe-code-gen.scala : update

adriaanm:
- [x] specialization 😧 
- [x] while/do-while/tailcalls and local lazy vals: each iteration should have the lazy values uninitialized

src/reflect/scala/reflect/internal/Definitions.scala
```diff
@@ -93,6 +93,9 @@ trait Definitions extends api.StandardDefinitions {
     lazy val boxedClass       = classesMap(x => getClassByName(boxedName(x)))
     lazy val refClass         = classesMap(x => getRequiredClass("scala.runtime." + x + "Ref"))
     lazy val volatileRefClass = classesMap(x => getRequiredClass("scala.runtime.Volatile" + x + "Ref"))
+    lazy val lazyHolders      = symbolsMap(ScalaValueClasses, x => getRequiredClass("scala.runtime.Lazy" + x))
+    lazy val LazyRefClass     = getRequiredClass("scala.runtime.LazyRef")
+    lazy val LazyUnitClass    = getRequiredClass("scala.runtime.LazyUnit")
```
retronym: - [x] `getClassIfDefined` is more tolerant of using old `scala-library.jar`.

adriaanm:
- [x] double-checked locking for local lazies?

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -412,58 +482,127 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
+      *     x$lzy.value
+      *  }
+      * ```
+      */
+    private def mkLazyLocalDef(lazyVal: Symbol, rhs: Tree): Tree = {
+      val lazyValType = lazyVal.tpe.resultType
+      val refClass = lazyHolders.getOrElse(lazyValType.typeSymbol, LazyRefClass)
+      val refTpe = if (refClass != LazyRefClass) refClass.tpe else appliedType(refClass.typeConstructor, List(lazyValType))
+
+      val flags = (lazyVal.flags & FieldFlags | ARTIFACT | MUTABLE) & ~(IMPLICIT | STABLE) // TODO: why include MUTABLE???
```
retronym: Getting rid of `MUTABLE` here avoids an unnecessary `ObjectRef`.

Before:

``` scala
⚡ qscalac -Xprint:mixin $(f "class C { def foo = { lazy val x = 42; x }}")
[[syntax trees at end of                     mixin]] // a.scala
package <empty> {
  class C extends Object {
    def foo(): Int = {
      lazy <artifact> var x$lzy: runtime.ObjectRef = scala.runtime.ObjectRef.create(new scala.runtime.LazyInt());
      C.this.x$1(x$lzy)
    };
    final private[this] def x$1(x$lzy$1: runtime.ObjectRef): Int = scala.Int.unbox(x$lzy$1.elem.$asInstanceOf[scala.runtime.LazyInt]().synchronized({
      if (x$lzy$1.elem.$asInstanceOf[scala.runtime.LazyInt]().initialized().unary_!())
        {
          x$lzy$1.elem.$asInstanceOf[scala.runtime.LazyInt]().initialized_=(true);
          x$lzy$1.elem.$asInstanceOf[scala.runtime.LazyInt]().value_=(42)
        };
      scala.Int.box(x$lzy$1.elem.$asInstanceOf[scala.runtime.LazyInt]().value())
    }));
    def <init>(): C = {
      C.super.<init>();
      ()
    }
  }
}
```

After:

``` scala
[[syntax trees at end of                     mixin]] // a.scala
package <empty> {
  class C extends Object {
    def foo(): Int = {
      lazy <artifact> val x$lzy: scala.runtime.LazyInt = new scala.runtime.LazyInt();
      C.this.x$1(x$lzy)
    };
    final private[this] def x$1(x$lzy$1: scala.runtime.LazyInt): Int = scala.Int.unbox(x$lzy$1.synchronized({
      if (x$lzy$1.initialized().unary_!())
        {
          x$lzy$1.initialized_=(true);
          x$lzy$1.value_=(42)
        };
      scala.Int.box(x$lzy$1.value())
    }));
    def <init>(): C = {
      C.super.<init>();
      ()
    }
  }
}
```

retronym: The redundant box/unbox is an artifact of the way that `synchronized` is modelled as a generic, unspecialized method. It is eliminated by our local optimizations under `-opt`. 

But we can also avoid generating it altogether with a `return`. I've prototyped that in:

https://github.com/retronym/scala/pull/25

adriaanm: I think it may be easier to write to a local variable to store the result if it turns out this is a problem. In the mean time, now that we have an optimizer we can recommend everyone use, I think we can let it do its box/unbox magic disappearing trick.

adriaanm:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/491/consoleFull

# scalaz compile error

```
core/src/main/scala/scalaz/Maybe.scala:230: could not find implicit value for parameter F: scalaz.Monad[scalaz.Maybe]
 implicit def maybeFirstMonad: Monad[FirstMaybe] = Tags.First.subst1[Monad, Maybe](Monad[Maybe])
                                                                                        ^
core/src/main/scala/scalaz/Maybe.scala:242: could not find implicit value for parameter F: scalaz.Monad[scalaz.Maybe]
 implicit def maybeLastMonad: Monad[LastMaybe] = Tags.Last.subst1[Monad, Maybe](Monad[Maybe])
                                                                                     ^
core/src/main/scala/scalaz/Maybe.scala:259: could not find implicit value for parameter F: scalaz.Monad[scalaz.Maybe]
 implicit def maybeMinMonad: Monad[MinMaybe] = Tags.Min.subst1[Monad, Maybe](Monad[Maybe])
                                                                                  ^
core/src/main/scala/scalaz/Maybe.scala:276: could not find implicit value for parameter F: scalaz.Monad[scalaz.Maybe]
 implicit def maybeMaxMonad: Monad[MaxMaybe] = Tags.Max.subst1[Monad, Maybe](Monad[Maybe])
```

# akka compile error

```
akka-actor/src/main/scala/akka/routing/ConsistentHashing.scala:160: 
Cannot find LogSource for Class[_ <: akka.routing.ConsistentHashingRoutingLogic]
please see ScalaDoc for LogSource for how to obtain or construct one.
  private lazy val log = Logging(system, getClass)
                                ^
```

# scala-js test failure

```
Failed: Total 5307, Failed 4, Errors 0, Passed 5153, Skipped 150
Failed tests: org.scalajs.testsuite.jsinterop.ExportsTest
    Test org.scalajs.testsuite.jsinterop.ExportsTest.exporting_all_members_of_a_class failed: java.lang.AssertionError: expected:<3> but was:<undefined>, took 0.006999808 sec
    Test org.scalajs.testsuite.jsinterop.ExportsTest.exporting_lazy_values_issue_977 failed: java.lang.AssertionError: expected:<1> but was:<undefined>, took 0.02 sec
    Test org.scalajs.testsuite.jsinterop.ExportsTest.exports_for_abstract_properties_in_class_issue_2513 failed: java.lang.AssertionError: expected:<5> but was:<undefined>, took 0.010000128 sec
    Test org.scalajs.testsuite.jsinterop.ExportsTest.exports_for_abstract_properties_in_trait_issue_2513 failed: java.lang.AssertionError: expected:<5> but was:<undefined>, took 0.004999936 sec

```

adriaanm:
https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/492/console

adriaanm:
Specialized lazy vals: https://issues.scala-lang.org/browse/SI-5552

SethTisue:
is this the PR that fixes https://issues.scala-lang.org/browse/SI-9814 ?

adriaanm:
Ah, good call -- it's not. I don't think we have a pr for this one. Looks
fixable in 2.12.1 tho

adriaanm:
From akka in the community build:

```
package foo { trait HasLazy { private[foo] lazy val myLazy = "my lady" } }
package bar { class MixInSuperLazy extends foo.HasLazy }
```

adriaanm:
✅ https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/495/console

src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -234,16 +234,15 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
-    *  @see http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
-   */
-  def mkSynchronizedCheck(clazz: Symbol, cond: Tree, syncBody: List[Tree], stats: List[Tree]): Tree =
-    mkSynchronizedCheck(mkAttributedThis(clazz), cond, syncBody, stats)
-
+    * The idiom works only if the condition is using a volatile field.
+    *
+    * @see http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
+    *
+    * TODO: update to state of the art on java 8 (https://shipilev.net/blog/2014/safe-public-construction/)
```
lrytz: it would help if the comment was bit more specific; maybe create a scala-dev issue for it?

adriaanm: https://github.com/scala/scala-dev/issues/204

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1451,9 +1446,9 @@ abstract class RefChecks extends Transform {
             )
       }
 
-      sym.isSourceMethod &&
+      sym.name == nme.apply &&
+        !(sym hasFlag STABLE) && // ???
```
lrytz, line 1450: hmm.. also wondering about this comment `// 🤔🤔🤔`

adriaanm, line 1450: The best I can come up with is that STABLE is assumed to only be added to
compiler-generated methods. An indirect way of detecting those versus ones
written by the user? (Note that lazy accessors have user written code and
the STABLE flag though)

lrytz:
I did a pass over the full diff, a really nice improvement, I especially love the cleanups in mixin. Great to see the tests passing!

adriaanm:
Thanks! Do you have any additional action items for me before this can be merged? I'm investigating the latest failures (EDIT: they are benign check file updates). I was thinking of keeping the commit history mostly as is, to document the refactoring steps.

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -412,58 +495,123 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
+      // lifted into a trait (TODO: not sure about the details here)
+      lazyVal.resetFlag(LAZY | STABLE | ACCESSOR)
 
+      Thicket(mkField(holderSym, New(refTpe)) :: accessor :: Nil)
+    }
 
     // synth trees for accessors/fields and trait setters when they are mixed into a class
     def fieldsAndAccessors(clazz: Symbol): List[ValOrDefDef] = {
-      def fieldAccess(accessor: Symbol): Option[Tree] = {
+      def fieldAccess(accessor: Symbol): List[Tree] = {
```
soc, line 598: Why did the result types of `fieldAccess`, `getterBody` and `setterBody` get changed from `Option` to `List`? It seems only empty and one-element lists are constructed.

lrytz:
Still looking at the discussion in https://github.com/scala/scala-dev/issues/133; so are you saying that dotty's approach (as outlined in https://github.com/scala/scala-dev/issues/133#issuecomment-237384661) is incorrect because reading `true` on the volatile `initialized` field does not necessarily mean that the initialized value will be read from the `value` field?

If so, then how is it different from the current approach for member lazy vals? Here, also the `bitmap` field is volatile, but the value field is read outside any synchronized block; so how does reading `true` from the bit mask ensure that reading the value field will yield the initialized value?

adriaanm:
Ok -- let's chat about the local lazies because I think I may be missing something or not explaining myself well.  (First, I'm pushing --force a slightly squashed version with reworded commit messages. No changes since last HEAD.) EDIT: Lukas resolved my confusion.

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm, line 598: The `superLazy` method produces more than one tree.

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
adriaanm: To clarify, I just inlined `isSourceMethod` here, acting on the TODO in its doc comment.

adriaanm:
I think I'm done here! 🎉  Running one last community build: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/496/console

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -513,51 +513,60 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
-        // (there's no happens-before relation with the read of the volatile initialized field)
-        // TODO: double-checked locking (https://github.come/scala/scala-dev/issues/204)
-        gen.mkSynchronized(Ident(holderSym))(
-          If(Ident(holderSym) DOT initializedGetter, getValue, Block(List(setValue, setInitialized), getValue)))
+      val refClass    = lazyHolders.getOrElse(lazyValType.typeSymbol, LazyRefClass)
+      val refTpe      = if (refClass != LazyRefClass) refClass.tpe else appliedType(refClass.typeConstructor, List(lazyValType))
+
+      val lazyName  = lazyVal.name.toTermName
+      val pos       = lazyVal.pos.focus
+      val flags     = (lazyVal.flags & FieldFlags | ARTIFACT | MUTABLE) & ~(IMPLICIT | STABLE) // TODO: why include MUTABLE???
```
lrytz: the `MUTABLE` flag causes an ObjectRef to be created in lambdalift, and we get

```
    public int t1() {
        ObjectRef v$lzy = ObjectRef.create((Object)new LazyInt());
        return this.v$1(v$lzy);
    }

    private final int v$1(ObjectRef v$lzy$1) {
        return ((LazyInt)v$lzy$1.elem).initialized() ? ((LazyInt)v$lzy$1.elem).value() : this.v$lzycompute$1(v$lzy$1);
    }

    private final /* synthetic */ int v$lzycompute$1(ObjectRef v$lzy$1) {
      ...
    }
```

Removing the flag fixes that issue. Can we just do that, or will it cause other problems?

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -513,51 +513,60 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
+      val holderSym = owner.newValue(lazyName append nme.LAZY_LOCAL_SUFFIX_STRING, pos, flags) setInfo refTpe
+
+      val initializedGetter = refTpe.member(nme.initialized)
+      val isUnit            = refClass == LazyUnitClass
+      val valueGetter       = if (isUnit) NoSymbol else refTpe.member(nme.value)
+      val getValue          = if (isUnit) UNIT     else Apply(Select(Ident(holderSym), valueGetter), Nil)
```
lrytz: this should be a `def` in order not to share trees

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -513,51 +513,60 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
+        owner.newMethod(lazyName append nme.LAZY_SLOW_SUFFIX, pos, ARTIFACT | PRIVATE) setInfo MethodType(Nil, lazyValType)
+
+      val rhsAtComputer = rhs.changeOwner(lazyVal -> computerSym)
+      val computer = mkAccessor(computerSym) {
+        val setInitialized = Apply(Select(Ident(holderSym), initializedGetter.setterIn(refClass)), TRUE :: Nil)
+        val setValue =
+          if (isUnit) rhsAtComputer
+          else Apply(Select(Ident(holderSym), valueGetter.setterIn(refClass)), rhsAtComputer :: Nil)
+
+        gen.mkSynchronized(Ident(holderSym))(mkChecked(Block(List(setValue, setInitialized), getValue)))
```
lrytz: The issue with having `getValue` within the synchronized block is that erausre creates a box/unbox operations (because the synthetic method `def synchronized[T](body: T): T` is not specialized).

Would it be safe to just read the value after the `synchronized` block?

```
    private final /* synthetic */ int v$lzycompute$1(LazyInt v$lzy$1) {
        LazyInt lazyInt = v$lzy$1;
        synchronized (lazyInt) {
            if (!v$lzy$1.initialized()) {
                v$lzy$1.value_$eq(this.getInt());
                v$lzy$1.initialized_$eq(true);
            }
        }
        return v$lzy$1.value();
    }
```

lrytz:
community build went green!

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm: Good point! I've removed it in my WIP.

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm: good catch!

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm: I don't think we can do that, but I have a WIP where we store the result in a local var outside of the synch block. That should take care of the boxing

adriaanm:
Scheduled another community build for the last commit (e61f06e): https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/501/console

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -534,30 +535,48 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
       val rhsAtComputer = rhs.changeOwner(lazyVal -> computerSym)
       val computer = mkAccessor(computerSym) {
         val setInitialized = Apply(Select(Ident(holderSym), initializedGetter.setterIn(refClass)), TRUE :: Nil)
-        val setValue =
-          if (isUnit) rhsAtComputer
-          else Apply(Select(Ident(holderSym), valueGetter.setterIn(refClass)), rhsAtComputer :: Nil)
 
-        gen.mkSynchronized(Ident(holderSym))(mkChecked(Block(List(setValue, setInitialized), getValue)))
+        // we write to a local var outside of the synchronized block to avoid boxing/unboxing (synchronized method is polymorphic)
+        val resVarSym = computerSym.newVariable(localLazyName, pos, ARTIFACT) setInfo lazyValType
```
lrytz: this should be moved into the `else` branch below

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -534,30 +535,48 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
...
+
+        if (isUnit)
+          gen.mkSynchronized(Ident(holderSym))(mkChecked(alreadyComputed = UNIT, compute = Block(List(rhsAtComputer, setInitialized), UNIT)))
+        else {
+          val alreadyComputed = Assign(Ident(resVarSym), getValue)
+          val storeComputed   = Apply(Select(Ident(holderSym), valueGetter.setterIn(refClass)), Ident(resVarSym) :: Nil)
+          val compute = Block(List(Assign(Ident(resVarSym), rhsAtComputer), storeComputed, setInitialized), Ident(resVarSym))
```
lrytz: the `expr` of the block could be simply `UNIT`, no?

lrytz: Even when doing that, there's an unnecessary value stored from the `if`:

```
    private final /* synthetic */ int v$lzycompute$1(LazyInt v$lzy$1) {
        int v$lzy = 0;
        LazyInt lazyInt = v$lzy$1;
        synchronized (lazyInt) {
            BoxedUnit boxedUnit;
            if (v$lzy$1.initialized()) {
                v$lzy = v$lzy$1.value();
                boxedUnit = BoxedUnit.UNIT;
            } else {
                v$lzy = this.i();
                v$lzy$1.value_$eq(v$lzy);
                v$lzy$1.initialized_$eq(true);
                boxedUnit = BoxedUnit.UNIT;
            }
        }
        return v$lzy;
    }
```

Could we get rid of the `boxedUnit` local somehow? I didn't manage to build the right tree shape so far..

adriaanm: Keen eye! I also haven't been able to figure out why erasure emits the box of unit there. Both branches are of type unit already. I'll fiddle a bit more with it.

adriaanm: Looks like it's because the expected type for the `if` inside the `synchronized` is `Object` (the erased type of its type param `T`), and thus the branches are adapted to `Object` instead of `Unit`. Maybe we should give `synchronized` special treatment like `isInstanceOf`

retronym: Without changing erasure, you could make it slightly more palatable with:

```
def compute(x: LazyInt) {
  var result = 0
  x.synchronized[AnyRef] {
    if (x.isInitialized) result = x.value
    else { result = <<rhs>>; x.value = result; x.initialized = result }
    null
  }
  result
}
```

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
retronym: The use of `STABLE` to identify compiler-created methods dates back to https://github.com/scala/scala/commit/9acfa7693d146fc2ea6fdad61c74907354910ea8#diff-d5789e5ae5061197d782d08324b260dbR97.

Through refactorings and comment ping pong, Martin commented that it was to "exclude all accessors" from the designation of `isSourceMethod`. 

In this instance, I think the subsequent check of `isCase` makes the check for `!STABLE` redundant.

```
scala> case class C(apply: Int) { }; object C { private val apply = 42 }
defined class C
defined object C

scala> :power

scala> typeOf[C].member("apply").isSourceMethod
<console>:39: error: type mismatch;
 found   : String("apply")
 required: $r.intp.global.Name
Note that implicit conversions are not applicable because they are ambiguous:
 both method stringToTermName in trait Names of type (s: String)$r.intp.global.TermName
 and method stringToTypeName in trait Names of type (s: String)$r.intp.global.TypeName
 are possible conversion functions from String("apply") to $r.intp.global.Name
       typeOf[C].member("apply").isSourceMethod
                        ^

scala> typeOf[C].member(TermName("apply")).isSourceMethod
res1: Boolean = false

scala> typeOf[C].member(TermName("apply")).isCase
res2: Boolean = false

scala> typeOf[C.type].member(TermName("apply")).alternatives
res6: List[$r.intp.global.Symbol] = List(method apply, value apply)

scala> typeOf[C.type].member(TermName("apply")).alternatives.map(_.isCase)
res7: List[Boolean] = List(true, false)
```

retronym: The obvious question is: why not check `ACCESSOR` rather than `STABLE` if we want to "exclude all accessors"? 

I think the answer to this is that there are accessor-like methods that don't have the `ACCESSOR` flag set. One example I found is case field accessors:

```

scala> case class C(private val privParam: Int)
defined class C

scala> val caseFieldAccessor = typeOf[C].decl(TermName("privParam$1"))
caseFieldAccessor: $r.intp.global.Symbol = value privParam$1

scala> caseFieldAccessor.isStable
res22: Boolean = true

scala> caseFieldAccessor.isAccessor
res23: Boolean = false
```

src/compiler/scala/tools/nsc/typechecker/Unapplies.scala
```diff
@@ -128,7 +128,11 @@ trait Unapplies extends ast.TreeDSL {
    */
   def factoryMeth(mods: Modifiers, name: TermName, cdef: ClassDef): DefDef = {
     val tparams   = constrTparamsInvariant(cdef)
-    val cparamss  = constrParamss(cdef)
+    // don't propagate e.g. @volatile annot to apply's argument
+    def valDefWithoutAnnots(vd: ValDef): ValDef =
+      treeCopy.ValDef(vd, vd.mods.mapAnnotations(_ => Nil), vd.name, vd.tpt, vd.rhs)
```
retronym: One example of an annotation that we might want to copy to the parameter is `@deprecatedName`. Failure to do so would regress the following:

```
scala> case class A(@deprecatedName('oldName) newName: Any)
defined class A

scala> new A(oldName = "methuselah")
res24: A = A(methuselah)

scala> A(oldName = "methuselah")
res25: A = A(methuselah)
```

adriaanm: good catch! I have a more robust approach that drops non-param-targeting annots in the info completer

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm: I quickly hacked together a version of erasure that essentially sees through `synchronized` (it retains its polymorphic info), which avoids the boxed unit. I'll push the commit for discussion.

adriaanm:
Pushed some commits for discussion, addressing review feedback.

sjrd:
db1b671 will most probably require adaptations in Scala.js, if it goes in.

retronym:
@adriaanm Needs a (trivial) rebase a la https://github.com/scala/scala/compare/2.12.x...retronym:rebase/5294?expand=1

adriaanm:
@retronym -- rebased! How do you feel about db1b671?

retronym:
That approach in erasure looks right me.

sjrd:
I'm not a big fan of specializing `synchronized` in erasure, TBH. So far, there are exactly 2 methods that are specialized so: `isInstanceOf` and `asInstanceOf`. That's fine because they _need_ specialization for _correctness_.

If we open the door of special-case-in-erasure to methods that want specialization "only" for performance (but not for correctness), we create a precedent to justify many more such specializations. For example, what about `Predef.locally` (which looks like a free syntactic construct)?

Removing the overhead of boxing in those cases should be left to an optimizer, IMO. In this case, even intra-method optimizations (which are always safe wrt separate compilation) are enough.

adriaanm:
I guess we'll live with the extra boxes for now then, just hope they will be unpacked before our next move. Tracking follow-up here: https://github.com/scala/scala-dev/issues/209

retronym:
I think that `synchronized` ought to emit the same bytecode as the Java version, with the default compiler settings. There are a few ways to get there, and we can discuss them further for 2.12.1.

DarkDimius:
The easiest way to implement it for me in linker was to add several 
specialized "synchronized" methods before erasure and make genbcode 
collapse them back together.

adriaanm:
I would argue that [my solution](https://github.com/adriaanm/scala/commit/328b50cee58698d99c8841ddd67752d0a93c19de) is a simpler way to get that specialization (its latest incarnation had no impact on scala-js, btw). It makes sense to me to special-case `synchronized`'s erasure. It doesn't result in an actual method call, so it doesn't set a precedent for other actual methods, eve if predefined. In any case, we should have the invariant that `synchronized(x)` and `x` erase to the same tree, modulo wrapping it in a marker for the backend to emit some extra monitor-related bytecode.

adriaanm:
/rebuild

failure was due to overload again...

adriaanm:
I diffed the bytecode as follows, will include some representative diffs in comments below:

``` bash
starr=a360470
locker=328b50c
g co $starr
sbt publishLocal
g co $locker
sbt -Dstarr.version=2.12.0-local-$starr clean publishLocal
cp -a build/quick/classes .
sbt -Dstarr.version=2.12.0-local-$locker compile
cd classes
for i in library reflect compiler; do ../build/quick/bin/scala scala.tools.nsc.backend.jvm.AsmUtils $(find $i -name "*.class" ); done
git init
for i in library reflect compiler; do git add  $(find $i -name "*.class.asm" ); done
g commit -m"quick"
cp -a ../build/quick/classes/* .
for i in library reflect compiler; do ../build/quick/bin/scala scala.tools.nsc.backend.jvm.AsmUtils $(find $i -name "*.class" ); done
git --no-pager diff > strap-$starr-to-$locker.diff

```

adriaanm:
# local lazy

``` diff
-  // access flags 0x2
-  private default in$lzycompute$1(Lscala/runtime/ObjectRef;Lscala/runtime/VolatileByteRef;)Ljava/io/BufferedInputStream;
+  // access flags 0x1002
+  private synthetic default in$lzycompute$1(Lscala/runtime/LazyRef;)Ljava/io/BufferedInputStream;
     // parameter final  in$lzy$1
-    // parameter final  bitmap$0$1
     TRYCATCHBLOCK L0 L1 L2 java/lang/Throwable
-    ALOAD 0
+    ACONST_NULL
+    ASTORE 2
+    ALOAD 1
     DUP
     ASTORE 3
     MONITORENTER
    L0
-    ALOAD 2
-    GETFIELD scala/runtime/VolatileByteRef.elem : B
-    ICONST_1
-    IAND
-    I2B
-    ICONST_0
-    IF_ICMPNE L3
     ALOAD 1
+    INVOKEVIRTUAL scala/runtime/LazyRef.initialized ()Z
+    IFEQ L3
+    ALOAD 1
+    INVOKEVIRTUAL scala/runtime/LazyRef.value ()Ljava/lang/Object;
+    CHECKCAST java/io/BufferedInputStream
+    ASTORE 2
+    GOTO L1
+   L3
     ALOAD 0
     INVOKEINTERFACE scala/reflect/io/Streamable$Bytes.bufferedInput ()Ljava/io/BufferedInputStream;
-    PUTFIELD scala/runtime/ObjectRef.elem : Ljava/lang/Object;
-    ALOAD 2
+    ASTORE 2
+    ALOAD 1
     ALOAD 2
-    GETFIELD scala/runtime/VolatileByteRef.elem : B
+    INVOKEVIRTUAL scala/runtime/LazyRef.value_$eq (Ljava/lang/Object;)V
+    ALOAD 1
     ICONST_1
-    IOR
-    I2B
-    PUTFIELD scala/runtime/VolatileByteRef.elem : B
-   L3
-    GETSTATIC scala/runtime/BoxedUnit.UNIT : Lscala/runtime/BoxedUnit;
-    POP
+    INVOKEVIRTUAL scala/runtime/LazyRef.initialized_$eq (Z)V
    L1
     ALOAD 3
     MONITOREXIT
```

adriaanm:
I stubbornly included my specialized erasure for `synchronized`, which resulted in the following over 2000 times:

``` diff
-    GETSTATIC scala/runtime/BoxedUnit.UNIT : Lscala/runtime/BoxedUnit;
-    POP
```

It also dropped the following unneeded casts:

``` diff
- CHECKCAST scala/tools/nsc/CompilationUnits$CompilationUnit
- CHECKCAST scala/tools/nsc/util/ClassPath
- CHECKCAST scala/collection/mutable/SynchronizedBuffer
- CHECKCAST scala/collection/mutable/SynchronizedBuffer
- CHECKCAST scala/collection/mutable/Buffer
- CHECKCAST scala/collection/mutable/SynchronizedBuffer
- CHECKCAST scala/collection/mutable/SynchronizedBuffer
- CHECKCAST scala/collection/mutable/Buffer
- CHECKCAST scala/collection/Iterator
- CHECKCAST scala/collection/mutable/SynchronizedMap
- CHECKCAST scala/collection/mutable/SynchronizedMap
- CHECKCAST scala/collection/mutable/Map
- CHECKCAST scala/Option
- CHECKCAST scala/collection/Iterator
- CHECKCAST scala/collection/Set
- CHECKCAST scala/collection/Iterable
- CHECKCAST scala/collection/Iterator
- CHECKCAST scala/Option
- CHECKCAST scala/Option
- CHECKCAST scala/collection/mutable/SynchronizedMap
- CHECKCAST scala/collection/mutable/SynchronizedMap
- CHECKCAST scala/collection/Iterable
- CHECKCAST scala/collection/Iterator
- CHECKCAST scala/collection/mutable/SynchronizedSet
- CHECKCAST scala/collection/mutable/SynchronizedSet
- CHECKCAST scala/collection/mutable/SynchronizedSet
- CHECKCAST scala/collection/mutable/SynchronizedSet
- CHECKCAST scala/collection/mutable/Set
- CHECKCAST scala/collection/mutable/Set
- CHECKCAST scala/collection/immutable/List
- CHECKCAST java/lang/String
- CHECKCAST scala/collection/mutable/ArrayBuffer
- CHECKCAST scala/collection/mutable/ArrayBuffer
- CHECKCAST scala/reflect/internal/Names$TypeName
- CHECKCAST scala/reflect/internal/Names$TermName
- CHECKCAST scala/reflect/internal/Names$TermName
- CHECKCAST scala/reflect/internal/Types$Type
- CHECKCAST scala/reflect/internal/Types$Type
```

adriaanm:
# signature for lazy compute methods

``` diff
+  // signature ()Lscala/collection/mutable/HashMap<Ljava/lang/String;Lscala/reflect/io/ZipArchive$DirEntry;>;
+  // declaration: scala.collection.mutable.HashMap<java.lang.String, scala.reflect.io.ZipArchive$DirEntry> allDirs$lzycompute()
   private allDirs$lzycompute()Lscala/collection/mutable/HashMap;
```

adriaanm:
# no more compute method for module

``` diff

-  // access flags 0x2
-  private MacroImplSig$lzycompute()Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
-    TRYCATCHBLOCK L0 L1 L2 java/lang/Throwable
-    ALOAD 0
-    DUP
-    ASTORE 1
-    MONITORENTER
-   L0
-    ALOAD 0
-    GETFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
-    IFNONNULL L3
-    ALOAD 0
-    NEW scala/reflect/macros/compiler/Validators$Validator$MacroImplSig$
-    DUP
-    ALOAD 0
-    INVOKESPECIAL scala/reflect/macros/compiler/Validators$Validator$MacroImplSig$.<init> (Lscala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler;)V
-    PUTFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
-   L3
-    GETSTATIC scala/runtime/BoxedUnit.UNIT : Lscala/runtime/BoxedUnit;
-    POP
-   L1
-    ALOAD 1
-    MONITOREXIT
-    GOTO L4
-   L2
-    ALOAD 1
-    MONITOREXIT
-    ATHROW
-   L4
-    ALOAD 0
-    GETFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
-    ARETURN
-    MAXSTACK = 4
-    MAXLOCALS = 2
-

   // access flags 0x1
   public scala$reflect$macros$compiler$Validators$Validator$$MacroImplSig()Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
+    TRYCATCHBLOCK L0 L1 L2 java/lang/Throwable
     ALOAD 0
     GETFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
-    IFNONNULL L0
+    IFNONNULL L3
     ALOAD 0
-    INVOKESPECIAL scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$lzycompute ()Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
-    GOTO L1
+    DUP
+    ASTORE 1
+    MONITORENTER
    L0
     ALOAD 0
     GETFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
+    IFNONNULL L1
+    ALOAD 0
+    NEW scala/reflect/macros/compiler/Validators$Validator$MacroImplSig$
+    DUP
+    ALOAD 0
+    INVOKESPECIAL scala/reflect/macros/compiler/Validators$Validator$MacroImplSig$.<init> (Lscala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler;)V
+    PUTFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
    L1
+    ALOAD 1
+    MONITOREXIT
+    GOTO L3
+   L2
+    ALOAD 1
+    MONITOREXIT
+    ATHROW
+   L3
+    ALOAD 0
+    GETFIELD scala/reflect/macros/compiler/DefaultMacroCompiler$MacroImplRefCompiler.MacroImplSig$module : Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig$;
     ARETURN
-    MAXSTACK = 1
-    MAXLOCALS = 1
+    MAXSTACK = 4
+    MAXLOCALS = 2
```

adriaanm:
# `lazy val`'s underlying field no longer `final`

``` diff
-  // access flags 0x12
-  private final Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig; scala$reflect$macros$compiler$Validators$Validator$$referenceMacroImplSig
+  // access flags 0x2
+  private Lscala/reflect/macros/compiler/Validators$Validator$MacroImplSig; scala$reflect$macros$compiler$Validators$Validator$$referenceMacroImplSig

```

adriaanm:
# special erasure for synchronized

``` scala
class VC(val x: String) extends AnyVal

class C {
  def synchVc = {println("vc"); synchronized(new VC("a"))}
  def synch_??? = {println("???"); synchronized(???)}
  def synch_() = {println("println"); synchronized(println("()"))}
}
```

after posterasure:

``` diff
     def synchVc(): String = {
       scala.Predef.println("vc");
-      C.this.synchronized(new VC("a")).$asInstanceOf[VC]().x()
+      C.this.synchronized[String]("a")
     };
     def synch_???(): Nothing = {
       scala.Predef.println("???");
-      C.this.synchronized(scala.Predef.???()).$asInstanceOf[Nothing]()
+      C.this.synchronized[Nothing](scala.Predef.???())
     };
     def synch_(): Unit = {
       scala.Predef.println("println");
-      {
-        C.this.synchronized({
-          scala.Predef.println("()");
-          scala.runtime.BoxedUnit.UNIT
-        });
-        ()
-      }
+      C.this.synchronized[Unit](scala.Predef.println("()"))
     }
   }
 }
```

I find this compelling enough to push my commit once more for discussion.

sjrd:
> I find this compelling enough to push my commit once more for discussion.

I don't have anything to add to my previous arguments. So if you think it's worth it, I yield ;)

We'll wait until a firm decision about this is reached before releasing Scala.js 0.6.12, as we'll need an adaptation in our back-end, as I mentioned earlier. (The whole point of releasing 0.6.12 one of these days is to support RC1, so it doesn't make sense for us _not_ to wait until we know what to do to ensure that support.)

adriaanm:
@sjrd, the version I pushed shouldn't need support in scala-js, assuming you run after cleanup? My argument against this opening the gates is that synchronized is not a method, while locally is. We only do this special casing for these magic methods that are truly primitive (and thus not definable in the library).

sjrd:
> the version I pushed shouldn't need support in scala-js, assuming you run after cleanup?

Ah but we don't ;-) We run right after `mixin` (before `delambdafy` and `cleanup`).

> My argument against this opening the gates is that synchronized is not a method, while locally is. We only do this special casing for these magic methods that are truly primitive (and thus not definable in the library).

Yes, being a primitive is a reasonable stance on what qualifies for special treatment.

adriaanm:
Ah I checked. Since you run after Mixins My assumption doesn't hold.

adriaanm:
Ok, great! Could you port the change from cleanups in my commit here to the suitable spot in scala-js, please?

sjrd:
Yes, OK, I'll do that right now.

sjrd:
https://github.com/scala-js/scala-js/pull/2575

src/reflect/scala/reflect/internal/transform/Erasure.scala
```diff
@@ -345,14 +344,14 @@ trait Erasure {
 
   /**  The symbol's erased info. This is the type's erasure, except for the following symbols:
    *
-   *   - For $asInstanceOf      : [T]T
+   *   - For $asInstanceOf and synchronized      : [T]T
```
sjrd: `synchronized` has in fact type `[T](T)T`, doesn't it?

adriaanm: - [x] Ah yes indeed! I'll  fix that

adriaanm:
Regarding the byte code diff and `lazy val` fields no longer being `ACC_FINAL`: that's progress according to the note at `BTypesFromSymbols::javaFlags`

> // For fields, only eager val fields can receive ACC_FINAL. vars or lazy vals can't

From [JLS 17.5.3](http://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5.3):

> Another problem is that the specification allows aggressive optimization of final fields. Within a thread, it is permissible to reorder reads of a final field with those modifications of a final field that do not take place in the constructor.

Moreover, [jvms-6.5.putfield](http://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.putfield) says:

> Otherwise, if the field is `final`, it must be declared in the current class, and the instruction must occur in an instance initialization method (`<init>`) of the current class. Otherwise, an `IllegalAccessError` is thrown.

... and Java 9 may actually close this loophole (on older VMs, the `IllegalAccessError` isn't actually thrown)

adriaanm:
From JLS 17.5.3:
> Another problem is that the specification allows aggressive optimization of final fields. Within a thread, it is permissible to reorder reads of a final field with those modifications of a final field that do not take place in the constructor.

src/library/scala/runtime/LazyRef.scala
```diff
@@ -0,0 +1,52 @@
...
+}
+
+class LazyDouble {
+  var value: Double = _
+  @volatile var initialized: Boolean = false
+}
+
+class LazyUnit {
+  @volatile var initialized: Boolean = false
+}
```
retronym, line 52: Compare these with `VolatileXxxRef`, e.g.

``` java
public class VolatileLongRef implements java.io.Serializable {
    private static final long serialVersionUID = -3567869820105829499L;

    volatile public long elem;
    public VolatileLongRef(long elem) { this.elem = elem; }
    public String toString() { return java.lang.Long.toString(elem); }

    public static VolatileLongRef create(long e) { return new VolatileLongRef(e); }
    public static VolatileLongRef zero() { return new VolatileLongRef(0); }
}
```

Differences to discuss:
- implementation in Java vs Scala
- Serializability
- toString

retronym:
Bytecode diffs all LGTM

retronym:
Could the local lazy `lzycompute` method be simplified to remove local variables after the erasure change?

``` java
    private static final /* synthetic */ int b2$lzycompute$1(LazyInt b2$lzy$1) {
        int b2$lzy = 0;
        LazyInt lazyInt = b2$lzy$1;
        synchronized (lazyInt) {
            if (b2$lzy$1.initialized()) {
                b2$lzy = b2$lzy$1.value();
            } else {
                b2$lzy = 42;
                b2$lzy$1.value_$eq(b2$lzy);
                b2$lzy$1.initialized_$eq(true);
            }
        }
        return b2$lzy;
    }
```

``` diff
diff --git a/src/compiler/scala/tools/nsc/transform/Fields.scala b/src/compiler/scala/tools/nsc/transform/Fields.scala
index 6753016..7dafe51 100644
--- a/src/compiler/scala/tools/nsc/transform/Fields.scala
+++ b/src/compiler/scala/tools/nsc/transform/Fields.scala
@@ -564,15 +564,11 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
           // we write to a local var outside of the synchronized block to avoid boxing/unboxing (synchronized method is polymorphic)
           val resVarSym = computerSym.newVariable(localLazyName, pos, ARTIFACT) setInfo lazyValType

-          val alreadyComputed = Assign(Ident(resVarSym), getValue)
+          val alreadyComputed = getValue
           val storeComputed   = Apply(Select(Ident(holderSym), valueGetter.setterIn(refClass)), Ident(resVarSym) :: Nil)
-          val compute = Block(List(Assign(Ident(resVarSym), rhsAtComputer), storeComputed), setInitialized)
+          val compute = Block(rhsAtComputer, setInitialized, getValue)

-          Block(
-            mkTypedValDef(resVarSym) ::
-            gen.mkSynchronized(Ident(holderSym))(mkChecked(alreadyComputed, compute)) :: Nil,
-            Ident(resVarSym)
-          )
+          gen.mkSynchronized(Ident(holderSym))(mkChecked(alreadyComputed, compute))
         }
       }
```

I'm happy for this to come in a followup PR.

src/library/scala/runtime/LazyRef.scala
adriaanm: /cc @lrytz, @DarkDimius

sjrd: The implementation in Scala is good for Scala.js. If it's in Java, we have to reimplement it in Scala on our end. We had to do this for the `XyzRef` classes. (the public field is particularly annoying, and requires a hack in the compiler back-end)

adriaanm:
The earlier test suite failures are due to jenkins load.

retronym:
LGTM

adriaanm:
🎉

PR #5296, opened 2016-07-21
SD-186 Fix positions in trait method bytecode

retronym:
Concrete, non private methods in traits are translated into a static
method with an explicit `$this` parameter. After this translation,
the references to `$this` (subistuted for `this` in user written code)
where being positioned at the position of the method, which makes
debugging unpleasant.

This commit leaves the `Ident($this)` trees unpositioned. This is
analagous to what we do in the body of extension methods, which
is the other user of `ThisSubstitutor`.

It would be more correct to copy the position of each `This`
tree over to the substituted tree. That would let us set a breakpoint
on a line that _only_ contained `this`. But in 99% of cases users
won't be able to spot the difference, so I've opted for the tried
and tested approach here.

Review by @lrytz


retronym:
https://github.com/scala/scala/pull/5296

lrytz:
LGTM

PR #5297, opened 2016-07-22
Disable stub warning by default.

retronym:
When we create a class symbols from a classpath elements, references
to other classes that are absent from the classpath are represented
as references to "stub symbols". This is not a fatal error; for instance
if these references are from the signature of a method that isn't called
from the program being compiled, we don't need to know anything about them.
A subsequent attempt to look at the type of a stub symbols will trigger a
compile error.

Currently, the creation of a stub symbol incurs a warning. This commit
removes that warning on the basis that it isn't something users need
to worry about. javac doesn't emit a comparable warning.

The warning is still issued under any of `-verbose` / `-Xdev` / `-Ydebug`.

Rebase  of #5268, both to squash and expand the commit message, and to step away from the transient build failures that seemed to plague that PR 🙏

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
@@ -366,7 +366,7 @@ abstract class ClassfileParser {
       //   - better owner than `NoSymbol`
       //   - remove eager warning
       val msg = s"Class $name not found - continuing with a stub."
-      if (!settings.isScaladoc) warning(msg)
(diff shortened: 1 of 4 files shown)


retronym:
LGTM (echoing my review from the #5268)

soc:
Would it make sense to have a test with these options enabled to prevent the stub warnings from regressing?

retronym:
I see your point, but in general we don't bother with test for diagnostic output like that, and I don't think the code in question is likely to be accidentally removed.

soc:
Alright!

PR #5298, opened 2016-07-22
SD-121 Remove now-unneeded J{Function,Proc}N functional interfaces

retronym:
Non specialized functions can directly use `scala.FunctionN` as the
functional interface, now that mixin generates default methods in
the new trait encoding.

Unfortunately we can't do this for specialized functions as things
stand: specialization leaves the wrong method abstract. In principle,
we could/should amend the specialization transform to fix this. But
my earlier attempts at this weren't sucessful, so for now we have
to stick with the fallback plan of keeping some hand rolled functional
interfaces around.

This commit reduces the surface area of `scala.runtime.java8` to
the minimal requiremnt: one functional interface for each specialized
variant of `Function{0,1,2}`.

Fixes https://github.com/scala/scala-dev/issues/121


retronym:
Review by @lrytz | @adriaanm

lrytz:
LGTM, cool!

szeiger:
Unfortunately this removes `JFunction.proc`, so now you have to manually `return scala.runtime.BoxedUnit.UNIT` from Java code to create a Scala function that returns Unit.

szeiger:
Another issue: We lose serializability of lambdas once again. JFunction\* classes were Serializable, scala.Function\* classes are still not.

retronym:
Maybe we should leave those concerns to Scala-java8-compat and leave things here minimal

retronym:
Scalac adds serializability to anon classes without the need for the functional interface itself to extend serializable.

PR #5299, opened 2016-07-22
Improve sbt-based IntelliJ integration in case of Windows

mpociecha:
Use File.pathSeparator when processing classpath instead of just : to don't end up with classpath like "C:\sth\a.jar:C:\sth\b.jar", what was causing problems after split(":").

Display questions first and then wait for user input. I tested on 3 computes and without flush the printed  questions were always displayed just after user replied - never before.


lrytz:
LGTM

PR #5301, opened 2016-07-25
SD-167 Fine tuning constructor pattern translation

retronym:
- Avoid calling NoSymbol.owner when checking whether we're
  dealing with a case class constructor pattern or a general
  extractor. Tested manually with the test case in the ticket,
  no more output is produced under `-Xdev`.
- Be more conservative about the conversion to a case class
  pattern: rather than looking just at the type of the pattern
  tree, also look at the tree itself to ensure its safe to
  elide. This change is analagous to SI-4859, which restricted
  rewrites of case apply calls to case constructors.

I've manually tested that case class patterns are still efficiently
translated:

```
object Test {
    Some(1) match { case Some(x) => }
  }
}

```

```
% qscalac -Xprint:patmat sandbox/test.scala
[[syntax trees at end of                    patmat]] // test.scala
package <empty> {
    def <init>(): Test.type = {
      Test.super.<init>();
      ()
    };
      case <synthetic> val x1: Some[Int] = scala.Some.apply[Int](1);
      case4(){
        if (x1.ne(null))
          matchEnd3(())
        else
          case5()
      };
      case5(){
        matchEnd3(throw new MatchError(x1))
      };
      matchEnd3(x: Unit){
        x
      }
    }
  }
}
```


retronym:
Review by @adriaanm 

Spun out of a fix in #5263.

adriaanm:
LGTM

PR #5302, opened 2016-07-26
Reduce deprecations and warnings

soc:


soc:
@retronym Can we spin up a new partest build?

There are cleanups that need to be done before we can release 2.12: https://github.com/scala/scala/blob/2.12.x/src/library/scala/Predef.scala#L423

soc:
```
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00007f9074a80000, 123207680, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 123207680 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/jenkins/workspace/scala-2.12.x-validate-test/hs_err_pid24410.log
```

Is there any week in which CI isn't broken one way or another? :-)

soc:
I don't get it. This compiled on my machine with exactly these errors fixed. It's hard to tell what the errors actually are as the diff isn't shown anymore. Maybe we should give up on CI if it's broken every week anyway?

test/files/run/bitsets.check
```diff
@@ -82,4 +82,3 @@ i2_r0 = true
 i2_r1 = true
 i2_r2 = true
 i2_r3 = true
-
```
retronym, line 85: Did you mean to remove this line? That's what is showing up in the diff.

image

retronym:
BTW, we're looking into the memory exhaustion problem in Jenkins as https://github.com/scala/scala-jenkins-infra/issues/181

lrytz:
LGTM

SethTisue:
thanks, Simon, we really appreciate your ongoing attention to this kind of thing.

PR #5303, opened 2016-07-26
avoid boxing

xuwei-k:
scala.runtime.Rich{Double, Float} has `isNaN` and these are value class.
Also java.lang.{Double, Float} has `isNaN`.
- https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#isNaN--
- https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#isNaN--

We can't call `RichDouble#isNaN` because
`implicit def double2Double(x: Double): java.lang.Double`
is higher priority than
`implicit def doubleWrapper(x: Double): RichDouble`

```
$ scala -version
Scala code runner version 2.11.8 -- Copyright 2002-2016, LAMP/EPFL

$ scala -Xprint:jvm -e "1.0.isNaN"
[[syntax trees at end of                       jvm]] // scalacmd616162202928036892.scala
package <empty> {
  object Main extends Object {
      new <$anon: Object>();
      ()
    };
    def <init>(): Main.type = {
      Main.super.<init>();
      ()
    }
  };
  final class anon$1 extends Object {
    def <init>(): <$anon: Object> = {
      anon$1.super.<init>();
      scala.this.Predef.double2Double(1.0).isNaN();
      ()
    }
  }
}
```

### DIFF
--- a/src/library/scala/concurrent/duration/Duration.scala
+++ b/src/library/scala/concurrent/duration/Duration.scala
@@ -120,7 +120,7 @@ object Duration {
   def fromNanos(nanos: Double): Duration = {
     if (nanos.isInfinite)
       if (nanos > 0) Inf else MinusInf
-    else if (nanos.isNaN)
+    else if (JDouble.isNaN(nanos))
       Undefined
     else if (nanos > Long.MaxValue || nanos < Long.MinValue)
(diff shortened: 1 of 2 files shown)


retronym:
This implicit situation here is pretty unfortunate. Would be nice to train `-opt` to deal with this, but for now your patch looks like a pragmatic solution.

LGTM

PR #5304, opened 2016-07-26
SI-4826 Fixes to javadoc [ci: last-only]

jodersky:
Some issues were discovered in my last contribution #5246 through the community build. This PR aims to solve them.

Posting this as a WIP so I can get a discussion going with the compiler maintainers, specifically @adriaanm and @VladUreche 

Issues to be fixed:
- [x] A doc comment in head position (such as a copyright header) will crash the compiler with an NPE, due to the way the javadoc analyzer is initialised. Fixed (thanks @adriaanm)
- [x] Documented java fields fail. Fixed in latest commit
- [x] Interfaces cause problems due to methods being placed into companion objects
- [x] Comments on constructors lead to double definition of constructor symbol
- [x] Ambiguous reference error to same-named objects defined in outer and inner classes (thanks @adriaanm)


jodersky:
An update on the issue: it seems that nesting is irrelevant, all interfaces fail if they have abstract methods (so in their essence). According to their trees, interfaces are represented as a trait with companion object.
As the following example shows, it looks like methods are placed in the wrong object.

The following java code (taken from Akka):

``` java
public interface Apply<I, R> {
    public R apply(I i) throws Exception;
}
```

gets parsed to this:

``` scala
object Apply extends  {
  def <init>();
  def apply(i: I): R
};
abstract trait Apply[I, R] extends _root_.java.lang.Object {
  import Apply._;
  def <init>()
}
```

resulting in a type error (R and I not found) as well as an abstract method error (only classes can have declared but undefined members).

Putting the interface's method in the trait rather than its companion object would solve the issues, and also make more sense in my opinion.

@adriaanm do you have any pointers where about in the compile I could look into this further?
Also, I am purely speculating here, but could this be related to the way default methods in interfaces are treated by the java parser of scalac? I imagine they could be implemented as members of a companion object; maybe something is messed up there.

adriaanm:
Aha! Looks like the mystery is slowly unraveling. Are they somehow being seen as static? I'm not familiar with the java parsing side, so I'll have to investigate a bit as well tomorrow. No immediate suggestions off the top of my head.

jodersky:
I just stumbled upon something even more interesting: if declarations in interfaces don't have a doc comment they are defined in the trait and therefore compile fine! As soon as you add a doc comment, they appear to be treated as static and get put into the companion object.
Looks like an issue with the way the parser is treating `DocDef`s

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -726,8 +726,15 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
-          (if (mods.hasStaticFlag || inInterface && !(decls exists (_.isInstanceOf[DefDef])))
+          val decls = joinComment(memberDecl(mods, parentToken))
+
+          def isDefDef(tree: Tree): Boolean = tree match {
+            case _: DefDef => true
+            case DocDef(_, defn) => isDefDef(defn)
+            case _ => false
+          }
+
+          (if (mods.hasStaticFlag || inInterface && !(decls exists isDefDef))
```
jodersky, line 716: FYI, this used to check whether a tree was an instance of a `DefDef`, therefore causing methods that had a doc comment (thus being nested under a `DocDef`) to end up as statics

jodersky:
Regarding the last item, secondary constructors aren't the issue. The combination of class declaration documentation and constructor documentation lead to double definition of the constructor symbol.

jodersky:
All issues should be fixed now, removing WIP status. I was able to compile and "doc" akka-actor without any errors.

@adriaanm and/or @VladUreche could you re-run the community build and give me your opinion on the changes. Thanks!

adriaanm:
Great! Running build here: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/462/console

Some failures are expected (twitter-util/shapeless)

adriaanm:
That took care of those failures, but there's one last issue:

```
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:8883: reference to Builder is ambiguous;
 it is both defined in object VectorClock and imported subsequently by 
 import Version._
       public Builder newBuilderForType() { return newBuilder(); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:8887: reference to Builder is ambiguous;
 it is both defined in object VectorClock and imported subsequently by 
 import Version._
       public Builder toBuilder() { return newBuilder(this); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:8890: reference to Builder is ambiguous;
 it is both defined in object VectorClock and imported subsequently by 
 import Version._
       protected Builder newBuilderForType(
                 ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:12469: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import Number._
       public Builder newBuilderForType() { return newBuilder(); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:12473: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import Number._
       public Builder toBuilder() { return newBuilder(this); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:12476: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import Number._
       protected Builder newBuilderForType(
                 ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:13048: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import EWMA._
       public Builder newBuilderForType() { return newBuilder(); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:13052: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import EWMA._
       public Builder toBuilder() { return newBuilder(this); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:13055: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import EWMA._
       protected Builder newBuilderForType(
                 ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:13626: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import Metric._
       public Builder newBuilderForType() { return newBuilder(); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:13630: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import Metric._
       public Builder toBuilder() { return newBuilder(this); }
              ^
 akka-cluster/src/main/java/akka/cluster/protobuf/msg/ClusterMessages.java:13633: reference to Builder is ambiguous;
 it is both defined in object NodeMetrics and imported subsequently by 
 import Metric._
       protected Builder newBuilderForType(
                 ^
 No documentation generated with unsuccessful compiler run
 12 errors found

```

jodersky:
Sorry for the delay, I was having some difficulty in condensing the problem (the file in Akka from which the error originated is almost 17kloc!)

So here is a minimal snippet to reproduce the ambiguity error:

``` java
// this compiles in javac, but gives an error when running scaladoc on it
public class Buildable {

    public static class Builder {}

    public static class InnerBuildable {
        public static class Builder {}
        public Builder foo() { return new Builder(); } // this line gives an error, that Builder is ambiguous
    }

}
```

scalac's java parser creates this tree:

``` scala
// Buildable.java
package test.scaldoc {
  object Buildable extends  {
    def <init>();
    object Builder extends  {
      def <init>()
    };
    <static> class Builder extends _root_.java.lang.Object {
      def <init>()
    };
    object InnerBuildable extends  {
      def <init>();
      object Builder extends  {
        def <init>()
      };
      <static> class Builder extends _root_.java.lang.Object {
        def <init>()
      }
    };
    <static> class InnerBuildable extends _root_.java.lang.Object {
      import InnerBuildable._; // imports InnerBuildable.Builder
      def <init>();
      def foo(): Builder // error manifests itself here
    }
  };
  class Buildable extends _root_.java.lang.Object {
    import Buildable._;
    def <init>()
  }
}
```

it looks like importing `InnerBuildable`'s companion object makes the reference to `Builder` ambiguous, since `Builder` can now be either the companion object's one or the outer class' one.

jodersky:
The most straight-forward fix I can think of is to make the type's path explicit, i.e. change

``` scala
def foo(): Builder
```

to 

``` scala
def foo(): InnerBuildable.Builder
```

This however can't be done in the JavaParser phase, so would probably require changes in the Namers phase.
@adriaanm , do you have any other ideas on how to tackle this problem?

adriaanm:
No, your conclusion is correct. this import business technical debt was going to bite us at some point. At a conf but I'll see if I can dig up the history around this tomorrow.

adriaanm:
I didn't get to this today, and am off the grid until Monday night. Thanks for the small repro -- let's find some time next week to discuss if no solution has presented itself by then.

jodersky:
I found a very very dirty, proof-of-concept hack, just overriding ambiguous situations in `Contexts`, always choosing the local type when typing java code.

jodersky:
@adriaanm just saw your PR, it looks alot cleaner :)

adriaanm:
Cool, it's also still a proof of concept, but it looks like it should work :-)

jodersky:
Adriaan's latest commit should fix the last issue. Hopefully all the corner cases are fixed now!

adriaanm:
```
[info] Project akka--------------------: SUCCESS (project rebuilt ok)
```

🎉

jodersky:
Glad the community build passes now!

Regarding review of this PR: every commit represents one component as to ease understanding and reviewing the changes. Not sure if you do it automatically, but I would recommend to squash all commits before merging this.

adriaanm:
I went over your changes -- they look good to me! I'd like a second opinion from @lrytz or @retronym on my changes around static member lookup in Java compilation units before merging this.

jodersky:
FYI: although an intermediary commit failed, the **final result passed** the jenkins verification.

I'll keep the intermediary commits for review and change the title back to "ci: last-only". I'll change back the title and squash commits upon final review.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
retronym, line 716: `DocDef`-s have a long history of such bugs :(

e.g. 1c6969945f209dafbe3c3fb38f1c3f9a5aaadb82 / c4561c1d4945a38febc41436ed333569d0e9a063

If you're motivated to do so, I'd love to see a followup change that changes the parser to no longer emit `DocDef` nodes at all, but instead parse the regular tree with an attachment `tree.updateAttachment(DocComment(...))`.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -135,6 +132,18 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
       DefDef(Modifiers(Flags.JAVA), nme.CONSTRUCTOR, List(), List(vparams), TypeTree(), blankExpr)
     }
 
+    def withConstructor(stats: List[Tree]): List[Tree] = {
+      val pureStats = stats map {
+        case DocDef(_, defn) => defn
+        case defn => defn
+      }
+      if(treeInfo.firstConstructor(pureStats) == EmptyTree) {
```
retronym: Move the `DocDef` unwrapping into `override def firstConstructor` in `scala.tools.nsc.ast.TreeInfo` to keep te `DocDef` special cases in one place?

retronym: That might even fix a latent bug in:

```
    def enterClassDef(tree: ClassDef) {
      val ClassDef(mods, _, _, impl) = tree
      val primaryConstructorArity = treeInfo.firstConstructorArgs(impl.body).size
```

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
jodersky, line 716: I agree that it would make things simpler by completely removing DocDef trees, however I can't promise anything as to when and if I'll get around to it.

retronym, line 716: I've raised a ticket for this (and linked to a quick prototype): https://github.com/scala/scala-dev/issues/198

adriaanm, line 716: Understood. The priority is to get RC1 out in any case. It's great to have Scaladoc finally work for Java sources!

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
jodersky: Thanks for the tip! I initially tried unwrapping in `scala.reflect.internal.TreeInfo` however `DocDefs` are only available in package `scala.tools.nsc.ast`. With so many trees and infos around, it makes it hard to see the forest for the trees! :smile:

jodersky:
@retronym, I took care of your comment in the last commit

src/compiler/scala/tools/nsc/ast/TreeInfo.scala
```diff
@@ -97,4 +97,12 @@ abstract class TreeInfo extends scala.reflect.internal.TreeInfo {
     case DocDef(_, definition) => isPureDef(definition)
     case _ => super.isPureDef(tree)
   }
+
+  override def firstConstructor(stats: List[Tree]): Tree = {
+    val unwrapped = stats map {
+      case DocDef(_, defn) => defn
```
adriaanm: I don't know whether it's needed, but the above methods recurse to handle nested `DocDef`s -- should this one do the same?

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -116,9 +116,9 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
     def makePackaging(pkg: RefTree, stats: List[Tree]): PackageDef =
       atPos(pkg.pos) {  PackageDef(pkg, stats) }
 
-    def makeTemplate(parents: List[Tree], stats: List[Tree]) = {
-      Template(parents, noSelfType, withConstructor(stats))
-    }
+    def makeTemplate(parents: List[Tree], stats: List[Tree]) =
+      Template(parents, noSelfType, if (treeInfo.firstConstructor(stats) == EmptyTree)
```
adriaanm, line 120: Nice! Much cleaner 👍

src/compiler/scala/tools/nsc/ast/TreeInfo.scala
jodersky: good catch! It's probably rare, but from all the bugs I've seen so far, I wouldn't be surprised if there is a `/**comment about */ /** a comment*/` somewhere out there

test/scaladoc/resources/SI-4826.java
```diff
@@ -254,6 +254,15 @@ private FI() {
 	public boolean defined(Object o);
     }
 
+    /** comment about */
+    /** a comment about */
+    /** a comment */
+    void foo() {}
+
+    /** someone forgot to uncomment */
```
adriaanm, line 262: 😀

src/compiler/scala/tools/nsc/ast/TreeInfo.scala
```diff
@@ -97,4 +97,12 @@ abstract class TreeInfo extends scala.reflect.internal.TreeInfo {
     case DocDef(_, definition) => isPureDef(definition)
     case _ => super.isPureDef(tree)
   }
+
+  override def firstConstructor(stats: List[Tree]): Tree = {
+    def unwrap(stat: Tree): Tree = stat match {
+      case DocDef(_, defn) => unwrap(defn)
+      case tree => tree
+    }
+    super.firstConstructor(stats map unwrap)
```
jodersky, line 106: It appears that `DocDef,DocDef` sequences actually also work without the recursive unwrapping. Can't hurt to change it though

test/files/pos/t2377b/a.scala
```diff
@@ -0,0 +1,3 @@
+object Test {
+  (new Q.Inner).foo // make sure foo's type is well-formed
```
retronym: It would strengthen the test to have: `.foo.someMethodOnInner_Builder` to check the right type is resolved.

test/files/pos/t2377b/a.scala
```diff
@@ -0,0 +1,3 @@
...
+}
```
retronym, line 5: Both `typedIdent` and `typedSelect` were modified. Are both new code paths exercised by the test?

I think we to add:

```
 public Inner.Builder baz() { return new Builder(); }
```

... and a call to it from `a.scala`.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4703,7 +4713,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           dyna.wrapErrors(t, (_.typed1(t, mode, pt)))
         }
 
-        val sym = tree.symbol orElse member(qual, name) orElse {
+        val sym = tree.symbol orElse member(qual, name) orElse inCompanionForJavaStatic(qual.tpe.prefix, qual.symbol, name) orElse {
```
retronym, line 4716: It is safe to just look at the type+symbol of the `qual` and discard the tree itself because we are only typechecking Java type signatures, not arbitrary Java expressions.

retronym:
Could you please introduce an option to the scaladoc tool to disable the new feature of processing Java sources? This will help people workaround further bugs or feature interactions that are discovered (e.g. what will this change mean for Scala IDE?), and concievably some people might not want Java sources reflected in the Scaladocs.

I think the right default for this option is to enable the feature.

Scaladoc options are defined in `src/scaladoc/scala/tools/nsc/doc/Settings.scala`

test/files/pos/t2377b/a.scala
retronym: Indeed, looks like we're binding to the outer `Builder` before the fallback case triggers

```
qscalac -Xprint:typer -uniqid test/files/pos/t2377b/*
[[syntax trees at end of                     typer]] // Q.java
package <empty> {
  object Q#6192 extends  {
    def <init>#14019();
    object Builder#14020 extends  {
      def <init>()
    };
    <static> class Builder#14022 extends _root_.java.lang.Object {
      def <init>()
    };
    object Inner#14023 extends  {
      def <init>();
      object Builder extends  {
        def <init>()
      };
      <static> class Builder extends _root_.java.lang.Object {
        def <init>()
      }
    };
    <static> class Inner#14025 extends _root_#5.java#10.lang.Object {
      def <init>#14026(): Q#6193.Inner#14025;
      def foo#14027(): Q#6193.Builder#14022
    }
  };
  class Q#6191 extends _root_.java.lang.Object {
    def <init>()
  }
}

 // a.scala
package <empty>#4 {
  object Test#6198 extends scala#22.AnyRef#1976 {
    def <init>#6708(): Test#6199.type = {
      Test#6199.super.<init>#2308();
      ()
    };
    new Q#6192.Inner#14025().foo#14027()
  }
}
```

```
⚡ tail -n 100 test/files/pos/t2377b/*
==> test/files/pos/t2377b/Q.java <==
public class Q {

  public static class Builder {}

  public static class Inner {
    public static class Builder {}
    public Builder foo() { return new Builder(); } // this line gives an error, that Builder is ambiguous
  }

}

==> test/files/pos/t2377b/a.scala <==
object Test {
  (new Q.Inner).foo.innerMethod // make sure foo's type is well-formed
}
/code/scala on pr/5304*
⚡ qscalac test/files/pos/t2377b/*
test/files/pos/t2377b/a.scala:2: error: value innerMethod is not a member of Q.Builder
  (new Q.Inner).foo.innerMethod // make sure foo's type is well-formed
                    ^
one error found
```

retronym: I think the logic needs to live in `lookupInPrefix` in `Context#lookupIdent`, rather than in `typedIdent`.

retronym: This closer, but still not quite right.

```
diff --git a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
index 16ef75c..cb432e3 100644
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -1016,13 +1016,22 @@ trait Contexts { self: Analyzer =>
           || unit.exists && s.sourceFile != unit.source.file
         )
       )
-      def lookupInPrefix(name: Name)    = pre member name filter qualifies
+      def lookupInPrefix(name: Name)    = {
+        val sym = pre.member(name).filter(qualifies)
+        def isNonPackageNoModuleClass(sym: Symbol) =
+          sym.isClass && !sym.isModuleClass && !sym.isPackageClass
+        if (!sym.exists && unit.isJava && isNonPackageNoModuleClass(pre.typeSymbol)) {
+          val found = companionSymbolOf(pre.typeSymbol, this).typeOfThis.member(name).filter(qualifies)
+          if (found.exists) pre = companionSymbolOf(pre.typeSymbol, this).typeOfThis
+          found
+        } else sym
+      }
       def accessibleInPrefix(s: Symbol) = isAccessible(s, pre, superAccess = false)

       def searchPrefix = {
         cx = cx.enclClass
         val found0 = lookupInPrefix(name)
-        val found1 = found0 filter accessibleInPrefix
+        var found1 = found0 filter accessibleInPrefix
         if (found0.exists && !found1.exists && inaccessible == null)
           inaccessible = LookupInaccessible(found0, analyzer.lastAccessCheckDetails)

diff --git a/test/files/pos/t2377b/a.scala b/test/files/pos/t2377b/a.scala
index 14be23c..5b0b655 100644
--- a/test/files/pos/t2377b/a.scala
+++ b/test/files/pos/t2377b/a.scala
@@ -1,3 +1,3 @@
 object Test {
-  (new Q.Inner).foo // make sure foo's type is well-formed
+  (new Q.Inner).foo.innerMethod // make sure foo's type is well-formed
 }
```

retronym:
@jodersky @adriaanm Here's a working version of my patch, including an expanded test.

https://github.com/scala/scala/compare/scala:pull/5304/head...retronym:review/5304?expand=1

The rationale is that in both typechecking a `Select` or an `Ident` we need to put the special case immediately after the failed call to `pre.member(name)`.

jodersky:
Thanks @retronym ! I'll go ahead and add and add a java processing option to scaladoc.

jodersky:
There was a conflict in Jason's review branch, I rebased now.

jodersky:
Hmm, looks like something broke while fixing the conflict.

retronym:
This patch should fix the test failures. I hadn't properly reverted the change in `typedIdent`.

``` diff
index eabd35c..f69612c 100644
@@ -4885,7 +4885,11 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         nameLookup match {
           case LookupAmbiguous(msg)         => issue(AmbiguousIdentError(tree, name, msg))
           case LookupInaccessible(sym, msg) => issue(AccessError(tree, sym, context, msg))
-          case LookupNotFound               => issue(SymbolNotFoundError(tree, name, context.owner, startContext))
+          case LookupNotFound               =>
+            inEmptyPackage orElse lookupInRoot(name) match {
+              case NoSymbol => issue(SymbolNotFoundError(tree, name, context.owner, startContext))
+              case sym      => typed1(tree setSymbol sym, mode, pt)
+                }
           case LookupSucceeded(qual, sym)   =>
             (// this -> Foo.this
             if (sym.isThisSym)
```

retronym:
I've included my fix, squashed a little and rebased to fix other conflicts and submitted as https://github.com/scala/scala/pull/5332

jodersky:
@retronym I really appreciate your help in this matter, but do you think creating a new PR from this one is the best way forward? I was going to squash everything once the final review was complete; it will now be up to you to do the cleanup and deal with any review comments.
If timing/response delay is a concern (I know you're in a different timezone), I can make you a collaborator on my scala fork, so you can update things as you please.
cheers,

retronym:
I was just expediting the process of getting a green tick on the end result and saving you the busywork of resolving merge conflicts and interpreting my comments on what to squash. 

You are credited as the author or the commits in the new PR, too. But I'm just as happy for you to reopen this one with a squash of my branch.

jodersky:
Ok, it's all yours then ;) I just wanted to emphasize that my intermediary commits will fail the community build (they pass validation checks though) and therefore should be squashed into one before merging into master.

PR #5305, opened 2016-07-26
Dotty compat: move Numeric#Ops to Numeric.Ops

adriaanm:
Dotty does not allow name clashes of inner classes,
and the inner class is polluting Numeric's interface anyway.

Note that a similar case can be made for Ordering#Ops.

Also note that this breaks source compatibility.

Fix scala/scala-dev#156


adriaanm:
ugh

PR #5306, opened 2016-07-27
SI-8576 Minimal changes for `-Xcheckinit` compatibility

szeiger:
As explained in https://issues.scala-lang.org/browse/SI-8576, I expect
serialization compatibility between builds with and without
`-Xcheckinit` to be unattainable. This commit contains some minor fixes
for issues discovered while running builds with `-Xcheckinit`:
- Add `@SerialVersionUID` to `scala.collection.immutable.Vector`, as
  requested in SI-8576. Note that this does not make `Vector`
  serialization compatible.
- Use lazy initialization for `global` in `PresentationCompilation`. It
  used to access the uninitialized `self` variable (which seems to be
  inconsequential in practice and only fails under `-Xcheckinit`).

We should consider using `Externalizable` instead of `Serializable` for
collections in 2.13 to make collection classes serialization compatible.

### DIFF
--- a/src/library/scala/collection/immutable/Vector.scala
+++ b/src/library/scala/collection/immutable/Vector.scala
@@ -59,6 +59,7 @@ object Vector extends IndexedSeqFactory[Vector] {
  *  @define mayNotTerminateInf
  *  @define willNotTerminateInf
  */
+@SerialVersionUID(-1334388273712300479L)
 final class Vector[+A] private[immutable] (private[collection] val startIndex: Int, private[collection] val endIndex: Int, focus: Int)
 extends AbstractSeq[A]
    with IndexedSeq[A]
--- a/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala
+++ b/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala
@@ -69,7 +69,7 @@ trait PresentationCompilation {
     val interactiveGlobal = new interactive.Global(copySettings, storeReporter) { self =>
       override lazy val platform: ThisPlatform = {
         new JavaPlatform {
-          val global: self.type = self
+          lazy val global: self.type = self
           override private[nsc] lazy val classPath: ClassPath = mergedFlatClasspath
         }
       }


szeiger:
Successful run with `-Xcheckinit`: https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/2648/

soc:
Isn't `Externalizable` even more deprecated than `Serializable`?

szeiger:
Neither one is deprecated in Java. Java serialization may not be the most efficient but it's still the simplest and most ubiquitous. I wouldn't want to mix `Externalizable` into any class that's meant to be user-extensible but using it in final collection classes like `Vector` makes sense. You don't really care about preserving the exact internal structure of a collection. If you want the best possible performance, you use something else anyway. We could store all collections simply as class name plus length plus contents. Deserializing could be implemented in a generic way: Get the class, initialize a `newBuilder` with the length and feed it the contents.

retronym:
Is this equivalent to the serialization proxy pattern we already use in some cases with `readResolve` / `writeReplace`?

szeiger:
Yes, I think so. I wonder why it was done with extra proxy classes and `readObject` / `writeObject` instead of just implementing `Externalizable` directly. The latter would be simpler.

retronym:
Effective Java recommends the pattern as an alternative to direct serialization.  It doesn't discuss `Externalizable` as an alternative.

soc:
Example of this pattern: https://github.com/soc/scala-java-time/blob/master/shared/src/main/scala/org/threeten/bp/Ser.scala and related classes.

szeiger:
For the record, after working on SI-7838 I now know why you want to use a serialization proxy instead of Externalizable: The latter requires mutable state in the object that is being deserialized, so you cannot get the state properly published without external synchronization or making all those mutable fields volatile.

SethTisue:
was the change to project/ScriptCommands.scala intended to be included? normally I don't think having commented-out stuff lying around is good, but if you're attached to it, so be it.

otherwise LGTM

szeiger:
It was meant to show how to enable Xcheckinit, but it's not essential.

SethTisue:
LGTM then once Travis passes

szeiger:
What's up with Jenkins today? Lots of spurious failures (probably OOM).

adriaanm:
https://github.com/scala/scala-jenkins-infra/issues/181

PR #5307, opened 2016-07-27
Propagate overloaded function type to expected arg type

adriaanm:
Infer missing parameter types for function literals passed
to higher-order overloaded methods by deriving the
expected argument type from the function types in the
overloaded method type's argument types.

This eases the pain caused by methods becoming overloaded
because SAM types and function types are compatible,
which used to disable parameter type inference because
for overload resolution arguments are typed without
expected type, while typedFunction needs the expected
type to infer missing parameter types for function literals.

It also aligns us with dotty. The special case for
function literals seems reasonable, as it has precedent,
and it just enables the special case in typing function
literals (derive the param types from the expected type).

Since this does change type inference, you can opt out
using the Scala 2.11 source level.

Fix scala/scala-dev#157

### DIFF
--- a/spec/06-expressions.md
+++ b/spec/06-expressions.md
@@ -1426,9 +1426,14 @@ Let $\mathscr{B}$ be the set of alternatives in $\mathscr{A}$ that are [_applica
 to expressions $(e_1 , \ldots , e_n)$ of types $(\mathit{shape}(e_1) , \ldots , \mathit{shape}(e_n))$.
 If there is precisely one alternative in $\mathscr{B}$, that alternative is chosen.
 
-Otherwise, let $S_1 , \ldots , S_m$ be the vector of types obtained by
-typing each argument with an undefined expected type.  For every
-member $m$ in $\mathscr{B}$ one determines whether it is applicable
+Otherwise, let $S_1 , \ldots , S_m$ be the list of types obtained by typing each argument as follows.
+An argument `$e_i$` of the shape `($p_1$: $T_1 , \ldots , p_n$: $T_n$) => $b$` where one of the `$T_i$` is missing,
+i.e., a function literal with a missing parameter type, is typed with an expected function type that
+propagates the least upper bound of the fully defined types of the corresponding parameters of
+the ([SAM-converted](#sam-conversion)) function types specified by the `$i$`th argument type found in each alternative.
+All other arguments are typed with an undefined expected type.
+
+For every member $m$ in $\mathscr{B}$ one determines whether it is applicable
 to expressions ($e_1 , \ldots , e_m$) of types $S_1, \ldots , S_m$.
 
 It is an error if none of the members in $\mathscr{B}$ is applicable. If there is one
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -731,7 +731,7 @@ trait Infer extends Checkable {
       // If args eq the incoming arg types, fail; otherwise recurse with these args.
       def tryWithArgs(args: List[Type]) = (
            (args ne argtpes0)
-        && isApplicable(undetparams, mt, args, pt)
+        && isApplicableToMethod(undetparams, mt, args, pt) // used to be isApplicable(undetparams, mt, args, pt), knowing mt: MethodType
       )
       def tryInstantiating(args: List[Type]) = falseIfNoInstance {
         val restpe = mt resultType args
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3271,40 +3271,74 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       }
 
       val fun = preSelectOverloaded(fun0)
+      val argslen = args.length
 
       fun.tpe match {
         case OverloadedType(pre, alts) =>
           def handleOverloaded = {
             val undetparams = context.undetparams
+
+            def funArgTypes(tps: List[Type]) = tps.map { tp =>
+              val relTp = tp.asSeenFrom(pre, fun.symbol.owner)
+              val argTps = functionOrSamArgTypes(relTp)
+              //println(s"funArgTypes $argTps from $relTp")
+              argTps.map(approximateAbstracts)
+            }
+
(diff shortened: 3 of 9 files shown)


adriaanm:
@DarkDimius, @smarter -- does this sound like what you are doing in Dotty? I tried to understand how this is implemented in dotty (https://github.com/lampepfl/dotty/issues/1378?) but couldn't figure out how you actually propagate the argument types of the function types in the arguments of the overloaded method type.

retronym:
IIRC, dotty does something questionable for:

``` scala
trait C[A] {
  def foo(a: A => String, z: A): Unit
  def foo(a: Int => String, z: Int): Unit
}

trait Test {
  def test(c: C[Int]) = c.foo(x => x.toString, 0)
}
```

When computing the members of `C[Int]`, it considers the two foos to overlap and picks one over the other. This happens in:

```
      at dotty.tools.dotc.core.Denotations$DenotUnion.matches(Denotations.scala:970)
      at dotty.tools.dotc.core.Denotations$SingleDenotation.filterDisjoint(Denotations.scala:817)
      at dotty.tools.dotc.core.Denotations$SingleDenotation.mapInherited(Denotations.scala:821)
      at dotty.tools.dotc.core.Denotations$SingleDenotation.mapInherited(Denotations.scala:457)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation$$anonfun$computeNPMembersNamed$1.collect$1(SymDenotations.scala:1578)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation$$anonfun$computeNPMembersNamed$1.apply(SymDenotations.scala:1586)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation$$anonfun$computeNPMembersNamed$1.apply(SymDenotations.scala:1563)
      at dotty.tools.dotc.util.Stats$.track(Stats.scala:36)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation.computeNPMembersNamed(SymDenotations.scala:1562)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation.nonPrivateMembersNamed(SymDenotations.scala:1552)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation.membersNamed(SymDenotations.scala:1539)
      at dotty.tools.dotc.core.SymDenotations$ClassDenotation.findMember(SymDenotations.scala:1591)
```

I remember discussing this on a previous dotty ticket or PR, but I can't find it at the moment.

odersky:
@adriaanm The dotty PR that introduced something similar was: 

https://github.com/lampepfl/dotty/pull/1160

Hope this helps!

adriaanm:
It does, thanks Martin! At first sight, looks like we ended up at the same solution, but I'll experiment a bit more to make sure.

lrytz:
I started reviewing this and hit a weird case. Not introduced by this PR, it already exists in M5, but it looks related:

```
scala> trait MF1[-A,+R] { def apply(a: A): R }
defined trait MF1

scala> (x => x) : MF1[String, String]
res0: MF1[String,String] = $$Lambda$1140/992955027@213bd3d5

scala> trait A extends MF1[String, String]
defined trait A

scala> (x => x) : A
res1: A = $$Lambda$1139/1562090557@69e05f61

scala> (x => x) : Function1[String, String]
res2: String => String = $$Lambda$1149/102709691@249e0271
```

So far so good, but now:

```
scala> trait B extends Function1[String, String]
defined trait B

scala> (x => x) : B
<console>:13: error: missing parameter type
       (x => x) : B
        ^
```

Adding the param type: note that we don't get an indyLMF-lambda, but an AOT anonymous class

```
scala> ((x: String) => x) : B
res4: B = <function1>
```

lrytz:
The problem of requiring a param type seems to be here: https://github.com/scala/scala/blob/961c8bea011559af65cfae9818d71ad35b6ecaf4/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L776

The sam symbol is the `apply` method inherited from `Function1`, so `sam.owner == correspondingFunctionSymbol`. The expected type is therefore set to `NoType`.

This simple fix seems fine: https://github.com/lrytz/scala/commit/b7a5f0e9c0c266121d93b0c46f976dfd6729dcf7

lrytz:
The problem of not using indyLMF is because the parent `Function1` has a constructor method `$init$`, so `compilesToPureInterface` is `false` (https://github.com/scala/scala/blob/961c8bea011559af65cfae9818d71ad35b6ecaf4/src/compiler/scala/tools/nsc/transform/Erasure.scala#L1226). It seems that the presence of any concrete trait method causes an `$init$` method to be added, which disqualifies the interface for LMF:

```
scala> trait A { def f(x: Int): Int }
defined trait A

scala> (x => x) : A
res0: A = $$Lambda$1722/816302479@36361ddb

scala> trait B { def f(x: Int): Int ; def blip = 1 }
defined trait B

scala> (x => x) : B
res1: B = $anonfun$1@7a2b1eb4
```

Related to https://github.com/scala/scala-dev/issues/191

Maybe we should allow no-op init methods?

lrytz:
the change looks good so far - i'll avoid the magic letters for now as you mention you'd like to experiment a bit more.

adriaanm:
Thanks, I'll also incorporate your fix. Regarding no-op ctors, maybe we should look into that before RC1? :-s OR we could reuse FunctionalInterface as an indication that a trait should not receive a ctor, and add it as an other sufficient condition for sammyness.

retronym:
> OR we could reuse FunctionalInterface as an indication that a trait should not receive a ctor, and add it as an other sufficient condition for sammyness.

I like this idea a lot.

retronym:
> Maybe we should allow no-op init methods?

The problem here is separate compilation.

adriaanm:
Tracking the `@FunctionalInterface` idea as https://github.com/scala/scala-dev/issues/197

adriaanm:
Community build ✅  (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/473/console)

adriaanm:
With the community build green, that concludes my experimentation/sanity checking :-)

adriaanm:
Rebased.

adriaanm:
LGTM-echo of Lukas's earlier review (which was pending my experimenting a bit with the community build)

adriaanm:
This caused [a regression](https://issues.scala-lang.org/browse/SI-9943). Maybe we need to follow dotty more closely and require identical argument types.. :-/

PR #5308, opened 2016-07-28
SI-3809 Avoid cycles during classfile parsing

retronym:
Manually tested with:

```
scalaHome := Some(file("/code/scala/build/pack"))
libraryDependencies += "com.datastax.cassandra" % "dse-driver" % "1.0.0"
```

```
class Test {
  new com.datastax.driver.dse.DseCluster.Builder()
}
```

I was also able to compile `pos/cycle{-jsoup,}` without the experimental
`-Ybreak-cycles`. I've removed the entire `-Ybreak-cycles` option and
supporting code in favour of the approach in this patch.

What's changed? I've used lazy types for methods infos, which is analagous
to what we do in `Unpickler` for scala originated types.

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/AbstractFileReader.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/AbstractFileReader.scala
@@ -19,11 +19,9 @@ import scala.tools.nsc.io.AbstractFile
  * @author Philippe Altherr
  * @version 1.0, 23/03/2004
  */
-class AbstractFileReader(val file: AbstractFile) {
+class AbstractFileReader(val file: AbstractFile, var buf: Array[Byte]) {
 
-  /** the buffer containing the file
-   */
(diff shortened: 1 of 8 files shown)


retronym:
```
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 123207680 bytes for committing reserved memory.
# Possible reasons:
#   The system is out of physical RAM or swap space
#   In 32 bit mode, the process size limit was hit
# Possible solutions:
```

retronym:
I've realized that the addition of laziness here will increase the footprint: the `ConstantPool`, which we access in the type completer, references `$outer.in.buf`, the entire contents of the class file.

I've just pushed a WIP commit to might alleviate this problem.

Swoorup:
This is also occuring on dotty. https://github.com/lampepfl/dotty/issues/9492
Any ideas how to port the fix over? @retronym

PR #5310, opened 2016-07-29
SIP-23 Implementation of literal types

milessabin:
Implements literal types. Enabled by `-Yliteral-types` or `-Xexperimental`.

This is a continuation of work by @folone and @adriaanm. The work on @adriaanm's [branch](https://github.com/adriaanm/scala/tree/sip23) was the starting point and all the TODO items on the corresponding [PR](https://github.com/adriaanm/scala/pull/21) have been completed.
- Literals can now appear in type position, designating the corresponding singleton type.
- Widening/narrowing behaviour is consistent with 2.11.x and between 2.12.x with the flag enabled and disabled. The use of `final` on a val definition to explicitly request an unwidened type to be inferred is preserved. Type parameter inference remains the same unless the inference of a singleton type is explicitly requested via a `<: Singleton` bound.
- A `scala.ValueOf[T]` type class and corresponding `scala.Predef.valueOf[T]` operator has been added yielding the unique value of types with a single inhabitant.
- `isInstanceOf`/`asInstanceOf` tests/conversions are implemented via equality/identity tests for singleton types.
- Support for `scala.Symbol` literal types has been added.
- The spec markdown files have been updated.
- Many tests for all of the above have been added.
- All (par)tests pass.
- With `-Yliteral-types` enabled on _all_ tests there are 15 test result changes, all of which are expected. See below for details.
- Types reported in error messages are widened where appropriate, matching behaviour with the flag disabled (eg. we report 'Found String expected Int' instead of 'Found "foo" expected Int').
- The public macro API is unchanged relative to 2.11.x. In contrast to @adriaanm's branch there are now non-API `LiteralConstantType` and `FoldableConstantType` subtypes of the API type `ConstantType`.
- shapeless is likely to be the library most directly affected by the above changes and, with only minor 2.11.x source-compatible [changes](https://github.com/milessabin/shapeless/commit/b9774c2ab2d8ebc2232399a2e7c2b21c028ba710), it compiles and passes all tests with `-Yliteral-types` both enabled and disabled.
### Related bugs fixed in this PR
- [SI-1273](https://issues.scala-lang.org/browse/SI-1273) Singleton type has wrong bounds
- [SI-5103](https://issues.scala-lang.org/browse/SI-5103) Singleton types not inferred in all places they should be
- [SI-8323](https://issues.scala-lang.org/browse/SI-8323) Duplicate method name & signature with singleton type parameters over constant types
- [SI-8564](https://issues.scala-lang.org/browse/SI-8564) Methods with ConstantType results get the inhabitant of ConstantType as their body
#### New warnings due to Symbol literals becoming pure:
- test/files/instrumented/indy-symbol-literal.scala
#### Error messages changed due to more precise types being reported:
- test/files/neg/t8323.scala
- test/files/neg/names-defaults-neg.scala
#### Pos tests which correctly become neg with -Yliteral-types
- test/files/pos/unchecked-a.scala
#### Neg tests which correctly become pos with -Yliteral-types:
- test/files/neg/any-vs-anyref.scala
- test/files/neg/not-possible-cause.scala
- test/files/neg/t6263.scala
- test/files/neg/t900.scala
#### Error messages changed correctly:
- test/files/neg/t8463.scala
- test/files/run/repl-colon-type.scala
#### Compiler/REPL output changed due to more precise types being reported:
- test/files/neg/logImplicits.scala
- test/files/run/analyzerPlugins.scala
- test/files/run/delambdafy_uncurry_byname_inline.scala
#### Compiler/REPL output changed due to Symbols becoming first-class literals:
- test/files/run/t6549.scala
#### Run tests which correctly fail due to improved inference
- test/files/run/StubErrorHK.scala
- test/files/run/repl-kind.scala
#### Unavoidable failure due to the enablement of the feature:
- test/files/run/settings-parse.scala


spec/12-the-scala-standard-library.md
```diff
@@ -642,6 +642,12 @@ object Predef {
   def classOf[T]: Class[T] = null
    // this is a dummy, classOf is handled by compiler.
 
+  // valueOf -----------------------------------------------------------
+
+  /** Retrieve the single value of a type with a unique inhabitant. */
+  @inline def valueOf[T](implicit vt: ValueOf[T]): T {} = vt.value
```
soc, line 644: The signature should be `def valueOf[T]: T {}`. Avoid exposing the implicit parameter. (Have a look at `classOf` for how it is done.)

src/library/scala/ValueOf.scala
```diff
@@ -0,0 +1,35 @@
...
+ * case class SList[T, L <: Int](l: List[T]) {
+ *   def length(implicit l: ValueOf[L]): Int = l.value
+ * }
+ *
+ * val l3 = Sized[Int, 3](List(1, 2, 3))
+ * l3.length // == 3
+ * }}}
+ */
+@scala.annotation.implicitNotFound(msg = "No singleton value available for ${T}.")
+final class ValueOf[T](val value: T) extends AnyVal
```
soc: I think we should not expose this, especially not in the main `scala` namespace.

milessabin: We definitely should. Between `def valueOf` and `ValueOf` it's the type class which is the most important ... see the (many) uses of `Witness` in shapeless.

spec/12-the-scala-standard-library.md
milessabin, line 644: See comment on `ValueOf` below. It's the type class which is important, this def very much less so.

xeno-by:
Why is the macro API left unchanged? What if metaprogrammers want to explicitly control the flavor of constant types?

Also, could you document the differences between LiteralConstantType and FoldableConstantType? At the moment, these are classes with no comments or documentation.

src/library/scala/ValueOf.scala
lihaoyi: I've wanted to enforce singleton-ness before, in random places, e.g. to ensure that you only pass in literal strings to a macro-DSL that reads the string at compile time and does stuff to it. Currently you can use `assert`s in the macro to make sure you get a literal, but having a typeclass would be way better.

soc: @milessabin 

What's is the benefit of letting users write code like

```
class Horse
val horse = valueOf(new ValueOf(new Horse))
```

@lihaoyi

How is this achieved with this code? Every user can define arbitrary instances of `ValueOf`s and pass them around.

spec/12-the-scala-standard-library.md
soc, line 644: The style of the signature is not related to its importance. It should not be defined this way, regardless of importance.

milessabin:
@xeno-by four reasons,
- an API change can't be enabled via a feature flag
- source compatibility with 2.11.x
- the decision about whether to fold or not is one for the compiler, not a macro author
- given that the 2.11/12 macro API is deprecated in favour of Scala.meta it seems inadvisable to change it unless absolutely necessary.

src/library/scala/ValueOf.scala
milessabin: @soc that's not the intended usage, but it seems harmless not to prevent it.

For an example of the intended use see the examples in the Scaladoc or any use of shapeless's `Witness[T]` in any of the many libraries which use it for automatically deriving type class instances for codecs.

soc: > that's not the intended usage, but it seems harmless not to prevent it.

I really like your proposal, but even the library-level modifications `valueOf`, `ValueOf` you suggest are so deeply interconnected that they become language features, and they have to be treated that way:

There is no place for hoping that people only follow _intended usage_ or assume that it's _harmless not to prevent_ the usage of public methods you are adding to the main namespace of Scala that gets imported into every source file.

The _intended usage_ needs to be the only usage the compiler allows. It can and should leverage the compiler's facilities that already exist to do so.

milessabin: @soc would you be happy with the constructor being made private?

soc: Could be an option. Alternatively, I'd suggest looking into things like `sealed abstract ValueOf[T] private ...`, `final abstract ValueOf[T] private ...`, ... to close as many reflection/serialization/unsafe/... loopholes as possible.

Additionally, I think it's necessary to think about a proper name for `ValueOf`. `classOf` doesn't return `ClassOf` either.

milessabin: Agreed on `final` and a private constructor. I think the name `ValueOf` works well.

soc: There is no precedence of the "XyzOf" style in the standard library and I think it would inconsistent to introduce it.

I think something like `scala.reflect.Value` or `scala.reflect.Singleton` would be better.

src/library/scala/ValueOf.scala
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * `ValueOf[T]` provides the unique value of the type `T` where `T` is a type which has a
+ * single inhabitant. Eligible types are singleton types of the form `stablePath.type`,
+ * Unit and, if the -Yliteral-types compiler option has been specified, singleton types
+ * corresponding to value literals.
+ *
+ * Instances of `ValueOf[T]` are provided implicitly for all eligible types. Typically
+ * an instance would be required where a runtime value corresponding to a type level
+ * computation is needed. For example, we might define a type of list which encodes its
+ * length in its type, in which case the lenght of the list can be computed as a
```
julienrf: Typo: leng**ht**

spec/12-the-scala-standard-library.md
soc, line 644: Have a look at `resolveClassTag` to see how it's possible to resolve an implicit for your call without having to specify that your method takes one (and therefore allowing people to accidentally or maliciously pass one explicitly).

notxcain, line 644: @soc What's wrong with this signature? stdlib have a lot of methods with implicit parameters. Isn't it the same as `def classTag[T](implicit ctag: ClassTag[T]) = ctag`?

milessabin:
I think those failures are spurious ...

spec/12-the-scala-standard-library.md
soc, line 644: There are methods where the user is never supposed to be passing the evidence, like `classOf` or `new Array ...`, and `valueOf` seems to be a new addition. This new addition should be consistent with what we have already.

There will always be a conflict between making things bullet-proof and "it's just a standard method", but in the case of things living in `scala` they should absolutely be bullet-proof.

propensive:
Just found a couple of minor things from early experiments:
- `foo.isInstanceOf[Singleton]` will always return true. Additionally, you get a spurious "fruitless type test" warning if you check `foo.isInstanceOf[String with Singleton]`.
- interestingly, in the REPL, `val x: "foo" = "foo"`'s return type is printed as `"foo"`, whereas `final val x = "foo"`'s is printed as `String("foo")`. Is there a deliberate distinction?

@milessabin Would you like me to continue reporting issues on this PR, or is there somewhere better?

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -1255,9 +1264,12 @@ self =>
      */
     def literal(isNegated: Boolean = false, inPattern: Boolean = false, start: Offset = in.offset): Tree = atPos(start) {
       def finish(value: Any): Tree = try newLiteral(value) finally in.nextToken()
-      if (in.token == SYMBOLLIT)
-        Apply(scalaDot(nme.Symbol), List(finish(in.strVal)))
-      else if (in.token == INTERPOLATIONID)
+      if (in.token == SYMBOLLIT) {
+        if(settings.YliteralTypes)
+          finish(Symbol(in.strVal.intern()))
```
retronym: I'd prefer to represent these with a distinct type from `Symbol` to avoid a memory leak in a resident compiler (`Symbol`-s are stored in a hash map).

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -1099,7 +1101,14 @@ self =>
...
             mkOp(infixType(InfixMode.RightOp))
         }
+        def isNegatedLiteralType = settings.YliteralTypes && (
+          t match { // the token for `t` (Ident("-")) has already been read, thus `isLiteral` below is looking at next token (must be a literal)
+            case Ident(name) if isLiteral => name == nme.MINUS.toTypeName
+            case _ => false
+          }
+        )
         if (isIdent) checkRepeatedParam orElse asInfix
+        else if (isNegatedLiteralType) atPos(t.pos.start){SingletonTypeTree(literal(isNegated = true, start = t.pos.start))}
```
retronym: It would be good to include a test that covers all the parser changes and that runs with `-Yrange-pos` enabled. We can also eyeball the output of `-Yrange-pos -Xprint:parser -Xprint-pos` for that file to check that positions are precise.

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
retronym: I forgot that it is actually a weak map, so we can keep things as they are.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1123,6 +1162,8 @@ abstract class Erasure extends AddInterfaces
           }
           treeCopy.Literal(tree, Constant(erased))
 
+        case Literal(ct) if ct.tag == SSymbolTag =>
```
retronym: Let's add `Constant#isSymbol` and use that here.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1123,6 +1162,8 @@ abstract class Erasure extends AddInterfaces
...
+          treeCopy.Apply(tree, Select(gen.scalaDot(nme.Symbol), "apply"), List(Literal(Constant(ct.scalaSymbolValue.name))))
```
retronym:
```
atPos(tree.pos) {
  gen.mkMethodCall(definitions.Symbol_apply, Literal(Constant(ct.scalaSymbolValue.name)))
}
```

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1155,6 +1196,17 @@ abstract class Erasure extends AddInterfaces
...
               treeCopy.ArrayValue(
                 tree1, elemtpt setType specialScalaErasure.applyInArray(elemtpt.tpe), trees map transform).clearType()
+            case ValDef(_, _, tpt, rhs) =>
+              try super.transform(tree1).clearType()
+              finally {
+                tpt.tpe match {
+                  case FoldableConstantType(_) if !rhs.isInstanceOf[Literal] =>
+                    val deconst = tpt.tpe.deconst
+                    tpt setType deconst
+                    tree1.symbol.setInfo(deconst)
```
retronym: This section of code assumes that `super.transform(tree1)` resulted in a tree with `tpt` was preserved. That's likely to be the case looking at the other cases in the transform, but it seems somewhat fragile. I'd prefer:

``` scala
val vd1 = super.transform(....).clearType().asInstanceOf[ValDef]
vd1.tpt1 match { ... }
vd1
```

src/compiler/scala/tools/nsc/transform/TailCalls.scala
```diff
@@ -116,7 +116,9 @@ abstract class TailCalls extends Transform {
       def tailLabels: Set[Symbol]
 
       def enclosingType = method.enclClass.typeOfThis
-      def isEligible    = method.isEffectivelyFinalOrNotOverridden
+      def isEligible    = method.isEffectivelyFinalOrNotOverridden && !isExcludedBySingletonArg && !isExcludedBySingletonResult
+      def isExcludedBySingletonArg = method.paramss.exists(_.exists(_.tpe_*.isInstanceOf[SingletonType]))
+      def isExcludedBySingletonResult = method.tpe_*.finalResultType.isInstanceOf[SingletonType]
```
retronym: We've got `mexists` for the `exists(_.exists(_ => )` pattern. More utilities like it are with `reflect.internal.util.Collections`.

test/files/neg/t8323.scala
```diff
@@ -0,0 +1,7 @@
+object Test {
+  final val a = "Bippy"
+  final val b = "Dingo"
+  def f(x: a.type) = x
+  def f(x: b.type) = x
+  def main(args: Array[String]): Unit = ()
+}
```
julienrf, line 7: Would it be possible, at some point, to be able to define overloads like here? That could be convenient for defining JavaScript facades.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1324,6 +1324,20 @@ trait Implicits {
       }
     }
 
+    /** Creates a tree will produce a ValueOf instance for the requested type.
+      * An EmptyTree is returned if materialization fails.
+      */
+    private def valueOfType(tp: Type): SearchResult = {
+      def success(t: Tree) = wrapResult(Apply(Select(New(gen.scalaDot(tpnme.ValueOf)), nme.CONSTRUCTOR), List(t)))
```
retronym:
``` scala
gen.mkMethodCall(defintitions.ValueOfClass.primaryConstructor, t :: Nil)
```

src/library/scala/Predef.scala
```diff
@@ -125,6 +125,22 @@ object Predef extends LowPriorityImplicits with DeprecatedPredef {
...
+   * val foo = valueOf[Foo.type]
+   * // foo is Foo.type = Foo
+   *
+   * // If -Yliteral-types has been specified
+   * val bar = valueOf[23]
+   * // bar is 23.type = 23
+   * }}}
+   * @group utilities
+   */
+  @inline def valueOf[T](implicit vt: ValueOf[T]): T {} = vt.value
```
retronym: Why `T{}` as the result type?

src/reflect/scala/reflect/internal/Definitions.scala
```diff
@@ -525,6 +526,8 @@ trait Definitions extends api.StandardDefinitions {
 
     lazy val StringContextClass           = requiredClass[scala.StringContext]
 
+    lazy val ValueOfClass                 = getRequiredClass("scala.ValueOf")
```
retronym: `getClassIfDefined` to support running this compiler with older standand libraries gracefully.

retronym:
Could you add @adriaanm and @folone as co-authors in the commit comment? I think the convention in Git is:

```
 Co-authored-by: Some One <some.one@example.foo>
```

src/library/scala/ValueOf.scala
milessabin: Well spotted.

milessabin:
@propensive yes, please do report issues here.

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
milessabin: :+1:

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
milessabin: Agreed.

src/compiler/scala/tools/nsc/transform/Erasure.scala
milessabin: :+1:

src/compiler/scala/tools/nsc/transform/Erasure.scala
milessabin: :+1:

src/compiler/scala/tools/nsc/transform/Erasure.scala
milessabin: :+1:

src/compiler/scala/tools/nsc/transform/TailCalls.scala
milessabin: Thanks for the pointer.

test/files/neg/t8323.scala
milessabin, line 7: It's possible now if the result types differ sufficiently to distinguish the overloads for the JVM post-erasure.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: Thanks.

src/library/scala/Predef.scala
milessabin: The empty refinement is an "accidental feature" (present since forever, exploited by shapeless for a couple of years) which prevents the result type from being widened in the calling context.

src/reflect/scala/reflect/internal/Definitions.scala
milessabin: :+1:

milessabin:
@retronym re. Co-authored-by ... yes, of course.

src/library/scala/Predef.scala
milessabin: @retronym See the test [here](https://github.com/milessabin/scala/blob/topic/sip-23-redux/test/files/pos/sip23-narrow.scala).

spec/03-types.md
```diff
@@ -102,16 +103,46 @@ forms.
...
 
-A singleton type is of the form $p.$`type`, where $p$ is a
-path pointing to a value expected to [conform](06-expressions.html#expression-typing)
-to `scala.AnyRef`. The type denotes the set of values
-consisting of `null` and the value denoted by $p$.
+A singleton type is of the form $p.$`type`. Where $p$ is a path pointing to a
+value which [conforms](06-expressions.html#expression-typing) to
+`scala.AnyRef`, the type denotes the set of values consisting of `null` and the
+value denoted by $p$ (i.e., the value $v$ for which `v eq p`). Where the path
+does not conform to `scala.AnyRef` the type denotes the set consiting of only
```
fthomas: s/consiting/consisting/

spec/03-types.md
```diff
@@ -102,16 +103,46 @@ forms.
...
+A literal type `lit` is a special kind of singleton type which denotes the
+single literal value `lit`.  Thus, the type ascription `1: 1` gives the most
+precise type to the literal value `1`:  the literal type `1`.
+
+At run time, an expression `e` is considered to have literal type `lit` if `e ==
+lit`.  Concretely, the result of `e.isInstanceOf[lit]` and `e match { case _ :
+lit => }` is determined by evaluating `e == lit`.
+
+Literal types are available for all types, apart from `Unit`, for which there is
+dedicated syntax. This includes the numeric types (other than `Byte` which
```
fthomas: `Short` also has no dedicated syntax, right?

spec/03-types.md
milessabin: Well spotted.

spec/03-types.md
milessabin: Yes, you're right ... will fix.

src/library/scala/ValueOf.scala
nafg: I don't think there's anything that establishes XXXOf as a choice that has been avoided. `Singleton` is not a typeclass, so it's not comparable. `classOf` doesn't have a corresponding typeclass so it's also not comparable.

@soc perhaps you could list all the analogous typeclasses in stdlib and derive a convention that is relevant? If yes it would be more constructive than just rejecting this name -- and if not then I don't hear an argument that it's inconsistent.

src/library/scala/ValueOf.scala
```diff
@@ -0,0 +1,35 @@
...
+ * runtime value directly from the type without needing either runtime storage or a
+ * traversal of the list value,
+ *
+ * {{{
+ * // List of T of length L ...
+ * case class SList[T, L <: Int](l: List[T]) {
+ *   def length(implicit l: ValueOf[L]): Int = l.value
+ * }
+ *
+ * val l3 = Sized[Int, 3](List(1, 2, 3))
```
fthomas: s/Sized/SList/

Also in the definition of `SList` `l` is used for the wrapped list and the `ValueOf` parameter. I would change the constructor argument to `list` to make the example more clear.

src/library/scala/ValueOf.scala
soc: > If yes it would be more constructive than just rejecting this name

@nafg I think it's the duty of those who want to add things to show that it's worth it, not those who end up doing the maintenance that it's not.

Given that there are too many things in the default namespace already (`java.lang` alone adds almost 100 names) and we struggled with deprecations for the last 4 major releases, I think a convincing argument needs to be made how this class is significantly more popular than the things already in `scala` and especially than those that were deprecated and removed.

This weights even stronger considering considering that `ValueX` would exist in that namespace, regardless of whether a compiler option for which we officially don't provide any support is switched on or off.

@nafg If you want to follow existing conventions, how about `scala.reflect.CanValueFrom` or `scala.reflect.ValueTag`?

dwijnand: I think "duty" is too strong a word, but I agree to your general sentiment. But it's still a lot more constructive and collaborative to provide more than just rejection.

Ultimately online communication is a lot hard to judge and deliver, and I believe we're all trying to do the right thing ("we're on the same team"), but I must say (at least to me) your comments felt a bit harsher than strictly necessary.

soc: Hi @dwijnand, thanks for your note. I'll try to phrase things in a less harsher way.

> But it's still a lot more constructive and collaborative to provide more than just rejection.

I don't want to start "he said this, he said that", but I have reviewed my comments again and I believe I proposed alternatives to all things I rejected, as well as explaining why they are not acceptable from my point of view.

soc: One thing I'd like to learn from the proponents of this: How much of the usage of `ValueX` do you imagine to be in the form of an implicit parameter, and how much to be as a context bound?

spec/03-types.md
alexeyr: It seems to be slightly non-obvious to me whether "for which there is dedicated syntax" binds to "for all types" or to "`Unit`". Also, there is dedicated syntax for tuples, functions, and XML, which also needs to be excluded. It also possibly should be specified that the dedicated syntax has to be for _values_ of the type: there is dedicated syntax for e.g. refinement types, just not for their values.

So I suggest something like: "Literal types are available for all types which have [value literals](01-lexical-syntax.html#literals), apart from (or except for) `Unit` and XML types."

src/library/scala/ValueOf.scala
alexeyr: Perhaps `L <: Int with Singleton`?

src/compiler/scala/tools/nsc/transform/Erasure.scala
milessabin: Happy to make this change. This eliminates a try/finally that looks slightly mysterious, but not accidental ... is that reasonable? The same pattern is also present in the immediately following case for `DefDef`.

spec/03-types.md
milessabin: @alexeyr agreed ... I'll clarify.

src/library/scala/ValueOf.scala
milessabin: @fthomas yes agreed. Do you think you could help come up with more/better examples?

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: The change you suggest here generates,

```
Apply(Select(This(TypeName("ValueOf")), termNames.CONSTRUCTOR), List(Literal(Constant(...))))
```

rather than,

```
Apply(Select(New(Select(Ident(scala), TypeName("ValueOf"))), termNames.CONSTRUCTOR), List(Literal(Constant(...))))
```

which is what the original generates. The new version fails, I think because the type argument isn't inferred correctly. I'll stick with what I had before for now.

src/library/scala/ValueOf.scala
milessabin: @alexeyr Yes, that would be better. I've updated the comment, but on reflection I think that it shows that we need a less artificial example, ie. one which makes non-trivial use of the singleton-type by, eg. indexing a type class with it.

jvican:
[SIP-23 has been reviewed in today's SIP meeting](https://plus.google.com/u/1/events/cj5bfohn6dnetj1o39dnteujjls). The assigned reviewer is @adriaanm, which will provide actionable feedback shortly.

milessabin:
@propensive apropos your observation about the different rendering of types in the REPL: the distinction is between inferred constant types (`String("foo")`) vs. explicit literal types (`"foo"`). I kept the representation of inferred constant types the same for backwards compatibility: very many neg tests would have different output if the representation changed. OTOH, where types have been provided explicitly, having a representation which is also syntactically valid under the new rules seems desirable.

alexeyr:
> Type parameter inference remains the same unless the inference of a singleton type is explicitly requested via a `<: Singleton` bound.

Does this apply to non-literal singleton types as well? It would be quite useful in some cases I've had. This change should also be documented in 6.26.4.

milessabin:
@alexeyr yes it does.

milessabin:
@retronym I've added a `-Yrangepos` test. I can't work out how to persuade the `-Xprint-pos` output to get sent to the corresponding check file ... is that possible?

retronym:
@milessabin I'm happy if you just run that manually and paste the output here for us to eyeball.

milessabin:
```
# starting 1 test in run
[[syntax trees at end of                    parser]] // sip23-rangepos.scala
[0:96]package [0:0]<empty> {
  [0:96]object Test extends [12:96][20:23]App {
    [12]def <init>() = [12]{
      [NoPosition][NoPosition][NoPosition]super.<init>();
      [12]()
    };
    [28:50]val foo: [37:42][37:42]"foo".type = [45:50]"foo";
    [53:73]val bar: [62:66][62:66]'bar.type = [69:73]'bar;
    [76:94]val baz: [85:88][85:88]-23.type = [92:94]-23
  }
}

ok 1 - run/sip23-rangepos.scala                

[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[success] Total time: 6 s, completed 12-Aug-2016 16:19:40
```

milessabin:
Squashed, rebased and `Co-Authored-By`'s added.

adriaanm:
I added a [comment](https://github.com/scala/scala.github.com/pull/346) to the SIP to summarize our discussion during last week's meeting.

spec/12-the-scala-standard-library.md
sjrd, line 644: @soc I disagree with your reasoning. The fact that `classOf[T]` is a builtin instead of taking an implicit `ClassTag[T]` is an artifact of history. Were we to define `classOf` and `ClassTag` _today_, we would make the implicit materialization of `ClassTag` the only builtin, and implement `classOf` in user-space as

``` scala
def classOf[T](implicit ct: ClassTag[T]): Class[T] = ct.runtimeClass
```

This would also have a nice benefit that we could simply write `classOf[T]` instead of `classTag[T].runtimeClass` when we have an implicit `ClassTag[T]` in scope.

Similarly, were `new Array[T](len)` be defined _today_, it would have an implicit `ClassTag[T]` as well.

Since we are defining `ValueOf` today, we can do it the right way, with the implicit materialization be the only builtin, and `valueOf[T]` being purely user-space.

I do not understand your argument of "bullet-proof". If I have a `v: ValueOf[T]` at my disposal, there's nothing wrong in passing it explicitly to `valueOf[T](v)`. The type of `v` already guarantees that `v` is the right `ValueOf[T]` to be passed to `valueOf[T]` anyway.

src/library/scala/ValueOf.scala
sjrd: The _consistent_ thing to do would be `ValueTag`, and have methods `valueTag` and `valueOf`:

``` scala
// consistent with ClassTag[T]
final class ValueTag[T](val value: T) extends AnyVal

// equivalent to implicitly[ValueTag[T]]
// consistent with classTag[T], which is equivalent to implicitly[ClassTag[T]]
def valueTag[T](implicit vt: ValueTag[T]): ValueTag[T] = vt

// equivalent to valueTag[T].value
// consistent with classOf[T], which is equivalent to classTag[T].runtimeClass
def valueOf[T](implicit vt: ValueTag[T]): T = vt.value
```

spec/12-the-scala-standard-library.md
nafg, line 644: Well.. isn't classOf considered a stable value? (Although it wasn't
originally.) It's pushing it when it takes type arguments, but if it took a
value argument, even implicitly, that would really be stretching it. (Just
being pedantic...)

sjrd, line 644: Not really a stable value, no. But a literal constant, indeed (it does have a `ConstantType`). It is true that this property makes `classOf` more magic than `valueOf` needs to be.

soc, line 644: @sjrd I think that's incorrect from multiple points of view:

> This would also have a nice benefit that we could simply write classOf[T] instead of classTag[T].runtimeClass when we have an implicit ClassTag[T] in scope.
- Adding `ClassTag[T]` to `classOf` would be wrong, as `classOf` requires a `ClassTag` only in some cases, it doesn't require a `ClassTag` when the argument is a constant type.
- Same for `Array[T]`.

I agree that `classTag[T].runtimeClass --> classOf[T]`  is desirable. In fact, I think I have an implementation of this lying around somewhere, which is only blocked by an unrelated scalac issue. Nothing in that requires adding `ClassTag` to `classOf`, and it would be wrong for the reasons mentioned.¹

This eliminates the argument for:

> Since we are defining ValueOf today, we can do it the right way, with the implicit materialization be the only builtin, and valueOf[T] being purely user-space.

The only question remaining now is whether we want to have `valueOf` consistent with `classOf` or doing its own thing. I prefer consistency.

> I do not understand your argument of "bullet-proof". If I have a v: ValueOf[T] at my disposal, there's nothing wrong in passing it explicitly to valueOf[T](v). The type of v already guarantees that v is the right ValueOf[T] to be passed to valueOf[T] anyway.

What's the use-case for that? The compiler is able to create `ValueOf`s for all valid cases out of thin air, I don't see any benefit where the user can pass his own `ValueOf` instance. Either it is identical to the one supplied by the compiler, or there is a serious issue– as the type can't guarantee that it's actually one of the "allowed" singleton types. That's why the `ValueOf` constructor needed to be private (see comment below). Otherwise a user can do 

```
class Horse
val horse = valueOf(new ValueOf(new Horse)) // or just pass null: Horse
```

which sabotages the whole idea of `valueOf`.

¹ In fact your biggest issue here is that Dmitry strongly disagrees that it's the right thing to do as `classOf[Int]` and `classTag[T].runtimeClass` have different performance characteristics.

soc, line 644: @nafg Mentions correctly that `classOf` needs to be a built in, as the type for concrete types needs to compile down to a `Constant`.

src/library/scala/ValueOf.scala
soc: I agree on @sjrd on this.
- `scala.reflect.ValueTag` would be consistent with the other tags.
- I'm not sure we even need or should add `scala.reflect.valueTag`, but if we do, the name is consistent with what we have.

spec/12-the-scala-standard-library.md
sjrd, line 644: > - Adding ClassTag[T] to classOf would be wrong, as classOf requires a ClassTag only in some cases, it doesn't require a ClassTag when the argument is a constant type.
> - Same for Array[T].

And in both those cases, an implicit `ClassTag[T]` can be materialized when `T` is a constant type. So demanding an `implicit ClassTag[T]` is just as correct.

> Nothing in that requires adding ClassTag to classOf

Sure, but then it requires `classOf` to be magic. You either need `classOf` to take an implicit `ClassTag` or to be magic. I prefer not-magic. Now, as per @nafg's comment, `classOf` needs to be magic for an unrelated reason. But `valueOf` doesn't need to.

> and it would be wrong for the reasons mentioned.

Nope, see my first answer.

> This eliminates the argument for

It doesn't. Less magic is still better.

> as the only question remaining now is whether we want to have valueOf consistent with classOf or doing its own thing. I prefer consistency.

There is the second question: more magic or less magic.

> That's why the ValueOf constructor needed to be private (see comment below).

I agree that the constructor of `ValueOf` should be private. This is indeed necessary to guarantee that an existing instance of `ValueOf[T]` is correct by construction. When that is guaranteed, there cannot be any harm done (whether involuntarily or maliciously) by passing an explicit `ValueOf[T]` to `valueOf[T]`. There is therefore no need for the compiler to take protection measures for `valueOf[T]`, so it need not be magic, so it can be defined in user-space with a perfectly normal `implicit ValueOf[T]`.

src/library/scala/ValueOf.scala
soc: > One thing I'd like to learn from the proponents of this: How much of the usage of ValueX do you imagine to be in the form of an implicit parameter, and how much to be as a context bound?

I'm asking this because I think `[T : ValueOf]` looks really really terrible in my opinion.

spec/12-the-scala-standard-library.md
soc, line 644: > Now, as per @nafg's comment, classOf needs to be magic for an unrelated reason. But valueOf doesn't need to.

It still needs to be consistent. If `classOf` doesn't take a `ClassTag` it doesn't need, `valueOf` should not either.

Regardless of the reasons `classOf` is like this, they are not changeable, so `valueOf` needs to adapt and be consistent with it, not the other way around.

sjrd, line 644: At this point, the argument boils down to a battle between consistency and magicless.
In my book, magicless wins. In yours, apparently consistency wins. We can never agree with each other, so other people need to take over this argument and eventually decide.

soc, line 644: In my book magicless also wins, except in the most important circumstances.
A method that will be available to every single developer by default, on every IDE auto-complete, is one of those important circumstances. It needs to consistent. It's the standard library, not some random third-party library people can opt out of.

After years of doing the deprecation/removal dance, `valueOf` with `ValueTag` is one of the things that would get fixed or removed in short order.

dwijnand, line 644: I think perhaps it would be appropriate to add a comment to `classOf`, explaining (or trying to explain) why it don't have the obvious implementation.

soc, line 644: I can have a look at my new `classOf[T] == classTag[T].runtimeClass` implementation, I'm pretty sure I added some comment there.

milessabin:
Backport to 2.11.8 available [here](https://github.com/typelevel/scala).

milessabin:
Test failure looks spurious to me ...

milessabin:
Rebased.

milessabin:
The failing tests here appear to be scaladoc related and not run locally via the default `partest` command ... how should they be run?

SethTisue:
`scripts/jobs/validate/test` runs `testAll` and in the definition of `testAll` in `build.sbt` we see`(testOnly in IntegrationTest in testP).toTask(" -- --srcpath scaladoc").result`. if you transmogrify that by rotating it through 11-dimensional space, it comes out as `test/it:testOnly -- --srcpath scaladoc`, voila

or you could take the shortcut and just say `partest --srcpath scaladoc`

needless to say, this could be all be made a great deal less obtuse. there is still a lot of ant and partest baggage here

milessabin:
Squashed.

milessabin:
Rebased.

milessabin:
Rebased.

milessabin:
Rebased, fixed https://github.com/typelevel/scala/issues/128 (with test).

adriaanm:
I hope to still get this in for 2.12.2, but we have a big backlog of PRs we need to get into 2.12.2

paulp:
I don't know if this is still planned for 2.12.2 but you will probably want to rework [this case](https://github.com/milessabin/scala/blob/c89e962d34e95598de9f2f93230c646a4f6d2d87/src/compiler/scala/tools/nsc/transform/Erasure.scala#L966) because the side-effecting unapply is called before the setting is checked to be enabled.

adriaanm:
Thanks! I'd buy you a tuna gnome, if only they were still in business. (I'm hoping to get to this for 2.12.3. EDIT: not the gnomes, the PR)

milessabin:
Minimised some more to,

```scala
object Test {
  class Covar[-A]

  def foo[A, CC[X] <: Option[X]]: Covar[CC[_ <: A]] = ???

  val bar: Covar[Option[Int]] = foo
}
```

The culprit is actually [here](https://github.com/milessabin/scala/blob/c89e962d34e95598de9f2f93230c646a4f6d2d87/src/reflect/scala/reflect/internal/Types.scala#L2861) ... the subtype test `<:< typeOf[Singleton]` is adding a bound of `Singleton` to the `TypeVar` corresponding to `A` preventing it from being instantiated to `Int`.

milessabin:
Fixed https://github.com/typelevel/scala/issues/146 and https://github.com/typelevel/scala/issues/140. Squashed and rebased.

Atry:
Will there be the `narrow` method in standard library?

milessabin:
@Atry shapeless has demonstrated that `narrow` is useful. I'd be open to adding it, but where? `Predef`?

Atry:
If there is an implicit conversion from `A` to `ValueOf[A]`, then `Predef.valueOf(a)` is enough

Atry:
`narrow` is useful because the rule to infer singleton type is too limited.

Is it possible to enable narrow type by default, at least in some context, in case of https://github.com/typelevel/scala/issues/151

spec/12-the-scala-standard-library.md
Atry: `T {}` is a little confusion because it does not respect `.widen`:

``` scala
@ val t = typeOf[0 {}] 
t: Type = RefinedType(List(FoldableConstantType(Constant(0))), SynchronizedOps())
@ val w = t.widen 
w: Type = RefinedType(List(FoldableConstantType(Constant(0))), SynchronizedOps())
@ println(t eq w) 
true
```

Atry: How about encode it as `Id[0]`, which is less magic.

``` scala
@ type Id[+A] = A 
defined type Id
@ val dontWidenMe: Id[0] = 0 
dontWidenMe: Id[0] = 0
@ Some(dontWidenMe) 
res72: Some[Id[0]] = Some(0)
```

tindzk:
Great work!

I migrated [Pine](https://github.com/sparsetech/pine) to literal types and I am very pleased with the syntactic expressiveness that allowed me to drastically reduce boilerplate code. Note that I came across some [corner-cases](https://github.com/typelevel/scala/issues/created_by/tindzk) which may not represent the expected behaviour.

For mainstream adoption, it is crucial to collaborate with the IDEA team to properly support [syntax highlighting](https://youtrack.jetbrains.com/issue/SCL-11193) and [implicit resolution](https://youtrack.jetbrains.com/issue/SCL-12080).

adriaanm:
Sorry this slipped again, had too much on my plate.

milessabin:
Rebased.

retronym:
As discussed in #6025, we'll have to revisit/retest weak subtyping as its implemementation assumes that numeric types can be identified either directly or with a `deconst`, but now we could have type aliases that expand to constant types that deconst to numeric types.

A test along these lines will help flush out the issues.
```scala
  val i1 = 1
  val l1 = 1L

  type I1 = i1.type
  type L1 = l1.type
```

```scala
assert(weakLub(typeOf[I1], typeOf[L1]) == typeOf[Long])
```

Atry:
I wonder if it is possible to add an evidence of `value.type =:= T` in `ValueOf`. The evidence will help make `ValueOf` more useful in situations described at https://github.com/milessabin/shapeless/issues/753

Atry:
Typelevel Scala seems not to support `valueOf[this.type]`:
https://scastie.scala-lang.org/Atry/5j8id6acR3KZauDMFJD9DA

Note that shapeless has the same limitation : https://github.com/milessabin/shapeless/issues/755

adriaanm:
Tentatively scheduling for 2.12.5 as the pending work items likely amount to more than a week's worth of work.

soronpo:
>Tentatively scheduling for 2.12.5 as the pending work items likely amount to more than a week's worth of work.

Ouch, and 2.12.5 is only due March 😟 . Thank god, flying spaghetti monster and Miles Sabin for TLS!

adriaanm:
FYI this work has been pending for a while: [my implementation](https://github.com/adriaanm/scala/tree/sip23) was 95% ready 2.5 years ago. I wish I had time myself to get this across the finish line, but I don't right now.

milessabin:
I'm going to retarget this at 2.13.x and backport to 2.12.x.

soronpo:
@milessabin If it's not urgent, maybe you should consider waiting until SIP33 #6142 is complete and merged, since both sips modify the parser.

milessabin:
Rebased against 2.13.x.

Updated in response to feedback on the implementation in Typelevel Scala.

Updated SIP here: http://docs.scala-lang.org/sips/42.type.html.

Atry:
@milessabin Will there be a backport of SIP-23 for 2.12?

milessabin:
@Atry for Typelevel Scala, yes. It's @adriaanm's call on whether it'll be merged in Lightbend Scala.

adriaanm:
My (default) position is that 2.12 is in bug fix mode and I’d rather not add new features to it retroactively.

lrytz:
Literal types in libraries leak to users that are not enabling `-Yliteral-types`:

```
$ cat A.scala
object A {
  def f(x: 1): 1 = x
}
$ qsc3 A.scala -Yliteral-types
$ qs3 -e 'A.f(23)'
/var/folders/c_/8cjdngcx4msf2f5gj334xxnc0000gn/T/scalacmd1751070577770488733.scala:1: error: type mismatch;
 found   : Int(23)
 required: Int(1)
A.f(23)
    ^
one error found
```

Opinions on that? Are we going to enable literal types by default in 2.13 anyway?

lrytz:
Corner case (already mentioned in [adriaan's PR](https://github.com/adriaanm/scala/pull/21)) - can / should we do something about this?

```
scala> val t: 1 = t
                  ^
       warning: value t does nothing other than call itself recursively
t: 1 = 0

scala> val t: 1 = if (t == 1) t else t  // no warning
t: 1 = 0

scala> val t: 1 = t + 0  // we get 1 here..
t: 1 = 1
```

johnynek:
In my opinion, the ideal is to enable by default (which of course means we don't know of any issues with doing so) for 2.13.

milessabin:
@lrytz I think those examples are orthogonal to literal types ... we can get the same behaviour from non-literal and non-singleton types (this is vanilla 2.12.4),

```scala
scala> val t: Int = t
<console>:11: warning: value t does nothing other than call itself recursively
       val t: Int = t
                    ^
t: Int = 0

scala> val t: Int = if (t == 1) t else t
t: Int = 0

scala> val t: Int = t + 0
t: Int = 0
```
```scala
scala> final val foo = "foo"
foo: String("foo") = foo

scala> val t: foo.type = t
<console>:13: warning: value t does nothing other than call itself recursively
       val t: foo.type = t
                         ^
t: String = null

scala> val t: foo.type = if (t == foo) t else t
t: String = null

scala> val t: foo.type = t + ""
t: String = null
```
That's not to say that the lack of warnings is unproblematic in any of these cases, but I think the fix needs to be more general.

lrytz:
@milessabin I think the situation is slightly worse for literal types. For a singleton type in 2.12, `null` is allowed (by the spec)

```
scala> val a = "hi"
a: String = hi

scala> var b: a.type = null
b: a.type = null

scala> b = a
b: a.type = hi

scala> b = null
b: a.type = null
```

Having a value of type `1` actually holding `0` seems wrong though.

```
scala> val a: 1 = a
                  ^
       warning: value a does nothing other than call itself recursively
a: 1 = 0

scala> println(a == 1) // hello, constant folding
true

scala> println(a)
0
```

lrytz:
Dotty seems to get it right:

```
[info] Running dotty.tools.repl.Main -classpath /Users/luc/scala/dotty/library/target/scala-2.12/dotty-library_2.12-0.5.0-bin-SNAPSHOT-nonbootstrapped.jar
scala> val x: Int = x
val x: Int = 0
scala> val y: 1 = y
val y: 1 = 1
```

Atry:
Behaviors are different between literal types and class types. Is it really right?

milessabin:
I'm not sure that Dotty is completely consistent here either. It certainly behaves the way that you're expecting wrt literal types which are subtypes of `AnyVal`, but it does the same for String literals, which makes them different from other singleton subtypes of `AnyRef`. Eg. with current Dotty,
```scala
scala> object Foo 
// defined object Foo

scala> val t: Foo.type = if (t == Foo) t else t 
val t: Foo.type = null

scala> val x: "foo" = if (x == "foo") x else x 
val x: foo = "foo"
```

I'm happy to change the behaviour here, but I think for now I'd like to wait and see what the feedback on the SIP is ... all of these are options and I could live with any of them.

spec/03-types.md
```diff
@@ -102,16 +103,46 @@ forms.
...
+A literal type `lit` is a special kind of singleton type which denotes the
+single literal value `lit`.  Thus, the type ascription `1: 1` gives the most
+precise type to the literal value `1`:  the literal type `1`.
+
+At run time, an expression `e` is considered to have literal type `lit` if `e ==
+lit`.  Concretely, the result of `e.isInstanceOf[lit]` and `e match { case _ :
+lit => }` is determined by evaluating `e == lit`.
+
+Literal types are available for all types for which there is dedicated syntax
+except `Unit`. This includes the numeric types (other than `Byte` and `Short`
```
lrytz, line 133: Not that I have a use case for it, but why is `Unit` excluded?

spec/03-types.md
```diff
@@ -102,16 +103,46 @@ forms.
...
+which don't currently have syntax), `Boolean`, `Char`, `String` and `Symbol`.
+
+<!-- TODO: use eq when we lift it up to Any -->
```
lrytz: I don't follow here, can you make the comment more explicit (what is `it`)?

spec/03-types.md
```diff
@@ -102,16 +103,46 @@ forms.
...
+single literal value `lit`.  Thus, the type ascription `1: 1` gives the most
+precise type to the literal value `1`:  the literal type `1`.
+
+At run time, an expression `e` is considered to have literal type `lit` if `e ==
+lit`.  Concretely, the result of `e.isInstanceOf[lit]` and `e match { case _ :
+lit => }` is determined by evaluating `e == lit`.
+
+Literal types are available for all types for which there is dedicated syntax
+except `Unit`. This includes the numeric types (other than `Byte` and `Short`
+which don't currently have syntax), `Boolean`, `Char`, `String` and `Symbol`.
```
lrytz, line 134: How about `null`? The PR implementation supports it

```
scala> val x: null = null
x: null = null
```

Dotty doesn't

spec/06-expressions.md
```diff
@@ -1570,6 +1570,8 @@ means finding a substitution $\sigma$ of types $T_i$ for the type
 parameters $a_i$ such that
 
 - None of the inferred types $T_i$ is a [singleton type](03-types.html#singleton-types)
+  unless it is a singleton type corresponding to an object or a constant value
+  definition.
```
lrytz: I don't follow your addition here. Could you clarify? It seems the expected type can cause a singleton to be inferred.
```
scala> def f[T]: List[T] = null
f: [T]=> List[T]

scala> f : List[Object]
res9: List[Object] = null

scala> f : List["hai"]
res10: List["hai"] = null

scala> f : List[Predef.type]
res12: List[scala.Predef.type] = null
```

Also, I don't see the special role of a `<: Singleton` bound being mentioned in the spec diff.

src/compiler/scala/reflect/macros/contexts/Evals.scala
```diff
@@ -8,7 +8,7 @@ trait Evals {
   self: Context =>
 
   private lazy val evalMirror = ru.runtimeMirror(universe.analyzer.defaultMacroClassloader)
-  private lazy val evalToolBox = evalMirror.mkToolBox()
+  private lazy val evalToolBox = evalMirror.mkToolBox(options = if(global.settings.YliteralTypes) "-Yliteral-types" else "")
```
lrytz: It's not obvious to me why we need to forward this specific flag but not any other compiler / language flags..

src/compiler/scala/reflect/reify/phases/Reify.scala
```diff
@@ -51,6 +51,8 @@ trait Reify extends GenSymbols
     case mods: global.Modifiers   => reifyModifiers(mods)
     case xs: List[_]              => reifyList(xs)
     case s: String                => Literal(Constant(s))
+    case s: scala.Symbol if settings.YliteralTypes
+                                  => Literal(Constant(s))
```
lrytz: Could we do this unconditionally?

src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
```diff
@@ -363,6 +363,14 @@ abstract class CoreBTypesFromSymbols[G <: Global] extends CoreBTypes {
     })
   }
 
+  def SSymbol_apply: Symbol = _SSymbol_apply.get
+  private[this] lazy val _SSymbol_apply: LazyVar[Symbol] = runLazy {
+    getMember(SymbolModule, nme.apply) filter (sym => sym.info.paramTypes match {
+      case List(pt) => pt.typeSymbol == StringClass
+      case _        => false
+    })
+  }
```
lrytz: I don't think you need to do the `filter` here, `Symbol.apply` is not overloaded. `getMember(SymbolModule, nme.apply)` should be enough.

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -1839,24 +1839,47 @@ trait Types
   class PackageClassInfoType(decls: Scope, clazz: Symbol)
   extends ClassInfoType(List(), decls, clazz)
 
-  /** A class representing a constant type.
+  /** A class representing a constant type. Constant types are constant-folded during type checking.
+   *  To avoid constant folding, use the type returned by `deconst` instead.
```
lrytz: Could you expand the documentation here and explain why we need the distinciton between `FoldableConstantType` and `LiteralType`? Which of the two is used when / where?

lrytz (commented):
I'm only mid-way through the first (of the many) commits, but here are a few comments I have so far. I also ran the last commit of this PR through our benchmark, there seems to be a slight slowdown. Maybe the `settings.YliteralTypes` check in the parser (`def simpleType`) is expensive?

milessabin:
Rebased.

milessabin:
Combined the two `Any`/`Nothing` commits to eliminate an earlier failed commit.

milessabin:
The next commit unconditionally enables `-Yliteral-types`. I've updated the breakdown of test result changes in the description at the top of this PR.

milessabin:
Nb. the previous commit is expected to cause a few tests to fail.

milessabin:
Now with test/check files updated to accommodate the new behaviour.

spec/03-types.md
milessabin: `().type` would be equivalent to `Unit` so there doesn't seem to be any need for it.

spec/03-types.md
milessabin: That's a bug ... I'll fix it.

spec/03-types.md
milessabin: I'll tweak the wording.

spec/06-expressions.md
milessabin: In your examples those are all explicit types ... the text relates to inferred types. See the differences in these three cases,
```
scala> val Foo = "foo"
Foo: String = foo

scala> List(Foo)
res0: List[String] = List(foo)

scala> object Foo
defined object Foo

scala> List(Foo)
res1: List[Foo.type] = List(Foo$@286866cb)

scala> final val foo = 23
foo: Int(23) = 23
```
Singleton types are only inferred for objects and for final vals.

I thought I had added some text about `<: Singleton` ... I'll check.

src/compiler/scala/reflect/macros/contexts/Evals.scala
milessabin: If we make this unconditional then that can go away.

src/compiler/scala/reflect/reify/phases/Reify.scala
milessabin: Yes :-)

src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
milessabin: I'll take a look.

src/reflect/scala/reflect/internal/Types.scala
milessabin: Will do.

milessabin:
Bootstrap failure due to pickling ...

milessabin:
CLA bot wedged?

src/compiler/scala/reflect/reify/phases/Reify.scala
milessabin: Qualifying that a bit: yes, we can make this unconditional if the whole feature is unconditional, otherwise not.

spec/03-types.md
milessabin: Looking at this now ... those TODOs are very stale and I think can be deleted. The one you're flagging up here relates to the specification being updated to use `eq` rather than `==` in the definition of `isInstanceOf` for literal types _when_ `eq` is defined for values of type `Any`. I don't think this is on the table at the moment. The following comment about `Byte` relates to a hypothetical future SIP which adds `Byte` and `Short` literals ... this is mentioned in SIP 23 in the section "Follow on work from this SIP" and I don't think needs to be mentioned as a comment in the spec source. The final comment relates to inferred constant types and that's covered in the local type inference section of the spec.

milessabin:
@lrytz your comments addressed ...

milessabin:
I don't understand the failure indications above ... 4ca2cb is an abort, 65d121 seems fine, and the combined failure refers the two of them. Will a rebuild sort this out?

milessabin:
Rebasing in the hope of getting rid of the (spurious?) error indications.

spec/03-types.md
lrytz: Conformance to `Singleton` could be an argument to allow `()` as a type

```
def f[T <: Singleton](x: T): T = x
f(1); f("") // ok
f(())
       error: type mismatch;
        found   : Unit
        required: Singleton
```

But I don't have a strong opinion, it's probably not useful.

spec/06-expressions.md
```diff
@@ -1607,6 +1609,8 @@ finding a substitution $\sigma$ of types $T_i$ for the type parameters
 $a_i$ such that
 
 - None of the inferred types $T_i$ is a [singleton type](03-types.html#singleton-types)
+  unless it is a singleton type corresponding to an object or a constant value
+  definition or the corresponding bound $U_i$ is a subtype of `scala.Singleton`.
```
lrytz, line 1613: Thanks adding a mention of `<: Singleton`. I think I misunderstood the text on my previous comment here; anyway, in my current interpretation, it sounds like a singleton type should always be inferred if it denotes an object/constant value.

Where does the spec say that `List[Int]` is inferred for the example below, not `List[1]`?

```
scala> def f[T](x: T): List[T] = Nil
f: [T](x: T)List[T]

scala> object t { final val x = f(1) }
defined object t

scala> t.x: List[1]
         ^
       error: type mismatch;
        found   : List[Int]
        required: List[1]
```

src/compiler/scala/reflect/macros/contexts/Evals.scala
lrytz: OK that's good, but I still don't get why this was needed in this case, but not for other flags..

spec/03-types.md
milessabin: Are you suggesting that `().type` should be equivalent to `Unit with Singleton`? Or that `Unit` should be a subtype of `Singleton`. Both of those make me a little uncomfortable. Either way, the current behaviour is consistent with the latest Dotty.

spec/06-expressions.md
milessabin: More or less exactly in the area you're commenting on. The right hand side of `val x` is `f(1)` and the type parameter `T` of `f` at that call site will be inferred as the widened type ("None of the inferred types Ti is a singleton type unless ..."). If `x` had been given an explicit `: 1` annotation that would have driven the inference of `T`.

lrytz: Thanks, I'm getting there :) Shouldn't this type check then:

```
scala> :pa

def f[T](x: T): List[T] = Nil
object o
object a { final val x = o; final val y = 1 }
object t { final val x = f(a.x); final val y = f(a.y) }

scala> a.x : o.type
res0: o.type = o$@b46e103

scala> a.y: 1
res2: Int = 1

scala> t.x : List[o.type]
res4: List[o.type] = List()

scala> t.y : List[1]
         ^
       error: type mismatch;
        found   : List[Int]
        required: List[1]
```

spec/03-types.md
lrytz: We don't need to do anything in this PR, maybe add a test for the status quo and refernece the discussion. You could write down why `() : Unit with Singleton` makes you uncomfortable, which would be helpful if the question comes up again in the future.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1259,6 +1299,16 @@ abstract class Erasure extends InfoTransform
             case ArrayValue(elemtpt, trees) =>
               treeCopy.ArrayValue(
                 tree1, elemtpt setType specialScalaErasure.applyInArray(elemtpt.tpe), trees map transform).clearType()
+            case ValDef(_, _, tpt, rhs) =>
```
lrytz, line 1302: a one-line comment saying which case is handled here would be welcome

spec/03-types.md
Atry: Considering implicit resolution issue for types related to `HNil.type`. I vote 👍  for `Unit =:= ().type`.

lrytz (commented):
I'm almost through, the remaining things for me to look at are Namers.scala (which is quite important) and read through the tests. The code looks great, and I really like the new feature!

Some other comments / questions.

1. There's the trick with the empty refinement to prevent widening:

```
    scala> def f[T <: Singleton](x: T): T {} = x
    scala> List(f(1))
    res11: List[1] = List(1)


    scala> def f[T <: Singleton](x: T): T = x
    scala> List(f(1))
    res12: List[Int] = List(1)
```

IUC there's nothing new in this PR, but it gets more exposure (in particular, it's in the signature of `def valueOf`), correct? I see this PR adds some regression tests for the (accidental?) feature, which is great. It should probably be documented somewhere - in the spec? Does it work in dotty?

2. It bothers me that a cyclic val with a literal type can get a value outside its type

```
scala> val t: 1 = t
                  ^
       warning: value t does nothing other than call itself recursively
t: 1 = 0
```

(`val t: 1 = if (t == 1) t else t` for a version that doesn't trigger the warning). This can be addressed in a follow-up PR, but I think it should. We discussed this before and left the question open (https://github.com/scala/scala/pull/5310#issuecomment-346876424). Does the SIP say anything on this topic?

spec/06-expressions.md
milessabin: No, in `f(a.y)` the inferred type for `T` is widened to `Int` because `T` doesn't have a `<: Singleton` bound.

milessabin:
Dotty does this,
```scala
scala> val t: 1 = t 
val t: 1 = 1
```
But it also does this,
```scala
scala> object Foo 
// defined object Foo
scala> val f: Foo.type = f 
val f: Foo.type = null
```
I think the behaviour of this PR is consistent and Dotty's isn't here. But on the other hand I think there's a good case for initializing with the singleton value if that's what the type ascription is asking for. I really do think that's orthogonal to this SIP/PR though. /cc @odersky @smarter

milessabin:
@lrytz on the empty refinement trick. Actually I don't want to add it to the spec, and no, it doesn't work in Dotty and I'd be surprised if I was able to persuade Martin to make it work.

There are tests which use it as part of this PR because it's used in the wild in contexts which are directly affected by the addition of literal types, so I wanted to make sure that the two combined did the right thing. But it's not used or supported in any essential way in the literal types implementation itself.

There's a replacement for it which works in Scala (and should also work in Dotty) which is to wrap results in an `Id` type. I've added `test/files/pos/sip23-narrow-no-empty-refinements.scala` which is mostly the same as `test/files/pos/sip23-narrow.scala` except the it uses `Id[T]` in place of `T {}`.

In fact this highlights one place in the original test where the trick doesn't work as intended ... the empty refinement interferes with type class instance selection in one case because the compiler doesn't treat `23 {}` as equivalent to `23`. Conversely `Id[23]` does the right thing in the same circumstances.

Atry:
@milessabin Another reason why use `Id` instead of empty refinement type is, `narrow` and `widen` should be inverse operator to each other.

https://github.com/scala/scala/pull/5310#discussion_r118456765

Atry:
```
Welcome to the Ammonite Repl 1.0.3
(Scala 2.12.4 Java 1.8.0_152)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import shapeless._  
import shapeless._ 

@ type Id[+A] = A  
defined type Id

@ type Zero = Witness.`0`.T  
defined type Zero

@ Widen[Zero] 
res3: Widen[0]{type Out = Int} = shapeless.Widen$$anon$4@36120a8b

@ Widen[Zero {}]  
cmd4.sc:1: could not find implicit value for parameter widen: shapeless.Widen[ammonite.$sess.cmd2.Zero]
val res4 = Widen[Zero {}] 
                ^
Compilation Failed

@ Widen[Id[Zero]]  
res4: Widen[0]{type Out = Int} = shapeless.Widen$$anon$4@491f3fb0

```
See, `Id` works better with `shapeless.Widen`.

spec/06-expressions.md
lrytz: OK, I get it, the compiler knows `implicitly[a.x.type =:= o.type]`, so widening to `o.type` doesn't discard information (or, `scala> o : a.x.type` type checks).

lrytz:
Thanks for the additional tests and elaborating on the empty refinement trick.

On the cyclic initialization question, I see what you mean by consistency, but also consider the examples in my earlier comment https://github.com/scala/scala/pull/5310#issuecomment-346898712. Deciding on this later is fine with me.

lrytz (commented):
Tiny cleanup: remove `test/files/neg/t6263.check` and `test/files/neg/t900.check`.

LGTM otherwise, really nice work, great cleanups on the side, I also like the extensive tests. Thanks a lot!

As discussed, we should run the community build before merging - do you need help with that?

milessabin:
I'll kick off a community build as soon as this goes green.

dwijnand:
JFYI this is green now @milessabin.

milessabin:
Community build here: https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/889/.

Although it looks like the community build hasn't be green for a while ... or am I misreading the Jenkins dashboard?

SethTisue:
yeah the 2.13 community build is kind of a shambles right now. a bunch of projects are green, but many aren't. so it can give some assurance, but far less assurance, currently, than the 2.12 community build gives.

I haven't prioritized it because as soon as the new collections land in 2.13, I expect everything to break anyway. the laborious process of getting everything green again will begin in earnest at that point.

lrytz:
compared to the previous run, there seem to be 3 additional failures. Theoretically they could be due to changes in the respective repos. To investigate.

scala-continuations: `FAILED (RuntimeException: )`

```
[scala-continuations] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.9/project-builds/scala-continuations-c9ed3b11fe80d175b140dc7c083132fd534a1de8/library/src/test/scala/scala/tools/selectivecps/TestSuite.scala:215: cannot compute type for CPS-transformed function result
[scala-continuations] [error]     test { shift { k: (Int => String) => 9 }; { test3(0); 2 } }
[scala-continuations] [error]                  ^
```



scala-parallel-collections: `FAILED (RuntimeException: Tests unsuccessful)`

```
[scala-parallel-collections] [error] Test scala.SerializationStabilityTest.testAll failed: java.io.InvalidClassException: scala.collection.immutable.HashMap$SerializationProxy; local class incompatible: stream classdesc serialVersionUID = 2, local class serialVersionUID = 3693899892909726219, took 0.023 sec
```

https://github.com/scala/scala-parallel-collections/blob/master/junit/src/test/scala/scala/SerializationStabilityTest.scala#L91




slick: `FAILED (RuntimeException: )`

```
[slick] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.9/project-builds/slick-95bd2107fe1d80c30d5fba7d421dc31395b01f96/slick/src/main/scala/slick/relational/RelationalProfile.scala:82: pattern type is incompatible with expected type;
[slick] [error]  found   : slick.relational.ProductResultConverter[_(in value $anonfun),(some other)_(in value $anonfun)]
[slick] [error]  required: slick.relational.ResultConverter[slick.relational.ResultConverterDomain,Any]
[slick] [error] Note: _ <: slick.relational.ResultConverterDomain, but trait ResultConverter is invariant in type M.
[slick] [error] You may wish to define M as +M instead. (SLS 4.5)
[slick] [error] Note: _ <: Any, but trait ResultConverter is invariant in type T.
[slick] [error] You may wish to define T as +T instead. (SLS 4.5)
[slick] [error]       case tm @ TypeMappingResultConverter(_: ProductResultConverter[_, _], _, _) =>
[slick] [error]                                               ^
```

milessabin:
The scala-continuations one is probably due to the introduction of the new subtypes of `ConstantType` ... should be an easy fix. The parallel collections problem looks unrelated. I'm guessing the Slick failure is a type inference change, probably also fairly easily fixed.

soronpo:
I don't know how difficult is the backport to 2.12, but if you do it for TLS anyways, then you can use it as dummy PR to 2.12, just to test the community build.

Atry:
There is a comment in https://github.com/milessabin/scala/blob/352a0c8c4eccbca1c0c98de784c2ed87cc5b7fe9/src/library/scala/Predef.scala#L203

> @inline def implicitly[T](implicit e: T) = e    // for summoning implicit values from the nether world -- TODO: when dependent method types are on by default, give this result type `e.type`, so that inliner has better chance of knowing which method to inline in calls like `implicitly[MatchingStrategy[Option]].zero`

I guess it's the time now to make `implicitly` have `shapeless.the` semantics.

adriaanm:
We’re not accepting new features in 2.12 anymore, even under a flag. Sorry,
we have to keep this train moving forward. Any maintenance burden we take
on for old versions is less time to spend on 2.13 and 2.14 (soon!)

Atry:
@adriaanm I meant 2.13 https://github.com/milessabin/scala/pull/6

soronpo:
> We’re not accepting new features in 2.12 anymore, even under a flag. 

Was this in reply to me @adriaanm ? I was just suggesting how to test for community-build regressions, since the 2.13 build is currently failing.

lrytz:
You could also use the [`freeze`](https://github.com/scala/community-builds/blob/2.13.x/freeze) script pick the last green run of the 2.13.x community build and then [run it locally](https://github.com/scala/community-builds/wiki/Local-runs) / https://github.com/scala/community-builds/issues/58 / https://github.com/scala/community-builds/issues/408 / https://github.com/scala/community-builds/issues/667.

milessabin:
@lrytz I'm not quite sure what you're proposing ... do you mean rebase this PR onto the commit corresponding to the last green community build and then run against that?

TBQH, I think it would probably be quicker and easier to checkout and try building the three newly failing projects.

lrytz:
Ah, my assumption is that the community build is not broken due to 2.13.x being broken, but because the community projects moved ahead with incompatible changes - that's the more common case anyway. So if you take the last green run and freeze all community projects at the SHAs they had at this point, the build will probably also pass with current 2.13.x (or the baseline of this PR).

lrytz:
Instead of running locally, you can push the frozen SHAs to a branch of your scala/community-builds fork, and then specify that when starting the community build (https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/build?delay=0sec)

SethTisue:
> Ah, my assumption is that the community build is not broken due to 2.13.x being broken, but because the community projects moved ahead with incompatible changes

My understanding is that the _main_ reason the 2.13 community build is currently broken is that we had inadvertently released 2.13.0-M1 and 2.13.0-M2 with `-Xsource:2.12` enabled by default. But we've made incompatible language changes since 2.12, so when that switch was flipped (by #6092), a number of projects broke. iirc it had especially (but perhaps not only) to do with eta expansion.

There are two paths forward for fixing it:

* attempt the fixes ourselves
* go ahead and release M3 anyway and let the project maintainers change their code themselves

we sometimes go the fix-it-ourselves route, but in this case several of the broken projects are pretty big/complicated (e.g. scalaz and shapeless), so I think it might be better to just cross our fingers and ship M3 without strong assurance from the community build.

jasmeet107 (approved):
LGTM 👍

adriaanm:
Let's deal with the community build in the next weeks. As Seth said, this isn't the first PR to affect the community build in 2.13.x.

adriaanm:
🎉

soronpo:
Wow! Congratulation @milessabin for threading this needle!

johnynek:
Woo! Thanks @milessabin and all reviewers and contributors to this effort.

milessabin:
Thanks for helping me getting this across the line @adriaanm!

PR #5312, opened 2016-07-29
SI-8434 Make generic Set operations build the same kind of Set

szeiger:
When building from a `Set` implementation that was statically seen as
a `collection.GenSet` or `collection.Set`, we used to build a default
`Set` implementation determined by `GenSetFactory.setCanBuildFrom`.
This change modifies `setCanBuildFrom` to determine the correct
implementation at runtime by asking the source `Set`’s companion object
for the `Builder`.

Tests are in `NewBuilderTest.mapPreservesCollectionType`, including lots
of disabled tests for which I believe there is no solution under the
current collection library design.

`Map` suffers from the same problem as `Set`. This _can_ be fixed in the
same way as for `Set` with some non-trivial changes (see the note in
`NewBuilderTest`), so it is probably best left for Scala 2.13.

JIRA: https://issues.scala-lang.org/browse/SI-8434


SethTisue:
The ticket number in the commit message is wrong (should be 8434 not 8432).

SethTisue:
LGTM once the commit message is fixed

PR #5314, opened 2016-07-31
SI-9813 Add Java 1.6. and 1.8 java.lang.Math methods into scala.math (reopened)

golem131:
reopen of https://github.com/scala/scala/pull/5271

### DIFF
--- a/src/library/scala/math/package.scala
+++ b/src/library/scala/math/package.scala
@@ -63,6 +63,9 @@ package scala
   *
   * @groupname randomisation Pseudo Random Number Generation
   * @groupprio randomisation 130
+  *
+  * @groupname exact Functions to avoid overflow on primitive types
+  * @groupprio exact 140
   */
 package object math {
   /** The `Double` value that is closer than any other to `e`, the base of
@@ -215,9 +218,37 @@ package object math {
     */
   def signum(x: Long): Long     = java.lang.Long.signum(x)
   /** @group signum */
-  def signum(x: Float): Float   = java.lang.Math.signum(x)
+  def signum(f: Float): Float   = java.lang.Math.signum(f)
   /** @group signum */
-  def signum(x: Double): Double = java.lang.Math.signum(x)
+  def signum(d: Double): Double = java.lang.Math.signum(d)
+
+  def floorDiv(x: Int, y: Int): Int = java.lang.Math.floorDiv(x, y)
+
+  def floorDiv(x: Long, y: Long): Long = java.lang.Math.floorDiv(x, y)
+
+  def floorMod(x: Int, y: Int): Int = java.lang.Math.floorMod(x, y)
+
+  def floorMod(x: Long, y: Long): Long = java.lang.Math.floorMod(x, y)
+
+  def copySign(magnitude: Double, sign: Double): Double = java.lang.Math.copySign(magnitude, sign)
+
+  def copySign(magnitude: Float, sign: Float): Float = java.lang.Math.copySign(magnitude, sign)
+
+  def nextAfter(start: Double, direction: Double): Double = java.lang.Math.nextAfter(start, direction)
+
+  def nextAfter(start: Float, direction: Double): Float = java.lang.Math.nextAfter(start, direction)
+
+  def nextUp(d: Double): Double = java.lang.Math.nextUp(d)
+
+  def nextUp(f: Float): Float = java.lang.Math.nextUp(f)
+
+  def nextDown(d: Double): Double = java.lang.Math.nextDown(d)
+
+  def nextDown(f: Float): Float = java.lang.Math.nextDown(f)
+
+  def scalb(d: Double, scaleFactor: Int): Double = java.lang.Math.scalb(d, scaleFactor)
+
+  def scalb(f: Float, scaleFactor: Int): Float = java.lang.Math.scalb(f, scaleFactor)
 
   // -----------------------------------------------------------------------
   // root functions
@@ -267,6 +298,16 @@ package object math {
     */
   def expm1(x: Double): Double = java.lang.Math.expm1(x)
 
+  /**
+    *  @group explog
+    */
+  def getExponent(f: Float): Int = java.lang.Math.getExponent(f)
+
+  /**
+    *  @group explog
+    */
+  def getExponent(d: Double): Int = java.lang.Math.getExponent(d)
+
   // -----------------------------------------------------------------------
   // logarithmic functions
   // -----------------------------------------------------------------------
@@ -315,13 +356,83 @@ package object math {
   /** Returns the size of an ulp of the given `Double` value.
     * @group ulp
     */
-  def ulp(x: Double): Double = java.lang.Math.ulp(x)
+  def ulp(d: Double): Double = java.lang.Math.ulp(d)
 
   /** Returns the size of an ulp of the given `Float` value.
     * @group ulp
     */
-  def ulp(x: Float): Float = java.lang.Math.ulp(x)
+  def ulp(f: Float): Float = java.lang.Math.ulp(f)
 
-  /** @group rounding */
+  /** @group exact */
   def IEEEremainder(x: Double, y: Double): Double = java.lang.Math.IEEEremainder(x, y)
+
+  // -----------------------------------------------------------------------
+  // exact functions
+  // -----------------------------------------------------------------------
+
+  /**
+    * @group exact
+    */
+  def addExact(x: Int, y: Int): Int = java.lang.Math.addExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def addExact(x: Long, y: Long): Long = java.lang.Math.addExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def subtractExact(x: Int, y: Int): Int = java.lang.Math.subtractExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def subtractExact(x: Long, y: Long): Long = java.lang.Math.subtractExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def multiplyExact(x: Int, y: Int): Int = java.lang.Math.multiplyExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def multiplyExact(x: Long, y: Long): Long = java.lang.Math.multiplyExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def incrementExact(a: Int): Int = java.lang.Math.incrementExact(a)
+
+  /**
+    * @group exact
+    */
+  def incrementExact(a: Long) =  java.lang.Math.incrementExact(a)
+
+  /**
+    * @group exact
+    */
+  def decrementExact(a: Int) =  java.lang.Math.decrementExact(a)
+
+  /**
+    * @group exact
+    */
+  def decrementExact(a: Long) =  java.lang.Math.decrementExact(a)
+
+  /**
+    * @group exact
+    */
+  def negateExact(a: Int) =  java.lang.Math.negateExact(a)
+
+  /**
+    * @group exact
+    */
+  def negateExact(a: Long) =  java.lang.Math.negateExact(a)
+
+  /**
+    * @group exact
+    */
+  def toIntExact(value: Long): Int = java.lang.Math.toIntExact(value)
+
 }


src/library/scala/math/package.scala
```diff
@@ -324,4 +368,74 @@ package object math {
 
   /** @group rounding */
   def IEEEremainder(x: Double, y: Double): Double = java.lang.Math.IEEEremainder(x, y)
+
```
janekdb, line 368: exact group LGTM

src/library/scala/math/package.scala
```diff
@@ -219,6 +222,37 @@ package object math {
   /** @group signum */
   def signum(x: Double): Double = java.lang.Math.signum(x)
 
+
+  def floorDiv(x: Int, y: Int): Int = java.lang.Math.floorDiv(x, y)
+
+
```
janekdb: Fix spacing

src/library/scala/math/package.scala
```diff
@@ -219,6 +222,37 @@ package object math {
...
+  def floorMod(x: Long, y: Long): Long = java.lang.Math.floorMod(x, y)
+
+  def copySign(magnitude: Double, sign: Double): Double = java.lang.Math.copySign(magnitude, sign)
+
+  def copySign(magnitude: Float, sign: Float): Float = java.lang.Math.copySign(magnitude, sign)
+
+  def nextAfter(start: Double, direction: Double): Double = java.lang.Math.nextAfter(start, direction)
+
+  def nextAfter(start: Float, direction: Float): Float = java.lang.Math.nextAfter(start, direction)
```
janekdb: Why is the signature different to the Java version?

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#nextAfter-float-double-

The Java version is `nextAfter(float start, double direction)` i.e. a deviation from the float/float or double/double pattern seen elsewhere. Find out why Java is different in this case?

src/library/scala/math/package.scala
```diff
@@ -219,6 +222,37 @@ package object math {
...
+
+  def nextUp(d: Double): Double = java.lang.Math.nextUp(d)
+
+  def nextUp(d: Float): Float = java.lang.Math.nextUp(d)
+
+  def nextDown(d: Double): Double = java.lang.Math.nextUp(d)
```
janekdb: `nextDown` on RHS.

src/library/scala/math/package.scala
```diff
@@ -219,6 +222,37 @@ package object math {
...
+
+  def nextDown(d: Float): Float = java.lang.Math.nextUp(d)
```
janekdb: `nextDown` on RHS.

src/library/scala/math/package.scala
```diff
@@ -219,6 +222,37 @@ package object math {
...
+
+  def scalb(d: Double, scaleFactor: Int): Double = java.lang.Math.scalb(d, scaleFactor)
+
+
+  def scalb(d: Float, scaleFactor: Int): Float = java.lang.Math.scalb(d, scaleFactor)
```
janekdb: Follow the Java convention of using `f` for float parameters here and in other methods.

janekdb:
Some comments added. I have not yet verified that all methods that could have been added have been added.

What's the plan with the method grouping? Add now or add into subsequent commit? Now is better?

golem131:
@janekdb java.lang.Math have 73 methods and scala.math in this PR have 73 methods binded to  java.lang.Math
About grouping - what group you recommend for floorDiv/floorMod and nextUp/nextAfter?

src/library/scala/math/package.scala
```diff
@@ -315,13 +356,83 @@ package object math {
   /** Returns the size of an ulp of the given `Double` value.
     * @group ulp
     */
-  def ulp(x: Double): Double = java.lang.Math.ulp(x)
+  def ulp(d: Double): Double = java.lang.Math.ulp(d)
```
janekdb, line 359: While `d` and `f` is now consistent with the naming convention used in the added methods which is based on the names used in Java I don't see this as an improvement because although this acts as a mnemonic to the parameter type it does not remind the user of the role of the parameter. Let the IDE and compiler remind us of the type; Use the name to describe the parameter along a different axis.

Although it's tricky to pick a name based on the role of the parameter here I would suggest `middle` based on the description of `ulp` in the Java docs. Even just reverting to `x` is better than switching to `d` and `f`.

WDYT?

janekdb, line 359: In fact this change has to be reverted or extended with a deprecated parameter name to be binary compatible. To maintain the focus of this PR to the addition of missing methods defer this renaming to a separate PR.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
...
+
+  def copySign(magnitude: Double, sign: Double): Double = java.lang.Math.copySign(magnitude, sign)
+
+  def copySign(magnitude: Float, sign: Float): Float = java.lang.Math.copySign(magnitude, sign)
+
+  def nextAfter(start: Double, direction: Double): Double = java.lang.Math.nextAfter(start, direction)
+
+  def nextAfter(start: Float, direction: Double): Float = java.lang.Math.nextAfter(start, direction)
+
+  def nextUp(d: Double): Double = java.lang.Math.nextUp(d)
```
janekdb, line 241: I'd like to see if we can take this opportunity to add some value to the parameter name. As mentioned below `d` and `f` provide no additional information that the typing doesn't already provide. Can we have more meaningful names?

Happy to make some suggestions later.

src/library/scala/math/package.scala
golem131, line 359: I agree with you, and at first in Java docs you can see - https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#ulp-double-

ulp(double d)/ulp(float f)

and second, i think, if we use convention, that we base naming on Java docs/Java source code, we need to do this in all places in file, not just in new bindings.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
     */
   def signum(x: Long): Long     = java.lang.Long.signum(x)
   /** @group signum */
-  def signum(x: Float): Float   = java.lang.Math.signum(x)
+  def signum(f: Float): Float   = java.lang.Math.signum(f)
```
janekdb, line 221: The parameter name cannot be changed with a deprecation annotation. This is out of scope for this PR don't you think?

Revert to `x`.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
...
   /** @group signum */
-  def signum(x: Double): Double = java.lang.Math.signum(x)
+  def signum(d: Double): Double = java.lang.Math.signum(d)
```
janekdb, line 223: Would require deprecation annotation.

Revert to `x`.

janekdb:
Confirmed all methods present in the Java 8 class are present here. :+1: 

Changes to debate further,
1. Follow the established pattern of using `x` instead of `d` and `f` for parameter names

src/library/scala/math/package.scala
janekdb, line 241: I think `x` is good enough.

src/library/scala/math/package.scala
janekdb, line 359: So I think we are in agreement: Use `x`.

src/library/scala/math/package.scala
```diff
@@ -315,13 +356,83 @@ package object math {
...
+
+  /**
+    * @group exact
+    */
+  def multiplyExact(x: Long, y: Long): Long = java.lang.Math.multiplyExact(x, y)
+
+  /**
+    * @group exact
+    */
+  def incrementExact(a: Int): Int = java.lang.Math.incrementExact(a)
```
janekdb, line 406: `a` is carried over from the Java 8 names. Maybe the Java engineers are working to some hidden convention but I can't figure it out. For example in Java

```
log(double a)
log10(double a)
log1p(double x)
toIntExact(long value)
```

Possibly there are some conventions from the world of floating point arithmetic at play here.

Overall I would just stick with `x` unless there is reason not too.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
     */
   def signum(x: Long): Long     = java.lang.Long.signum(x)
   /** @group signum */
-  def signum(x: Float): Float   = java.lang.Math.signum(x)
+  def signum(f: Float): Float   = java.lang.Math.signum(f)
   /** @group signum */
-  def signum(x: Double): Double = java.lang.Math.signum(x)
+  def signum(d: Double): Double = java.lang.Math.signum(d)
+
+  def floorDiv(x: Int, y: Int): Int = java.lang.Math.floorDiv(x, y)
```
janekdb, line 225: @golem131 - WDYT of _Modulus and Quotient_ as the group name for `floorDiv` and `floorMod`? This fits in with the existing group names and is clear.

While this group contains just these four methods a short note in `@groupdesc` could be added

> Calculate quotient values by rounding to negative infinity.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
...
+
+  def floorMod(x: Int, y: Int): Int = java.lang.Math.floorMod(x, y)
+
+  def floorMod(x: Long, y: Long): Long = java.lang.Math.floorMod(x, y)
+
+  def copySign(magnitude: Double, sign: Double): Double = java.lang.Math.copySign(magnitude, sign)
+
+  def copySign(magnitude: Float, sign: Float): Float = java.lang.Math.copySign(magnitude, sign)
+
+  def nextAfter(start: Double, direction: Double): Double = java.lang.Math.nextAfter(start, direction)
```
janekdb, line 237: @golem131 - Suggestion for `next*` group name: _Adjacent Floats_

It's short and to the point. For anyone hunting out the functionality of a `next*` method this is a strong cue.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
...
+
+  def floorDiv(x: Int, y: Int): Int = java.lang.Math.floorDiv(x, y)
+
+  def floorDiv(x: Long, y: Long): Long = java.lang.Math.floorDiv(x, y)
+
+  def floorMod(x: Int, y: Int): Int = java.lang.Math.floorMod(x, y)
+
+  def floorMod(x: Long, y: Long): Long = java.lang.Math.floorMod(x, y)
+
+  def copySign(magnitude: Double, sign: Double): Double = java.lang.Math.copySign(magnitude, sign)
```
janekdb, line 233: @golem131 `copySign` can be included in the "Signs" group. The group name is `signum` so rename that to `@group signs` to make it less specific to `signum`.

src/library/scala/math/package.scala
```diff
@@ -215,9 +218,37 @@ package object math {
...
+
+  def nextUp(d: Double): Double = java.lang.Math.nextUp(d)
+
+  def nextUp(f: Float): Float = java.lang.Math.nextUp(f)
+
+  def nextDown(d: Double): Double = java.lang.Math.nextDown(d)
+
+  def nextDown(f: Float): Float = java.lang.Math.nextDown(f)
+
+  def scalb(d: Double, scaleFactor: Int): Double = java.lang.Math.scalb(d, scaleFactor)
```
janekdb, line 249: @SethTisue  - Add a new group called `scaling` for `scalb`

```
@group scaling
```

Use this description,

```
@groupdesc scaling Scaling with rounding guarantees
```

janekdb:
@golem131 - Comments on group names added.

janekdb:
@golem131 - The final commit message should follow the commit message guidelines found here: 

```
https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md
Clean commits, clean history
```

src/library/scala/math/package.scala
```diff
@@ -63,6 +63,9 @@ package scala
   *
   * @groupname randomisation Pseudo Random Number Generation
   * @groupprio randomisation 130
+  *
+  * @groupname exact Functions to avoid overflow on primitive types
```
janekdb, line 67: @SethTisue - Change to

```
@groupname exact Exact Arithmetic
@groupdesc exact Integral addition, multiplication, stepping and conversion throwing ArithmeticException instead of underflowing or overflowing
```

janekdb:
@SethTisue - Reviewed. Changes suggested.

janekdb:
@SethTisue - Grant me push permission for this PR and I will happily take it over the finishing line.

SethTisue:
@janekdb do I have that power? is there a GitHub feature I don't know about? if not, just submit your own PR that includes @golem131's commits

janekdb:
Noone has that power as far as I can tell! Progressing with #5382

SethTisue:
@janekdb well hey looky here https://github.com/blog/2247-improving-collaboration-with-forks . too late for this PR though ("pull requests created before the option was available will default to not allowing collaboration"), so let's stick with the new one

janekdb:
@SethTisue - Well, well, well! Nice find.

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

src/library/scala/math/package.scala
janekdb: Done in #5382

PR #5315, opened 2016-08-01
Fix race condition in lambda deserialization

retronym:
Review of the code made me aware that concurrent calls to `$deserializeLambda$` for some lambda hosting class could result in concurrent calls to operations on `j.u.HashMap`.

I've added a synchronized block to avoid this problem. I don't think this is likely to be a bottleneck in practical use cases, but if so we could come up with a lock-free scheme in the future.


retronym:
Review by @lrytz

szeiger:
LGTM

PR #5316, opened 2016-08-01
SI-7838 Mark tail var in `::` as volatile

szeiger:
As recommended in SI-7838. I did some benchmarking by building Scala (`sbt clean dist/mkQuick`) with both, this version and the previous commit as STARR. Since scalac uses `List` heavily in its internal data structures, this should make a good real-life test case.

Before: 315 305 300 313 292 302 299 302 301 300 300
After: 314 318 302 301 299 298 297 316 303 301 302

All build times in seconds. Running on an otherwise idle machine, there's still more variation than I'd like but it doesn't look like the new version is substantially slower.

### DIFF
--- a/src/library/scala/collection/immutable/List.scala
+++ b/src/library/scala/collection/immutable/List.scala
@@ -431,7 +431,7 @@ case object Nil extends List[Nothing] {
  *  @since   2.8
  */
 @SerialVersionUID(509929039250432923L) // value computed by serialver for 2.11.2, annotation added in 2.11.4
-final case class ::[B](override val head: B, private[scala] var tl: List[B]) extends List[B] {
+final case class ::[B](override val head: B, @volatile private[scala] var tl: List[B]) extends List[B] {
   override def tail : List[B] = tl
   override def isEmpty: Boolean = false
 }
--- a/src/library/scala/collection/immutable/List.scala
+++ b/src/library/scala/collection/immutable/List.scala
@@ -431,9 +431,10 @@ case object Nil extends List[Nothing] {
  *  @since   2.8
  */
 @SerialVersionUID(509929039250432923L) // value computed by serialver for 2.11.2, annotation added in 2.11.4
-final case class ::[B](override val head: B, @volatile private[scala] var tl: List[B]) extends List[B] {
+sealed case class ::[B](override val head: B, private[scala] var tl: List[B]) extends List[B] {
   override def tail : List[B] = tl
-  override def isEmpty: Boolean = false
+  final override def isEmpty: Boolean = false
+  private[collection] final def freeze: ::[B] = new SafeCons(head, tl)
 }
 
 /** $factoryInfo
--- a/src/library/scala/collection/immutable/SafeCons.java
+++ b/src/library/scala/collection/immutable/SafeCons.java
@@ -1,0 +1,25 @@
+package scala.collection.immutable;
+
+import scala.collection.mutable.Builder;
+
+/** A subclass of `::` which routes access to the usually mutable `tl` var through a final val.
+ * This ensures thread-safe publishing of the tail (cf. SI-7838).
+ */
+final class SafeCons extends $colon$colon {
+  private final List tl2;
+
+  public SafeCons(Object _head, List<?> _tail) {
+    super(_head, _tail);
+    this.tl2 = _tail;
+    // freeze action for `tl2` happens here
+  }
+
+  // Scala generates two identical accessor methods which read the `tl` field directly.
+  // We have to override both of them. All other generated methods go through one of
+  // these accessors:
+  @Override public List tl$access$1() { return tl2; }
+  @Override public List tl() { return tl2; }
+
+  // Disambiguate `newBuilder`, which is inherited twice
+  @Override public Builder newBuilder() { return super.newBuilder(); }
+}
--- a/src/library/scala/collection/mutable/ListBuffer.scala
+++ b/src/library/scala/collection/mutable/ListBuffer.scala
@@ -308,7 +308,10 @@ final class ListBuffer[A]
    */
   override def toList: List[A] = {
     exported = !isEmpty
-    start
+    start match {
+      case cons: ::[A] => cons.freeze
+      case _ => start
+    }
   }
 
 // New methods in ListBuffer
--- a/test/benchmarks/README.md
+++ b/test/benchmarks/README.md
@@ -5,9 +5,7 @@ that makes use of the [SBT plugin](https://github.com/ktoso/sbt-jmh) for [JMH](h
 
 ## Running a benchmark
 
-The benchmarks require first building Scala into `../../build/pack` with `ant`.
-If you want to build with `sbt dist/mkPack` instead,
-you'll need to change `scalaHome` in this project.
+The benchmarks require first building Scala into `../../build/pack`.
 
 You'll then need to know the fully-qualified name of the benchmark runner class.
 The benchmarking classes are organized under `src/main/scala`,
--- a/test/benchmarks/src/main/scala/scala/collection/immutable/VolatileListBenchmark.scala
+++ b/test/benchmarks/src/main/scala/scala/collection/immutable/VolatileListBenchmark.scala
@@ -1,0 +1,51 @@
+package scala.collection.immutable;
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+
+import scala.collection.mutable.ListBuffer
+
+/** Benchmark some list operations to check the impact of making `::.tl` volatile. */
+class VolatileListBenchmark {
+
+  @Benchmark
+  def buildSum10: Int = {
+    val l = new ListBuffer[Int]
+    var i = 0
+    while(i < 10) {
+      l += i
+      i += 1
+    }
+    l.result.sum
+  }
+
+  @Benchmark
+  def buildSum1000: Int = {
+    val l = new ListBuffer[Int]
+    var i = 0
+    while(i < 1000) {
+      l += i
+      i += 1
+    }
+    l.result.sum
+  }
+
+  @Benchmark
+  def buildImmutableSum1000: Int = {
+    var l1: List[Int] = Nil
+    var i = 0
+    while(i < 1000) {
+      l1 = i :: l1
+      i += 1
+    }
+    l1.reverse.sum
+  }
+
+  @Benchmark
+  def sum1000: Int = VolatileListBenchmark.l1000.sum
+}
+
+object VolatileListBenchmark {
+  val l1000 = (0 to 1000).toList
+}


Jasper-M:
Shouldn't they be volatile instead of transient?

szeiger:
Yes, obviously. Time for another round of benchmarks...

szeiger:
Updated to use `volatile`. The build will fail due to SI-8873 though.

soc:
@szeiger Mhhh, looks like a case for the general improvement of the target meta annotations.
I had that work planned after getting in the other annotation improvements, but I'm kind of stuck on that until now. :-(

szeiger:
@soc I'm looking into it. At least the description in the ticket makes it sound like this should be easy to fix. (Maybe you'll want to rewrite it in a more comprehensive way later)

adriaanm:
SI-8873 sounds like something I can take on in #5206

szeiger:
Using https://github.com/scala/scala/compare/2.12.x...szeiger:issue/8873 as a tentative fix for SI-8873, I did new benchmarks, this time building only `clean compile/library` instead of `clean dist/mkQuick` and making sure the search indexer wasn't running in the background.

Before this PR: 92 93 92 92 94 92
With this PR: 93 93 90 92 91 91

Despite doing everything wrong initially, the conclusion from the new version is still the same: No obvious performance degradation.

adriaanm:
Could you look into devising a JMH test case that represents the worst case of contested reads/writes to a volatile var? I suspect the compiler may not reveal the overhead because there are so many things going on, and it's also single threaded.

szeiger:
I can certainly do more micro-level testing, but there will never be any contention over this var. In a correct program it is only ever accessed from a single thread or with external synchronization. Making it volatile is only necessary to close the loophole that occurs when building a List via mutation in one thread and then handing the resulting immutable List off to another thread.

Ichoran:
The benchmark should create it in one thread and hand it off to 2+ other threads that traverse it repeatedly.  There isn't any genuine contention there (it's all reading) but it might uncover a slowdown. Also, try stuff like `xs.head * xs.head + xs.head`.

adriaanm:
You're right, I didn't use the correct term 😬, but there should be some
cost to enforcing the memory model, right? If I understood correctly
writing a volatile var is like releasing a lock and reading is like
acquiring one.

szeiger:
I don't see why you'd want to benchmark the hand-off from one thread to another. It's broken under the current implementation. Comparing the performance of a working implementation and a broken one doesn't seem very interesting.

I focused on single-threaded benchmarks instead. Those are the scenarios that already work correctly under the old implementation and we want to know whether a bug fix has a negative impact on them.

I've added some JMH benchmarks for this purpose. `buildSum1000` appends 1000 Ints to a ListBuffer, converts it to a List and traverses that to compute the sum. The old implementation gets 104k ops/s, the new one with volatile `tl` only 47k. With only 10 elements instead of 1000 we get the same picture: 13M for the old version, 5.4M for the new one. All with standard settings in JDK 8 on x64. So the JVM is unable to optimize the write barriers away and the volatile version ends up being much slower.

I also did a test with a pre-constructed List which is only traversed during the benchmark run in `sum1000`. With this I get 175k ops/s on both versions. The volatile `tl` has absolutely no impact on List traversal. This is to be expected (with Hotspot on x64) because `volatile` only changes the write semantics. There are no read barriers.

I'll think some more about exploiting the freezing of `final` values for an alternative fix. The strong guarantees that a volatile `tl` gives us (you always get consistent reads from another thread while a List is being built in a mutable way) aren't really necessary. Maybe there's a different solution that would only guarantee safe publishing from `ListBuffer.toList`.

szeiger:
@adriaanm My high-level understanding (which has fortunately been backed up by the benchmarks so far) is that reading a volatile var is free but writing is expensive because it flushes cache lines.

Ichoran:
The "one creator many readers" version works correctly now as long as there's a decent amount of time in practice between the creation and the reading.  The idea that immutable data structures make concurrency easier only makes sense if you have concurrent access at some point!  So I agree that testing read/write contention is pointless (as right now it's broken so any speed is better than what we've got).  I am pretty sure that the way that volatile is implemented in practice (as a memory barrier on write) gives only a penalty on writing not reading, but I wouldn't be 100% sure without testing read-read contention.  And the description of volatile insists that for operations like a*a+a, you have to access the variable three times instead of once (doing the rest in registers), which really ought to show a slowdown.

Is it possible to use a volatile var as a deployment mechanism for the buffer so we only pay the price once per list creation instead of per element, while still gaining the benefits of a memory barrier?

szeiger:
@Ichoran I don't think so. "We" (i.e. ListBuffer) only operate single-threaded. It's the caller who would have to use a synchronized var, which brings us back to square one because we want to give the caller the illusion that List is frozen. Exploiting [freeze actions on finals](https://shipilev.net/blog/2014/jmm-pragmatics/#_what_do_we_have_5) looks more promising because it can establish a transitive happens-before relationship between the freeze action and a subsequent read on another thread.

Ichoran:
It would be great if that could be made to work.  I never was able to conceive of a scheme that would do it given that the mutation happens at the end of the list, what is passed is the beginning of the list, and my reading of the happens-befores was that establishing transitivity was tricky.  Given that the JMM doesn't seem to align very well with any real memory architecture, though, I imagine that there might be a way to get things to work in practice that would not break for any reasonable hardware.

szeiger:
Here's a sketch of what I think might work:

The usual Scala List classes:

```
package foo

sealed abstract class MyList[+T] {
  def head: T
  def tail: MyList[T]
}

case object MyNil extends MyList[Nothing] {
  override def head = ???
  override def tail = ???
}

sealed case class MyCons[T](override val head: T, private[foo] var tl: MyList[T]) extends MyList[T] {
  override def tail = tl
}
```

Unfortunately this part has to be done in Java:

```
package foo;

public final class SafeCons<T> extends MyCons<T> {
  private final MyList<T> tl2;

  public SafeCons(T _head, MyList<T> _tail) {
    super(_head, _tail);
    this.tl2 = _tail;
  }

  @Override public MyList<T> tl$access$1() {
    return tl2;
  }

  @Override public MyList<T> tl() {
    return tl2;
  }
}
```

When you construct a List in a mutable way you can use the usual `::` (or `MyCons` as I called it above). When you want to publish the result (for example, in `ListBuffer.toList`), you have to replace the first cons cell by a new `SafeCons`. This initializes the final field `tl2`, thereby freezing all the vars. Any access to `::.tl` from within `::` goes through either `tl()` or `tl$access$1()`, so we override them to read the frozen field.

adriaanm:
Stefan, how far should we go in the alternative design route in 2.12.0 versus 2.13?

See also https://shipilev.net/blog/2014/safe-public-construction for some examples of stress testing various safe publication mechanisms.

Ichoran:
You've now got three child classes instead of two, though, and they're all visited.  I worry that multiple dispatch will impact performance.  Definitely worth a try, though!

szeiger:
@adriaanm Our only constraint is time. It's been broken all along, so it's not a blocker. We can keep it broken for another release. But I can't think of any solution that would benefit from the lower source compatibility constraints in 2.13. Keeping it compatible is easy, making it fast is the difficult part.

@Ichoran Yes, that's my worry as well. OTOH, we already have multiple dispatch with `Nil` and `::` so it may not be that different after all.

szeiger:
Does anyone have a snippet that can actually demonstrate an error due to SI-7838? I tried several variations on https://gist.github.com/jrudolph/6552186 but cannot make them fail.

szeiger:
Updated with a proper `SafeCons`-based implementation. Here are the benchmarks results:

```
old:

[info] Benchmark                                     Mode  Cnt         Score       Error  Units
[info] VolatileListBenchmark.buildSum10             thrpt  200  13413346.222 ± 51462.101  ops/s
[info] VolatileListBenchmark.buildSum1000           thrpt  200    102901.091 ±   372.979  ops/s
[info] VolatileListBenchmark.sum1000                thrpt  200    171224.368 ±   618.348  ops/s
[info] VolatileListBenchmark.buildImmutableSum1000  thrpt  200     84930.092 ±   333.690  ops/s

SafeCons:

[info] Benchmark                                     Mode  Cnt        Score       Error  Units
[info] VolatileListBenchmark.buildSum10             thrpt  200  3236840.930 ± 12309.165  ops/s
[info] VolatileListBenchmark.buildSum1000           thrpt  200   106761.615 ±   360.415  ops/s
[info] VolatileListBenchmark.sum1000                thrpt  200   166865.492 ±   584.621  ops/s
[info] VolatileListBenchmark.buildImmutableSum1000  thrpt  200    84891.105 ±   282.742  ops/s
```

As a bit of a surprise, building a large List is actually faster than before but traversal is slower. Both are within a few percent of the old score though. The problem is building a small List, which has become 4x slower.

Apart from the performance considerations, I'm also not 100% convinced that the solution is correct (unlike the `volatile` version before, where it is more obvious) but I have no way to verify that without being able to reproduce SI-7838 in the first place.

szeiger:
I wasn't able to come up with a better solution. I think we should document the current limitations and put SI-7838 into the backlog. The performance penalty is too big for this corner case which has a simple workaround.

adriaanm:
Sounds good. Thanks for the thorough investigation

szeiger:
Closing in favor of https://github.com/scala/scala/pull/5323

lrytz:
> still more variation than I'd like

one thing you could try to improve this situation is disabling turbo boost (https://github.com/rugarciap/Turbo-Boost-Switcher for mac)

szeiger:
@lrytz I benchmarked the compiler on a Windows box. The variation was probably due to the indexing service running in the background.

Ichoran:
@szeiger - Since we're not doing this it doesn't matter right now, but the JVM special-cases multiple dispatch with two alternatives.  IIRC in my hands I saw a ~3 ns penalty per dispatch on 3 vs 2 in typical cases on 3-4 GHzish Intel CPUs; 2 is usually ~1 ns penalty over one, which is about the cost of an if-statement with decent but imperfect branch prediction.

It may have more to do with CPUs than the JIT compiler, though, because I couldn't find any alternative with 3+ branches that was substantially better (ifs, switch, etc., all seemed about the same).  The testing was not exhaustive, but I did do a lot of it with an attempt at a hybrid OptionEither with three states (equivalent to "None", "Left", "Right"), which is very much like what you suggested here.  The only possible difference is that if the JIT compiler did loop unrolling, it might be able to special case the first entry and then fall back to 2-ary dispatch on the tail of the list, which would presumably rescue performance.

szeiger:
@Ichoran That could be the explanation for the slowdown in traversal. It was small but measurable. The builder performance is still not clear to me but I didn't investigate it further. Since building a 1000 element list didn't become slower, it can't be a fundamental problem with this approach. Of course, throwing away one cons cell and replacing it with a new one has a much higher overhead in a 10 element list but it's still only 1/10th of the number of elements built, hardly enough to explain a 4x slowdown.

PR #5317, opened 2016-08-02
SD-192 Change scheme for trait super accessors

retronym:
Rather than putting the code of a trait method
body into a static method, leave it in the default
method. The static method (needed as the target
of the super calls) now uses `invokespecial`
to exactly call that method.

The [JVM spec](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.invokespecial) is slightly ambiguous for this case. "If all of the following are true, let C be the direct superclass of the current class" should apply, all of the conditions in the list are true. But there's no superclass, as T is an interface.

Relevant implementation in HotSpot:

http://hg.openjdk.java.net/jdk/jdk/file/c30c35118303/src/hotspot/share/interpreter/linkResolver.cpp#l1126
http://hg.openjdk.java.net/jdk/jdk/file/c30c35118303/src/hotspot/share/interpreter/linkResolver.cpp#l867

Here's the inclusion of the check for `is_same_or_direct_interface`

http://hg.openjdk.java.net/jdk/jdk/annotate/b7dba4cde1c6/hotspot/src/share/vm/interpreter/linkResolver.cpp#l924

Implementing part of [JSR-335 Lambda Expressions](https://jcp.org/aboutJava/communityprocess/final/jsr335/index.html)

image

Even thought the discussion is all about the use case of naming a super-interface method, its explicitly says its okay for invokespecial to name a method in the current interface, which is what we're doing.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
@@ -8,12 +8,12 @@ package scala.tools.nsc
 package backend
 package jvm
 
-import scala.collection.{ mutable, immutable }
+import scala.collection.{immutable, mutable}
 import scala.tools.nsc.symtab._
-
 import scala.tools.asm
 import GenBCode._
 import BackendReporting._
+import scala.tools.nsc.backend.jvm.BCodeHelpers.InvokeStyle
 
 /*
  *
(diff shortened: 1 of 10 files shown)


retronym:
https://github.com/scala/scala/pull/5317

retronym:
This one needs to be in 2.12.0

retronym:
Looks like I'll need some checkfile updates.

BTW, here are the new stack traces:

```
scala> trait T { def foo = ??? }; class C extends T; class Test { def test = new C().foo }; new Test().test
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:284)
  at T.foo(<console>:11)
  at Test.test(<console>:11)
  ... 30 elided
```

```
scala> trait T { def foo = ??? }; trait U extends T { override def foo = ??? }; class Test extends T with U { def test = super[T].foo }; new Test().test
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:284)
  at T.foo(<console>:11)
  at T.foo$(<console>)
  at Test.test(<console>:11)
```

retronym:
Pushed a checkfile update and removed extraneous parts of the diff.

src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
```diff
@@ -491,18 +491,22 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
...
+              val statified = global.gen.mkStatic(dd, sym.name, _.cloneSymbol)
+              genDefDef(statified)
+            } else {
+              val forwarderDefDef = {
+                val dd1 = global.gen.mkStatic(deriveDefDef(dd)(_ => EmptyTree), traitSuperAccessorName(sym), _.cloneSymbol)
+                dd1.symbol.setFlag(Flags.ARTIFACT).resetFlag(Flags.OVERRIDE)
+                deriveDefDef(dd1)(_ =>
+                  atPos(dd1.pos)(
+                    Apply(Select(global.gen.mkAttributedIdent(dd1.symbol.paramss.head.head).setType(sym.owner.typeConstructor), dd.symbol),
+                    dd.vparamss.head.map(p => global.gen.mkAttributedIdent(p.symbol))).updateAttachment(UseInvokeSpecial))
```
lrytz: It seems a bit hackish to refer to `dd`'s parameter symbols within the body of `dd1`. Should it be `dd1.vparamss.head.tail`?

test/junit/scala/collection/mutable/OpenHashMapTest.scala
```diff
@@ -52,6 +52,7 @@ class OpenHashMapTest {
 
   /** Test that an [[OpenHashMap]] frees references to a deleted key (SI-9522). */
   @Test
+  @Ignore
```
lrytz: did this slip in accidentally?

lrytz:
the use of `invokespecial` looks legit ([spec](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.invokespecial)): the class name in invocation's descriptor should always be the current class (in BCodeSkelBuilder, the static forwarder has an `Apply(Select(.., dd.symbol), ..)`, where `dd` is a method in the `Template` of the class that's currently generated. So `dd.owner` should be the current class.), so method selection starts at the same class, and the method will be immediately found.

retronym:
@lrytz Fixed the reference to the symbols of `dd1`, and remove the stray `@Ignore` annotation.

retronym:
```
[info] 'compiler-interface' not yet compiled for Scala 2.12.0-20160813-112758-eb236ce. Compiling...
java.util.NoSuchElementException: key not found: value phase
    at scala.collection.MapLike.default$(MapLike.scala:232)
    at scala.collection.MapLike.default(MapLike.scala:231)
    at scala.collection.mutable.HashMap.apply(HashMap.scala:65)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder$locals$.load(BCodeSkelBuilder.scala:385)
    at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:352)
    at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.$anonfun$genLoadArguments$1(BCodeBodyBuilder.scala:940)
    at 
```

https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/2804/console

❓

adriaanm:
I aborted the job -- something fishy there :-)

lrytz:
LGTM

PR #5321, opened 2016-08-07
SD-193 Lock down lambda deserialization

retronym:
The old design allowed a forged `SerializedLambda` to be
deserialized into a lambda that could call any private method
in the host class.

This commit passes through the list of all lambda impl methods
to the bootstrap method and verifies that you are deserializing
one of these.

The new test case shows that a forged lambda can no longer call
the private method, and that the new encoding is okay with a large
number of lambdas in a file.

We already have method handle constants in the constant pool
to support the invokedynamic through LambdaMetafactory, so
the only additional cost will be referring to these in
the boostrap args for `LambdaDeserialize`, 2 bytes per lambda.

I checked this with an example:

  https://gist.github.com/retronym/e343d211f7536d06f1fef4b499a0a177

fixes scala/scala-dev#193

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -124,12 +124,13 @@ abstract class BTypes {
    */
   val indyLambdaImplMethods: mutable.AnyRefMap[InternalName, mutable.LinkedHashSet[asm.Handle]] = recordPerRunCache(mutable.AnyRefMap())
   def addIndyLambdaImplMethod(hostClass: InternalName, handle: Seq[asm.Handle]): Unit = {
-    indyLambdaImplMethods.getOrElseUpdate(hostClass, mutable.LinkedHashSet()) ++= handle
+    if (handle.nonEmpty)
+      indyLambdaImplMethods.getOrElseUpdate(hostClass, mutable.LinkedHashSet()) ++= handle
   }
-  def getIndyLambdaImplMethods(hostClass: InternalName): List[asm.Handle] = {
+  def getIndyLambdaImplMethods(hostClass: InternalName): Iterable[asm.Handle] = {
     indyLambdaImplMethods.getOrNull(hostClass) match {
       case null => Nil
-      case xs => xs.toList.distinct
+      case xs => xs
     }
   }
 
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -122,7 +122,16 @@ abstract class BTypes {
    * inlining: when inlining an indyLambda instruction into a class, we need to make sure the class
    * has the method.
    */
-  val indyLambdaHosts: mutable.Set[InternalName] = recordPerRunCache(mutable.Set.empty)
+  val indyLambdaImplMethods: mutable.AnyRefMap[InternalName, mutable.LinkedHashSet[asm.Handle]] = recordPerRunCache(mutable.AnyRefMap())
+  def addIndyLambdaImplMethod(hostClass: InternalName, handle: Seq[asm.Handle]): Unit = {
+    indyLambdaImplMethods.getOrElseUpdate(hostClass, mutable.LinkedHashSet()) ++= handle
+  }
+  def getIndyLambdaImplMethods(hostClass: InternalName): List[asm.Handle] = {
+    indyLambdaImplMethods.getOrNull(hostClass) match {
+      case null => Nil
+      case xs => xs.toList.distinct
+    }
+  }
 
   /**
    * Obtain the BType for a type descriptor or internal name. For class descriptors, the ClassBType
--- a/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
@@ -289,19 +289,6 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
         coreBTypes.jliCallSiteRef
       ).descriptor,
       /* itf = */ coreBTypes.srLambdaDeserialize.isInterface.get)
-  lazy val lambdaDeserializeAddTargets =
-    new scala.tools.asm.Handle(scala.tools.asm.Opcodes.H_INVOKESTATIC,
-      coreBTypes.srLambdaDeserialize.internalName, "bootstrapAddTargets",
-      MethodBType(
-        List(
-          coreBTypes.jliMethodHandlesLookupRef,
-          coreBTypes.StringRef,
-          coreBTypes.jliMethodTypeRef,
-          ArrayBType(coreBTypes.jliMethodHandleRef)
-        ),
-        coreBTypes.jliCallSiteRef
-      ).descriptor,
-      /* itf = */ coreBTypes.srLambdaDeserialize.isInterface.get)
 }
 
 /**
--- a/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
@@ -283,7 +283,21 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
         List(
           coreBTypes.jliMethodHandlesLookupRef,
           coreBTypes.StringRef,
-          coreBTypes.jliMethodTypeRef
+          coreBTypes.jliMethodTypeRef,
+          ArrayBType(jliMethodHandleRef)
+        ),
+        coreBTypes.jliCallSiteRef
+      ).descriptor,
+      /* itf = */ coreBTypes.srLambdaDeserialize.isInterface.get)
+  lazy val lambdaDeserializeAddTargets =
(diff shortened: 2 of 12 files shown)


retronym:
This encoding avoids the javac's problem ([JDK-8008413](https://bugs.openjdk.java.net/browse/JDK-8008413)) of `$deserializeLambda$` growing too large, but delivers the same security AFAICT.

retronym:
@lrytz Before we merge this, we need to refactor the backend a little to add the fully elaborated `$deserializeLambda$` method after the inliner has run. This PR removes and readds the method after inlining lambda captures which is a bit of hack. Suggestions for how to structure this would be welcome.

retronym:
Hat tip to @forax for highlighting this issue at JVM Language Summit.

retronym:
@lrytz I've pushed another commit that tries to add this method once and once only.

retronym:
I've [posted](http://mail.openjdk.java.net/pipermail/mlvm-dev/2016-August/006699.html) a summary of this change to the OpenJDK mailing list, mlvm-dev, to solicit further feedback and to highlight it as a potential fix for JDK-8008413.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -122,7 +122,16 @@ abstract class BTypes {
...
    */
-  val indyLambdaHosts: mutable.Set[InternalName] = recordPerRunCache(mutable.Set.empty)
+  val indyLambdaImplMethods: mutable.AnyRefMap[InternalName, mutable.LinkedHashSet[asm.Handle]] = recordPerRunCache(mutable.AnyRefMap())
+  def addIndyLambdaImplMethod(hostClass: InternalName, handle: Seq[asm.Handle]): Unit = {
+    indyLambdaImplMethods.getOrElseUpdate(hostClass, mutable.LinkedHashSet()) ++= handle
+  }
+  def getIndyLambdaImplMethods(hostClass: InternalName): List[asm.Handle] = {
+    indyLambdaImplMethods.getOrNull(hostClass) match {
+      case null => Nil
+      case xs => xs.toList.distinct
```
lrytz: why `distinct`, given `xs` is a set?

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -104,16 +105,16 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
    * a boolean indicating if the instruction list contains an instantiation of a serializable SAM
```
lrytz: no longer returns a boolean

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
retronym: That should be removed. I switched from `List` to `Set` along the way but neglected to remove this.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -87,12 +87,13 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
     val nilLookupDesc = MethodBType(Nil, jliMethodHandlesLookupRef).descriptor
     val serlamObjDesc = MethodBType(jliSerializedLambdaRef :: Nil, ObjectRef).descriptor
+    val addTargetMethodsObjDesc = MethodBType(ObjectRef :: Nil, UNIT).descriptor
```
lrytz: unused

src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
```diff
@@ -283,7 +283,21 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
...
           coreBTypes.jliMethodHandlesLookupRef,
           coreBTypes.StringRef,
-          coreBTypes.jliMethodTypeRef
+          coreBTypes.jliMethodTypeRef,
+          ArrayBType(jliMethodHandleRef)
+        ),
+        coreBTypes.jliCallSiteRef
+      ).descriptor,
+      /* itf = */ coreBTypes.srLambdaDeserialize.isInterface.get)
+  lazy val lambdaDeserializeAddTargets =
```
lrytz: unused

src/library/scala/runtime/LambdaDeserialize.java
```diff
@@ -2,28 +2,41 @@
...
 
 import java.lang.invoke.*;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 
 public final class LambdaDeserialize {
+    public static final MethodType DESERIALIZE_LAMBDA_MT = MethodType.fromMethodDescriptorString("(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object;", LambdaDeserialize.class.getClassLoader());
+    public static final MethodType ADD_TARGET_METHODS_MT = MethodType.fromMethodDescriptorString("([Ljava/lang/invoke/MethodHandle;)V", LambdaDeserialize.class.getClassLoader());
```
lrytz: unused

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -405,10 +405,7 @@ class Inliner[BT <: BTypes](val btypes: BT) {
 
     callsiteMethod.maxStack = math.max(callsiteMethod.maxStack, math.max(stackHeightAtNullCheck, maxStackOfInlinedCode))
 
-    if (hasSerializableClosureInstantiation && !indyLambdaHosts(callsiteClass.internalName)) {
-      indyLambdaHosts += callsiteClass.internalName
-      addLambdaDeserialize(byteCodeRepository.classNode(callsiteClass.internalName).get)
-    }
+    addIndyLambdaImplMethod(callsiteClass.internalName, targetHandles)
```
lrytz, line 408: maybe this should be done only `if (targetHandles.nonEmpty)`? or that test could be added to the implementation of `addIndyLambdaImplMethod`.

src/library/scala/runtime/LambdaDeserialize.java
```diff
@@ -2,28 +2,41 @@
 
 
 import java.lang.invoke.*;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
```
lrytz: the new imports are also unused

src/library/scala/runtime/LambdaDeserialize.java
```diff
@@ -2,28 +2,41 @@
...
-        MethodType type = MethodType.fromMethodDescriptorString("(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object;", lookup.getClass().getClassLoader());
-        MethodHandle deserializeLambda = lookup.findVirtual(LambdaDeserialize.class, "deserializeLambda", type);
-        MethodHandle exact = deserializeLambda.bindTo(new LambdaDeserialize(lookup)).asType(invokedType);
+                                     MethodType invokedType, MethodHandle... targetMethods) throws Throwable {
+        MethodHandle deserializeLambda = lookup.findVirtual(LambdaDeserialize.class, "deserializeLambda", DESERIALIZE_LAMBDA_MT);
+        MethodHandle exact = deserializeLambda.bindTo(new LambdaDeserialize(lookup, targetMethods)).asType(invokedType);
         return new ConstantCallSite(exact);
     }
+    public static String nameAndDescriptorKey(String name, String descriptor) {
+        return name + " " + descriptor;
```
lrytz: could save the `" "`

src/library/scala/runtime/LambdaDeserializer.scala
```diff
@@ -69,7 +71,15 @@ object LambdaDeserializer {
 
       // Lookup the implementation method
       val implMethod: MethodHandle = try {
-        findMember(lookup, getImplMethodKind, implClass, getImplMethodName, implMethodSig)
+        if (targetMethodMap != null) {
```
lrytz: what case does this cover, is it for backwards compatibility? maybe this can be exploited. could we just remove the case (and the `findMember` method)?

lrytz:
aside from these details, looks good!

src/library/scala/runtime/LambdaDeserializer.scala
retronym: I left it as a convenience for the existing unit test. But I agree its extraneous here and I'll move it into the test itself instead.

retronym: (The caller has to provide the `Lookup`, so there isn't an exploit, but minimizing surface area is good practice.)

src/library/scala/runtime/LambdaDeserialize.java
retronym: will do

src/library/scala/runtime/LambdaDeserialize.java
retronym: that's what you get pushing branches on 30 minutes of free wifi in Aukland airport 😄 

originally I thought that bootstrap arg list would be limited to 255 params (like regular param lists), so i'd worked on a "builder style".  But in fact it can be longer if the final param is an array type.

retronym:
@lrytz details polished

lrytz:
LGTM!

adriaanm:
Good stuff, @retronym! (I first removed the WIP label assuming that it was referring to the review comments, but maybe I should let you make that call, so restored the label.)

PR #5322, opened 2016-08-08
SD-194 Tweak module initialization to comply with JVM spec

retronym:
Top level modules in Scala currently desugar as:

```
class C; object O extends C { toString }
```

```
public final class O$ extends C {
  public static final O$ MODULE$;

  public static {};
    Code:
       0: new           #2                  // class O$
       3: invokespecial #12                 // Method "<init>":()V
       6: return

  private O$();
    Code:
       0: aload_0
       1: invokespecial #13                 // Method C."<init>":()V
       4: aload_0
       5: putstatic     #15                 // Field MODULE$:LO$;
       8: aload_0
       9: invokevirtual #21                 // Method java/lang/Object.toString:()Ljava/lang/String;
      12: pop
      13: return
}
```

The static initalizer `<clinit>` calls the constructor `<init>`, which
invokes superclass constructor, assigns `MODULE$= this`, and then runs
the remainder of the object's constructor (`toString` in the example
above.)

It turns out that this relies on a bug in the JVM's verifier: assignment to a
static final must occur lexically within the <clinit>, not from within `<init>`
(even if the latter is happens to be called by the former).

I'd like to move the assignment to <clinit> but that would
change behaviour of "benign" cyclic references between modules.

Example:

```
package p1; class CC { def foo = O.bar}; object O {new CC().foo; def bar = println(1)};

scala> p1.O
1
```

This relies on the way that we assign MODULE$ field after the super class constructors
are finished, but before the rest of the module constructor is called.

Instead, this commit removes the ACC_FINAL bit from the field. It actually wasn't
behaving as final at all, precisely the issue that the stricter verifier
now alerts us to.

```
scala> :paste -raw

package p1; object O

scala> val O1 = p1.O
O1: p1.O.type = p1.O$@ee7d9f1

scala> scala.reflect.ensureAccessible(p1.O.getClass.getDeclaredConstructor()).newInstance()
res0: p1.O.type = p1.O$@64cee07

scala> O1 eq p1.O
res1: Boolean = false
```

We will still achieve safe publication of the assignment to other threads
by virtue of the fact that `<clinit>` is executed within the scope of
an initlization lock, as specified by:

  https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.5

Fixes scala/scala-dev#SD-194


sjrd:
> I'd like to move the assignment to <clinit> but that would
> change behaviour of "benign" cyclic references between modules.

FTR, there are a lot of those "benign" cyclic references in the modules of the `ClassTag`/`Manifest` family. I learned that the hard way when I implemented Ozma (the "ancestor" of Scala.js) which is why Scala.js has been mimicking the "assign the static var after calling the super constructor" behavior forever.

adriaanm:
I vote for maintaining the status quo. We can change the encoding when we're targeting bytecode version 53 (java 9), once we do. Rumor has it that under version 52, our current init pattern will continue to be accepted.

adriaanm:
So, does this achieve [safe publication](https://shipilev.net/blog/2014/safe-public-construction/#_safe_publication) by:

> 1. Use static initializer to do the initializing stores ([JLS 12.4](http://docs.oracle.com/javase/specs/jls/se7/html/jls-12.html#jls-12.4))

EDIT: I don't think it does, so we shouldn't drop `ACC_FINAL`

adriaanm:
(see ongoing discussion at https://github.com/scala/scala-dev/issues/194)

adriaanm:
Ok, I forgot that the constructor is called from the `<clinit>` block, which means that we achieve safe publication under the class initializer's lock. Now LbetterTM

adriaanm:
Needs a rebase (--> wip label).

retronym:
rebased

lrytz:
Just noting that a final field can be changed through reflection:

```
$ cat Test.scala
class C {
  val f = 0
  var g = 0
}

$ scalac Test.scala
$ asm C.class
$ cat C.asm
public class C {

  // access flags 0x12
  private final I f

  // access flags 0x2
  private I g
}
$ scala
scala> val c = new C
c: C = C@6818d900

scala> val Array(f, g) = c.getClass.getDeclaredFields()
f: java.lang.reflect.Field = private final int C.f
g: java.lang.reflect.Field = private int C.g

scala> f.set(c, 123)
java.lang.IllegalAccessException: Class  can not access a member of class C with modifiers "private final"
  at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
  at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
  at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
  at java.lang.reflect.Field.set(Field.java:761)
  ... 28 elided

scala> g.set(c, 123)
java.lang.IllegalAccessException: Class  can not access a member of class C with modifiers "private"
  at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
  at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
  at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
  at java.lang.reflect.Field.set(Field.java:761)
  ... 28 elided

scala> f.setAccessible(true); g.setAccessible(true)

scala> f.set(c, 123)

scala> g.set(c, 123)

scala> c.f
res5: Int = 123

scala> c.g
res6: Int = 123
```

lrytz:
With the latest [comments on the ticket](https://github.com/scala/scala-dev/issues/194#issuecomment-239257830) and a bit of reading on Aleksey's blog, this LGTM for 2.12.0. Leaving the second review to @adriaanm.

adriaanm:
I'm also on board, but would like for the various clarifications from this PR and the issue to find their way to the commit message (regarding the memory model and the final bit not being needed for safe publication because we always initialize the field with the `<clinit>` lock). Then, LGTM :)

retronym:
I've included a note about safe publication to the commit messsage.

adriaanm:
LGTM

PR #5323, opened 2016-08-08
SI-7838 Document the multi-threading semantics of List and Vector

szeiger:
Making them completely thread-safe would be too expensive (in terms
of performance of single-threaded use cases).

Replaces #5316.

JIRA: https://issues.scala-lang.org/browse/SI-7838

### DIFF
--- a/src/library/scala/collection/immutable/List.scala
+++ b/src/library/scala/collection/immutable/List.scala
@@ -25,6 +25,8 @@ import java.io.{ObjectOutputStream, ObjectInputStream}
  *  This class is optimal for last-in-first-out (LIFO), stack-like access patterns. If you need another access
  *  pattern, for example, random access or FIFO, consider using a collection more suited to this than `List`.
  *
+ *  $usesMutableState
+ *
  *  ==Performance==
  *  '''Time:''' `List` has `O(1)` prepend and head/tail access. Most other operations are `O(n)` on the number of elements in the list.
  *  This includes the index-based lookup of elements, `length`, `append` and `reverse`.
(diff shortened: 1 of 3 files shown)


SethTisue:
unclear on what meaning "which is obtained in some way, for example from a Builder" is supposed to add. does it simply mean "regardless of how the instance was obtained"?

it's difficult to come up with a wording which is both clear and accurate, especially given the likelihood that most readers will be vague on JMM details. Given that, perhaps the best thing would be to include a link to https://issues.scala-lang.org/browse/SI-7838?

SethTisue:
over at #5316 you call this a "corner case which has a simple workaround", but what is the workaround?

szeiger:
The emphasis should be on "a new instance". This caveat does not apply to an existing list which was already published but only to freshly built instances.

There are many workarounds, like using a volatile var for the exchange, using a synchronized accessor, or wrapping it in an object with a final field. And to put "corner case" into perspective: I cannot think of _any_ case that is both useful and correct (under strict Java Memory Model semantics) but would fail due to SI-7838. Provoking the race condition that triggers an erroneous `null` read of the tail requires exchanging a freshly constructed `::` between threads through a non-volatile, non-synchronized read of a field. But there is no guarantee that such a read would _ever_ see the written value, so you cannot rely on it in a correct program. Once you establish a happens-before relationship to guarantee the value is visible, you are also guaranteed to see the correct, non-null tail.

szeiger:
Here's a revised version

varming:
You may want to mention [JEP 193](http://openjdk.java.net/jeps/193) and that the cheapest safe way to pass a list to another thread is by the pair java.lang.invoke.VarHandle.setRelease / VarHandle.getAcquire (available in Unsafe on JDK8). With standardization of these operations in JDK9 (Jep 193) programmers should have plenty of tools for safe concurrency.

szeiger:
@varming Thanks, I've added this to the JIRA ticket.

SethTisue:
LGTM

PR #5324, opened 2016-08-09
use Any instead of type-bound to clarify unboundedness

cvogt:
.contains accepts anything in these cases. While the previous code worked fine,
this signature is simpler and makes it slightly more obvious (especially to beginners)
that there is no type-safety here.

Do others agree with this judgement? Not married to this, but seems like a minor improvement to me to scare away type-system newbies a little bit less.

### DIFF
--- a/src/library/scala/Option.scala
+++ b/src/library/scala/Option.scala
@@ -227,7 +227,7 @@ sealed abstract class Option[+A] extends Product with Serializable {
    *  @return `true` if the option has an element that is equal (as
    *  determined by `==`) to `elem`, `false` otherwise.
    */
-  final def contains[A1 >: A](elem: A1): Boolean =
+  final def contains(elem: Any): Boolean =
     !isEmpty && this.get == elem
 
   /** Returns true if this option is nonempty '''and''' the predicate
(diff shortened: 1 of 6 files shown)


cvogt:
(tried my best to sign the CLA via that link, not sure that worked, it kept saying I should log in)

smarter:
It seems that ideally the signature of `contains` should be something like:

``` scala
def contains[B](elem: B)(implicit ev: (A <:< B) | (B <:< A)): Boolean 
```

This way you can't do:

``` scala
List(1).contains(true)
```

Not sure if that can be expressed without union types, or if it's worth the extra complexity

milessabin:
@smarter a single type class which captures that relationship can be defined like this,

``` scala
sealed class <:>[A, B]
object <:> {
  implicit def left[A, B <: A]: A <:> B = new (A <:> B)
  implicit def right[A,  B >: A]: A <:> B = new (A <:> B)
}
```

I don't have an opinion on whether or not using it would improve `contains` in these cases.

cvogt:
@smarter Potentially using a type-class here to improve safety is not in the scope of this PR. This PR keeps the status quo in terms of type-restrictions and simply makes the signature more obvious. It would be a separate issue to discuss potential restrictions and their implication and not in the scope of this PR.

soc:
I'd prefer keeping it this way. Two reasons:
- It breaks people's code who provided the type argument for this call (e. g. for documentation)
- It reminds us that we would like to provide better bounds, but currently can't (for technical reasons)

cvogt:
Not sold on the argument that we should let our signatures document desired future changes.

But not being able anymore to explicitly provide the type argument is a good point. Loosing that would probably be undesirable. Closing this.

PR #5325, opened 2016-08-09
rename toSet to toSetUp to highlight upcasting

cvogt:
toSet's behavior while sound can be surprising.
Despite Set being invariant .toSet happily upcasts the source collection if
this makes the compiler happy, even to a collection of Any is necessary,
because the source collection is co-variant.

This compiles: `Seq(1,2,3).toSet.contains("foo")`
This does not compile: `Seq(1,2,3).toSet.map(_*2)`

Renaming to toSetUp highlights this fact.

In most cases the less surprising operation is to[Set]:
This does not compile: `Seq(1,2,3).to[Set].contains("foo")`
This compiles: `Seq(1,2,3).to[Set].map(_*2)`

This hopefully helps users to avoid problems like these:
http://stackoverflow.com/questions/5544536/type-inference-fails-on-set-made-with-toset/6124177#6124177
http://stackoverflow.com/questions/27357353/scala-missing-parameter-type
https://issues.scala-lang.org/browse/SI-9116
https://issues.scala-lang.org/browse/SI-7743

Not married to the new name.


cvogt:
Note that deprecating toSet can probably not be considered binary compatible, because overrides of it would not affect consumers of toSetUp. Maybe this should go into 2.12.0 or 2.13.0.

soc:
I dislike the name `toSetUp`. I think it isn't clear what it does (and it looks like someone misspelled "setup").
I'd leave the name as is, especially as people think about reducing variance in various parts of collections.

While documenting it is desirable, I think it would make the APIs more cumbersome if we would start documenting every detail in the method's name.

cvogt:
@soc "documenting every detail" would indeed be overkill. But letting the method name reflect dangerous or confusing caveats is something desirable in my eyes. This is most prominent in some libraries prefixing some methods with "unsafe". The same argument applies here.

I am not totally sold on toSetUp either. What makes it ok to me is that without it people have frequently made the wrong assumption about what `toSet` does. The suffix `Up` is not self-explanatory, but at least forces people to wonder "what the heck is this" and once they read the docs, the word "up" can serve as a reminder that it finds an appropriate "up"per bound for the element type. I also realized the similarity to "setup", but that doesn't break the former reasoning - confusion about the exact meaning prompting to read the docs is better than what we have now I think.

cvogt:
`toLubSet` would be another possible name

cvogt:
(are these build failures legit?)

cvogt:
Another name suggestion `toSetOf` https://twitter.com/apmarki/status/763027148324032512

Ichoran:
A change of this magnitude should go in 2.13.  Also, I agree with Simon that the name would need to be improved, and that we should not get in the habit of writing documentation into our names.  Instead, if a method does something confusing we should explore avenues of making it work more intuitively, or accept that it inherently is that confusing and write better tutorials that explain exactly why and how it has to be this way.

So I agree that `.toSet` should undergo some serious scrutiny, but I don't think this PR is the obvious way to go.  (Especially not this late in the 2.12 release process!)

SethTisue:
We should continue to explore the solution space here, but I don't think this is mergeable.

PR #5326, opened 2016-08-10
SI-9881 Fix ImportHandler's reporting of importedNames and importedSymbols

jasonxh:
This PR attempts to fix `importedNames` and `importedSymbols` in `ImportHandler`. Currently they are broken, causing the following issues:
- https://issues.scala-lang.org/browse/SI-9881
- https://issues.scala-lang.org/browse/SI-9880

Changes in this PR:
- Keep track of both selector `names` & `renames`. Use `names` to lookup symbols from owner. Use `renames` to construct `importedNames`.
- `sym.thisType` will never work for module symbols. Change to ~~`sym.typeOfThis`~~ `sym.moduleClass.thisType`.
- Fix the bug where `typeOfExpression` inadvertently resolves type under `jvm` phase, disregarding requested phase from caller. This is used in `ImportHandler` to find `importableMembers` from objects.
- Cherry-picked @retronym 's change from https://github.com/retronym/scala/tree/review/4698 to filter out flattened inner symbols from `importableMembers`.

### DIFF
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -4549,7 +4549,12 @@ trait Types
 
   /** Members which can be imported into other scopes.
    */
-  def importableMembers(pre: Type): Scope = pre.members filter isImportable
+  def importableMembers(pre: Type): Scope = {
+    def isFlattenedSymbol(sym: Symbol) = sym.owner.isPackageClass && sym.name.containsName(nme.NAME_JOIN_STRING) && {
+      sym.owner.info.member(sym.name.take(sym.name.indexOf(nme.NAME_JOIN_STRING))) != NoSymbol
+    }
+    pre.members filter (m => !isFlattenedSymbol(m) && isImportable(m))
+  }
 
   def objToAny(tp: Type): Type =
     if (!phase.erasedTypes && tp.typeSymbol == ObjectClass) AnyTpe
--- a/src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
+++ b/src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
@@ -214,13 +214,16 @@ trait MemberHandlers {
     val Import(expr, selectors) = imp
     def targetType = intp.global.rootMirror.getModuleIfDefined("" + expr) match {
       case NoSymbol => intp.typeOfExpression("" + expr)
-      case sym      => sym.thisType
+      case sym      => sym.typeOfThis
     }
-    private def importableTargetMembers = importableMembers(targetType).toList
+    private def importableTargetMembers = importableMembers(exitingTyper(targetType)).toList
     // wildcard imports, e.g. import foo._
     private def selectorWild    = selectors filter (_.name == nme.USCOREkw)
-    // renamed imports, e.g. import foo.{ bar => baz }
-    private def selectorRenames = selectors map (_.rename) filterNot (_ == null)
+
+    // non-wildcard imports
+    private def individualSelectors = selectors filter { x =>
+      x.name != nme.USCOREkw && x.name != null && x.rename != nme.USCOREkw & x.rename != null
+    }
 
     /** Whether this import includes a wildcard import */
     val importsWildcard = selectorWild.nonEmpty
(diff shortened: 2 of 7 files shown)


jasonxh:
@retronym I cherry-picked some of your changes from https://github.com/retronym/scala/tree/review/4698 to filter out flattened symbols. It seems there hasn't been activity in that branch in the past year. Are you still working on it? If there is pending work in the area, please let me know and we can coordinate that.

retronym:
Thanks for picking this up. My branch is dormant, I'm not currently working on this.

jasonxh:
@retronym can you help review this PR, especially the changes I cherry-picked from your branch?

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4549,7 +4549,12 @@ trait Types
 
   /** Members which can be imported into other scopes.
    */
-  def importableMembers(pre: Type): Scope = pre.members filter isImportable
+  def importableMembers(pre: Type): Scope = {
```
adriaanm, line 4552: I think this change is too risky for 2.11.9. How about moving it to the invocation of `importableMembers` in the repl?

src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
```diff
@@ -214,27 +214,34 @@ trait MemberHandlers {
     val Import(expr, selectors) = imp
     def targetType = intp.global.rootMirror.getModuleIfDefined("" + expr) match {
       case NoSymbol => intp.typeOfExpression("" + expr)
-      case sym      => sym.thisType
+      case sym      => sym.moduleClass.thisType
     }
-    private def importableTargetMembers = importableMembers(targetType).toList
+    private def importableTargetMembers = importableMembers(exitingTyper(targetType)).toList
```
adriaanm, line 219: Does this PR introduce any new call-sites of `importableTargetMembers` that aren't already in `exitingTyper`?

Currently, IntelliJ says:
```
    lazy val individualSymbols: List[Symbol] = exitingTyper(importableTargetMembers filter (m => selectorNames(m.name)))
    lazy val wildcardSymbols: List[Symbol]   = exitingTyper(if (importsWildcard) importableTargetMembers else Nil)
```

src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
```diff
@@ -214,27 +214,34 @@ trait MemberHandlers {
     val Import(expr, selectors) = imp
     def targetType = intp.global.rootMirror.getModuleIfDefined("" + expr) match {
       case NoSymbol => intp.typeOfExpression("" + expr)
-      case sym      => sym.thisType
+      case sym      => sym.moduleClass.thisType
```
adriaanm, line 217: Good catch! 
```
    /** If symbol is a class, the type `this.type` in this class,
     * otherwise `NoPrefix`.
     * We always have: thisType <:< typeOfThis
     */
    def thisType: Type = NoPrefix
```

From the repl's `:power` mode:
```
scala> object o
defined object o

scala> val module = typeOf[o.type].termSymbol
module: $r.intp.global.Symbol = object o

scala> module.thisType
res7: $r.intp.global.Type = <noprefix>

scala> module.moduleClass.thisType
res8: $r.intp.global.Type = o.type
```

However, this can be had more directly:
```
scala> module.tpe
res12: $r.intp.global.Type = o.type
```

src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
```diff
@@ -214,27 +214,34 @@ trait MemberHandlers {
...
-    private def importableTargetMembers = importableMembers(targetType).toList
+    private def importableTargetMembers = importableMembers(exitingTyper(targetType)).toList
     // wildcard imports, e.g. import foo._
     private def selectorWild    = selectors filter (_.name == nme.USCOREkw)
-    // renamed imports, e.g. import foo.{ bar => baz }
-    private def selectorRenames = selectors map (_.rename) filterNot (_ == null)
+
+    // non-wildcard imports
+    private def individualSelectors = selectors filter { x =>
+      x.name != nme.USCOREkw && x.name != null && x.rename != nme.USCOREkw & x.rename != null
```
adriaanm, line 225: Could we encapsulate this stuff in `Contexts`, say? Looks like we already have `warnUnusedImports`, which compares against the more logical `nme.WILDCARD`, by the way.

adriaanm (changes requested):
There are some solid improvements in here, thank you! I know it's been too long, but could you take a look at a few of my suggestions? To avoid test failures, I'd suggest moving the fix for index of out bounds in names first, and squashing the WIP commit into your own.

adriaanm:
Ping! If you'd like this to be considered for 2.11.9, please let us know by the end of the week and we'll do our best to help you get this over the finish line.

adriaanm:
Please consider resubmitting for 2.12.2!

jasonxh:
@adriaanm Sorry I didn't get back to this in time. I'll rebase against 2.12.x and address your comments.

src/reflect/scala/reflect/internal/Types.scala
jasonxh: Will move to MemberHandlers

src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
jasonxh: Will use the simpler version.

src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
jasonxh: Nope it doesn't introduce new call sites. I just moved `exitingTyper` from both call sites to `importableMembers` itself.

src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
jasonxh: Will create some helpers in `Contexts`.

PR #5328, opened 2016-08-11
Improve log output of the `testAll` task

szeiger:
It’s a lot of code for little benefit but makes the output more useful
when test tasks fail. Unfortunately there doesn’t seem to be any way to
get the `summary` reported by a test framework at this point. The
arguments of `toTask` for InputTasks with applied arguments have also
been lost, so we keep track of the commands separately.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -786,11 +786,51 @@ lazy val root: Project = (project in file("."))
           doc in Compile in scalap
         ).result
       )).value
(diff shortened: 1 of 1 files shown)


adriaanm:
Every improvement to test output is worth it to me! User-friendliness to new contributors as well as keeping our own eyes from bleeding scanning through unnecessary punctuation!

adriaanm:
LGTM!

PR #5329, opened 2016-08-11
Remove doc-version was too long warning

adriaanm:
It would trigger 3091 times per test run (adding ca. 568k to the log file).
Use CSS to handle overflow instead.

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
@@ -154,19 +154,7 @@ trait EntityPage extends HtmlPage {
 
   def search =
     <div id="search">
-        <span id="doc-title">
-          {universe.settings.doctitle.value}
-          <span id="doc-version">
-          {
-            val version = universe.settings.docversion.value
-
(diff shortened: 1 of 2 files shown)


adriaanm:
Verified locally on chrome, so I'm going to merge. /cc @felixmulder in case my CSS hack should be tweaked

felixmulder:
I'm sure it's fine 😄  - did you check:
- [ ] Chrome, Android
- [ ] Chrome, iOS
- [ ] Safari, iOS

?

Also this was remedied by https://github.com/scala/scala/pull/5063 - something broken?

EDIT: keeping desktop + mobile in sync is usually what requires the most tuning.

adriaanm:
I can check the iOS variants on the nightly docs, but don't have access to android. I'm not sure why #5063 did not kick in -- maybe too many PRs simply were branched before it was merged. In any case, I'd rather not truncate the dev version in the html -- it seems CSS should be able to handle that just fine.

felixmulder:
Looks fine on droid 👍

adriaanm:
Thanks for checking!

soc:
Thanks a lot, Adriaan!

PR #5330, opened 2016-08-12
SI-9885 Don't return offset past EOF

som-snytt:
On bad line number, `lineToOffset` should not return
an offset past EOF (which was sentinel, internally).

JIRA: https://issues.scala-lang.org/browse/SI-9885

### DIFF
--- a/src/reflect/scala/reflect/internal/util/SourceFile.scala
+++ b/src/reflect/scala/reflect/internal/util/SourceFile.scala
@@ -154,18 +154,23 @@ class BatchSourceFile(val file : AbstractFile, content0: Array[Char]) extends So
     case _       => false
   }
 
-  def calculateLineIndices(cs: Array[Char]) = {
-    val buf = new ArrayBuffer[Int]
-    buf += 0
-    for (i <- 0 until cs.length) if (isAtEndOfLine(i)) buf += i + 1
-    buf += cs.length // sentinel, so that findLine below works smoother
-    buf.toArray
+  private lazy val lineIndices: Array[Int] = {
+    def calculateLineIndices(cs: Array[Char]) = {
+      val buf = new ArrayBuffer[Int]
+      buf += 0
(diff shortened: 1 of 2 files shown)


retronym:
LGTM

som-snytt:
That's pretty tricky. You're saying the weird test is to verify that Position's lines are 1-based, but `lineToOffset` is 0-based. That's unfortunate.

fommil:
turns out scala-refactoring was relying on the old behaviour.

src/reflect/scala/reflect/internal/util/SourceFile.scala
```diff
@@ -154,18 +154,23 @@ class BatchSourceFile(val file : AbstractFile, content0: Array[Char]) extends So
...
+      buf.toArray
+    }
+    calculateLineIndices(content)
   }
-  private lazy val lineIndices: Array[Int] = calculateLineIndices(content)
 
-  def lineToOffset(index : Int): Int = lineIndices(index)
+  def lineToOffset(index: Int): Int = {
+    val offset = lineIndices(index)
+    if (offset < length) offset else throw new IndexOutOfBoundsException(index.toString)
```
fommil, line 170: this doesn't seem to support asking for the first line of an empty file

fommil: I've minimised the case where we make this assumption in scala-refactoring

```scala
object Main
```

we ask for the source of line 0 and line 1.

som-snytt: The use case is to probe whether the next line is empty. Does the semantics of "line 1" change if the current line is not terminated by NL? Arguably there is an empty line 1 if there is a NL. Anyway, I proposed a different expression on the linked ticket which probes offsets for lines and then tests isEmpty on lineContent.

PR #5331, opened 2016-08-13
SD-128 fix override checks for default methods

adriaanm:
The check for inheriting two conflicting members was wrong for default
methods, leading to a missing error message.

We were also not issuing "needs `override' modifier" when overriding a
default method.

Removes two methods:
- `isDeferredOrJavaDefault` had a single use that is removed in this commit.
- `isDeferredNotJavaDefault` is redundant with `isDeferred`, because
  no default method has the `DEFERRED` flag:
  - For symbols originating in the classfile parser this was the case
    from day one: default methods don't receive the `DEFERRED` flag.
    Only abstract interface methods do, as they have the `JAVA_ACC_ABSTRACT`
    flag in bytecode, which the classfile parser translates to `DEFERRED`.
  - For symbols created by the Java source parser, we don't add the
    `DEFERRED` to default methods anymore since 373db1e.

Fixes scala/scala-dev#128

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -419,7 +419,7 @@ abstract class RefChecks extends Transform {
             overrideError("cannot be used here - classes can only override abstract types")
           } else if (other.isEffectivelyFinal) { // (1.2)
             overrideError("cannot override final member")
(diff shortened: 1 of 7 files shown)


adriaanm:
LGTM :-)

adriaanm:
This will need a few follow ups, as discovered by the community build: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/475/console

PR #5332, opened 2016-08-13
Fixes to Java source support in Scaladoc

retronym:
[Rebase of #5304 by @jodersky]

Some issues were discovered in my last contribution #5246 through the community build. This PR aims to solve them.

Posting this as a WIP so I can get a discussion going with the compiler maintainers, specifically @adriaanm and @VladUreche 

Issues to be fixed:
- [x] A doc comment in head position (such as a copyright header) will crash the compiler with an NPE, due to the way the javadoc analyzer is initialised. Fixed (thanks @adriaanm)
- [x] Documented java fields fail. Fixed in latest commit
- [x] Interfaces cause problems due to methods being placed into companion objects
- [x] Comments on constructors lead to double definition of constructor symbol
- [x] Ambiguous reference error to same-named objects defined in outer and inner classes (thanks @adriaanm)


retronym:
Review by @adriaanm

adriaanm:
LGTM

adriaanm:
I think we can squash on merge and thus only check the last commit.

retronym:
I've manually squashed and removed the `last-only` tag from the title.

jodersky:
@retronym , @adriaanm thanks for all your help! :tada: :tada: :tada:

PR #5336, opened 2016-08-13
Removing BooleanSort method

SakalyaDeshpande:
Do we really need Booleansort method??????

### DIFF
--- a/src/library/scala/util/Sorting.scala
+++ b/src/library/scala/util/Sorting.scala
@@ -213,25 +213,6 @@ object Sorting {
     }
   }
 
-  // Why would you even do this?
-  private def booleanSort(a: Array[Boolean]): Unit = {
-    var i = 0
-    var n = 0
-    while (i < a.length) {
-      if (!a(i)) n += 1
-      i += 1
-    }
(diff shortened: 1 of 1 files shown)


Ichoran:
It's a bit silly, but sometimes it's nicer to work at a higher level than maintaining counts.

Anyway, thank you for noticing that it seems a bit pointless, but all removals have to go through a deprecation process.  Because of the work it takes to make a deprecation and removal happen, I doubt it's worth it to remove this method that at least has the advantage of making the set complete.  If it is worth it in your opinion, file a bug report and perhaps issue a PR that has a decent deprecation message.

SethTisue:
> all removals have to go through a deprecation process

this one's `private`. but I actually agree that keeping it is better anyway. (and note that Jenkins failed because the code doesn't even compile)

Ichoran:
Technically, yes, we can remove private methods.  But in this case it doesn't compile because despite being private it changes the API of the public method that calls it (which wasn't fixed), and if that were to change it would need to be deprecated: it's not merely binary incompatible but source incompatible too.

PR #5337, opened 2016-08-13
SI-9717 Implicit class parameters in block in super class

diesalbla:
This is a WIP PR for discussion and advice, which addresses [Scala Issue 9717](https://issues.scala-lang.org/browse/SI-9717). 

The issue was that, when using a _block_ expressions as arguments in the call to the super-class constructor, any `implicit` parameter of the class was not visible within that block expression.
We add a `pos`  test file, that highlights the existing error. Note that, when using the implicit parameter _without_ the block, it was compiling fine. 

We have traced the issue back to the implicit resolution method. Here, there is an auxiliary method `nextOuter`, which is used when climbing up the context hierarchy, to skip some intermediate contexts. We guess that this was done as an optimization of implifit search, which may have unintentionally broken the semantics. Disabling this context filtering fixes the problem, so the compiler can now handle the positive test.

We do not know if this change could have a detrimental impact on compiler performance. Advice on this issues would be welcome.

This was done with the help of @ausmarton,  @flowcont, and @milessabin.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -608,7 +608,7 @@ trait Contexts { self: Analyzer =>
       // -- If the first statement is in the constructor, scopingCtx == (constructor definition)
       // -- Otherwise, scopingCtx == (the class which contains the constructor)
       val scopingCtx =
-        if (owner.isConstructor) nextEnclosing(c => !c.tree.isInstanceOf[Block])
+        if (owner.isConstructor) nextEnclosing(c => true)
         else this
 
       scopingCtx.outer
--- a/test/files/pos/t9717.flags
+++ b/test/files/pos/t9717.flags
@@ -1,0 +1,1 @@
+-Yno-predef
--- a/test/files/pos/t9717.scala
+++ b/test/files/pos/t9717.scala
@@ -1,0 +1,8 @@
+import scala.Predef.implicitly
+
+class A(val a: Int)
+class B(implicit F: Int) extends A( implicitly[Int] )
+class C(implicit F: Int) extends A( {
+  val v = implicitly[Int]
+  v
+})


milessabin:
The build failure looks odd and unrelated to the change in this PR. I suggest rebasing against the current 2.12.x which will trigger a rebuild.

adriaanm:
To rebase, I suggest `git pull --rebase https://github.com/scala/scala.git 2.12.x` while you have your `t9717_constructors_blocks_class_implicits` checked out.

adriaanm:
Also, no need for the rebuild command when you change your PR -- the automation will do this automatically.

som-snytt:
TIL `actual context` versus `conceptual context`.

diesalbla:
@milessabin The failure is that in the `PriorityQueueProxy` class, it is no longer finding in scope an `implicit Ordering[A]` despite the class having such an implicit parameter, so I am afraid that it could be related. I am trying to slice it down.

diesalbla:
I have reduced the failing code from the [PriorityQueue file]() to the following lines,  

``` scala
class Ann(implicit val i: Int)

abstract class Bob(implicit i: Int) extends Ann {
  def foo: Int
  def dee(): Bob = new Bob {
    def foo = 23
  }
}
```

Of course, the `2.12.x` version does not give an error with the reduced code. 

The interesting part is that in the super-class `Ann`,  the implicit parameter is declared as `val`, which makes it public, while in the sub-class it is not. This is like the PriorityQueue file, in the implicit `ord` parameter of the classes [`PriorityQueue`](https://github.com/scala/scala/blob/804133f60dd3c78909dc9e557e91b5c9923240ff/src/library/scala/collection/mutable/PriorityQueue.scala#L49) and [`PriorityQueueProxy`](https://github.com/scala/scala/blob/804133f60dd3c78909dc9e557e91b5c9923240ff/src/library/scala/collection/mutable/PriorityQueue.scala#L364).

In the reduced error example, the error disappears if **a)** you remove the `val` qualifier in the superclass, or **b)** rename the sub-class variable to another identifier, or **c)** remove the sub-class relation between classes `Ann` and `Bob`, or **d)** make the `i` parameter in `Bob` also a `val`, and add the `override` modifier .

_Edited:_ a more complete example includes an abstract method in `Bob`. Also, I add a few more fixes to the code example.

SethTisue:
Closing for inactivity. We can always reopen.

PR #5338, opened 2016-08-13
Add a new flag to support importing an extra user defined predef

andyscott:
This is an attempt to add a flag to support an "extra" user defined predef. The extra predef is always added when the flag is set, regardless of the `-Yno-predef` and `-Yno-imports` flags.

From my initial local testing, all tests pass. I've added one positive test for the new flag and two negative tests for behavior interplay with `-Yno-predef` and `-Yno-imports`.


andyscott:
Closing in favor of #5350.

PR #1521, opened 2012-10-23
Saving symbol lookup from typedIdent.

paulp:
Anticipating the appeals to performance, I offer this coarse measurement. My times are the good ones. Better measurements are welcome, but I convinced myself at least that I wasn't hurting anything.

```
% for i in 1 2 3; do time scalac3 -J-Xmx1g -d $(mktemp -dt foo) $(find /scratch/trunk5/src/library/scala/collection/mutable -name '*.scala') ; done
28.139 real, 68.872 user, 1.186 sys
30.963 real, 70.071 user, 1.260 sys
29.741 real, 73.389 user, 1.249 sys
% for i in 1 2 3; do time ./pack-whats-in-a-name/bin/scalac -J-Xmx1g -d $(mktemp -dt foo) $(find /scratch/trunk5/src/library/scala/collection/mutable -name '*.scala') ; done
28.129 real, 68.396 user, 1.176 sys
27.571 real, 67.318 user, 1.081 sys
27.240 real, 66.336 user, 1.046 sys
```

Review by @odersky, @adriaanm.

PR #5339, opened 2016-08-14
SI-2458 Import shadows def in other unit

som-snytt:
Fixes the implementation of point 4. from SLS chapter 2:

> Definitions made available by a package clause not in the
> compilation unit where the reference occurs have lowest precedence.

We had code to deal with this:

if (defSym.exists && impSym.exists) {
if (isPackageOwnedInDifferentUnit(defSym))
defSym = NoSymbol

But, in the `pos/t2458`, no `impSym` was around, due to filtering
in `depthOk`. The depth of the imported symbol was the same as the
depth of the defined symbol, which didn't pass:

imp.depth > symbolDepth

`depthOk` already had special handling for Java sources: if the
import is non-wildcard and depths are equal, retain the import.

This commit does the same thing for Scala sources, using the
`isPackageOwnedInDifferentUnit` as the criterion. It also makes
the check for the "imported ... is permanently hidden" warning
to avoid the spurious warning.

Fix by retronym
Excavation and preparation of bones for mounting by community

Fixes scala/bug#2458
Fixes scala/bug#9552


som-snytt:
This doesn't fix SI-9552. The import should be higher precedence than definition in other file. Maybe the package object fools a depth check?

```
package object p {
  type Foo = Int 
} 
```

and separate unit

```
package p {
  object Dingo { type Foo = String }
  import Dingo._
  trait Bippy  {
    // import Dingo._  // this version compiles
    def z: Foo
    z: String
  }
}
```

som-snytt:
Fixes the Java check for explicit import. Scala lets wildcard import hide foreign definition.

som-snytt:
Suffers from the broken build.

som-snytt:
Some documentation notes the bug and should be updated:

http://docs.scala-lang.org/tutorials/FAQ/finding-implicits.html

Blaisorblade:
@som-snytt what happened to this PR? Are the bugtracker entries up-to-date currently? (https://issues.scala-lang.org/browse/SI-2458 is closed, https://issues.scala-lang.org/browse/SI-9552 is open, and I'm confused).

som-snytt:
@Blaisorblade The old issue was closed prematurely when the spec was clarified; this PR was meant to be implementation which also nails down the newer issue (which also objects to the behavior as specified). I don't remember the build status, so I'll have to revisit it.

Blaisorblade:
@som-snytt Thanks — it sounds like all the open parts are tracked by the newer issue. I dread unfixed closed issues and forgotten bugs.

som-snytt:
Previous failure was due to cycle in `ProcessBuilder.scala` with wildcard import from companion.

There was also an import in scaladoc that was not to spec.

som-snytt:
This idiom now inoculates against accidental local dir named `util`:
```
package p

import scala._      // $ mkdir p/util
import util.Try     // was: error: object Try is not a member of package p.util

class C {
  def f = Try(42)
}
```
Too bad that's not more useful.

build.sbt
```diff
@@ -102,6 +102,9 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   autoScalaLibrary := false,
   // Avoid circular dependencies for scalaInstance (see https://github.com/sbt/sbt/issues/1872)
   managedScalaInstance := false,
+  //maxErrors := 500,
+  maxErrors := 5,
```
som-snytt: Meant to set that to zero.

src/compiler/scala/reflect/reify/Reifier.scala
```diff
@@ -69,7 +70,7 @@ abstract class Reifier extends States
           val rtree = pipeline(tree)
 
           val tpe = typer.packedType(tree, NoSymbol)
-          val ReifiedType(_, _, tpeSymtab, _, rtpe, tpeReificationIsConcrete) = `package`.reifyType(global)(typer, universe, mirror, tpe, concrete = false)
+          val ReifiedType(_, _, tpeSymtab, _, rtpe, tpeReificationIsConcrete) = reifyTypeGlobal(global)(typer, universe, mirror, tpe, concrete = false)
```
som-snytt: This was good, when you got packages and methods named reify.

src/compiler/scala/tools/nsc/reporters/EditReporter.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scala.collection.mutable
+import scala.reflect.internal.util.Position
+import scala.util.Properties
+import java.io.BufferedWriter
+
+/**
+ * This class implements a Reporter that stores erroneous files in a script for editing.
+ */
+class EditReporter(settings: Settings) extends Reporter {
```
som-snytt: If @lrytz was up for a reporters module, I'd contribute.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -984,6 +984,14 @@ trait Contexts { self: Analyzer =>
         }
       )
     }
+    // If ambiguous import resolves to the conflicting symbol, allow it.
+    // Typically, implicit import scala._ conflicts with definitions in scala packages.
+    // For that case, check if the root import dealiases to the definition.
+    private def resolveAmbiguity(name: Name, defSym: Symbol, impSym: Symbol, imp: ImportInfo): Boolean = {
+      defSym == impSym || imp.isRootImport && {
+        impSym.isAliasType && impSym.info.typeSymbol.fullName == defSym.fullName
```
som-snytt: @retronym Help. I didn't have time to improve this yet.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -1109,37 +1112,47 @@ trait Contexts { self: Analyzer =>
...
+       *  4) Definitions made available by a package clause, but not also defined in the same compilation unit
+       *     as the reference, have lowest precedence.
+       */
+
+      def depthOk(imp: ImportInfo) = imp.depth > symbolDepth || (
+        if (unit.isJava) imp.depth == symbolDepth && imp.isExplicitImport(name)
+        else if (depthCheckEnabled) foreignDefined
+        else imp.depth == symbolDepth && foreignDefined
       )
+      def depthCheckEnabled = true  // settings.isScala213
```
som-snytt: I was going to add the kill switch just for the last commit, where package-defined sym is amby with import in enclosing scope, the cause of many subtle edits, but I'd have to move those edits to 2.13.x.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
retronym: I'm having trouble paging in the context for this from my mental swap space... could you elaborate a little?

retronym:
Heads up for https://github.com/scala/scala/pull/6297 which is likely to collide with this.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
som-snytt: If root `import scala._` alias `Seq` collides with `Seq` defined in enclosing package, recognize that they are the same. There is related code for when two imports collide.

som-snytt:
Don't forget to update the spec deficiency about root imports and aliasing. https://gitter.im/scala/contributors?at=5a7b33b96117191e610894a1

EDIT: That was added in https://github.com/scala/scala/pull/5970 but note the limitation mentioned on gitter. The original paulpation was https://github.com/scala/scala/pull/1521. Not sure it's obvious that the spec includes the behavior here, where not only two imports, but any two name bindings that can be resolved to the same type are acceptable.

The spec needs definition of root imports per se and that they are like
```
import java.lang._ { import scala._ { import Predef._ { /* code */ }}}
```
and implementations may provide platform-specific `-Yno-imports` or `-predef` or whatever.

SethTisue:
rebase?

som-snytt:
My plan is to re-do this on 2.13 and then add `-Yimport` to make preamble imports tractable. Not sure exactly how it will work, but similarly to the leading `import Predef._` exclusion, a source file can opt out of the preamble; however, it seems to me that the syntax should align with ordinary binding.

If I compile with `-Yimport:my.Predef._` then `import my.Predef.{_ => _}` turns it off with no special magic.

If I compile with `-Yimport:my.Predef.any2stringadd`, then `import my.Predef.{_ => _}` does not hide? Unless the rule is construed that "level 4" includes the preamble imports. I don't think anyone considered that option, i.e., the preamble is not just stuck at the front of your source; it has level 4 "binding force" like bindings from a package clause that is not in the current unit.

som-snytt:
Rebased with the fix for root imports aka preamble imports. No longer impacts the world like an asteroid.

som-snytt:
Gitter is down? so I'll ask here whether this existing behavior seems too loose:

```
scala> class C { type InputStream = java.io.InputStream }
defined class C

scala> def f(c: C) = { import java.io._, c._ ; null.asInstanceOf[InputStream] }
f: (c: C)c.InputStream

scala> def f(c: C) = { import c._, java.io._ ; null.asInstanceOf[InputStream] }
f: (c: C)java.io.InputStream

```
I was considering whether the ambiguity in a test class
```
import p._ ; package p { S } ; package q { S }
```
deserves a pass. (Update: Yes, it does, even though the fix (moving import inside packages that need it) is easy. If the import goes unused, the user can ask for a warning.)

som-snytt:
I needed this today.


> imported `Configurator' is permanently hidden by definition of object Configurator in package apps

som-snytt:
I need this basically all the time.

> [warn] /Users/andrew/workspace/scala/test/benchmarks/src/main/scala/scala/tools/reporters/ReportingBenchmark.scala:9: imported `Reporter' is permanently hidden by definition of class Reporter in package reporters

adriaanm:
Moving this to the next 2.13 milestone.

som-snytt:
@adriaanm There's a forward port PR for 2.13; I have a `-Xsource:2.13` commit to push here; would you consider that?

adriaanm:
Sure, as a backport once it has landed in 2.13.x. I'm concerned with how thin our attention is spread across the PR queue these days, so I'd like to compact the queue. (It's nice to be so outnumbered, but also regret the delays this is causing.)

som-snytt:
Superseded for 2.13 by https://github.com/scala/scala/pull/6589

PR #5340, opened 2016-08-15
SI-8873: Annotations on case class constructor parameters

edmundnoble:
Fixes SI-8873, allowing annotations on case class constructor parameters by stripping them from the generated apply method's parameter list and keeping them on the generated fields. Not sure if this is exactly the approach to take, but I'm not sure which annotations should be removed exactly if not all of them.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Unapplies.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Unapplies.scala
@@ -128,7 +128,7 @@ trait Unapplies extends ast.TreeDSL {
    */
   def factoryMeth(mods: Modifiers, name: TermName, cdef: ClassDef): DefDef = {
     val tparams   = constrTparamsInvariant(cdef)
-    val cparamss  = constrParamss(cdef)
+    val cparamss  = mmap(constrParamss(cdef))(param => param.copy(mods = param.mods.copy(annotations = Nil)))
     def classtpe = classType(cdef, tparams)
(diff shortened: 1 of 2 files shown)


sjrd:
Something top think about and test: `@deprecatedName`

edmundnoble:
@sjrd Thank you for the suggestion. I'm not sure if apply's annotations should have a blacklist (just exclude volatile, synchronized, other builtin annotations which shouldn't be on function parameters) or a whitelist (only include deprecatedName, deprecated, and other builtin annotations which are okay for function parameters). I'm concerned about macro annotations in the whitelist case (though according to scalameta/paradise#16, macro annotations on parameters may be on their way to removal in dotty), so a blacklist is where I'm leaning at the moment.

sjrd:
The very existence of a blacklist and/or a whitelist suggests that there's something wrong. You will never be able to anticipate the requirements of all user-defined annotations.

IMO this issue is calling for more meta-annotations, e.g., `@meta.applyParam`.

edmundnoble:
I see now that @adriaanm plans to tackle this as part of the scala/scala/pull/5294, and has taken the same approach initially as I did. I am not sure if this issue meriting meta-annotations would make its inclusion overcomplicate his PR, but I agree that it should definitely be considered.

lrytz:
closing, as https://issues.scala-lang.org/browse/SI-8873 is fixed

PR #5342, opened 2016-08-15
Make SearchImpl a value class

zizhengtai:


zizhengtai:
Signed the CLA.

retronym:
Retargetted to 2.13.x as the change would not be binary compatible with 2.12.0.

lrytz:
rebased in https://github.com/scala/scala/pull/5599

PR #5102, opened 2016-04-15
SI-2712 Add support for partial unification of type constructors

milessabin:
An improvement to type inference for type constructors has been added, enabled by the `-Ypartial-unification` command line option (also enabled in `-Xexperimental` mode). This fixes [SI-2712](https://issues.scala-lang.org/browse/SI-2712) in the way suggested by Paul Chiusano in a [comment on the issue](https://issues.scala-lang.org/browse/SI-2712?focusedCommentId=61270). This has many benefits for libraries, such as Cats and Scalaz, which make extensive use of higher-kinded types.

With the feature enabled the following compiles,

``` scala
def foo[F[_], A](fa: F[A]): String = fa.toString
foo { x: Int => x * 2 }
```

with the type variable `F[t]` solved as `Int => t` and the type variable `A` solved as `Int`. Previously the compiler would have rejected this programme because,
- The type parameter `F` of `foo` is a type constructor with a single type parameter, also known as kind * -> *.
- The provided type, `Int => Int`, which is a synonym for `Function1[Int, Int]`, has an outer type constructor with two type parameters, or kind * -> * -> *.
- The type inference algorithm required that the kinds of type parameters and their corresponding type arguments must be the same.

Partial unification solves this problem by treating outer type constructors at call sites as partially applied. In the example above, the compiler does the equivalent of creating a local type alias with the correct kind and using that at the call site,

``` scala
type Partial[t] = Int => t 
foo[Partial, Int] { x: Int => x * 2 }
```

`Partial` has the same kind as the type argument `F` of `foo` and so this compiles successfully.

The implementation partially applies type constructors from left to right, leaving the rightmost type parameters free. This works well with binary type constructors with a natural right bias, such as `Either`, `Xor` in Cats and Scalaz's disjunction. For example a `map` operation defined with the signature illustrated below will naturally map over the value of the righthand type, corresponding to validity, whilst leaving the value of the lefthand type untouched,

``` scala
def map[F[_], A, B](fa: F[A])(f: A => B): F[B] = ...
val right: Either[String, Int] = Right(23)
val left: Either[String, Int] = Left("Invalid")
map(right)(_ + 1) // yields Right(24)
map(left)(_ + 1) // yields Left("Invalid")
```

An [article](https://gist.github.com/djspiewak/7a81a395c461fd3a09a6941d4cd040f2) by Daniel Spiewak expands on the consequences of this strategy.

[...]

PR #5343, opened 2016-08-15
SI-2712 Add support for higher order unification

milessabin:
This is a backport of #5102 to 2.11.x for 2.11.9.


sundarms:
is this availabe now? i can't see this in 2.11.11

PR #5346, opened 2016-08-16
SIP-23 Implementation of literal types

milessabin:
Co-Authored-By: George Leontiev folone@gmail.com
Co-Authored-By: Adriaan Moors adriaan@lightbend.com

Backport of #5310 to 2.11.x for 2.11.9.

### DIFF
--- a/src/reflect/scala/reflect/internal/Constants.scala
+++ b/src/reflect/scala/reflect/internal/Constants.scala
@@ -8,6 +8,7 @@ package reflect
 package internal
 
 import java.lang.Integer.toOctalString
+import scala.{ Symbol => SSymbol }
 import scala.annotation.switch
 
 trait Constants extends api.Constants {
@@ -30,6 +31,7 @@ trait Constants extends api.Constants {
   final val ClazzTag   = 12
   // For supporting java enumerations inside java annotations (see ClassfileParser)
   final val EnumTag    = 13
+  final val SSymbolTag = 14
 
   case class Constant(value: Any) extends ConstantApi {
     import java.lang.Double.doubleToRawLongBits
@@ -49,6 +51,7 @@ trait Constants extends api.Constants {
       case x: Char      => CharTag
       case x: Type      => ClazzTag
       case x: Symbol    => EnumTag
+      case x: SSymbol   => SSymbolTag
       case _            => throw new Error("bad constant value: " + value + " of class " + value.getClass)
     }
 
@@ -60,6 +63,8 @@ trait Constants extends api.Constants {
     def isFloatRange: Boolean = ByteTag <= tag && tag <= FloatTag
     def isNumeric: Boolean    = ByteTag <= tag && tag <= DoubleTag
     def isNonUnitAnyVal       = BooleanTag <= tag && tag <= DoubleTag
+    def isSymbol: Boolean     = tag == SSymbolTag
+    def isSuitableLiteralType = (BooleanTag <= tag && tag <= NullTag) || tag == SSymbolTag
     def isAnyVal              = UnitTag <= tag && tag <= DoubleTag
 
     def tpe: Type = tag match {
@@ -76,6 +81,7 @@ trait Constants extends api.Constants {
       case NullTag    => NullTpe
       case ClazzTag   => ClassType(typeValue)
       case EnumTag    => EnumType(symbolValue)
+      case SSymbolTag => SymbolTpe
     }
 
     /** We need the equals method to take account of tags as well as values.
@@ -242,6 +248,7 @@ trait Constants extends api.Constants {
     }
     def typeValue: Type     = value.asInstanceOf[Type]
     def symbolValue: Symbol = value.asInstanceOf[Symbol]
+    def scalaSymbolValue: SSymbol = value.asInstanceOf[SSymbol]
 
     /**
      * Consider two `NaN`s to be identical, despite non-equality
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -1816,22 +1816,45 @@ trait Types
   class PackageClassInfoType(decls: Scope, clazz: Symbol)
   extends ClassInfoType(List(), decls, clazz)
 
-  /** A class representing a constant type.
+  /** A class representing a constant type. Constant types are constant-folded during type checking.
+   *  To avoid constant folding, use the type returned by `deconst` instead.
    */
-  abstract case class ConstantType(value: Constant) extends SingletonType with ConstantTypeApi {
-    override def underlying: Type = value.tpe
-    assert(underlying.typeSymbol != UnitClass)
+  abstract class ConstantType extends SingletonType with ConstantTypeApi {
+    //assert(underlying.typeSymbol != UnitClass)
+    val value: Constant
+
     override def isTrivial: Boolean = true
-    override def deconst: Type = underlying.deconst
-    override def safeToString: String =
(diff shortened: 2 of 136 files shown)


milessabin:
Jenkins seems to be wedged? @SethTisue ... help?

SethTisue:
Jenkins seems to have a problem with your PRs in particular. Conspiracy theorists, start your engines!

Tracking at https://github.com/scala/scala-jenkins-infra/issues/185 — it might have to wait until RC1 is out

milessabin:
Rebased.

milessabin:
I'm not able to reproduce that one test failure locally. It's a run-output-differs failure ... is there any way to coax Jenkins into showing what the mismatched output was?

lrytz:
I thought the diff is supposed to be in the logs.. maybe @szeiger has a guess (https://scala-ci.typesafe.com/job/scala-2.11.x-validate-test/2076/consoleFull) - note that it's 2.11.x.

I managed to reproduce it locally with the same command that the CI uses (see the jenkins log):

```
➜  scala git:(844ecb1) sbt -sbt-version 0.13.12 -Dstarr.version=2.11.9-844ecb1-SNAPSHOT --warn 'setupValidateTest https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/' clean 'partest test/files/run/sip23-valueof.scala --update-check'
...
# starting 1 test in run
++ 1 - run/sip23-valueof.scala                   [updated check file]

➜  scala git:(844ecb1) ✗ cat test/files/run/sip23-valueof.check
warning: there were 7 inliner warnings; re-run with -Yinline-warnings for details
```

So the difference is that `setupValidateTest` enables the optimizer, and you get some inliner warnings:

```
➜  sandbox git:(844ecb1) ✗ qsc ../test/files/run/sip23-valueof.scala -optimise -Yinline-warnings -Yliteral-types
../test/files/run/sip23-valueof.scala:12: warning: At the end of the day, could not inline @inline-marked method implicitly
  implicitly[ValueOf['foo]]
            ^
...
7 warnings found
```

Not sure why the (2.11-)inliner fails here. I think your best solution is to add the following to the top of `sip23-valueof.scala`:

```
/*
 * filter: inliner warnings; re-run with
 */
```

milessabin:
Tweaked and squashed.

milessabin:
This last failure in the scaladoc tests is the same as the one which affected the 2.12.x version ... which is promising I think.

milessabin:
Backported scaladoc fix from #5310.

milessabin:
Added MiMa exclusions for `ValueOf`.

szeiger:
@lrytz Some newer improvements to the sbt build are only in the 2.12.x branch. This is what enables logs on 2.12.x: https://github.com/scala/scala/blob/2.12.x/project/ScriptCommands.scala#L28

lrytz:
thanks for checking!

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -1816,24 +1816,47 @@ trait Types
...
-  final class UniqueConstantType(value: Constant) extends ConstantType(value)
-
   object ConstantType extends ConstantTypeExtractor {
+    def apply(c: Constant): ConstantType = FoldableConstantType(c)
+    def unapply(tpe: ConstantType): Option[Constant] = Some(tpe.value)
+  }
+
+  abstract case class FoldableConstantType(value: Constant) extends ConstantType {
+    override def underlying: Type =
+      if (settings.YliteralTypes && value.isSuitableLiteralType) LiteralType(value) else value.tpe
```
adriaanm, line 1837: Let's not create a new `LiteralType` on every invocation of `underlying`

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -2789,18 +2814,28 @@ trait Types
...
+
       /* We can seed the type constraint with the type parameter
        * bounds as long as the types are concrete.  This should lower
        * the complexity of the search even if it doesn't improve
        * any results.
        */
-      if (propagateParameterBoundsToTypeVars) {
-        val exclude = bounds.isEmptyBounds || (bounds exists typeIsNonClassType)
+      val constr =
+        if ((propagateParameterBoundsToTypeVars || settings.YliteralTypes) && !isPastTyper) {
```
adriaanm, line 2824: Can we move the other side of the disjunction to propagateParameterBoundsToTypeVars's rhs? Should also dig in why we're getting the SOE -- is it because of existential type subtyping?

src/reflect/scala/reflect/internal/pickling/Translations.scala
```diff
@@ -33,7 +33,7 @@ trait Translations {
   def picklerTag(ref: AnyRef): Int = ref match {
     case tp: Type                       => picklerTag(tp)
     case sym: Symbol                    => picklerTag(sym)
-    case const: Constant                => LITERAL + const.tag
+    case const: Constant                => if (settings.YliteralTypes) assert (const.tag != NoTag) ; LITERAL + const.tag
```
adriaanm, line 36: let's drop this assert, verify invariant in ConstantType factory

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -223,7 +223,8 @@ trait TypeComparers {
       case PolyType(ps1, res1)        => tp2 match { case PolyType(ps2, res2)        => equalTypeParamsAndResult(ps1, res1, ps2, res2)       ; case _ => false }
       case ExistentialType(qs1, res1) => tp2 match { case ExistentialType(qs2, res2) => equalTypeParamsAndResult(qs1, res1, qs2, res2)       ; case _ => false }
       case ThisType(sym1)             => tp2 match { case ThisType(sym2)             => sym1 == sym2                                         ; case _ => false }
-      case ConstantType(c1)           => tp2 match { case ConstantType(c2)           => c1 == c2                                             ; case _ => false }
+      case FoldableConstantType(c1)   => tp2 match { case FoldableConstantType(c2)   => c1 == c2                                             ; case _ => false }
```
adriaanm, line 226: use a type test to avoid adding another case here? this is called a lot

src/reflect/scala/reflect/internal/Constants.scala
```diff
@@ -60,6 +63,8 @@ trait Constants extends api.Constants {
     def isFloatRange: Boolean = ByteTag <= tag && tag <= FloatTag
     def isNumeric: Boolean    = ByteTag <= tag && tag <= DoubleTag
     def isNonUnitAnyVal       = BooleanTag <= tag && tag <= DoubleTag
+    def isSymbol: Boolean     = tag == SSymbolTag
```
adriaanm, line 66: I don't think we should add support for Symbol constants. This will break older compilers, right?

adriaanm (changes requested):
I'm having second thoughts about merging this into the last 2.11.x release.

More precisely, I'm concerned about:
  - the overall risk of regression (changes to subtyping & inference) and performance (e.g., repeated checking of `settings.YliteralTypes`, reduced performance of pattern matches on ConstantType)
  - class files generated under the `-Yliteral-types` flags having constant types in unexpected places, which could break older compilers
  - ValueOf cannot be introduced in a minor version
  - the changes should be broken into smaller commits,  with detailed motivation

milessabin:
I'm happy to see this validated in practice in the TLS 2.11.x series.

adriaanm:
Agreed, sounds like a good differentiator for your 2.11 series. Happy New Year!

PR #5347, opened 2016-08-17
SI-6967 Primitive ClassTag unapply is deprecated

som-snytt:
Follow-up to add deprecation in 2.12.0.

### DIFF
--- a/src/library/scala/reflect/ClassTag.scala
+++ b/src/library/scala/reflect/ClassTag.scala
@@ -83,15 +83,24 @@ trait ClassTag[T] extends ClassManifestDeprecatedApis[T] with Equals with Serial
        ) Some(x.asInstanceOf[T])
     else None
 
-  // TODO: deprecate overloads in 2.12.0, remove in 2.13.0
+  // deprecate overloads in 2.12.0, remove in 2.13.0
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Byte)    : Option[T] = unapplyImpl(x, classOf[Byte])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Short)   : Option[T] = unapplyImpl(x, classOf[Short])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Char)    : Option[T] = unapplyImpl(x, classOf[Char])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Int)     : Option[T] = unapplyImpl(x, classOf[Int])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Long)    : Option[T] = unapplyImpl(x, classOf[Long])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Float)   : Option[T] = unapplyImpl(x, classOf[Float])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Double)  : Option[T] = unapplyImpl(x, classOf[Double])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Boolean) : Option[T] = unapplyImpl(x, classOf[Boolean])
+  @deprecated("Specialized unapply will be removed.", since = "2.12.0")
   def unapply(x: Unit)    : Option[T] = unapplyImpl(x, classOf[Unit])
 
   private[this] def unapplyImpl(x: Any, primitiveCls: java.lang.Class[_]): Option[T] =
--- a/test/files/run/t6318_primitives.check
+++ b/test/files/run/t6318_primitives.check
@@ -1,3 +1,4 @@
+warning: there were 18 deprecation warnings (since 2.12.0); re-run with -deprecation for details
 Checking if byte matches byte
 Some(1)
 Checking if byte matches short


sjrd:
Can we not remove them immediately? It should be a source-compatible change, as a call that would previously resolve to a specialized overload will simply now resolve to the generic overload, which does the same thing.

soc:
I agree with @sjrd. scala.reflect is not covered by any compatibility promises anyway.

adriaanm:
Thanks for catching that oversight, @som-snytt -- we should deprecate this in 2.12.x so we can remove in 2.13.

It would be great if telling people scala.reflect is experimental would actually be heard and give us the freedom to remove stuff, but in practice it does not...

I'd be less concerned if we were talking scala.reflect._internal_, which is explicitly excluded from mima (not to mention its name), but scala.reflect is not.

sjrd:
> It would be great if telling people scala.reflect is experimental would actually be heard and give us the freedom to remove stuff, but in practice it does not...

Sure. Although in this particular case, as I was saying, I think removing them is fine because it is still 100% source compatible, thanks to the overload that takes `Any` (which is not going anywhere).

In fact, I argue that it is even _more_ source compatible to remove them than deprecating them. If they are deprecated, I have to go through an artificial hoop with an ascription `ct.unapply(x: Any)` to bypass the deprecated and choose the one taking `Any`.

adriaanm:
Right, good point! That's convincing to me. As I'll be traveling, I'll let @szeiger decide here. 

(On my phone and buttons seem to jump around.)

som-snytt:
I don't have a strong opinion because of https://issues.scala-lang.org/browse/SI-9380 which goes the other way, deprecating the unapply(Any) for regex. People don't like messages, and they can also have subtle use cases.

The only behavior change is that primitives are boxed in user code.

Happy to type in whatever change is deemed OK, so long as, to quote Simon Pegg's Scotty, I'm not blamed for it.

sjrd:
SI-9380 is not comparable to this case. In `Regex.unapply`, _only_ `CharSequence`s can ever be successfully matched, so it makes sense to narrow the type down to `CharSequence`.

Here, values of arbitrary types can be matched (depending on what `Class` the `ClassTag` represents), so we need the overload with `Any`. The other ones can therefore be removed.

> The only behavior change is that primitives are boxed in user code.

Not even that. The primitives were already boxed when given to `unapplyImpl`.

szeiger:
There is a subtle difference in the implementation of the overloads vs the generic method:

``` scala
scala> class C[T](implicit val ct: scala.reflect.ClassTag[T]) { var x: T = _ }
defined class C

scala> class I extends C[Int] { println(ct.unapply(x)); println(ct.unapply(x: Any)) }
defined class I

scala> new I
Some(0)
None
res9: I = I@5c33f1a9
```

I'd argue that the overloads get it right and the right course of action is to only deprecate the overloads now and fix the generic implementation.

sjrd:
@szeiger Your example is mostly illustrating the problems with `= _` for generics instantiated to primitive values. It has very little to do with classtags in particular.

sjrd:
If you don't rely on `= _` to expose weird bugs, the behavior is completely fine:

``` scala
doeraene@lamppc37:~$ ~/opt/scala-2.11.8/bin/scala
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_40).

scala> class C[T](var x: T)(implicit val ct: scala.reflect.ClassTag[T])
defined class C

scala> class I extends C[Int](0) { println(ct.unapply(x)); println(ct.unapply(x: Any)) }
defined class I

scala> new I
Some(0)
Some(0)
res0: I = I@5ba23b66
```

szeiger:
I doubt that there's an easy way to avoid this issue (other than flat out disallowing default initializers for polymorphic AnyVals, which would limit the problem to uninitialized values). You could argue whether `unapply(x: Any)` should ever see a `null` in this case. Ideally it should not but it's probably unavoidable, too.

But my main point is that the overloads are not 100% source compatible, therefore I would prefer to deprecate them first and remove them in 2.13, unless there's a reason for a more urgent solution that I'm missing.

szeiger:
OTOH, since the generic implementation is the broken one, if we fix it now, we could also delete the overloads immediately.

sjrd:
You cannot fix the generic implementation for the `= _` case. But that's because `= _` is broken all over the place; not because the generic implementation is broken.

There are a lot of other things that are completely broken with `= _`.

szeiger:
I know, but unless someone has a silver bullet, we can only fix them one at a time. You're right that the current `unapply(x: Any)` is unfixable. `unapply(x: T)` could work but it would be pointless as a standalone method. As far as usage in the pattern matcher is concerned, I think the null checks for primitive types need to be removed for correctness. This does not look right to me:

``` scala
scala> implicitly[ClassTag[Int]].unapply(null: Any)
res14: Option[Int] = None

scala> (null: Any).asInstanceOf[Int]
res15: Int = 0
```

If the pattern matcher already knows that the types are incompatible, there's no reason to call `unapply`. The only case where the change is relevant is a match `x: T` where `T <:< AnyVal` and `x.type <:< Any` (but not `x.type <:< AnyRef`). I would expect such a test to succeed for `x == null` because a cast also succeeds.

sjrd:
> I know, but unless someone has a silver bullet, we can only fix them one at a time.

We shouldn't. Every time you think you've fixed an `= _` bug, you can make it resurface with one more indirection of generics. For example, in this case:

``` scala
scala> class C[T](implicit val ct: scala.reflect.ClassTag[T]) { var x: T = _ }
defined class C

scala> class I extends C[Int] { println(ct.unapply(x)); println(ct.unapply(x: Any)) }
defined class I

scala> new I
Some(0)
None
res0: I = I@52c3cb31

scala> class J[T](implicit ct: scala.reflect.ClassTag[T]) extends C[T] { println(ct.unapply(x)); println(ct.unapply(x: Any)) }
defined class J

scala> new J[Int]
None
None
res1: J[Int] = J@331acdad
```

The `= _` is so fundamentally broken that any decision making based on the behavior of `= _` is making Scala more complex for no reason whatsoever.

som-snytt:
Interesting discussion. I wonder if there's an argument to neither deprecate nor remove.

My previous reference to the regex.unapply debacle was just to say that use cases are subtle (for instance, reliance on that unapplySeq producing a List and not a Vector). It's worth assuming that edge cases like unapply(null) are exploited.

So the change pushes the bytecodes for boxing to user code (instead of in unapply). Maybe that only matters if you're counting bytecodes; but people do count them.

Maybe there's no harm in the status quo. Or, I could head over to StackOverflow. Actually, in this form it's still a puzzler (in the sense of not unreasonable to expect the obvious):

```
scala> class C[A] { var a: A = _
     | def g(seq: Seq[Any])(implicit tag: reflect.ClassTag[A]): Seq[A] = seq.collect { case tag(t) => t } }
defined class C

scala> val c = new C[Double]
c: C[Double] = C@9accff0

scala> c.g(List(1, c.a, "s"))
res7: Seq[Double] = List()

scala> c.a
res8: Double = 0.0
```

Also, I'm not sure I understand that `null` should match for any type test. Maybe I misunderstood the question.

```
scala> def f[A: ClassTag](a: A) = a match { case _: A => }
f: [A](a: A)(implicit evidence$1: scala.reflect.ClassTag[A])Unit

scala> f(null: String)
scala.MatchError: null
  at .f(<console>:12)
  ... 32 elided

scala> (null: String) match { case _: String => }
scala.MatchError: null
  ... 32 elided
```

szeiger:
@som-snytt Not matching null for primitives because it doesn't match for AnyRefs is a good point. The difference for AnyVal is that null is not a legal value, so when you get a null you know that it must be an artifact of the compiler leaving a boxed AnyVal uninitialized. But since you don't know what the type of that AnyVal should be (or if it's an AnyVal at all, and not a legitimate null AnyRef), there's really no correct answer after all. Not matching nulls is at least consistent between AnyVal and AnyRef.

In light of this, I'd say remove the overloads. It may not get much better, but it's simpler and at least no worse than the current situation.

What case are you thinking of where this would require boxing in user code? If you match on an unboxed value, you don't have to do a type check in the first place. You only call `unapply` if you have an already boxed value.

adriaanm:
Let's try to make a final call here today, so we can have all RC1 PRs merged tomorrow.

sjrd:
@adriaanm From @szeiger's last comment, it seems to me the consensus would be to remove the overloads.

> In light of this, I'd say remove the overloads. It may not get much better, but it's simpler and at least no worse than the current situation.

som-snytt:
Superseded.

PR #5348, opened 2016-08-17
SI-9841 Regression test for init SO

som-snytt:
Verifies example behavior in ticket.

JIRA: https://issues.scala-lang.org/browse/SI-9841


adriaanm:
Could you clarify that this documents the (regretted) status quo, and not the desired outcome?

som-snytt:
OK. But I thought regression test means we regretted it ever happened.

SethTisue:
LGTM

PR #5349, opened 2016-08-17
SI-9841 Progression test for SO on init

som-snytt:
Test as reported in ticket. Works on 2.12.

JIRA: https://issues.scala-lang.org/browse/SI-9841


adriaanm:
LGTM! Thanks @som-snytt!

PR #5350, opened 2016-08-18
Add flags -Ysysdef and -Ypredef to configure root imports [ci: last-only]

andyscott:
This adds a `-Ysysdef` and `-Ypredef` flag to give you full control over global imports for Scala compilation units.

_Background information:_

By default, scalac imports `java.lang._`, `scala._`, and `scala.Predef._` into the namespace of Scala compilation units.

As it stands, we have two existing flags that adjust the behavior of predefined imports.
- `-Yno-imports` disables all of these.
- `-Yno-predef` disables `scala.Predef._`.

There are a few other nuances to the behavior of imports. The most important behavior is that `scala.Predef._` is unimported/evicted if there are any direct leading imports for anything within `scala.Predef`. For example, an import to `scala.Predef.DummyImplicit` will cause everything else normally imported by predef to be unimported.

#### Changes
- Introduces the notion of "sysdefs" and "predefs" for Scala compilation units. Sysdefs are global imports that are imported at the top of all Scala compilation units _under all circumstances_. Predefs are imported after sysdefs and can be masked by top level imports in the source file (to match the existing behavior of directly importing from `scala.Predef`).
- Adds `-Ysysdef` which controls the global sysdef imports for all Scala compilation units. The default value is set to `java.lang._; scala._`.
- Added `-Ypredef` which controls the predef imports for Scala compilation units. The default value is set to `scala.Predef._`.

I'd be great to get any feedback. I would also to back port this to 2.11.


andyscott:
It looks like it would be reasonable to implement support for importing individual symbols. For example, `-Ypredef:java.lang._,scala.{ any2stringAdd => _, _},scala.Predef._`. Should I push this a little further?

tpolecat:
I would love to see this feature, with fine-grained support as described above.

som-snytt:
That was my thinking as well. There might be considerations about symbol init which makes it not work as simply as an import? The idiom would be `-Ypredef:java.lang._,scala._,scala.Predef.{any2stringAdd => _, _}`.

Does the refactor of that part work the same? I remember the import has to be at the top of the unit, because otherwise the semantics looks like unimporting. The REPL supports it at the top. The other part is don't import our own package.

There are postSetHooks for settings which you can use to wire -Ypredef to -Yno-imports and -Yno-predef. I guess -Yno-imports resets to empty, -Yno-predef strips Predef. (As you have it in the code.) `noimports andThen (flag => if (flag) Ypredef.value = Nil)`. `noassertions` uses that.

src/compiler/scala/tools/nsc/settings/AbsScalaSettings.scala
```diff
@@ -33,7 +33,7 @@ trait AbsScalaSettings {
   def ChoiceSetting(name: String, helpArg: String, descr: String, choices: List[String], default: String): ChoiceSetting
   def ChoiceSettingForcedDefault(name: String, helpArg: String, descr: String, choices: List[String], default: String): ChoiceSetting
   def IntSetting(name: String, descr: String, default: Int, range: Option[(Int, Int)], parser: String => Option[Int]): IntSetting
-  def MultiStringSetting(name: String, helpArg: String, descr: String): MultiStringSetting
+  def MultiStringSetting(name: String, helpArg: String, descr: String, default: List[String] = Nil): MultiStringSetting
```
som-snytt: This doesn't look right to me. Compare `MultiChoiceSetting` where default arg is on MutableSettings.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -172,8 +172,9 @@ trait ScalaSettings extends AbsScalaSettings
   val log             = PhasesSetting     ("-Ylog", "Log operations during")
   val Ylogcp          = BooleanSetting    ("-Ylog-classpath", "Output information about what classpath is being applied.")
   val Ynogenericsig   = BooleanSetting    ("-Yno-generic-signatures", "Suppress generation of generic signatures for Java.")
-  val noimports       = BooleanSetting    ("-Yno-imports", "Compile without importing scala.*, java.lang.*, or Predef.")
-  val nopredef        = BooleanSetting    ("-Yno-predef", "Compile without importing Predef.")
+  val Ypredef         = MultiStringSetting("-Ypredef", "package", "Supply a list of packages/objects to be imported into the global scope", List("java.lang", "scala", "scala.Predef"))
+  val noimports       = BooleanSetting    ("-Yno-imports", "Deprecated. Compile without importing scala.*, java.lang.*, or Predef.")
+  val nopredef        = BooleanSetting    ("-Yno-predef", "Deprecated. Compile without importing Predef.")
```
som-snytt: Use `withDeprecationMessage`.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -76,26 +70,45 @@ trait Contexts { self: Analyzer =>
...
+        if (settings.nopredef) error("Deprecated flag -Yno-predef isn't supported with -Ypredef")
+        if (settings.noimports) error("Deprecated flag -Yno-imports isn't supported with -Ypredef")
+        settings.Ypredef.value
+      }
+      else if (settings.noimports) Nil
+      else if (settings.nopredef) settings.Ypredef.value.filterNot(_ == "scala.Predef")
+      else settings.Ypredef.value
+
+    paths.map(path => rootMirror.getPackage(path))
+  }
```
som-snytt: I think it's easier just to use Ypredef here. If you want to enforce exclusive use, you can move that into withPostSetHook.

milessabin:
I'm very much in favour of this. However if deprecation warnings are added for existing compiler flags I'd like to see some mechanism to disable them. Maybe flag deprecation warnings could also be covered by `-deprecation`?

dwijnand:
@milessabin 

> I'm very much in favour of this. However if deprecation warnings are added for existing compiler flags I'd like to see some mechanism to disable them. Maybe flag deprecation warnings could also be covered by -deprecation?

If that's for cross-building sbt reasons, you can make compiler flags conditional to the scala version. I had to do that with, for example, 2.12 renaming `-Yopt` to `-opt`.

milessabin:
@dwijnand no, not really. If a warning really shouldn't be ignored it should be an error. If it is permissible to ignore it, at least temporarily, then it should be possible to note the warning, disable it and return a build uncluttered with noise. People don't ignore warnings for fun ... typically there's some constraint that prevents them from fixing them immediately ... that being so we should avoid imposing additional burdens on them.

som-snytt:
TIL

```
$ scalam -Ymacro-no-expand
Welcome to Scala 2.12.0-M5 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_92).

scala> [init] warning: there was one deprecation warning; re-run with -deprecation for details


scala> 42
res0: Int = 42

scala> :replay -Ymacro-no-expand -deprecation
Replaying: 42
[init] warning: -Ymacro-no-expand is deprecated: Use -Ymacro-expand:none
warning: -Ymacro-no-expand is deprecated: Use -Ymacro-expand:none
res0: Int = 42


scala> :replay -Ymacro-no-expand -deprecation -nowarn
Replaying: 42
res0: Int = 42
```

I think there's a proposal for warnings to be typed, so it's easier to filter them (in a reporter). Then maybe they'll do warning suppression of a sort.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -312,6 +321,11 @@ trait ScalaSettings extends AbsScalaSettings
 
   private def replacedByYpredef = "This flag is slated for removal. Its behavior can be achieved with the more generic -Ypredef flag."
 
+  private def requireExclusiveUse(peer: Setting): (Setting) => Unit = { s =>
+    if (peer.isSetByUser)
+      errorFn(s"Flag ${s.name} can not be used in conjunction with flag ${peer.name}")
+  }
+
```
som-snytt, line 337: This is a bit odd. Is there a word for currying in this way? I think this is the same:

```
def exclusively(peer: Setting)(setting: Setting): Unit = if (peer.isSet) fail(setting)
```

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -72,11 +72,7 @@ trait Contexts { self: Analyzer =>
...
     val paths =
-      if (settings.Ypredef.isSetByUser) {
-        if (settings.nopredef) error("Deprecated flag -Yno-predef isn't supported with -Ypredef")
-        if (settings.noimports) error("Deprecated flag -Yno-imports isn't supported with -Ypredef")
-        settings.Ypredef.value
-      }
+      if (settings.Ypredef.isSetByUser) settings.Ypredef.value
       else if (settings.noimports) Nil
       else if (settings.nopredef) settings.Ypredef.value.filterNot(_ == "scala.Predef")
       else settings.Ypredef.value
```
som-snytt: You may not want to go this way, but I was thinking this could be `val paths = settings.Ypredef.value` if setting noimports or nopredef updated Ypredef in the hook. Then the real code doesn't care how Ypredef was composed.

andyscott: Aah! Gotcha.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -76,31 +70,55 @@ trait Contexts { self: Analyzer =>
...
-      //         message for this ticket for an example.
-      debuglog("Omitted import of Predef._ for " + unit)
-      RootImports.javaAndScalaList
+    if (unit.isJava) gen.mkWildcardImport(JavaLangPackage) :: Nil
+    else {
+
+      /** Cleanup import `imp` such that any selectors covered by
+        * any imports in `parents` are removed.
+        */
+      def dedupe(imp: Import, parents: List[Import]): Option[Import] = {
```
andyscott: I am unsure of what the specific behavior should be for skipping predef imports. Also, what's the appropriate way to test this logic?

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -76,31 +70,55 @@ trait Contexts { self: Analyzer =>
...
 
-  /** List of symbols to import from in a root context.  Typically that
-   *  is `java.lang`, `scala`, and [[scala.Predef]], in that order.  Exceptions:
+  lazy val generalRootImports: List[Import] =
+    settings.Ypredef.value
+      .map(global.newUnitParser(_, "<flag-Ypredef>").parseRule(_.importExpr()))
+      .collect { case imp: Import =>
+        // I can't seem to figure out how to use the freshly parsed tree...
+        // Instead we get to recreate it
+        gen.mkImportFromSelector(
```
andyscott: See the inline code comments. The parsed tree seems to be lacking symbols on the import's expr tree. I couldn't figure out how to assign them without throwing errors.

andyscott:
I made an attempt at handling import selectors-- some help needed for this one, so I've added a few comments to my own code.

andyscott:
Getting the `neg/predef-masking` test to pass is going to be "fun" (read: challenging), regardless of support for import selectors.

For this test to pass, hard coded logic for `Predef` needs to be generalized to handle all predefs.

``` scala
  // Belongs in TreeInfo but then I can't reach it from Printers.
  def isReferenceToScalaMember(t: Tree, Id: Name) = t match {
    case Ident(Id)                                          => true
    case Select(Ident(nme.scala_), Id)                      => true
    case Select(Select(Ident(nme.ROOTPKG), nme.scala_), Id) => true
    case _                                                  => false
  }
```

To make this easier, I may put a constraint that all predef imports would automatically be attached to `_root_`. This means `-Ypredef:scala._;collection._;immutable._;` wouldn't behave but `-Ypredef:scala._;scala.collection._;scala.collection.immutable._` would be fine.

soc:
@andyscott While you are on it, can you add something to treat named imports as wildcard imports?

andyscott:
@soc I'm not familiar with what you mean by start trees. Can you explain?

soc:
@andyscott  Sorry, I was still editing => wildcard imports.

andyscott:
One issue I am still having: Many of the tests, such as `test/files/run/dynamic-updateDynamic.scala` are failing with a diff error due to the following warning output:

```
-newSettings: allArgs = List(-d, dynamic-updateDynamic-run.obj, -usejavacp, -Xprint-pos, -Xprint:typer, -Yrangepos, -Ystop-after:typer, -d, dynamic-updateDynamic-run.obj)
```

I noticed a few existing tests extending `scala.tools.partest.DirectTest` override the debug flag to suppress this warning. Do I need to do this for all of the failing tests or is there a better solution for  this?

som-snytt:
I guess you're talking about your local test output, not the jenkins failure 

```
scala.reflect.internal.MissingRequirementError: object <0:9>java.lang in compiler mirror not found.
```

That log output is when `partest.debug`, which is not on for builds, right?

I haven't followed the last changes, so I don't understand the semicolon separator change.

andyscott:
I made the delimiter configurable for multi string settings, since commas might be present in the import selector syntax. `-Ypredef:scala._;scala.Predef.{any2stringadd,println}`.

I believe I fixed that particular error in one of the last commits. It had to do with calling `toString` on an Import's expr. It looks like it was attaching some range information in certain cases.

som-snytt:
Ah. I wonder if it isn't better just to make the option parser aware of brackets.

Needless to add, if we're imitating import syntax, `import concurrent._, ExecCtx.Implicits.global` and so on. It's commas all the way down.

Or, it's not a multistring at all, but a string. When you receive it, prepend "import" and parse it and typecheck it.

andyscott:
I'm okay with any of those approaches. The custom delimiter seems to work fine, but if you'd prefer that I go a different route let me know and I can switch it over.

som-snytt:
Just noticed `-language:experimental._` doesn't actually work. I remember the options are hardcoded (because the package isn't loaded yet) but it would be nice if all these options could be handled uniformly. Then `-language:foo` would be `-Ypredef:language.foo` and so on.

adriaanm:
This is very cool, and I support developing this, but we can't commit to shipping this with 2.12.1, so moving to the WIP milestone for now.

andyscott:
@adriaanm No problem! I have been slow to wrap this up. I will find some time over the upcoming holidays to put together a proper solution in a rebased PR for you all to review.

adriaanm:
Sounds good, thanks!

andyscott:
I simplified the solution and also split the one new flag into two new flags to accommodate for the different behavior of different kinds of global imports. The main PR description has mostly been updated to describe this.

Also, the first commit in the rebased PR intentionally fails. I wanted to run the new implementation against all the existing tests. Everything seems to work and the two tests that fail only fail because I completely disabled the existing "no predef/import" flags and I didn't bother to update the tests. In the next PR, I feature flagged the new behavior to make sure all existing flags work exactly as they should.

I'm sure I need to add more exhaustive tests, but I figure might be in a good place for a review.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -186,7 +186,16 @@ trait ScalaSettings extends AbsScalaSettings
   val Ylogcp          = BooleanSetting    ("-Ylog-classpath", "Output information about what classpath is being applied.")
   val Ynogenericsig   = BooleanSetting    ("-Yno-generic-signatures", "Suppress generation of generic signatures for Java.")
   val noimports       = BooleanSetting    ("-Yno-imports", "Compile without importing scala.*, java.lang.*, or Predef.")
+    .withPostSetHook(postSetExclusive(Ysysdef, Ypredef))
   val nopredef        = BooleanSetting    ("-Yno-predef", "Compile without importing Predef.")
+    .withPostSetHook(postSetExclusive(Ysysdef, Ypredef))
+  val Ysysdef: StringSetting =
+                        StringSetting     ("-Ysysdef", "imports", "Supply a list of packages/objects to be imported into the global scope as system imports", "java.lang._, scala._")
```
lrytz: the messages here and below could mention "comma-separated" and show the default value, and say that `_` means nothing.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -413,6 +414,38 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
 
   import syntaxAnalyzer.{ UnitScanner, UnitParser, JavaUnitParser }
 
+  private def resolveSettingImports(settingValue: String, settingName: String): List[Import] = {
```
lrytz: could take a `StringSetting` as argument

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -99,8 +99,68 @@ trait Contexts { self: Analyzer =>
   }
 
 
+  // Collects the leading imports in a given unit body.
+  //
+  // This method should return the list of imports that's
+  // checked in `noPredefImportImport`
```
lrytz: what is `noPredefImportImport `?

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -99,8 +99,68 @@ trait Contexts { self: Analyzer =>
...
+    }
+  }
+
+  def localPredefImports(unit: CompilationUnit): List[Import] = {
+    val leadingImports = collectLeadingImports(unit.body)
+    val leadingImportsContext =
+      globalRootImports.foldLeft(startContext)((c, imp) => c.make(imp))
+
+    val (_, rootImports) = leadingImports
+      .foldLeft((leadingImportsContext, globalPredefImports)) { (acc, imp) =>
```
lrytz: could deconstruct (an name) the parts with `foldLeft(..) { case ((contxtImports, predefImports), leadingImport) => ... }` (with better names)

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -99,8 +99,68 @@ trait Contexts { self: Analyzer =>
...
+      case head :: tail =>
+        try {
+          val start = context.lookupSymbol(head, _ => true).symbol
+          tail.foldLeft(start)((parent, name) => parent.info member name)
+        } catch { case _: Throwable => NoSymbol }
+      case Nil => NoSymbol
+    }
+  }
+
+  def localPredefImports(unit: CompilationUnit): List[Import] = {
```
lrytz, line 145: a doc comment on this method would be really helpful

lrytz: IUC, this implements the magic that you point out in the PR description: importing `Predef.X` causes the `Predef._` import to be removed. That's why this method only looks at the Import tree's `expr`, but not at the `selectors`, right? This is not very obvious, should be documented.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -99,8 +99,68 @@ trait Contexts { self: Analyzer =>
...
+    val leadingImports = collectLeadingImports(unit.body)
+    val leadingImportsContext =
+      globalRootImports.foldLeft(startContext)((c, imp) => c.make(imp))
```
lrytz: the naming here is a bit confusing: the name `leadingImportsContext` sounds related to `leadingImports`, but it's not.

test/files/presentation/ypredef-completion-implicit-chained/src/Completions.scala
```diff
@@ -0,0 +1,12 @@
+package test
+
+import scala.Predef.DummyImplicit // turn off other predef implicits for a cleaner .check file.
```
lrytz: Add a comment explaining that `Predef.x` eliminates the `Predef._` import - not really obvious :)

test/files/presentation/ypredef-completion-implicit-chained/src/Completions.scala
```diff
@@ -0,0 +1,12 @@
...
+
+object O {
+  def map(x: Int => Int)(implicit a: DummyImplicit): O.type = this
+  val prefix123 : Int = 0
+}
+
+class Foo {
+  O.map(x => x)./*!*/ // we want the presentation compiler to apply the implicit argument list.
```
lrytz, line 14: Why are you actually adding a `/presentation` test in this PR? Just for coverage? Maybe add a comment about that, as this PR doesn't change anything specific to the presentation compiler.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -99,8 +99,68 @@ trait Contexts { self: Analyzer =>
...
+  }
+
   def rootContext(unit: CompilationUnit, tree: Tree = EmptyTree, throwing: Boolean = false, checking: Boolean = false): Context = {
-    val rootImportsContext = (startContext /: rootImports(unit))((c, sym) => c.make(gen.mkWildcardImport(sym)))
+
+    val localRootImports =
+      if (settings.Ysysdef.isSetByUser || settings.Ypredef.isSetByUser) {
+        // 'new' feature flagged implementation
+        if (unit.isJava) RootImports.javaList.map(gen.mkWildcardImport)
+        else (globalSysdefImports ::: localPredefImports(unit))
```
lrytz: unnecessary parentheses

lrytz (commented):
This looks very good, I think it's very close!

Two more test suggestions:
  - The current behavior (without `-Ysysdef` / `-Ypredef`) of an explicit `Predef.X` import eliminating the `Predef._` - if such a test doesn't exist already
  - The fact that `-Ysysdef` / `-Ypredef` arguments need to be fully qualified imports (no `collection.mutable`)

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
andyscott: Whoops! Looks like a typo from some refactoring I did. It should be `noPredefImportForUnit`.

test/files/presentation/ypredef-completion-implicit-chained/src/Completions.scala
andyscott: This test was a blocker for the initial iterations/implementation of `-Ypredef`, so I wanted to ensure that it passed when I finalized the approach. Would it be best to remove the test (since it passes now) or amend it with a comment?

src/compiler/scala/tools/nsc/Global.scala
andyscott: Got it. I have to use a type projection `ScalaSettings#StringSetting` instead of `settings.StringSetting`, otherwise scalac laments about it not being a stable identifier.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
andyscott: Yep! I took a stab at documenting this and I'll push it soon.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
andyscott: I'll name them more appropriately and add a comment to clarify.

andyscott:
> Two more test suggestions:
> - The current behavior (without -Ysysdef / -Ypredef) of an explicit Predef.X import eliminating the Predef._ - if such a test doesn't exist already

This is already "covered" (but not documented clearly) by `test/files/presentation/completion-implicit-chained/`. I went ahead and updated the documentation for this test to be consistent with the updated documentation for the parallel test for `-Ypredef`.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -190,10 +190,10 @@ trait ScalaSettings extends AbsScalaSettings
   val nopredef        = BooleanSetting    ("-Yno-predef", "Compile without importing Predef.")
     .withPostSetHook(postSetExclusive(Ysysdef, Ypredef))
   val Ysysdef: StringSetting =
-                        StringSetting     ("-Ysysdef", "imports", "Supply a list of packages/objects to be imported into the global scope as system imports", "java.lang._, scala._")
+                        StringSetting     ("-Ysysdef", "imports", "Supply a comma separated list of fully qualified packages/objects to be imported into the global scope as system imports. Use an empty string or the string \"_\" to configure for no system imports.", "java.lang._, scala._")
```
lrytz: Now it's a long line, most other `-Y` descriptions fit on a single terminal line.. Something like:

"Comma-separated list of fully qualified system imports, added to every source file. `_` for none."

"Comma-separated list of fully qualified Predef imports, added if not clashing with top-level imports. `_` for none."

lrytz (commented):
LGTM otherwise.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
som-snytt: Maybe time to make it easy for options to have extended help, like `-opt:help`.  `StringSetting(...) withVerboseHelp "Long explanations..."`. That could enable `isHelping` and set `help` string. Not sure how to ask for that help in general. Maybe `scala -Ysysdef` by itself? So, `isHelping` if the StringSetting has missing argument? Not as explicit as `-opt:help`.

som-snytt: Or `-help -Ysysdef` should do the obvious thing and check `isHelping` for Ysysdef, which would forward to `settings.help`.

andyscott:
@lrytz Fixed. Would you like me to rebase and/or squash any commits?
/cc @milessabin (re: TLS 2.12.1)

lrytz:
@andyscott thanks! This PR needs to be rebased on top of current 2.12.x, there's a conflict.

Leaving a second review to @adriaanm.

adriaanm:
I'll take a look. When you rebase, I'd suggest squashing the commits.

lrytz:
My high-level questions:
  - Some duplicated logic: if none of `-Ypredef` / `-Ysysdef` flags is present, vs. when one of them is (e.g., what to do `if (unit.isJava)`). The new code is much more complex as it needs to do lookups, while in the existing, default case we can just refer to the symbols in definitions (see `RootImports`, e.g., `JavaLangPackage :: ScalaPackage :: PredefModule :: Nil`). So it probably makes sense to keep both.
  - It's maybe a bit of a historical artifact / implementation detail to have the split between predef and sysdef?

andyscott:
Rebased and squashed.

@lrytz 
- I kept the duplicate logic to ensure perfect compatibility with the existing predef behavior. I figure, after a period of transition, we we could remove the original logic.
- The split behavior of sysdef/predef may be a historical artifact. This PR, at the very least, documents it a bit in the implementation. If the sysdef/predef behavior is simplified, it should be pretty straightforward to update this code to match it.

adriaanm:
I don't think we should let history impede good design, especially when we're talking about unsupported/experimental features such as `-X`/`-Y` flags.

I much prefer your original approach where you can just specify the default imports, where we could have three options: `default`, `none`, and arbitrary list of imports. How about calling it `-Yimports`, and dropping the old flags?

andyscott:
Correct me if I am wrong, but what I'm hearing is:

- Drop `-Yno-imports` and `-Yno-predef` in favor of the new flag(s)
- Rename/unify `-Ysysdef` and `-Ypredef` to `-Yimports`

This sounds great to me-- but how do you want to unify the behavior of "sysdef" vs "predef" style imports?

adriaanm:
Yes, that's what I meant. Could you help me understand the issue with 
> unify the behavior of "sysdef" vs "predef" style imports?

adriaanm:
I mean, your table in the description lists the values for `-Yimports` that correspond to the sysdef/predef options, right? If you specify any imports, you're going to have to be exhaustive. Is that what you're referring to?

andyscott:
The default root imports are `java.lang._`, `scala._`, and `scala.Predef._`. In the vast majority of cases, they all behave "the same"-- just like regular leading imports at the top of a file.

`scala.Predef._` gets special treatment. Leading imports into the import space of `scala.Predef._` will cause the import to `scala.Predef._` to be excluded from the compilation unit. So importing `scala.Predef.DummyImplicit` at the top of your file will prevent the import to `scala.Predef._` from being available.

I needed to make a distinction between the kind of root import, so I called one "sysdef" and the other "predef". System defined imports can't be excluded while predefined imports can.

If one flag covers both of these styles of imports, then I think we need a special syntax for the flag to distinguish the "style" of import, or we need to unify the behavior of all root imports.

andyscott:
@adriaanm I removed the table, since it was out dated (for the original iteration of this PR).

adriaanm:
Ah, thanks for reminding me. Should we preserve this behavior for the setting, though? We could just require you to use renaming to hide imports from `Predef`, like you would for any other import.

EDIT: hmm, this is confusing -- let me rephrase in another comment

adriaanm:
I think we should not have this special behavior for predef imports in the setting. If you don't explicitly list `scala.Predef._` in the list passed to `-Yimports`, you don't get it; if you want to just import one thing from `Predef`, name it explicitly.

andyscott:
Renaming works. I have a stashed implementation that supported using renaming to "unimport" any existing import (predef, leading, or otherwise).

For this PR, maybe we could stick to preserving the existing behavior for 2.12.2 and I could begin additional work to cleanup the overall behavior? That could be slated for 2.12.3 or beyond.

adriaanm:
I'd rather not do this gradually, but instead figure out the design we can stick with. This option is not used much, and I'd rather spend our energy on this once.

andyscott:
Questions:
- Should we switch to one implementation for root imports? Right now I have it feature flagged.
- Should the new implementation have two style of imports or a unified predef import? A unified import would be a breaking change, IMHO, because an extremely small subset of Scala files wouldn't compile correctly anymore
- Can we do both of the above at once?

adriaanm:
I don't want to change the meaning of code when not using the new `-Yimports` flag. So, yes, we'll have to preserve the logic that interprets predef imports differently when type checking source code.  I assume we can use different/simpler/unified logic to build the root context when deriving it from the flag.

andyscott:
So:
- Keep the "feature flagged" behavior
- Update the new implementation to have one style of import. To "exclude" this import, we will use import renaming à la `import scala.Predef.{_ => _}`.

I have an implementation of this already tucked away. The logic to support unimportant already imports was... wonky. I had to introduce a private symbol `UnimportedSymbol` in `Contexts.scala` to keep track of which predefined imports had been unimproved. I can get a branch of this going again, if you'd like.

adriaanm:
Concretely, I'd take the current implementation

```scala
/** List of symbols to import from in a root context.  Typically that
 *  is `java.lang`, `scala`, and [[scala.Predef]], in that order.  Exceptions:
 *
 *  - if option `-Yno-imports` is given, nothing is imported
 *  - if the unit is java defined, only `java.lang` is imported
 *  - if option `-Yno-predef` is given, if the unit body has an import of Predef
 *    among its leading imports, or if the tree is [[scala.Predef]], `Predef` is not imported.
 */
protected def rootImports(unit: CompilationUnit): List[Symbol] = {
  assert(definitions.isDefinitionsInitialized, "definitions uninitialized")

  if (settings.noimports) Nil
  else if (unit.isJava) RootImports.javaList
  else if (settings.nopredef || treeInfo.noPredefImportForUnit(unit.body)) {
    // SI-8258 Needed for the presentation compiler using -sourcepath, otherwise cycles can occur. See the commit
    //         message for this ticket for an example.
    debuglog("Omitted import of Predef._ for " + unit)
    RootImports.javaAndScalaList
  }
  else RootImports.completeList
}


def rootContext(unit: CompilationUnit, tree: Tree = EmptyTree, throwing: Boolean = false, checking: Boolean = false): Context = {
  val rootImportsContext = (startContext /: rootImports(unit))((c, sym) => c.make(gen.mkWildcardImport(sym)))

  // there must be a scala.xml package when xml literals were parsed in this unit
  if (unit.hasXml && ScalaXmlPackage == NoSymbol)
    reporter.error(unit.firstXmlPos, "To compile XML syntax, the scala.xml package must be on the classpath.\nPlease see https://github.com/scala/scala-xml for details.")

  // scala-xml needs `scala.xml.TopScope` to be in scope globally as `$scope`
  // We detect `scala-xml` by looking for `scala.xml.TopScope` and
  // inject the equivalent of `import scala.xml.{TopScope => $scope}`
  val contextWithXML =
    if (!unit.hasXml || ScalaXmlTopScope == NoSymbol) rootImportsContext
    else rootImportsContext.make(gen.mkImport(ScalaXmlPackage, nme.TopScope, nme.dollarScope))

  val c = contextWithXML.make(tree, unit = unit)

  c.initRootContext(throwing, checking)
  c
}
```

and turn it into:
```scala
protected def rootImports(unit: CompilationUnit): List[Symbol] = {
  assert(definitions.isDefinitionsInitialized, "definitions uninitialized")

  if (unit.isJava) RootImports.javaList
  else if (treeInfo.noPredefImportForUnit(unit.body)) {
    // SI-8258 Needed for the presentation compiler using -sourcepath, otherwise cycles can occur. See the commit
    //         message for this ticket for an example.
    debuglog("Omitted import of Predef._ for " + unit)
    RootImports.javaAndScalaList
  }
  else RootImports.completeList
}


def rootContext(unit: CompilationUnit, tree: Tree = EmptyTree, throwing: Boolean = false, checking: Boolean = false): Context = {
  // scala-xml needs `scala.xml.TopScope` to be in scope globally as `$scope`
  // We detect `scala-xml` by looking for `scala.xml.TopScope` and
  // inject the equivalent of `import scala.xml.{TopScope => $scope}`
  val standardImports =
    if (!settings.Yimports.isSetByUser) {
      val rootImportsContext = (startContext /: rootImports(unit)) ((c, sym) => c.make(gen.mkWildcardImport(sym)))

      // there must be a scala.xml package when xml literals were parsed in this unit
      if (unit.hasXml && ScalaXmlPackage == NoSymbol)
        reporter.error(unit.firstXmlPos, "To compile XML syntax, the scala.xml package must be on the classpath.\nPlease see https://github.com/scala/scala-xml for details.")

      if (!unit.hasXml || ScalaXmlTopScope == NoSymbol) rootImportsContext
      else rootImportsContext.make(gen.mkImport(ScalaXmlPackage, nme.TopScope, nme.dollarScope))
    } else {
      (startContext /: settings.Yimports.value) (_ make _)
    }

  val c = standardImports.make(tree, unit = unit)

  c.initRootContext(throwing, checking)
  c
}

```

andyscott:
If you just create a regular context using the value of `Yimports`, I think you won't be able to exclude any of the predefs. Do we want to allow exclusions on a per file basis with the current mechanism or with import renaming `{_ => _}`.

adriaanm:
if you just want any2stringadd, you would do `-Yimport:scala.Predef.any2stringadd`, right?

adriaanm:
If you don't want it, you'd pass in `scala.Predef.{any2stringadd => _, _}`

andyscott:
Whoops. I should clarify: On a per .scala file basis, do you want to allow excluding root imports? With the code you posted, I believe you won't be able to exclude `Yimport` root imports within a single file.

For excluding a root import in a .scala file, the current pattern is to just import something else from the same scope. `import scala.Predef.DummyImplicit` to exclude `import scala.Predef._`. But I could see supporting a wildcard rename to force an exclusion. `import scala.Predef.{_ => _}`.

adriaanm:
Ah, I see -- hadn't thought of that use case. I'm not sure how useful that is. Seems like it would be subsumed by full control over predef imports? If you have some exotic sources that require opting out of some Predef imports mixed in with others that want the full set, they could either be in a different project that uses different settings or you'd have to explicitly `import Predef._` in the files that need it, and not add it to the `-Yimports` list.

For a more ambitious effort, a SIP for abstracting over imports could be pretty cool -- something like:

```
import myDefault = scala._, scala.Predef.{notThisOne => _, _}
```

and then you can do `import myDefault` to get that list of imports. Import identifiers would live in a separate namespace and have priority over all other identifiers.

andyscott:
For how useful this exclusion behavior is, I'm not sure! Honestly I don't think the exclusion trick extremely useful for `scala.Predef._` on its own, but for broader custom predefs it might be really nice-- especially if it's given a specific syntax such as `{_ => _}`.

Regardless: If I remove the current exclusion behavior (with the split sysdef/predef flags), then the new root import logic won't be backwards compatible with existing code. Breaking changes (extremely uncommon, I'm guessing) under an experimental flag seem fine-- but my long term goal would be to dump the existing implementation in favor of this more general implementation. What's the best path forward on this?

adriaanm:
I don't know. I would prefer to modularize Predef so that its parts can be wildcard-imported, rather than relying on a special interpretation of what it means to import a named entity from Predef to avoid importing all of it.

som-snytt:
Just noticed there's a ticket for this. https://issues.scala-lang.org/browse/SI-9341

fommil:
I really like the concept of this!

andyscott:
Any updates/ideas on how to best approach this? As it stands, I don't break any existing behavior-- but that means I've built some logic to accommodate the uncommon predef exclusion trick in the current compiler.

That being said... with the ability to specify custom predefs, I'm beginning to like the idea of retaining the predef exclusion logic.

If I have a predef of `foo.bar.predef._` and I import `foo.bar.predef.someImplicit`, then it seems reasonable to me that the rest of `foo.bar.predef._` should be automatically excluded. I suspect this isn't used much because most people take all of `scala.Predef._` for granted and have no reason to individually import items. With custom predefs, that probably won't hold true.

Thoughts?

adriaanm:
> I suspect this isn't used much because most people take all of scala.Predef._ for granted and have no reason to individually import items. 

Good point -- that made something click for me :-) 

My thinking about this case had become very uptight

> With custom predefs, that probably won't hold true.

How so? A predef module should be designed so that you can import all of it at once, no¿

BTW, I just had a neat idea on how to model abstracting over imports:

```
object MyPredef {
  export Predef.usefulMember
  export Predef.{awkwardName => betterName}
}

import MyPredef._ // same as having those `export` statements as `import`s here
```

An interesting question: should an export result in a member? that way you could implement members by "exporting" them.

clhodapp:
If you could export `Foo._`, it would open up possibilities for abstraction that were previously only accessible via inheritance.

adriaanm:
Indeed :-)

lihaoyi:
I can't be the only one to remember this has been talked about before :P https://docs.google.com/document/d/1dlT6NgB9610jqLscCJW2LRB7TapDh3q4d2S3YA_q5zs/edit

adriaanm:
Thanks -- I for one had forgotten.

milessabin:
Some sort of "export" construct could be extremely useful. See also my [export-hook](https://github.com/milessabin/export-hook), in particular the description of [`@reexports`](https://github.com/milessabin/export-hook#bundling-and-reexporting-type-class-instances).

jvican:
@adriaanm Are these just cool ideas or are they going to be materialized in a SIP? :wink:

milessabin:
FYI: this has been [merged](https://github.com/typelevel/scala/blob/typelevel-readme/notes/2.12.1.md#configurable-default-imports-pull5350-andyscott) in Typelevel Scala 2.12.1.

adriaanm:
Before we merge this, I'd like to reduce this to one command line option. In code, here's what I have in mind:

```scala
/** Standard imports to add to the root context for each compilation unit.
  *
  *  - A Java-defined unit imports only `java.lang._`
  *  - Scala-defined units:
  *    - import `java.lang._`, `scala._`, `scala.Predef._` by default, unless the `-Yimports` setting is set by the user,
  *      in which case the specified imports are used instead.
  *    - If there is a top-level import from `Predef`, or if the unit defines `scala.Predef` itself,
  *      the `scala.Predef._` import will be removed from the default imports (even if specified using `-Yimports`).
  *
  *  TODO: allow configuring the list of modules that receive the `Predef`-style import filtering?
  */
protected def rootImports(unit: CompilationUnit): List[Import] = {
  assert(definitions.isDefinitionsInitialized, "definitions uninitialized")

  if (unit.isJava) RootImports.javaList.map(gen.mkWildcardImport)
  else {
    def makeImports(value: List[String]): List[Import] = ???

    val scalaImports =
      if (settings.YImports.isSetByUser) makeImports(settings.YImports.value)
      else RootImports.completeList.map(gen.mkWildcardImport)
    
    // SI-8258 Needed for the presentation compiler using -sourcepath, otherwise cycles can occur. See the commit
    //         message for this ticket for an example.
    if (treeInfo.noPredefImportForUnit(unit.body)) scalaImports.filter(_.symbol != PredefModule)
    else scalaImports
  }
}


def rootContext(unit: CompilationUnit, tree: Tree = EmptyTree, throwing: Boolean = false, checking: Boolean = false): Context = {
  val rootImportsContext = (startContext /: rootImports(unit))(_ make _)
```

andyscott:
I'll take a stab at this over the weekend when I'll have some hack time.

lrytz:
Thank you! It's been a lot of back and forth on this one..

adriaanm:
In a discussion with retronym, it just occurred to me this is also useful during migration. We could remove stuff from Predef, and leave DeprecatedPredef around for you to import using this flag while you migrate from 2.12 to 2.13

fommil:
it's not overly obvious, but the docs of the syntax would lead one to believe that `-Ysysdef:foo._,bar._` should work, but it does not because it needs to be two separate parameters `"-Ysysdef", "foo._,bar._"`.

A few niggles I noticed:

1. the predef seems to need to be defined outside of the compilation? e.g. I cannot put a `predef.scala` into my project and then use `-Ypredef fommil.Predef._`.
2. not possible to use `-Ysysdef` with `-Yno-predef` :disappointed: which would be useful. Workaround is to use `"-Ypredef", "_"`

fommil:
~~eek, most importantly, this doesn't seem to be recognised by the presentation compiler!~~ never mind, bug in sbt-ensime / coursier.

som-snytt:
Also +1 for a `MultiStringSetting` with single arg syntax `-Ysysdef:foo._,bar._`. Reviewing the transcript, I may have suggested the opposite last year.

fommil:
Some feedback.

I've been using this feature in the TL build of scala (in fact that's why I'm using TL Scala). It's great, I love it!

I've found I use it to minimise `scala.Predef` rather than provide a custom, extended, one. Indeed, it's made me conclude that I don't want a `Predef` at all, I just want to control the system imports.

This is what I am using:

```scala
  "-Ysysdef", Seq(
    "java.lang.String",
    "scala.{Any,AnyRef,AnyVal,Boolean,Byte,Double,Float,Short,Int,Long,Char,Unit,Null,Nothing,Option,Some,None,Either,Left,Right,StringContext}",
    "scala.annotation.{tailrec,inductive}",
    "scala.collection.immutable.{Map,Seq,List,::,Nil,Set,Vector}",
    "scala.util.{Try,Success,Failure}",
    // prefer `the` to `implicitly`
    "scala.Predef.{???,ArrowAssoc,identity,<:<,=:=}"
    // best not to import non-stdlib stuff or it breaks inferior IDEs
    // "cats._,cats.data._,cats.implicits._",
    // "freestyle._,freestyle.implicits._",
    // "shapeless.{::=>#:,_}",
    // "fs2._"
  ).mkString(","),
  "-Ypredef", "_"
```

There are some problems with going beyond the std imports / predefs:

1. developers using inferior IDEs are unable to get the benefit of this compiler feature
2. the system imports do not behave like a preamble, it feels more like compiler magic and perhaps there is an opportunity to simplify the semantics in a (breaking) future version of scala.

As an example, I was using `fs2._` in my sysdefs and the following code was fine

```
import io.circe._
```

but if I were to use the default imports/predef (or none) then this code would *not* compile

```
import fs2._
import io.circe._
```

because `fs2` has a subpackage called `io`.

So, as much as I love this feature (and I'm really looking forward to it), I must admit to being a little bit suspicious about the whole concept of a sysdef / predef (including the one in the current releases of scala). What I actually want, and what I think a lot of people will assume this is: is the ability to provide a custom set of imports that behave **exactly** like a preamble in all source files.

fommil:
one more thing, my import of `scala.collection.immutable.::` means that when I have an explicit `import shapeless._` in my code, I get ambiguous imports for uses of `::`. Workaround is to `import shapeless.{:: => #:, _}`... but I would have expected the `shapeless` import to win because it comes after the preamble. Further evidence that the system / predef imports are magical?

som-snytt:
Name binding precedence and shadowing in chapter 2 of the spec is surprisingly surprising.

The spec says `Predef is implicitly imported` but doesn't specify how. I noticed you get a scoping level; the following says it's a wildcard import.

```
scala> assert(true)

scala> :pa

package object p { def assert(b: Boolean): Unit = () }

scala> { import p._ ; assert(false) }

scala> { import Predef.assert ; { import p._ ; assert(false) } }
<console>:12: error: reference to assert is ambiguous;
it is imported twice in the same scope by
import p._
and import Predef.assert
       { import Predef.assert ; { import p._ ; assert(false) } }
                                               ^

scala> { import Predef._ ; { import p._ ; assert(false) } }

scala> 
```

alexandru:
Just dropping a note - my use-case for this feature is to be able to provide "orphaned type class instances", which are implicits that aren't visible by default due to them not being defined in the companion objects of the types involved.

I'd like to be able to instruct users to just use some import by default.

Seeing that this is scheduled for shipping in 2.12.3 (hopefully, fingers crossed) makes me very happy.

alexandru:
Also, I see that because this is about default system definitions and Predef, there might be conflicts involved.

I'm personally fine with library authors preparing packages or objects for this feature (clean, conflict free).

fommil:
btw, this feature doesn't work very well with `warn-unused`... lots of warnings.

alexandru:
Ah, that might be a blocker.

How does warn-unused work with the current sys imports and Predef? Does it do hard-coded filtering? Can it be fixed?

som-snytt:
There's an old jedi mind trick:
```
if (settings.warnUnusedImport && !isRootImport) // excludes java.lang/scala/Predef imports
```
where
```
override final def isRootImport = !tree.pos.isDefined
```

adriaanm:
> shipping in 2.12.3

The milestone mostly indicates we're open to shipping this in that release, but we likely will need some help to address the super valuable feedback by @fommil and others before this can actually ship.

adriaanm:
I'm sorry to bump this again.

fommil:
I'm delighted to bump it again.

andyscott:
I'd like to finish this feature and get it shipped. I'll haven't reviewed
the code or discussion in quite a while, so over the next few weeks I'll
try to dive back in and post back here with an update so we can sort out
how to move forward.

som-snytt:
I could have used this feature today. I see I commented last April about name binding. Today I fixed the handling of "level 4" names, with disastrous results: from the point of view of the collections library, many implicit imports introduce ambiguity, for example, if there is a `mutable.Seq` in your package, it no longer shadows the `Seq` from either `Predef` or `scala.package`. Using the blunt instrument of `-Yno-imports`, I trade those errors for not knowing what `String` is. So I would like to try a minimal `Predef` for compiling the standard library.

som-snytt:
It would also be nice for the leading import escape hatch to work for e.g. `import scala.{util => _, _}`. Otherwise, if I have package `myproject.util` and I'm in `package myproject`, I can't `import util.C` using the relative package, without explicitly importing `myproject.util`.

SethTisue:
closing for inactivity. can always be reopened

andyscott:
sad kitten

SethTisue:
@andyscott if you'd still like to work on this — and/or if someone else wants to take it up — we'd be happy to reopen

SethTisue:
(P.S. if having some help might help this move towards the finish line, you might post about it on https://contributors.scala-lang.org and see if you can generate some interest/assistance that way)

fommil:
to summarise my feedback from above:

1. I personally don't care about Predef. I think everything can be achieved with custom imports.
2. The system imports are triggering unused import warnings, they should be special cased and the warning not emitted.

other than that, I can live with all the other glitches because I'd really like system imports to behave as if I wrote them explicitly at the top of the file.

som-snytt:
I'm interested in helping, because I think fixing name binding in another PR will make more folks (common folk) consider controlling the implicit imports in projects, even if they are a handy default for beginners and casual REPLing. I'll also look at whether they warn. How can they warn if they are not in code?

andyscott:
If we can come up with a reasonable path forward I'd be very happy to allocate some time to wrap this up (or something similar in functionality).

Is this PR the best place to continue the discussion?

lrytz:
Yes please! I spinned of a scala-dev ticket to continue the discussion: https://github.com/scala/scala-dev/issues/474.

som-snytt:
I left a note on scala-dev that I've proposed a simplified version of this feature, as part of a PR that has been hanging around about as long as this one.

adriaanm:
Heads up: if this is going to make it into 2.13, it will need to be ready to merge by M5 (due Aug 10).

som-snytt:
My simpler `-Yimports` is at https://github.com/scala/scala/pull/6589/commits/1bae6f4311e7f6f70e64b3b3b7a4f7e3dca8b65b

part of https://github.com/scala/scala/pull/6589/commits

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -99,8 +99,79 @@ trait Contexts { self: Analyzer =>
...
 
+  // Collects the leading imports in a given unit body.
+  //
+  // This method should return the list of imports that's
+  // checked in `noPredefImportForUnit`
+  //
+  private[this] def collectLeadingImports(body: Tree): List[Import] = {
+    @tailrec def loop(rem: List[Tree], acc: List[Import]): List[Import] = rem match {
+      case tree :: tail => tree match {
+        case PackageDef(_, stats) => loop(stats ::: tail, acc)
```
som-snytt, line 110: This drills down through package statements, different from current check.

som-snytt: Sorry, I retract that! It does in fact enter nested packages looking for the exclusionary import.

som-snytt:
Looking this over again for ideas and LOC for my `-Yimports`.

Thinking about this during unbearably long Bay Area commutes, I wanted to minimize scans of unit trees, and also minimize context lookups, in part to avoid forcing infos. I was also thinking that you only need to look at previous imports (including renames).

But the approach taken here doesn't work for this case where a definition aliases a name in the full name of the predef object `hello.world.minidef` (from a test):

```
import hello.world._

object minidef {
  val somethingElse = 17
}

// does this look like predef import?
import minidef._
```

In a normal context lookup, the definition shadows the import.

The problem is partly that the existing implementation doesn't just look at "leading" imports. But switching the order of the object and the import from it shouldn't change semantics. 

The revised words at [this puzzler](http://scalapuzzlers.com/#pzzlr-046) are still not right, but the gist is that the import can be at the bottom of the file, like vi modelines. Possibly we should just disallow that, for the sake of efficiency.

Here is the working example, where `INT` and `Magic` come from `-Ypredef hello.world.minidef._`. Uncommenting the import breaks it as shown. `selectively import` was to show that the nested import doesn't disqualify the predef.

```
//import hello.world._

// does this look like predef import?
import minidef._

object minidef {
  val somethingElse = 17
}

object Test {
  // selectively import
  //import hello.world.minidef.INT
  val a: INT = Magic
  println("this should be OK")
  def f = somethingElse
}
```
Edit: there are couple of heuristics or limitations that we could apply, such as taking only leading imports, up to a def, and only at top-level or top-level package; and either requiring fully-qualified import (with the exception of `scala.Predef`) or warning for the case shown here, which is easy? to check for.

By fully-qualified, I wasn't even thinking "absolute", `_root_`-rooted.

Proposed warning, torn from the headlines of The Onion:

`import minidef._ permanently shadows import of -Ypredef`

Or is there a fifth level of name binding that ignores definitions in imports?

som-snytt:
This work was brought forward in a slightly different form in https://github.com/scala/scala/pull/6764 which supersedes this PR.

The merged PR is one flag, and distinguishes "sysdef" as packages and "predef" as objects that are not packages. You get wildcard imports from either sort, and the order is respected, where later shadows earlier, as though nested. The way predefs are excluded is similar but arguably more workable.

"Fine-grained" control is not supported. Root contexts work more like nested packages than like imports.

I commented that a true `-Ypreamble` could be added separately, where a preamble is just code prepended to source text. That preamble could be an arbitrary import, for example.

PR #5351, opened 2016-08-19
SI-9823 Rephrase builders for cache friendliness

retronym:
The HotSpot VM contains a one element cache in the metadata
of each class to record the most recent successful test for
a secondary interface.

For example:

```
classOf[Serializable].isAssignableFrom(classOf[Some[_]])
```

Would, in addition to returning `true`, set:

```
classOf[Some[_]]._secondary_super_cache = classOf[Serializable]
```

This is done to hide the fact that interface tests are O(N),
where N is the number of inherited interfaces.

This scheme is discussed in "Fast Subtype Checking for the
HotSpot JVM" (Click, Rose) [1]

However, if this cache repeatedly misses, not only are we
exposed to the linear search of the secondary super type array,
but we are also required to write back to the cache. If other
cores are operating on the same cache line, this can lead to
a significant slowdown ("cache thrashing"). This effect will
by most (or only?) visible on multi socket machines.

The pathological case is:

```
scala> Iterator.continually(List(classOf[Product], classOf[Serializable])).flatten.take(100).map(intf => intf.isAssignableFrom(classOf[Some[_]])).count(x => x)
res19: Int = 100
```

Which, if run on a multi-socket machine, should be much slower
than:

```
scala> (Iterator.continually(classOf[Product]).take(50) ++ Iterator.continually(classOf[Serializable]).take(50)).map(intf => intf.isAssignableFrom(classOf[Some[_]])).count(x => x)
res20: Int = 100
```

This commit avoids a interface test in a hot path in the collections
by instead using virtual dispatch to differentiate between
IndexedSeqLike and other collections. HotSpot will use some
shared bookkeeping ("inline cache" [2]) at the callsites of
this method, but these aren't prone to need to write to
memory frequently.

[1] https://www.researchgate.net/publication/221552851_Fast_subtype_checking_in_the_HotSpot_JVM
[2] https://wiki.openjdk.java.net/display/HotSpot/PerformanceTechniques

### DIFF
--- a/src/library/scala/collection/TraversableLike.scala
+++ b/src/library/scala/collection/TraversableLike.scala
@@ -142,7 +142,7 @@ trait TraversableLike[+A, +Repr] extends Any
 
   def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Repr, B, That]): That = {
     val b = bf(repr)
-    if (that.isInstanceOf[IndexedSeqLike[_, _]]) b.sizeHint(this, that.seq.size)
+    if (that.isInstanceOf[IndexedSeqLike[_, _]]) sizeHintTo(b, that.seq.size)
     b ++= thisCollection
     b ++= that.seq
     b.result
@@ -182,7 +182,7 @@ trait TraversableLike[+A, +Repr] extends Any
    */
   def ++:[B >: A, That](that: TraversableOnce[B])(implicit bf: CanBuildFrom[Repr, B, That]): That = {
     val b = bf(repr)
-    if (that.isInstanceOf[IndexedSeqLike[_, _]]) b.sizeHint(this, that.size)
+    if (that.isInstanceOf[IndexedSeqLike[_, _]]) sizeHintTo(b, that.size)
     b ++= that
     b ++= thisCollection
     b.result
@@ -227,7 +227,7 @@ trait TraversableLike[+A, +Repr] extends Any
   def map[B, That](f: A => B)(implicit bf: CanBuildFrom[Repr, B, That]): That = {
     def builder = { // extracted to keep method size under 35 bytes, so that it can be JIT-inlined
       val b = bf(repr)
-      b.sizeHint(this)
+      sizeHintTo(b, 0)
       b
     }
     val b = builder
@@ -368,7 +368,7 @@ trait TraversableLike[+A, +Repr] extends Any
 
   def scanLeft[B, That](z: B)(op: (B, A) => B)(implicit bf: CanBuildFrom[Repr, B, That]): That = {
     val b = bf(repr)
-    b.sizeHint(this, 1)
+    sizeHintTo(b, 1)
     var acc = z
     b += acc
     for (x <- this) { acc = op(acc, x); b += acc }
@@ -452,7 +452,7 @@ trait TraversableLike[+A, +Repr] extends Any
     var lst = head
     var follow = false
     val b = newBuilder
-    b.sizeHint(this, -1)
+    sizeHintTo(b, -1)
     for (x <- this) {
       if (follow) b += lst
       else follow = true
@@ -466,7 +466,7 @@ trait TraversableLike[+A, +Repr] extends Any
   def drop(n: Int): Repr =
     if (n <= 0) {
       val b = newBuilder
-      b.sizeHint(this)
+      sizeHintTo(b, 0)
       (b ++= thisCollection).result
     }
     else sliceWithKnownDelta(n, Int.MaxValue, -n)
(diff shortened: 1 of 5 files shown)


retronym:
I'm proposing this as a late addition to 2.12.0. I'm going to add a benchmark and then find a big multi-socket box to run it on.

adriaanm:
We need a bigger box!

som-snytt:
Today in the mail, I got my "lurk and learn" t-shirt!

sjrd:
The same should probably be done for [`sizeHintBounded`](https://github.com/scala/scala/blob/99a82be91cbb85239f70508f6695c6b21fd3558c/src/library/scala/collection/mutable/Builder.scala#L114).

It seems that this removes all usages of the overloads of `sizeHint` that take a `coll`. The same would apply for `sizeHintBounded`. In which case, I guess [these methods](https://github.com/scala/scala/blob/99a82be91cbb85239f70508f6695c6b21fd3558c/src/library/scala/collection/mutable/Builder.scala#L65-L117) should be deprecated.

retronym:
> The same should probably be done for sizeHintBounded.

Yep, that would be a consistent change. That method use used in `slice`, `takeRight`, `splitAt`.

> It seems that this removes all usages of the overloads of sizeHint that take a coll.

There was one left in `Tuple2Zipped#map`. I could either make `sizeHintTo` et al public, rather than protected, or just inline the type test into that usage and proceed with the deprecation.

That said, I'm not 100% sure about deprecating as the new alternative isn't battle tested (maybe it has different performance gotchas under some workloads?) But at least we should mention the new methods from the Scaladoc.

src/library/scala/collection/TraversableLike.scala
```diff
@@ -142,7 +142,7 @@ trait TraversableLike[+A, +Repr] extends Any
 
   def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Repr, B, That]): That = {
     val b = bf(repr)
-    if (that.isInstanceOf[IndexedSeqLike[_, _]]) b.sizeHint(this, that.seq.size)
+    if (that.isInstanceOf[IndexedSeqLike[_, _]]) sizeHintTo(b, that.seq.size)
```
shanielh, line 145: Don't you need to remove this if and replace it with "sizeHintTo" (that will do nothing in case that 'that' isn't an IndexedSeqLike)?

retronym, line 145: No, we can't easily use virtual dispatch to check if `this` and `that` are both `IndexedSeqLike`.

shanielh, line 145: Oh, Got it. Didn't see the implementation of `mutable.Builder` before it. Just for curiosity - Can't we use double dispatch for that?

Ichoran:
I think the performance impact of this change is quite unlikely to be negative in any important case.  The JVM has virtual dispatch more carefully optimized than practically anything else.  I would recommend adding it to 2.12 unless an actual correctness bug is found (assuming it has the performance benefit in the condition it's supposed to).

retronym:
The reporter of the bug ran this branch against the microbenchmark posted in JIRA and reports that the performance problem is resolved.

``` scala
// Running environment is critical; on 2 socket system with 18 cores/socket, HT disabled
// Use -XX:-TieredCompilation
class Task extends Runnable {
  val c1 = Array(-1, 1)
  val c2 = Array(-2, 2)
  def run() {
    for (i <- 1 to 200000000) {
      c1.zip(c2).map(_._1).toSeq.zip(1 until 3).map(_._1)
    }
  }
}

object ScalaInstanceof {
    val numThreads = args(0).toInt
    val es = java.util.concurrent.Executors.newFixedThreadPool(numThreads)
    for (i <- 1 to numThreads) es.submit(new Task)
    es.shutdown()
  }
}
```

> Before: 24m12s, CPI at 8.43
> After: 1m48s, CPI 0.53
> System is Haswell-EP, 2 sockets, HT disabled, 36 cores.
> [details](https://docs.google.com/spreadsheets/d/1z4gouPNRBCrtcfnjNZvyWxxV186hYEivwCorOVaQYNc/pubhtml)

The bottleneck in the "before" test was in the compiled code [for `sizeHint`](https://github.com/scala/scala/blob/v2.11.8/src/library/scala/collection/mutable/Builder.scala#L68)

image

I was able to reproduce similar results on a 2x6core Xeon HT disabled. On the same machine, I also confirmed that single threaded performance for the benchmark was unchanged by this patch wrt 2.12.0-M5.

In my experiments, I found that I didn't see the pathologically bad performance with hyperthreading _enabled_. I've asked the reporter whether he can reproduce that result and offer insight into why that might be. That aspect of needn't hold up this patch, but will help complete the picture.

Intel has a useful [reference](https://software.intel.com/en-us/articles/avoiding-and-identifying-false-sharing-among-threads) that explains "false sharing", which is the phenomona we're witnessing here. Although in this instance, the sharing isn't completely "false", it would be better described as "needless". We'd really like to have a "socket local" copy of the secondary_super_cache rather than one shared across the machine.

I will address @sjrd's feedback today.

retronym:
Closing in favour of #5364, which I believe is a better solution.

PR #5352, opened 2016-08-19
Deprecate scala.Enumeration, add @enum annotation

soc:
Rough draft, some things are necessarily ugly, others temporarily.

The syntax

```
@enum
class Week(val isWeekend: Boolean = false) {
  Monday  (false)
  ...
  ...
  Friday  (false)            { def isGoodDay = true }
  Saturday(true)             { val isGoodDay = true }
  Sunday  (isWeekend = true) { def isGoodDay = true }

  def isGoodDay = false
}
```

is a possible implementation example, intention is to also support ADTs
that conform to certain restrictions (no nesting, recursion or
varying constructor parameters), e. g.:

```
@enum
sealed trait Toggle
case object ON  extends Toggle
case object OFF extends Toggle
```

Deprecates the unmitigated disaster that is scala.Enumeration.
Advantages of @enum over scala.Enumeration:
- Actually works
- Java interop
- No erasure issues
- No confusing mini-DSL to learn when defining enumerations

Disadvantages: None.

This addresses the issue of not being able to have one codebase that
supports Scala-JVM, Scala.js and Scala-Native (Java source code
not supported on Scala.js/Scala-Native, Scala source code not able to
define enums that are accepted by existing APIs on Scala-JVM).

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -665,9 +665,14 @@ trait Namers extends MethodSynthesis {
       if (isScala && deriveAccessors(tree)) enterGetterSetter(tree)
       else assignAndEnterFinishedSymbol(tree)
 
+      val sym = tree.symbol
+
       if (isEnumConstant(tree)) {
-        tree.symbol setInfo ConstantType(Constant(tree.symbol))
-        tree.symbol.owner.linkedClassOfClass addChild tree.symbol
+        sym setInfo ConstantType(Constant(sym))
+        if (sym.isJavaDefined)
+          sym.owner.linkedClassOfClass addChild sym
+        else
+          sym.owner addChild sym
       }
     }
 
@@ -698,6 +703,38 @@ trait Namers extends MethodSynthesis {
       tree.symbol = enterClassSymbol(tree)
       tree.symbol setInfo completerOf(tree)
 
+      if (mods.hasAnnotationNamed(TypeName("enum"))) {
+        var ordinal = 0
+        def newEnumItemOrdinalAttachment() = {
+          val att = new EnumConstantOrdinalAttachment(ordinal)
+          ordinal += 1
+          att
+        }
+
+        def isPossibleEnumConstantTree(tree: Tree): Boolean = { val result = tree.isInstanceOf[Ident] || tree.isInstanceOf[Apply]; println(s"$tree is enum? $result"); result }
+        def enumConstantName(enumConstant: Tree): TermName = enumConstant match {
+          case Ident(termName: TermName)                     => termName
+          case Apply(Ident(termName: TermName), _)           => termName
+          case Apply(Apply(Ident(termName: TermName), _), _) => termName
+        }
+
+        val enumConstants: List[TermName] =
+          tree.impl.body.iterator
+            .dropWhile(tree => !isPossibleEnumConstantTree(tree))
+            .takeWhile(isPossibleEnumConstantTree)
+            .map(const => {
+              const.updateAttachment(newEnumItemOrdinalAttachment())
+              const.symbol.setFlag(JAVA_ENUM)
+              enumConstantName(const)})
+            .toList
+
+        tree.symbol.setFlag(JAVA_ENUM)
+        tree.symbol.updateAttachment(EnumConstantsAttachment(enumConstants))
+
+        // We keep this for the moment, otherwise the compiler doesn't see the members.
+        ensureCompanionObject(tree)
+      }
+
       if (mods.isCase) {
         val m = ensureCompanionObject(tree, caseModuleDef)
         m.moduleClass.updateAttachment(new ClassForCaseCompanionAttachment(tree))
@@ -718,7 +755,7 @@ trait Namers extends MethodSynthesis {
       // Suggested location only.
       if (mods.isImplicit) {
         if (primaryConstructorArity == 1) {
-          log("enter implicit wrapper "+tree+", owner = "+owner)
+          log(s"enter implicit wrapper $tree, owner = $owner")
           enterImplicitWrapper(tree)
         }
         else reporter.error(tree.pos, "implicit classes must accept exactly one primary constructor parameter")
@@ -1028,6 +1065,7 @@ trait Namers extends MethodSynthesis {
 
     private def templateSig(templ: Template): Type = {
       val clazz = context.owner
+
       def checkParent(tpt: Tree): Type = {
         if (tpt.tpe.isError) AnyRefTpe
         else tpt.tpe
@@ -1040,6 +1078,10 @@ trait Namers extends MethodSynthesis {
       val decls = newScope
       val templateNamer = newNamer(context.make(templ, clazz, decls))
       templateNamer enterSyms templ.body
+
+      if (clazz.hasJavaEnumFlag)
+        clazz.attachments.get[EnumConstantsAttachment]
+          .map(att => addEnumMembers(att.constants, templateNamer))
 
       // add apply and unapply methods to companion objects of case classes,
       // unless they exist already; here, "clazz" is the module class
@@ -1617,7 +1659,6 @@ trait Namers extends MethodSynthesis {
       }
     }
 
-
     /** Given a case class
      *   case class C[Ts] (ps: Us)
      *  Add the following methods to toScope:
@@ -1641,6 +1682,24 @@ trait Namers extends MethodSynthesis {
 
     def addCopyMethod(cdef: ClassDef, namer: Namer) {
       caseClassCopyMeth(cdef) foreach namer.enterSyntheticSym
+    }
+
+    /** Given an enum class §EnumClass with enum constants §EnumConstant1 ... §EnumConstantN,
+      * add the following members to scope:
+      * - a private static field $VALUES containing the enum constants
+      *   `<static> private[this] val $VALUES: Array[§EnumCass]`
+      * - a public static method that returns a copy of $VALUES
+      *   `<static> def values: Array[§EnumCass]`
+      *   a public static method that returns the enum constant denoted by the string if such an enum constant exists
+      * - `<static> def valueOf(name: String): §EnumCass`
+      *
+      * @param constants a list of the enum constants defined by this enum
+      * @param namer     the namer of this class
+      */
+    def addEnumMembers(constants: List[TermName], namer: Namer) = {
+      namer.enterSyntheticSym(enumValuesField(owner, constants))
+      namer.enterSyntheticSym(enumValuesMethod(owner))
+      namer.enterSyntheticSym(enumValueOfMethod(owner))
     }
 
     /**
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -2773,7 +2773,15 @@ self =>
       val annots = annotations(skipNewLines = true)
       val pos    = caseAwareTokenOffset
       val mods   = modifiers() withAnnotations annots
-      tmplDef(pos, mods)
+      // We need to come up with a more principled way to communicate the ability for
+      // exhaustiveness checks to the pattern matcher instead of abusing various flag
+      // combinations that need to be undone again in a later phase.
+      // It should be possible to address both Java-defined and Scala-defined enums with a simple
+      // check for the JAVA_ENUM flag.
+      val mods1 =
+        if (mods.hasAnnotationNamed("enum")) mods | Flags.JAVA_ENUM | Flags.SEALED | Flags.FINAL
+        else mods
+      tmplDef(pos, mods1)
     }
 
     /** {{{
@@ -2828,9 +2836,8 @@ self =>
           val (constrMods, vparamss) =
             if (mods.isTrait) (Modifiers(Flags.TRAIT), List())
             else (accessModifierOpt(), paramClauses(name, classContextBounds, ofCaseClass = mods.isCase))
-          var mods1 = mods
-          val template = templateOpt(mods1, name, constrMods withAnnotations constrAnnots, vparamss, tstart)
-          val result = gen.mkClassDef(mods1, name, tparams, template)
+          val template = templateOpt(mods, name, constrMods withAnnotations constrAnnots, vparamss, tstart)
+          val result = gen.mkClassDef(mods, name, tparams, template)
           // Context bounds generate implicit parameters (part of the template) with types
           // from tparams: we need to ensure these don't overlap
           if (!classContextBounds.isEmpty)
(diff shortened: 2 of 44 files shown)


fommil:
What does the annotation do? Make a val available that aggregates all known children?

soc:
It arranges the class file so that it is accepted as a JVM enum and usable from both Scala and Java.
This includes adding a superclass, setting class file flags, emitting static fields and members, teaching scalac to also look in Scala class files for static stuff, etc.

xerial:
Big 👍 for this change. Even in Scala, I needed to use Java enumeration for avoiding many shortcomings of Scala.Enumeration (e.g., cannot define any method on enum values, the value type becomes MyEnum.Value, not MyEnum, etc.)

mslinn:
Does this construct define an `Ordering`? Java enums can be compared by importing

```
import scala.math.Ordering.Implicits._
```

I don't want to forgo that capability. Also Java enums `valueOf` and `values` methods are important.

sjrd:
I do welcome a means to define proper Java enums in Scala. That said, this PR needs:
1. Tests
2. A SIP

soc:
@mslinn I'm not sure what you are asking ... it compiles to a Java enum, so whatever works for an enum works with it, too.

EDIT: Oh, ok, understood. It can't extend `Ordered` due to some brokenness in the `Ordered` <-> `Comparable` hierarchy, and `Enum` declaring all methods final, but if `Ordering` provides implicits for `enum` that will of course work.

mslinn:
I did not look at the implementation, so wasn't entirely sure if this compiled to a Java enum or not. Presumably the automatically generated ordinal values are derived from the order in which the enum's values are declared, as is the case for Java, and hopefully the automatic string representation for each enum would also be provided to Java.

soc:
@mslinn Yes, everything as expected.

soc:
@sjrd Working on tests, there should be a few failures right now anyway. Have to think about the SIP, there is of course the existing one from 3.5 years ago if someone wants to push that again ...

djspiewak:
Big :+1: for addressing the flaming barnyard that is `scala.Enumeration`.  Some concerns on `@enum`…

To be useful, I think it needs to support at a minimum the following two cases:

``` scala
@enum
sealed trait Number

case object One extends Number
case object Two extends Number
```

And this:

``` scala
@enum
sealed trait Number

object Number {
  case object One extends Number
  case object Two extends Number
}
```

In my experience, the latter is more commonly used in larger code bases.

Handling final case classes as well would be icing on the cake.

soc:
@djspiewak Yes, agree that both would need to work. I need to have a look whether it can work reliably in general, maybe after @milessabin's fix ...

fanf:
Does it have something to do with https://github.com/LPTK/Boilerless/blob/master/README.md ? 

If not, there seems to be a lot of motivation for the subject at the moment, and perhaps a sip is effectively the way to go.

soc:
@fanf No, I'm working on it in various ways since 2013.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -665,7 +665,7 @@ trait Namers extends MethodSynthesis {
       else assignAndEnterFinishedSymbol(tree)
 
       if (isEnumConstant(tree)) {
-        tree.symbol setInfo ConstantType(Constant(tree.symbol))
+        //tree.symbol setInfo ConstantType(Constant(tree.symbol))
```
soc: Currently some tests fail with `annotation argument needs to be a constant; found: Enum.VALUE`, but enabling setting the constant type causes:

```
Week.scala:20: error: type mismatch;
 found   : foo.Week.type(Saturday)
 required: foo.Week
```

So we either need to make it understand that the constant type is a subtype of the enum type, or change the way we handle constants.

EDIT: Nevermind, was an issue in `def EnumType`. Currently the compiler considers Java-defined enum constants to be added to an imaginary companion object, while Scala-defined enum constants go into the class. It's kind of weird, and might warrant cleanup, but it works now.

Thinking about it, we should really consider treating Java static members as static members, not members of an imaginary companion object. I think that trying to hide the difference causes more trouble than it is avoiding.

soc:
Ok, I fixed all test failures except of one, t6613:

```
nsc> 
nsc> 
nsc> java.lang.AssertionError: assertion failed: 
  Cannot create ClassBType from NoSymbol
     while compiling: t6613/Broken.scala
        during phase: jvm
     library version: version 2.12.0-20160819-055227-d96cbbbe44
    compiler version: version 2.12.0-20160819-230807-f419dc967d
  reconstructed args: <snip>

  last tree to typer: TypeTree(class Enummy$Broke)
       tree position: line 1 of t6613/Broken.scala
            tree tpe: Enummy$Broke
              symbol: (final sealed abstract) class Enummy$Broke
   symbol definition: final sealed abstract class Enummy$Broke extends Enum (a ClassSymbol)
      symbol package: <empty>
       symbol owners: class Enummy$Broke
           call site: constructor Broken in class Broken in package <empty>

     0 class Broken() { def broken() = Enummy.Broke.CHIP }
    at scala.tools.nsc.Global.assert(Global.scala:229)
    at scala.tools.nsc.backend.jvm.BTypesFromSymbols.classBTypeFromSymbol(BTypesFromSymbols.scala:109)
    at scala.tools.nsc.backend.jvm.BCodeHelpers$BCInnerClassGen.internalName(BCodeHelpers.scala:529)
    at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.fieldOp(BCodeBodyBuilder.scala:412)
    at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.fieldLoad(BCodeBodyBuilder.scala:401)
    at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:341)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.emitNormalMethodBody$1(BCodeSkelBuilder.scala:597)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.genDefDef(BCodeSkelBuilder.scala:629)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.gen(BCodeSkelBuilder.scala:503)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.$anonfun$gen$7(BCodeSkelBuilder.scala:505)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.$anonfun$gen$7$adapted(BCodeSkelBuilder.scala:505)
    at scala.collection.immutable.List.foreach(List.scala:378)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.gen(BCodeSkelBuilder.scala:505)
    at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.genPlainClass(BCodeSkelBuilder.scala:113)
    at scala.tools.nsc.backend.jvm.GenBCode$BCodePhase$Worker1.visit(GenBCode.scala:190)
    at scala.tools.nsc.backend.jvm.GenBCode$BCodePhase$Worker1.$anonfun$run$1(GenBCode.scala:139)
    at scala.tools.nsc.Global$GlobalPhase.withCurrentUnitNoLog(Global.scala:395)
    at scala.tools.nsc.backend.jvm.GenBCode$BCodePhase$Worker1.run(GenBCode.scala:139)
    at scala.tools.nsc.backend.jvm.GenBCode$BCodePhase.buildAndSendToDisk(GenBCode.scala:381)
    at scala.tools.nsc.backend.jvm.GenBCode$BCodePhase.run(GenBCode.scala:350)
    at scala.tools.nsc.Global$Run.compileUnitsInternal(Global.scala:1412)
    at scala.tools.nsc.Global$Run.compileUnits(Global.scala:1397)
    at scala.tools.nsc.Global$Run.compileSources(Global.scala:1392)
    at scala.tools.nsc.Global$Run.compile(Global.scala:1486)
        <snip>
error: Error while emitting Broken.scala
```

I'm not too familiar with the resident compiler, do you have an idea what could cause the `NoSymbol` @retronym?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
soc: Looks like the field phase deletes all static fields under the mistaken assumption that all constant types don't need any storage...

soc: Also needed to adapt Constructors. It's kind of weird now, it looks something unlinked the fields, as the right tree is displayed in every phase, but when emitting the bytecode and retrieving the fields with `classFields(clasz)` it's not in the decls.

sjrd:
@djspiewak IMO only your second snippet should be valid. The first one has problems wrt scoping: the scopes are not the same in the Scala definition than they would be in a Java definition. This is disturbing from a user point of view, and will probably cause huge implementation difficulties as well.

We should aim for a syntax that at least respects scoping.

soc:
@sjrd In what sense? I would the static methods to the class in both cases. Whether the initializers refer to an enum constant instantiation within the enum class or to something outside shouldn't matter much. Although we could decide to not enter the static val's in the first case to make sure usage mirrors definition.

retronym:
Unfortunately we can't consider this for the 2.12 series due to the timing and the need to gather feedback on the design choices. This feedback should happen with a SIP. Hopefully having a working implementation will help to spur that discussion, though.

sjrd:
If I define an enum in Java like this:

``` java
package foo;

enum Weekday {
  Monday, Tuesday
}
```

then `foo.Monday` does not exist. It is only available as `foo.Weekday.Monday`. This matches the scoping behavior of the second snippet.

However, if I define it in Scala like this:

``` scala
package foo

@enum
sealed abstract class Weekday

case object Monday extends Weekday
case object Tuesday extends Weekday
```

then _in Scala_, `foo.Monday` exists (and `foo.Weekday.Monday` doesn't). If you compile that down to a Java enum, from Java's point of view `foo.Monday` does not exist but `foo.Weekday.Monday` does. The Java scoping and the Scala scoping don't match, which is bad from a language design point of view.

It would also probably be tricky to get the Scala compiler right when reading `Weekday` from a class file. The Scala scoping needs to be recreated somehow (although there is no top-level `Monday$.class`) and the Java scoping hidden away. This is asking for serious trouble.

Both from a language design point of view and from an implementation perspective, this snippet has issues. The following alternative has none of those issues:

``` scala
package foo

@enum
sealed abstract class Weekday

object Weekday {
  case object Monday extends Weekday
  case object Tuesday extends Weekday
}
```

soc:
@retronym Yes, 2.12 is a bit too early. I would have targeted that at 2.13 but the branch doesn't exist yet.

Regarding the SIP, I think I'll rather get the compiler fixes in, and split all user-visible changes into a macro annotation. This would be nice because with macro annotations I don't need all those crazy parser hacks.

Alternative would be that someone from the community (@djspiewak maybe?) who wants these changes steps up and handles the SIP, because I really don't have time to deal with the 30 comments from people–who didn't bother to read the proposal nor the implementation–about "did you try library X" (that doesn't address the requirements) and another 20 comments about "why not pick this syntax" (that doesn't work).

There is only is only so much stuff I can handle in parallel, and the turnaround in terms of weeks and months for PRs is already pushing it.

maxaf:
I've built an [`@enum` macro annotation](https://github.com/maxaf/numerato) in library form a while back. Its syntax is somewhat different from the one proposed here, and it doesn't target generating Java `enum`s, but I'm glad to see others working towards the same goal.

som-snytt:
Also big :+1:  for working implementation. No doubt the specification will invite lots of bikeshedding, so it's a brave undertaking. Possibly, generating consensus is the hard part, which suggests the feature would benefit from vetting as a library first. That's easier to do in a scaladex world, perhaps.

I hope we can call the feature Numerato. Wasn't that one of the X-Men?

`scala.Enumeration` doesn't actually deserve the bashing, and it ought to be rehabilitated in death with a headstone that reads: 

> Did One Thing Well
> Pushed the Envelope
> in a Time sans macros
> and path-dependent types
> 2.0 - 2.13

djspiewak:
@soc I'll have to bow out of SIP duty on this one.  Not because I think it's a bad idea, but because I simply don't get impacted personally by this functionality.  If anything, I'm more interested in deprecating `scala.Enumeration` than I am in providing support for compiling down to Java enums.  Like I said, I think the latter is an excellent idea, it just has literally no impact on me.  :-)

BTW, I've seen talk of syntax like this:

``` scala
@enum
class Stuff(val flag: Boolean) {
  One(true),
  Two(false)
}
```

I'm _really_ really really really opposed to this syntax.  It's basically introducing a brand new, non-orthogonal method for introducing types.  This is definitely concise, and it's very reminiscent of the Java syntax, but those are the only two things which really recommend it and I'm not convinced that those two properties are really worth all that much.

My proposal would be to start with the ADT-based syntax and then go from there.  Reminder: odds are very good that Dotty will be getting first-class syntax sugar for declaring ADTs (similar to what Scala had back in 1.x).  If and when that happens, I would expect that such syntax will find its way into NSC in the reasonably near future (or at least, near-er than the Dotty transition).

@sjrd  Good point on scoping.  The scoped ADT pattern is a lot better IMO (it's literally all I use in real code bases), so restricting `@enum` to just that kind of pattern works well I think.

dwijnand:
.. and please no more trailing comma problems... :^)

djspiewak:
@som-snytt 

> scala.Enumeration doesn't actually deserve the bashing, and it ought to be rehabilitated in death with a headstone that reads:

Ehhhhhh…  Even back in those days, we had significantly superior solutions to the enumeration problem.  For example: https://gist.github.com/viktorklang/1057513  As far back as 2.7.2, I was telling people that literally any and all use of the type was a mistake.

som-snytt:
If the enumeration problem is some named ints that can be enumerated, in minimal syntax, then that's what `scala.Enumeration` did. The link to the `Numerato` project agrees that the succinct syntax, a phrase I can't say ten times fast, was its best (only) feature. I don't mean to defend the old dead horse from further beating other than that. In the past year, I've used it for a list of command options to feed to the local Java command line parser. Look ma, no quoted strings. But of course, I defer to anyone who was born before 2.8 and experienced those troubled times.

Off the cuff, does SIP-23 help? If `MyEnum { val myVal }` has a singleton type that is the top-level `MyVal`, then maybe the scoping problem can be reconciled, in that Java sees `MyEnum.MyVal` and Scala sees them as the same type.

One can imagine an interpolator syntax `myEnum"myVal"` that produces the required literal. `import MyEnum._` to pick up the enrichment?  I'm just guessing.

lare96:
I made a thread about this on reddit about a month ago, great to see this PR!

soc:
@djspiewak The "new" syntax mainly exists because the proponent of `Enumeration` said that one-class-per-enum-constant is an unacceptable cost compared to `Enumeration`. Also, the ADT thing doesn't work at all currently (https://issues.scala-lang.org/browse/SI-7046). I hope Miles' fixes improve the situation, but I haven't compiled and tried his code yet.

To be honest, given proposals like SIP-commas, the proposed new extension keyword for ADTs or the general onslaught of new keyword proposals (`inline`, `local`) I'd rather drop the "new" syntax than to set a precedence for these things.

I think the right way forward for this without a SIP champion might be to submit the things that are pure compiler fixes, and delegate user-facing syntax additions to macro annotations.

PR #5353, opened 2016-08-19
SI-9896 Join operations at `MapLike`

pfcoperez:
This PR adds `innerJoin`, `leftJoin` and `rightJoin` methods to `MapLike` interface thus providing functionality analogous to Spark's `PairRDD' [join, leftOuterJoin, ...](http://spark.apache.org/docs/latest/api/scala/#org.apache.spark.rdd.PairRDDFunctions).

It isn't difficult to find Scala users needing to join maps in Scala ([e.g.](http://stackoverflow.com/questions/39031928/finding-the-totalspent-for-each-customer-in-scala/39032738#39032738)).

Besides, these operations could be easily used along side `TraversableLike`'s [groupBy](http://www.scala-lang.org/api/2.11.8/index.html#scala.collection.TraversableLike@groupBy[K]%28f:A=>K%29:scala.collection.immutable.Map[K,Repr]) to build powerful and concise algorithms.

### DIFF
--- a/src/library/scala/collection/MapLike.scala
+++ b/src/library/scala/collection/MapLike.scala
@@ -369,4 +369,36 @@ self =>
   override /*PartialFunction*/
   def toString = super[IterableLike].toString
 
+  /**
+    * Joins this map with a given one, keeping just entries with keys present at both maps and returning a new map
+    * with their combined values.
+    *
+    * @param that map to join with.
+    * @tparam VT the type of the other map values.
+    * @return a new map of tuple values with each map value for those keys present at both this map and the other.
+    */
+  def joinInner[VT](that: Map[K, VT]): Map[K, (V,VT)] =
+    for((k,va) <- this; vb <- that.get(k)) yield k -> (va, vb)
+
+  /**
+    * Joins this map with a given one, keeping entries with keys present at both maps or just `this` and returning a
+    * new map with the values at `this` and, optionally, their values at the right map.
+    *
+    * @param that map to join with.
+    * @tparam VT the type of the other map values.
+    * @return a new map of tuples of values from `this` and `None` or some values from the right map.
+    */
+  def joinOuterLeft[VT](that: Map[K, VT]): Map[K, (V,Option[VT])] =
+    for((k,va) <- this) yield k -> (va, that.get(k))
+
+  /**
+    * Joins this map with a given one, keeping entries with keys present at both or just the joined map and returning a
+    * new map with the values at `that` and, optinonally, their values at `this`
+    * @param that map to join with.
+    * @tparam VT the type of the other map values.
+    * @return a new map of tuples of `None` or some values from `this` and values from the right map.
+    */
+  def joinOuterRight[VT](that: Map[K, VT]): Map[K, (Option[V],VT)] =
+    for((k,vb) <- that) yield k -> (get(k), vb)
+
 }
--- a/test/junit/scala/collection/MapLikeTest.scala
+++ b/test/junit/scala/collection/MapLikeTest.scala
@@ -1,0 +1,47 @@
+package scala.collection
+
+import org.junit.Assert._
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+@RunWith(classOf[JUnit4])
+class MapLikeTest {
+
+  val ma = Map(0 -> " ", 1 -> "a", 2 -> "b")
+  val mb = Map(1 -> "A", 2 -> "B", 3 -> "C")
+
+  @Test
+  def joinInnerRightValues(): Unit = {
+
+    assertEquals(Map(1 -> ("a", "A"), 2 -> ("b", "B")), ma joinInner mb)
+
+  }
+
+  @Test
+  def joinOuterLeftRightValues(): Unit = {
+
+    val expected = Map(
+      0 -> (" ", None     ),
+      1 -> ("a", Some("A")),
+      2 -> ("b", Some("B"))
+    )
+
+    assertEquals(expected, ma joinOuterLeft mb)
+
+  }
+
+  @Test
+  def joinOuterRightRightValues(): Unit = {
+
+    val expected = Map(
+      1 -> (Some("a"), "A"),
+      2 -> (Some("b"), "B"),
+      3 -> (None     , "C")
+    )
+
+    assertEquals(expected, ma joinOuterRight mb)
+
+  }
+
+}


soc:
Minor nitpick: I'd prefer it if the names all started with `join...`. This makes discoverability in an editor much easier and groups related methods together.

pfcoperez:
@soc As you surely know, these operations are usually named after the English nominal group describing them and I was worried that changing the customary could be poorly regarded by the community.

However I fully agree with you remark regarding editor discoverability (for example, I believe part of the appeal of internal DSLs resides in them becoming a kind of "wizard" within an editor with indexing and auto-completing).
To start writing _join_ and auto-completing will present the available choices within this family of operations.

Hence, I propose the following names:
- `joinInner`
- `joinOuterLeft`
- `joinOuterRight`

What do you think?

In another vein, I've been thinking about adding a fourth operation to complete the family: `joinOuterFull` keeping keys present at `this` or `that` associated with tuples of two optional values.

soc:
Yes, these names look much better imho! :-)

Not sure how much time and effort you want to spend on this, but if other people agree with the general direction of adding these methods (it currently seems as if people want to avoid any large changes/additions considering the upcoming redesign of collections), it would make sense to investigate whether the ordered maps enable any kind of algorithmic optimizations for these operations.

pfcoperez:
@soc  No worries, I don't mind expending some time on this if it has any chance of being included. Moreover, that kind of optimizations might be a good exercise. I am open to any other suggestion.

Thanks for your time reviewing this PR.

pfcoperez:
@soc I've just realized that the suggested optimization for ordered maps could consist on applying a modified _merge_ phase of the _merge sort_ algorithm which would allow calculating any operation of the join family in `O(N)` (with N = number of keys of `this` + number of keys of `that`). Implementing that is quite easy.

I'll try to update the PR within next week.

Ichoran:
Regardless of the underlying algorithm, I wonder whether all these joins are really capturing the natural operation adequately.  Wouldn't it be more general to have

``` scala
def joinByKey[V2,R](that: Map[K, V2])(pf: PartialFunction[(Option[V1], Option[V2]), R]) =
  // The obvious implementation
```

which handles all the cases?  In particular

``` scala
  val inner = m1.joinByKey(m2){ case (Some(a), Some(b)) => (a,b) }
  val outerLeft = m1.joinByKey(m2) { case (Some(a), x) => (a, x) }
  val outerRight = m1.joinByKey(m2) { case (x, Some(b)) => (x, a) }
```

If you want an optimized version for the case where both keys exist, then I would call it `zipByKey` instead of `join`, because it has exactly the behavior of `zip` on indices (i.e. it truncates when you run out of either collection).

```
  def zipByKey[V2](that: Map[K, V2]): Map[K, (V, V2)] = ???
```

pfcoperez:
@Ichoran I might be wrong but, if I understood you correctly,  the `joinByKey` operation you suggest will require calculating the cartesian product of `this` and `that`. In time complexity terms, it is  like `joinOuterFull`.  From a full outer join result, the cartesian product of both Maps, it is possible to get the results of left, right outer and even inner join operations by just filtering each _entry_ (e.g: passing `pf` to `collect` over the product result: `this.product(that).collect(pf)`).

You can save time and memory if you limit your operation from the beginning. When you have an specific operation as, lets say, left outer join (`joinOuterLeft`) you know you don't need to calculate the cartesian product, just _iterate_ over `this` and lookup each key at `that`.

If `this` and `that` are two maps of sizes `N` and `M` respectively, then:
- Cartesian product would always take `O(N*M)`, no matter what optimization you perform. 
- On the other hand, if `that` were, for example, a `HashMap`, its lookup time would be effectively constant (eC _Programming in Scala_ notation) and the left join operation would effectively take `O(N)` .

That's why I think it is better to keep - at least three - four distinct join operations.

The scenario would be different if `joinByKey` internally analysed `pf` in order to determine its needs as to whether calculate the cartesian product or not.

Finally, I fully agree with your observation that inner join is `zipByKey` . Thus, the operation could be named `zipByKey`. I would keep `joinInner` as just another name for it for the reason exposed by @soc that its nice to have the whole set sharing a prefix for discoverability.

Ichoran:
I'm not suggesting the Cartesian product, just the union.  You can already get the Cartesian product from `for { (k,v) <- m; (k2, v2) <- m2 } yield f(k, k2, v, v2)`.

It is true that a full union considers keys in both maps instead of just one.  But the left and right outer join are obligated to create a tuple with an option when the next step might be to consume that and map it with a function or select it somehow.  If you care about efficiency _that_ much, you'll already be writing it by hand somehow to avoid that step.  Plus the outer left and right joins are not exactly hard to write already: `m.map{ case (k,v) => (k, (v, m2 get k)) }` and `m2.map{ case (k,v) => (k, (m get k, v)) }`.  (Or the comprehension versions, `for { (k,v) <- m } yield (k, (v, m2 get k))` etc..)

It's the inner and full outer joins that are interesting in that it's not completely straightforward to write them (especially the outer join); and in that when you have two things, merging via tupling them is not always what you want to do (c.f. `reduceByKey` which Scala's collections are also missing).

pfcoperez:
@Ichoran  Absolutely true, it is the union of keys what matters, I found myself momentarily forgetting  we were dealing with maps, in fact the first implementation I came with for `joinFullOuter` consisted on joining both maps key-sets and the performing lookups which (still in the context of the example I used above), would be `O(M+N)`.  Sorry for misinterpreting your comment.

Isn't merging via tupling enough, given the key uniqueness of maps?  Unless, of course, we had Multi Maps, which we have under the mutable package `mutable` as a trait. If I am not wrong, it is just an interface, for Maps whose values are Sets, to be used as Multi Maps. So it seems to me that, at the end of the day, merging by tupling could be a good - enought? - decision and that `reduceByKey` could be a method of the `MultiMap` trait.

I think, though this assumption could be too far fetched,  that most users performing joins would not iterate all over the map but perform lookup operations. In this context, there is no need for the additional step of decomposing the tuple of values, values-options, options-values or options-options.

I share your believe that implementing left and right joins is almost trivial but it is customary to rely on the collection library operations as the most optimal algorithms building blocks. It is so much that that using those method is usually better regarded than writing ad-hoc code since the latter has been reviewed by a more reduced community of users.

Ichoran:
Well, despite the motivation coming from Spark, I think it is more important to be internally consistent and to cover the right use cases.  I like the idea overall, but I'm uneasy about the implementation (even given the big advantage of existing over not-existing).

Some of this has to do with naming.  Scala already has a `zip` method, and this is exactly analogous except works by key instead of iteration order.  Calling it `join` seems peculiar: why use a different name for the same operation?

Some of it has to do with completeness. You've written an inner join, a left join, and a right join, but not a full outer join.  But the full outer join is actually the hardest to write in an efficient manner.  This doesn't seem like an ideal design.

Some of it has to do with how generic the operations are.  Given `mergeKeys[U, W](that: Map[K, U])(f: (V, U) = W): Map[K, W]` you can `m.mergeKeys(m2)(_ -> _)` to get tuples if that's what you want; given `zipKeys[U](that: Map[K, U]): Map[K, (V, U)]` you can `m.zipKeys(m2).mapValues{ case (v,u) => f(v,u) }` to get `mergeKeys`.  Given merging, zipping is easy (and efficient).  Given zipping, merging is more awkward and less efficient.  This suggests to me that the functional merge should be preferred (or both should be given).

I would certainly like to have something like this in 2.13 for the collections revamp.  It's getting pretty late in 2.12 to consider it for that, especially given that the best API isn't exquisitely clear.

Maybe a lengthier justification for this particular set of methods would help.  Or maybe it should go through a SLIP.  Or maybe someone else e.g. @SethTisue or @szeiger can comment on whether this sort of thing is, from what they've heard, a big enough pain point to just go ahead with something decent without worrying about how ideal it is.  I would tend to think that since 2.13 is a library-focused release anyway, and with 2.12 dropping Java 6 support it's better if 2.11 and 2.12 are more source-compatible, that this sort of thing is better left to 2.13.  (But right-biasing Either is a counterexample.)

SethTisue:
> this sort of thing is better left to 2.13

agree

adriaanm:
Yep -- 2.12.x has been in feature freeze for a while now. We're about to cut [2.12.0-RC1](https://github.com/scala/scala/milestone/47). FYI, we announce these things on scala-internals, with an overview at https://github.com/scala/scala/milestones.

szeiger:
For reference, here are the join operators in Slick: https://github.com/slick/slick/blob/master/slick/src/main/scala/slick/lifted/Query.scala#L62-L134

In addition to the standard SQL inner / left / right / full outer joins, we also have zip joins, which are modeled the same way in the AST. In the API `zip`, `zipWith` and `zipWithIndex` have their usual names and semantics. I don't think that you need a uniform API for zip joins and other joins. zip joins are foreign to database users, the other joins are foreign to FP.

Slick's join operators require a mapping function for the key (so you write something like `xs joinLeft ys on (_.id === _.id)` but I wouldn't do the same for collections where we could never use a hash index in these cases (which is what you usually expect from a join).

pfcoperez:
@Ichoran: 
- Concerning names: I agree on  `mergeKeys`, `zipKeys` ... being more consistent from a functional programming standpoint. I would be more than happy to change that in my PR for the sake of consistency.  But I wonder whether  it would be possible to have a "join" family interface for the sake of simplicity of use since most users are familiar with these operations after being exposed to relational databases. **This would be optional, I definitively find consistency has a greater priority**. 
- Concerning generic operations: Providing `mergeKeys` which could easily be used to implement `zipKeys` is neat and nice.
- Still, I keep thinking that operations with the notion of the master key set (left or right) would drive to further optimizations but, maybe, not worth of it. What about `mergeKeys[U, W](that: Map[K, U])(f: (V, U) => W)(masterKeySet: (Map[K, _], Map[K, _]) => Set[K]): Map[K, W]`? It could have a default parameter value for `masterKeySet` returning the union of keys thus resulting in your `mergeKeys`:

``` scala
mergeKeys[U, W](that: Map[K, U])(f: (V, U) => W)(
    masterKeySet: (Map[K, _], Map[K, _]) => Set[K] = { case (a,b) => a.keySet ++ b.keySet }
): Map[K, W]
```
- In relation to completeness: I wanted to start the discussion by providing a minimal set, in fact I've already mentioned I wanted to implement full outer join.

@szeiger I am not native English speaker, by your setence _I don't think that you need a uniform API for zip joins and other joins. zip joins are foreign to database users, the other joins are foreign to FP._, do you mean you would rather keep two sets of operations (one for users familiar to databases and other to those closer to functional programming) even though one API would call the other (presumably, the join interfaces implemented from FP operations)? 

@adriaanm Despite the fact that it is already too late to include this or anything related to this in Scala 2.12, will you keep the PR open so I can change my implementation following the conclusions of this discussion in the same context?

Thanks guys for your time reviewing and discussing this feature.

szeiger:
@pfcoperez I didn't mean that there should be two APIs for joins, but that a join API wouldn't necessarily have to be consistent with a zip API.

After thinking some more about this PR, I wonder what the typical use cases would be. In my experience with writing database code and working on Slick, a join is rarely what you want but it's the best you can get when you're limited to flat tuples of data.

Case in point: The StackOverflow question that you used as an example. The layout of the data and the way the question is phrased clearly point to a relational database background where "join" is the natural answer to this sort of question. But when you're not confined to relational agebra there's a more direct solution:

```
customers.map { case (id,_) => (id, purchases.get(id).map(_.map(_._3).sum).getOrElse(0)) }
```

pfcoperez:
@szeiger Thanks for making it clearer for me. Providing a dual interface, was just an optional suggestion in any case and now I am now convinced that the functional operation set would be more than enough.

Following the example ([StackOverflow answer](http://stackoverflow.com/a/39032738/1893995) ), I suggested the OP to use the `leftJoin` operation I quickly drafted as:

``` scala
...
    def leftJoin[B](b: Map[K, B], default: B): Map[K, (A,B)] =
        for((k,va) <- a; vb = b.getOrElse(k, default)) yield k -> (va, vb)
...
   customers leftJoin(purchases, Seq()) mapValues { case (_, l) => l.map(_._3).sum }
```

Which, if am not mistaken, is as writing: 

``` scala
        a.map { case (k,va) => k -> (va, b.getOrElse(k, default))  } mapValues {
            case (_, l) => l.map(_._3).sum 
        }
```

That is the same as the `map` (the higher order method equivalent to FP's `fmap` I mean) you wrote if it wasn't for the second iteration over the resulting Map which performs the last transformation. That adds some overhead so:  Yes, I do prefer the more generic `mergeKeys` @Ichoran has pointed out.

If we had it (with the addition on the master key selector)  [Note that all of this isn't an actual implementation, just a thought experiment] :

``` scala
def mergeKeys[U, W](that: Map[K, U])(f: (V, U) => W)(
    masterKeySet: (Map[K, _], Map[K, _]) => Set[K] = { case (a,b) => a.keySet ++ b.keySet }
): Map[K, W]
```

We could express the same solution as:

``` scala
customers.mergeKeys(purchases) { 
  case (_, Some(pur)) => pur.map(_._3).sum
  case _=> 0
 }   { case (a,_) => a.keySet }
```

Not performing the union, as we are providing a key set selector, and not adding overhead from additional transformations either. 

Then, `zipByKey` would naturally come as:

``` scala
def zipByKey[V2](that: Map[K, V2]): Map[K, (V, V2)] = mergeKeys(that)  {
  case (Some(v1), Some(v2)) => (v1, v2)
} ()
```

But I don't like that final parenthesis at the end so I'd rather implement the default key selector  with a second overloaded method or some other redesign.

szeiger:
I don't think `mergeKeys` with the extra `masterKeySet` carries its weight as an abstraction. By abstracting out both `f` and `masterKeySet`, there's no room for optimization, and by the time I've finally understood the complicated signature, I have already implemented a solution with a single `map` call, which ends up being shorter, too.

@Ichoran's original `mergeKeys` could make a worthwhile addition. I suggest calling it `zipByKeyWith`. Similar to how a (pairwise) `zipWith` (not present in Scala collections, but in Haskell, for example) leads to `zip` via `zipWith(_ -> _)`, `zipByKeyWith(_ -> _)` gives you `zipByKey` (inner join). And, like all `zip` operations, it discards any value that isn't matched on both sides:

``` scala
// Seq:
def zipWith[B, C](that: Seq[B])(f: (A, B) => C): Seq[C]
def zip[B](that: Seq[B]): Seq[(A, B)] = zipWith(that)(_ -> _)

// Map:
def zipByKeyWith[W, X](that: Map[K, W])(f: (V, W) => X): Map[K, X]
def zipByKey[W](that: Map[K, W]): Map[K, (V, W)] = zipByKeyWith(that)(_ -> _)
```

"merge" could be used for more generic operations that do not discard non-matching values:

``` scala
def mergeByKeyWith[W, X](that: Map[K, W](f: PartialFunction[(Option[V], Option[W]), X]: Map[K, X]
def mergeByKey[W](that: Map[K, W]): Map[K, (Option[V], Option[W])] = mergeByKeyWith(that) { case (a, b) => (a, b) }
```

This give you a consistent naming scheme encompassing inner and full outer joins plus the more abstract methods from which they are built. I wouldn't bother with left or right outer joins. They can be built with `mergeByKeyWith` where required but I'm still unconvinced that they are useful enough to add to the standard library.

ctongfei:
I propose adding an additional `fullOuterJoin`:

```
Map[K, A] fullOuterJoin Map[K, B] == Map[K, Either[A, B]]
```

Ichoran:
@ctongfei - That doesn't have the right type signature.  `Either` cannot simultaneously be inhabited by both `Right` and `Left`.

ctongfei:
@Ichoran Sorry.. `symmetricDiff` results in `Map[K, Either[A, B]]`.
`fullOuterJoin` results in `Map[K, (Option[A], Option[B])]`.

SethTisue:
Closing this particular 2.12.x PR. But this idea could certainly be revisited in a 2.13 context, after https://github.com/scala/collection-strawman either does or doesn't move forward (although we're pretty confident at this point it will).

PR #1359, opened 2012-09-19
SI-6407 Better defaults for text encoding

soc:
Traditionally there were two ways of getting a “default” text encoding:
- Assume UTF-8
- Use the underlying platform's encoding

We shouldn't invent an additional, third way to do that.

PR #5354, opened 2016-08-21
SI-6407 Default to UTF-8 for text encoding

soc:
Traditionally there were two ways of getting a “default” text encoding:
- Use UTF-8
- Use the underlying platform's encoding

We shouldn't invent an additional, third way to do that.

Instead we use UTF-8 by default, and encourage people to be explicit.
Most of the code written out there is never designed with arbitrary
encoding changes via `-Dfile.encoding` or other means in mind, and only
causes pain, suffering and a larger testing surface for developers.

Deprecates the `Codec.ISO8859` constant, because the number is NOT
optional. "ISO-8859" is like "UTF" ... it's not meaningful and doesn't
denote a specific encoding.

Also fixes code that used `toLowerCase` or `toUpperCase` without a
locale. This is dangerous as code and scalac itself will behave
differently under some locales, namely azeri, lithuanian and turkish.

### DIFF
--- a/src/compiler/scala/tools/nsc/ClassPathMemoryConsumptionTester.scala
+++ b/src/compiler/scala/tools/nsc/ClassPathMemoryConsumptionTester.scala
@@ -48,7 +48,7 @@ object ClassPathMemoryConsumptionTester {
     println("Memory consumption can be now measured")
 
     var textFromStdIn = ""
-    while (textFromStdIn.toLowerCase != "exit")
+    while (textFromStdIn.toLowerCase(java.util.Locale.ENGLISH) != "exit")
       textFromStdIn = readLine("Type 'exit' to close application: ")
   }
 
--- a/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
@@ -158,7 +158,7 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
   lazy val javaUnboxMethods: Map[InternalName, MethodNameAndType] = {
     ScalaValueClassesNoUnit.map(primitive => {
       val boxed = boxedClass(primitive)
-      val name = primitive.name.toString.toLowerCase + "Value"
+      val name = primitive.name.toString.toLowerCase(java.util.Locale.ENGLISH) + "Value"
       (classBTypeFromSymbol(boxed).internalName, methodNameAndType(boxed, newTermName(name)))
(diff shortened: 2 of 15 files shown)


soc:
Resubmit of https://github.com/scala/scala/pull/1359.

Ichoran:
I didn't check everything, but I did quickly check to make sure that specifying locale information doesn't slow anything down.  (It doesn't seem to.)

soc:
@Ichoran It should actually be faster than before, because the JDK doesn't need to figure out the default locale (which might load additional baggage we don't need like all the unicode locale-specific case mapping stuff).

It's just very annoying that doing "the right thing" needs more effort and is easy to forget. That's basically the design I want to get away from with this patch.

soc:
Would it be possible to get this into the next RC? This would give it a bit more time for testing than a point release. There might be some obscure, broken code out there which might be impacted by this fix.
I'd rather see this now than having it slip through the cracks and breaking things in a point release.

SethTisue:
provisionally retargeted, but @adriaanm should make the final call here

adriaanm:
@soc, I really appreciate your work, but we closed the RC1 queue weeks ago. The PRs that are still open have been in the queue for weeks or months, except for a few tiny ones that have 0 risk of causing regressions. Your PR comes as a surprise to me, and we can't afford to take the necessary time to assess the risk before cutting the RC. 

Your improvement is a perfect fit for the library cleanups in 2.13. As soon as 2.12.0 is out, we'll branch for 2.13.x and I would be happy to see this PR retarget that branch!

mintty:
Two comments:

> We shouldn't invent an additional, third way to do that.

Also, this should be handled consistently. Currently, parameters for `scalac` and `scala` are different (`-encoding` vs `-Dfile.encoding=`). Last time I tried with 2.11 on Windows, even the default assumptions were different; `scalac` defaulted to UTF-8 while the runtime system would default to the Windows' ANSI encoding; what a mess!

> - Use the underlying platform's encoding

When using Scala from cygwin, the additional problem arises that we are having two platforms, with unfortunately different paradigms to set up the encoding (`chcp` vs `locale`/LC_*).
There are two trivial patches to the wrapper scripts that fix this problem for cygwin.
See https://issues.scala-lang.org/browse/SI-9744

mintty:
I'd like to submit my patch proposal explicitly as an issue but I don't find the wrapper scripts (`bin/scalac` and `bin/scala`) and their repository...

SethTisue:
@mintty The wrapper scripts are in this repo; see the files `src/compiler/scala/tools/ant/templates`, invoke by the `mkScalaTool` calls in the sbt build.

PR #5355, opened 2016-08-21
Italicize more defining occurrences of terms in Scala lang. spec.

dsbos:
The Scala language specification partially follows the style of italicizing the defining occurrence of a term.  However, it does not do so consistently.

Commit ...3459 italicizes a lot more of those defining occurrences.  (It does not cover all terms.)


dsbos:
How do I get the Scala CLA check to refresh? (I signed it, but the check above [edit: that was above and is now below] for it is still showing a red X.)

dsbos:
How did a test fail when the only thing I modified was non-code documentation?  Was the master branch already broken?

adriaanm:
Your changes are based on a commit from Dec 22, 2015, which is likely why this is not passing the test suite (I'm guessing because the failing test is sensitive to a jdk upgrade?) 

In any case, thanks for the polish! Since this is purely a spec update, we can merge it without the test suite.

PR #5356, opened 2016-08-23
Switch remaining uses of ant over to sbt

szeiger:
This is the first half of #5311, i.e. the final stages of switching build scripts over to sbt, short of actually remvoving the ant build.
- Modify `tools/scaladoc-diff` to use sbt instead of ant.
- Move `stability-test.sh` from `tools` to `scripts`. With the new
  build process without separate `locker` and `strap` stages, it doesn’t
  make sense to call this script without first setting up the proper
  test environment in a CI build.
- Replace the use of `build.number` in `bootstrap` with a new
  `SHA-NIGHTLY` mode for `baseVersionSuffix`.
- Make `partest` call sbt instead of ant for initializing the classpath
  and use the new classpath location (`quick` instead of `pack`).


test/partest
```diff
@@ -57,11 +57,11 @@ fi
 # Let ant construct the classpath used to run partest (downloading partest from maven if necessary)
```
lrytz: ant -> sbt

lrytz:
LGTM

PR #3992, opened 2014-09-22
Add composition and subst methods to =:= and <:<

puffnfresh:
Both scalaz.Leibniz and scalaz.Liskov have useful methods which are not
present in Predef.

https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Leibniz.scala

https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Liskov.scala

I don't see a reason why some of the methods should not be present in
scala-library itself, other than the implementations of these methods
have to be done carefully, since Predef uses the singleton trick. You
can read how scalaz makes a different trade-off on the Typelevel blog:

http://typelevel.org/blog/2014/07/06/singleton_instance_trick_unsafe.html

This commit adds the following:
- <:<#compose
- <:<#andThen
- <:<#subst
- =:=#symm
- =:=#compose
- =:=#andThen
- =:=#subst

This means equality is now an equivalence relation:
- Reflexivity (already existed)
- Symmetry
- Transitivity

I find these very useful when trying to do proofs of equality in Scala.

---

Carefully reviewing this code is important, all implementations are casts.

@djspiewak

@S11001001

PR #5359, opened 2016-08-25
Add methods for manual proof

Atry:


sjrd:
Is there any previous discussion/rationale/motivation about this?

Atry:
Not yet.

https://github.com/mpilquist/simulacrum will benefit from this PR. The `@typeclass` macro can use `implicitly[A =:= X].invariant[F].apply(fa)`, instead of `fa.asInstanceOf[F[X]]` at the moment.

soc:
I think we had a PR like this before. might make sense to compare and discuss why this is supposed to be better.

Atry:
@soc I could not find the PR. Could you paste it here?

Atry:
Looks like https://github.com/scala/scala/pull/3992

Atry:
There are only three basic atomic rules `asymmetry`, `invariant` and `covariant`.

 `andThen`, `compose` and all other helpers can be implemented from `invariant` and `covariant`.

`invariant` and `covariant` are especially important for type safety, because the two methods give a chance to manually prove subtyping or equality. They are not necessary to be `implicit` though.

Ichoran:
It would really help to have more extensive justification, documentation, and a full set of tests.

You don't use contravariant or asymmetry at all.  What problem do they solve?  What behavior ought they have?  Do they actually have that behavior?

(Some of them look wrong to me.)

SethTisue:
the test failure is spurious and should go away if you rebase on top of current head

SethTisue:
@Atry rebase?

adriaanm:
We can't change the public API of 2.12 anymore, targeting for 2.13.

Atry:
@adriaanm There is not such a 2.13.x branch yet

Atry:
Regardless the ability of covariance/contravariance of introduced by this PR, there is another issue make `<:<` less useful - type applying with type constraints.

``` scala
trait MyTypeClass[+A]
final class MyTypeClassForByteSeq[+A <: Seq[Byte]] extends MyTypeClass[A]

// Does not work because `<:<` is meaningless for type constraints
def myTypeClassForByteSeq[A0](implicit ev: A0 <:< Seq[Byte]): MyTypeClass[A0] = new MyTypeClassForByteSeq[A0]
```

SethTisue:
Closing for 2.12.x. 2.13.x branch exists now, if you want to make a new PR (or reopen and retarget this one).

PR #5362, opened 2016-08-27
increase stack size when running JUnit tests

SethTisue:
`scala.collection.immutable.PagedSeqTest.test_SI6615` was failing
intermittently during PR validation

I verified that this change fixes the problem by:
- changing the PageSize constant in PagedSeq (as shown by
  Jason Zaugg) until the test failed every time locally
- making this build change and seeing the test pass again

5M is just an arbitrary number, considerably over the default (which
varies according to platform & CPU). 5M is a lot of stack, but not so
vastly much that it appreciably eats into the heap

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -556,6 +556,7 @@ lazy val junit = project.in(file("test") / "junit")
   .settings(disablePublishing: _*)
   .settings(
     fork in Test := true,
+    javaOptions in Test += "-Xss5M",
     libraryDependencies ++= Seq(junitDep, junitInterfaceDep, jolDep),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-v"),
     testFrameworks -= new TestFramework("org.scalacheck.ScalaCheckFramework"),


SethTisue:
I'm going to go ahead and merge this since all PR's are affected.

SethTisue:
@adriaanm writes:

> I'm not sure we should be running all our tests under such an unusually big stack. I'm concerned that would mask regressions in other tests.

shall we try 2M for a while? 1M? ANT_OPTS had 1M

adriaanm:
I'd rather we disable a single test (or, ideally, fix the underlying issue in the impl) than change the behavior of the whole test suite. Could you revert this change and just disable `scala.collection.immutable.PagedSeqTest.test_SI6615` for now? Let's open a ticket against 2.12.1 to make the impl stack safe and revive the test.

adriaanm:
The stack size is already 1M: https://github.com/scala/scala-jenkins-infra/blob/master/attributes/default.rb#L7

retronym:
Does Sbt use that env variable for forked processes?

adriaanm:
Aha! Sounds like that explains all our observations. So, instead of disabling the test, let's begin with reducing the stack size to 1M.

SethTisue:
> Let's open a ticket against 2.12.1 to make the impl stack safe

note that PagedSeq is deprecated and almost certainly nobody has ever used it except the scala-parser-combinators folks (see https://github.com/scala/scala-parser-combinators/issues/71), so it isn't worth putting any effort into

PR #5363, opened 2016-08-27
SI-9892 Avoid pointless list creations

soc:
JIRA: https://issues.scala-lang.org/browse/SI-9892

### DIFF
--- a/src/library/scala/collection/immutable/List.scala
+++ b/src/library/scala/collection/immutable/List.scala
@@ -243,6 +243,19 @@ sealed abstract class List[+A] extends AbstractSeq[A]
     else this drop lo take (until - lo)
   }
 
+  override def tails: Iterator[List[A]] = {
+    new AbstractIterator[List[A]] {
+      var list: List[A] = List.this
+      def hasNext: Boolean = list != null
+      def next(): List[A] = {
+        val nextList = list
+        if (list.isEmpty) list = null
+        else              list = list.tail
+        nextList
+      }
+    }
+  }
+
   override def takeRight(n: Int): List[A] = {
     @tailrec
     def loop(lead: List[A], lag: List[A]): List[A] = lead match {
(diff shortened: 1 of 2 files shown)


SethTisue:
Stream has the same bug:

``` scala
scala> val s = Stream.from(1)
s: scala.collection.immutable.Stream[Int] = Stream(1, ?)

scala> s.tails.toStream
res6: scala.collection.immutable.Stream[scala.collection.immutable.Stream[Int]] = Stream(Stream(1, ?), ?)

scala> res6.head eq s
res7: Boolean = false

scala> res6.head
res8: scala.collection.immutable.Stream[Int] = Stream(1, ?)

scala> s
res9: scala.collection.immutable.Stream[Int] = Stream(1, ?)
```

which suggests that the problem might be better tackled at the LinearSeq level... what do you think?

soc:
I'll have a look, I usually try to stay away from Stream as far as possible.

(Btw, wouldn't this change behavior if evaluation caused a side-effect?)

SethTisue:
One other question, is the custom iterator definition necessary? How about just `Iterator.iterate(this)(_.tail)`?

soc:
@SethTisue 

```
scala> Iterator.iterate(List(1,2,3))(_.tail).toList
java.lang.UnsupportedOperationException: tail of empty list
  at scala.collection.immutable.Nil$.tail(List.scala:417)
  at scala.collection.immutable.Nil$.tail(List.scala:412)
```

soc:
@SethTisue Just had a look, the issue is that `immutable.LinearSeq` is pretty much an empty trait. I'm not sure it's the right thing to add random methods to it.

SethTisue:
@soc with `.takeWhile(_.nonEmpty)`, then. or maybe that's off by one at the end, I don't know offhand — I'm just sketching here

paulp:
@SethTisue Yes since there's no `takeWhileAndAlsoThatOneWhereItWentNotWhile` you could write `Iterator.iterate(xs)(_.tail).takeWhile(_.nonEmpty) ++ Iterator(Nil)`.

PR #5364, opened 2016-08-29
SI-9823 Collections perf: favor virtual call over instanceof

retronym:
This avoids concurrent usages of collections in NUMA architectures
from falling of a performance cliff due to an implementation detail
of interface instanceof checks in HotSpot JVM.

The VM contains a one element cache in the metadata of each class
to record the most recent successful test for a interface.

For example:

```
classOf[Serializable].isAssignableFrom(classOf[Some[_]])
```

Would, in addition to returning `true`, set:

```
classOf[Some[_]]._secondary_super_cache = classOf[Serializable]
```

This is done to hide the fact that interface tests are O(N),
where N is the number of inherited interfaces.

This scheme is discussed in "Fast Subtype Checking for the
HotSpot JVM" (Click, Rose) [1]

However, if this cache repeatedly misses, not only are we
exposed to the linear search of the secondary super type array,
but we are also required to write back to the cache. If other
cores are operating on the same cache line, this can lead to
a significant slowdown ("cache thrashing"). This effect will
by most (or only?) visible on multi socket machines.

The pathological case is:

```
scala> Iterator.continually(List(classOf[Product], classOf[Serializable])).flatten.take(100).map(intf => intf.isAssignableFrom(classOf[Some[_]])).count(x => x)
res19: Int = 100
```

Which, if run on a multi-socket machine, should be much slower
than:

```
scala> (Iterator.continually(classOf[Product]).take(50) ++ Iterator.continually(classOf[Serializable]).take(50)).map(intf => intf.isAssignableFrom(classOf[Some[_]])).count(x => x)
res20: Int = 100
```

This commit avoids a interface test in a hot path in the collections
by instead using virtual dispatch to differentiate between
IndexedSeqLike and other collections. HotSpot will still use some
shared bookkeeping ("inline cache" [2]) at the callsites of
this method, but these stabilize in the megamorphic usage and no
longer force expensive cache line synchronization.

[1] https://www.researchgate.net/publication/221552851_Fast_subtype_checking_in_the_HotSpot_JVM
[2] https://wiki.openjdk.java.net/display/HotSpot/PerformanceTechniques

### DIFF
--- a/src/library/scala/collection/GenTraversableOnce.scala
+++ b/src/library/scala/collection/GenTraversableOnce.scala
@@ -96,6 +96,12 @@ trait GenTraversableOnce[+A] extends Any {
    */
   def size: Int
 
+   /** The size of this $coll if it is can be cheaply computed
+   *
+   *  @return    the number of elements in this $coll, or -1 if the size cannot be determined cheaply
+   */
+  protected[collection] def sizeHintIfCheap: Int = -1
+
   /** Tests whether the $coll is empty.
    *
    *  Note: Implementations in subclasses that are not repeatedly traversable must take
--- a/src/library/scala/collection/IndexedSeqLike.scala
+++ b/src/library/scala/collection/IndexedSeqLike.scala
@@ -92,4 +92,6 @@ trait IndexedSeqLike[+A, +Repr] extends Any with SeqLike[A, Repr] {
     copyToBuffer(result)
     result
   }
+
+  override protected[collection] def sizeHintIfCheap: Int = size
 }
--- a/src/library/scala/collection/mutable/Builder.scala
+++ b/src/library/scala/collection/mutable/Builder.scala
@@ -65,18 +65,18 @@ trait Builder[-Elem, +To] extends Growable[Elem] {
   /** Gives a hint that one expects the `result` of this builder
    *  to have the same size as the given collection, plus some delta. This will
    *  provide a hint only if the collection is known to have a cheap
-   *  `size` method. Currently this is assumed to be the case if and only if
-   *  the collection is of type `IndexedSeqLike`.
-   *  Some builder classes
-   *  will optimize their representation based on the hint. However,
+   *  `size` method, which is determined by calling `sizeHint`.
+   *
+   *  Some builder classes will optimize their representation based on the hint. However,
    *  builder implementations are still required to work correctly even if the hint is
    *  wrong, i.e. a different number of elements is added.
    *
    *  @param coll  the collection which serves as a hint for the result's size.
    */
   def sizeHint(coll: TraversableLike[_, _]) {
-    if (coll.isInstanceOf[collection.IndexedSeqLike[_,_]]) {
-      sizeHint(coll.size)
+    coll.sizeHintIfCheap match {
+      case -1 =>
+      case n => sizeHint(n)
     }
   }
 
@@ -94,8 +94,9 @@ trait Builder[-Elem, +To] extends Growable[Elem] {
    *  @param delta a correction to add to the `coll.size` to produce the size hint.
    */
   def sizeHint(coll: TraversableLike[_, _], delta: Int) {
-    if (coll.isInstanceOf[collection.IndexedSeqLike[_,_]]) {
-      sizeHint(coll.size + delta)
+    coll.sizeHintIfCheap match {
+      case -1 =>
+      case n => sizeHint(n + delta)
     }
   }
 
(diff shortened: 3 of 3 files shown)


retronym:
Review by @sjrd @Ichoran 

This is an an alternative version of #5351 that has a smaller delta to the collections API, by just adding `sizeHint` that is `-1` for non cheap sizes.

adriaanm:
I'm not a performance expert, but LGTM on the correctness front.

src/library/scala/collection/GenTraversableOnce.scala
```diff
@@ -96,6 +96,12 @@ trait GenTraversableOnce[+A] extends Any {
    */
   def size: Int
 
+   /** The size of this $coll if it is can be cheaply computed
+   *
+   *  @return    the number of elements in this $coll, or -1 if the size cannot be determined cheaply
+   */
+  protected[collection] def sizeHint: Int = -1
```
sjrd: I'm not sure about restricting the visibility of this method. Supposedly, if I create my own type of collection, I might want to call `sizeHint` on another collection? Or should I only ever call the methods `sizeHint(coll)` & co.?

sjrd: Btw, using `sizeHint` for the name of this method doesn't look like a good idea to me:
- It does not have the same meaning as the other methods called `sizeHint` (which are _also_ in the hierarchy): this one is a "getter" of sorts, while the others are "setters".
- It creates an overload between a `()`-less `def` and other defs _with_ `()`. I've always found this was not very clean (and, IIRC, it's not even allowed in dotty anymore)

Can we find another name?

adriaanm: `protected def sizeIfCheap`?

soc:
`sizeInfo`? Assuming we don't want to introduce some kind of more sophisticated design as in psp?

src/library/scala/collection/GenTraversableOnce.scala
soc: `sizeInfo`? Assuming we don't want to introduce some kind of more sophisticated design as in psp?

sjrd: > `protected def sizeIfCheap`

Sounds good to me.

@soc `sizeInfo` doesn't convey the "optionality" or "inaccuracy" of what it does, IMO.

retronym:
`sizeHintOf` ?

src/library/scala/collection/GenTraversableOnce.scala
soc: I think it has done a good job in this use-case: https://github.com/non/psp-view/blob/master/core/src/main/scala/SizeInfo.scala#L8

But yeah, `sizeIfCheap` is fine, at least we can add something more advanced with `sizeInfo` later.

retronym: > I'm not sure about restricting the visibility of this method.

I was mostly being conservative about offering this for public consumption when it is designed to be used in the collection/builder protocol.

adriaanm:
Let's go with `protected def sizeIfCheap: Int`, since there isn't much more information that we can convey with an `Int` anyway (i.e., it's either equal to `size` or a marker to denote `size` is not "cheap" -- happy to pick another adjective, but I can't think of any and I don't think it's terribly crucial since it's internal).

retronym:
`sizeIfCheap` it is, then. I've left it `protected[collection]` for now (it needs to be accessible from within `Builder#sizeHint(TraversableLike)`, so protected is insufficient. @sjrd was asking whether it should in fact be public. While it would be okay as a public method, I'm hesitant to do this without a good name.

adriaanm:
LGTM

sjrd:
LGTM as well.

src/library/scala/collection/GenTraversableOnce.scala
```diff
@@ -96,6 +96,12 @@ trait GenTraversableOnce[+A] extends Any {
    */
   def size: Int
 
+   /** The size of this $coll if it is can be cheaply computed
```
analytically, line 99: Small typo in the comment.

SethTisue: @analytically thanks, I'll get it in my next batch of typo fixes

retronym:
The underlying performance problem in the JVM is tracked as [JDK-8180450](https://bugs.openjdk.java.net/browse/JDK-8180450)

He-Pin:
optimization in netty: https://github.com/netty/netty/pull/12709

PR #5366, opened 2016-08-29
SI-6967 Primitive ClassTag.unapply is removed

som-snytt:
Follow-up to remove the overloads, which is source
compatible, in favor of unapply(Any).

There ought to be a `@removed(since = "2.12", signature = "")` annotation for folks who go looking for removed API.

Removal approval at https://github.com/scala/scala/pull/5347#issuecomment-243129126


soc:
@som-snytt Could you create a ticket for `@removed`? :-)

szeiger:
LGTM

PR #5367, opened 2016-08-29
Ensure trait var accessor type is widened

adriaanm:
If we don't widen, we'll fail to find the setter when
typing `x = 42`, because `x` is constant-folded to `0`,
as its type is `=> Int(0)`. After widening, `x` is
type checked to `x` and its symbol is the getter in the
trait, which can then be rewritten to the setter.

Regression spotted and test case by @szeiger.


szeiger:
LGTM

PR #5369, opened 2016-08-30
Fixes to mixin forwarders

lrytz:


src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -243,11 +243,17 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
...
+                if (owner.isJavaDefined && owner.isInterface && !clazz.parentSymbols.contains(owner)) {
+                  val text = s"Unable to implement a mixin forwarder for $member in $clazz unless interface ${owner.name} is directly extended by $clazz."
+                  if (required) reporter.error(clazz.pos, text)
+                  else warning(clazz.pos, text)
+                } else
+                  cloneAndAddMixinMember(mixinClass, member).asInstanceOf[TermSymbol] setAlias member
               }
 
-              if (settings.XgenMixinForwarders) genForwarder()
+              if (settings.XgenMixinForwarders) genForwarder(required = false)
```
sjrd: I think this will erroneously report a warning instead of an error if both of the following apply:
1. `-XgenMixinForwarders` is used, _and_
2. the forwarder is required for correctness.

I would expect an error to be raised every time a forwarder is required for correctness, regardless of the value of `settings.XgenMixinForwarders`.

sjrd:
:+1: for the commit about JUnit. I think it's a much better experience to offer those by default, even though it does feel like burning library-specific things in the compiler.

src/compiler/scala/tools/nsc/transform/Mixin.scala
lrytz: good point, i have to change the order of tests

lrytz:
will re-open

lrytz:
About the last commit: by default the compiler would now generate mixin forwarders whenever a class inherits a JUnit-annotated default method from a trait. For

```
trait T {
  @Test def t(): Unit = { ...}
}
class C extends T
```

we now get (by default):

```
public interface T {
    @Test
    default public static /* synthetic */ void t$(T $this) { $this.t(); }

    @Test
    default public void t() { }

    default public static void $init$(T $this) { }
}
public class C implements T {
    @Test
    @Override
    public void t() { T.t$(this); }
    public C() { T.$init$(this); }
}
```

This is certainly good for everyone using JUnit 4, which will ignore the default method and the static method (both annotated `@Test`) in T, and it will only execute `C.t`.

I did a quick test with the current master branch of JUnit 5 (https://github.com/junit-team/junit5). First of all there's a new `@Test` annotation in package `org.junit.jupiter.api`, but in some ways JUnit 5 depends on / supports JUnit 4, so the old `@org.junit.Test` might also be available and work, not sure. In any case, for the above source code, JUnit 5 executes the test `t()` in both cases, i.e., with and without a mixin forwarder. It doesn't (try to) run the static `t$` method.

So the current behavior (add forwarders for JUnit-annotated methods) seems to be fine also for JUnit 5, at least for its current state - not sure how things may change until their release.

Note that we don't emit forwarders when using the new `@org.junit.jupiter.api.Test`, which is fine: the annotation only exists in JUnit 5, which doesn't need the forwarder.

If we want to be more careful, instead of emitting the forwarders by default, we could provide a triple-choice command-line flag `-Yjunit-default-methods:<warn | gen-forwarders | nothing>` and set the default to `warn`.

adriaanm:
Sounds like the right approach to me as well. Small nit about the option choices if we do that: true|false|warn is easier to remember

lrytz:
> Sounds like the right approach to me as well

Which, to only warn by default?

lrytz:
Also, should it be a `-Y` option? Any thoughts on naming are welcome :)

adriaanm:
Yes to all the above :-) the main use case will be junit, which is done by
default, right? The only way to turn that off will be to explicitly pass
false to the option that I shall not name :-)

lrytz:
> Yes to all the above

OK, so "Yes, it should only warn, but not emit forwarders by default" and "yes, it should be a `-Y` option".

> the main use case will be junit

This is the only use case. We're not talking about emitting mixin forwarders in general, just for JUnit-annotated methods.

> which is done by default, right?

Well, if we set the default to warn, it will only warn.

adriaanm:
Ok then I misunderstood. I thought the default was to mix in forwarders for
test methods and warn for problems when we can't be sure to invoke the
right super method. So, yes to that. I interpreted 'true' to mean mix in
all forwarders and don't warn, and 'false' results in the minimal set of
forwarders, thus no junit test methods and no warnings.

lrytz:
OK. To recap, currently there are two independent flags:
-  `-Xgen-mixin-forwarders` enables generating forwarders **always**. In case a forwarder cannot be generated, but the resulting bytecode is still correct, there's a warning
- `-Yjunit-trait-methods-no-forwarders` disables generating forwarder for JUnit methods - by default they are generated.

Whenever a forwarder is required but cannot be generated, there's an error.

My proposal was to possibly extend the `-Y` option to a triple-choice (as explained above), and possibly set the default to `warn`. This is only about the JUnit case.

Another possibility, which is probably where you're getting at, would be to have one single option `-XYZmixin-forwarders` with the following choices:
- `all` to generate all mixin forwarders, warn for those that cannot be emitted but are not required
- `junit` to generate only forwarders for JUnit methods, error if one cannot be emitted
- `junit-warn` to warn about JUnit methods in traits, but generate only forwarders required for correctness
- `required-only` to only generate forwarders required for correctness

In all cases, if a required forwarder cannot be emitted, there's an error.

The default could be any of the above - to be decided. We might still go for `all` because of the performance impact (www.scala-lang.org/blog/2016/07/08/trait-method-performance.html), otherwise we could go for `junit` or `junit-warn`, I don't feel strongly about this.

adriaanm:
Yeah, I think a single option would be best. `-Xmixin-force-forwarders`?
- `true`: all forwarders
- `false`: minimal set of forwarders
- `junit`: between `true` and `false` -- only ones we know are needed for JUnit 4

Thinking about it, how useful is `warn`? We know it's broken, and our users might not, so they won't know to turn on the option. Maybe we can just enable it implicitly, and disable when `-Xmixin-force-forwarders` is set explicitly (implying the user knows what's going on and doesn't need a warning).

lrytz:
> how useful is warn? We know it's broken

It's only broken for JUnit 4, when JUnit 5 arrives Scala 2.12 will still be around. OTOH, the forwarders should not hurt for JUnit 5 (as I tested above). So I think the `warn` level only makes sense if we put it as default. As an explicit option, I don't think it's too useful. The potential use case would be: somebody is using JUnit 4 and wants to make sure he's not writing any test methods in traits.

If we do have the warning, I'd prefer it to be an explicit option. The implicit may be a bit confusing.

lrytz:
The test failures in my commits are `PagedSeqTest.test_SI6615 failed: java.lang.StackOverflowError`. I rebased on current 2.12.x - is that expected? @szeiger

adriaanm:
Yeah apparently 1m of stack is not always enough. Jason has a commit in his
PR to just disable the test.

lrytz:
OK, I think I'm done with the discussed changes, PTAL!

src/compiler/scala/tools/nsc/transform/Mixin.scala
```diff
@@ -243,51 +243,59 @@ abstract class Mixin extends InfoTransform with ast.TreeDSL {
...
+                case _ => false
               }
+
+              def generateJUnitForwarder: Boolean = {
+                settings.mixinForwarderChoices.isJunit &&
+                  member.annotations.nonEmpty &&
+                  JUnitAnnotations.exists(annot => annot.exists && member.hasAnnotation(annot))
+              }
+
+              if (existsCompetingMethod(clazz.baseClasses) || generateJUnitForwarder)
```
adriaanm, line 295: So, ignoring the annotations, `generateJUnitForwarder` is `false` iff `-Xmixin-force-forwarder:false`? When no value is supplied, when it's `true`, and when it's `junit` we get an error when annotations are present.

lrytz: For JUnit methods, if `-Xmixin-force-forwarders` is either `true` or `junit` (which is the default, i.e., no value supplied), we generate the forwarder with `required = true`, i.e., there's an error if it cannot be generated.

If `-Xmixin-force-forwarder` is `false` we don't generate the JUnit forwarder.

adriaanm:
LGTM

lrytz:
pushed a small cleanup to the help texts of ChoiceSetting (fdfdd253bfb4e44faf4afce1da56989ce132e15c)

PR #5371, opened 2016-08-31
SI-9906: override ListBuffer.last/lastOption to run in O(1) time

chrisokasaki:
Also update scaladocs for those two methods.

review by @Ichoran

### DIFF
--- a/src/library/scala/collection/mutable/ListBuffer.scala
+++ b/src/library/scala/collection/mutable/ListBuffer.scala
@@ -386,6 +386,25 @@ final class ListBuffer[A]
     this
   }
 
+  /** Selects the last element.
+   *
+   *  Runs in constant time.
+   *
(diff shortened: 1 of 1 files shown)


Ichoran:
LGTM

SethTisue:
thanks Chris!

mosyp:
[2.11](https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/mutable/ListBuffer.scala) & [2.13](https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/mutable/ListBuffer.scala) versions are missing the fix. Is there anyway to port it?

NthPortal:
I don't think 2.11 is likely to get this (or almost any other) fix at this point. It should definitely be ported to 2.13.

lrytz:
https://github.com/scala/scala/pull/7386

PR #5373, opened 2016-09-01
SI-9909: corrected stream example so it does not give forward reference

TimWSpence:
JIRA: https://issues.scala-lang.org/browse/SI-9909

### DIFF
--- a/src/library/scala/collection/immutable/Stream.scala
+++ b/src/library/scala/collection/immutable/Stream.scala
@@ -23,7 +23,7 @@ import scala.language.implicitConversions
  *  import scala.math.BigInt
  *  object Main extends App {
  *
- *    val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }
+ *    lazy val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }
  *
  *    fibs take 5 foreach println
  *  }
@@ -46,7 +46,7 @@ import scala.language.implicitConversions
  *  import scala.math.BigInt
  *  object Main extends App {
  *
- *    val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(
+ *    lazy val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(
  *      fibs.tail).map(n => {
  *        println("Adding %d and %d".format(n._1, n._2))
  *        n._1 + n._2
@@ -162,7 +162,7 @@ import scala.language.implicitConversions
  *  // The first time we try to access the tail we're going to need more
  *  // information which will require us to recurse, which will require us to
  *  // recurse, which...
- *  val sov: Stream[Vector[Int]] = Vector(0) #:: sov.zip(sov.tail).map { n => n._1 ++ n._2 }
+ *  lazy val sov: Stream[Vector[Int]] = Vector(0) #:: sov.zip(sov.tail).map { n => n._1 ++ n._2 }
  *  }}}
  *
  *  The definition of `fibs` above creates a larger number of objects than


TimWSpence:
I've signed the CLA now

som-snytt:
I commented on the ticket. It's confusing, but the proposed change is extraneous or misleading, and doesn't explain it to me.

TimWSpence:
I'm totally open to other options eg removing the fibonnacci example entirely or making it a def instead of a val but the example as it currently stands in the docs does not compile. What do you suggest?

som-snytt:
Well, it does compile. There's a difference between a member definition and a local definition, see the SO linked from the ticket. Since the Scaladoc is talking about the difference between val and def ("The definition of `fibs` is a `val` not a method"), I think the explanation of what is used where should be easy enough so even I can understand it.

The problem of "I added braces and now the example is broken" isn't the fault of Stream, but it's clearly a goal that the doc shouldn't set us up to fail.  I'm not sure adding lazy to code that works is optimal for that.

I would suggest footnoting the example with the info on SO and make sure it's clear when lazy is required.

Also remove the import of BigInt.

We're discussing this difference, I hope:

```
scala> :pa

object Main extends App {

  val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }

  fibs take 5 foreach println
}

defined object Main

scala> Main main null
0
1
1
2
3

scala> def f() = {
     | val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }
     | (fibs take 5).toList
     | }
<console>:12: error: forward reference extends over definition of value fibs
       val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }

```

xevix:
It works in the Scala console when pasting, but does not seem to compile on my system via sbt or scalac.

```
$ sbt compile
[info] Loading global plugins from /Users/xevix/.sbt/0.13/plugins
[info] Loading project definition from /Users/xevix/fibs/project
[info] Set current project to fibs (in build file:/Users/xevix/fibs/)
[info] Compiling 2 Scala sources to /Users/xevix/fibs/target/scala-2.11/classes...
[error] /Users/xevix/fibs/src/main/scala/Fibs.scala:617: forward reference extends over definition of value fibs
[error]     val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }
[error]                                                            ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
[error] Total time: 4 s, completed Sep 1, 2016 10:52:32 PM
```

som-snytt:
```
$ cat streamex.scala

package streamex

object Main extends App {

  val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }

  fibs take 5 foreach println
}
$ scalac streamex.scala && scala streamex.Main
0
1
1
2
3
$ scala -version
Scala code runner version 2.11.8 -- Copyright 2002-2016, LAMP/EPFL
```

TimWSpence:
There's definitely some weirdness here. I tried running it in a new file in Intellij and had no problems. However, adding 

```
@Test
def foo(): Unit = {
    val fibs: Stream[BigInt] = BigInt(0) #:: BigInt(1) #:: fibs.zip(fibs.tail).map { n => n._1 + n._2 }

    fibs take 5 foreach println
}
```

to StreamTest.scala gives me a compiler error "forward reference extends over definition of value fibs"

som-snytt:
You have not read the SO links I added to the ticket. Also please re-read my previous comment.

SethTisue:
I think I would suggest simply changing `val` to `lazy val`, since `lazy val` will work in any context, and not including any explanation at all. It's Stream doc, it can't include coverage of every language issue it touches.

SethTisue:
even the `object Main extends App` doesn't seem necessary to me to include

xevix:
@som-snytt Ah, I missed those, sorry. Yeah, it seems an Object with a main() method fails, but an Object that extends App works fine.

Side note: 2.11.7 and 2.11.8, same results.

som-snytt:
I didn't mean to take an egg-stream position, as Egghead would have put it.

The same confusion between a member and local def would hound REPL examples.

Also, `lazy` is an accepted way to deal with init order issues, so slapping on `lazy` is not itself lazy. As Wallace tells Gromit, "It's amazing what a lick of paint can do."

SethTisue:
> Side note: 2.11.7 and 2.11.8, same results

the difference between the scoping of val members vs. local vals in a block goes back at least to Scala 2.7, probably farther:

```
Welcome to Scala version 2.7.7.final (Java HotSpot(TM) 64-Bit Server VM, Java 1.6.0_65).

scala> object O { val x: Int = x + 1 }
defined module O

scala> O.x
res0: Int = 1

scala> { val x: Int = x + 1; x }
<console>:5: error: forward reference extends over definition of value x
       { val x: Int = x + 1; x }
                      ^
```

it's as spec'ed.

SethTisue:
@TimWSpence interested in getting this across the finish line?

TimWSpence:
@SethTisue Yes! Thanks for the reminder. Did we decide to just leave the lazy and remove the added comment?

SethTisue:
that's what I'm suggesting, yeah

TimWSpence:
@SethTisue Done :)

SethTisue:
I just noticed the code is repeated down on line 49, I think that one needs updating too...?

TimWSpence:
Thanks @SethTisue fixed that as well. I changed at least one other similar problem I spotted as well

SethTisue:
haha nice catch!

SethTisue:
thanks Tim!

PR #5374, opened 2016-09-02
More elegant holders for local lazy vals.

adriaanm:
Follow up for #5294 based on @retronym's feedback.

I'm going to defer teaching the optimizer about stack allocating the lazy holders: https://github.com/scala/scala-dev/issues/215

### DIFF
--- a/src/library/scala/runtime/LazyRef.scala
+++ b/src/library/scala/runtime/LazyRef.scala
@@ -1,52 +1,157 @@
+/*                     __                                                      *\
+**     ________ ___   / /  ___     Scala API                                   **
+**    / __/ __// _ | / /  / _ |    (c) 2002-2016, LAMP/EPFL and Lightbend, Inc **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/                      **
+** /____/\___/_/ |_/____/_/ | |                                                **
+**                          |/                                                 **
+\*                                                                             */
+
 package scala.runtime
 
-/* Classes used as holders for local lazy vals */
+/** Classes used as holders for lazy vals defined in methods. */
 
 class LazyRef[T] {
-  var value: T = _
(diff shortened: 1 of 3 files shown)


lrytz:
LGTM!

retronym:
LGTM

src/library/scala/runtime/LazyRef.scala
```diff
@@ -1,52 +1,157 @@
...
+
 package scala.runtime
 
-/* Classes used as holders for local lazy vals */
+/** Classes used as holders for lazy vals defined in methods. */
 
 class LazyRef[T] {
-  var value: T = _
-  @volatile var initialized: Boolean = false
+  @volatile private[this] var _initialized: Boolean = _
```
viktorklang, line 14: Why is this @volatile btw?

adriaanm: Quoting @retronym:
> In double checked locking, the write to the value is in a synchronized block
> the fast path (already initialized) just needs to volatile read on the initialized field. if that is set, we know the value was safely published

viktorklang: Apologies, I guess what threw me of course is that this class (or rather these classes) are public, and non-final, classes which cannot be used properly without sprinkling of "magic"/macros. Are these intended to be possible to use *directly* by users?

adriaanm:
True, the runtime package is not meant for public consumption. Good point
that they should likely be final. They have to be public though, as they
are referenced from compiler-generated code.

src/library/scala/runtime/LazyRef.scala
adriaanm: True, the runtime package is not meant for public consumption. Good point
that they should likely be final. They have to be public though, as they
are referenced from compiler-generated code.

https://github.com/scala/scala-dev/issues/339

PR #5376, opened 2016-09-05
Make -Xexperimental features available separately

milessabin:
I've removed all the code which is guarded by a bare -Xexperimental flag setting. In the case of the things that were covered by `enableTypeVarExperimentals` the status of all tests remains the same, so I think that the functionality it was intended to enable has become part of the default case.

I have also disabled the code supporting match virtualization ... I'd be happy to put it back under a `-Yvirtpatmat` flag.

And I've removed the disabling of the handling of SAMs where 2.11.x source code compatibility has been requested. Again, I would be happy to put this back under a `-Ysam` flag, but I'm not convinced that partial downgrades make a lot of sense, so I would prefer to just remove this functionality.


adriaanm:
But what about the users of the improved type inference for existentials, say?  We'd also like to preserve the escape hatch to opt out of SAM conversion for older compilers. We can discuss refinements of this PR during 2.12.x, as we would like to have a good experience for `-Xexperimental`. The 2.12.0 release is not the right time for this.

We have one more tweak to make that was discovered by the community build (https://github.com/scala-js/scala-js/issues/2580#issuecomment-244802306). This is the only kind of change we will make during the whole 2.12.0 RC cycle: fixing bugs that prevent use of non-experimental language features and that do not have a reasonable workaround. We look forward to getting back to the regular swing of subsequent 2.12.1 and up releases, but we _have_ to get 2.12.0 out first!

milessabin:
On improved type inference for existentials: no tests have been affected by this change ... see my commit message. Do you have any particular test in mind which would fail if this was no longer supported? Either this is now covered by the default case or there is no test coverage for it at all at present.

On the Scala 2.11.x compatibility front, yes, I guess this should go back. I will need to make similar changes in TLS for 2.11.x, and it's really there that an explicit flag for SAMs becomes an issue.

You didn't mention the match virtualization issue ... what are your thoughts on that for > 2.12.0?

I understand your desire to get 2.12.0 out the door, however, if `-Xexperimental` continues to be a "mystery meat" option, then we should **not** recommend that people use it in preference to `-Ypartial-unification` and that should be reflected in the release notes.

adriaanm:
Ok, I'm happy with that compromise. I'll add this PR to the release notes label to remind ourselves.

milessabin:
I'm not quite sure what the compromise is? :-D

adriaanm:
I prefer recommending -Xexperimental. You prefer merging this PR for RC1. We'll do neither for 2.12.0

milessabin:
Gotcha ... yes, that's a good compromise :-)

milessabin:
Rebased.

milessabin:
Squashed and rebased.

milessabin:
Rebased.

@adriaanm ... are there any changes you'd like to see before merging?

adriaanm:
I'm sad to see my various experimental features go. I know of at least some users of the improved type inference for existentials, and virtualized pattern matching is kind of neat too, wouldn't you say? I don't think we can just drop these and rely on the test suite to catch this, sadly -- our test coverage is far from 100%.

milessabin:
Understood, but this did result in problems for someone enabling `-Xexperimental` in preference to `-Ypartial-unification`. Do you consider these things mature enough to make non-experimental? If not, do you (or anyone else who's invested in them) have time to get them there?

The virtualization stuff is for @TiarkRompf and @namin mainly, isn't it? We can ask them if they still need it.

milessabin:
Here's the "mystery meat" issue we had with `-Xexperimental` in TLS: https://github.com/typelevel/scala/issues/113. I'm not sure exactly what feature is being enabled by `-Xexperimental` which is responsible for that, so it's not clear to me if the failure is a bug or an intended change. Either way, it would be unfortunate to pull in something with an unclear status as a side effect of encouraging people to use `-Xexperimental` to enable features that they definitely do want.

At a minimum we need to be able to turn off undesirable features. For that we would need explicit `-Y` options for each feature which is currently guarded by bare `Xexperimental`. I'm happy to do that if you agree it's the right way to go.

milessabin:
Another option might be to separate out a `-Xearly-access` which is a strict subset of `-Xexperimental`.

It occurs to me that the Scala virtualized infrastructure will probably never be enabled by default, so I think it probably warrants a flag of its own, independent from `-Xexperimental`.

adriaanm:
I propose we add Y flags for the esoteric stuff like virtpatmat. The type
inference changes should be safe under experimental. Do you have reports of
it breaking something?

milessabin:
@adriaanm it turns out the issue I linked to earlier only fails with `-Xexperimental` on 2.11.x.

Can you point to an explanation of what the type inference changes actually are?

adriaanm:
The code had gotten a bit overfactored, but generally I think the comment explained it well:

> When comparing two types containing skolems, remember the highest level
> of skolemization. If that highest level is higher than our initial
> skolemizationLevel, we can't re-use those skolems as the solution of this
> typevar, which means we'll need to repack our inst into a fresh existential.

Apparently we'll need to fix https://issues.scala-lang.org/browse/SI-5729 before it can be enabled by default, but it should eventually become default behavior.

milessabin:
Do you have small example which illustrates the difference with the flag enabled vs. not?

milessabin:
I have to be honest, I don't think I could recommend anyone to use `-Xexperimental` if we know up front that it enables a bug which looks like it could have an impact in not particularly improbable code.

If we know that we have to fix SI-5729 before it can be enabled by default, can we either fix SI-5729 now, or defer the inclusion of this in `-Xexperimental` until it is fixed?

milessabin:
@dwijnand suggested (as a way to find out about real-world impact),

> One way would be if you created a PR to the community build enabling it by default and requested a Jenkins build of your PR. Might find you some real world cases.

Which I think is a great idea. Unfortunately it won't work because `-Xexperimental` enables _all_ the features it covers, not just this one. If nothing else, I think this is an argument in favour of having a flag to enable the feature individually even if we only expect and encourage people to enable it via `-Xexperimental`.

dwijnand:
That's true.

Given this PR is about `-Xexperimental` in general, perhaps a community build run with `-Xexperimental` would be informative.

milessabin:
We know that will fail because the SI-2712 fix will break projects which already have unapply-like workarounds for SI-2712.

retronym:
Perhaps try `-Xexperimental -Ypartial-unification:false ...`

```
⚡ printf ":settings\n" | scala -Xexperimental -Ypartial-unification:false
Welcome to Scala 2.12.0 (OpenJDK 64-Bit Server VM, Java 1.8.0_112-release).

scala> :settings
-Xexperimental = true
-Ypartial-unification = false
-d = .
-deprecation = true
-encoding = UTF-8
-feature = true
-nowarn = false

scala> :quit
```

milessabin:
@retronym true. Good suggestion.

But hey, just look at that ... isn't it helpful to be able to use flags to isolate the effects of individual features ... just sayin' ...

milessabin:
Can someone point me to instructions for how to setup/request a community build with a set of flags enforced globally?

dwijnand:
Change https://github.com/scala/community-builds/blob/93ffc15cf9c164d0fe6ccead2621f73c3a88d38b/community.dbuild#L6, PR it, requesting a Jenkins job start.

SethTisue:
also possible to run locally with `scalac_opts=whatever ./run.sh`

milessabin:
I've now managed to complete a community build with `-Xexperimental -Ypartial-unification:false` (yes, it really did take me this long) and it's successful, meaning that @adriaanm's type inference improvements are safe, at least to the extent that the community build is representative.

However, I'm firmly in favour of all features under `-Xexperimental` either being individually controllable, or becoming default. Given that the type inference changes appear to be safe, I'd like to see them made default as part of this `-Xexperimental` cleanup.

If that's deemed too risky, then I think it's unreasonable to deny prospective users of `-Xexperimental` the option of turning it off individually (other than by continuing to not use `-Xexperimental` and only enabling the features they can enable individually ... which is the opposite of what we're trying to achieve here).

I'll also reinstate the virtpatmat code under a `-Y` flag as @adriaanm suggested.

milessabin:
I've taken a close look at [SI-5729](https://issues.scala-lang.org/browse/SI-5729), [SLS 6.26.3](http://scala-lang.org/files/archive/spec/2.11/06-expressions.html#overloading-resolution) and concluded that the implementation of [`asSpecificAs`](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/typechecker/Infer.scala#L808-L833) matches the spec, hence that the ambiguity reported in SI-5729 is the correct behaviour, not a bug: the spec provides no justification for preferring the polymorphic `join[S]` over the monomorphic (albeit existential) `join`.

So I've enabled the improved inference by default and moved t5729.scala from pos to neg.

milessabin:
I've also left the 2.11.x source code compatibility handling of SAMs alone.

milessabin:
It seems that [Dotty agrees](https://gitter.im/lampepfl/dotty?at=583c447b16207f7b0eb5ea3b) that the SI-5729 case should be ambiguous.

adriaanm:
I ran out of time (for today) while convincing myself that that is true. Here's my sketch of a counter-example, where I've tried to reduce method application to assignment:

```
trait C[+X]

object Test {
  def join(in: C[_]): Int = ???
  def join[S](in: C[S]): String = ???
  trait Equiv { 
    type S 

    val alt1: C[S] = alt2 // can't supply a C[_] when a C[S] was expected for universally quantified S
    val alt2: C[_] = alt1 // a C[S] with unknown S can pass as the existentially quantified C[_]
  }

  join(null: C[_]) // should pick the polymorphic one, since it's more specific
}
```

Abbreviating the spec to the parts we need:
# Overloading Resolution

An alternative A is _more specific_ than an alternative B if
A is as specific as B and B is not as specific as A:

- A parameterized method m of type `(p_1:T_1, ... , p_n:T_n)U` is _as specific as_ some other member m'
	if m' is applicable to arguments `(p_1 , ... , p_n)` of types T_1 , ... , T_n.
- A polymorphic method of type `[a_1 >: L_1 <: U_1 , ... , a_n >: L_n <: U_n]T` is _as specific as_ some other member of type S
  if T is as specific as S under the assumption that each a_i is an abstract type name bounded from below by L_i and from above by U_i for i = 1 , ... , n.

	
	
# Function Applications
An application `f(e_1 , ... , e_m)` applies the function `f` to the argument expressions `e_1, ... , e_m`. For this expression to be well-typed, the function must be *applicable* to its arguments, which is defined next by case analysis on f's type.

If f has a method type `(p_1:T_1 , ... , p_n:T_n)U`, each argument expression e_i is typed with the corresponding parameter type T_i as expected type. Let S_i be the type of argument e_i (i = 1 , ... , m). The function f must be _applicable_ to its arguments e_1, ... , e_n of types S_1 , ... , S_n. We say that an argument expression e_i is a _named_ argument if it has the form `x_i=e'_i` and `x_i` is one of the parameter names `p_1, ..., p_n`.

Once the types S_i have been determined, the function f of the above method type is said to be applicable if all of the following conditions hold:
  - for every named argument p_j=e_i' the type S_i is [compatible](03-types.html#compatibility) with the parameter type T_j;
  - for every positional argument e_i the type S_i is [compatible](03-types.html#compatibility) with T_i;
  - if the expected type is defined, the result type U is [compatible](03-types.html#compatibility) to it.

If f is a polymorphic method, [local type inference](#local-type-inference) is used to instantiate f's type parameters.
The polymorphic method is applicable if type inference can determine type arguments so that the instantiated method is applicable.

milessabin:
@adriaanm you've missed out,
> A member of any other type is always as specific as a parameterized method or a polymorphic method.

Which applied to this case I read as saying that an applicable monomorphic method is as specific as a polymorphic method. And the implementation seems to directly reflect that ... matches on `MethodType` appear before matches on `PolyType` in the ranking function.

milessabin:
For comparison,
```scala
Welcome to Scala 2.12.0 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> object C {
     |   def join(in: Seq[List[Any]]): Int = 23
     |   def join[S](in: Seq[List[S]]): String = "foo"
     | }
defined object C

scala> C.join(Seq[List[Int]]())
<console>:13: error: ambiguous reference to overloaded definition,
both method join in object C of type [S](in: Seq[List[S]])String
and  method join in object C of type (in: Seq[List[Any]])Int
match argument types (Seq[List[Int]])
       C.join(Seq[List[Int]]())
         ^
```
I believe that the case reported in SI-5729 should be handled exactly the same as this one (which it is after this PR).

adriaanm:
> A member of any other type is always as specific as a parameterized method or a polymorphic method.

this case does not apply here, since we're comparing only `PolyType` and `MethodType`. The quoted case above is a base case for when non-method members are thrown in the mix.

milessabin:
How about my example where the existential is replaced by `Any`?

milessabin:
I think that the first clause is enough to argue that the monomorphic method is as specific as the polymorphic method (and hence that these two should be ambiguous), _if_ your improved type inference for existentials is in play. That rule says that a method _m_ is as specific as a method _m'_ if _m'_ is applicable to an argument list typed as the parameters of _m_.

So consider,
```scala
scala> object C {
     |   def join(in: Seq[List[_]]): Int = 23
     |   def join[S](in: Seq[List[S]]): String = "foo"
     |   def join2[S](in: Seq[List[S]]): String = "foo"
     | }
defined object C
```
Without `-Xexperimental` this condition is (unintuitively, incorrectly IMO) not met because of a failure of type inference. We can see this without being distracted by ambiguity by attempting this with `join2`,
```scala
scala> C.join2(Seq[List[Int]](): Seq[List[_]]): String
<console>:13: error: no type parameters for method join2: (in: Seq[List[S]])String exist so that it can be applied to arguments (Seq[List[_]])
 --- because ---
argument expression's type is not compatible with formal parameter type;
 found   : Seq[List[_]]
 required: Seq[List[?S]]
       C.join2(Seq[List[Int]](): Seq[List[_]]): String
         ^
 found   : Seq[List[_]]
 required: Seq[List[S]]
       C.join2(Seq[List[Int]](): Seq[List[_]]): String
```
On the other hand, with `-Xexperimental` the same application of `join2` is valid,
```scala
scala> C.join2(Seq[List[Int]](): Seq[List[_]]): String
res0: String = foo
```
which I assume is the main objective of your change.

But from this it immediately follows that the first clause of the rule is met: `join[S]` is applicable to argument lists which are typed as the parameters of `join`, so `join` is at least as specific as `join[S]`.

So the ambiguity due to enabling `-Xexperimental` is, correctly, due to the polymorphic method becoming more widely applicable due to improved type inference.

adriaanm:
Ok, you convinced me. Here's how I see it:

```scala
class Test {
trait T[x] // OR `class T[+x]` OR `class T[-x]`

// if these two definitions compile...
def joinX(in: T[_]): String = joinP(in)
def joinP[S](in: T[S]): String = joinX(in)

def joinO(in: T[_]): Int = 1
def joinO[S](in: T[S]): String = "a"

// ... this overloaded call is ambiguous, since the definitions above 
// imply neither is more specific (since they are "inter-applicable")
// (if joinX/joinP do not compile, over overload is more specific than 
// the other, and the call below should type check)
joinO(null)
}
```

adriaanm:
Thanks for pushing this through, @milessabin!

milessabin:
Awesome ... thanks!

SethTisue:
thanks Miles, this is a really nice cleanup.

retronym:
Regarding discussions about `isAsSpecific`, I'd like to cross reference some problem I found with the implementation that I tried to fix, but wasn't able to without causing too much breakage. In the end, I've punted that work to  https://github.com/scala/scala-dev/issues/207.

```scala
trait M1[A]; trait M2[A] extends M1[A]

trait N1[A[_]]; trait N2[A[_]] extends N1[A]

object Test {
  type D = DummyImplicit
  def m[A](implicit d: D): M1[A] = ???
  def m[A](implicit d1: D, d2: D): M2[A] = ???
  m // not ambiguous, M1[_] <:< M2[_] (type params of the methods are existentially abstracted in `isAsSpecificValueType`)

  def n[A[_]](implicit d: D): N1[A] = ???
  def n[A[_]](implicit d1: D, d2: D): N2[A] = ???
  n // was not ambiguous in 2.11:
    // existentialAbtraction(A :: Nil, N2[A]) did not notice that `N2[A]` contained
    // `A` (because it was eta expanded to [_]A[_] before matching for `case TypeRef(_, ASymbol, _)`)
}

```

I'm not sure if whether it is relevant to this discussion or not.

milessabin:
@adriaanm @retronym can you elaborate on "too much breakage"? I was able to successfully run a community build with this enabled, and it pretty clearly _is_  in line with the spec.

milessabin:
Also I see no difference in behaviour on @retronym's new example in 2.12.x with the inference improvements enabled or disabled.

PR #5377, opened 2016-09-05
SD-143 allow super calls to methods defined in indirect super classes

lrytz:
The restriction for super calls to methods defined in indirect super
classes introduced in a980fde was over-restrictive. In particular, it
ruled out a valid code pattern to select a method from a superclass
when an unqualified `super` would resolve to an override defined in a
trait (example in the commit as a test).

See https://github.com/scala/scala-dev/issues/143


adriaanm:
LGTM

PR #5378, opened 2016-09-06
SI-9913 Lead span iterator finishes at state -1

som-snytt:
Even if no elements fail the predicate (so that the trailing
iterator is empty).

Review by @Ichoran 

JIRA: https://issues.scala-lang.org/browse/SI-9913

### DIFF
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -648,15 +648,15 @@ trait Iterator[+A] extends TraversableOnce[A] {
      * handling of structural calls. It's not what's intended here.
      */
     class Leading extends AbstractIterator[A] {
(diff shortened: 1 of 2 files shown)


som-snytt:
I noticed leading.status goes toward negative, while trailing.status goes toward positive.

It would be handy if one could introduce a private trait of constant value defs (inlined constants) and then for the compiler not to emit the trait class. I would name the constants if it were free. Or if it didn't emit the field at least for `private[this] final val K = 42`.

szeiger:
LGTM

SethTisue:
thanks Som!

Ichoran:
@som @adriaanm - Sorry, still swamped.  This one slipped through the cracks.  I am not sure how you get cracks in a swamp, but there you go.

PR #5379, opened 2016-09-06
Store buildcharacter.properties in scala-compiler.jar

szeiger:
In addition to all the individual projects’ version properties files
that we already keep in `scala-compiler.jar` we now write a new
`scala-buildcharacter.properties` which is identical to the
`buildcharacter.properties` written to the root directory by
`generateBuildCharacterPropertiesFile`. The new task
`extractBuildCharacterPropertiesFile` extracts it from the bootstrap
Scala compiler and writes the usual `buildcharacter.properties`.

This can be used to reproduce the exact version information for all
modules in builds that start from an arbitrary published Scala version
instead of being triggered directly by a bootstrap job.

(see https://github.com/scala/community-builds/issues/274 for context)


SethTisue:
LGTM

PR #5380, opened 2016-09-06
remove outdated ENSIME info

SethTisue:
as suggested by @fommil

if up-to-date info on hacking Scala using ENSIME becomes
available in the future, better we simply link from our readme to that
info on the ENSIME site where it can be more easily updated


fommil:
http://ensime.org/contributing/scalac/

SethTisue:
@fommil thanks, I've pushed a revised commit that adds that link to README.md

SethTisue:
see also https://github.com/ensime/ensime.github.io/pull/162

SethTisue:
review by @lrytz?

lrytz:
LGTM

PR #5368, opened 2016-08-30
SD-208 Restore 2.11 names for arrayOps implicits

retronym:

PR #5381, opened 2016-09-07
bump STARR to 2.12.0-RC1

SethTisue:
because PR #5368 is binary-incompatible

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -8,7 +8,7 @@
 # The scala version used for bootstrapping. This has no impact on the final classfiles:
 # there are two stages (locker and quick), so compiler and library are always built
 # with themselves. Stability is ensured by building a third stage (strap).
-starr.version=2.12.0-M5
+starr.version=2.12.0-RC1
 
 # These are the versions of the modules that go with this release.
 # These properties are used during PR validation and in dbuild builds.
@@ -19,7 +19,7 @@ starr.version=2.12.0-M5
 #  - After 2.x.0 is released, the binary version is 2.x.
 #  - During milestones and RCs, modules are cross-built against the full version.
 #    So the value is the full version (e.g. 2.12.0-M2).
-scala.binary.version=2.12.0-M5
+scala.binary.version=2.12.0-RC1
 
 # external modules shipped with distribution, as specified by scala-library-all's pom
 scala-xml.version.number=1.0.5


SethTisue:
note that this is slightly different from what @adriaanm suggested on 5368, where he wrote:

> I was thinking we could use the nightly produced from this PR as starr

@milessabin calls foul [on Gitter](https://gitter.im/scala/contributors?at=57cfc59dcdbf820f7fa37c24):

> I really think a pre-RC1 starr should have been published. Requiring RC1 for RC1 is surely not a good thing

but if we had published a pre-RC1 starr, it would have been built from the literally the same commit as RC1 was built from, namely 4fa4b8302bffb2f35cfdaa6015ada5f8e047564d. so that seems like a distinction without a difference.

Miles: here's how I'd put it. the STARR for RC1 itself remains M5. this PR is not itself part of RC1; it's immediately post-RC1. (and in fact, the `v2.12.0-RC1` tag does not include 7507765)

adriaanm:
Exactly. Thanks for explaining, Seth

milessabin:
Binary, but not morally equivalent I think. It's not now possible to start from 0 and roll forward building released compilers from incrementally published binaries.

Maybe that's not a big deal, and maybe it's happened before, but nevertheless on balance I think it would have been better stick to a process which can be replicated without manual intervention.

milessabin:
Also,

> here's how I'd put it. the STARR for RC1 itself remains M5

Tests fail with M5 as STARR.

lrytz:
Test don't fail in a full bootstrap build. It's just that in our CI testing, we don't re-build partest, but make a liberal assumption about binary compatibility. We use the partest version compiled against the current STARR together with a compiler/library binary compiled from the current 2.12.x branch. Since we're in pre-.0-mode, there may be binary incompatible changes. In a bootstrap build, partest (and the other modules) are re-built.

milessabin:
OK, I don't think the sky is falling, but I do think that together with https://github.com/scala/scala-dev/issues/201 this suggests that making the process of updating STARR much smoother so that it can be done very easily at every relevant juncture would be a good plan.

adriaanm:
I really don't think there's a problem here. Why couldn't you perform the
same actions as we did now in this hypothetical future where someone wanted
to bootstrap from zero? How far back does zero lie, by the way? The first C
compiler? Java 1? The obscure Java dialect Scala 1 was written in?

In any case, we're always looking to improve our tooling. Everything is in
the open and we welcome contributions that align with our design goals.
Happy to discuss those.

SethTisue:
the coverage of these issues in the current README.md is minimally adequate, perhaps, but it could be made a lot better by folding in stuff from this discussion and from https://github.com/scala/scala-dev/issues/201, not to mention https://groups.google.com/d/topic/scala-internals/gp5JsM1E0Fo/discussion (which the readme currently links to)

SethTisue:
I agree that talking/thinking about "from zero" isn't helpful. there is no "from zero" anymore, it's just about getting from n - 1 to n.

adriaanm:
Right. Sorry about the snarky tone. I should've just said what Seth said.

milessabin:
From zero probably wasn't helpful.

I have had two (actually three) concrete issues with the lack of an updated STARR, recently,
- The roadblock I hit reported in https://github.com/scala/scala-dev/issues/201
- Today attempting to rebase my two main outstanding PRs against 2.12.x after 2.12.0-RC1 was published but before this PR was merged. This resulted in failures when running tests.

I've also had a third problem ... after following @adriaanm's `publishLocal`/`local.sbt`/`sbt -Dstarr.version` advice I've on several occasions kicked off `partest` runs only to realize 15 or so mins in that `local.sbt` or the SBT command line option was missing or something similar and had to abort and restart.

None of this is incredibly serious, but it is a bump on the road which if you want to have more community participation it would be good to have smoothed out.

adriaanm:
Agreed, those annoyances sound very familiar :-) only good routine can save
us here.

I should have submitted the bump Starr PR as soon as I tagged RC1.

Another tweak I've been thinking about is to include the starr version in
the synthetic local version number to make it easier to switch between
branches with different starrs.

PR #5384, opened 2016-09-08
SI-9915 Utf8_info are modified UTF8

som-snytt:
Use DataInputStream.readUTF to read CONSTANT_Utf8_info.

This fixes reading embedded null char and supplementary chars.

JIRA: https://issues.scala-lang.org/browse/SI-9915

### DIFF
--- a/test/files/run/t9915/C_1.java
+++ b/test/files/run/t9915/C_1.java
@@ -1,0 +1,18 @@
+
+public class C_1 {
+    public static final String NULLED = "X\000ABC";
+    public static final String SUPPED = "𐒈𐒝𐒑𐒛𐒐𐒘𐒕𐒖";
+
+    public String nulled() {
+        return C_1.NULLED;
+    }
+    public String supped() {
+        return C_1.SUPPED;
+    }
+    public int nulledSize() {
+        return C_1.NULLED.length();
+    }
+    public int suppedSize() {
+        return C_1.SUPPED.length();
+    }
+}
--- a/test/files/run/t9915/Test_2.scala
+++ b/test/files/run/t9915/Test_2.scala
@@ -1,0 +1,12 @@
+
+object Test extends App {
+  val c = new C_1
+  assert(c.nulled == "X\u0000ABC")    // "X\000ABC"
+  assert(c.supped == "𐒈𐒝𐒑𐒛𐒐𐒘𐒕𐒖")
+
+  assert(C_1.NULLED == "X\u0000ABC")  // "X\000ABC"
+  assert(C_1.SUPPED == "𐒈𐒝𐒑𐒛𐒐𐒘𐒕𐒖")
+
+  assert(C_1.NULLED.size == "XYABC".size)
+  assert(C_1.SUPPED.codePointCount(0, C_1.SUPPED.length) == 8)
+}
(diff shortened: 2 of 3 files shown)


adriaanm:
LGTM!

test/files/run/t9915/C_1.java
```diff
@@ -0,0 +1,18 @@
+
+public class C_1 {
+    public static final String NULLED = "X\000ABC";
```
vjagadish1989, line 3: This is a pretty good selection of test cases. Just to ensure complete parity with the Java spec, I wonder if it'll be a good idea to add additional test cases and asserts?

Set 1 : asserts that test encoded single byte, 2 byte and 3 bytes of zeros

```
public class TestNull {
  public static  String null1 = "A\0B"; <- access from scala code and verify that it should be of length 3
  public static  String null2 = "A\00B"; <- should be of length 3
  public static  String null3 = "A\000B"; <- should be of length 3
  public static   String null4 = "A\0000B"; <- should be of length 4
  public static  String null5 = "A\u0000B"; <- should be of length 3
}
```

It wonder if we could verify TestNull.null1 == TestNull.null2 == TestNull.null3 == TestNull.null5 == "A\0B" == "A\u0000B" from scala code? 

Please let me know if the additional asserts will be useful (or may be an overkill).

test/files/run/t9915/Test_2.scala
```diff
@@ -0,0 +1,14 @@
+
+object Test extends App {
+  val c = new C_1
+  assert(c.nulled == "X\u0000ABC")
+//assert(c.nulled == "X\000ABC")
```
vjagadish1989: I'm curious if this assert not be true? I verified in Java that:

`"A\u0000B".equals("A\0B") = true`?

test/files/run/t9915/C_1.java
som-snytt: Well, Scala doesn't see the source-level difference; Java encodes them the same. One funny thing about your test case was that octal escapes are deprecated in Scala.

test/files/run/t9915/Test_2.scala
som-snytt: Right, the octal escape is deprecated, but I left it as a comment, since part of the charm of the test was that the two strings were textually the same length.

SethTisue:
Som, sorry to nitpick, but I don't think we should have commented out code unless there's also a comment saying why the code is being kept. "charm of the test" notwithstanding. I'm suggesting just axing those lines

som-snytt:
You'd like me to comment that this is textually the same as the Java constant, but octal literals are deprecated in Scala so it uses unicode instead?

SethTisue:
I just don't think we should have unexplained commented-out code. I don't have a strong feeling on whether it's better to simply leave out the currently commented-out lines, or to add a comment saying why they're there.

som-snytt:
Squashed, rebased.

SethTisue:
LGTM assuming Jenkins thinks so too

som-snytt:
Now I wish I had tipped him better. How much do you normally give Jenkins at the holidays?

SethTisue:
bah humbug! reduce the surplus population!

som-snytt:
/rebuild

What is that color? Bronze? Dijon mustard? Since it's not green, rebuilding. Previously, finished:

```
[success] Total time: 4065 s, completed Oct 21, 2016 4:41:22 PM
erreicht: 6101597
Archiving artifacts
Finished: SUCCESS
```

SethTisue:
thanks Som!

PR #5387, opened 2016-09-08
Define the root sbt project's scalaVersion

dwijnand:
The root project is commonly used as the reference point for values, such
as "the scala version" of the (entire) project.

For example ENSIME uses it like that.

Also disablePublishing already disables publish and publishLocal.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,6 +91,8 @@ baseVersion in Global := "2.12.1"
 baseVersionSuffix in Global := "SNAPSHOT"
 mimaReferenceVersion in Global := Some("2.12.0-RC1")
 
+scalaVersion in Global := versionProps("starr.version")
+
 lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings ++ Seq[Setting[_]](
   organization := "org.scala-lang",
   // we don't cross build Scala itself
@@ -113,7 +115,6 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
       s2
     }
   },
-  scalaVersion := (scalaVersion in bootstrap).value,
   // As of sbt 0.13.12 (sbt/sbt#2634) sbt endeavours to align both scalaOrganization and scalaVersion
   // in the Scala artefacts, for example scala-library and scala-compiler.
   // This doesn't work in the scala/scala build because the version of scala-library and the scalaVersion of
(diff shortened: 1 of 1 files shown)


dwijnand:
See https://github.com/sbt/sbt/issues/2579 for more `scalaVersion` context.

szeiger:
My take on this is that ENSIME uses it wrong, but I see no harm in changing it if it improves IDE support. We already do some hacks of this kind for IntelliJ.

We should change it in a different way though: Change the minimum number of settings (e.g. only set `version in root`) and add a comment why this settings is needed. The more stuff we change randomly because some IDE needs it, the harder it becomes to refactor the build to a more standard sbt layout (now that ant is gone) without breaking IDE support again.

dwijnand:
Adding a "commonSettings" is a common sbt practice by the way.

soc (approved):

fommil:
making my first contribution to scala/scala I got stung by this. Seems a trivial fix, made a similar change locally, thanks Dale! Will reference this from the ensime docs. Please backport to 2.11.

fommil:
looking closer, my fix was different, I did this

``` diff
@@ -135,7 +135,6 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
       s2
     }
   },
-  scalaVersion := (scalaVersion in bootstrap).value,
   // we always assume that Java classes are standalone and do not have any dependency
   // on Scala classes
   compileOrder := CompileOrder.JavaThenScala,
@@ -338,10 +337,10 @@ def regexFileFilter(s: String): FileFilter = new FileFilter {
   def accept(f: File) = pat.matcher(f.getAbsolutePath.replace('\\', '/')).matches()
 }

+scalaVersion in ThisBuild := versionProps("starr.version")
+
 // This project provides the STARR scalaInstance for bootstrapping
-lazy val bootstrap = (project in file("target/bootstrap")).settings(
-  scalaVersion := versionProps("starr.version")
-)
+lazy val bootstrap = (project in file("target/bootstrap"))

 lazy val library = configureAsSubproject(project)
   .settings(generatePropertiesFileSettings: _*)

```

SethTisue:
@dwijnand needs rebase

dwijnand:
Switched to defining the `scalaVersion` in `Global`, once, for all the projects, instead of defining and evaluating it multiple times.

retronym:
LGTM

PR #5390, opened 2016-09-09
Emit all mixin forwarders when building Scala

lrytz:


lrytz:
This is just to trigger a CI build.

lrytz:
https://scala-ci.typesafe.com/job/scala-2.12.x-validate-main/3613/

retronym:
I cherry picked this commit on top of my fix to a performance regression in 2.12.0-RC1 and benchmarked with [scala-jmh-suite](http://github.com/retronym/scala-jmh-suite).

image

It seems that it offers a modest improvement to cold performance (~10%). We're still behind 2.11.8 cold performance. That is no doubt also due to other changes, like the lambda encoding.

Hot performance seems unchanged.

lrytz:
thank you for doing the benchmark!

PR #5391, opened 2016-09-10
SI-6276 Exclude var from trivial recursion warning

som-snytt:
`var x: X = x` is the new `var x: X = _`.

Unlike explicitly setting a default, it preserves any value
that was set earlier during initialization.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -1127,7 +1127,7 @@ abstract class RefChecks extends Transform {
     // SI-6276 warn for trivial recursion, such as `def foo = foo` or `val bar: X = bar`, which come up more frequently than you might think.
     // TODO: Move to abide rule. Also, this does not check that the def is final or not overridden, for example
     def checkInfiniteLoop(sym: Symbol, rhs: Tree): Unit =
-      if (!sym.isValueParameter && sym.paramss.isEmpty) {
+      if (!sym.isValueParameter && !sym.isVariable && sym.paramss.isEmpty) {
         rhs match {
           case t@(Ident(_) | Select(This(_), _)) if t hasSymbolWhich (_.accessedOrSelf == sym) =>
             reporter.warning(rhs.pos, s"${sym.fullLocationString} does nothing other than call itself recursively")
--- a/test/files/neg/t6276.scala
+++ b/test/files/neg/t6276.scala
@@ -41,4 +41,10 @@ object Test {
       }
     }
   }
+
+  class F {
+    g()
+    var f: String = f
+    def g() = this.f = "hi"
+  }
 }


SethTisue:
I'm not sure how I feel about this...

som-snytt:
This must have come up on gitter or SO. In fact, I suddenly recall you wanted to eliminate underscore syntax? Maybe that was the context.

SethTisue:
ah, the discussion about `_` vs `null.asInstanceOf[T]`. I would like to reduce the number of cases where null can sneak into your code even though you didn't textually say `null`. This PR seems like a step backwards on that.

as for the "preserves any value that was set earlier during initialization" part, that gives my hives the heebie jeebies, or is that the other way around? indeed

```
scala> object O { x = 3; var x: Int = x; println(x) }
<console>:11: warning: variable x in object O does nothing other than call itself recursively
       object O { x = 3; var x: Int = x; println(x) }
                                      ^
<console>:11: warning: Reference to uninitialized variable x
       object O { x = 3; var x: Int = x; println(x) }
                  ^
defined object O

scala> O
3
res0: O.type = O$@3b90a30a
```

but is there a point to reducing this from 2 warnings to 1? I think it'd rather it printed 10 warnings.

maybe you've spent longer thinking about than I have and you actually have a case, but if so, I'm asking you to state that case. what's an example of code that it's actually good not to issue this warning for?

som-snytt:
It can be useful to stash a value somewhere for debugging purposes, that is, to be read with a debugger.

With https://github.com/scala/scala/pull/5402
```
$ skala -Ywarn-unused
Welcome to Scala 2.12.2-20161212-232841-f048339 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_111).

scala> object X { private var x: Int = _ }
<console>:11: warning: private var x in object X is never used
       object X { private var x: Int = _ }
                              ^
defined object X

scala> object X { private var x: Int = x }
<console>:11: warning: private setter of x in object X is never used
       object X { private var x: Int = x }
                              ^
<console>:11: warning: variable x in object X does nothing other than call itself recursively
       object X { private var x: Int = x }
                                       ^
defined object X

scala> object X { private var x: Int = x ; x = 42 }
<console>:11: warning: variable x in object X does nothing other than call itself recursively
       object X { private var x: Int = x ; x = 42 }
                                       ^
defined object X
```
The warning about private setter should suggest using a val, so I broke that message. (Edit: actually it always worked like that, only local vars get the val prompt.)

SethTisue:
further thoughts or opinions on this, anyone?

som-snytt:
I'm not anyone, but re-reading it again, it's pretty funny.

`import language.scaryScala` might take care of it.

lrytz:
I also don't feel like `var x: X = x` is something we should promote.

adriaanm:
I'm not sensing a lot of enthusiasm for this one. If the main use case is for debugging purposes, wouldn't it be desirable to have a warning remind you to remind that debug line before you submit your PR? Happy to reopen if someone feels strongly about getting this in.

som-snytt:
Thanks, I'll reopen when attempts to squash the underscore initializer break down in tears. The syntax from the dotty implicit args issue might work, namely, `var x: X = implicit`. Aha! Or `implicitly`. I'll have to consult @ merriam-webster.

PR #5393, opened 2016-09-10
No warn when discarding r.f(): r.type

som-snytt:
The paradigm is `def add(x: X): Unit = listBuffer += x`.

The value that is discarded is not new information.

Also cleans up the recent tweaks to help messaging.

Adds newlines in case they ask for multiple helps.

This commit is also a test whether warnings and help
fall under documentation tweaks which don't require
a tracking issue. That doesn't mean the tweak can't
crash the compiler.


lrytz:
LGTM, thanks!

PR #5394, opened 2016-09-11
WIP: fix for regression in 2.12.0-RC1 compiling shapeless tests.

milessabin:
This change allows the shapeless tests to be compiled with 2.12.0-RC1+. I don't claim that this fix is correct, but it should help see what the issue is.

It appears that the new fields phase is [eliminating constant typed vals](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/transform/Fields.scala#L695-L698) but not ensuring that all references to them are replaced with the corresponding literal values. The included test doesn't completely cover the failure in the shapeless tests, but it's close ... if this is along the right lines I should be able to come up with a shapeless-independent test which covers the rest, but I think it will require a macro to do it.

Nb. in this instance the problematic val is the synthetic one introduced by the parser when translating right associative operators. Compiling the test case without this fix and with `-Xprint:all` shows the constant typed synthetic val definition being eliminated in fields, but leaving an untransformed reference to it. This leaks right the way through to code generation and blows up then. The fix here eliminates the dangling reference in erasure.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Erasure.scala
+++ b/src/compiler/scala/tools/nsc/transform/Erasure.scala
@@ -1179,7 +1179,9 @@ abstract class Erasure extends InfoTransform
           copyDefDef(tree)(tparams = Nil)
         case TypeDef(_, _, _, _) =>
           EmptyTree
-
+        case Ident(_) if isSingleConstantType(tree.tpe) =>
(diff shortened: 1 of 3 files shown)


retronym:
Thanks, @milessabin. I'm taking a look at this one now to see if we can rewrite the reference in the fields phase rather than leaving that to erasure.

retronym:
My fix: #5395

PR #5396, opened 2016-09-13
SI-9920 Avoid linkage errors with captured local objects + self types

retronym:
An outer parameter of a nested class is typed with the self type
of the enclosing class:

```
class C; trait T { _: C => def x = 42; class D { x } }
```

leads to:

```
    class D extends Object {
      def <init>($outer: C): T.this.D = {
        D.super.<init>();
        ()
      };
      D.this.$outer().$asInstanceOf[T]().x();
```

Note that a cast is inserted before the call to `x`.

If we modify that a little, to instead capture a local module:

```
class C; trait T { _: C => def y { object O; class D { O } } }
```

Scala 2.11 used to generate (after lambdalift):

```
    class D$1 extends Object {
      def <init>($outer: C, O$module$1: runtime.VolatileObjectRef): C#D$1 = {
        D$1.super.<init>();
        ()
      };
      D$1.this.$outer().O$1(O$module$1);
```

That isn't type correct, `D$1.this.$outer() : C` does not
have a member `O$1`.

However, the old trait encoding would rewrite this in mixin to:

```
T$class.O$1($outer, O$module$1);
```

Trait implementation methods also used to accept the self type:

```
trait T$class {
   final <stable> def O$1($this: C, O$module$1: runtime.VolatileObjectRef): T$O$2.type
}
```

So the problem was hidden.

This commit changes `outerPath` (used in both explicitouter and in lambdalift to
insert a cast if required.)

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/ExplicitOuter.scala
+++ b/src/compiler/scala/tools/nsc/transform/ExplicitOuter.scala
@@ -8,7 +8,7 @@ package tools.nsc
 package transform
 
 import symtab._
-import Flags.{ CASE => _, _ }
+import Flags.{CASE => _, _}
 import scala.collection.mutable.ListBuffer
 
 /** This class ...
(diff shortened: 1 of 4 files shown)


retronym:
Review by @lrytz

PR #5398, opened 2016-09-14
SD-225 Use a "lzycompute" method for module initialization

retronym:
The monitors and module instantation were inliuned into the module accessor
method in b2e0911. However, this seems to have had a detrimental impact on
performance. This might be because the module accessors are now above the
"always inline" HotSpot threshold of 35 bytes, or perhaps because they
contain monitor-entry/exit and exception handlers.

This commit returns to the the 2.11.8 appraoch of factoring
the the second check of the doublecheck locking into a method.
I've done this by declaring a nested method within the accessor;
this will be lifted out to the class level by lambdalift.

This represents a slight deviation from the implementation strategy used
for lazy accessors, which create a symbol for the slowpath method in
the info transform and generate the corresponding DefDef as a class
member. I don't believe this deviation is particular worrisome, though.

I have bootstrapped the compiler through this commit and found that
the drastic regression in compiling the shapeless test suite is solved.

Fixes https://github.com/scala/scala-dev/issues/225

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Fields.scala
+++ b/src/compiler/scala/tools/nsc/transform/Fields.scala
@@ -22,7 +22,7 @@ import symtab.Flags._
   * in the first (closest in the subclassing lattice) subclass (not a trait) of a trait.
   *
   * For lazy vals and modules, we emit accessors that using double-checked locking (DCL) to balance thread safety
-  * and performance. A lazy val gets a compute method for the DCL's slow path, for a module it's all done in the accessor.
+  * and performance. For both lazy vals and modules, the a compute method contains the DCL's slow path.
   *
   * Local lazy vals do not receive bitmaps, but use a Lazy*Holder that has the volatile init bit and the computed value.
   * See `mkLazyLocalDef`.
@@ -236,7 +236,9 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
       *
       * TODO: optimize using local variable?
       */
-    Block(If(needsInit, gen.mkSynchronized(monitorHolder)(If(needsInit, init, EmptyTree)), EmptyTree) :: Nil, moduleVarRef)
+    val computeName = nme.newLazyValSlowComputeName(module.name)
+    val computeMethod = DefDef(NoMods, computeName, Nil, ListOfNil, TypeTree(UnitTpe), gen.mkSynchronized(monitorHolder)(If(needsInit, init, EmptyTree)))
+    Block(computeMethod :: If(needsInit, Apply(Ident(computeName), Nil), EmptyTree) :: Nil, moduleVarRef)
   }
 
   // NoSymbol for lazy accessor sym with unit result type
--- a/test/files/run/delambdafy_t6028.check
+++ b/test/files/run/delambdafy_t6028.check
@@ -42,10 +42,11 @@ package <empty> {
       <synthetic> <stable> <artifact> def $outer(): T = MethodLocalObject$2.this.$outer;
       <synthetic> <stable> <artifact> def $outer(): T = MethodLocalObject$2.this.$outer
     };
+    final private[this] def MethodLocalObject$lzycompute$1(barParam$1: String, MethodLocalObject$module$1: runtime.VolatileObjectRef): Unit = T.this.synchronized[Unit](if (MethodLocalObject$module$1.elem.$asInstanceOf[T#MethodLocalObject$2.type]().eq(null))
+      MethodLocalObject$module$1.elem = new T#MethodLocalObject$2.type(T.this, barParam$1));
     final <stable> private[this] def MethodLocalObject$1(barParam$1: String, MethodLocalObject$module$1: runtime.VolatileObjectRef): T#MethodLocalObject$2.type = {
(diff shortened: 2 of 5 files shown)


retronym:
Review by @adriaanm | @lrytz

adriaanm:
Looks like it needs a junit test update. Otherwise LGTM

lrytz:
LGTM, too

retronym:
I've run a few more benchmarks that show the same regression in 2.12.0-RC1 and restored performance after bootstrapping through this change (highlighted below).

image

milessabin:
:clap:

milessabin:
After this I'm seeing,

2.12.0 branch
coreJVM/compile         39s
coreJVM/test:compile    238s

Which is a significant improvement over RC1 but still almost twice as long as 2.11.8.

retronym:
@milessabin Did you use a bootstrapped compiler when collecting those numbers? Until https://github.com/scala/scala/pull/5403 is merged, we're still using 2.12.0-RC1 as STARR.

milessabin:
I did a publishLocal using @adriaanm's recipe, then rebuilt with `-Dstarr.version=...`, then `dist/mkPack` and built shapeless and its tests using `++2.12.0-RC1=... path to pack ...`.

milessabin:
If you have an alternative recipe for testing please let me know and I'll try it out.

retronym:
I've just tested against 2.12.0-d274405-SNAPSHOT (the artifact from #5403).

I benchmarked the cold performance of compiling shapeless core with this version and 2.11.8.

Results: https://gist.github.com/retronym/220345f697a18a0facc259556a8a73b2

```
[info] Benchmark                    (_scalaVersion)  (extraArgs)                                      (source)  Mode  Cnt      Score     Error  Units
[info] ColdScalacBenchmark.compile           2.11.8               @/Users/jz/code/shapeless/args-2.11-core.txt    ss   12  19463.820 ± 453.189  ms/op
[info] ColdScalacBenchmark.compile  2.12.0-d274405-SNAPSHOT               @/Users/jz/code/shapeless/args-2.12-core.txt    ss   12  22751.384 ± 727.376  ms/op
```

This does show a performance regression of around ~15%, which I'm interested to investigate further.

For the tests project (only one iteration this time)

```
[info] Benchmark                    (_scalaVersion)  (extraArgs)                                      (source)  Mode  Cnt       Score   Error  Units
[info] ColdScalacBenchmark.compile           2.11.8               @/Users/jz/code/shapeless/args-2.11-test.txt    ss       113177.188          ms/op
[info] ColdScalacBenchmark.compile  2.12.0-d274405-SNAPSHOT               @/Users/jz/code/shapeless/args-2.12-test.txt    ss       131956.216          ms/op
```

That suggests a similar slowdown. But neither of these matches with your report of a 2x slowdown.

milessabin:
My latest timings ...

For 2.12.0 + #5403 + dist/mkPack, latest from shapeless master,

```
% sbt -Dsbt.profile=2.12.x
> ++2.12.0-RC1=/home/miles/projects/scala/build/pack
> clean
> coreJVM/compile
> coreJVM/test:compile

coreJVM/compile 37s
coreJVM/test:compile 200s
```

For 2.11.8, latest from shapeless master,

```
% sbt -Dsbt.profile=2.12.x
> ++2.11.8
> clean
> coreJVM/compile
> coreJVM/test:compile

coreJVM/compile 33s
coreJVM/test:compile 123s
```

So, I'm seeing about the same slowdown as you for coreJVM/compile but a much bigger slowdown for coreJVM/test:compile ... consistently 1.6x with my latest 2.12.0 build.

Are we compiling the exact same tests? Is it possible that you excluded some to avoid the constant val elimination problem?

Alternatively, am I doing something wrong?

milessabin:
(nb. the `-Dsbt.profile=2.12.x` just disables `sbt-scoverage` which doesn't have 2.12.x support yet).

retronym:
@milessabin Just doing `dist/mkPack` will not use `-opt` when building. I suggest using `;setupPublishCore;dist/mkPack;publishLocal`

Here are the compiler command lines I used (includind selection of test source files) https://gist.github.com/retronym/8bca64252f68fbfb354593af7b53ae74

I was also compiling outside of SBT. It is conceivable that there is an additional regression in the incremental compiler (which adds two compiler phases to extract dependencies and APIs from the compiled files).

milessabin:
I'm afraid I can't reproduce your results, even building with scalac outside SBT.

I'm quite prepared to believe that -optimise (or rather the lack of it) is the issue here, but if it is then none of the instructions above seem to be enough to coax it into being applied.

milessabin:
Could you publish a snapshot that I can test here?

milessabin:
Compiling with plain scalac, command line similar to yours above I see,

```
2.11.8  129s
2.12.x  155s
```

So 1.2x in this case ... does that line up with what you're seeing?

In SBT I'm seeing,

```
2.11.8  138s
2.12.x  237s
```

ie. 1.7x, which suggests that your hypothesis that there's a regression in the incremental compiler might be right.

retronym:
I have since found what might have been the root cause of this problem: our code gen from `synchronized` has subtly changed in a way that rendered them unJITtable.

I have proposed a fix for that in https://github.com/scala/scala/pull/5417

I think this change makes sense in any case, but if anyone want to revisit the decision, we could do some benchmarking without the `lzycompute` methods for module accessors.

PR #5399, opened 2016-09-14
Only compare group ID and artifact ID when looking for JLine JAR

szeiger:
If the JLine dependency is resolved through Ivy instead of Maven (which
can happen with a local proxy setup in sbt that funnels all artifacts
through an Ivy repo), the artifact may have a “configuration” set,
which prevents a naive comparison of ModuleIDs from finding it.

@mgttlinger ran into this issue during the Typelevel Workshop  at Scala World.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -463,7 +463,7 @@ lazy val replJlineEmbedded = Project("repl-jline-embedded", file(".") / "target"
       import collection.JavaConverters._
       val inputs: Iterator[JarJar.Entry] = {
         val repljlineClasses = (products in Compile in replJline).value.flatMap(base => Path.allSubpaths(base).map(x => (base, x._1)))
-        val jlineJAR = (dependencyClasspath in Compile).value.find(_.get(moduleID.key) == Some(jlineDep)).get.data
+        val jlineJAR = findJar((dependencyClasspath in Compile).value, jlineDep).get.data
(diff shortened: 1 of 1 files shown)


mgttlinger:
For reference: I ran into this because I use coursier (http://get-coursier.io/) for dependency resolution instead of the built in resolver in sbt.

build.sbt
```diff
@@ -1126,3 +1126,9 @@ intellijToSample := {
   } else
     s.log.info("Aborting.")
 }
+
+/** Find a specific module's JAR in a classpath, comparing only organization and name */
+def findJar(files: Seq[Attributed[File]], dep: ModuleID): Option[Attributed[File]] = {
+  def extract(m: ModuleID) = (m.organization, m.name)
+  files.find(_.get(moduleID.key).map(extract _) == Some(extract(dep)))
```
lrytz, line 1133: `Some(extract(dep))` could be stored in a local instead of re-creating it over again

lrytz (approved):

lrytz:
LGTM

dwijnand:
Comparing ModuleId's with == isn't very reliable.

I improved the situation a little for sbt 1.0 https://github.com/sbt/librarymanagement/pull/48/files

It was working before this change before because JLine is a java dependency, so is defined with `CrossVerison.Disabled`, which is an object - no function to mess things up.

PR #5400, opened 2016-09-14
Rewrite TraversableLike.stringPrefix not to blow up code size in Scala.js.

sjrd:
The commit 30876fe2dd8cbe657a6cad6b11bbc34f10c29b36 changed `TraversableLike.stringPrefix` to report nicer results for inner classes and method-local classes. The changes included calls to `String.split()`, `Character.isDigit()` and `Character.isUpperCase()`. This was particularly bad for Scala.js, because those methods bring with them huge parts of the JDK (the `java.util.regex.*` implementation on the one hand, and the Unicode database on the other hand), which increased generated code size by 6 KB after minimification and gzip for an application that does not otherwise use those methods. This sudden increase is tracked in the Scala.js bug tracker at https://github.com/scala-js/scala-js/issues/2591.

This commit rewrites `TraversableLike.stringPrefix` in a very imperative way, without resorting to those methods. The behavior is (mostly) preserved. There can be different results when `getClass().getName()` contains non-ASCII lowercase letters and/or digits. Those will now be recognized as user-defined instead of likely compiler-synthesized (which is a progression). There still are false positives for ASCII lowercase letters, which cause the `stringPrefix` to be empty (as before).

Since the new implementation is imperative anyway, at least I made it not allocate anything but the result `String` in the common case where the result does not contain any `.`.

### DIFF
--- a/src/library/scala/collection/TraversableLike.scala
+++ b/src/library/scala/collection/TraversableLike.scala
@@ -605,22 +605,69 @@ trait TraversableLike[+A, +Repr] extends Any
    *           applied to this $coll. By default the string prefix is the
    *           simple name of the collection class $coll.
    */
-  def stringPrefix : String = {
+  def stringPrefix: String = {
+    /* This method is written in a style that avoids calling `String.split()`
+     * as well as methods of java.lang.Character that require the Unicode
+     * database information. This is mostly important for Scala.js, so that
+     * using the collection library does automatically bring java.util.regex.*
+     * and the Unicode database in the generated code.
+     *
+     * This algorithm has the additional benefit that it won't allocate
+     * anything except the result String in the common case, where the class
+     * is not an inner class (i.e., when the result contains no '.').
+     */
     val fqn = repr.getClass.getName
-    val cls = {
-      val idx1 = fqn.lastIndexOf('.' : Int)
-      if (idx1 != -1) fqn.substring(idx1 + 1) else fqn
+    var pos: Int = fqn.length - 1
+
+    // Skip trailing $'s
+    while (pos != -1 && fqn.charAt(pos) == '$') {
+      pos -= 1
     }
-    val parts = cls.split('$')
-    val last = parts.length - 1
-    parts.zipWithIndex.foldLeft("") { case (z, (s, i)) =>
-      if (s.isEmpty) z
-      else if (i != last && s.forall(java.lang.Character.isDigit)) "" // drop prefix in method-local classes
-      else if (i == 0 || java.lang.Character.isUpperCase(s.charAt(0))) {
-        if (z.isEmpty) s else z + '.' + s
+    if (pos == -1 || fqn.charAt(pos) == '.') {
+      return ""
+    }
+
+    var result: String = ""
+    while (true) {
+      // Invariant: if we enter the loop, there is a non-empty part
+
+      // Look for the beginning of the part, remembering where was the last non-digit
+      val partEnd = pos + 1
+      while (pos != -1 && fqn.charAt(pos) <= '9' && fqn.charAt(pos) >= '0') {
+        pos -= 1
       }
-      else z
+      val lastNonDigit = pos
+      while (pos != -1 && fqn.charAt(pos) != '$' && fqn.charAt(pos) != '.') {
+        pos -= 1
+      }
+      val partStart = pos + 1
+
+      // A non-last part which contains only digits marks a method-local part -> drop the prefix
+      if (pos == lastNonDigit && partEnd != fqn.length) {
+        return result
+      }
+
+      // Skip to the next part, and determine whether we are the end
+      while (pos != -1 && fqn.charAt(pos) == '$') {
+        pos -= 1
+      }
+      val atEnd = pos == -1 || fqn.charAt(pos) == '.'
+
+      // Handle the actual content of the part (we ignore parts that are likely synthetic)
+      def isPartLikelySynthetic = {
+        val firstChar = fqn.charAt(partStart)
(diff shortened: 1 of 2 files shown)


sjrd:
Review by @Ichoran.

I'm happy to discuss alternative ways to tackle this issue.

som-snytt:
`StringOps.split(Char)` should just inline its simple implementation instead of calling out to Java. It would mean I can do `name.split(c)` in JS without breaking out in a cold sweat.

(It had been implemented once before, to correct a problem with escaping, then unimplemented with a different fix for escaping that dances with Java.)

src/library/scala/collection/TraversableLike.scala
```diff
@@ -605,22 +605,65 @@ trait TraversableLike[+A, +Repr] extends Any
...
-    parts.zipWithIndex.foldLeft("") { case (z, (s, i)) =>
-      if (s.isEmpty) z
-      else if (i != last && s.forall(java.lang.Character.isDigit)) "" // drop prefix in method-local classes
-      else if (i == 0 || java.lang.Character.isUpperCase(s.charAt(0))) {
-        if (z.isEmpty) s else z + '.' + s
+
+    var pos: Int = fqn.length - 1
+
+    // Skip trailing $'s
+    while (pos != -1 && fqn.charAt(pos) == '$')
```
Ichoran: It's hard to read all of these short statements split onto separate lines but without braces.  One gets lost in all the whitespace.

src/library/scala/collection/TraversableLike.scala
```diff
@@ -605,22 +605,65 @@ trait TraversableLike[+A, +Repr] extends Any
...
+
+      val partStart = pos + 1
+
+      // Skip to the next part
+      while (pos != -1 && fqn.charAt(pos) == '$')
+        pos -= 1
+
+      // Handle the actual content of the part
+      val atEnd = pos == -1 || fqn.charAt(pos) == '.'
+      if (atEnd || (fqn.charAt(partStart) <= 'Z' && fqn.charAt(partStart) >= 'A')) {
```
Ichoran: Isn't this algorithm wrong?  What if someone uses a lower-case class name?  (The algorithm hasn't changed, but is it correct?)

Ichoran:
It looks reasonable but can you add some tests, including those that illustrate the difference in behavior e.g. with some of `ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑ`...?

src/library/scala/collection/TraversableLike.scala
sjrd: I didn't stop to question the original algorithm, TBH ^^ I found that a bit weird, but I thought there was a good reason for that. You were the one to write the previous version of the algorithm. Do you remember why it was only checking for uppercase? Was it some clever way to filter out `$$lambda` something?

sjrd:
@Ichoran OK I'll add some tests.

@som-snytt Implementing `split(Char)` ourselves would make sense as well. I first started with that, but it got rid of only half of the code size increase. So I figured I'd do small steps and only touch `stringPrefix`.

src/library/scala/collection/TraversableLike.scala
sjrd: OK I guess we have different styles ;-) I'll add braces and remove some of the blank lines.

src/library/scala/collection/TraversableLike.scala
Ichoran: Yeah, looking back, that was it: a heuristic to filter out whatever the compiler stuck in.  But my version doesn't use foldLeft or split and is at https://github.com/Ichoran/scala/blob/issue/9019/src/library/scala/collection/TraversableLike.scala#L661 (and I don't know if it was correct or not, since Stefan (apparently?--not fully trusting git blame here) modified it before committing it).

Ichoran:
Crikey, GitHub seems to have added some sort of awful make-your-comments-and-then-say-you-want-to-make-them "feature".  My apologies if any of my comments go missing while I'm getting used to this and/or figuring out how to turn it off.

src/library/scala/collection/TraversableLike.scala
Ichoran: Which suggests that if it's to avoid compiler symbols and the compiler only includes ASCII, the test could be narrower (e.g. allow anything over 0x7E).

sjrd:
I have rearranged the method for less whitespace and more braces, and added a few more test cases.

test/junit/scala/collection/TraversableLikeTest.scala
```diff
@@ -5,29 +5,65 @@ import org.junit.Test
...
+
     val bar = Foo.mkBar()
     assertEquals("Bar", bar.stringPrefix)  // Previously would have been outermost class, TraversableLikeTest
 
     val baz = new Baz[Int]()
     assertEquals("TraversableLikeTest.Baz", baz.stringPrefix)  // Make sure we don't see specialization $mcI$sp stuff
+
+    // The false positive unfortunately produces an empty stringPrefix
+    val falsePositive = Foo.mkFalsePositiveToSyntheticTest()
+    assertEquals("", falsePositive.stringPrefix)
```
sjrd, line 61: Note that this happened with the previous algorithm as well.

szeiger:
Should this be targeting 2.12.0 instead?

sjrd:
I can target 2.12.0 if you guys prefer. I thought 2.12.x was less demanding for you on the RC cycle. Since it's not critical and is binary compatible, I don't have a problem waiting for 2.12.1 to see it fixed. I'm happy to change if you want.

SethTisue:
thx @sjrd and reviewers!

PR #5402, opened 2016-09-14
SI-8040 Improve unused warnings

som-snytt:
Add symbol names, don't warn for both getters and setters
or for synthetics (except default arg getters).

Tweak messages for readability.

Warn for unused `case X(x) =>` but, as an escape hatch,
not for `case X(x @ _) =>`. The latter form is deemed
documentary. (Named args could serve a similar purpose,
`case X(x = _) =>`.)


som-snytt:
Exit code 137.

```
Exception in thread "Thread-29" java.io.EOFException
    at java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:2626)
    at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1321)
    at java.io.ObjectInputStream.readObject(ObjectInputStream.java:373)
    at sbt.React.react(ForkTests.scala:114)
    at sbt.ForkTests$$anonfun$mainTestTask$1$Acceptor$2$.run(ForkTests.scala:74)
    at java.lang.Thread.run(Thread.java:745)
[error] Error during tests:
[error]     Running java with options -classpath /home/jenkins/workspace/scala-2.12.x-validate-test@2/target/test/it-classes:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/test:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/compiler:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/library:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/reflect:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/interactive:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/repl-jline-embedded:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/repl-jline:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/repl:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/scalap:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/partest-extras:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/pack/lib/scala-partest-javaagent.jar:/home/jenkins/workspace/scala-2.12.x-validate-test@2/build/quick/classes/scaladoc:/home/jenkins/.ivy2/cache/org.apache.ant/ant/jars/ant-1.9.4.jar:/home/jenkins/.ivy2/cache/org.apache.ant/ant-launcher/jars/ant-launcher-1.9.4.jar:/home/jenkins/.ivy2/cache/org.scala-lang.modules/scala-asm/bundles/scala-asm-5.1.0-scala-1.jar:/home/jenkins/.ivy2/cache/org.scala-lang.modules/scala-xml_2.12.0-RC1/bundles/scala-xml_2.12.0-RC1-1.0.5.jar:/home/jenkins/.ivy2/cache/jline/jline/jars/jline-2.14.1.jar:/home/jenkins/.ivy2/cache/org.scala-lang.modules/scala-partest_2.12.0-RC1/jars/scala-partest_2.12.0-RC1-1.0.17.jar:/home/jenkins/.ivy2/cache/com.googlecode.java-diff-utils/diffutils/jars/diffutils-1.3.0.jar:/home/jenkins/.ivy2/cache/org.scala-sbt/test-interface/jars/test-interface-1.0.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/jsoup-1.3.1/jars/jsoup-1.3.1-346d3dff4088839d6b4d163efa2892124039d216.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/macro210/jars/macro210-3794ec22d9b27f2b179bd34e9b46db771b934ec3.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/enums/jars/enums-981392dbd1f727b152cd1c908c5fce60ad9d07f7.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/annotations/jars/annotations-02fe2ed93766323a13f22c7a7e2ecdcd84259b6c.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/genericNest/jars/genericNest-b1ec8a095cec4902b3609d74d274c04365c59c04.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/methvsfield/jars/methvsfield-be8454d5e7751b063ade201c225dcedefd252775.jar:/home/jenkins/.ivy2/cache/org.scala-lang.scala-sha-bootstrap.test.files.lib/nest/jars/nest-cd33e0a0ea249eb42363a2f8ba531186345ff68c.jar:/home/jenkins/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-agent-0.13.11.jar:/home/jenkins/.sbt/boot/scala-2.10.6/org.scala-sbt/sbt/0.13.11/test-interface-1.0.jar sbt.ForkMain 33528 failed with exit code 137
```

test/files/neg/warn-unused-privates.check
```diff
@@ -61,6 +61,24 @@ warn-unused-privates.scala:137: warning: private method x in class OtherNames is
...
+                     ^
+warn-unused-privates.scala:161: warning: local val z in method h is never used
+    val C(x @ _, y @ _, z @ Some(_)) = c  // warn?
+                        ^
+warn-unused-privates.scala:166: warning: local val x in method v is never used
+    val D(x) = d                          // warn
+         ^
+warn-unused-privates.scala:170: warning: local val x in method w is never used
+    val D(x @ _) = d                      // fixme
+         ^
```
kiritsuku: I'm not sure if this should be enabled with the `-Ywarn-unused` flag alone. I expect many occurrences of this in real world code and people often have fatal warnings enabled but may not have the time to immediately fix all of the occurrences. I would prefer an additional flag to explicitly enable this new (and very useful) warning.

test/files/neg/warn-unused-privates.scala
```diff
@@ -52,13 +52,31 @@ trait Accessors {
   }
 }
 
+class StableAccessors {
+  private var s1: Int = 0 // warn
+  private var s2: Int = 0 // warn, never set
+  private var s3: Int = 0 // warn, never got
+  private var s4: Int = 0 // no warn
```
kiritsuku, line 59: How is an identifier like

``` scala
  var `s5 with spaces and backticks` = 0
```

rendered?

src/compiler/scala/reflect/reify/codegen/GenAnnotationInfos.scala
```diff
@@ -10,7 +10,8 @@ trait GenAnnotationInfos {
   // however, when reifying free and tough types, we're forced to reify annotation infos as is
   // why is that bad? take a look inside
   def reifyAnnotationInfo(ann: AnnotationInfo): Tree = {
-    val reifiedArgs = ann.args map { arg =>
+    //val reifiedArgs = ann.args map
```
kiritsuku: is it normal here to keep old code as comments?

src/compiler/scala/tools/reflect/FormatInterpolator.scala
```diff
@@ -19,7 +19,7 @@ abstract class FormatInterpolator {
   @inline private def truly(body: => Unit): Boolean = { body ; true }
   @inline private def falsely(body: => Unit): Boolean = { body ; false }
 
-  private def fail(msg: String) = c.abort(c.enclosingPosition, msg)
+  //private def fail(msg: String) = c.abort(c.enclosingPosition, msg)
```
kiritsuku: useless comment, see above

test/files/neg/warn-unused-privates.check
som-snytt: The position of x is different here, hence "fixme". It's not feature complete yet. I suspect it will eventually require `-Ywarn-unused:patterns,parameters,vars` or whatever to suit individual needs.

test/files/neg/warn-unused-privates.scala
som-snytt: I noticed `Symbol.nameString`, that may be a way to consolidate backtickage.

src/compiler/scala/reflect/reify/codegen/GenAnnotationInfos.scala
som-snytt: I'll squash the "clean ups". Here, the comment should add "SI-8915".

src/compiler/scala/tools/reflect/FormatInterpolator.scala
som-snytt: That's my code, so I kept it.

test/files/neg/warn-unused-privates.check
som-snytt: I pushed the upgraded `-Y`.

som-snytt:
Rebased with @sschaef advice not to leave extraneous commented code.

Also last commit undertakes handling his https://issues.scala-lang.org/browse/SI-7860

src/compiler/scala/tools/nsc/settings/Warnings.scala
```diff
@@ -99,7 +127,9 @@ trait Warnings {
     helpArg = "warning",
     descr   = "Enable or disable specific warnings",
     domain  = LintWarnings,
-    default = Some(List("_")))
+    default = Some(List("_"))) //.withPostSetHook (s => if (s contains Unused) warnUnused.add("_"))
```
lrytz: should this be enabled now?

src/compiler/scala/tools/nsc/settings/Warnings.scala
```diff
@@ -15,15 +15,40 @@ trait Warnings {
...
+    val Implicits = Choice("implicits", "Warn if an implicit parameter is unused.")
+  }
+
+  // The -Ywarn-unused warning group.
+  val warnUnused = MultiChoiceSetting(
+    name    = "-Ywarn-unused",
+    helpArg = "warning",
+    descr   = "Enable or disable specific `unused' warnings",
+    domain  = UnusedWarnings,
+    default = Some(List("_"))
```
lrytz, line 39: what do you think of using `default = List("-patvars")`? just wanted to bring it up.

lrytz (commented):
Really impressive work, thank you!

src/compiler/scala/tools/nsc/settings/Warnings.scala
som-snytt: I have one more experimental commit that enables it for scalac, with fixes, to see if turning everything on, including patvars, is doable. You may be right that just leaving it off by default is least controversial or annoying.

src/compiler/scala/tools/nsc/settings/Warnings.scala
som-snytt: Oh right. Might as well shoot for the stars!

som-snytt:
@lrytz Thanks for the pat on the back. It really goes a long way. Also thanks to the whole team for their time in reviewing PRs.

som-snytt:
The way it does constrParamAccessors is false pos on super.init(p).

Also defaultCase$ and the copy/paste in $anonfun.isDefinedAt which always has true/false RHS so nothing used.

The "documentary pattern bind" `X(p@_)` is actually helpful to see what's used and what's just showing what the placeholder represents.

som-snytt:
I suspect scalacOptions from build.sbt are communicated to partest.

test/files/neg/warn-unused-privates.check
```diff
@@ -106,6 +106,15 @@ warn-unused-privates.scala:201: warning: pattern var z in method f is never used
...
+warn-unused-privates.scala:136: warning: parameter value i in method x_= is never used
+  private def x_=(i: Int): Unit = ???
+                  ^
+warn-unused-privates.scala:138: warning: parameter value i in method y_= is never used
+  private def y_=(i: Int): Unit = ???
+                  ^
 error: No warnings can be incurred under -Xfatal-warnings.
-36 warnings found
+39 warnings found
```
lrytz: this change fails in CI, and i don't see what parts of this commit should cause these additional warnings.

som-snytt: I was running the build.sbt with `-Ywarn-unused` and didn't realize it got passed additively to the test. Also, I think I have improvements to follow up? [edit: yes, false positives from pattern matching anonfuns, as noted a few days ago] or was I only doing a "heed advice" clean-up? I'l fix the green first.

lrytz: just my two review comments / questions above (default for `-Ywarn-unused`, post-set-hook for `-Xlint`)

som-snytt:
The funky idiom is `-Ywarn-unused -Xlint:-unused`, otherwise lint turns off `patvars` and Ywarn is not allowed to turn it back on again. So maybe amend not to special case patvars.

som-snytt:
Rebased and added settings tweaks, tweaks to anonfun handling.  There's a big "heed the warnings" delta in waiting somewhere. I'll submit that as a separate PR. After I get green checks again here...

som-snytt:
Together with the green snytt icons, the red checks are kind of cheery in a seasonal way.

adriaanm:
Now if only we could make them blink.

adriaanm:
There was no room left in the worker's stocking. I've stretched it a bit.

adriaanm:
/rebuild  (one at a time for now)

som-snytt:
I hope you're flattered that we take you for granted.

som-snytt:
Green checks! It's all I want for xmas.

som-snytt:
If you ran out and got me green checks patterned socks for Christmas, or boxer shorts, or in particular boxer briefs, please understand that I'll receive them gratefully. Who doesn't need socks?

som-snytt:
Riddle me this: What might you pass that is useful yet unused? DummyImplicit. (And other evidence.)

adriaanm:
Also, wind.

som-snytt:
I think it is time to reimagine scalapuzzlers.

retronym:
I'm somewhat skeptical that unused parameter checks will have a high enough signal-to-noise ratio to be that useful in practice. If it doesn't cost us too much code to include it, and it is for opt-inners only, I'm okay with including it though.

som-snytt:
By now I'll have to review what I attempted, but the search for a sweet spot included don't warn if implementing an API.

I think I didn't add yet: don't warn for abstract API, that is, for methods in a trait or abstract class. That's a way of signalling that there will be subclasses with overrides which use the parameter.

The other possible distinction was to warn for unused implicit params only.

The other noisy warning is `patvars`, where `case x @ _` disables the warning. I hope the settings ergonomics will make it easy to do what you prefer. So maybe `patvars` and `params` should be disabled under `-Xlint`, but with `-Ywarn-unused` you have to turn them off explicitly.

Also check that `for (s: String <- Option("")) yield s` doesn't warn in the filter.

som-snytt:
@adriaanm @lrytz @retronym Please give me a heads up on how to push this over the line for 2.12.2.

The ergonomical bits are not to overwarn on unused parameters and unused `case p(x)`. Is "ease of disabling" sufficient, or should it default not to warn? I tend to think default to warn all is easier to reason about. (I turned on `javac -Xlint:all` in my colleague's gradle build and it's confusing how to exclude a heap pollution warning, for example. So I sympathize with the easily annoyed. Or sympathise.)

I wouldn't want to have to merge this code, as it is old and I am old and the earth is no longer young either.

adriaanm:
Thanks for your patience here. Let's get this merged before your beard is the length of Jon Stewart's tie. @SethTisue, could you look at this with a fresh pair of 👀 ?

lrytz:
To summarize

  - `-Xlint:unused` enables `-Ywarn-unused:-patvars,_`
    - `params` is enabled - is that too verbose?
  - `-Ywarn-unused` is now a multi-choicesetting
  - `-Ywarn-unused` enables all unused warnings, including `patvars` and `params`
  - Unused params exceptions:
    - deprecated methods
    - overriding methods
    - `args` in `main`
    - `DummyImplicit` (other unused implicit params: warn)
    - `val` class parameters
    - https://github.com/som-snytt/scala/blob/issue/8040-unused/test/files/neg/warn-unused-params.scala
    - https://github.com/som-snytt/scala/blob/issue/8040-unused/test/files/pos/t8040.scala
  - Unused privates exceptions (not new in this PR): "readResolve" and friends, constants (possibly inlined, no longer referenced)
  - Unused patvars exceptions:
    - `x @ _` bindings
    - bindings originating in for-comprehensions (right?)
    - https://github.com/som-snytt/scala/blob/issue/8040-unused/test/files/neg/warn-unused-privates.scala#L147
    - with `-Ywarn-unused:-patvars`: https://github.com/som-snytt/scala/blob/issue/8040-unused/test/files/neg/warn-unused-patvars.scala

There are also some `fixme` comments in some tests, these can probably be addressed later.

som-snytt:
I'll try this out on the communal build. Thanks for the exhaustive and exhausting summarisation, @lrytz , it makes me want to sleep right now.  Edit: I'll try it again tomorrow.

som-snytt:
Rebased.

SethTisue (approved):

som-snytt:
The other day, I did actually kick off a community build on my home machine, but I'm busy at work, depressed about life, etc, and didn't wade through output. Probably I should find projects with `-Xfatal-warnings` and focus on those.

SethTisue:
let's do a community build run on Jenkins against 2.12.2-bin-47754ef-SNAPSHOT: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1387/consoleFull (404 while Jenkins untangles that mass of old USB cables in the closet)

SethTisue:
discussed at team meeting. we'll push to get this into 2.12.2 if reasonably possible

SethTisue:
so the community build run has:

```
Result: FAILED (failed: scala-partest, tut, pcplod, macro-paradise, kind-projector)
```

about half of the projects in the community build depend, directly or indirectly, on either macro-paradise or kind-projector, so these are just one round of failures, fixing these would very likely uncover others.

I glanced through the logs and these all appear to be projects that have `-Xfatal-warnings` enabled, and there are new warnings, hence, failed. (note that we can turn off `-Xfatal-warnings` in a project's entry in the dbuild config if we want to)

@som-snytt would you like to look through these and see if any of them give you pause...?

the macro-paradise one is:

```
[macro-paradise] [error] -Ywarn-unused-import is deprecated: Enable -Ywarn-unused:imports
[macro-paradise] [error] one error found
```

in tut we see:

```
[tut] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/tut-04e719af31535d7e99ddb329a6d9811a743a23e3/core/src/main/scala/tut/Spigot.scala:8: private setter of baos in class Spigot is never used
[tut] [error]   private var baos = new ByteArrayOutputStream()
...
[tut] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/tut-04e719af31535d7e99ddb329a6d9811a743a23e3/core/src/main/scala/tut/Zed.scala:121: parameter value a in method void is never used
[tut] [error]   def void[A](a: A): Unit = ()
[tut] [error]               ^
```

the errors list in scala-partest is quite long, some unused imports, some unused parameters, not possible to know at a glance whether spurious.

pcplod has:

```
[pcplod] [info] Mr Plod
[pcplod] [info] - should typecheck a compilable valid noddy file *** FAILED ***
[pcplod] [info]   List(PcMessage(com/acme/foo.scala,Warning,parameter value a in method bar is never used)) was not empty (PcPlodTest.scala:37)
```

kind-projector looks like unused imports and a unused parameter or two.

som-snytt:
@SethTisue Thanks, I'll spend some time with it this evening.

tpolecat's `void` is especially poignant.

Probably we'll have to set that warning off under Xlint. Who knew people like unused params?

som-snytt:
I'm not comfortable reading the dbuild output yet. Why does the tut output show `Spigot.scala` in its own file? I only witness it in `Tut.scala`. The warning is correct. Also, apparently -Xlint doesn't warn about `private[this] var` that is never set. Spigot should spew a couple more warnings.

lrytz:
> Why does the tut output show Spigot.scala in its own file?

I don't understand the question here..

The partest log revealed the following problem:

```
scala> def f(x: Int)(y: Int = 1) = x + y
<console>:11: warning: parameter value x in method f$default$2 is never used
       def f(x: Int)(y: Int = 1) = x + y
             ^
f: (x: Int)(y: Int)Int
```

and also this one - maybe we have to back off for overridable methods?

```
scala> :pa

class C { def f(x: Int, y: Int) = y }
class D extends C { override def f(x: Int, y: Int) = x + y }

<console>:11: warning: parameter value x in method f is never used
       class C { def f(x: Int, y: Int) = y }
                       ^
```

The [`KindRewriter` class in kind-projector](https://github.com/non/kind-projector/blob/master/src/main/scala/KindProjector.scala#L24) takes an unused `plugin` parameter, which is a bit surprising, but the warning seems to be correct, the parameter is unused.

I just realized that maintaining binary compatibility can be a valid reason to keep an unused parameter. There's no mechanism for suppressing a warning locally..

lrytz:
I would suggest to turn `-Xlint:unused` into `-Ywarn-unused:-patvars,-params,_` (currently it's `-patvars,_`), then this change is fine for 2.12.2 IMO. The `-Ywarn-unused:params` implementation can be refined later.

som-snytt:
@lrytz thanks so much for looking. I agree about `-params`. I had an idea to facilitate exceptions: `@deprecated` unused params, but I haven't experimented yet.

I'll add a commit to turn off `-params` for the default linter.

My question about the tut output:

```
[tut] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/tut-04e719af31535d7e99ddb329a6d9811a743a23e3/core/src/main/scala/tut/Spigot.scala:8: private setter of baos in class Spigot is never used
```

I didn't see a file `Spigot.scala`, but maybe I was just too bleary-eyed.

I thought it avoided warning about synthetics and would avoid warning about defaults, but I guess there is a hole to plug.

It's a fundamental usage question, whether lint should warn about generated code; there was another question whether to warn about REPL synthetics. In general, can a lint rule be usefully triggered in generated code if it's not already triggered in the source text?

Looking at scalac warnings under -Xlint, I think missing-interpolator should give a pass to `"$conforms"` because normally you'd write `conforms.toString`, although in general an arbitrary interpolator could be doing an arbitrary operation on `conforms`.

BTW, when I say thanks for assistance with sussing this out, I mean it's a challenge coming back to this feature after time passes, and other commitments are pressing, so I really appreciate the boost. I know you all are very busy with the releases.

lrytz:
Thanks for the kind words!

https://github.com/tpolecat/tut/blob/master/core/src/main/scala/tut/Spigot.scala

som-snytt:
I expected https://github.com/tpolecat/tut/tree/v0.4.8/core/src/main/scala/tut but I'm probably picking up the wrong cues.

lrytz (approved):

SethTisue:
new community build run: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1397/consoleFull (404 til Jenkins finishes re-grouting the tub like he promised)

SethTisue:
test run finished, results ready for inspection

lrytz:
It looks good to me. Can you do the necessary workarounds for the community build?

som-snytt:
Actually `-Ywarn-unused -Xlint:-unused,_` don't play well together. As shown, you must disable the lint in order to enable `patvars,params`. Maybe `-Ywarn-unused -Xlint` would work if the hook on `unused` only enabled settings instead of disabling. Also the help text is stale.

lrytz:
Right, it would be good to get this piece in place before we release 2.12.2. It can be a follow-up if @SethTisue prefers to have this merged sooner.

som-snytt:
I'll submit to fix the settings in just a minute.

SethTisue:
> Can you do the necessary workarounds for the community build?

I've spun that out as https://github.com/scala/community-builds/issues/517. (summary: let's temporarily disable `-Xfatal-warnings` community-build-wide to get green for release, then do better later)

som-snytt:
I also detuned the deprecation of `-Ywarn-unused-import` [sic] which is in the paradise build. That complicates builds to no purpose, so it remains an alias for `-Ywarn-unused:imports`.

A trivial PR for KP https://github.com/non/kind-projector/pull/48 which sees the unused imports under `-Xlint`.

lrytz (approved):

lrytz (approved):
@SethTisue I'll let you merge it, as you're coordinating the 2.12.2 release.

SethTisue:
one more community build run with the latest: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1402/consoleFull

adriaanm:
One test failure in pcplod in the community build due to the additional warnings:
```
example/src/test/scala/org/ensime/testing/classes.scala:9: private method log in class Me is never used
class Me
      ^
example/src/test/scala/org/ensime/testing/classes.scala:12: private method log in class Myself is never used
class Myself(val foo: String, val bar: Long)
      ^
example/src/test/scala/org/ensime/testing/classes.scala:17: private method log in trait Mallgan is never used
trait Mallgan
      ^
example/src/test/scala/org/ensime/testing/classes.scala:25: private method log in class Foo is never used
class Foo(foo: String, bar: Long) {
      ^
example/src/test/scala/org/ensime/testing/classes.scala:25: private method log in object Foo is never used
class Foo(foo: String, bar: Long) {
         ^
example/src/test/scala/org/ensime/testing/classes.scala:33: private method log in class Baz is never used
class Baz[T](val fred: T)
      ^
example/src/test/scala/org/ensime/testing/classes.scala:36: private method log in class Mine is never used
class Mine(val foo: String = "foo", val bar: Long = 13)
      ^
example/src/test/scala/org/ensime/testing/classes.scala:38: private method log in class Covariant is never used
@noddy class Covariant[+I](item: I)
             ^
example/src/test/scala/org/ensime/testing/classes.scala:39: private method log in class Contravariant is never used
@noddy class Contravariant[-I](item: I)
             ^
9 warnings found
NoddyPcSpec:
@noddy
- should handle definitions of @noddy *** FAILED ***
  List(PcMessage(classes.scala,Warning,private method log in class Me is never used), PcMessage(classes.scala,Warning,private method log in object Me is never used), PcMessage(classes.scala,Warning,private method log in class Myself is never used), PcMessage(classes.scala,Warning,private method log in object Myself is never used), PcMessage(classes.scala,Warning,private method log in trait Mallgan is never used), PcMessage(classes.scala,Warning,private method log in object Foo is never used), PcMessage(classes.scala,Warning,private method log in class Foo is never used), PcMessage(classes.scala,Warning,private method log in class Baz is never used), PcMessage(classes.scala,Warning,private method log in object Baz is never used), PcMessage(classes.scala,Warning,private method log in class Mine is never used), PcMessage(classes.scala,Warning,private method log in object Mine is never used), PcMessage(classes.scala,Warning,private method log in class Covariant is never used), PcMessage(classes.scala,Warning,private method log in object Covariant is never used), PcMessage(classes.scala,Warning,private method log in class Contravariant is never used), PcMessage(classes.scala,Warning,private method log in object Contravariant is never used)) was not equal to List() (NoddyPcSpec.scala:16)
org.scalatest.exceptions.TestFailedException:
```

The following projects don't compile under -Xfatal-warnings:
  - tut    
  - kind-projector
  - scala-partest 

Should we add `extra.commands: ${vars.default-commands} [ "removeScalacOptions -Xfatal-warnings" ]` to their dbuild settings?

som-snytt:
I did a PR for kind-proj's unused imports, but not yet for tut's unused; I'll have a PR for partest. I think the answer is yes, cut them some slack for now. I didn't look at pcplod.

SethTisue:
> Should we add extra.commands: ${vars.default-commands} [ "removeScalacOptions -Xfatal-warnings" ] to their dbuild settings

Yes, I'll do that.

SethTisue:
the imps are wearing parkas today

fommil:
@som-snytt can this be disabled on a per-method basis. I have lots of genuine use cases (e.g. in mocks, or to shadow an implicit that I don't want in scope).

mpilquist:
Echoing @fommil -- we really need a use site disable annotation or some such. I compile everything with -Xfatal-warnings and don't want to give that up. As a result, upon upgrading to 2.12.2, I need to do stuff like this:
```scala
// On an implicit class for Traversable[A]:
def toMultiMap[K, V](implicit ev: A <:< (K, V)): scala.collection.immutable.Map[K, Set[V]] = {
  val _ = ev // Convince scalac that ev is used
  self.
    asInstanceOf[Traversable[(K, V)]].
    groupBy { case (k, v) => k }.
    mapValuesStrict { t => t.map { case (k, v) => v }.toSet }
}
```
Ed. changed `identity(ev)` to `val _ = ev` thanks to tip from @djspiewak.

SethTisue:
please open a bug at https://github.com/scala/bug, comments on already-merged PRs are ephemeral

fommil:
If anybody knows how to disable this warning across everything, I'd like to be able to do that. In combination with `-Xlint:_`

godenji:
@fommil -Xlint brings in `unused` which "Enable -Ywarn-unused:imports,privates,locals,implicits"

fommil:
@godenji yeah I know, but I want to now remove the ones introduced by this PR. I don't think it's possible to remove things once everything has been enabled, right?

som-snytt:
Just `-Xlint:-unused,_`. The same discussion was had about `-Xlint:missing-interpolator`, you either love it or hate it, but it's trivial to disable.

som-snytt:
@mpilquist It doesn't warn on unused params that are specified by superclass or that are deprecated.

`def f(@deprecated("unused","") x: Int) = ()`

But it is handier just to disable what annoys you. You can use `-Ywarn-unused` more selectively, of course.

(Deprecating a param has no other consequence; so it's a quirky but harmless workaround. Maybe someone will propose a different annotation at some point, or a reporter mechanism to suppress the warning.)

mpilquist:
@som-snytt Thanks. I love the new feature and want to keep it enabled (it's already found tons of unused implicits -- vestiges of various refactorings). I just need a way to disable it for witness style implicits, where the presence of the implicit param is enough. I'll try out the `deprecated` trick, as that won't generate any additional bytecode like the `val _ = ev` trick.

fommil:
@mpilquist but won't you now get warnings where you call that method?

godenji:
@fommil it's the interplay of Xlint and -Ywarn that got me; I didn't notice that -Xlint brings in `unused`, which then led to much head scratching as to why my -Ywarn negations weren't getting applied

fommil:
@som-snytt is there any way (when using `-Xlint:-unused,_`) to selectively turn on `imports,privates,locals,implicits`? I want the imports warning, but some of these other ones are giving me a lot of false positives.

som-snytt:
@fommil Yes, you can `-Ywarn-unused:help` to one's heart's content.

I'll take a look if this is mentioned somewhere besides here.

godenji:
@fommil use -Ywarn-unused and selectively negate

som-snytt:
@godenji I'd expect turning off an option to always override enablement, as a rule, but I'll check whether that is hono(u)red here. The class is called `MutableSettings` for a reason. paulp tried to make it immutable once...

fommil:
@som-snytt I don't have a `scalac` to do that on and I don't really want to have to start tricking `sbt` into doing this

@godenji if I knew how to selectively negate... :confused:

som-snytt:
@fommil Ah, you mean, in use-case speak, as an sbt user, how do I discover scalac sub-options!

People also ask for updated manpages on stackoverflow.

I'll check that out.

fommil:
`-Xlint -Ywarn-unused:imports,-patvars,-privates,-locals,-params,-implicits` I am still getting

```
[warn] /home/fommil/Projects/ensime-server/util/src/main/scala/org/ensime/util/enums.scala:36: parameter value gen in method fromSingletonByName is never used
[warn]       gen: Generic.Aux[A, C],
[warn]       ^
[warn] /home/fommil/Projects/ensime-server/util/src/main/scala/org/ensime/util/file.scala:72: parameter value cs in method readLines is never used
[warn]     def readLines()(implicit cs: Charset): List[String] = {
[warn]                              ^
```

fommil:
this *seems* to do what I want

```
    // WORKAROUND https://github.com/scala/scala/pull/5402
    scalacOptions := {
      val orig = scalacOptions.value
      CrossVersion.partialVersion(scalaVersion.value) match {
        case Some((2, 12)) => orig.map {
          case "-Xlint"               => "-Xlint:-unused,_"
          case "-Ywarn-unused-import" => "-Ywarn-unused:imports,-patvars,-privates,-locals,-params,-implicits"
          case other                  => other
        }
        case _             => orig
      }
    }
```

godenji:
@fommil it's because -Xlint enables `unused` which pulls in above referenced -Ywarn flags and there's *no* way override with -Ywarn flags

som-snytt:
@godenji it sounds like negating with -Ywarn isn't respected by -Xlint:_, we can either use https://github.com/scala/bug/issues/10270 for tweaking or you can create another issue.

som-snytt:
@fommil the explicit "negations" should be documentary (unnecessary) on `-Ywarn-unused`. Thanks for the sbt snippet.

godenji:
@som-snytt up to you, whatever makes it clear to the user what's going on. We have a workaround but likely many users will get tripped up by this.

fommil:
@som-snytt how can I suppress the warnings here?

```scala
  implicit def NamedTypeMemberInfoFormat: SexpFormat[NamedTypeMemberInfo] = {
    @deprecated("local", "") def NamedTypeMemberInfoFormat = ???
    implicitly[SexpFormat[NamedTypeMemberInfo]]
  }
```

the `@deprecated` doesn't work. It's when I'm mirroring an implicit with a non-implicit.

som-snytt:
That's a clever idiom.

I've never thought of `Sex` as having a `pFormat`, but I'll go with it. Or, how could it not have a `pFormat`?

I'm sure it's easy to exclude deprecated locals as with params.

Warning suppression has been bounced around a bit; hopefully there's a more general solution down the line.

fommil:
@som-snytt s-expressions :stuck_out_tongue_closed_eyes: this has found a tonne of unused things in ensime! Including a `var` that should be a `val` but until I can suppress this kind of pattern, I can't turn it on in the build.

som-snytt:
Thanks for your patience. I'll go with the easy albeit ugly workaround with deprecation.

fommil:
I'm also suffering the same problem as @mpilquist with the implicit evidence

fommil:
@som-snytt I have a false positive implicit in ensime-server if you're interested.

```
[warn] /home/fommil/Projects/ensime-server/core/src/main/scala/org/ensime/indexer/orientdb/SchemaFormat.scala:85: parameter value gen in method familySchemaFormat is never used
[warn]     gen: LabelledGeneric.Aux[T, Repr],
[warn]     ^
[warn] /home/fommil/Projects/ensime-server/core/src/main/scala/org/ensime/indexer/orientdb/SchemaFormat.scala:87: parameter value tpe in method familySchemaFormat is never used
[warn]     tpe: Typeable[T]
[warn]     ^
[warn] /home/fommil/Projects/ensime-server/core/src/main/scala/org/ensime/indexer/orientdb/SimpleOrient.scala:191: parameter value tpe in method insertE is never used
[warn]       tpe: shapeless.Typeable[E],
[warn]       ^
[warn] three warnings found
```

removing these implicits, `core` no longer compiles. Fun.

SethTisue:
@fommil can you isolate it and open a new bug in http://github.com/scala/bug?

fommil:
@SethTisue I have absolutely no idea what it could be. Seems to be about chaining of implicits. Chances are it'll come up again in spray-json-shapeless which is easier to minimise.

som-snytt:
@fommil again, it's not pretty, but you can deprecate the params to suppress warning. Or maybe I don't see it.

fommil:
@som-snytt oh, interesting, I thought that'd cause a deprecation warning at the point of use, and it didn't work for the implicit shadowing.

som-snytt:
@fommil yes only for params, and no other tool cares about deprecated params, I hope.

fommil:
I checked the 2.11 build and it seemed to accept that, so that's a reasonable workaround (would still be good to work out why there is a false positive here though)

som-snytt:
Those params are actually not used, so it's not a false pos, or do I miss the point?

fommil:
@som-snytt the ones in the `Orient` files... if I remove them, then the build fails. So they must be used.

som-snytt:
You mean graph.addEdge takes an implicit Typeable? I haven't deep-dived yet. That API is a dependency.

som-snytt:
I think we can chat forever on a merged PR and no one will care.

PR #4819, opened 2015-10-26
SI-7916: ScriptEngine support

som-snytt:
Refactor the ScriptEngine support to an adaptor atop the
IMain API.

Allow references to resolve to context attributes. (The
attributes must be defined at compilation time, though
they may resolve to updated values at evaluation time.)

This means that attributes are not bound statically in
REPL history. In particular, we forgo the trick of binding
attributes named "name: Type" as typed values.

Instead, an `x` bound in dynamic context is injected into
the script as a dynamic selection `$ctx.x` where `ctx`
performs the look-up in the script context.

When a compiled script is re-evaluated, a new instance of
the script class is created and defined symbols are
rebound.

PR #5404, opened 2016-09-16
Fixes reference to script engine factory in META-INF/services

rjolly:
This prevents the interpreter to run as a JSR-223 scripting engine (using e.g. jrunscript).


som-snytt:
Thanks! I didn't think to check the sbt build for that.

The value changed in https://github.com/scala/scala/pull/4819

som-snytt:
LGTM

SethTisue:
should this be tested in https://github.com/scala/scala-dist-smoketest/blob/2.12.x/scripts/jobs/release/smoketest ?

adriaanm:
Yeah, that would be a nice addition. Could you file a ticket to track that?

PR #5408, opened 2016-09-19
Enable MiMa for 2.12.0

szeiger:


szeiger:
Updated with the brand new MiMa 0.1.10

szeiger:
Build failure is due to higher parallelization of partest on the new machines (not enough memory). The actual MiMa results look good.

adriaanm:
LGTM

PR #5410, opened 2016-09-20
upgrade to scala-xml 1.0.6

SethTisue:
just because in general we want to ship the latest versions of the
modules, and some desirable-looking fixes went into 1.0.6

let's at least see if Jenkins passes

1.0.6 release notes: https://github.com/scala/scala-xml/releases/tag/v1.0.6

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -22,7 +22,7 @@ starr.version=2.12.0-RC1
 scala.binary.version=2.12.0-RC1
 
 # external modules shipped with distribution, as specified by scala-library-all's pom
-scala-xml.version.number=1.0.5
+scala-xml.version.number=1.0.6
 scala-parser-combinators.version.number=1.0.4
 scala-swing.version.number=2.0.0-M2
 scala-swing.version.osgi=2.0.0.M2


SethTisue:
probably not to be merged right away? it would be nice to have some reports from the field that the new version is usable

at the time this is merged, we should also bump scala-xml-version in the 2.12 community build

SethTisue:
oh, it's waiting on https://github.com/scala/scala-xml/issues/97 anyway which isn't quite done

SethTisue:
at the time this is merged, we should also bump `scala-xml-version` in the 2.12 community build

lrytz:
How about running a community build using the new version and merging this onSuccess?

adriaanm:
Let's just bump it in 2.12.1.

SethTisue:
the 2.12 community build already tracks master of scala-xml, and properly includes it in the Scala used by the rest of the projects, so there's no need to run that specially.

waiting for 2.12.1 is sensible.

SethTisue:
re my comment above, scala-xml-ref no longer exists so my comment no longer applies.

SethTisue:
fyi @biswanaths @ashawley

ashawley:
Are there a lot of people who use `scala-library-all`?

In addition to getting shipped, does incrementing the version change the version used in parts of the compiler system?

SethTisue:
> Are there a lot of people who use scala-library-all?

I doubt it. Only beginners ought to use it (unless nobody at all should use it, which you could certainly argue), and I don't see beginner resources recommending it, so I bet it's used pretty much never. (That's just off the top of my head though, I haven't tried looking at e.g. Maven download numbers.)

I think we could axe it for 2.13, but it may be that by then something will be happening with the Scala Platform that will change all this around.

> In addition to getting shipped, does incrementing the version change the version used in parts of the compiler system

yes

ashawley:
Thanks, Seth. Good to know.

Yeah, usage seems pretty low for scala-library-all.

https://mvnrepository.com/artifact/org.scala-lang/scala-library-all

ashawley:
I see no reason to rush shipping 1.0.6 in to the 2.12.0 release.  I predict people will more likely just use scala-xml package directly, if they really want to use 1.0.6 and the fixes it contains.

SethTisue:
@adriaanm let's merge?

SethTisue:
oops, we missed that 1.0.6 hasn't actually been published for 2.12

SethTisue:
> oops, we missed that 1.0.6 hasn't actually been published for 2.12

fixed: https://github.com/scala/scala-xml/issues/120

PR #5412, opened 2016-09-21
SD-226 Be lazier in Fields info transform for better performance

retronym:
Only mixin fields + accessors into class infos of classes that are
either in the current run, or appear in a superclass chain of a class
in the current run.

This is analagous to what happens in the mixin phase.

Review by @adriaanm 

I'll post some benchmark results below.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Fields.scala
+++ b/src/compiler/scala/tools/nsc/transform/Fields.scala
@@ -305,6 +305,10 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
       lazyCallingSuper setInfo tp
     }
 
+    private def needsMixin(cls: Symbol): Boolean = {
+      !(cls.isPackageClass || cls.isJavaDefined) && (currentRun.compiles(cls) || refChecks.isSeparatelyCompiledScalaSuperclass(cls))
+    }
+
     def apply(tp0: Type): Type = tp0 match {
       // TODO: make less destructive (name changes, decl additions, flag setting --
       // none of this is actually undone when travelling back in time using atPhase)
@@ -360,9 +364,12 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
           ClassInfoType(parents, allDecls, clazz)
         } else tp
 
-      // mix in fields & accessors for all mixed in traits
 
-      case tp@ClassInfoType(parents, oldDecls, clazz) if !clazz.isPackageClass =>
+      case tp@ClassInfoType(parents, oldDecls, clazz) if !needsMixin(clazz) => tp
+
+      // mix in fields & accessors for all mixed in traits
+      case tp@ClassInfoType(parents, oldDecls, clazz) =>
+
         val site = clazz.thisType
 
         // setter conflicts cannot arise independently from a getter conflict, since a setter without a getter does not a val definition make
--- a/src/compiler/scala/tools/nsc/transform/Fields.scala
+++ b/src/compiler/scala/tools/nsc/transform/Fields.scala
@@ -305,7 +305,7 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
       lazyCallingSuper setInfo tp
     }
 
-    private def needsMixin(cls: Symbol): Boolean = {
+    private def classNeedsInfoTransform(cls: Symbol): Boolean = {
       !(cls.isPackageClass || cls.isJavaDefined) && (currentRun.compiles(cls) || refChecks.isSeparatelyCompiledScalaSuperclass(cls))
     }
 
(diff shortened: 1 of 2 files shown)


retronym:
```
[info] Benchmark                                           (_scalaVersion)  (extraArgs)  (source)    Mode  Cnt     Score   Error  Units
[info] HotScalacBenchmark.compile                           2.11.8               better-files  sample  742  498.946 ± 1.597  ms/op
[info] HotScalacBenchmark.compile                  2.12.0-6df14e5-SNAPSHOT               better-files  sample  595  625.133 ± 2.822  ms/op
[info] HotScalacBenchmark.compile                  2.12.0-c0450f0-SNAPSHOT               better-files  sample  677  545.357 ± 2.928  ms/op
[info] HotScalacBenchmark.compile                           2.11.8                 scalap  sample  324  1177.065 ± 4.638  ms/op
[info] HotScalacBenchmark.compile                  2.12.0-6df14e5-SNAPSHOT                 scalap  sample  320  1209.683 ± 9.359  ms/op
[info] HotScalacBenchmark.compile                  2.12.0-c0450f0-SNAPSHOT                 scalap  sample  342  1111.957 ± 3.753  ms/op
```

retronym:
I have some additional optimizations in https://github.com/scala/scala/compare/2.12.0...retronym:opt/settings?expand=1

which in aggregate improve performance further to:

```
[info] HotScalacBenchmark.compile                  2.12.0-2ddbc8f-SNAPSHOT               better-files  sample  735  501.089 ± 1.742  ms/op
[info] HotScalacBenchmark.compile                  2.12.0-2ddbc8f-SNAPSHOT                 scalap  sample  360  1047.903 ± 3.885  ms/op
```

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -360,9 +364,12 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
           ClassInfoType(parents, allDecls, clazz)
         } else tp
 
+
+      case tp@ClassInfoType(parents, oldDecls, clazz) if !needsMixin(clazz) => tp
```
retronym: Thinking about this some more, we probably need to guard this with `globalPhase.isAfterRefchecks`, as we only populate the map of separated compiled superclasses in the refchecks tree transform.

If something at or before refchecks where to run `exitingFields(cls.info)`, we would get here before that map was populated and falsely skip this info transform.

retronym: I've added a commit to conservatively approximate this as false.

adriaanm: ~~I'm not sure this will fly, since we also need to hit the next case if there's a module or lazy val that needs expansion, for example.~~ <-- scratch that, but would be good to rename `needsMixin`, since it's more than just about mixing in vals now

retronym: Renamed to `classNeedsInfoTransform`

retronym:
I've just repeated some benchmarking after rebasing/restarring on the synchronized performance fix and the other major open PRs:

```
* 43535e8 Bump to prerelease SBT
* c3c4bba Bump sbt.version to 0.13.12, without breaking
* cb39978 Restarr on faster compiler
*   4aff108 Merge remote-tracking branch 'origin/pr/5419' into merge/2.12-prs
|\
| * e735c34 Avoid needless Boolean.unbox in hot code paths.
| * 0d74b87 Cache setting-derived values, like isScala211, in the Run
*   ab6507a Merge remote-tracking branch 'origin/pr/5412' into merge/2.12-prs
|\
| * e07585c Make isSeparateCompiled... robust against rogue phase time travel
| * c0450f0 SD-226 Be lazier in Fields info transform for better performance
*   a3a8bdb Merge remote-tracking branch 'origin/pr/5397' into merge/2.12-prs
|\
| * e3e1e30 SI-9920 Avoid linkage errors with captured local objects + self types
*   62552f3 Merge remote-tracking branch 'origin/pr/5388' into merge/2.12-prs
|\
| * 5f64ee5 Cast more pro-actively in synthetic accessor trees.
| * 9a39e0c Avoid mismatched symbols in fields phase
* 9212340 Restarr on synchronized fix
* 3ed4496 Merge remote-tracking branch 'origin/pr/5417' into merge/2.12-prs
* e994c1c SD-233 synchronized blocks are JIT-friendly again
```

```
Benchmark                                           (_scalaVersion)  (extraArgs)      (source)    Mode  Cnt    Score   Error  Units
HotScalacBenchmark.compile                  2.12.0-ab6507a-SNAPSHOT               better-files  sample  522  468.867 ± 2.662  ms/op
HotScalacBenchmark.compile                  2.12.0-a3a8bdb-SNAPSHOT               better-files  sample  649  570.785 ± 3.450  ms/op
HotScalacBenchmark.compile                  2.12.0-ab6507a-SNAPSHOT                 scalap  sample  249  1004.473 ± 6.369  ms/op
HotScalacBenchmark.compile                  2.12.0-a3a8bdb-SNAPSHOT                 scalap  sample  348  1098.468 ± 6.179  ms/op
```

After that we still see a big improvement from the laziness here: compile time changes by 0.82x for better-files, 0.91x for scalap.

adriaanm:
LGTM

PR #5414, opened 2016-09-22
Do not build partest-javaagent and partest-extras for `pack`

szeiger:
Building them as part of `dist/mkQuick` (and thus, by extension,
`dist/mkPack`) was not necessary in the first place. Partest does not
rely on these tasks for its dependencies.

And when we do build the jars, they now go into their standard location
under `target` instead of `build/pack/lib` so they don’t confuse
sbt (see https://github.com/sbt/sbt/issues/2748).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -510,8 +510,10 @@ lazy val scalap = configureAsSubproject(project)
   )
   .dependsOn(compiler)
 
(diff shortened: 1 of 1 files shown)


dwijnand (approved):
thank you :)

adriaanm:
LGTM

adriaanm:
Adding this one to my mental list of unexpected consequences of insourcing scalacheck.

PR #5415, opened 2016-09-22
WIP: SD-228 eliminate static super accessors in traits

lrytz:
WIP, don't merge: the first commit updates starr to a version on scala-release-temp that corresponds to the tip of this PR. This way I can find out if all commits in this PR pass CI validation.

The temporary starr is build from this branch: https://github.com/lrytz/scala/tree/traitSuperAccessorsTmp
It's the same as this PR, but without the starr update in the first commit. The bootstrap job that built it passed all tests (https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-bootstrap/593/)

For submission, the right procedure is probably to keep the current starr, have CI fail on the commits, and then re-starr right after merging it.

PR #5417, opened 2016-09-26
SD-233 synchronized blocks are JIT-friendly again

retronym:
GenBCode, the new backend in Scala 2.12, subtly changed
the way that synchronized blocks are emitted.

It used `java/lang/Throwable` as an explicitly named exception
type, rather than implying the same by omitting this in bytecode.

This appears to confuse HotSpot JIT, which reports a error
parsing the bytecode into its IR which leaves the enclosing method
stuck in interpreted mode.

This commit passes a `null` descriptor to restore the old pattern
(the same one used by javac.) I've checked that the JIT warnings
are gone and that the method can be compiled again.

Fixes https://github.com/scala/scala-dev/issues/233

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSyncAndTry.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSyncAndTry.scala
@@ -73,9 +73,11 @@ abstract class BCodeSyncAndTry extends BCodeBodyBuilder {
       /* ------ (4) exception-handler version of monitor-exit code.
        *            Reached upon abrupt termination of (2).
        *            Protected by whatever protects the whole synchronized expression.
+       *            null => "any" exception in bytecode, like we emit for finally.
+       *            Important not to use j/l/Throwable which dooms the method to a life of interpretation! (SD-233)
        * ------
        */
-      protect(startProtected, endProtected, currProgramPoint(), jlThrowableRef)
+      protect(startProtected, endProtected, currProgramPoint(), null)
       locals.load(monitor)
       emit(asm.Opcodes.MONITOREXIT)
       emit(asm.Opcodes.ATHROW)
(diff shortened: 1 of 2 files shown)


retronym:
Review by @lrytz

lrytz:
LGTM

Another lesson in the arts of taming the mighty JIT gods..

lrytz:
@retronym here's a regression test: https://github.com/lrytz/scala/tree/sd233-test

I tried pushing to your branch, but 

```
 ! [remote rejected] ticket/SD-233 -> ticket/SD-233 (permission denied)
```

Is "Allow edits from maintainers" checked? https://github.com/blog/2247-improving-collaboration-with-forks

retronym:
@lrytz I haven't seen that option when pushing to scala/scala. I did see it pushing to sbt/sbt. Not sure why.

I've pushed your commit with the test, thanks!

I'd like to merge this soon and then restarr the 2.12.0 branch. I can then rebase my other performance oriented PRs and benchmark them (along with the residual commits in https://github.com/scala/scala/compare/2.12.x...retronym:opt/settings?expand=1) to see which ones still are worthwhile.

lrytz:
I don't see it when creating a PR, but in my PR https://github.com/scala/scala/pull/5415 it's there on the right

image

retronym:
Clicked!

lrytz:
OK, so it was off - for some reason it was on in my PR by default..

retronym:
I created the PR with the command line tool, `hub`. Perhaps the API defaults to the status quo, whereas the UI is different.

adriaanm:
Wow! Could you add a comment to the code as well? Also, I would be curious to see if this removes the need for the compute method for modules.

retronym:
@adriaanm Comment added. I'll benchmark the compiler bootstrapped through this change and reversion of the module lzycompute method, but I'd like to do that as a followup task.

adriaanm:
Thanks, sounds good!

retronym:
Restarring proposed: #5422

PR #5418, opened 2016-09-26
Optimize javaBinaryName callers

retronym:
... by calling javaBinaryNameString, instead.

They all are happy with a throw away String, there is no advantage
to interning this into the name table.

Review by @lrytz


lrytz:
LGTM, thanks

PR #5420, opened 2016-09-26
expunge a stray reference to sbaz in the Scala man page

SethTisue:
and fix a typo while we're at it


som-snytt:
LGTM

`-Ylog-classpath` confirms that cwd is at the end of the class path.

PR #5421, opened 2016-09-26
SI-9936 SeqLike.indexWhere starts at zero

som-snytt:
This follows the Scaladoc, and makes

```
"abcdef".indexOf('c', -1)
```

work like

```
"abcdef".toVector.indexOf('c', -1)
```

### DIFF
--- a/src/library/scala/collection/SeqLike.scala
+++ b/src/library/scala/collection/SeqLike.scala
@@ -113,7 +113,7 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
   }
 
   def indexWhere(p: A => Boolean, from: Int): Int = {
-    var i = from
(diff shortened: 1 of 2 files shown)


lrytz:
LGTM, thanks!

tzachz:
Should there be a similar fix for what seems to be the same issue in `LineraSeqOptimized`? See [here](https://github.com/scala/scala/blob/v2.12.1/src/library/scala/collection/LinearSeqOptimized.scala#L294). 

```scala
scala> "abcde".indexOf('c', -1)
res3: Int = 2

scala> "abcde".toVector.indexOf('c', -1)
res4: Int = 2

scala> "abcde".toList.indexOf('c', -1)
res5: Int = 1

```

som-snytt:
@tzachz Thanks!

PR #5424, opened 2016-09-27
SI-4700 Add `@infix` annotation for type printing

milessabin:
Cherry pick of #5401 onto 2.11.x. All credit belong to @VladUreche.

```
scala> import scala.annotation.infix
import scala.annotation.infix

scala> @infix class &&[T, U]
defined class $amp$amp

scala> def foo: Int && Boolean = ???
foo: Int && Boolean
```

### DIFF
--- a/bincompat-forward.whitelist.conf
+++ b/bincompat-forward.whitelist.conf
@@ -387,6 +387,10 @@ filter {
     {
         matchName="scala.concurrent.impl.Promise.toString"
         problemName=MissingMethodProblem
+    },
+    {
+        matchName="scala.annotation.showAsInfix"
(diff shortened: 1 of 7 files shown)


paulp:
If you add this capability but do not default to printing operators infix, we will spend years with a patchwork of infix-and-not, helplessly waiting for whatever libraries turn out to lag, or worse, forking them just to print the types right.

It's so, so easy to see it coming.

Just print operators infix.

At the very (very) least, add a compiler option to do it.

SethTisue:
The design and implementation need to be worked out over at #5401 first, so I don't think this is going to make 2.11.x. (@adriaanm can reopen if he disagrees)

PR #5425, opened 2016-09-27
SI-7707 Case apply transform private access

som-snytt:
Case.apply transform is allowed from companion,
even if constructor has restricted access.

Default args are also transformed, from the
apply defaults to the ctor defaults. Usually
those expressions are identical.

Unused check is moved to refcheck, since it
is a ref check, and since the check must be
done after the case.apply transform.


som-snytt:
There's no change to the unused check, which is relocated to refcheck, so if this passes tests, probably better to tack it onto https://github.com/scala/scala/pull/5402 which does touch the check.

PR #5413, opened 2016-09-21
Make output of errors in testAll less verbose

szeiger:
In order to fix the far too verbose output from https://github.com/scala/scala/pull/5397#issuecomment-248588645.

Previously a failure in an early task with lots of dependent tasks could lead to many megabytes of log output (or even gigabytes? or non-termination?) because the deeply nested failure tree would be traversed and printed in entirety. Now we only print the root cause and its nearest enclosing task for each top-level task, and we prune the tree to remove all duplicates.

A worst-case failure in `library/compile` (relatively early in the dependency tree; affects every possible test) now looks like this:

```
> testAll
[info] Compiling 1 Scala source to /Users/szeiger/code/scala/build/quick/classes/library...
[error] /Users/szeiger/code/scala/src/library/scala/Equals.scala:30: expected class or object definition
[error] xxx
[error] ^
[error] one error found
[info] Packaging /Users/szeiger/code/scala/build/pack/lib/scala-partest-javaagent.jar ...
[info] Done packaging.
[error] 11 of 11 test tasks failed:
[error] - junit/test
[error]   - library/compile:compileIncremental failed: Compilation failed
[error] - partest run
[error]   - (library/compile:compileIncremental failed)
[error] - partest pos neg jvm
[error]   - (library/compile:compileIncremental failed)
[error] - partest res scalap specialized scalacheck
[error]   - (library/compile:compileIncremental failed)
[error] - partest instrumented presentation
[error]   - (library/compile:compileIncremental failed)
[error] - partest --srcpath scaladoc
[error]   - (library/compile:compileIncremental failed)
[error] - osgiTestFelix/test
[error]   - (library/compile:compileIncremental failed)
[error] - osgiTestEclipse/test
[error]   - (library/compile:compileIncremental failed)
[error] - library/mima
[error]   - (library/compile:compileIncremental failed)
[error] - reflect/mima
[error]   - (library/compile:compileIncremental failed)
[error] - doc
[error]   - (library/compile:compileIncremental failed)
[trace] Stack trace suppressed: run last root/*:testAll for the full output.
[error] (root/*:testAll) java.lang.RuntimeException
[error] Total time: 1 s, completed Sep 21, 2016 5:27:46 PM
>
```

PR #5427, opened 2016-09-28
CI validation: run test:compile before testAll

retronym:
Seeks to fail fast and in an normal manner for bootstrapping errors,
rather than the enourmous diagnostic error seen in:

https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/3175/consoleText

Review by @szeiger

### DIFF
--- a/scripts/jobs/validate/test
+++ b/scripts/jobs/validate/test
@@ -23,6 +23,7 @@ case $prDryRun in
        --warn \
        "setupValidateTest $prRepoUrl" \
        $testExtraArgs \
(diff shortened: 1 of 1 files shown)


szeiger:
Requiring `test:compile` would prevent partest from running when there's a compilation error in one of the JUnit tests. The goal of `testAll` is to avoid these unnecessary early failures and get a complete error report. You could run `compile` first but with https://github.com/scala/scala/pull/5413 this is no longer necessary.

PR #3711, opened 2014-05-05
SI-8549 Serialization: fix regression with @SerialVersionUID / start enforcing backwards compatibility

retronym:

PR #5429, opened 2016-09-28
Default -Xmixin-force-forwarders to true

lrytz:
Also eliminates the warning when a mixin forwarder cannot be implemented because the target method is a java-defined default method in an interface that is not a direct parent of the class.

Adds some test cases, and fixes https://github.com/scala/scala-dev/issues/224.
## Discussion Summary

Since https://github.com/scala/scala/pull/5085, classes inheriting default methods no longer get a forwarder method that delegates to the default method (except in a few corner cases, like trait methods overriding class methods). The JVMs method resolution will pick the correct method.

This PR reverts that change.

We realized that missing forwarders cause a performance regression in startup time (www.scala-lang.org/blog/2016/07/08/trait-method-performance.html). Further benchmark numbers are available in comments in this PR. In summary:
- Startup is slower without forwarders
- For a large program with a short running time (compiling a small file) the difference can be significant (> 40%)
- Hot performance is not affected
- There are probably multiple underlying causes (CHA, populating vtables), we don't have a clear picture

### Two Forwarders

When trait forwarders are added, invoking a trait method involves indirection through **two** forwarders:

``` scala
trait T { def f = 1 }
class A extends T   // generates `def f = super[T].f`
```

generates

``` java
class A implements T {
    public int f() { return T.f$(this); }
}
interface T {
    default public static int f$(T $this) { return $this.f(); }

    default public int f() { return 1; }
}
```

The reason for having a static accessor method (`f$`) and using it for the super calls is summarized in https://github.com/scala/scala-dev/issues/228.

We could not identify any performance issues due to having two forwarders. Note that the JVM will unconditionally inline small methods (< 35 bytes) without querying any heuristics, the forwarders fall into this category.
### Serialization Stability

If a class does not have an explicit `@SerialVersionUID`, it is computed by the VM. The computed value is sensitive to changes in the classfile that don't affect deserialization, for example adding a public method.

This means that we cannot change our strategy for mixin forwarders in a 2.12.x minor release, as it would affect serialization compatibility. So even if the performance issue with missing forwarders is fixed in a future JVM update, we need to keep the forwarders. We can re-consider for 2.13.

See also https://github.com/scala/scala-dev/issues/237.


lrytz:
Test failure due to a changed `serialVersionUID` of `scala.reflect.ClassTag$GenericClassTag` (there's no explicit value, the generated one changes due to the addition of mixin forwarders). The test was added in https://github.com/scala/scala/pull/3711 for SI-8549 to ensure serialization stability.

Looking at the class hierarchy around `ClassTag` and `Manifest`, it seems the only class that has a `serialVersionUID` is `AnyValManifest`, where the hierarchy is something like:

```
trait ClassTag                      // extends Serializable
|- class GenericClassTag
|- trait Manifest
   |- class ClassTypeManifest
   |- class SingletonTypeManifest
   |- ...
   |- abstract class AnyValManifest // has @SerialVersionUID
      |- class DoubleManifest
      |- ...
```

A [comment in #3711](https://github.com/scala/scala/pull/3711#issuecomment-42175413) says

> the annotation [`@SerialVersionUID`] can be defined up the class hierarchy

I think this comment is not correct (or it means something else than I think it means):

```
$ cat Test.scala
@SerialVersionUID(1L)
class A extends Serializable
class B extends A

$ scalac Test.scala && serialver -classpath .:/Users/luc/scala/scala-2.12.0-RC1/lib/scala-library.jar B
B:    private static final long serialVersionUID = -4759527637665705469L;

$ cat Test.scala
@SerialVersionUID(1L)
class A extends Serializable
class B extends A { def f = 1 }

$ scalac Test.scala && serialver -classpath .:/Users/luc/scala/scala-2.12.0-RC1/lib/scala-library.jar B
B:    private static final long serialVersionUID = -6380942902981420110L;
```

The `@SerialVersionUID` annotation on `AnyValManifest` was added in the same PR (https://github.com/scala/scala/pull/3711 / e7ab41e), probably with the intention to have a stable value for all subclasses.

I think the right solution here is to add explicit `@SerialVersionUID` annotations to all concrete Manifest / ClassTag subclasses.

adriaanm:
LGTM

lrytz:
The SerialVersionUID test failed for `Set$EmptySet$` now. I patched it up, but the way we handle this annotation in the std lib is really ad-hoc..

adriaanm:
rebased with update to t8549 (git diff ffc8e3e..0e0614c):

``` diff
diff --git a/test/files/run/t8549.scala b/test/files/run/t8549.scala
index 2bf648f..7ec3635 100644
--- a/test/files/run/t8549.scala
+++ b/test/files/run/t8549.scala
@@ -79,7 +79,7 @@ object Test extends App {
     }
   }

-  // Generated on 20160930-19:54:01 with Scala version 2.12.0-local-d86377e)
+  // Generated on 20160930-16:09:23 with Scala version 2.12.0-local-ffc8e3e)
   overwrite.foreach(updateComment)

   check(Some(1))("rO0ABXNyAApzY2FsYS5Tb21lESLyaV6hi3QCAAFMAAV2YWx1ZXQAEkxqYXZhL2xhbmcvT2JqZWN0O3hyAAxzY2FsYS5PcHRpb27+aTf92w5mdAIAAHhwc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAAE=")
@@ -163,7 +163,7 @@ object Test extends App {
   // TODO SI-8576 Uninitialized field: IndexedSeqLike.scala: 56
   // check(immutable.Stream(1, 2, 3))( "rO0ABXNyACZzY2FsYS5jb2xsZWN0aW9uLmltbXV0YWJsZS5TdHJlYW0kQ29uc/ekjBXM3TlFAgADTAACaGR0ABJMamF2YS9sYW5nL09iamVjdDtMAAV0bEdlbnQAEUxzY2FsYS9GdW5jdGlvbjA7TAAFdGxWYWx0ACNMc2NhbGEvY29sbGVjdGlvbi9pbW11dGFibGUvU3RyZWFtO3hyACFzY2FsYS5jb2xsZWN0aW9uLmltbXV0YWJsZS5TdHJlYW0552RDntM42gIAAHhwc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAAFzcgAtc2NhbGEuY29sbGVjdGlvbi5JdGVyYXRvciQkYW5vbmZ1biR0b1N0cmVhbSQxRWR4We0SX0UCAAFMAAYkb3V0ZXJ0ABtMc2NhbGEvY29sbGVjdGlvbi9JdGVyYXRvcjt4cHNyAChzY2FsYS5jb2xsZWN0aW9uLkluZGV4ZWRTZXFMaWtlJEVsZW1lbnRzGF+1cBwmcx0CAANJAANlbmRJAAVpbmRleEwABiRvdXRlcnQAIUxzY2FsYS9jb2xsZWN0aW9uL0luZGV4ZWRTZXFMaWtlO3hwAAAAAwAAAAFzcgArc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLldyYXBwZWRBcnJheSRvZkludMmRLBcI15VjAgABWwAFYXJyYXl0AAJbSXhwdXIAAltJTbpgJnbqsqUCAAB4cAAAAAMAAAABAAAAAgAAAANw")

-  check(immutable.TreeSet[Int]())(   "rO0ABXNyACJzY2FsYS5jb2xsZWN0aW9uLmltbXV0YWJsZS5UcmVlU2V0sRdVIDjbWAsCAAJMAAhvcmRlcmluZ3QAFUxzY2FsYS9tYXRoL09yZGVyaW5nO0wABHRyZWV0AC5Mc2NhbGEvY29sbGVjdGlvbi9pbW11dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQkKTb4nP6aAqoCAAB4cHA=")
+  check(immutable.TreeSet[Int]())(   "rO0ABXNyACJzY2FsYS5jb2xsZWN0aW9uLmltbXV0YWJsZS5UcmVlU2V0sRdVIDjbWAsCAAJMAAhvcmRlcmluZ3QAFUxzY2FsYS9tYXRoL09yZGVyaW5nO0wABHRyZWV0AC5Mc2NhbGEvY29sbGVjdGlvbi9pbW11dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQk8u7cgruVzucCAAB4cHA=")

   // TODO SI-8576 unstable under -Xcheckinit
   // check(immutable.TreeSet(1, 2, 3))( "rO0ABXNyACJzY2FsYS5jb2xsZWN0aW9uLmltbXV0YWJsZS5UcmVlU2V0sRdVIDjbWAsCAAJMAAhvcmRlcmluZ3QAFUxzY2FsYS9tYXRoL09yZGVyaW5nO0wABHRyZWV0AC5Mc2NhbGEvY29sbGVjdGlvbi9pbW11dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQkC4BMdr1Z51wCAAB4cHNyADFzY2FsYS5jb2xsZWN0aW9uLmltbXV0YWJsZS5SZWRCbGFja1RyZWUkQmxhY2tUcmVlzRxnCKenVAECAAB4cgAsc2NhbGEuY29sbGVjdGlvbi5pbW11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWVrqCSyHJbsMgIABUkABWNvdW50TAADa2V5dAASTGphdmEvbGFuZy9PYmplY3Q7TAAEbGVmdHEAfgACTAAFcmlnaHRxAH4AAkwABXZhbHVlcQB+AAh4cAAAAANzcgARamF2YS5sYW5nLkludGVnZXIS4qCk94GHOAIAAUkABXZhbHVleHIAEGphdmEubGFuZy5OdW1iZXKGrJUdC5TgiwIAAHhwAAAAAnNxAH4ABgAAAAFzcQB+AAoAAAABcHBzcgAXc2NhbGEucnVudGltZS5Cb3hlZFVuaXR0pn1HHezLmgIAAHhwc3EAfgAGAAAAAXNxAH4ACgAAAANwcHEAfgAQcQB+ABA=")
@@ -179,12 +179,12 @@ object Test extends App {
   check(mutable.HashMap())( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuSGFzaE1hcAAAAAAAAAABAwAAeHB3DQAAAu4AAAAAAAAABAB4")
   check(mutable.HashMap(1 -> 1))( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuSGFzaE1hcAAAAAAAAAABAwAAeHB3DQAAAu4AAAABAAAABABzcgARamF2YS5sYW5nLkludGVnZXIS4qCk94GHOAIAAUkABXZhbHVleHIAEGphdmEubGFuZy5OdW1iZXKGrJUdC5TgiwIAAHhwAAAAAXEAfgAEeA==")
   check(mutable.HashSet(1, 2, 3))( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuSGFzaFNldAAAAAAAAAABAwAAeHB3DQAAAcIAAAADAAAABQBzcgARamF2YS5sYW5nLkludGVnZXIS4qCk94GHOAIAAUkABXZhbHVleHIAEGphdmEubGFuZy5OdW1iZXKGrJUdC5TgiwIAAHhwAAAAAXNxAH4AAgAAAAJzcQB+AAIAAAADeA==")
-  check(mutable.TreeMap[Int, Int]())( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZU1hcNx8qC229ZvwAgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZU1hcCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQkKTb4nP6aAqoCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAABw")
-  check(mutable.TreeMap(1 -> 1, 3 -> 6))( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZU1hcNx8qC229ZvwAgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZU1hcCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQkKTb4nP6aAqoCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAAJzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSROb2RlGxHsFtValgACAAZaAANyZWRMAANrZXl0ABJMamF2YS9sYW5nL09iamVjdDtMAARsZWZ0cQB+AAdMAAZwYXJlbnRxAH4AB0wABXJpZ2h0cQB+AAdMAAV2YWx1ZXEAfgAKeHAAc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAAFwcHNxAH4ACQFzcQB+AAwAAAADcHEAfgALcHNxAH4ADAAAAAZxAH4ADg==")
-  check(mutable.TreeMap(1 -> 1, 3 -> 6).range(1, 2))( "rO0ABXNyACxzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZU1hcCRUcmVlTWFwVmlldx7MCZxLhVQ8AgADTAAGJG91dGVydAAiTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9UcmVlTWFwO0wABGZyb210AA5Mc2NhbGEvT3B0aW9uO0wABXVudGlscQB+AAJ4cgAgc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlRyZWVNYXDcfKgttvWb8AIAAkwACG9yZGVyaW5ndAAVTHNjYWxhL21hdGgvT3JkZXJpbmc7TAAmc2NhbGEkY29sbGVjdGlvbiRtdXRhYmxlJFRyZWVNYXAkJHRyZWV0ACxMc2NhbGEvY29sbGVjdGlvbi9tdXRhYmxlL1JlZEJsYWNrVHJlZSRUcmVlO3hwc3IAGHNjYWxhLm1hdGguT3JkZXJpbmckSW50JCk2+Jz+mgKqAgAAeHBzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSRUcmVlAEynNPA1phUCAAJJAARzaXplTAAEcm9vdHQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJE5vZGU7eHAAAAACc3IAKnNjYWxhLmNvbGxlY3Rpb24ubXV0YWJsZS5SZWRCbGFja1RyZWUkTm9kZRsR7BbVWpYAAgAGWgADcmVkTAADa2V5dAASTGphdmEvbGFuZy9PYmplY3Q7TAAEbGVmdHEAfgAKTAAGcGFyZW50cQB+AApMAAVyaWdodHEAfgAKTAAFdmFsdWVxAH4ADXhwAHNyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAABcHBzcQB+AAwBc3EAfgAPAAAAA3BxAH4ADnBzcQB+AA8AAAAGcQB+ABFzcQB+AANxAH4ACHEAfgALc3IACnNjYWxhLlNvbWURIvJpXqGLdAIAAUwABXZhbHVlcQB+AA14cgAMc2NhbGEuT3B0aW9u/mk3/dsOZnQCAAB4cHEAfgARc3EAfgAWc3EAfgAPAAAAAg==")
-  check(mutable.TreeSet[Int]())( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZVNldM10nxFQDpt4AgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZVNldCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQkKTb4nP6aAqoCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAABw")
-  check(mutable.TreeSet(1, 3))( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZVNldM10nxFQDpt4AgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZVNldCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQkKTb4nP6aAqoCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAAJzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSROb2RlGxHsFtValgACAAZaAANyZWRMAANrZXl0ABJMamF2YS9sYW5nL09iamVjdDtMAARsZWZ0cQB+AAdMAAZwYXJlbnRxAH4AB0wABXJpZ2h0cQB+AAdMAAV2YWx1ZXEAfgAKeHAAc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAAFwcHNxAH4ACQFzcQB+AAwAAAADcHEAfgALcHBw")
-  check(mutable.TreeSet(1, 3).range(1, 2))( "rO0ABXNyACxzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZVNldCRUcmVlU2V0Vmlld2JdAzqy0DpGAgADTAAGJG91dGVydAAiTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9UcmVlU2V0O0wABGZyb210AA5Mc2NhbGEvT3B0aW9uO0wABXVudGlscQB+AAJ4cgAgc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlRyZWVTZXTNdJ8RUA6beAIAAkwACG9yZGVyaW5ndAAVTHNjYWxhL21hdGgvT3JkZXJpbmc7TAAmc2NhbGEkY29sbGVjdGlvbiRtdXRhYmxlJFRyZWVTZXQkJHRyZWV0ACxMc2NhbGEvY29sbGVjdGlvbi9tdXRhYmxlL1JlZEJsYWNrVHJlZSRUcmVlO3hwc3IAGHNjYWxhLm1hdGguT3JkZXJpbmckSW50JCk2+Jz+mgKqAgAAeHBzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSRUcmVlAEynNPA1phUCAAJJAARzaXplTAAEcm9vdHQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJE5vZGU7eHAAAAACc3IAKnNjYWxhLmNvbGxlY3Rpb24ubXV0YWJsZS5SZWRCbGFja1RyZWUkTm9kZRsR7BbVWpYAAgAGWgADcmVkTAADa2V5dAASTGphdmEvbGFuZy9PYmplY3Q7TAAEbGVmdHEAfgAKTAAGcGFyZW50cQB+AApMAAVyaWdodHEAfgAKTAAFdmFsdWVxAH4ADXhwAHNyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAABcHBzcQB+AAwBc3EAfgAPAAAAA3BxAH4ADnBwcHNxAH4AA3EAfgAIcQB+AAtzcgAKc2NhbGEuU29tZREi8mleoYt0AgABTAAFdmFsdWVxAH4ADXhyAAxzY2FsYS5PcHRpb27+aTf92w5mdAIAAHhwcQB+ABFzcQB+ABVzcQB+AA8AAAAC")
+  check(mutable.TreeMap[Int, Int]())( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZU1hcNx8qC229ZvwAgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZU1hcCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQk8u7cgruVzucCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAABw")
+  check(mutable.TreeMap(1 -> 1, 3 -> 6))( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZU1hcNx8qC229ZvwAgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZU1hcCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQk8u7cgruVzucCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAAJzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSROb2RlGxHsFtValgACAAZaAANyZWRMAANrZXl0ABJMamF2YS9sYW5nL09iamVjdDtMAARsZWZ0cQB+AAdMAAZwYXJlbnRxAH4AB0wABXJpZ2h0cQB+AAdMAAV2YWx1ZXEAfgAKeHAAc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAAFwcHNxAH4ACQFzcQB+AAwAAAADcHEAfgALcHNxAH4ADAAAAAZxAH4ADg==")
+  check(mutable.TreeMap(1 -> 1, 3 -> 6).range(1, 2))( "rO0ABXNyACxzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZU1hcCRUcmVlTWFwVmlldx7MCZxLhVQ8AgADTAAGJG91dGVydAAiTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9UcmVlTWFwO0wABGZyb210AA5Mc2NhbGEvT3B0aW9uO0wABXVudGlscQB+AAJ4cgAgc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlRyZWVNYXDcfKgttvWb8AIAAkwACG9yZGVyaW5ndAAVTHNjYWxhL21hdGgvT3JkZXJpbmc7TAAmc2NhbGEkY29sbGVjdGlvbiRtdXRhYmxlJFRyZWVNYXAkJHRyZWV0ACxMc2NhbGEvY29sbGVjdGlvbi9tdXRhYmxlL1JlZEJsYWNrVHJlZSRUcmVlO3hwc3IAGHNjYWxhLm1hdGguT3JkZXJpbmckSW50JPLu3IK7lc7nAgAAeHBzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSRUcmVlAEynNPA1phUCAAJJAARzaXplTAAEcm9vdHQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJE5vZGU7eHAAAAACc3IAKnNjYWxhLmNvbGxlY3Rpb24ubXV0YWJsZS5SZWRCbGFja1RyZWUkTm9kZRsR7BbVWpYAAgAGWgADcmVkTAADa2V5dAASTGphdmEvbGFuZy9PYmplY3Q7TAAEbGVmdHEAfgAKTAAGcGFyZW50cQB+AApMAAVyaWdodHEAfgAKTAAFdmFsdWVxAH4ADXhwAHNyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAABcHBzcQB+AAwBc3EAfgAPAAAAA3BxAH4ADnBzcQB+AA8AAAAGcQB+ABFzcQB+AANxAH4ACHEAfgALc3IACnNjYWxhLlNvbWURIvJpXqGLdAIAAUwABXZhbHVlcQB+AA14cgAMc2NhbGEuT3B0aW9u/mk3/dsOZnQCAAB4cHEAfgARc3EAfgAWc3EAfgAPAAAAAg==")
+  check(mutable.TreeSet[Int]())( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZVNldM10nxFQDpt4AgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZVNldCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQk8u7cgruVzucCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAABw")
+  check(mutable.TreeSet(1, 3))( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZVNldM10nxFQDpt4AgACTAAIb3JkZXJpbmd0ABVMc2NhbGEvbWF0aC9PcmRlcmluZztMACZzY2FsYSRjb2xsZWN0aW9uJG11dGFibGUkVHJlZVNldCQkdHJlZXQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJFRyZWU7eHBzcgAYc2NhbGEubWF0aC5PcmRlcmluZyRJbnQk8u7cgruVzucCAAB4cHNyACpzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuUmVkQmxhY2tUcmVlJFRyZWUATKc08DWmFQIAAkkABHNpemVMAARyb290dAAsTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9SZWRCbGFja1RyZWUkTm9kZTt4cAAAAAJzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSROb2RlGxHsFtValgACAAZaAANyZWRMAANrZXl0ABJMamF2YS9sYW5nL09iamVjdDtMAARsZWZ0cQB+AAdMAAZwYXJlbnRxAH4AB0wABXJpZ2h0cQB+AAdMAAV2YWx1ZXEAfgAKeHAAc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAAFwcHNxAH4ACQFzcQB+AAwAAAADcHEAfgALcHBw")
+  check(mutable.TreeSet(1, 3).range(1, 2))( "rO0ABXNyACxzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuVHJlZVNldCRUcmVlU2V0Vmlld2JdAzqy0DpGAgADTAAGJG91dGVydAAiTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9UcmVlU2V0O0wABGZyb210AA5Mc2NhbGEvT3B0aW9uO0wABXVudGlscQB+AAJ4cgAgc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlRyZWVTZXTNdJ8RUA6beAIAAkwACG9yZGVyaW5ndAAVTHNjYWxhL21hdGgvT3JkZXJpbmc7TAAmc2NhbGEkY29sbGVjdGlvbiRtdXRhYmxlJFRyZWVTZXQkJHRyZWV0ACxMc2NhbGEvY29sbGVjdGlvbi9tdXRhYmxlL1JlZEJsYWNrVHJlZSRUcmVlO3hwc3IAGHNjYWxhLm1hdGguT3JkZXJpbmckSW50JPLu3IK7lc7nAgAAeHBzcgAqc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLlJlZEJsYWNrVHJlZSRUcmVlAEynNPA1phUCAAJJAARzaXplTAAEcm9vdHQALExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUmVkQmxhY2tUcmVlJE5vZGU7eHAAAAACc3IAKnNjYWxhLmNvbGxlY3Rpb24ubXV0YWJsZS5SZWRCbGFja1RyZWUkTm9kZRsR7BbVWpYAAgAGWgADcmVkTAADa2V5dAASTGphdmEvbGFuZy9PYmplY3Q7TAAEbGVmdHEAfgAKTAAGcGFyZW50cQB+AApMAAVyaWdodHEAfgAKTAAFdmFsdWVxAH4ADXhwAHNyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAABcHBzcQB+AAwBc3EAfgAPAAAAA3BxAH4ADnBwcHNxAH4AA3EAfgAIcQB+AAtzcgAKc2NhbGEuU29tZREi8mleoYt0AgABTAAFdmFsdWVxAH4ADXhyAAxzY2FsYS5PcHRpb27+aTf92w5mdAIAAHhwcQB+ABFzcQB+ABVzcQB+AA8AAAAC")
   // TODO SI-8576 Uninitialized field under -Xcheckinit
   // check(new mutable.History())( "rO0ABXNyACBzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuSGlzdG9yeUhuXxDIFJrsAgACSQAKbWF4SGlzdG9yeUwAA2xvZ3QAIExzY2FsYS9jb2xsZWN0aW9uL211dGFibGUvUXVldWU7eHAAAAPoc3IAHnNjYWxhLmNvbGxlY3Rpb24ubXV0YWJsZS5RdWV1ZbjMURVfOuHHAgAAeHIAJHNjYWxhLmNvbGxlY3Rpb24ubXV0YWJsZS5NdXRhYmxlTGlzdFJpnjJ+gFbAAgADSQADbGVuTAAGZmlyc3QwdAAlTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9MaW5rZWRMaXN0O0wABWxhc3QwcQB+AAV4cAAAAABzcgAjc2NhbGEuY29sbGVjdGlvbi5tdXRhYmxlLkxpbmtlZExpc3Sak+nGCZHaUQIAAkwABGVsZW10ABJMamF2YS9sYW5nL09iamVjdDtMAARuZXh0dAAeTHNjYWxhL2NvbGxlY3Rpb24vbXV0YWJsZS9TZXE7eHBwcQB+AApxAH4ACg==")
   check(mutable.LinkedHashMap(1 -> 2))( "rO0ABXNyACZzY2FsYS5jb2xsZWN0aW9uLm11dGFibGUuTGlua2VkSGFzaE1hcAAAAAAAAAABAwAAeHB3DQAAAu4AAAABAAAABABzcgARamF2YS5sYW5nLkludGVnZXIS4qCk94GHOAIAAUkABXZhbHVleHIAEGphdmEubGFuZy5OdW1iZXKGrJUdC5TgiwIAAHhwAAAAAXNxAH4AAgAAAAJ4")
```

lrytz:
Green now, i guess you just re-generated the test?

lrytz:
Wait, you had to update the test in the "Default -Xmixin-force-forwarders to true" commit, which means that there's still some classes being tested by t8549 that change SVUID under `-Xmixin-force-forwarders:ture`. Instead, these classes should get the annotation.

lrytz:
So this time it would be `Ordering$Int$` that should get the annotation it seems. Strangely I couldn't reproduce the failure locally (i did a local bootstrap). So yeah, not sure if we should go through more of the std lib and add the annotations everywhere..

This is a reminder that changing `-Xmixin-force-forwarders`, even though binary compatible, is not serialization compatible.

adriaanm:
> This is a reminder that changing -Xmixin-force-forwarders, even though binary compatible, is not serialization compatible.

Yep :-/ Do we still commit to having it on, knowing we can't turn it off?

lrytz:
I'm still in favor of the change. Do we have any guarantees in serialization compatibility? We have t8549, but it covers only a slice of the serializable classes in the sdt lib. For users, anyone relying on serialization compatibility across binary versions _should_ be using `@SVUID` anyway.

adriaanm:
We don't explicitly promise it because we can't mechanically check, but we
have tended to shy away from making changes that could impact
serializability.

If the performance benefit is significant, I'm on board but I don't think
we can revert it later knowing the impact on serialization.

lrytz:
I'm going a bit further into the off-topic..

Maybe we should have an `-Xlint` warning for subclasses (abstract and concrete, not traits) of `Serializable` that don't have the `@SVUID` annotation?

The problem with explicit `@SVUID` is that it not only allows benign changes such as adding a public method to a class, but also lets problematic changes pass. For example:

```
@SerialVersionUID(1L)
class ArrayBuffer extends Serializable {
  private var elems = new Array[AnyRef](8)
  private def ensureCapacity(n: Int): Unit = {
    // var newSize = elems.length * 2
    // while (n >= newSize) newSize *= 2
  }
  def clear(): Unit = { elems = new Array[AnyRef](8) }
}
```

Now assume there's a refactoring:

```
@SerialVersionUID(1L)
class ArrayBuffer(initSize: Int) extends Serializable {
  assert(initSize > 1)
  def this() = this(8)
  private var elems = new Array[AnyRef](initSize)
  private def ensureCapacity(n: Int): Unit = {
    // ...
    // var newSize = elems.length * 2
    // while (n >= newSize) newSize *= 2
    // 
  }
  def clear(): Unit = { elems = new Array[AnyRef](initSize) }
}
```

Now if you de-serialize a stream from the first version into the second version, the `initSize` field will be `0`. After a call to `clear()`, `ensureCapacity` will loop forever.

This is of course not a Scala-specific problem, but we don't have a way of tracking actual breaking changes in the std lib.

adriaanm:
Let's spin the serialization topic of to its own scala-dev issue. Maybe we could consider assigning a SVUID that's 1:1 with the Scala binary version for all classes in the stdlib? Together with MiMa this would ensure serialization "compatibility" (somewhat trivially).

lrytz:
Sounds good to me (discussing scala-dev). I think this PR could be merged, though it has a bit of an ad-hoc mix now for the `@SVUID` question: for some classes the annotation is added to make `t8549` pass (the `ClassTag` subclasses), for others we adjusted the serialized value in the test itself (`Ordering$Int$`). But that's no different than what we've been doing so far.

@retronym could you take another look at the PR and the various points we discussed here? Could you also run your benchmarks before/after this commit? We can also merge it beforehand if that makes benchmarking easier.

I've been talking to Adriaan today, our (current) conclusion is to merge this PR if the performance benefit is significant, knowing we cannot change it during the 2.12.x cycle because of serialization compatibility.

retronym:
`t8549` is designed to check that we don't break serialization within minor releases.

I think we should use the serizaliazation proxy pattern more pervasively in the collections, rather than directly serializing the internal representation. But this would also expose us more to the seemingly intractable problem of https://issues.scala-lang.org/browse/SI-9237. sigh..

retronym:
I'll benchmark again. I feel a bit uncomfortable going ahead with this change based purely on the result of the cold benchmark improvement without actually having understood _why_ it makes the difference. It might not be possible to produce such an answer in time, but we can try some variations (e.g. like turning this on for library- or compiler-only).

adriaanm:
Let's prioritize finding the best answer to the performance question within
the time available (by end of week).

lrytz:
First numbers

Cold performance of compiling better-files is 12% faster with forwarders
  - with forwarders: 7770.704 ± 182.328 ms/op
  - without forwarders: 8741.837 ± 150.133 ms/op

Unfortunately, hot performance is 8% worse with forwarders (also compiling better-files)
  - with forwarders: 870.223 ± 15.266
  - without forwarders: 803.554 ± 6.324

I'll continue benchmarking later, but I think it would be good to have it done on multiple machines. I'll also try my slow linux box at home.

lrytz:
ok, scratch that, i was not using the same compiler options; i need to run both with `-Xmixin-force-forwarders:junit`; i'll give it another go :)

lrytz:
OK, so **hot** performance seems to be the same. (n = no forwarders, y = with forwarders, unit = ms/op).

better-files
- n: 821.011 ± 9.608 
- y: 818.349 ± 3.649

hello (a very simple class)
- n: 45.312 ± 0.196
- y: 44.975 ± 0.324

For **cold** performance, having the forwarders helps. The longer the compiler runs, the smaller the difference. For a minimal class, the difference is 41%.

scalap (2117 loc): 9%
- n: 12240.296 ± 177.272
- y: 11223.632 ± 313.381

better-files (924 loc): 12%
- no forwarders: 8742.618 ± 142.837
- with forwarders: 7757.120 ± 138.085

vector (935 loc): 17%
- n: 7223.134 ± 191.534
- y: 6145.266 ± 174.945

hello (3 loc): 41%
- n: 3969.111 ± 24.397
- y: 2804.122 ± 29.709

Wild speculation: the relation between slowdown and running time _could_ confirm the hypothesis "due to no CHA for default methods". CHA is used for inlining decisions in C1, so it has a big impact in the beginning. C2 uses type profiling information for inlining decisions, so the lack of CHA does not matter anymore.

lrytz:
will do more benchmarks, but a bit later :) ideas (mostly by jason):
- effect of having forwarders only in the compiler, only in the library
- check the effect of `-XX:-UseCHA` in 2.11.8, and in 2.12.x with / without forwarders
- check the effect of `-XX:TieredStopAtLevel=1` (disable C2) for various compilers
- compare the compiler phase timings
- compare `-XX:+PrintCompilation` logs

retronym:
I'm also interested in the cold performance using JDK9-ea, rather than JDK8.

For 2.11.8, I've measured an improvement of 1.6s -> 1.4s from 1.8_102 to 9-ea-128.

I haven't quite been able to measure on 2.12.0-SNAPSHOT or this PR yet as I discovered another bug in the new optimizer (https://github.com/scala/scala-dev/issues/242).

retronym:
> I'm also interested in the cold performance using JDK9-ea, rather than JDK8.

I didn't see a difference in the cold performance between these JDK version in 2.12.0-SNAPSHOT.

Following on in the direction of your  "compile hello world" benchmark, I tried benchmarking `scalac -version`. This takes 0.2s in 2.11.8, and 1s (!!!) on 2.12.0-SNAPSHOT.

Zooming in with the XCode profiler:

```
% instruments -t "Time Profiler" java ...
```

I don't see an obvious hotspot, but I do see an old enemy, `generate_default_methods` spread through the profile:

image

I remember this one from the time I tried adding _all_ ancestors redundantly as parents of all classes, which severely degraded startup time. IIRC, the analysis doesn't maintain a set of already processed ancestors, so it repeats work when an interface is inherited through multiple routes. Perhaps we have enough instances of this in collections and/or the reflect/compiler to be a problem.

retronym:
Example of our "strongly woven" family tree:

``` scala
scala> def interfaceCount(cls: Class[_]) = {
  val m = collection.mutable.Map[Class[_], Int]().withDefaultValue(0)
  def loop(cls1: Class[_]) 
    cls1.getInterfaces.foreach(i => m(i) = (m(i) + 1))
    cls1.getInterfaces.foreach(loop)
    if (cls1.getSuperclass != null) loop(cls1.getSuperclass)
  }
  loop(cls)
  m
}
interfaceCount: (cls: Class[_])scala.collection.mutable.Map[Class[_],Int]

scala> println(interfaceCount(Nil.getClass).toSeq.sortBy(-_._2).mkString("\n"))
(interface scala.collection.GenTraversableOnce,75)
(interface scala.collection.Parallelizable,58)
(interface scala.collection.GenTraversableLike,58)
(interface scala.collection.generic.HasNewBuilder,32)
(interface scala.collection.GenIterableLike,26)
(interface scala.Equals,19)
(interface scala.collection.TraversableOnce,17)
(interface scala.collection.generic.FilterMonadic,17)
(interface scala.collection.TraversableLike,17)
(interface scala.collection.generic.GenericTraversableTemplate,15)
(interface scala.collection.GenTraversable,15)
(interface scala.collection.IterableLike,10)
(interface scala.collection.GenSeqLike,8)
(interface scala.collection.GenIterable,8)
(interface scala.collection.Traversable,7)
(interface scala.collection.SeqLike,5)
(interface scala.collection.Iterable,5)
(interface scala.collection.Seq,3)
(interface scala.PartialFunction,3)
(interface scala.Function1,3)
(interface scala.collection.GenSeq,3)
(interface java.io.Serializable,2)
(interface scala.collection.LinearSeqLike,2)
(interface scala.collection.immutable.Traversable,1)
(interface scala.collection.immutable.LinearSeq,1)
(interface scala.collection.immutable.Seq,1)
(interface scala.collection.LinearSeqOptimized,1)
(interface scala.collection.immutable.Iterable,1)
(interface scala.Immutable,1)
(interface scala.Serializable,1)
(interface scala.Product,1)
(interface scala.collection.LinearSeq,1)
```

HotSpot code involved:

http://hg.openjdk.java.net/jdk9/jdk9/hotspot/file/tip/src/share/vm/classfile/defaultMethods.cpp

retronym:
This recently fixed JDK bug is in the same ballpark, but appears distinct: https://bugs.openjdk.java.net/browse/JDK-8163969

lrytz:
It's an interesting thing to keep in mind, but I guess there's nothing we can do about it for RC2 - the presence / absence of mixin forwarders does not have an impact here, right?

I'll do some more benchmarks for the effects related to mixin forwarders when I get back.

lrytz:
When using `-XX:TieredStopAtLevel=1`, i.e., disabling C2, therefore relying more on CHA, there is a difference in **hot** performance (with C2 enabled, hot performance was the same):

better-files
- n: 1834.449 ± 8.637
- y: 1893.169 ± 12.478

hello
- n: 101.173 ± 0.382
- y: 96.638 ± 0.361

retronym:
I've lodged a bug for the HotSpot problem with default method resolution performance, I'll link it here when it appears in the tracker.

Here's an attempt to distill the problem in a Java-only test case: https://gist.github.com/retronym/ddedbadc3c764da66ffdb9a911915b40

lrytz:
I did a bit of benchmarking with the debug JVM that I built on linux, with the goal of comparing the impact of the `-XX:-UseCHA` flag.

But it looks like a bad idea to use such a VM for benchmarking: it is extremely slow. For example, compiling better-files takes 11s with Oracle 1.8.0_102, but 47s with the debug vm i built (jdk8u branch).

Anyway, here are some numbers I collected on the debug VM, using 2.11.8.

hello, hot performancewith CHA
  -

lrytz:
I did a bit of benchmarking with the debug JVM that I built on linux, with the goal of comparing the impact of the `-XX:-UseCHA` flag.

But it looks like a bad idea to use such a VM for benchmarking: it is extremely slow (plenty of debug code all over). For example, compiling better-files takes 11s with Oracle 1.8.0_102, but 47s with the debug VM I built (jdk8u branch).

Anyway, here are some numbers I collected on the debug VM, using 2.11.8.

hello, hot performance
- default: 160.882 ± 5.780
- -UseCHA: 158.800 ± 5.996

hello, cold performance
- default: 9072.075 ± 354.355
- -UseCHA: 6988.007 ± 212.776

better-files, hot performance
- default: 1578.248 ± 140.107
- -UseCHA: 1874.994 ± 47.585

better-files, cold performance
- default: 32128.875 ± 653.279
- -UseCHA: ~ 55000

Not sure if we can draw any conclusions here.. CHA slows down, maybe due to additional debug logging?

lrytz:
Interestingly, the difference cold performance is caused mostly by forwarders in scala-library.jar.

hello, cold performance
- no forwarders: 3973.261 ± 23.066
- forwarders only in compiler + reflect: 3926.395 ± 34.604
- forwarders only in library: 2861.180 ± 22.530
- forwarders in all jars: 2808.490 ± 31.637

lrytz:
One more number: when disabling JIT entirely (`-Xint`), the forwarders still improve performance. For compiling hello, cold performance, it's 15%:
- no forwarders: 7031.654 ± 60.182
- with forwarders: 6099.804 ± 24.948

If we assume that CHA is only used in the optimizers (not sure if that is true), this shows that there is some other factor at play.

Maybe this is related to method resolution, but this is just a guess.

```
trait T { def f = 1 }
class A extends T
class C { def g(a: A) = a.f }
```

In C we get an `invokevirtual A.f` (with or without the forwarder). If there's a forwarder, resolution should be quick (it's right there in A). The forwarder itself does:

```
class A implements T {
  public f()I
    ALOAD 0
    INVOKESTATIC T.f$ (LT;)I
    IRETURN
}
```

The lookup of the static method should be quick, too. The static accessor in T does:

```
interface T {
  public static synthetic f$(LT;)I
    ALOAD 0
    INVOKESPECIAL T.f ()I
    IRETURN
}
```

The lookup of `INVOKESPECIAL T.f` should also be quick, it's in the named class.

If `A` doesn't have a forwarder, resolving `A.f` is more costly, especially if `A` extends a large number of interfaces and `f` is defined somewhere high up (think: collections). We should benchmark a bit more in this area at some point.

Note that the interpreter needs to go through two forwarders, but this is still faster than the version where the default method is directly executed.

lrytz:
I think we should take a decision based on the numbers we have now. (Suggestions for other benchmarks are welcome, if they can be done quickly. I guess we want to have RC2 soon.)

To summarize:
- Startup is slower without forwarders
- For a large program with a short running time (compiling a small file) the difference can be significant (40%)
- Hot performance is not affected
- There are probalby multiple underlying causes (CHA, populating vtables), we don't have a clear picture
- We cannot revert in 2.12.x due to serialization compatibility

adriaanm:
Since emitting the forwarders doesn't adversely affect hot performance, while substantially improving startup on current JVMs, I agree we should merge this PR. Our concern remains that this is an area where we are exposed to intricacies of the JIT and classloading, which could change in future updates, while we are bound to the choice we make now (because flipping the switch affects serialization compatibility).

retronym:
LGTM

retronym:
I've spun out the issue with startup performance as https://github.com/scala/scala-dev/issues/243

ramkishanbetta:
> The SerialVersionUID test failed for `Set$EmptySet$` now. I patched it up, but the way we handle this annotation in the std lib is really ad-hoc..

Well, how do one is expected to handle this as part of upgrade? Say, I serialized using scala 2.11 and saved, and deserialize in scala 2.12 post upgrade?

PR #5428, opened 2016-09-28
Emit local module like lazy val

retronym:
The motivation is to use the new fine-grained lock scoping that
local lazies have since #5294.

Fixes https://github.com/scala/scala-dev/issues/235

Co-Authored-By: Jason Zaugg jzaugg@gmail.com

Review by @adriaanm

TODO:
- [ ] the outer pointer for the object after this PR is `null`:
        `(C$O$2$)O$lzy$1.initialize((Object)new C$O$2$(null));` versus `x$1.elem = new C$O$2$(this);`
- [ ] the name for the local variable changed `O$module` -> `O$lzy`

PR #5430, opened 2016-09-28
Emit local module like lazy val

adriaanm:
The motivation is to use the new fine-grained lock scoping that
local lazies have since #5294.

Fixes scala/scala-dev#235

Co-Authored-By: Jason Zaugg jzaugg@gmail.com

Review by @adriaanm

TODO:
- [X] the outer pointer for the object after this PR is `null`:
        `(C$O$2$)O$lzy$1.initialize((Object)new C$O$2$(null));` versus `x$1.elem = new C$O$2$(this);`
- [X] the name for the local variable changed `O$module` -> `O$lzy` 

Follow up for #5428


adriaanm:
I couldn't help but make some constants final in  d04cda1 when I was fixing the name of the val in the expansion. @retronym, @lrytz: any reservations about this?

The outer pointer being null is progress, actually, since delambdafy correctly elides it. I added a test case where it isn't elided because it's used.

adriaanm:
/rebuild

Hrm. Is the ec2 instance type not the issue here?

SethTisue:
```
# starting 63 tests in run
!!  1 - run/SI-4676.scala                         [compilation failed]
!!  2 - run/SI-4360.scala                         [compilation failed]
!!  3 - run/SI-4887.scala                         [compilation failed]
##### Log file '/home/jenkins/workspace/scala-2.12.0-validate-test/test/scaladoc/run/SI-4360-run.log' from failed test #####

error: Java heap space

##### Log file '/home/jenkins/workspace/scala-2.12.0-validate-test/test/scaladoc/run/SI-4887-run.log' from failed test #####

error: Java heap space

##### Log file '/home/jenkins/workspace/scala-2.12.0-validate-test/test/scaladoc/run/SI-4676-run.log' from failed test #####

error: Java heap space
```

adriaanm:
Right, so that error occurred on the behemoth that had already been upgraded to 4xlarge...

adriaanm:
How come the max heap space is 1024m in that log? 

> Java options are:    -Xmx1024M -Xms64M -XX:MaxPermSize=128M

I thought we had it at 2g?

SethTisue:
`git grep Xmx` shows all sorts of different numbers in different places :-\

SethTisue:
even more numbers if you do the same `git grep` in the scala-jenkins-infra, but in a partest context, I think it's the scala/scala numbers that matters since it's forked JVMs

adriaanm:
`build.sbt:635:    testOptions in IntegrationTest += Tests.Argument("-Dpartest.java_opts=-Xmx1024M -Xms64M -XX:MaxPermSize=128M"),`

retronym:
consting those strings seems okay to me.

build.sbt
```diff
@@ -632,7 +632,7 @@ lazy val test = project
     javaOptions in IntegrationTest += "-Xmx2G",
     testFrameworks += new TestFramework("scala.tools.partest.sbt.Framework"),
     testFrameworks -= new TestFramework("org.scalacheck.ScalaCheckFramework"),
-    testOptions in IntegrationTest += Tests.Argument("-Dpartest.java_opts=-Xmx1024M -Xms64M -XX:MaxPermSize=128M"),
+    testOptions in IntegrationTest += Tests.Argument(s"-Dpartest.java_opts=${(javaOptions in IntegrationTest).value.mkString(" ")}"),
```
retronym: Using a small initial heap for all the forked partest processes seems desirable to me, this change drops that:

```
show test/it:javaOptions
[info] List(-Xmx2G)
```

retronym: If we actually have partests that require large heaps, we should probably segregated them into a new category that is run sequentially. That would let us run the vast majority of them with a small heap.

adriaanm: I like the idea of a large heap section.

Regarding the initial heap size, after this PR it would be 472M according to `java -Xmx2g -XX:+PrintFlagsFinal`. Should I move the `-Xms64m` setting to `javaOptions in IntegrationTest += "-Xmx2G",`?

retronym: Yep, the JVM that forks for the partest framework itself can probably get away with a small heap, too. I have a feeling that we bumped up the flags of that JVM to try to get around intermittent failures that were actually fixed later on this line.

Happy to do whatever gets things working again for 2.12.0 and leave the cleanup work until 2.12.x; I've noted my ideas in https://github.com/scala/scala-jenkins-infra/issues/181#issuecomment-250353894

retronym:
Still failing. Very hard from the output to see which JVM is actually heap bound.

Should we add `-XX:+CrashOnOutOfMemoryError` to all our JVMs to fail fast with a crash log to help diagnosis?

adriaanm:
Added it to the tests, for a start. No idea what's going on here. This is running on an EC2 c4.4xlarge instance, which has 30G ram.

adriaanm:
We'll have to wait until we can upgrade to a b92 or later jdk for the crash on out of memory option.

src/compiler/scala/tools/nsc/transform/Fields.scala
```diff
@@ -541,10 +536,13 @@ abstract class Fields extends InfoTransform with ast.TreeDSL with TypingTransfor
       *  }
       * def x(): Int = if (x$lzy.initialized()) x$lzy.value() else x$lzycompute()
       * ```
+      *
+      * The expansion is the same for local lazy vals and local objects,
+      * except for the name of the val ($lzy or
```
lrytz: or? `$module`?

lrytz (commented):
LGTM, also the name constants.

I don't understand what you're referring to by "outer pointer being null is progress / added a test case"

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm: darn, yes, or `$module` :-)

adriaanm:
> "outer pointer being null is progress / added a test case"

I was surprised by the diff in generated code for a local test I was doing, but forgot you can't see my screen, so yeah, that was a mysterious comment 😎  The test looked something like `class C { def foo = object O }`.

In that test, the outer pointer passed to the local module's class's constructor is now `null`, instead of `C.this`. It turns out delambdafy now correctly detects it's not being used.

src/compiler/scala/tools/nsc/transform/Fields.scala
adriaanm: fixed

adriaanm:
I dropped all the CI-related stuff (moved discussion to https://github.com/scala/scala-dev/issues/236) and fixed the doc

PR #5433, opened 2016-09-29
don't deprecate Either.left and Either.right yet

SethTisue:
as per https://github.com/scala/scala/pull/5135

review by @szeiger, @Ichoran, @soc, community.  (this will likely be
merged rapidly, but that doesn't mean additional changes, e.g. to the
Scaladoc, can't follow)

### DIFF
--- a/src/library/scala/util/Either.scala
+++ b/src/library/scala/util/Either.scala
@@ -95,13 +95,15 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
   /**
    * Projects this `Either` as a `Left`.
    */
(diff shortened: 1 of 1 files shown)


szeiger (approved):
I don't agree with the premise that source compatibility between two releases implies being able to build without warnings on both, but I'm in no hurry to deprecate these methods.

SethTisue:
> I don't agree with the premise that source compatibility between two releases implies being able to build without warnings on both

noted. in the general case, I agree.

Ichoran:
LGTM

adriaanm:
I'm going to ignore the spurious test failures and merge this.

PR #5434, opened 2016-09-29
Limit partest concurrency to 8

retronym:
In an attempt to restore stability to CI builds after
the move to a bigger box with 16 processors.


lrytz:
Any idea why the commits here are not being tested by scaboo?

adriaanm:
I've reverted the bump to the bigger EC2 instance type. Let's deal with this post-2.12.0. See scala/scala-dev#236

PR #5435, opened 2016-09-29
Bump sbt-jmh version to 0.2.16

monkey-mas:
It'd be good to use the latest version( with the latest JMH version `1.14.1`). From sbt-jmh version `0.2.10`, Flight Recorder / Java Mission Control is available[1], which would be nice.

[1] https://github.com/ktoso/sbt-jmh#using-oracle-flight-recorder


SethTisue:
review by @performantdata ?

performantdata:
LGTM, as long as it compiles and runs, which I'd expect with a minor JMH release. You'll need to test that manually; the project tests don't check that automatically.

lrytz:
I tested it by running `jmh:runMain scala.collection.mutable.OpenHashMapRunner` (which seems to be running for long time)

SethTisue:
thanks @monkey-mas!

monkey-mas:
Thanks alot for your review! :)

PR #5436, opened 2016-09-29
Explicit set diff

m2rtin:
When dealing with sets, I run into troubles. Suppose we want to do operations on two sets **S1**, **S2**. Why is **&** implemented as **S1 filter S2** but **&~** is evidently implemented as **r=S1; S2.foreach(x => r = r - x); r**
The **-** operation creates a new instance of **S1**-like large set each time element is removed. 

Take into consideration the following:
We want to do **S1** diff **S2**.

(1)  |S1| = n, |S2| = M, |S1 intersect S2| = k, **k <= n << M**, 

The current version has to go through M values and creates k-times new Instance of a collection.
Whereas the proposed version has to do just n times a query to S2 and it creates just one final instance of Set (filter uses SetBuilder).

(2)  |S1| = N, |S2| = m, **N >> m >= k**

Although the current version goes through just m values, it creates k-times a massive collection.
The proposed version goes through a lot of (N) values, however it does create an output of just k items.

Complexity summary:
(1) - current - memory O(n \* k), **time O(M)**
(1) - proposed - memory k, time n*C, where C is time of evaluating whether element is in a large set

(2) - current - **memory O(N \* k)**, time O(m)
(2) - proposed - memory k, time M \* c, where c is time of evaluating whether element is in a small set

...Besides who would have thought that & and &~ behave so differently?
If **&** can do it by filtering so can **&~**

### DIFF
--- a/src/library/scala/collection/GenSetLike.scala
+++ b/src/library/scala/collection/GenSetLike.scala
@@ -83,7 +83,7 @@ extends GenIterableLike[A, Repr]
    *  @return     a set containing those elements of this
    *              set that are not also contained in the given set `that`.
    */
-  def diff(that: GenSet[A]): Repr
+  def diff(that: GenSet[A]): Repr = this filterNot that
 
   /** The difference of this set and another set.
    *
--- a/src/library/scala/collection/SetLike.scala
+++ b/src/library/scala/collection/SetLike.scala
@@ -171,14 +171,6 @@ self =>
    */
   def union(that: GenSet[A]): This = this ++ that
 
-  /** Computes the difference of this set and another set.
-   *
-   *  @param that the set of elements to exclude.
-   *  @return     a set containing those elements of this
-   *              set that are not also contained in the given set `that`.
-   */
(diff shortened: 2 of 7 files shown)


szeiger:
Can you add a [benchmark](https://github.com/scala/scala/blob/2.12.x/test/benchmarks/README.md) to show the improvements?

What are the implications for parallel collections?

strelec:
What if we checked which collection is bigger and reverse the operands as needed? Provided .size is efficient, of course.

m2rtin:
I have not noticed that this version of scala has already improved on the performance of this operation, and it is no longer implemented as --. 
I added micro benchmark declaring this. I used sbt to run the benchmark by the following command:
"jmh:run -i 3 -wi 3 -f1 -t1 .*SetBenchmark"

SethTisue:
confused about where we stand on this — you're now suggesting we _only_ add a benchmark? (but the PR still includes code changes?)

m2rtin:
The PR is not valid, i discovered that the bug is not present in this version

PR #5439, opened 2016-09-30
Shorten fields phase description

som-snytt:
Makes fields fit the field width, which is fitting.

`s/including/add` seems sufficient.

Possibly, "synthesize" is an extravagance for "add",
but "add" is used previously in that column.

Resolve, load, translate, add, synthesize, replace, erase,
move, eliminate, remove, generate.

Would love to learn a word that says what typer does, if
the word "type" is too redundant or overloaded, besides the
food metaphor. Also "meat-and-potatoes" implies basic,
simple, not fussy or fancy. There are many devices,
like the heart or a Ferrari engine, that are
fundamental without being unfussy.

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -676,7 +676,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
   }
 
   /** A description of the phases that will run in this configuration, or all if -Ydebug. */
-  def phaseDescriptions: String = phaseHelp("description", elliptically = true, phasesDescMap)
+  def phaseDescriptions: String = phaseHelp("description", elliptically = !settings.debug, phasesDescMap)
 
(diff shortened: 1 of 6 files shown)


som-snytt:
@SethTisue This may qualify as a doc-level fix for 2.12.0.

som-snytt:
@adriaanm Review by. Man, I know you like the long commit comments. You the novelist. But -Xshow-phases just wants to display in 80 chars. Think of the newbie who opens a terminal and does that. Do they want to see an ellipsis? That's all I'm saying. And I got a delta so that -Ydebug will show the whole string, in case you got a plugin with a short story baked in.

adriaanm:
This was going to my debut as a novelist, but fine! I stand redacted.

som-snytt:
/rebuild

Jenquins is so Euro! `erreicht: 50221`

som-snytt:
`scalac -Xshow-phases -Ydebug` used to show phases that weren't enabled (continuations, optimizations), but they aren't any now.

SethTisue:
thanks So...

PR #5440, opened 2016-10-02
SI-9944 Scan after interp expr keeps CR

som-snytt:
In an interpolated expression `s"""${ e }"""`, the scanner
advances input past the RBRACE. If a multiline string as
shown, get the next raw char, because CR is significant.


lrytz:
LGTM!

PR #5442, opened 2016-10-04
SI-9943 sealed class does not yield SAM type

adriaanm:
Cannot subclass such a class. (Well, we could subclass a sealed
class in the same compilation unit. We ignore this for simplicity.)

This is a bit of a sneaky fix for this bug, but our hand
is pretty much forced by other constraints, in this intersection
of overload resolution involving built-in function types and SAMs,
and type inference for higher-order function literals (#5307).

Luckily, in this particular issue, the overloading clash seems
accidental. The `sealed` `<:<` class is not a SAM type as it
cannot be subclassed outside of `Predef`. For simplicity,
we don't consider where the SAM conversion occurs and exclude
all sealed classes from yielding SAM types.

Thanks to Miles for pointing out that `final` was missing in my
first iteration of this fix.

JIRA: https://issues.scala-lang.org/browse/SI-9943

### DIFF
--- a/spec/06-expressions.md
+++ b/spec/06-expressions.md
@@ -1361,6 +1361,7 @@ Note that a function literal that targets a SAM is not necessarily compiled to t
 
 It follows that:
   - if class `C` defines a constructor, it must be accessible and must define exactly one, empty, argument list;
+  - class `C` cannot be `final` or `sealed` (for simplicity we ignore the possibility of SAM conversion in the same compilation unit as the sealed class);
   - `m` cannot be polymorphic;
   - it must be possible to derive a fully-defined type `U` from `S` by inferring any unknown type parameters of `C`.
 
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -840,14 +840,14 @@ trait Definitions extends api.StandardDefinitions {
      *
      * The method must be monomorphic and have exactly one parameter list.
      * The class defining the method is a supertype of `tp` that
-     * has a public no-arg primary constructor.
+     * has a public no-arg primary constructor and it can be subclassed (not final or sealed).
      */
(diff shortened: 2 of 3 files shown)


milessabin:
Can a type marked final (rather than sealed) be considered as the target for a SAM conversion?

adriaanm:
Good point! Pushing a fix.

adriaanm:
@retronym, what say you?

adriaanm:
FTR, we basically do the same as dotty (treating Scala functions the same as SAMs when resolving overloading), except that we don't require the higher-order function's argument types to be identical. Compiling our pos/overloaded_ho_fun.scala test case with dotc produces two errors:

```

trait Sam { def apply(x: Int): String }
trait SamP[U] { def apply(x: Int): U }

class OverloadedFun[T](x: T) {
  def foo(f: T => String): String = f(x)
  def foo(f: Any => T): T = f("a")

  def poly[U](f: Int => String): String = f(1)
  def poly[U](f: Int => U): U = f(1)

  def polySam[U](f: Sam): String = f(1)
  def polySam[U](f: SamP[U]): U = f(1)

  // check that we properly instantiate java.util.function.Function's type param to String
  def polyJavaSam(f: String => String) = 1
  def polyJavaSam(f: java.util.function.Function[String, String]) = 2
}

class StringLike(xs: String) {
  def map[A](f: Char => A): Array[A] = ???
  def map(f: Char => Char): String = ???
}

object Test {
  val of = new OverloadedFun[Int](1)

  of.foo(_.toString) // error: missing parameter type for parameter _$1, expected = ? => ?

  of.poly(x => x / 2 )
  of.polySam(x => x / 2 ) // error: missing parameter type for parameter x, expected = ?
  of.polyJavaSam(x => x)

  val sl = new StringLike("a")
  sl.map(_ == 'a')  // : Array[Boolean]
  sl.map(x => 'a')  // : String
}
```

lrytz:
LGTM

For the record, we considered an alternative solution that would use only the `Function`-typed higher-order method for inferring the parameter type, but this unequal treatment of functions and SAM types would be unnatural and lead to unexpected corner cases.

PR #5443, opened 2016-10-05
SI-5293 delete flaky collection performance tests

SethTisue:
review by @Ichoran, @adriaanm

note that the tests in question were already banished to "pending" once
already, by @adriaanm in 1729b26500506530733753d44f9ce2a2597e0e33.
no mercy for recidivists!


Ichoran:
I would just delete them.  This catches such a tiny fraction of the possible performance issues that it's hardly worth it, especially given a lousy benchmarking "framework" and various hard-coded magical values that may or may not be true depending on things like lack of warmup.

If we want performance regression tests we should do it properly.  They will necessarily take a while to run, though, so they should be part of the per-release test set, not the per-commit.

adriaanm:
Agreed. Benchmarks do not belong in the partest suite.

SethTisue:
OK, revised to delete rather than banish. I'll merge once Jenkins signs off.

PR #5445, opened 2016-10-06
SI-9946 Fix regressions with nullification of singly used fields

retronym:


sjrd:
Thanks for doing this in mixin rather than delambdafy :)

lrytz:
This adds a full tree traversal just for this optimization. I would prefer the previous proposal of collecting nullable fields in an earlier phase.

retronym:
@lrytz It doesn't add a traversal, but rather moves an existing traversal around. The traversal is only run for classes that contain lazy decls, so in practice the cost is pretty small. Furthermore, running it after `flatten` means we don't hit exponential performance for nested classes, which I think was the case in RC1.

My earlier patch that performed the analysis in explicitouter wasn't quite correct: the analysis should have been in a traversal after explicitouter. We could, for instance, detect the nullable fields in constructors and then emit the null assignments in mixin.

retronym:
That said, I'd still consider such a move a minor performance improvement that we could defer until 2.12.x

lrytz:
Oh right, I didn't realize it was an extra traversal already before.

Another small optimization / cleanup we could do in Mixin is avoid calling `preTransform` for every subtree, as it only ever handles `Template` trees (and `lzycompute` `DefDef`s after this patch).

lrytz:
LGTM!

PR #5448, opened 2016-10-07
Optimize List.reverseIterator

Atry:


Ichoran:
Do you have any evidence that this is meaningfully faster?

Atry:
@Ichoran It is quite obvious. Creating a temporary `Array` (this PR) is much faster than creating a temporary `List` (the original implementation).

xavier-fernandez:
@Atry, can you provide some benchmarks and its results?

Atry:
Benchmark: https://gist.github.com/Atry/3dc5d09d82fc6d88af590b35db9b7d97

Atry:
This PR implementation could be faster than now if `copyToArray` is inline.

Ichoran:
The benchmark is on 2.11 as it stands now, not your PR, I guess?

Can you use a proper benchmarking framework  (JMH or ScalaMeter or Caliper or at least Thyme)?  I'm a little worried about the loss of performance for short lists, but your numbers have not stabilized by the end of the stuff you printed out.

SethTisue:
are the JMH-based scaffolds in `test/benchmarks` usable here?

Ichoran:
Should be.  I haven't actually used them yet.

SethTisue:
@Atry interested in getting this over the finish line?

szeiger:
I've tried to benchmark the status quo, thinking I can compare the new `reverseIterator` against `reverse.iterator` (which is what `reverseIterator` currently does) but the results are surprising. The old `reverseIterator` is more than 3 times slower than `reverse.iterator`:

```
[info] Benchmark                                         (size)  Mode  Cnt      Score     Error  Units
[info] ListReverseIteratorBenchmark.reverseIterator          10  avgt   20    184.474 ±   8.826  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator         100  avgt   20   1718.719 ±  30.427  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator        1000  avgt   20  17854.919 ± 300.583  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator      10  avgt   20     52.709 ±   0.733  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator     100  avgt   20    558.202 ±   8.142  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator    1000  avgt   20   5723.814 ±  57.116  ns/op
```

lrytz:
Maybe the `toCollection` indirection in `def reverseIterator: Iterator[A] = toCollection(reverse).iterator` (which has a cast) prevents some optimizations, could you add this variant to the benchmark?

szeiger:
@lrytz Using a `def simpleReverseIterator = reverse.asInstanceOf[SeqLike[A, Repr]].iterator` defined right next to `reverseIterator` in `SeqLike` is only a few percent faster, even when it is made `final`. Moving the method down into `List` OTOH makes it as fast as `reverse.iterator`. The difference in dispatch must prevent some optimization (at the JVM level) of the actual `reverse` method (because the overhead scales linearly with the size of the list).

szeiger:
Benchmarking this change demonstrates two things:

1. Benchmarking is important - If you just guess you will guess wrong because of all the delicate optimizations happening in the JVM

2. Benchmarking is useless - Unless you benchmark exactly the use case you care about (which you never do in a micro-benchmark)

Here's the JMH code I used:

```
  @Param(Array("1", "10", "100", "1000"))
  var size: Int = _

  var values: List[Int] = _

  @Setup(Level.Trial) def initKeys(): Unit =
    values = (0 to size).toList

  @Benchmark def reverseIteratorOnly = values.reverseIterator

  @Benchmark def reverseThenIteratorOnly = values.reverse.iterator

  @Benchmark def reverseIterator = reverseIteratorOnly.foreach(_ => ())

  @Benchmark def reverseThenIterator = reverseThenIteratorOnly.foreach(_ => ())
```

Before this PR:

```
[info] Benchmark                                             (size)  Mode  Cnt      Score      Error  Units
[info] ListReverseIteratorBenchmark.reverseIteratorOnly           1  avgt    5     33.519 ±    0.869  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly          10  avgt    5    175.433 ±    7.491  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly         100  avgt    5   1447.529 ±  108.374  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly        1000  avgt    5  14076.380 ± 1078.826  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly       1  avgt    5     12.662 ±    0.691  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly      10  avgt    5     41.880 ±    2.279  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly     100  avgt    5    350.122 ±   25.178  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly    1000  avgt    5   3647.474 ±  104.670  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator               1  avgt    5     51.375 ±    4.103  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator              10  avgt    5    243.367 ±   13.722  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator             100  avgt    5   1988.538 ±  185.945  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator            1000  avgt    5  21960.133 ±  958.672  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator           1  avgt    5      9.693 ±    0.742  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator          10  avgt    5     52.145 ±    2.241  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator         100  avgt    5    560.490 ±   51.044  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator        1000  avgt    5   5633.995 ±   95.964  ns/op
```

With this PR:

```
[info] Benchmark                                             (size)  Mode  Cnt      Score     Error  Units
[info] ListReverseIteratorBenchmark.reverseIteratorOnly           1  avgt    5     44.522 ±   1.553  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly          10  avgt    5    127.666 ±   6.200  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly         100  avgt    5    997.544 ±  48.275  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly        1000  avgt    5   9637.756 ± 327.259  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly       1  avgt    5     12.921 ±   0.740  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly      10  avgt    5     43.003 ±   2.719  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly     100  avgt    5    355.850 ±  20.409  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly    1000  avgt    5   3744.242 ± 113.824  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator               1  avgt    5     56.166 ±   3.540  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator              10  avgt    5    180.745 ±   7.132  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator             100  avgt    5   1540.986 ±  66.046  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator            1000  avgt    5  14848.339 ± 682.518  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator           1  avgt    5      9.669 ±   0.736  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator          10  avgt    5     53.010 ±   4.426  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator         100  avgt    5    559.426 ±  24.375  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator        1000  avgt    5   5862.856 ± 416.236  ns/op
```

It's looking good so far for the usefulness of benchmarking. We can see that the change is clearly a win for most cases. Only lists of size 1 were faster to reverse under the old version. This is unfortunate because you generally try to avoid list reversal unless you know that the lists are very small.

We can also see that `reverseIterator` is still *much* slower than `reverse.iterator` in all cases.

And now to shake things up a bit, here is an alternate improvement that simply adds `override def reverseIterator = reverse.iterator` to `List` instead of this PR:

```
[info] Benchmark                                             (size)  Mode  Cnt      Score     Error  Units
[info] ListReverseIteratorBenchmark.reverseIteratorOnly           1  avgt    5     12.745 ±   1.077  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly          10  avgt    5     42.779 ±   2.118  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly         100  avgt    5    353.529 ±  28.702  ns/op
[info] ListReverseIteratorBenchmark.reverseIteratorOnly        1000  avgt    5   3666.602 ± 314.670  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly       1  avgt    5     12.795 ±   0.351  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly      10  avgt    5     42.340 ±   3.168  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly     100  avgt    5    347.912 ±  21.738  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIteratorOnly    1000  avgt    5   3664.193 ± 331.657  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator               1  avgt    5     35.115 ±   3.361  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator              10  avgt    5    210.960 ±  14.108  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator             100  avgt    5   1916.455 ± 139.488  ns/op
[info] ListReverseIteratorBenchmark.reverseIterator            1000  avgt    5  19229.420 ± 566.683  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator           1  avgt    5      9.666 ±   0.567  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator          10  avgt    5     52.759 ±   2.965  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator         100  avgt    5    557.937 ±  41.402  ns/op
[info] ListReverseIteratorBenchmark.reverseThenIterator        1000  avgt    5   5667.702 ± 336.258  ns/op
```

Now `reverseIterator` is just as fast as `reverse.iterator`! At least as long as you don't use it. When we also iterate through the results (the tests without `Only`), we're back to the original pathetic performance.

So what's the conclusion for this PR? From these results it's not clear which version is the best (or rather the least bad one).

lrytz:
If someone has more time to spend on this question, comparing the assembly (`-prof perfasm`) would probably reveal the interesting difference. Maybe also inliner logs. We could then search for a known issue or report it upstream.

Learning more about this issue might also point out performance problems for other operations.

Atry:
I guess we could add a `@inline` on `reverseIterator` method in order to make it faster when `-opt` is on.

szeiger:
I'm inclined to sit this out for 2.12 and revisit it in the context of https://github.com/scala/collection-strawman. Maybe we'll have more luck with the method dispatch on the somewhat simpler design.

adriaanm:
Closing based on @szeiger's feedback. Let me know if we should reopen!

PR #5449, opened 2016-10-08
SI-9953 Any Any aborts warn on equals

som-snytt:
Don't warn about equals if any `Any` is involved. cf SI-8965

The condition for warning is that both types lub to a supertype
of Object.

JIRA: https://issues.scala-lang.org/browse/SI-9953


SethTisue:
This warning seems tricky enough to get right (SI-5663, SI-5648, and maybe others) that one wonders whether it's worth having in the main codebase (as opposed to a type-aware linter like abide) at all :-\

delegating to a linter means you don't get warnings in the REPL though :-\

oh well. regardless, LGTM

PR #5452, opened 2016-10-09
Fix the interface flag when re-writing a closure call to the body method

lrytz:
When re-writing a closure invocation to the body method, the `itf` flag
of the invocation instruction was incorrect: it needs to be true if
the method is defined in an interface (including static methdos), not
if the method is invoked through `INVOKEINTERFACE`.

JDK 8 doesn't flag this inconsistency and executes the bytecode, but the
verifier in JDK 9 throws an `IncompatibleClassChangeError`.

Similar fixes went into e619b03.

Fixes https://github.com/scala/scala-dev/issues/242

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
@@ -325,8 +325,7 @@ class ClosureOptimizer[BT <: BTypes](val btypes: BT) {
         insns.insertBefore(invocation, new InsnNode(DUP))
         INVOKESPECIAL
     }
-    val isInterface = bodyOpcode == INVOKEINTERFACE
-    val bodyInvocation = new MethodInsnNode(bodyOpcode, lambdaBodyHandle.getOwner, lambdaBodyHandle.getName, lambdaBodyHandle.getDesc, isInterface)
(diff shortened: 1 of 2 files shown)


lrytz:
I vote for putting this into RC2..

src/compiler/scala/tools/nsc/backend/jvm/opt/ClosureOptimizer.scala
```diff
@@ -325,7 +325,7 @@ class ClosureOptimizer[BT <: BTypes](val btypes: BT) {
         insns.insertBefore(invocation, new InsnNode(DUP))
         INVOKESPECIAL
     }
-    val isInterface = bodyOpcode == INVOKEINTERFACE
+    val isInterface = classBTypeFromInternalName(lambdaBodyHandle.getOwner).isInterface.get
```
retronym: Could we ever hit a case where we don't `classBTypeFromInternalName.apply` or `.get` fails, say if the lambda target class was not in the current compilation classpath? (Maybe a lambda had previously been inlined into the class from which we are inlining.)

retronym: Oh, we can just call `lambdaBodyHandle.isInterface` here!

lrytz: ok, used that instead

retronym:
LGTM. :thumbsup: for 2.12.0

PR #5454, opened 2016-10-12
SI-9834 Improve error on failed op=

som-snytt:
If rewriting x += y fails to typecheck, emit error messages
for both the original tree and the assignment.

If rewrite is not attempted because x is a val, then say so.

Supersedes https://github.com/scala/scala/pull/5254 from last June. I must've forgotten it was against 2.11. What fashions did we wear and what music did we listen to back then? Who was president?

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4501,20 +4501,54 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         val appStart = if (Statistics.canEnable) Statistics.startTimer(failedApplyNanos) else null
         val opeqStart = if (Statistics.canEnable) Statistics.startTimer(failedOpEqNanos) else null
 
-        def onError(reportError: => Tree): Tree = fun match {
-          case Select(qual, name) if !mode.inPatternMode && nme.isOpAssignmentName(newTermName(name.decode)) =>
+        def isConversionCandidate(qual: Tree, name: Name): Boolean =
+          !mode.inPatternMode && nme.isOpAssignmentName(TermName(name.decode)) && !qual.exists(_.isErroneous)
+
+        def reportError(error: SilentTypeError): Tree = {
+          error.reportableErrors foreach context.issue
+          error.warnings foreach { case (p, m) => context.warning(p, m) }
+          args foreach (arg => typed(arg, mode, ErrorType))
+          setError(tree)
+        }
+        def advice1(errors: List[AbsTypeError], err: SilentTypeError): List[AbsTypeError] =
+          errors.map { e =>
+            if (e.errPos == tree.pos) {
+              val header = f"${e.errMsg}%n  Expression does not convert to assignment because:%n    "
+              NormalTypeError(tree, err.errors.flatMap(_.errMsg.lines.toList).mkString(header, f"%n    ", ""))
+            } else e
+          }
+        def advice2(errors: List[AbsTypeError]): List[AbsTypeError] =
+          errors.map { e =>
+            if (e.errPos == tree.pos) {
+              val msg = f"${e.errMsg}%n  Expression does not convert to assignment because receiver is not assignable."
+              NormalTypeError(tree, msg)
+            } else e
+          }
+        def onError(error: SilentTypeError): Tree = fun match {
+          case Select(qual, name) if isConversionCandidate(qual, name) =>
             val qual1 = typedQualifier(qual)
             if (treeInfo.isVariableOrGetter(qual1)) {
               if (Statistics.canEnable) Statistics.stopTimer(failedOpEqNanos, opeqStart)
-              convertToAssignment(fun, qual1, name, args)
+              val erred = qual1.isErroneous || args.exists(_.isErroneous)
+              if (erred) reportError(error) else {
+                val convo = convertToAssignment(fun, qual1, name, args)
+                silent(op = _.typed1(convo, mode, pt)) match {
+                  case SilentResultValue(t) => t
+                  case err: SilentTypeError => reportError(SilentTypeError(advice1(error.errors, err), error.warnings))
+                }
+              }
             }
             else {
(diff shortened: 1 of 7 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4506,16 +4506,34 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                    err.errors foreach (e => context.reporter.echo(s" -- ${e.errMsg}"))
+                    t
+                }
+              }
             }
             else {
               if (Statistics.canEnable) Statistics.stopTimer(failedApplyNanos, appStart)
-                reportError
+              val t = reportError
+              val Apply(Select(qual2, _), args2) = t
```
adriaanm: `val t@Apply(Select(qual2, _), args2) = reportError`?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4506,16 +4506,34 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
               if (Statistics.canEnable) Statistics.stopTimer(failedOpEqNanos, opeqStart)
-              convertToAssignment(fun, qual1, name, args)
+              val erred = qual1.isErroneous || args.exists(_.isErroneous)
+              if (erred) reportError else {
+                val convo = convertToAssignment(fun, qual1, name, args)
+                silent(op = _.typed1(convo, mode, pt)) match {
+                  case SilentResultValue(t) => t
+                  case err: SilentTypeError =>
+                    val t = reportError
+                    context.reporter.echo("Expression does not convert to assignment because:")
```
adriaanm: What's the reasoning behind `echo`ing? I'd expect to see `context.issue`/`context.warning`?

adriaanm (changes requested):
Looks good overall, but I have a few questions after a first pass.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
som-snytt: Because there's no `context.reporter.sendTextAlert`? I don't remember, so you're probably right as rain.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
som-snytt: Check! People think newlines grow on trees. Wait was that a compiler joke? Next time someone asks for a feature, People think syntax grows on trees!

src/compiler/scala/tools/nsc/typechecker/Typers.scala
som-snytt: I remember now, we're just appending to the reported error; we don't want to bump an error or warning count.

som-snytt: The reportError function should take the optional text to append.

som-snytt: Also, I don't know what 'start a review' and 'review changes' buttons do yet.

dragos: I think this may be missed by tools, like Ensime/Eclipse and maybe even Sbt. It might play better if the additional text was part of the original error.

som-snytt: I thought about it briefly last night and agree, but not sure about code yet. I haven't looked at dotty "explain" messages yet, and whether they include an "append advice" feature. I like the idea of decoupled, "something happened, and this additional text may help."

som-snytt:
More retro history is that the crasher fix on 2.12 is https://github.com/scala/scala/pull/5179

som-snytt:
Before issuing errors, update the error(s) at the tree.pos with the additional text. It would be nicer to `err.copy(errMsg = "more")` to preserve the actual `TypeError` in case a reporter cares.

som-snytt:
@SethTisue "Hello, Seth, can you tell me how I broke the build?" (If I could talk to Seth's Build Tool like Siri.)

adriaanm:
https://github.com/scala/scala-dev/issues/251 is looking less and less spurious.

som-snytt:
“Spuriouser and spuriouser!” Cried Alice (she was so much surprised, that for the moment she quite forgot how to speak good English).

som-snytt:
How about showing the expansion as well:

```
scala> val arr1 = new Array[Byte](10)
arr1: Array[Byte] = Array(0, 0, 0, 0, 0, 0, 0, 0, 0, 0)

scala> val arr2 = new Array[Byte](10)
arr2: Array[Byte] = Array(0, 0, 0, 0, 0, 0, 0, 0, 0, 0)

scala> arr1(4) ^= arr2(5)
<console>:14: error: value ^= is not a member of Byte
  Expression does not convert to assignment because:
    type mismatch;
     found   : Int
     required: Byte
    expansion: arr1.update(4, arr1.apply(4).^(arr2(5)))
       arr1(4) ^= arr2(5)
               ^

scala> val j = collection.mutable.Set("a")
j: scala.collection.mutable.Set[String] = Set(a)

scala> j("a") += "b"
<console>:13: error: value += is not a member of Boolean
  Expression does not convert to assignment because:
    type mismatch;
     found   : String
     required: Boolean
    expansion: j.update("a", j.apply("a").$plus("b"))
       j("a") += "b"
              ^
```
I see the caret operator is not shown as `$up`.

lrytz:
Looks very nice. Could we even include the receiver expression? `Expression does not convert to assignment because x is not assignable`.

retronym:
> Could we even include the receiver expression

This can lead to poor messages in cases when the expression is long, or contained desugarings. Usually we just identify it with the error position and only refer to the expressions type in the message.

som-snytt:
That was my original reason, although the word `receiver` could be poorly received. But maybe use a simple ident when you've got one.

I also forgot my comment that the expansion string shows `$plus` but not `$up`. What's `$up` with that? Or more colloquially, `Wass^?`

The message will be perfect in dotty...

adriaanm (approved):

PR #5455, opened 2016-10-12
Add clarification on preservation of order

coryfklein:
Future.sequence does not preserve order of the original collection. The
current implementation creates a function that adds the results to the
final collection via a builder, but that function does not get called
until the Future is resolved.

So if the futures do not resolve in the
same order in which they are passed, the final collection will not have
the same order as the original.

### DIFF
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -494,7 +494,7 @@ object Future {
   def apply[T](body: =>T)(implicit @deprecatedName('execctx) executor: ExecutionContext): Future[T] = impl.Future(body)
 
   /** Simple version of `Future.traverse`. Transforms a `TraversableOnce[Future[A]]` into a `Future[TraversableOnce[A]]`.
-   *  Useful for reducing many `Future`s into a single `Future`.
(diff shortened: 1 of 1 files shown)


coryfklein:
While writing a new `sequence` method that does maintain order, I realized that the original method signature may already communicate the intent of this PR:

``` scala
def sequence[A, M[X] <: TraversableOnce[X]](in: M[Future[A]])(...): Future[M[A]]
```

When I read this, I did not immediately recognize that order would be lost on any `Iterable` passed into the method due to the method accepting a `TraversableOnce`. I still think the clarification of this PR would make that intent stand out more, preventing others like me from misunderstanding the complete implications of the method signature.

coryfklein:
Huh. I thought I was seeing `sequence` return results out-of-order, but now I am having difficulty reproducing it...

coryfklein:
Pretty sure this PR is wrong. If you pass in an ordered collection, you'll get an ordered collection back out.

adriaanm:
Thanks for investigating!

coryfklein:
Some day I'll contribute something of worth here... some day. 👍

PR #5462, opened 2016-10-18
Avoid use of legacy JVM MaxPermSize option in partest

retronym:
In Java 8+, which we require on this branch, this option
is a no-op and triggers a JVM warning these days.

See  http://openjdk.java.net/jeps/122


SethTisue:
LGTM

PR #5464, opened 2016-10-18
[backport] Bump sbt.version to 0.13.12, without breaking

retronym:
Aiming to keep 2.11 and 2.12 SBT builds as common as possible.


retronym:
Review by @SethTisue

SethTisue:
LGTM. there’s not really any risk here given that in 2.11 world the sbt build isn’t really connected to anything

PR #5469, opened 2016-10-20
SI-10020 SI-10027 Scaladoc: keep Java comment scanning stack-friendly [ci: last-only]

adriaanm:
... and boy scout the area a bit, reducing duplication.

I was led to this part of the forest while investigating some stack overflows during Java comment scanning. The regression (in #5246, /cc @jodersky FYI) was a result of some hooks into formerly tail-recursive methods. They are once more stack-friendly, and also finally final.

Functionality is intended to be identical.

Review by @som-snytt because boy scout?


adriaanm:
/rebuild I'm pretty sure there's little difference between these commits in terms of what the test suite could reveal.

som-snytt:
I'll take a look because "friendly", although the scouts won't take me back.

adriaanm:
Last commit is green -- squash on merge?

adriaanm:
/nothingtoseehere -- test failures in earlier commits were due to jenkins load

adriaanm:
@jodersky, could I perhaps guilt you into a review? 😁

jodersky:
of course, such an embarrassing mistake 😳

adriaanm:
Not at all! I didn't catch it either. OO bad. `final` good! Also, that method should've been explicitly `@tailrec`

jodersky:
LGTM! Maybe we should add a regression test like this https://github.com/scala/scala/compare/2.12.x...jodersky:SI-10027-test ? I can do it in a separate PR if you prefer to just get this fix in

adriaanm:
Thanks! Cherry picked.

src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
```diff
@@ -38,6 +38,27 @@ trait ScannersCommon {
...
+    def flushDoc(): DocComment = try lastDoc finally lastDoc = null
+    def registerDocComment(raw: String, pos: Position) = {
+      lastDoc = DocComment(raw, pos)
+      signalParsedDocComment(raw, pos)
+    }
+
+    /** To prevent doc comments attached to expressions from leaking out of scope
+      *  onto the next documentable entity, they are discarded upon passing a right
+      *  brace, bracket, or parenthesis.
+      */
```
som-snytt, line 58: I guess this could wait for scalafmt.

adriaanm: Not my best spacing work.

som-snytt:
Sorry I didn't get back to this. Looking now to see how to make a scaladoc test...

adriaanm:
No worries, my guilt-trip efforts yielded a test and a review.

src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
```diff
@@ -103,11 +124,11 @@ trait Scanners extends ScannersCommon {
     }
   }
 
-  abstract class Scanner extends CharArrayReader with TokenData with ScannerData with ScannerCommon {
+  abstract class Scanner extends CharArrayReader with TokenData with ScannerData with ScannerCommon with DocScanner {
     private def isDigit(c: Char) = java.lang.Character isDigit c
 
     private var openComments = 0
-    protected def putCommentChar(): Unit = nextChar()
+    final protected def putCommentChar(): Unit = { processCommentChar(); nextChar() }
```
kiritsuku, line 131: Making this method final is a source incompatible change, which are supposed not to be done in minor Scala versions afaik. Concretely it broke the IDE. We can't update from 2.12.0 to 2.12.1. Since 2.12.1 is already released it is too late to revert the change. We may find a workaround for it but I wanted to mention in anyway.

SethTisue: We don't make those sorts of firm compatibility guarantees for compiler internals. 

In the past, our shared backstop for this sort of thing was `scripts/jobs/integrate/ide` — if we broke something the IDE was using, it would tell us. But that script was never updated to work with 2.12, so it hasn't been part of PR validation or nightly testing for a long time now.

If we want to avoid this kind of thing happening, I think either that script would need to be restored to be part of Scala's CI again, or (to catch the problem a little later, but still relatively fast), you would have to add something comparable to your own CI that tests your code against 2.12.x (and 2.13.x?) nightlies.

kiritsuku: Ok. I hope once we are fully compatible to 2.12.1 we can re-enable the IDE integration CI job.

PR #5473, opened 2016-10-20
make a compiler warning go away

SethTisue:
namely "patterns after a variable pattern cannot match (SLS 8.1.1)"


SethTisue:
review by @lrytz

som-snytt:
This is cleaned up in https://github.com/scala/scala/pull/5393

PR #5474, opened 2016-10-21
Improve hash code of Names

retronym:
The old approach of using the first, last, and middle characters
only lays a trap for generate names that have little or no entropy
at these locations. Fresh existential names generated in "as seen
from" operations are one such case, and when compiling large
batches of files the name table can become imbalanced.

I found such an example in ScalaTest:

   https://github.com/scala/scala-dev/issues/246#issuecomment-255338925

This commit uses all characters to compute the hashCode.

Review by @adriaanm @lrytz

I expect that this change will be beneficial on large code bases
that regularly exercise `AsSeenFrom#captureThis`, and neutral on
most smaller code bases. I'll perform some benchmarks to make sure
of that "neutral" claim and post the results here.

### DIFF
--- a/src/reflect/scala/reflect/internal/Names.scala
+++ b/src/reflect/scala/reflect/internal/Names.scala
@@ -47,17 +47,15 @@ trait Names extends api.Names {
   /** Hashtable for finding type names quickly. */
   private val typeHashtable = new Array[TypeName](HASH_SIZE)
 
-  /**
-   * The hashcode of a name depends on the first, the last and the middle character,
-   * and the length of the name.
-   */
-  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int =
-    if (len > 0)
-      (len * (41 * 41 * 41) +
-       cs(offset) * (41 * 41) +
-       cs(offset + len - 1) * 41 +
-       cs(offset + (len >> 1)))
-    else 0
+  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int = {
+    var h = 0
+    var i = 0
+    while (i < len) {
+      h = 31 * h + cs(i + offset)
+      i += 1
+    }
+    h
+  }
 
   /** Is (the ASCII representation of) name at given index equal to
    *  cs[offset..offset+len-1]?


retronym:
Related to https://github.com/scala/scala-dev/issues/246

src/reflect/scala/reflect/internal/Names.scala
```diff
@@ -41,16 +41,19 @@ trait Names extends api.Names {
   private val typeHashtable = new Array[TypeName](HASH_SIZE)
 
   /**
-   * The hashcode of a name depends on the first, the last and the middle character,
+   * The hashcode of is equivalent to cs name depends on the first, the last and the middle character,
```
dragos: Hm, looks like a partial update, you probably need to rephrase the whole line.

lrytz: this reads a bit strangely now..

lrytz:
great numbers for ScalaTest, hope there's no negative impact on other projects

src/reflect/scala/reflect/internal/Names.scala
```diff
@@ -41,16 +41,19 @@ trait Names extends api.Names {
...
    * and the length of the name.
    */
-  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int =
-    if (len > 0)
-      (len * (41 * 41 * 41) +
-       cs(offset) * (41 * 41) +
-       cs(offset + len - 1) * 41 +
-       cs(offset + (len >> 1)))
-    else 0
+  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int = {
```
DarkDimius, line 50: @retronym do you have a guarantee that strings don't repeat in your name-table?(Dotty does)
If yes, than you won't need to consider character values at all as default hashcode(system identity hashcode) would work correctly.

lrytz: @DarkDimius how do you guarantee that strings don't repeat? Let's say you first create a name "abc", then a name "bc", will the name table only contain "abc"?

Also, to me, dotty's [Names.scala](https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/Names.scala) looks quite similar to scala's. There's a `hashValue` method that looks the same as the one being replaced in this PR. Can you point out differences you have in mind?

DarkDimius: @lrytz, 

> @DarkDimius how do you guarantee that strings don't repeat? Let's say you first create a name "abc", then a name "bc", will the name table only contain "abc"?

It would contain both, but the next time you try to create `abc` you'll get the same one. See https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/Names.scala#L245

> Also, to me, dotty's Names.scala looks quite similar to scala's. There's a hashValue method that looks the same as the one being replaced in this PR. Can you point out differences you have in mind?

`hashValue` is only used when creating new Term names. It's not used when comparing hashcodes.
Hashcode is https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/Names.scala#L178

lrytz: This looks to me exactly the same as in scala https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/Names.scala#L233, so I still don't understand where the difference is to dotty..

DarkDimius: Oh, I misunderstood the intention of this PR. I thought it changes the hashcode. 
Dotty has the same issue.

src/reflect/scala/reflect/internal/Names.scala
retronym: Oops. I've just pushed the full change.

retronym:
Here's one way to perform the followup work (using per-compilation unit counters rather than a global counter to generate the fresh existential names):

``` diff
diff --git a/src/reflect/scala/reflect/internal/Symbols.scala b/src/reflect/scala/reflect/internal/Symbols.scala
index f870ecf..6f74b2f 100644
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -34,9 +34,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
   def recursionTable = _recursionTable
   def recursionTable_=(value: immutable.Map[Symbol, Int]) = _recursionTable = value

-  private var existentialIds = 0
-  protected def nextExistentialId() = { existentialIds += 1; existentialIds }
-  protected def freshExistentialName(suffix: String) = newTypeName("_" + nextExistentialId() + suffix)
+  protected def freshExistentialName(suffix: String) = newTypeName(currentFreshNameCreator.newName("_") + suffix)

   // Set the fields which point companions at one another.  Returns the module.
   def connectModuleToClass(m: ModuleSymbol, moduleClass: ClassSymbol): ModuleSymbol = {
diff --git a/src/reflect/scala/reflect/runtime/SynchronizedSymbols.scala b/src/reflect/scala/reflect/runtime/SynchronizedSymbols.scala
index 237afa0..2cbf0a7 100644
--- a/src/reflect/scala/reflect/runtime/SynchronizedSymbols.scala
+++ b/src/reflect/scala/reflect/runtime/SynchronizedSymbols.scala
@@ -10,9 +10,6 @@ private[reflect] trait SynchronizedSymbols extends internal.Symbols { self: Symb
   private lazy val atomicIds = new java.util.concurrent.atomic.AtomicInteger(0)
   override protected def nextId() = atomicIds.incrementAndGet()

-  private lazy val atomicExistentialIds = new java.util.concurrent.atomic.AtomicInteger(0)
-  override protected def nextExistentialId() = atomicExistentialIds.incrementAndGet()
-
   private lazy val _recursionTable = mkThreadLocalStorage(immutable.Map.empty[Symbol, Int])
   override def recursionTable = _recursionTable.get
   override def recursionTable_=(value: immutable.Map[Symbol, Int]) = _recursionTable.set(value)
```

retronym:
My first attempt to benchmark this for small programs showed no change (795ms->797ms) for `better-files`, and a 2% slowdown (1645ms -> 1678ms) for `scalap`. However, the results seem a little noiser on my laptop (even with TurboBoost disabled) than i was used to on my desktop where I've done previous benchmarking, so I'm not sure I have enough warmup time or forks to say whether or not that 2% is just noise.

I plan to re-run the benchmarks, and also compare profiles before/after to see is `hashValue` is starting to show up.

adriaanm:
Have you had a chance to confirm the benchmark results?

retronym:
Nope, I haven't. I'm going to close this one for now and revisit once we've got some more benchmarking coverage next year.

retronym:
Benchmark run: https://scala-ci.typesafe.com/view/scala-bench/job/compiler-benchmark/2573/console

retronym:
I'm happy with the performance, and will merge this now.

lrytz:
> I'm happy with the performance

What did you observe?

retronym:
@lrytz [neutral to ~1% regression](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?orgId=1&from=1549080972797&to=1555585465453&var-scalaSha=2d2b8959c3%7C3b36ff82c1%7Cc5877a73e7%7C652f49bbfb%7Cc2d25ac469%7Cf6d09146bd%7Cebef3cddc8%7Cf0ec40c1cc%7Cfff4dc0391%7C82e6d0871a%7C51230ecbeb%7C48c3e1dba1&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40)

I'm hopeful that #8019 will recover a little of the regression (if it is real, that is!). I'll keep an eye on the performance charts and the results our profile runs now both changes are merged.

retronym:
The motivation to revive this, BTW, was that we had a report from a customer that they had a name table bucket with 500 entries, suggesting that something in the compiler or a macro/plugin was running into the same problem as we found/fixed in https://github.com/scala/scala/pull/5506

PR #5475, opened 2016-10-22
Change synchronization method in TwoWayCache for more parallelism

AlexeyIvanov8:
Remove full synchronization of TwoWayCache methods and synchronize only cache maps. Because in multithreading app full synchronization cache operations implies performance degradation. Altough synchronized need only for two maps: toJava and toScala and only for write operations.

I'm replace they from WeakHashMap to synchronyzed wraper on WeakHashMap, because there are not exsists true concurrent WeakHashMap modification.

Before changes in my performance tests number of reflections operations decrease when I was add new threads. And in threads dump can see what most time thread try attempt lock. After changes number operations have linear dependensy of threads count.

### DIFF
--- a/src/reflect/scala/reflect/runtime/TwoWayCaches.scala
+++ b/src/reflect/scala/reflect/runtime/TwoWayCaches.scala
@@ -32,7 +32,7 @@ private[runtime] trait TwoWayCaches { self: SymbolTable =>
         } else None
     }
 
-    def toScala(key: J)(body: => S): S = {
+    def syncToScala(key: J)(body: => S): S = gilSynchronized {
       toScalaMap get key match {
         case SomeRef(v) =>
           v
@@ -43,7 +43,16 @@ private[runtime] trait TwoWayCaches { self: SymbolTable =>
       }
     }
 
-    def toJava(key: S)(body: => J): J = {
+    def toScala(key: J)(body: => S): S = {
+      toScalaMap get key match {
+        case SomeRef(v) =>
+          v
+        case _ =>
+          syncToScala(key)(body)
+      }
+    }
+
+    def syncToJava(key: S)(body: => J): J = gilSynchronized {
       toJavaMap get key match {
         case SomeRef(v) =>
           v
@@ -54,7 +63,16 @@ private[runtime] trait TwoWayCaches { self: SymbolTable =>
       }
     }
 
-    def toJavaOption(key: S)(body: => Option[J]): Option[J] = {
+    def toJava(key: S)(body: => J): J = {
+      toJavaMap get key match {
+        case SomeRef(v) =>
+          v
+        case _ =>
+          syncToJava(key)(body)
+      }
+    }
+
+    def syncToJavaOption(key: S)(body: => Option[J]): Option[J] = gilSynchronized {
       toJavaMap get key match {
         case SomeRef(v) =>
           Some(v)
@@ -62,6 +80,15 @@ private[runtime] trait TwoWayCaches { self: SymbolTable =>
           val result = body
           for (value <- result) enter(value, key)
           result
+      }
+    }
+
+    def toJavaOption(key: S)(body: => Option[J]): Option[J] = {
+      toJavaMap get key match {
+        case SomeRef(v) =>
+          Some(v)
+        case _ =>
+          syncToJavaOption(key)(body)
       }
     }
   }
--- a/src/reflect/scala/reflect/runtime/TwoWayCaches.scala
+++ b/src/reflect/scala/reflect/runtime/TwoWayCaches.scala
@@ -1,8 +1,10 @@
 package scala.reflect
 package runtime
 
-import scala.collection.mutable.WeakHashMap
 import java.lang.ref.WeakReference
+import java.util.{Collections, WeakHashMap}
+
+import scala.collection.convert.decorateAsScala._
 
 /** A cache that maintains a bijection between Java reflection type `J`
  *  and Scala reflection type `S`.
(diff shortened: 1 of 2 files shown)


lrytz:
@xeno-by do you have time to take a look?

AlexeyIvanov8:
I'm write benchmark for my changes, they also shows that exist more effictively solution(ConcurrentHashMap, but it no have full syntactically identity), for discussion. Results for my computer placed in readme:

[scala-performance-tests](https://github.com/AlexeyIvanov8/scala-performance-tests)

adriaanm:
ping @xeno-by

src/reflect/scala/reflect/runtime/TwoWayCaches.scala
```diff
@@ -52,7 +54,7 @@ private[runtime] trait TwoWayCaches { self: SymbolTable =>
       }
     }
 
-    def toJavaOption(key: S)(body: => Option[J]): Option[J] = gilSynchronized {
```
retronym, line 55: We need to be sure that calling code was not relying on the coarse grained lock here to safely execute the `body` thunk. We do have a test of runtime reflection thread safety in test/files/run/reflection-sync-potpourri.scala and test/files/run/reflection-sync-subtypes.scala, so it is encouraging to see that they still pass after your patch. But there could be race conditions that aren't exercised by that test.

xeno-by:
@AlexeyIvanov8 Do you have (possibly rough) benchmark numbers for situations before your patch and after your patch?

retronym:
@xeno-by https://github.com/AlexeyIvanov8/scala-performance-tests#test-results

retronym:
If we want to get more adventurous, we could actually implement the Java to Scala caches in terms of [`java.lang.ClassValue`](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassValue.html), and the Scala to Java cache in terms of `Symbol` attachments (using a fine grained lock on the particular `Symbol`)

retronym:
Actually, `ClassValue` would only be directly useful for maps with `java.lang.Class` as the key. We use the same pattern with `Field`, `Method`, etc as the keys.

xeno-by:
I second @retronym's comment. Performance gains are respectable, but we need to make sure that body-computing thunks passed to calls to `TwoWayCache` are fine to run in parallel.

AlexeyIvanov8:
@retronym, @xeno-by , thank for explanations, I'm accept the problem with race condition. I think this can be avoided by that modification(for one case, as example):

```
def syncToScala(key: J)(body: => S): S = synchronized {
    toScalaMap.get(key) match {
      case SomeRef(v) =>
        v
      case _ =>
        val result = body
        enter(key, result)
        result
    }
  }

  def toScala(key: J)(body: => S): S = {
    toScalaMap.get(key) match {
      case SomeRef(v) =>
        v
      case _ =>
        syncToScala(key)(body)
    }
  }
```

My benchmark also show linear depndency of operation count from number threads in this case.

retronym:
@AlexeyIvanov8 That looks better, so long as you're using `gilSynchronized` rather than `this.synchronized { .. .}` in the places where we used `gilSynchronized` before.

AlexeyIvanov8:
I leave the same sync methods as define currently: [TwoWayCache](https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/runtime/TwoWayCache.scala). In [TwoWayCaches](https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/runtime/TwoWayCaches.scala) planned still use gilSynchronyzed. But gilSynchronyzed take a lock only for body and really look better. I can use gilSynchronyzed instead of this.synchronyzed in both caches.

AlexeyIvanov8:
Added separate methods for synchronized write access to cache.

xeno-by:
@retronym I think that the patch looks alright now. Wdyt?

retronym:
@AlexeyIvanov8 Sorry that this has been left in limbo for so long, it fell off my radar.

I've taken a fresh look at the proposed change. By introducing locking on the maps (via the `synchronizedMap` wrappers, we are exposed to deadlocks due if different threads try to acquire these locks + the GIL in different orders. But the locks on the maps are never held while running a callback (the `body` thunk), so I _think_ this is not an issue.

However, we still have a potential race if two threads observe a missing key, and then each compute the value and write it into the map. The threads will mutate the Map and the symbol table serially, but what the callback for a given key could be called more than once. We would need some idempotency guarantees here. 

Furthermore, I find client code of these caches that I don't understand.

```
    private def jmethodAsScala(jmeth: jMethod): MethodSymbol =
      toScala(methodCache, jmeth)(_ jmethodAsScala1 _)

    private def jmethodAsScala1(jmeth: jMethod): MethodSymbol = {
      val clazz = sOwner(jmeth)
      val meth = clazz.newMethod(newTermName(jmeth.getName), NoPosition, jmeth.scalaFlags)
      methodCache enter (jmeth, meth) // eagerly enters the method symbol into the cache before completion of the `body` thunk!
      val tparams = jmeth.getTypeParameters.toList map createTypeParameter
      val paramtpes = jmeth.getGenericParameterTypes.toList map typeToScala
      val resulttpe = typeToScala(jmeth.getGenericReturnType)
      setMethType(meth, tparams, paramtpes, resulttpe)
      propagatePackageBoundary(jmeth.javaFlags, meth)
      copyAnnotations(meth, jmeth)
      if (jmeth.javaFlags.isVarargs) meth modifyInfo arrayToRepeated
      if (jmeth.getDefaultValue != null) meth.addAnnotation(AnnotationDefaultAttr)
      markAllCompleted(meth)
      meth
    }
```

Stepping back from this a little, I think it would be useful to experiment with changing the GIL into a `ReadWriteLock` and using lock free data structures in appropriate places.

retronym:
We should also rethink the design of `TwoWayCache`-s. In [this WIP branch](https://github.com/scala/scala/compare/2.12.x...retronym:topic/reflection-two-way-cache-purge?expand=1), I've:
  - Added a field to `Symbol` in runtime reflection to hold the underlying `java.lang.reflect.{Field,Method, ...}`. This replaces the global maps of weak references.
  - Used `java.lang.ClassValue` to store, for each class, maps of all `java.lang.reflect.{Field,Method, ...}` declared in that class to the corresponding Scala `Symbol`.
  - ... hence (I think), making it okay drop all of our weak reference handling. I'm not so sure of the intended lifecycle that `TwoWayCache` facilitated, though, so its hard to say.

SethTisue:
@retronym should this remain in the PR queue?

retronym:
Unfortunately the terrain here is too rough for me to accept the proposed PR or flesh out my ideas into something working. I'll close to reflect that reality. Thread safety in runtime reflection was a bit of a miraculous last minute addition. It's sort of amazing to me that it works at all, but it is certainly hard to reason about changes to improve the performance.

PR #5476, opened 2016-10-22
intern some doubles when boxing [ci: last-only]

fommil:
In applications that are number heavy, the number of boxings can contribute significantly to the heap size. This becomes a major GC problem when the `java.lang.Double` instances are strongly referenced.

In contrast, the JVM feels it is worth the effort to intern byte, int and long, but not double.

This PR works around the lack of interning in the JVM by manually caching a very small subset of `Double` instances that are generally the largest contributors to the heap in the typical application.

### DIFF
--- a/src/library/scala/runtime/BoxesRunTime.java
+++ b/src/library/scala/runtime/BoxesRunTime.java
@@ -73,10 +73,20 @@ public final class BoxesRunTime
         return java.lang.Float.valueOf(f);
     }
 
+    private static final Double DOUBLE_ZERO = java.lang.Double.valueOf(0.0);
+    private static final Double DOUBLE_NEG_ZERO = java.lang.Double.valueOf(-0.0);
+    private static final Double DOUBLE_UNIT = java.lang.Double.valueOf(1.0);
+    private static final Double DOUBLE_NEG_UNIT = java.lang.Double.valueOf(-1.0);
+
     public static java.lang.Double boxToDouble(double d) {
-        // System.out.println("box " + d);
-        // (new Throwable()).printStackTrace();
-        return java.lang.Double.valueOf(d);
+        // Double can detect negative zeros, but primitive == cannot
+        // also note that Double.NaN.equals(Double.NaN) is true, unlike primitive
+        Double value = java.lang.Double.valueOf(d);
+        if (value.equals(DOUBLE_ZERO)) return DOUBLE_ZERO;
+        else if (value.equals(DOUBLE_NEG_ZERO)) return DOUBLE_NEG_ZERO;
+        else if (value.equals(DOUBLE_UNIT)) return DOUBLE_UNIT;
+        else if (value.equals(DOUBLE_NEG_UNIT)) return DOUBLE_NEG_UNIT;
+        else return value;
     }
 
 /* UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING */
--- a/src/library/scala/runtime/BoxesRunTime.java
+++ b/src/library/scala/runtime/BoxesRunTime.java
@@ -74,19 +74,23 @@ public final class BoxesRunTime
     }
 
     private static final Double DOUBLE_ZERO = java.lang.Double.valueOf(0.0);
+    private static final long DOUBLE_ZERO_BITS = java.lang.Double.doubleToLongBits(0.0);
     private static final Double DOUBLE_NEG_ZERO = java.lang.Double.valueOf(-0.0);
+    private static final long DOUBLE_NEG_ZERO_BITS = java.lang.Double.doubleToLongBits(-0.0);
     private static final Double DOUBLE_UNIT = java.lang.Double.valueOf(1.0);
+    private static final long DOUBLE_UNIT_BITS = java.lang.Double.doubleToLongBits(1.0);
     private static final Double DOUBLE_NEG_UNIT = java.lang.Double.valueOf(-1.0);
+    private static final long DOUBLE_NEG_UNIT_BITS = java.lang.Double.doubleToLongBits(-1.0);
 
     public static java.lang.Double boxToDouble(double d) {
         // Double can detect negative zeros, but primitive == cannot
         // also note that Double.NaN.equals(Double.NaN) is true, unlike primitive
-        Double value = java.lang.Double.valueOf(d);
-        if (value.equals(DOUBLE_ZERO)) return DOUBLE_ZERO;
-        else if (value.equals(DOUBLE_NEG_ZERO)) return DOUBLE_NEG_ZERO;
-        else if (value.equals(DOUBLE_UNIT)) return DOUBLE_UNIT;
-        else if (value.equals(DOUBLE_NEG_UNIT)) return DOUBLE_NEG_UNIT;
-        else return value;
+        long bits = java.lang.Double.doubleToLongBits(d);
+        if (bits == DOUBLE_ZERO_BITS) return DOUBLE_ZERO;
+        else if (bits == DOUBLE_NEG_ZERO_BITS) return DOUBLE_NEG_ZERO;
+        else if (bits == DOUBLE_UNIT_BITS) return DOUBLE_UNIT;
+        else if (bits == DOUBLE_NEG_UNIT_BITS) return DOUBLE_NEG_UNIT;
+        else return java.lang.Double.valueOf(d);
     }
 
 /* UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING ... UNBOXING */
--- a/test/files/run/boxesRunTime-boxToDouble.scala
+++ b/test/files/run/boxesRunTime-boxToDouble.scala
@@ -1,18 +1,17 @@
 // partest test/files/run/boxesRunTime-boxToDouble.scala
 object Test {
   def main(args: Array[String]): Unit = {
-    assert(!(-0.0.asInstanceOf[java.lang.Double] eq 0.0.asInstanceOf[java.lang.Double]), "negative zero was positive")
-
     assert(0.0.asInstanceOf[java.lang.Double] eq 0.0.asInstanceOf[java.lang.Double], "zero was not interned")
     assert(-0.0.asInstanceOf[java.lang.Double] eq -0.0.asInstanceOf[java.lang.Double], "negative zero was not interned")
     assert(1.0.asInstanceOf[java.lang.Double] eq 1.0.asInstanceOf[java.lang.Double], "one was not interned")
     assert(-1.0.asInstanceOf[java.lang.Double] eq -1.0.asInstanceOf[java.lang.Double], "negative one was not interned")
 
-    // a reminder of the java.land.Double API vs Scala/Java primitives
+    // a reminder of the java.lang.Double vs primitive equality exceptions
     assert(java.lang.Double.NaN.equals(java.lang.Double.NaN))
     assert(Double.NaN != Double.NaN)
 
-    // canary, the JVM may start to cache these...
-    assert(!(100.0.asInstanceOf[java.lang.Double] eq 100.0.asInstanceOf[java.lang.Double]), "100 was interned")
+    assert(!(java.lang.Double.valueOf(0.0).equals(java.lang.Double.valueOf(-0.0))))
+    assert(0.0 == -0.0)
+    assert(!(-0.0.asInstanceOf[java.lang.Double] eq 0.0.asInstanceOf[java.lang.Double]), "negative zero was positive")
   }
 }
--- a/test/files/run/boxesRunTime-boxToDouble.scala
+++ b/test/files/run/boxesRunTime-boxToDouble.scala
@@ -1,0 +1,18 @@
+// partest test/files/run/boxesRunTime-boxToDouble.scala
+object Test {
+  def main(args: Array[String]): Unit = {
+    assert(!(-0.0.asInstanceOf[java.lang.Double] eq 0.0.asInstanceOf[java.lang.Double]), "negative zero was positive")
+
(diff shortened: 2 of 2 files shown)


test/files/run/boxesRunTime-boxToDouble.scala
```diff
@@ -0,0 +1,18 @@
...
+    assert(-0.0.asInstanceOf[java.lang.Double] eq -0.0.asInstanceOf[java.lang.Double], "negative zero was not interned")
+    assert(1.0.asInstanceOf[java.lang.Double] eq 1.0.asInstanceOf[java.lang.Double], "one was not interned")
+    assert(-1.0.asInstanceOf[java.lang.Double] eq -1.0.asInstanceOf[java.lang.Double], "negative one was not interned")
+
+    // a reminder of the java.land.Double API vs Scala/Java primitives
+    assert(java.lang.Double.NaN.equals(java.lang.Double.NaN))
+    assert(Double.NaN != Double.NaN)
+
+    // canary, the JVM may start to cache these...
+    assert(!(100.0.asInstanceOf[java.lang.Double] eq 100.0.asInstanceOf[java.lang.Double]), "100 was interned")
```
sjrd: > canary, the JVM may start to cache these...

And Scala.js actually does. So this test will fail on Scala.js already.

src/library/scala/runtime/BoxesRunTime.java
```diff
@@ -73,10 +73,20 @@ private static int typeCode(Object a) {
...
+    private static final Double DOUBLE_UNIT = java.lang.Double.valueOf(1.0);
+    private static final Double DOUBLE_NEG_UNIT = java.lang.Double.valueOf(-1.0);
+
     public static java.lang.Double boxToDouble(double d) {
-        // System.out.println("box " + d);
-        // (new Throwable()).printStackTrace();
-        return java.lang.Double.valueOf(d);
+        // Double can detect negative zeros, but primitive == cannot
+        // also note that Double.NaN.equals(Double.NaN) is true, unlike primitive
+        Double value = java.lang.Double.valueOf(d);
```
sjrd: This will always create the box, even if you end up using the interned box. What's the benefit, then?

sjrd (commented):
It's probably not a good idea to mess with boxing functions without benchmarks. This might actually significantly slow down pretty much every boxing operation of doubles that do not happen to have those 4 values.

Also, assuming this is desirable at all, the same should be done for `Float`s.

src/library/scala/runtime/BoxesRunTime.java
fommil: It's short lived

test/files/run/boxesRunTime-boxToDouble.scala
fommil: ok, I'll revert the canary

src/library/scala/runtime/BoxesRunTime.java
fommil: it might be possible possible to call out to `java.lang.Double.equals` statically, in which case we might even avoid the instance creation.

fommil:
how does one write performance tests for the scala standard library?

src/library/scala/runtime/BoxesRunTime.java
fommil: ok, looks like `java.lang.Double` just calls out to `doubleToLongBits` so I'll do that and avoid the alloc.

sjrd: `java.lang.Double.equals` is JDK8 only. But you can roll your own tests pretty easily:

``` scala
if (d == 0.0) {
  if (1.0 / d == Double.POSITIVE_INFINITY) return DOUBLE_ZERO;
  else return DOUBLE_NEG_ZERO;
} else {
  if (d == 1.0) return DOUBLE_UNIT;
  else if (d == -1.0) return DOUBLE_NEG_UNIT;
  else return java.lang.Double.valueOf(d);
}
```

VladUreche:
I agree with Seb on the benchmarking point, though I don't think we currently have good benchmarks for the Scala library. And this is a very deep change, that will impact all Scala programs.

One way to go, AFAICT, is to have the interning mechanism as a separate library and a compiler plugin which does the necessary redirects. WDYT?

fommil:
@VladUreche that won't catch problems in third party libraries. Of course, doing this in a `javaagent` is another way to go and something I am investigating.

VladUreche:
> @VladUreche that won't catch problems in third party libraries. Of course, doing this in a javaagent is another way to go and something I am investigating.

Now that you mention it, there's maybe an even better way, assuming:

``` scala
class ScalaRunTime {
  ...
}
object ScalaRunTime extends ScalaRunTime
```

You can change the `MODULE$` static variable so any mention of `ScalaRunTime` can direct to your version of `ScalaRunTime`, updating it for the entire stack :)

Here's how: http://stackoverflow.com/questions/3301635/change-private-static-final-field-using-java-reflection

fommil:
That'll end up being equivalent to a javaagent, but yeah, cool idea! I hadn't thought to do it that way!

fommil:
Jason pointed to some perftest ideas in https://gitter.im/scala/contributors?at=580bdae956121b9c7eae3a1b

Ichoran:
Do you have data indicating that negative zero is a common double value?

I just checked on a data processing application, and there the top values are 0, Double.NaN, and 1 respectively.  I am pretty sure that this patch will slow things down way too much; even with correct branch prediction, putting most accesses behind four conditionals is dubious.

Also, the GC churn may be important, but the performance is really not very good.  Taking arrays like this:

``` scala
val a = Array.fill(10000)(util.Random.nextDouble * util.Random.nextInt(2))
```

and the following three reference implementations for boxing:

``` scala
object D {
    private[this] final val DOUBLE_ZERO = java.lang.Double.valueOf(0.0);
    private[this] final val DOUBLE_ZERO_BITS = java.lang.Double.doubleToLongBits(0.0);
    private[this] final val DOUBLE_NEG_ZERO = java.lang.Double.valueOf(-0.0);
    private[this] final val DOUBLE_NEG_ZERO_BITS = java.lang.Double.doubleToLongBits(-0.0);
    private[this] final val DOUBLE_UNIT = java.lang.Double.valueOf(1.0);
    private[this] final val DOUBLE_UNIT_BITS = java.lang.Double.doubleToLongBits(1.0);
    private[this] final val DOUBLE_NEG_UNIT = java.lang.Double.valueOf(-1.0);
    private[this] final val DOUBLE_NEG_UNIT_BITS = java.lang.Double.doubleToLongBits(-1.0);

    def box(d: Double): java.lang.Double = {
        val bits = java.lang.Double.doubleToLongBits(d)
        if (bits == DOUBLE_ZERO_BITS) DOUBLE_ZERO
        else if (bits == DOUBLE_NEG_ZERO_BITS) DOUBLE_NEG_ZERO
        else if (bits == DOUBLE_UNIT_BITS) DOUBLE_UNIT
        else if (bits == DOUBLE_NEG_UNIT_BITS) DOUBLE_NEG_UNIT
        else java.lang.Double.valueOf(d)
    }
}

object Z {
  private[this] final val zero = java.lang.Double.valueOf(0.0);
  def box(d: Double) = if (java.lang.Double.doubleToLongBits(d) == 0) zero else java.lang.Double.valueOf(d)
}

object A {
  def box(d: Double) = java.lang.Double.valueOf(d)
}
```

This is just a disaster on a straight array-to-boxed-array conversion.  Z doesn't break even until nearly 90% of the copied elements are zero.  D requires even a bit more than 90%.  If there are are some elements, you suffer huge branch penalties (25% slowdown for Z, 50% for D).  If there are really none, though, branch prediction takes over and the penalty is small again (1% for Z, 10% for D).

Anyway, I assume the reason this isn't cached by the JVM is because it's a bad idea, at least in my preliminary testing.

mpociecha:
Could you include this description in the commit message and squash commits? (I'm afraid that just `respond to review` is not very valuable when in the future someone will need to read the history of changes)

fommil:
expanding on @Ichoran's performance test, here is an executable version

``` scala
import scala.util.Random

trait Box {
  def box(d: Double): java.lang.Double
}

// the current impl
object A extends Box {
  override def box(d: Double) = java.lang.Double.valueOf(d)
}

// just zero
object B extends Box {
  private[this] final val zero = java.lang.Double.valueOf(0.0);
  override def box(d: Double) = if (java.lang.Double.doubleToLongBits(d) == 0) zero else java.lang.Double.valueOf(d)
}

// proposed approach (0.0, -0.0, 1.0, -1.0)
object C extends Box {
  private[this] final val DOUBLE_ZERO = java.lang.Double.valueOf(0.0);
  private[this] final val DOUBLE_ZERO_BITS = java.lang.Double.doubleToLongBits(0.0);
  private[this] final val DOUBLE_NEG_ZERO = java.lang.Double.valueOf(-0.0);
  private[this] final val DOUBLE_NEG_ZERO_BITS = java.lang.Double.doubleToLongBits(-0.0);
  private[this] final val DOUBLE_UNIT = java.lang.Double.valueOf(1.0);
  private[this] final val DOUBLE_UNIT_BITS = java.lang.Double.doubleToLongBits(1.0);
  private[this] final val DOUBLE_NEG_UNIT = java.lang.Double.valueOf(-1.0);
  private[this] final val DOUBLE_NEG_UNIT_BITS = java.lang.Double.doubleToLongBits(-1.0);

  def box(d: Double): java.lang.Double = {
    val bits = java.lang.Double.doubleToLongBits(d)
    if (bits == DOUBLE_ZERO_BITS) DOUBLE_ZERO
    else if (bits == DOUBLE_NEG_ZERO_BITS) DOUBLE_NEG_ZERO
    else if (bits == DOUBLE_UNIT_BITS) DOUBLE_UNIT
    else if (bits == DOUBLE_NEG_UNIT_BITS) DOUBLE_NEG_UNIT
    else java.lang.Double.valueOf(d)
  }
}

object D extends Box {
  private[this] final val DOUBLE_ZERO = java.lang.Double.valueOf(0.0);
  private[this] final val DOUBLE_ZERO_BITS = java.lang.Double.doubleToLongBits(0.0);
  private[this] final val DOUBLE_NEG_ZERO = java.lang.Double.valueOf(-0.0);
  private[this] final val DOUBLE_NEG_ZERO_BITS = java.lang.Double.doubleToLongBits(-0.0);

  def box(d: Double): java.lang.Double = {
    if (d == 0.0) {
      if (java.lang.Double.doubleToLongBits(d) == DOUBLE_ZERO_BITS) DOUBLE_ZERO
      else DOUBLE_NEG_ZERO
    } else {
      java.lang.Double.valueOf(d)
    }
  }
}

object Perf {

  def timedArrayBox(name: String, impl: Box, array: Array[Double]): Unit = {
    System.gc(); System.gc(); System.gc()
    val start = System.currentTimeMillis()
    var i = 0
    while (i < array.length) {
      impl.box(array(i))
      i += 1
    }
    val end = System.currentTimeMillis()
    println(s"$name took ${end - start} millis")
  }

    println("creating arrays...")
    val size = 10000000
    val zeros = Array.fill(size)(0.0)
    val constant = Array.fill(size)(13.0)
    val random = Array.fill(size)(Random.nextDouble * Random.nextInt(2))
    println("... done")

    (1 to 10).foreach { _ =>
      List(A, B, C, D).foreach { impl =>
        val name = impl.getClass.getSimpleName
        timedArrayBox(s"$name - zeros", impl, zeros)
        timedArrayBox(s"$name - constant", impl, constant)
        timedArrayBox(s"$name - random", impl, random)
      }
    }

  }
}
```

on the 10th run, these are the results:

```
[info] A$ - zeros took 50 millis
[info] A$ - constant took 52 millis
[info] A$ - random took 51 millis
[info] B$ - zeros took 42 millis
[info] B$ - constant took 53 millis
[info] B$ - random took 92 millis
[info] C$ - zeros took 47 millis
[info] C$ - constant took 60 millis
[info] C$ - random took 104 millis
[info] D$ - zeros took 43 millis
[info] D$ - constant took 51 millis
[info] D$ - random took 96 millis
```

so it looks like the random case is about half the speed, but it's important to note that this is a reduction of 50millis for 10,000,000 elements... so the question is if people are willing to tolerate that in the typical app for the heap saving benefits it brings.

fommil:
@mpociecha github repo owners are fully in control of this https://github.com/blog/2141-squash-your-commits

mpociecha:
@fommil: Ha, nice one!

But anyway, if nothing changed in this repo, it means that there has to be run the full, passing build (including partest) for each of your commits, I think.

SethTisue:
@fommil as a rule, I don't trust results generated by hand-rolled benchmarking code. may I suggest using jmh? the scala repo has infrastructure for this in the `test/benchmarks` directory

fommil:
@SethTisue I might look into that out of interest, but this PR was primarily about me feeling my way around hacking on scala in ensime and making sure it all worked for scala contributors.

I ended up spending all of Saturday bringing `sbt-ensime` up to speed see http://ensime.github.io/build_tools/sbt/#customise :smile: I think I'd need to collaborate with somebody who understands the scala sbt build to be able to get the REPL (and assembly jars) working. The problem is that `sbt-ensime` needs to resolve the project's version of `scala-compiler` to jar files, but that's causing a cycle. Who'd be best to talk to about that?

The PR itself is a bit of a side effect and sounds too contentious anyway to be accepted. I'm already working on a different workaround.

fommil:
seems there are a few other places where boxing can happen, e.g. https://github.com/scala/scala/blob/v2.11.8/src/library/scala/Predef.scala#L356

Ichoran:
Since many simple operations on doubles (min, max, *, +, etc.) run in less than 1 ns, 5 ns extra (doubling the cost of the boxing) is pretty bad.

fommil:
too niche :smile: but could somebody please explain why the builds all failed?

lrytz:
@fommil the failures are due to a problem that shows up from time to time on 2.11.x, we haven't managed to nail it down yet: https://github.com/scala/scala-dev/issues/251

SethTisue:
> Who'd be best to talk to about that?

@fommil I'd suggest posting to scala-internals

PR #5477, opened 2016-10-24
Pattern Matching analysis improvements [Backport] #4919 [ci: last-only]

adriaanm:
Backport #4919, review by @retronym 

Straightforward rebase, dropping `test/junit/scala/tools/nsc/transform/patmat/PatmatBytecodeTest.scala`

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
+++ b/src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
@@ -177,6 +177,8 @@ trait TreeAndTypeAnalysis extends Debugging {
               filterChildren(subclasses)
             })
           }
+        case sym if sym.isCase if settings.isScala212 =>
+          List(List(tp))
 
         case sym =>
           debug.patmat("enum unsealed "+ ((tp, sym, sym.isSealed, isPrimitiveValueClass(sym))))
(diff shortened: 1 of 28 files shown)


retronym:
What's the motivation for the backport?

adriaanm:
A support ticket.

retronym:
> Backport note: unclear why I had to update test/files/jvm/future-spec

`FutureTest` in 2.12.x had the `@unchecked` annotation added in https://github.com/scala/scala/commit/da46355a#diff-e09a1fb00b466cd8d3385c8ebfb011d9L241, when that part of it was forward merged from 2.11.

adriaanm:
Thanks. My comments was a bit vague and now also outdated. I initially also updated the check file because on my machine I saw three deprecation warnings instead of one. On the CI it's still one.

src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
```diff
@@ -173,6 +177,8 @@ trait TreeAndTypeAnalysis extends Debugging {
               filterChildren(subclasses)
             })
           }
+        case sym if sym.isCase =>
+          List(List(tp))
```
retronym, line 181: I wonder if this part should be `-Xsource:2.12`

adriaanm: Done. I wonder if we should cache the 211/212 version settings.

adriaanm:
I'm going to back out of this backport, since there were some performance regressions in the 2.12 fixes I backported that we haven't resolved yet.

lloydmeta:
Just ran into something that I think this would have fixed in a Play project (same-named case accessors leading to non-exhaustive match warning). Would be nice to get this eventually.

PR #5479, opened 2016-10-25
Workaround in backend for linkedClassOfClass returning a type alias

lrytz:
Fixes scala/scala-dev/issues/248

This fix is the a workaround and does not fix the underlying issue.
The change propsed here is minimal and safe.

In principle, linkedClassOfClass should not return a type alias. This
could be easily achieved by adding an additional filter to
Symbol.companionClass, as proposed in a comment on the ticket. However,
I observed some further complex interactions in the classfile parser /
unpickler when having type aliases shadowing companions. This behavior
might be cause by the proposed fix. In any case, a change to
companionClass seems rather fundamental and needs to be carefully
investigated.


lrytz:
superseded by #5482

PR #5480, opened 2016-10-26
SI-6412 remove flaky test

SethTisue:
I have repeatedly seen this fail CI runs, including recently

as the comment in the test itself says: "I'm not sure this is a great
way to test for memory leaks, since we're also testing how good
the JVM's GC is, and this is not easily reproduced between
machines/over time"

review by @adriaanm


SethTisue:
note, this leaves just two tests that still use `partest.MemoryTest`:

```
test/files/run/reflection-mem-glbs.scala
test/files/run/reflection-mem-tags.scala
```

bit tempted to just toss all three of them. what do you think?

an alternative would be to try to make MemoryTest more robust. maybe sprinkle some additional `System.gc()`s and `Thread.sleep`s in there, who knows :-\

dragos:
I think the `MemoryTest` class should calculate a "slope" for memory usage and test against that. This one in [scala-ide](https://github.com/scala-ide/scala-ide/blob/6fd4dce1b90429a08d48fb5405e0e31e8817b607/org.scala-ide.sdt.core.tests/src/org/scalaide/core/leaks/MemoryLeaksTest.scala) was remarkably stable and at the same time caught some regressions.

szeiger (approved):

SethTisue:
Iulian, I have logged your suggestion as https://issues.scala-lang.org/browse/SI-10076

PR #5481, opened 2016-10-26
SI-10007 sys.process thread sync

som-snytt:
A previous change to replace `SyncVar.set` with `SyncVar.put`
breaks things.

This commit tweaks the thread synchronizing in `sys.process`
to actually use `SyncVar` to sync and pass a var.

Joining the thread about to exit is superfluous.

A result is put exactly once, and consumers use
non-destructive `get`.

Note that as usual, avoid kicking off threads in a static
context, since class loading cycles are somewhat dicier
with 2.12 lambdas. In particular, REPL is a static context
by default.

### DIFF
--- a/src/library/scala/sys/process/ProcessImpl.scala
+++ b/src/library/scala/sys/process/ProcessImpl.scala
@@ -86,17 +86,20 @@ private[process] trait ProcessImpl {
   private[process] abstract class CompoundProcess extends BasicProcess {
     def isAlive()   = processThread.isAlive()
     def destroy()   = destroyer()
-    def exitValue() = getExitValue._2() getOrElse scala.sys.error("No exit code: process destroyed.")
-    def start()     = getExitValue
+    def exitValue() = futureValue() getOrElse scala.sys.error("No exit code: process destroyed.")
+    def start()     = { futureThread ;() }
 
-    protected lazy val (processThread, getExitValue, destroyer) = {
+    protected lazy val (processThread, (futureThread, futureValue), destroyer) = {
       val code = new SyncVar[Option[Int]]()
-      code.put(None)
-      val thread = Spawn(code.put(runAndExitValue()))
+      val thread = Spawn {
+        var value: Option[Int] = None
+        try value = runAndExitValue()
+        finally code.put(value)
+      }
 
       (
         thread,
-        Future { thread.join(); code.get },
+        Future(code.get),          // thread.join()
         () => thread.interrupt()
       )
     }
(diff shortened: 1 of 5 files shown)


ghik (commented):
There are too many zeros in test file name :)

som-snytt:
Rebased and changed test file name to scientific notation. :)

test/junit/scala/sys/process/t10007.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.junit.Test
+import org.junit.Assert.assertEquals
+
+@RunWith(classOf[JUnit4])
+class ProcessTest {
+  @Test def t10007(): Unit = {
+    val res = ("cat" #< new ByteArrayInputStream("lol".getBytes)).!!
+    assertEquals("lol\n", res)
```
som-snytt: Test won't fly on windows, sorry, @SethTisue

lrytz:
LGTM - except for that test on windows.. should we just guard it with `System.getProperty("os.name").toLowerCase.contains("windows")`?

som-snytt:
I added the SI-10055 test because nothing good happens after midnight. It doesn't really add anything but time. Or I guess it tests the lambda initialization. The call-by-name arg could be invoked anywhere, including a static init where it could block. Sorry, Jenkins, I was too sleepy to compile and test before bed, would you do that, there's a good chap.

src/library/scala/sys/process/ProcessImpl.scala
```diff
@@ -91,12 +91,11 @@ private[process] trait ProcessImpl {
 
     protected lazy val (processThread, getExitValue, destroyer) = {
       val code = new SyncVar[Option[Int]]()
-      code.put(None)
       val thread = Spawn(code.put(runAndExitValue()))
```
som-snytt: Should this be try-finally like at ProcessBuilderImpl L61?

lrytz:
LGTM, thanks for these cleanups!

lrytz:
Not sure if `t9752.scala` is a spurious failure.

I just saw this now: could you rename `t10007.scala` to `ProcessTest.scala`?

som-snytt:
Hopefully spurious. I renamed both those junit files and squashed.

som-snytt:
Test rig is failing badly. I won't rebuild at this time.

adriaanm:
/rebuild

The OOMpas are at it again. Rig is feeling fresher now, it looks like.

lrytz:
Thanks, @som-snytt

PR #2963, opened 2013-09-18
Small refactorings giving hooks for Scala.js

sjrd:
I have reviewed the ugliest parts of Scala.js, where I need to circumvent the lack of appropriate hooks in scalac, and I came up with these three small refactorings.

Each one of these has its corresponding simplifying commit in Scala.js, in the following comparison:
https://github.com/sjrd/scala-js/compare/scala-2.11...topic/what-if-scalac-was-nicer-to-us
(in the same order).

There is no test accompanying this pull request, because there is absolutely no functionality change (it's only refactoring) and I don't quite see how to test that with partest. I'll be glad to add tests if someone can hint at me how.

(In case you care:) Incidentally, these hooks are the only ones on which Scala.js relies, in addition to being able to manipulate the phase set. Which means that _if_ they were (experimentally) overridable by plugins, then Scala.js could be implemented as a plugin, which would bring lots of advantages in terms of integration with our various tools (sbt and IDE mainly).

(There are still two files with a non-negligible amount of code duplication in Scala.js, [JSSymbolLoaders.scala](https://github.com/sjrd/scala-js/blob/topic/what-if-scalac-was-nicer-to-us/compiler/src/main/scala/scala/tools/nsc/scalajs/JSSymbolLoaders.scala) and [JSTypefileParser.scala](https://github.com/sjrd/scala-js/blob/topic/what-if-scalac-was-nicer-to-us/compiler/src/main/scala/scala/tools/nsc/scalajs/JSTypefileParser.scala), but a proper fix to these two would require much bigger refactorings in scalac; and I don't care so much about these because they are not specifically the hooks I am relying on.)

PR #5482, opened 2016-10-26
fix backend crash: Cannot create ClassBType from non-class symbol; also fix SI-7139

lrytz:
Frontend fixes for scala/scala-dev#248 

See individual commit messages.


retronym:
Awesome, this looks to fix https://issues.scala-lang.org/browse/SI-7139!

```
% cat sandbox/test.scala
package test {
  object A
}
package object test {
  type A = Int
}

% scalac sandbox/test.scala && printf ':silent\ndef bug = null.asInstanceOf[test.A.type]\ndef bug = null.asInstanceOf[test.A.type]\n' | scala
Welcome to Scala 2.11.8 (OpenJDK 64-Bit Server VM, Java 1.8.0_112-release).

scala> :silent

scala> def bug = null.asInstanceOf[test.A.type]

scala> def bug = null.asInstanceOf[test.A.type]
<console>:11: error: type mismatch;
 found   : test.A.type
 required: AnyRef
Note that A extends Any, not AnyRef.
Such types can participate in value classes, but instances
cannot appear in singleton types or in reference comparisons.
       def bug = null.asInstanceOf[test.A.type]
                                        ^

scala> :quit

% qscalac sandbox/test.scala && printf ':silent\ndef bug = null.asInstanceOf[test.A.type]\ndef bug = null.asInstanceOf[test.A.type]\n' | qscala
Welcome to Scala 2.12.0-20161026-205457-37b3df3 (OpenJDK 64-Bit Server VM, Java 1.8.0_112-release).

scala> :silent

scala> def bug = null.asInstanceOf[test.A.type]

scala> def bug = null.asInstanceOf[test.A.type]

scala> :quit
```

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -38,10 +38,6 @@ abstract class UnPickler {
       assert(classRoot != NoSymbol && moduleRoot != NoSymbol, s"The Unpickler expects a class and module symbol: $classRoot - $moduleRoot")
       new Scan(bytes, offset, classRoot, moduleRoot, filename).run()
     } catch {
-      case ex: IOException =>
-        throw ex
-      case ex: MissingRequirementError =>
-        throw ex
       case ex: Throwable =>
```
soc: Maybe `ex: NonFatal` instead?

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -356,63 +356,40 @@ abstract class ClassfileParser {
...
-      owner = sym.moduleClass
-      start = end + 1
-      end = name.indexOf('.', start)
-    }
-    ss = name.subName(0, start)
-    owner.info.decls lookup ss orElse {
-      sym = owner.newClass(ss.toTypeName) setInfoAndEnter completer
-      debuglog("loaded "+sym+" from file "+file)
-      sym
-    }
```
soc, line 396: This is a nice cleanup! :-)

lrytz:
added a test case for SI-7139. all commits are green, so review by @retronym. maybe also @adriaanm, as this touches some low-level internals.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
lrytz: that's reasonable, will do

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -162,17 +173,10 @@ abstract class SymbolLoaders {
...
-        enterClassAndModule(owner, classRep.name, newClassLoader(bin))
+        enterClassAndModule(owner, classRep.name, new ClassfileLoader(bin, _, _))
     }
   }
 
-  /** Create a new loader from a binary classfile.
-   *  This is intended as a hook allowing to support loading symbols from
-   *  files other than .class files.
-   */
-  protected def newClassLoader(bin: AbstractFile): SymbolLoader =
```
adriaanm, line 173: I verified this hook (re-introduced by #2963) is no longer used by scala-js (since https://github.com/scala-js/scala-js/commit/8d85779)

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -441,13 +416,10 @@ abstract class ClassfileParser {
     }
 
     val isTopLevel = !(currentClass containsChar '$') // Java class name; *don't* try to to use Scala name decoding (SI-7532)
-
-    val c = if (isTopLevel) pool.getClassSymbol(nameIdx) else clazz
     if (isTopLevel) {
-      if (c != clazz) {
-        if ((clazz eq NoSymbol) && (c ne NoSymbol)) clazz = c
```
adriaanm, line 448: Is there any scenario in e.g. presentation/interactive mode where we needed this? /cc @dragos

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -52,20 +52,28 @@ abstract class SymbolLoaders {
...
-    val clazz = owner.newClass(newTypeName(name))
+  def enterClass(owner: Symbol, name: String, completer: SymbolLoader): Symbol =
+    enterClass(owner, newClass(owner, name), completer)
+
+  def enterClass(owner: Symbol, clazz: Symbol, completer: SymbolLoader): Symbol = {
     clazz setInfo completer
     enterIfNew(owner, clazz, completer)
   }
 
+  def newModule(owner: Symbol, name: String): Symbol = owner.newModule(newTermName(name))
```
adriaanm: `: ModuleSymbol` as the result? (Allows tightening `enterClassAndModule` and `ClassfileLoader`).

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -113,9 +121,12 @@ abstract class SymbolLoaders {
   /** Enter class and module with given `name` into scope of `root`
    *  and give them `completer` as type.
    */
-  def enterClassAndModule(root: Symbol, name: String, completer: SymbolLoader) {
-    val clazz = enterClass(root, name, completer)
-    val module = enterModule(root, name, completer)
+  def enterClassAndModule(root: Symbol, name: String, getCompleter: (Symbol, Symbol) => SymbolLoader) {
```
adriaanm: `(Symbol, ModuleSymbol) => SymbolLoader` seems within reach to avoid confusion in the future.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -277,7 +281,7 @@ abstract class SymbolLoaders {
     }
   }
 
-  class ClassfileLoader(val classfile: AbstractFile) extends SymbolLoader with FlagAssigningCompleter {
+  class ClassfileLoader(val classfile: AbstractFile, clazz: Symbol, module: Symbol) extends SymbolLoader with FlagAssigningCompleter {
```
adriaanm: `module: ModuleSymbol`

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -132,17 +132,12 @@ abstract class ClassfileParser {
     finally loaders.parentsLevel -= 1
   }
 
-  def parse(file: AbstractFile, root: Symbol): Unit = {
-    debuglog("[class] >> " + root.fullName)
-
+  def parse(file: AbstractFile, clazz: Symbol, module: Symbol): Unit = {
```
adriaanm: `module: ModuleSymbol`

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
adriaanm: Implemented in https://github.com/scala/scala/commit/5c7b8ede926c6157ae9853316f1d4c90859985ee

adriaanm (changes requested):
Really good stuff! I have a few small suggestions for locking this down a bit more while we're at it.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -309,7 +313,7 @@ abstract class SymbolLoaders {
       // errors. More concretely, the classfile parser calls "sym.companionModule", which calls
       // "isModuleNotMethod" on the companion. After refchecks, this method forces the info, which
       // may run the classfile parser. This produces the error.
-      enteringPhase(phaseBeforeRefchecks)(classfileParser.parse(classfile, root))
+      enteringPhase(phaseBeforeRefchecks)(classfileParser.parse(classfile, clazz, module))
```
dragos: The comment seems outdated, the classfile parser doesn't call `companionModule` anymore. Maybe there's no need to call `enteringPhase` either?

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
dragos: This allows callers to pass unrelated symbols. Maybe some warning in a comment is in order (I assume calls to `companionModule` & friends are also "forbidden" inside this method). If so, also worth a comment.

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
dragos: No, I don't think so. This looks more like defensive coding.  I don't know if there's any scenario in which `pool.getClassSymbol` returns `NoSymbol`, but that seems to be the only purpose of this call. Maybe the new logic involving `StubSymbol` subsumes it?

lrytz: With this PR, clazz cannot ever be NoSymbol.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
lrytz: Hmm.. I think it's a bit risky to remove the phase travel here, the classfile parser and unpickler call various methods on symbols. I'll update the comment.

lrytz: Though I see it was added not too long ago: 0ccdb15

lrytz: I'll give it a try.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
lrytz: ok, done.

adriaanm (approved):

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
adriaanm: Ok, thanks.

adriaanm:
LGTM!

PR #5486, opened 2016-10-27
SI-6734 Synthesize companion near case class

som-snytt:
Tweak the "should I synthesize now" test for
case modules, so that the tree is inserted in
the same tree as the case class.

This for the case where there are multiple `package p { }` trees, which happens implicitly when there is a `package object p { }`.

No attempt made in this quick fix to improve locality-based synthesis.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3140,13 +3140,14 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           // SI-5877 The decls of a package include decls of the package object. But we don't want to add
           //         the corresponding synthetics to the package class, only to the package object class.
           // SI-6734 Locality test below is meaningless if we're not even in the correct tree.
+          //         For modules that are synthetic case companions, check that case class is defined here.
           def shouldAdd(sym: Symbol): Boolean = {
             def shouldAddAsModule: Boolean =
               sym.moduleClass.attachments.get[ClassForCaseCompanionAttachment] match {
                 case Some(att) =>
                   val cdef = att.caseClass
                   stats.exists {
-                    case t @ ClassDef(_, _, _, _) => t.symbol == cdef.symbol
+                    case t @ ClassDef(_, _, _, _) => t.symbol == cdef.symbol   // cdef ne t
                     case _ => false
                   }
                 case _ => true
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3139,10 +3139,24 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           val initElems = scope.elems
           // SI-5877 The decls of a package include decls of the package object. But we don't want to add
           //         the corresponding synthetics to the package class, only to the package object class.
-          def shouldAdd(sym: Symbol) =
-            inBlock || !context.isInPackageObject(sym, context.owner)
+          // SI-6734 Locality test below is meaningless if we're not even in the correct tree.
+          def shouldAdd(sym: Symbol): Boolean = {
+            def shouldAddAsModule: Boolean =
+              sym.moduleClass.attachments.get[ClassForCaseCompanionAttachment] match {
+                case Some(att) =>
+                  val cdef = att.caseClass
+                  stats.exists {
+                    case t @ ClassDef(_, _, _, _) => t.symbol == cdef.symbol
+                    case _ => false
+                  }
+                case _ => true
+              }
+
+            (!sym.isModule || shouldAddAsModule) && (inBlock || !context.isInPackageObject(sym, context.owner))
+          }
           for (sym <- scope)
-            for (tree <- context.unit.synthetics get sym if shouldAdd(sym)) { // OPT: shouldAdd is usually true. Call it here, rather than in the outer loop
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3139,10 +3139,24 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
           //         the corresponding synthetics to the package class, only to the package object class.
-          def shouldAdd(sym: Symbol) =
-            inBlock || !context.isInPackageObject(sym, context.owner)
+          // SI-6734 Locality test below is meaningless if we're not even in the correct tree.
+          def shouldAdd(sym: Symbol): Boolean = {
+            def shouldAddAsModule: Boolean =
+              sym.moduleClass.attachments.get[ClassForCaseCompanionAttachment] match {
+                case Some(att) =>
+                  val cdef = att.caseClass
+                  stats.exists {
```
dragos, line 3149: Doesn't this introduce quadratic complexity in the size of the package? This method is called for every member in scope, and the test searches linearly to find a `ClassDef` tree.

I'm also not sure if assuming source code is going to work always. What if the case class in scope comes from binaries?

som-snytt: Only invoked for synthetic modules of case classes with attachment introduced by Namers.

dragos: This code still looks strange to me. The attachment **is** a `ClassDef`, but the code does a linear search for a(nother?) `ClassDef` with the same symbol. Isn't it the same `ClassDef`? If not, why do they have the same symbol? In either case it might be good to document it.

This may be perfectly ok, but it's still not clear to me why it only treats symbols that are currently compiled (that have trees around).

som-snytt: Currently, the moduleclass has attachment pointing to a classdef; but meanwhile the in-flight classdef is something transformed, so `eq` fails. The test is to say, add synthetic companion only if the associated classdef is in these statements, too. (The big loop also tries to match associated synths in an ad-hoc way.)

I'll try a bigger clean-up, where the classdef bears an attachment (to mean it has an associated synthetic; similarly for other defs). Then the big loop should just traverse looking for things with synths to add; and then at the end, add whatever remaining synths belong here.

Maybe the attachment has a `Tree => Boolean` to match something invariant that says, yes, this tree is associated with me.

som-snytt: That would be useful for a unit with a huge number of case classes, since right now each tree is checked against all (remaining) synths. The check itself must always do work, as noted.

dragos: Thanks for explaining, I start to understand a bit more what's happening. If I understand correctly, this PR tightens the conditions under which such synthetics are generated, presumably making it "deterministic" -- the same place where the ClassDef is. But where were they added before this PR? Maybe there's an equivalent condition that doesn't require iterating?

I agree the code won't run for all stats in a package, but the typer is generally "hot", and even if this doesn't seem much, it's still going to be doing more work than before, work that's not a constant but a function of code size.

som-snytt: See the PR comment and test case -- they'd be added to the package tree containing the package object if that was first in source-text-order.

dragos: Yeah, I stared at that test case long and hard. It's impenetrable. Thanks again for explaining. I'd really add some explanation to the test

test/files/pos/t6734.scala
```diff
@@ -0,0 +1,23 @@
+
+//single file badimp.scala
+// adding package object gives not found: type SortedMap
+package object badimp
```
dragos: How about:

``` scala
// this package object used to contain the synthetic companion object of `Nodal` and fail because the 
// necessary import is not in this scope
```

test/files/pos/t6734.scala
```diff
@@ -0,0 +1,23 @@
...
+
+package badimp {
+
+  // move before package object works
```
dragos: This comment is more confusing than clarifying. This test case is the way it is now, and the ticket you link to gives the background and additional details.

test/files/pos/t6734.scala
```diff
@@ -0,0 +1,23 @@
...
+  import scala.collection.immutable.SortedMap
+
+  case class Nodal private[badimp] (value: String, children: SortedMap[String, Int])
+
+  // adding target object restores sanity
+  // but adding it before the import does not
+  //object Nodal
```
dragos: I would remove this part as well since the ticket describes all these variations.

dragos (dismissed):

src/compiler/scala/tools/nsc/typechecker/Typers.scala
som-snytt: I was just using the test case the guy submitted on the ticket. I agree that many test cases make you wonder what's being tested and require archaeology. I'll add a comment benefiting from hindsight.

test/files/pos/t6734.scala
som-snytt: The funny thing is that it's not the package object, but the package introduced by desugaring that received the synthetic.

retronym:
LGTM

dragos:
Sorry for missing this one, thanks for closing it @retronym

PR #5487, opened 2016-10-28
SI-3236 constant types for literal final static java fields

lrytz:


lrytz:
As mentioned in the original PR, this could be expanded to support (some) composite expressions, but it would be a bigger effort. For that, instead of trying to constant fold and assign types in the parser, I think we should just parse the expression and pass it on to the Scala type checker. But that's also not easy by any means :)

lrytz:
I think the code here is an OK compromise, even with its limitations. For example, it treats `!1` as `-1`. But the Java compiler would reject that of course.

lrytz:
@jrudolph please take a look.

jrudolph:
On first glance, I think it looks good :+1: 

Would it make sense to add tests for the additional feature of implicitly converting the constants?

lrytz:
yes, also some neg tests to make sure the limits are checked. will do.

adriaanm:
One last test failure: test/files/run/t3236 doesn't compile.

lrytz:
Added a neg test to frame the limitations

lrytz:
review by @adriaanm

jrudolph:
:+1: new tests LGTM

adriaanm (approved):
LGTM

adriaanm:
Thanks!

PR #5489, opened 2016-10-30
SI-9904 Add size/isEmpty/nonEmpty to VectorBuilder

slothspot:


Ichoran:
Is there a reason why `VectorBuilder` as opposed to all the other Builders should have these methods?

slothspot:
Originally this PR is to address https://issues.scala-lang.org/browse/SI-9904. If it's reasonable, this extension may be propagated to other builders as well.

Ichoran:
It's reasonable to consider something like this but less reasonable to do it piecemeal.  Anyway, 2.12 is out and this would not be binary compatible, so it can't be as a patch against 2.12, it'd have to target 2.13.

slothspot:
I see no 2.13.x branch to base my PR on it. So how could I target 2.13?

SethTisue:
we'll branch for 2.13.x soon, once 2.12.0 is fully out

slothspot:
@SethTisue updated to target 2.13.x

SethTisue:
All collections stuff is likely to get overhauled as part of the https://github.com/scala/collection-strawman, so at the moment, I don't think it would make sense to accept this (or to leave it in the queue). We can always revisit/reopen later.

PR #4832, opened 2015-11-01
Make scala.math compile under Dotty

odersky:
Need to change unrelated inner classes so that they
have different names. Dotty does not allow a type
"not-overriding" a class with same name in a basetype.

PR #5490, opened 2016-10-30
SI-9705 Add <,<=,>,>= to Numeric.Ops

slothspot:


slothspot:
@SethTisue updated to target 2.13.x

SethTisue:
is there a chance of conflicting implicits with `Ordered` and/or `Ordering`?

soc:
I think it might be good idea in general, but I'd rather wait until we know what's the deal with contravariance and type classes in the future before exposing things further. Also, what's the effect on floating point numbers here (which use PartialOrdering as far as I remember)?

retronym:
I think the current design is to expose these via [`Ordering.Implicits.infixOrderingOps`](https://github.com/scala/scala/blob/ca606e6943c405e487c54c3fd14a3a8ac917bb2e/src/library/scala/math/Ordering.scala#L192-L192):

``` scala
Welcome to Scala 2.11.8 (OpenJDK 64-Bit Server VM, Java 1.8.0_112-release).

scala> new java.math.BigDecimal("1") <= new java.math.BigDecimal("2")
<console>:12: error: value <= is not a member of java.math.BigDecimal
       new java.math.BigDecimal("1") <= new java.math.BigDecimal("2")
                                     ^

scala> import scala.math.Ordering.Implicits._
import scala.math.Ordering.Implicits._

scala> new java.math.BigDecimal("1") <= new java.math.BigDecimal("2")
res4: Boolean = true
```

Or via [Ordering#mkOrderingOps](https://github.com/scala/scala/blob/ca606e6943c405e487c54c3fd14a3a8ac917bb2e/src/library/scala/math/Ordering.scala#L139-L139)

``` scala
scala> def foo[T](t1: T, t2: T)(implicit ordering: Ordering[T]) = { t1 <= t2 }
<console>:11: error: value <= is not a member of type parameter T
       def foo[T](t1: T, t2: T)(implicit ordering: Ordering[T]) = { t1 <= t2 }
                                                                       ^

scala> def foo[T](t1: T, t2: T)(implicit ordering: Ordering[T]) = { import ordering._; t1 <= t2 }
foo: [T](t1: T, t2: T)(implicit ordering: Ordering[T])Boolean
```

I'm fairly hesitant to change this part of the library.

retronym:
See also #4832 and #5305 for (incomplete) recent attempts to fix a problem in the design of this part of the library.

SethTisue:
closing for inactivity.

@slothspot this could still perhaps move forward, but we'd need your response on the review feedback received so far.

PR #5496, opened 2016-11-01
Merge 2.12.0, starr=2.12.0, re-outsource scalacheck

adriaanm:
After this, the v2.12.0 tag will be in 2.12.x's history and the 2.12.0 branch can be deleted (and the 2.13.x one created!)


SethTisue:
LGTM

PR #5497, opened 2016-11-02
Don’t include scala-asm.jar in scala-compiler.jar

szeiger:
Fixes https://github.com/scala/scala-dev/issues/254

### DIFF
--- a/project/Osgi.scala
+++ b/project/Osgi.scala
@@ -37,9 +37,9 @@ object Osgi {
     },
     jarlist := false,
     bundle := Def.task {
-      val res = (products in Compile in packageBin).value
-      bundleTask(headers.value.toMap, jarlist.value, (products in Compile in packageBin).value,
-        (artifactPath in (Compile, packageBin)).value, res, streams.value)
+      val cp = (products in Compile in packageBin).value
+      bundleTask(headers.value.toMap, jarlist.value, cp,
+        (artifactPath in (Compile, packageBin)).value, cp, streams.value)
     }.value,
     packagedArtifact in (Compile, packageBin) := (((artifact in (Compile, packageBin)).value, bundle.value)),
     // Also create OSGi source bundles:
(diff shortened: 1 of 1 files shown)


retronym:
@szeiger Alternatively: https://github.com/scala/scala/compare/4e467c1...retronym:review/5479

This would include `scala-asm.properties` as a resource within scala-compiler.jar, which seems in line with the inclusion of, e.g. `scaladoc.properties`.

szeiger:
@retronym Good point. I didn't consider `scala-asm.properties`. Updated and rebased.

retronym:
LGTM.  Looks like we should backport to 2.11.x, too.

retronym:
> Looks like we should backport to 2.11.x, too.

@szeiger I'll leave that to you. There is a subtle difference in the surrounding code in the 2.11.x branch that you might know more about. In 2803d0ad50e94a3a88d98d709b7a38a5bbb6a383, you're passing `Nil` rather than the value now know as `val cp` as the param `resourceDirectories`.

If thats a mistake, the backport would be:

``` patch
commit 40526b4af1c4be132536fc9b5213a6d60619de95
Author: Stefan Zeiger <szeiger@novocode.com>
Date:   Wed Nov 2 15:55:27 2016 +0100

    Don’t include scala-asm.jar in scala-compiler.jar

    Fixes https://github.com/scala/scala-dev/issues/254

diff --git a/project/Osgi.scala b/project/Osgi.scala
index c5d4734..2103534 100644
--- a/project/Osgi.scala
+++ b/project/Osgi.scala
@@ -39,4 +39,5 @@ object Osgi {
     bundle <<= Def.task {
-      bundleTask(headers.value.toMap, jarlist.value, (products in Compile in packageBin).value,
-        (artifactPath in (Compile, packageBin)).value, Nil, streams.value)
+      val cp = (products in Compile in packageBin).value
+      bundleTask(headers.value.toMap, jarlist.value, cp,
+        cp, Nil, streams.value)
     },
@@ -58,3 +59,8 @@ object Osgi {
     headers foreach { case (k, v) => builder.setProperty(k, v) }
-    val includeRes = resourceDirectories.filter(_.exists).map(_.getAbsolutePath).mkString(",")
+
+    // https://github.com/scala/scala-dev/issues/254
+    // Must be careful not to include scala-asm.jar within scala-compiler.jar!
+    def resourceDirectoryRef(f: File) = (if (f.isDirectory) "" else "@") + f.getAbsolutePath
+
+    val includeRes = resourceDirectories.filter(_.exists).map(resourceDirectoryRef).mkString(",")
     if(!includeRes.isEmpty) builder.setProperty(INCLUDERESOURCE, includeRes)

```

szeiger:
@retronym No need to backport, this works fine in 2.11.x. The switch to a newer version of bnd in 2.12.x required the separate resourceDirectories, which resulted in this bug.

PR #5498, opened 2016-11-03
SI-10023: third iteration using member.keyword and special handling for parameterless methods

esarbe:
I guess that the strings _probably_ shouldn't be hard coded. What would be a more appropriate way to do it?

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
+++ b/src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
@@ -152,8 +152,10 @@ trait MemberHandlers {
     override def definesValue = flattensToEmpty(member.vparamss) // true if 0-arity
     override def resultExtractionCode(req: Request) = {
       val nameString = colorName(name)
-      val typeString = colorType(req typeOf name)
-      if (mods.isPublic) s""" + "$nameString: $typeString\\n"""" else ""
+      val defType = req.typesOfDefinedTerms(name)
+      val typeString = colorType(s"${paramString(defType)}: ${defType.resultType}")
+
+      if (mods.isPublic) s""" + "${member.keyword} $nameString$typeString\\n"""" else ""
     }
   }
 
--- a/src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
+++ b/src/repl/scala/tools/nsc/interpreter/MemberHandlers.scala
@@ -153,7 +153,11 @@ trait MemberHandlers {
     override def resultExtractionCode(req: Request) = {
       val nameString = colorName(name)
       val defType = req.typesOfDefinedTerms(name)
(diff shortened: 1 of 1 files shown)


som-snytt:
Probably. https://github.com/scala/scala/pull/5222

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -2490,7 +2490,7 @@ trait Types
     //TODO this may be generalised so that the only constraint is dependencies are acyclic
     def approximate: MethodType = MethodType(params, resultApprox)
 
-    override def safeToString = paramString(this) + resultType
+    override def safeToString = paramString(this) + ": " + resultType
```
adriaanm: I would recommend restricting the changes for this PR to the interpreter, as this method in reflect should stay the same.

esarbe: Thanks for the feedback. I thought that it was too easy. ;)

esarbe:
This version has the problem of parameterless methods being printed as 

```
scala> def one = 1
def one(): Int
```

instead of (as it currently is)

```
scala> def one = 1
one: Int
```

although one could argue that the current output is also not consistent, since according to
http://www.scala-lang.org/files/archive/spec/2.11/03-types.html#method-types
it should actually be

```
scala> def one = 1
one: => Int
```

So, what do do? I'm loath to match on the type. Suggestions? I'm sure I shouldn't go as far as to introduce a `NullaryDefHandler`...

SethTisue:
I believe you're trying to match the language syntax, not the method type notation in the spec. Method types are a spec thing, not a language thing. So `def one: Int` (not `def one(): Int`) seems correct to me.

som-snytt:
If the feature tends toward minimally two modes, beginner and expert, or verbose and terse, I'd go for "type" mode myself:

```
scala> :type val x = 42
Int

scala> :type def f = 42
=> Int

scala> def f = 42
f: => Int                            // conforms with :type output, no val or def keyword
```

Another possibility is for verbose mode to be "source mode" or "print-tab mode":

```
scala> def f = 42
def f: Int = 42

scala> def f = 42
def f = 42 : Int
```

print-tab puts the ascribed type behind a line comment, but the idea is to highlight that this is the inferred type. It's easy to cut/paste the prefix of the output that excludes it, though maybe there's no use case for that, unless:

```
scala> def g(implicit s: String) = s
g: (implicit s: String)String  // old-style output

scala> def f(implicit s: String) = g
def f(implicit s: String) = g(s) : String    // may want to cut/paste expanded version?
```

Another idea is to have a special color for inferred types, and put them in usual source order.

esarbe:
@som-snytt Do you think I should create another SI for the beginner/expert switch? Or could this be addressed in the context of SI-10024?

esarbe:
This version also handles parameterless methods. I don't like the if/else, so I'd be glad for any pointer to get it done without.

PR #5501, opened 2016-11-04
avoid deprecated sbt 0.12 operators

SethTisue:


SethTisue:
@dwijnand @szeiger can you help finish this?

```
[warn] /Users/tisue/scala.212/project/Osgi.scala:44: `<<=` operator is deprecated. Use `key := { x.value }` or `key ~= (old => { newValue })`.
[warn] See http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html
[warn]     packagedArtifact in (Compile, packageBin) <<= (artifact in (Compile, packageBin), bundle).identityMap,
[warn]                                               ^
[warn] /Users/tisue/scala.212/project/Osgi.scala:44: method t2ToTable2 in object Scoped is deprecated: The sbt 0.10 style DSL is deprecated: '(k1, k2) map { (x, y) => ... }' should now be '{ val x = k1.value; val y = k2.value }'.
[warn] See http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html
[warn]     packagedArtifact in (Compile, packageBin) <<= (artifact in (Compile, packageBin), bundle).identityMap,
[warn]                                                   ^
[warn] two warnings found
/Users/tisue/scala.212/build.sbt:879: warning: `<<=` operator is deprecated. Use `key := { x.value }` or `key ~= (old => { newValue })`.
See http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html
    mkPack <<= Def.task { (buildDirectory in ThisBuild).value / "pack" } dependsOn (packagedArtifact in (Compile, packageBin), mkBin),
           ^
/Users/tisue/scala.212/build.sbt:107: warning: value compilerJar in class ScalaInstance is deprecated: Only `allJars` and `jars` can be reliably provided for modularized Scala.
      val s2 = new ScalaInstance(s.version, s.loader, s.libraryJar, s.compilerJar, s.extraJars, Some(s.actualVersion))
                                                                      ^
/Users/tisue/scala.212/build.sbt:107: warning: value extraJars in class ScalaInstance is deprecated: Only `allJars` and `jars` can be reliably provided for modularized Scala.
      val s2 = new ScalaInstance(s.version, s.loader, s.libraryJar, s.compilerJar, s.extraJars, Some(s.actualVersion))
                                                                                     ^
/Users/tisue/scala.212/build.sbt:759: warning: value compilerJar in class ScalaInstance is deprecated: Only `allJars` and `jars` can be reliably provided for modularized Scala.
      val jar = (scalaInstance in bootstrap).value.compilerJar
                                                   ^
```

SethTisue:
@dwijnand I added you to https://github.com/orgs/scala/teams/admins and checked the "Allow edits from maintainers" checkbox here

dwijnand:
I guess even Admin isn't sufficient..

```
20:04:15 $ git push SethTisue sbt-deprecations
fatal: remote error:
  You can't push to git://github.com/SethTisue/scala.git
  Use https://github.com/SethTisue/scala.git
```

Removed 2 deprecations in https://github.com/SethTisue/scala/compare/sbt-deprecations...dwijnand:sbt-deprecations

I'm not sure about `ScalaInstance#compilerJar` and `ScalaInstance#extraJars`. Here's the commit (https://github.com/sbt/sbt/commit/5b5577a1878dcdbce36f84514c3ea6c0d04389bb#diff-3c4b2252d31738d0ea4eb81da88d2838R20). @eed3si9n any insights on this?

szeiger:
Unless the bug in sbt that prompted this workaround has been fixed, I don't think we can avoid the deprecated calls.

dwijnand:
@szeiger even if whatever is happening with isManagedVersion is fixed, there's still https://github.com/scala/scala/blob/10c609e750a7089055b126e6231e5ddb2f2e8623/build.sbt#L759

szeiger:
For this one we could ask sbt for _all_ jars to find the file. Not the most elegant solution but it should work with non-deprecated API.

szeiger:
Actually we could do some filtering by file name on both. That's what sbt does anyway. However, I wonder how sbt is supposed to get rid of the need to distinguish between compiler and library jars. You don't want the compiler jar on the classpath by default, right?

dwijnand:
We could do filtering on file name, but that's not what sbt does, it does filtering on the module information: https://github.com/sbt/sbt/blob/v0.13.13/main/src/main/scala/sbt/Defaults.scala#L395-L408

szeiger:
Here are the changes to avoid the deprecated ScalaInstance API: https://github.com/SethTisue/scala/compare/sbt-deprecations...szeiger:sbt-deprecations

Looks rather stupid though. The API do get the individual JARs from a ScalaInstance is deprecated but we still have to pass the individual JARs in, only to have ScalaInstance concatenate the lists...

szeiger:
@dwijnand That should amount to pretty much the same result. We don't get the module descriptors for the JARs, so we'll have to use the file names. If use use a local `scalaHome`, sbt already uses file names.

SethTisue:
I added Dale and Stefan's commits, so I think this is ready for merge, assuming Jenkins agrees

enjoy the silence

SethTisue:
I threw in an additional commit replacing all `.settings(x: _*)` with `.settings(x)`

retronym:
LGTM!

PR #5502, opened 2016-11-05
Spec.: Add language code for some Markdown code blocks that had none.

dsbos:
Some code blocks in the Scala language specification did not have a language code (to enable language-specific code highlighting). Most of those contained Scala (literal Scala or mostly Scala (e.g., with ellipses)) or Java.  (A few were non- or not-quite-Scala representations of mappings.)

This branch/PR adds "scala" and "java" language codes to the starting "```" line for those blocks.


SethTisue:
LGTM, but could you retarget onto the `2.12.x` branch? (I've changed the milestone.) also, I think this could be a single commit.

SethTisue:
thanks Daniel!

PR #5503, opened 2016-11-07
Spec:  Fix 2 pairs of quotes in text.

dsbos:
This change fixes two pairs of quotes that were still ASCII ` and ' (instead of the intended left and right single quotes (U+2018 and U+2019)).

(It also clarifies specification of quote characters in the spec's read-me file.)


SethTisue:
LGTM, thanks!

PR #5504, opened 2016-11-07
Silence SBT logging about macros and incremental compilation.

retronym:
Since upgrading to SBT 0.13.12, clean builds have incurred warnings like:

    Because JavaMirrors.scala contains a macro definition, the following
    dependencies are invalidated unconditionally: ....

This commit disables this behaviour of the SBT incremental compiler in
the library and reflect projects, as these aren't regular macros (the
macro implementations are hard coded in the compiler in `FastTrack`)
so the new behaviour isn't actually improving correctness of inc.
compilation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -358,6 +358,8 @@ lazy val library = configureAsSubproject(project)
         "-doc-root-content", (sourceDirectory in Compile).value + "/rootdoc.txt"
       )
     },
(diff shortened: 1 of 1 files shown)


SethTisue:
LGTM

milessabin:
Can we have this on 2.12.x and 2.13.x as well please :-)

retronym:
@milessabin #5510

PR #5505, opened 2016-11-08
Fix wrong description of deprecation policy

soc:


soc:
No idea, how I messed that up, spotted by @SethTisue.

SethTisue:
hmm, should the 2.12 portion that follows be updated now that 2.12.0 is out, or do you think the wording is satisfactory? "will be removed in this release" is ambiguous about whether it applies to 2.12.0 only or the whole 2.12.x series, but maybe it's clear enough.

soc:
@SethTisue Mhh, which sentence do you have in mind? It reads fine to me, but I could have missed something obviously wrong again. :-)

Jasper-M:
Shouldn't the line above it read:

> A deprecated element of the Scala language or a definition in the Scala standard library will be preserved **for** at least another major version. 

And that line still seems to mean that "an element deprecated since 2.12 will be preserved in 2.13".

soc:
@jasper-m You are right, I need to fix that, too.

PR #5506, opened 2016-11-08
Avoid name table pollution with fresh existentials

retronym:
During large compilations runs, the large numbers of globally unique
fresh names for existentials captured from prefixes of `asSeenFrom`.
is a) somewhat wasteful (all these names are interned in the name table)
, and, b) form a pathological case for the current implementation of
`Names#hashValue`, which leads to overfull hash-buckets in the name table.

`hashValue` should probably be improved, but my attempts to do so have
shown a small performance degradation in some benchmarks. So this commit
starts by being more frugal with these names, only uniquely naming
within an `asSeenFrom` operation.

References scala/scala-dev#246

### DIFF
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -4487,6 +4487,7 @@ trait Types
               debuglog(s"transposed irregular matrix!? tps=$tps argss=$argss")
               NoType
             case Some(argsst) =>
+              var capturedParamIds = 0
               val args = map2(sym.typeParams, argsst) { (tparam, as0) =>
                 val as = as0.distinct
                 if (as.size == 1) as.head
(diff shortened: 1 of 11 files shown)


retronym:
Review by @adriaanm

lrytz:
we could create a small class to encapsulate the variable

```
class FreshExistentialNameCreator {
  var counter
  def nm(suffix) = { counter += 1; newTypeName("_" + counter + suffix) }
}

def freshExistential(suffix: String, n: FreshExistentialNameCreator): TypeSymbol =
  newExistential(n.nm(suffix), pos)
```

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4508,8 +4509,10 @@ trait Types
                     else { // Martin: I removed this, because incomplete. Not sure there is a good way to fix it. For the moment we
                       // just err on the conservative side, i.e. with a bound that is too high.
                       // if(!(tparam.info.bounds contains tparam))   //@M can't deal with f-bounds, see #2251
+                      capturedParamIds += 1
+                      val capturedParamId = capturedParamIds
```
adriaanm, line 4513: What's the intent of the `val`? Is there some capture that this is avoiding? (I'm not seeing one.)

adriaanm:
I'm okay to merge as-is and come back and polish later.

adriaanm:
LGTM

PR #5507, opened 2016-11-08
SI-10034: Regression: Make Future.failed(e).failed turn into a success instead of failure

viktorklang:
JIRA: https://issues.scala-lang.org/browse/SI-10034

Massive thanks to the reporter who found this >2 year old lurking regression.

I'm surprised the community build didn't catch this, but then OTOH it is a non-happy-path thing.

Review by @retronym ?

Ping @phaller & @heathermiller


retronym:
LGTM. I don't see similar errors in the neighbourhood.

PR #5509, opened 2016-11-08
SI-10032 Fix code gen with returns in nested try-finally blocks

lrytz:
Return statements within `try` or `catch` blocks need special treatement
if there's also a `finally`

    try { return 1 } finally { println() }

For the return, the code generator emits a store to a local and a jump
to a "cleanup" version of the finally block. There will be 3 versions
of the finally block:

  - One reached through a handler, if the code in the try block
    throws; re-throws at the end
  - A "cleanup" version reached from returns within the try; reads the
    local and returns the value at the end
  - One reached for ordinary control flow, if there's no return and no
    exception within the try

If there are multiple enclosing finally blocks, a "cleanup" version is
emitted for each of them. The nested ones jump to the enclosing ones,
the outermost one reads the local and returns.

A global variable `shouldEmitCleanup` stores whether cleanup versions
are required for the curren finally blocks. By mistake, this variable
was not reset to `false` when emitting a `try-finally` nested within a
`finally`:

    try {
      try { return 1 }
      finally { println() } // need cleanup version
    } finally {             // need cleanup version
      try { println() }
      finally { println() } // no cleanup version needed!
    }

In this commit we ensure that the variable is reset when emitting
nested `try-finally` blocks.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -488,16 +488,11 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
           bc emitRETURN returnType
         case nextCleanup :: rest =>
           if (saveReturnValue) {
-            if (insideCleanupBlock) {
-              reporter.warning(r.pos, "Return statement found in finally-clause, discarding its return-value in favor of that of a more deeply nested return.")
-              bc drop returnType
-            } else {
-              // regarding return value, the protocol is: in place of a `return-stmt`, a sequence of `adapt, store, jump` are inserted.
-              if (earlyReturnVar == null) {
-                earlyReturnVar = locals.makeLocal(returnType, "earlyReturnVar")
-              }
-              locals.store(earlyReturnVar)
+            // regarding return value, the protocol is: in place of a `return-stmt`, a sequence of `adapt, store, jump` are inserted.
+            if (earlyReturnVar == null) {
+              earlyReturnVar = locals.makeLocal(returnType, "earlyReturnVar")
             }
+            locals.store(earlyReturnVar)
           }
           bc goTo nextCleanup
           shouldEmitCleanup = true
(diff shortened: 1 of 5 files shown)


lrytz:
review by @retronym

retronym:
Not sure if related to this change, but can you shed some light on:

``` java
class Test {
        System.out.println(test()); // prints 2
    }
    static int test() {
        try { return 1; } finally { try { return 2; } finally {} } 
    }
}
```

``` scala
object Test {
    // Scala 2.11: 2
    // Scala 2.12: "sandbox/test.scala:7: warning: Return statement found in finally-clause, discarding its return-value in favor of that of a more deeply nested return.", prints 1
    println(test);
  }
  def test: Int = {
    try { return 1 } finally { try { return 2 } finally {} }
  }
}
```

lrytz:
Uh, very interesting. I should have caught that. It's most likely a difference between GenASM and GenBCode. Need to read the spec.

lrytz:
I'd say the scala spec is incomplete about the fact that `finally` blocks are executed if a `return` occurs in a protected area: it [just says](http://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#return-expressions) "The evaluation of any statements or expressions following the return expression is omitted".

So we should follow what java does.

The [java spec](https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.17) says that a "return completes abruptly", it "attempts to transfer control to the invoker", but that attempt may fail: `finally` statements of enclosing `try` are executed first, and "Abrupt completion of a finally clause can disrupt the transfer of control initiated by a return statement".

[Abrupt completion](https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.1) is basically returning or throwing.

The spec of [completion of a `try-finally` statement](https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.2) is a bit complicated, but basically it says: if the `finally` blocks completes abruptly for reason S, then the entire `try` statement completes abruptly with reason S. An abrupt termination of the `try` block for a different reason R is discarded.

So the behavior after this PR is wrong for your example. Given that your example doesn't compile with 2.12.0 (crashes the compiler), it should be safe to fix it for 2.12.1.

lrytz:
OK, here's an example that will change semantics in 2.12.1, compared to 2.12.0:

```
def t: Int = try {
  try { return 1 }
  finally { return 2 }
} finally { println("hi") }
```

In 2.11.8, the result is 2, while in 2.12.0 it's 1, which is wrong. It's also wrong in 2.11.8 with `-Ybackend:GenBCode`. I'm pretty sure we should fix this for 2.12.1, and point it out in the release notes.

retronym:
LGTM. Would have been a good candidate bug to find with a fuzzer that compares behaviour to equivalent Java.

PR #5512, opened 2016-11-08
Add “test”, “scaladoc” and “repl” projects to scala-compiler docs

szeiger:
This makes the scaladoc bundle for scala-compiler consistent with the
binary and source bundles.


lrytz:
maybe this PR could target 2.11.x?

szeiger:
2.11 has some of the new sbt support but distributions are still published with ant, so this would be useless. We would need to do the same in the ant build.

lrytz:
i see, thanks!

lrytz:
LGTM. build failure looks spurious: `Internal error when running tests: sbt.ForkMain$Run$RunAborted: java.net.SocketException: Broken pipe`

PR #5513, opened 2016-11-08
improve top-level compiler/reflect doc text

SethTisue:
this shows up at http://www.scala-lang.org/api/2.12.0/scala-compiler/

ideally there'd be something better here, but we should at least not
link to egregiously outdated stuff


lrytz:
LGTM

PR #5515, opened 2016-11-08
SI-10033 First round of deprecation cleanups for 2.13

soc:
This addresses most of the low-hanging fruits in the standard library.


soc:
This shaves off 11% of the jar size. 5.5MB -> 4.9MB.

SethTisue:
I would like to run the community build against this, but we don't quite have the 2.13.x community build in good enough shape yet — fixes are in progress on macro-paradise, genjavadoc, etc.

adriaanm:
Great! It would be really really nice if we could split this up in phases, and do the deprecations that don't break the community build first (most of them look like pure wins), but I'm afraid removing e.g.,  `Pair` will break the community build and once it's broken we're flying blind until all deprecations are addressed.

adriaanm:
We can also remove some parts of the community build to make this task easier, but we should be able to build at least the core of the eco-system on 2.13.x by the end of the year.

soc:
> Build was aborted
> Aborted by Adriaan Moors

@adriaanm How can I figure out which changes break the community build?

> It would be really really nice if we could split this up in phases

Yes, that's the intention. I tried to avoid most things that looked dangerous in this PR. This PR has only a small subset of pending removals.

> I'm afraid removing e.g., Pair

Will restore `Pair`. (`Triple` too?) I wanted to have this PR tested on the server, because the ScalaCheck tests all seem to fail on my machine, and I think that's only a local issue.

adriaanm:
ah sorry, meant to abort the job of the closed pr. my bad!

adriaanm:
Weird, I got the right job, but somehow it's reported here. (Ah, they have the same sha but different target branch)

soc:
Thanks! Let's see what the ScalaCheck tests tell us. On my local machine I get error messages as if ScalaCheck wasn't there. This might be a result of the recent out-sourcing of ScalaCheck?

adriaanm:
Scalacheck tests pass on the CI! Can you reproduce the error from a clean build?

soc:
@adriaanm That depends on how Florida votes. Otherwise I get some well-deserved sleep and try to reproduce it tomorrow. :-)

SethTisue:
> How can I figure out which changes break the community build

You can run the community build locally: https://github.com/scala/community-builds/wiki/Maintenance#local-runs

You can also open an issue or PR in that repo and request that we trigger a Jenkins run.

(Ideally if you opened a PR, Scabot would do a Jenkins run for you; that's https://github.com/scala/scala-jenkins-infra/issues/92)

soc:
Also, please drop my name from the CLA list and notify me when it is done.

PR #5519, opened 2016-11-12
SI-6407 Default to UTF-8 for text encoding

soc:
Traditionally there were two ways of getting a “default” text encoding:
 - Use UTF-8
 - Use the underlying platform's encoding

We shouldn't invent an additional, third way to do that.

Instead we use UTF-8 by default, and encourage people to be explicit.
Most of the code written out there is never designed with arbitrary
encoding changes via `-Dfile.encoding` or other means in mind, and only
causes pain, suffering and a larger testing surface for developers.

Deprecates the `Codec.ISO8859` constant, because the number is NOT
optional. "ISO-8859" is like "UTF" ... it's not meaningful and doesn't
denote a specific encoding.

Also fixes code that used `toLowerCase` or `toUpperCase` without a
locale. This is dangerous as code and scalac itself will behave
differently under some locales, namely azeri, lithuanian and turkish.

### DIFF
--- a/src/library/scala/io/Codec.scala
+++ b/src/library/scala/io/Codec.scala
@@ -9,24 +9,11 @@
 package scala
 package io
 
-import java.nio.charset.{ Charset, CharsetDecoder, CharsetEncoder, CharacterCodingException, CodingErrorAction => Action }
-import scala.annotation.migration
 import scala.language.implicitConversions
+import scala.annotation.migration
+import java.nio.charset.{ Charset, CharsetDecoder, CharsetEncoder, CharacterCodingException, CodingErrorAction => Action, StandardCharsets }
 
-// Some notes about encodings for use in refining this implementation.
-//
-// Emails: encoding recorded in header, e.g. Content-Type: charset= "iso-8859-1"
-// HTML: optional content-type meta tag.
-//   <meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
-// XML: optional encoding parameter.
-//   <?xml version="1.0" encoding="ISO8859-1" ?>
-//
-// MacRoman vs. UTF-8: see http://osdir.com/ml/lang-jruby-devel/2009-04/msg00071.html
-// -Dfile.encoding: see http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4375816
-
-/** A class for character encoding/decoding preferences.
- *
- */
+/** A class for character encoding/decoding preferences. */
 class Codec(val charSet: Charset) {
   type Configure[T] = (T => T, Boolean)
   type Handler      = CharacterCodingException => Int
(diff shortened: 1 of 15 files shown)


soc:
Is there any way I can abort the 2.12 build?

SethTisue:
I assume Scabot is confused by the change of base branch :-\

SethTisue:
I'll have to tinker with this. but not today

src/library/scala/io/Codec.scala
```diff
@@ -9,24 +9,11 @@
...
-// Some notes about encodings for use in refining this implementation.
-//
-// Emails: encoding recorded in header, e.g. Content-Type: charset= "iso-8859-1"
-// HTML: optional content-type meta tag.
-//   <meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
-// XML: optional encoding parameter.
-//   <?xml version="1.0" encoding="ISO8859-1" ?>
-//
-// MacRoman vs. UTF-8: see http://osdir.com/ml/lang-jruby-devel/2009-04/msg00071.html
-// -Dfile.encoding: see http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4375816
```
som-snytt, line 25: That's a useful URL.

src/library/scala/io/Codec.scala
```diff
@@ -77,17 +64,20 @@ trait LowPriorityCodecImplicits {
...
-  /** Optimistically these two possible defaults will be the same thing.
+  /* Optimistically these two possible defaults will be the same thing.
    *  In practice this is not necessarily true, and in fact Sun classifies
    *  the fact that you can influence anything at all via -Dfile.encoding
    *  as an accident, with any anomalies considered "not a bug".
    */
-  def defaultCharsetCodec = apply(Charset.defaultCharset)
+  /** Always returns a UTF8 codec. */
+  def defaultCharsetCodec = UTF8
+  @deprecated("use `Codec(scala.util.Properties.encodingString)` to create a codec influenced by the `-Dfile.encoding` JVM flag", "2.12.0")
```
som-snytt, line 78: I wonder if all file.encoding usage should be deprecated. Also on util.Properties. It's a naming conflict with jvm internals and a bug to use it.

som-snytt:
Maybe there's a place for an extension string.casefolded in scala.tools.nsc to encapsulate toLowerCase in English locale. Maybe also a method for Names.

src/library/scala/io/Codec.scala
lrytz, line 78: `since` should be `2.13.0`

src/library/scala/io/Source.scala
```diff
@@ -109,34 +110,28 @@ object Source {
   def fromBytes(bytes: Array[Byte], enc: String): Source =
     fromBytes(bytes)(Codec(enc))
 
-  /** Create a `Source` from array of bytes, assuming
-   *  one byte per character (ISO-8859-1 encoding.)
-   */
+  /** Create a `Source` from array of bytes, assuming one byte per character (ISO-8859-1 encoding.) */
+  @deprecated("Use `fromBytes` and specify encoding explicitly.", "2.12.0")
```
lrytz, line 114: `since = 2.13.0`

PR #5521, opened 2016-11-12
Improve support for enum trees

soc:
These are the minimal changes required to allow macro authors to emit
trees that end up as valid JVM enums in the backend.


soc:
See https://github.com/scala/scala/pull/5352 for some earlier discussion.

soc:
Is there any way I can abort the 2.12 build?

PR #5522, opened 2016-11-12
SI-9507 Make Stream #:: and #::: allow type widening

ruippeixotog:
Self explanatory; simply adds a `[B >: A]` type parameter to `#::` and `#:::`, as suggested in [SI-9886](https://issues.scala-lang.org/browse/SI-9886).

This PR targets 2.13 as I'm not 100% sure if this change leads to binary or source incompatibility. If you think this change can be included in Scala 2.12, please tell me and I'll rebase this.

### DIFF
--- a/src/library/scala/collection/immutable/Stream.scala
+++ b/src/library/scala/collection/immutable/Stream.scala
@@ -1119,11 +1119,11 @@ object Stream extends SeqFactory[Stream] {
     /** Construct a stream consisting of a given first element followed by elements
      *  from a lazily evaluated Stream.
      */
-    def #::(hd: A): Stream[A] = cons(hd, tl)
+    def #::[B >: A](hd: B): Stream[B] = cons(hd, tl)
     /** Construct a stream consisting of the concatenation of the given stream and
      *  a lazily evaluated Stream.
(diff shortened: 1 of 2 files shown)


SethTisue:
(Scabot is confused by the base branch having been changed)

lrytz:
Should be fine for 2.12.x, @szeiger what do you think?

ruippeixotog:
There seems to be some problem with the build, as it never seems to finish even when retried. It seems to be because I changed the base branch as @SethTisue said, as I think the build passed before I changed it.

szeiger:
I can't think of a better reason than this being an oversight in the first place, so the change in general LGTM.

It breaks source compatibility in corner cases (like eta-expanding one of the affected methods) but it's probably not bad enough to delay it until 2.13.

ruippeixotog:
@SethTisue @szeiger I have just rebased my commit on top of 2.12.x. Sorry for taking so long, I completely forgot about this PR.

The build was having some issues due to the first change of the base branch. Let's see if it works now that it's back to 2.12.x.

szeiger (approved):

PR #5523, opened 2016-11-12
SI-9706 Add a fromString method to Numeric trait

ruippeixotog:
Adds a `fromString` method to the `Numeric` trait. This would allow methods accepting any `T` with a numeric instance to easily parse string inputs as `T` instances, while currently that isn't possible at all in the general case.

### DIFF
--- a/src/library/scala/math/Numeric.scala
+++ b/src/library/scala/math/Numeric.scala
@@ -10,6 +10,7 @@ package scala
 package math
 
 import scala.language.implicitConversions
+import scala.util.Try
 
 /**
  * @since 2.8
@@ -37,6 +38,7 @@ object Numeric {
     def rem(x: BigInt, y: BigInt): BigInt = x % y
     def negate(x: BigInt): BigInt = -x
     def fromInt(x: Int): BigInt = BigInt(x)
+    def parseString(str: String): Option[BigInt] = Try(BigInt(str)).toOption
     def toInt(x: BigInt): Int = x.intValue
     def toLong(x: BigInt): Long = x.longValue
     def toFloat(x: BigInt): Float = x.floatValue
@@ -52,6 +54,7 @@ object Numeric {
     def rem(x: Int, y: Int): Int = x % y
     def negate(x: Int): Int = -x
     def fromInt(x: Int): Int = x
+    def parseString(str: String): Option[Int] = Try(str.toInt).toOption
     def toInt(x: Int): Int = x
     def toLong(x: Int): Long = x.toLong
     def toFloat(x: Int): Float = x.toFloat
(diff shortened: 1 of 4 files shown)


soc:
Last I remember is that Haskell devs regret even adding `fromInt` to their `Num`. I think we need to investigate and understand that first...

ruippeixotog:
Hmm, interesting. Can you point me to more information or something else to read on that?

lrytz:
I'm not sure we should encourage unsafe conversions even more. Maybe with signature `def fromString(str: String): Option[T]`? But I vote to keep this external, it can be a type class as well (as discussed on http://stackoverflow.com/questions/35980321/scala-create-a-numeric-from-a-string).

SethTisue:
anyone else have an opinion on this?

xavier-fernandez:
IMO the method return should be either a `Option[T]` or something like `Either[InvalidNumberFormatError, T]`.

szeiger:
Possible risks:

- Parsing can fail or silently drop precision. But considering all the other `Numeric` methods which can silently drop precision I'm not too worried about this aspect. None of the other methods throw exceptions so I'd go with `Option[T]` to make failure explicit.

- There's no good default implementation so all implementors of `Numeric` need to be changed.

In general, I think it makes sense to make this a part of `Numeric`.

Ichoran:
I think that we should have `parse` in the name and the return type should be `Option[X]`; there is no substantive information about a parse failure other than that the string didn't actually contain the type we tried to parse.  The implementation should take care to avoid throwing an exception when attempting to parse data of the wrong type (but for floating point that will probably mean validation and then try-catch just to be sure, since FP parsing is tricky).

ruippeixotog:
I makes all sense for `fromString` to return an `Option[A]`. If everyone agrees, I can change the signature to `def parseString(str: String): Option[T]`.

SethTisue:
> I think that we should have `parse` in the name

why? I like the `.toX`, `.fromX` naming convention

Ichoran:
@SethTisue - Because `toX` and `X.fromY` generally return `X` not `Option[X]`.  They're not used, typically, for operations that frequently fail.  In this case, especially, `"fish".toInt` will throw an exception, so it's weird that `fromString` would not have the same return type.

SethTisue:
I like the `Option` return type.

And I agree with @szeiger and @Ichoran that this is worth including in stdlib. I think it would be irritating to have to pull in an extra library for this. stdlib should cover basic conversions like this in a good-enough-for-basic-usage way. more demanding users can use external libraries.

I don't know what the Haskell issue Simon referred to was, but according to http://kodu.ut.ee/~varmo/MFP2004/PreludeTour.pdf `fromInt` doesn't return a `Maybe` (Haskell's `Option` type), so perhaps using `Option` here addresses that.

SethTisue:
> But I vote to keep this external, it can be a type class as well (as discussed on http://stackoverflow.com/questions/35980321/scala-create-a-numeric-from-a-string)

while the idea of a general `Readable` typeclass is nice, we don't have one, even after 10+ years. (if we ever get one — and there is no concrete prospect of this — we could deprecate this addition.)

szeiger:
@ruippeixotog It looks like the prevailing opinion is to add the `Option`-based parsing methods to `Numeric`. Can you update this PR accordingly and rebase it? We're planning to close the queue for M2 at the end of next week.

ruippeixotog:
@szeiger I have just changed the signature to `def parseString(str: String): Option[T]` and rebased the commit.

PR #5524, opened 2016-11-13
[nomerge] SI-8433 SI-9689 test status quo failures

som-snytt:
This progressed on 2.12; in case the tests are useful.

### DIFF
--- a/test/files/run/t8433.scala
+++ b/test/files/run/t8433.scala
@@ -1,0 +1,54 @@
+
+import tools.partest.DirectTest
+import reflect.internal.util._
+
+// mimic the resident compiler failure by recompiling
+// the class with new run of same global.
+object Test extends DirectTest {
+
+  override def code = """
+    object Main {
+      def main(args: Array[String]): Unit = {
+        Surf xmain args
+        import trial.core.Rankable
+        object Surf {
+          def xmain(args: Array[String]): Unit = {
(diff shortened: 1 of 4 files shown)


som-snytt:
Not sure if this is a known syndrome, or whether I should rebuild in the face of it.

```
[error] symbol value x4#391586 does not exist in scala.reflect.internal.tpe.GlbLubs$class.glb0$1, which contains locals value t#246231,value o36#391634,value x#391628,value o34#391632,value x1#390515,variable x21#391611,variable rc30#391627,value depth$4#780046,value ts$6#780045,value ts0#245140,value $this#808330. 
[error] Method code: def glb0($this: reflect#2601.internal#7670.SymbolTable#7937, ts0: scala.`package`.List[GlbLubs.this.Type], ts$6: List#8738, depth$4: Int#1131): GlbLubs.this.Type = {
```

lrytz:
It's https://github.com/scala/scala-dev/issues/251. Been around for a while, but happening all the time lately.. We haven't nailed it down.

lrytz:
I think we can live without the progression tests in 2.11.x, given your other PR (#5525)

som-snytt:
If it's not obvious what fixed it, probably it won't get backported for 2.11.9 anyway.

Have they printed the "2.11.9 End of the Line" t-shirts yet?

PR #5527, opened 2016-11-13
String.replaceAllLiterally is literally String.replace

som-snytt:
The method is not deprecated outright because it avoids
the overloaded equivalent.

`replaceAllLiterally` literally just replaces `replace`. Arguably its name is more expressive.

A really useful extension would be `replaceAllAlliteratively`.


lrytz (approved):

lrytz:
Thanks!

som-snytt:
@SethTisue I must have missed this change of title. :laughing:

PR #5528, opened 2016-11-14
Changed HashMap.getOrElseUpdate to only calculate the index once

l0rinc:
Fixes https://issues.scala-lang.org/browse/SI-10049

Since `groupBy` uses this method extensively and suffered a measurable slowdown in `2.12.0`, this modification restores (and exceeds) its original speed.

---
included benchmarks:

(`ns/op` → smaller is better)

`before (2.12.0):`
```java
Benchmark                                     (size)  Mode  Cnt      Score      Error  Units
s.c.immutable.VectorMapBenchmark.groupBy          10  avgt   20    865.693 ±    7.869  ns/op
s.c.immutable.VectorMapBenchmark.groupBy         100  avgt   20   3095.657 ±   56.438  ns/op
s.c.immutable.VectorMapBenchmark.groupBy        1000  avgt   20  28247.005 ±  470.513  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate      10  avgt   20    836.561 ±   20.085  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate     100  avgt   20   7891.368 ±   56.808  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate    1000  avgt   20  97478.629 ± 1782.497  ns/op
```
`after:`
```java
Benchmark                                     (size)  Mode  Cnt      Score      Error  Units
s.c.immutable.VectorMapBenchmark.groupBy          10  avgt   20    627.007 ±    9.718  ns/op
s.c.immutable.VectorMapBenchmark.groupBy         100  avgt   20   2086.955 ±   19.042  ns/op
s.c.immutable.VectorMapBenchmark.groupBy        1000  avgt   20  19515.234 ±  173.647  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate      10  avgt   20    503.208 ±    2.643  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate     100  avgt   20   5526.483 ±   28.262  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate    1000  avgt   20  69265.900 ±  674.958  ns/op
```

i.e. for the given benchmark conditions `~40%` faster `groupBy` and `getOrElseUpdate`

### DIFF
--- a/src/library/scala/collection/mutable/HashMap.scala
+++ b/src/library/scala/collection/mutable/HashMap.scala
@@ -72,6 +72,37 @@ extends AbstractMap[A, B]
     else Some(e.value)
   }
 
+  override def getOrElseUpdate(key: A, defaultValue: => B): B = {
+    val i = index(elemHashCode(key))
+    val entry = findEntry(key, i)
+    if (entry != null) entry.value
+    else addEntry(createNewEntry(key, defaultValue), i)
+  }
+
+  /* inlined HashTable.findEntry0 to preserve its visibility */
+  private[this] def findEntry(key: A, h: Int): Entry = {
+    var e = table(h).asInstanceOf[Entry]
+    while (notFound(key, e))
+      e = e.next
+    e
+  }
+  private[this] def notFound(key: A, e: Entry): Boolean = (e != null) && !elemEquals(e.key, key)
+
+  /* inlined HashTable.addEntry0 to preserve its visibility */
+  private[this] def addEntry(e: Entry, h: Int): B = {
+    if (tableSize >= threshold) addEntry(e)
+    else addEntry0(e, h)
+    e.value
+  }
+
+  /* extracted to make addEntry inlinable */
+  private[this] def addEntry0(e: Entry, h: Int) {
+    e.next = table(h).asInstanceOf[Entry]
+    table(h) = e
+    tableSize += 1
+    nnSizeMapAdd(h)
+  }
+
   override def put(key: A, value: B): Option[B] = {
     val e = findOrAddEntry(key, value)
     if (e eq null) None
--- a/test/benchmarks/README.md
+++ b/test/benchmarks/README.md
@@ -5,9 +5,7 @@ that makes use of the [SBT plugin](https://github.com/ktoso/sbt-jmh) for [JMH](h
 
 ## Running a benchmark
 
-The benchmarks require first building Scala into `../../build/pack` with `ant`.
-If you want to build with `sbt dist/mkPack` instead,
-you'll need to change `scalaHome` in this project.
+The benchmarks require first building Scala into `../../build/pack`.
 
 You'll then need to know the fully-qualified name of the benchmark runner class.
 The benchmarking classes are organized under `src/main/scala`,
@@ -18,8 +16,7 @@ Using this example, one would simply run
 
     jmh:runMain scala.collection.mutable.OpenHashMapRunner
 
-in SBT.
-SBT should be run _from this directory_.
+in SBT, run _from this directory_ (`test/benchmarks`).
 
 The JMH results can be found under `target/jmh-results/`.
 `target` gets deleted on an SBT `clean`,
--- a/test/benchmarks/project/plugins.sbt
+++ b/test/benchmarks/project/plugins.sbt
@@ -1,2 +1,2 @@
 addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "4.0.0")
-addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.16")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.17")
--- a/test/benchmarks/src/main/scala/scala/collection/immutable/VectorMapBenchmark.scala
+++ b/test/benchmarks/src/main/scala/scala/collection/immutable/VectorMapBenchmark.scala
@@ -1,0 +1,32 @@
+package scala.collection.immutable
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra._
+import org.openjdk.jmh.runner.IterationType
+import benchmark._
+import java.util.concurrent.TimeUnit
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(2)
+@Threads(1)
+@Warmup(iterations = 10)
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class VectorMapBenchmark {
+  @Param(Array("10", "100", "1000"))
+  var size: Int = _
+
+  var values: Vector[Any] = _
+
+  @Setup(Level.Trial) def initKeys(): Unit = {
+    values = (0 to size).map(i => (i % 4) match {
+      case 0 => i.toString
+      case 1 => i.toChar
+      case 2 => i.toDouble
+      case 3 => i.toInt
+    }).toVector
+  }
+
+  @Benchmark def groupBy = values.groupBy(_.getClass)
+}
--- a/test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
+++ b/test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
@@ -1,0 +1,70 @@
+package scala.collection.mutable
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra._
+import org.openjdk.jmh.runner.IterationType
+import benchmark._
+import java.util.concurrent.TimeUnit
+
+import scala.collection.mutable
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(2)
+@Threads(1)
+@Warmup(iterations = 10)
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class HashMapBenchmark {
+  @Param(Array("10", "100", "1000"))
+  var size: Int = _
+
+  var existingKeys: Array[Any] = _
+  var missingKeys: Array[Any] = _
+
+  @Setup(Level.Trial) def initKeys(): Unit = {
+    existingKeys = (0 to size).map(i => (i % 4) match {
+      case 0 => i.toString
+      case 1 => i.toChar
+      case 2 => i.toDouble
+      case 3 => i.toInt
+    }).toArray
+    missingKeys = (size to 2 * size).toArray
+  }
+
+  var map = new mutable.HashMap[Any, Any]
+
+  @Setup(Level.Invocation) def initializeMutable = existingKeys.foreach(v => map.put(v, v))
+
+  @TearDown(Level.Invocation) def tearDown = map.clear()
+
+  @Benchmark def getOrElseUpdate(bh: Blackhole): Unit = {
+    var i = 0;
+    while (i < size) {
+      bh.consume(map.getOrElseUpdate(existingKeys(i), -1))
+      bh.consume(map.getOrElseUpdate(missingKeys(i), -1))
+      i += 1
+    }
+  }
+
+  @Benchmark def get(bh: Blackhole): Unit = {
+    var i = 0;
+    while (i < size) {
+      bh.consume(map.get(existingKeys(i), -1))
+      bh.consume(map.get(missingKeys(i), -1))
+      i += 1
+    }
+  }
+
+  @Benchmark def put(bh: Blackhole): Any = {
+    var map = new mutable.HashMap[Any, Any]
+
+    var i = 0;
+    while (i < size) {
+      map.put(existingKeys(i), i)
+      i += 1
+    }
+
+    map
+  }
+}
--- a/test/benchmarks/build.sbt
+++ b/test/benchmarks/build.sbt
@@ -1,5 +1,5 @@
 scalaHome := Some(file("../../build/pack"))
-scalaVersion := "2.12.0-dev"
+scalaVersion := "2.12.1-dev"
 scalacOptions ++= Seq("-feature", "-opt:l:classpath")
 
 lazy val root = (project in file(".")).
@@ -7,5 +7,5 @@ lazy val root = (project in file(".")).
   settings(
     name := "test-benchmarks",
     version := "0.0.1",
-	libraryDependencies += "org.openjdk.jol" % "jol-core" % "0.4"
+    libraryDependencies += "org.openjdk.jol" % "jol-core" % "0.6"
   )


l0rinc:
Signed the Scala CLA

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -72,6 +72,17 @@ extends AbstractMap[A, B]
     else Some(e.value)
   }
 
+  override def getOrElseUpdate(key: A, defaultValue: => B): B = {
```
l0rinc, line 75: In case the element was missing from the map, the hash code (with some added variance) was calculated first for verification, and again for the `put` operation.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -131,7 +131,7 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
   protected def findEntry(key: A): Entry =
     findEntry0(key, index(elemHashCode(key)))
 
-  private[this] def findEntry0(key: A, h: Int): Entry = {
+  protected[this] final def findEntry0(key: A, h: Int): Entry = {
```
l0rinc: These two parent methods needed to be accessed from `HashMap` to avoid calculation the hash code twice.
`final` was added to keep the semantics, changing only the visibility of the method.

adriaanm:
Review by @Ichoran

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -72,6 +72,17 @@ extends AbstractMap[A, B]
...
+    val h = index(elemHashCode(key))
+    var e = findEntry0(key, h)
+    if (e == null) {
```
paplorinc: could also return early via an else branch, please advise

src/library/scala/collection/mutable/HashTable.scala
retronym: I'm not 100% sure this change is forwards binary compatible. A custom subclass of 2.12.1 hashtable might rely on this newly exposed API and incur a linkage error if run with the 2.12.0 standard library. Our build is supposed to check for such violations, but it hasn't flagged this problem. I need to check out if there is a problem in our build, or in [MiMa](https://github.com/typesafehub/migration-manager) itself. This won't be a blocker for the change as we can always copy/paste these private methods into `HashMap` for the 2.12.x releases.

retronym:
The benchmark could be added to our (recently added) JMH suite, in the same place as [OpenHashMapBenchmark](https://github.com/scala/scala/blob/100d6374c282d94497ee9f0d4a206d427951c74c/test/benchmarks/src/main/scala/scala/collection/mutable/OpenHashMapBenchmark.scala).

We don't run these automatically, but it will at least provide a reference for someone studying this commit or making further changes to this area.

The benchmark itself could do with tests of workloads that exercise more code paths. For instance, using different key types will lead control through through more paths in `BoxesRuntime.{equals*, hash*}`, and different function arguments to produce the default to make the call to `Function.apply` within `getOrElseUpdate` megamorphic.

src/library/scala/collection/mutable/HashTable.scala
retronym: It would also be see whether performance of other methods in `mutable.HashMap` could be improved by using copy/pasted implementations of `findEntry` etc rather than those inherited from the `HashTable` trait. Obviously this isn't an ideal solution, but `HashMap` is so widely used, and these methods are so tiny that failure to JIT inline can have a real impact of observable performance, that I'd be open to this solution for 2.12.1.

l0rinc: My fix isn't about the inlining limit, that's a side-effect of collapsing the stack depth by calling the indexing methods earlier.

This fix would speed up 2.11 also.

Please advise on how you want me to change this PR.
I could start by adding a JMH benchmark for `HashMap` and verifying the performance of other methods also - and trying to optimize them on the way.
Are we sure we want to do the benchmarks from `Scala` and not `Java`?

test/benchmarks/README.md
```diff
@@ -5,9 +5,7 @@ that makes use of the [SBT plugin](https://github.com/ktoso/sbt-jmh) for [JMH](h
 
 ## Running a benchmark
 
-The benchmarks require first building Scala into `../../build/pack` with `ant`.
-If you want to build with `sbt dist/mkPack` instead,
-you'll need to change `scalaHome` in this project.
+The benchmarks require first building Scala into `../../build/pack`.
```
l0rinc, line 8: `sbt` does it correctly also, no need for `ant` anmore

test/benchmarks/project/plugins.sbt
```diff
@@ -1,2 +1,2 @@
 addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "4.0.0")
-addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.16")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.17")
```
l0rinc, line 2: Sent a [PR to update SBT-JMH](https://github.com/ktoso/sbt-jmh/pull/92), it was accepted and the new version containing the latest `JMH` was included here

paplorinc: Sent a [PR to update SBT-JMH](https://github.com/ktoso/sbt-jmh/pull/92), it was accepted and the new version containing the latest JMH was included

test/benchmarks/src/main/scala/scala/collection/immutable/VectorMapBenchmark.scala
```diff
@@ -0,0 +1,40 @@
...
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class VectorMapBenchmark {
+  @Param(Array("10", "100", "1000"))
+  var size: Int = _
+
+  var values: Vector[Any] = _
+
+  @Setup(Level.Trial) def initKeys(): Unit = {
```
paplorinc: @retronym, tried all different possibilities for greater variance.
Is this what you meant :)?

test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,56 @@
...
+  @Benchmark def getOrElseUpdate_found(bh: Blackhole): Unit = {
+    var i = 0
+    while (i < size) {
+      val value = map.getOrElseUpdate(i, -1)
+      bh.consume(value)
+      i += 1
+    }
+  }
+
+  @Benchmark def getOrElseUpdate_missing(bh: Blackhole): Unit = {
```
l0rinc: these two (found and not found in map) should probably be separated, as they have different speeds.
Otherwise the result would depend on their cardinality's ratio.

src/library/scala/collection/mutable/HashTable.scala
l0rinc: @retronym, we would be forced to include more than these two methods, if we don't want to make it `protected` :/

src/library/scala/collection/mutable/HashMap.scala
Ichoran: I'm not sure this is a binary-compatible change.  @retronym - is this?  I would not intuitively have thought it would be with the new trait encoding, but I'm not sure of the new rules.

test/benchmarks/src/main/scala/scala/collection/immutable/VectorMapBenchmark.scala
```diff
@@ -0,0 +1,40 @@
...
+    values = (0 to size).map(i => (i % 10) match {
+      case 0 => i.toString
+      case 1 => i.toByte
```
Ichoran: This seems excessive.  Object and non-object are different paths, and maybe constant-box vs. newly allocated box (byte vs double) is good to check.  Having to worry about so many seems excessive.

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -72,6 +72,18 @@ extends AbstractMap[A, B]
...
+    val i = index(elemHashCode(key))
+    val entry = findEntry0(key, i)
+    if (entry != null)
+      entry.value
+    else {
+      val newEntry = createNewEntry(key, defaultValue)
+      addEntry0(newEntry, i)
+      newEntry.value
+    }
+  }
```
Ichoran, line 104: This implementation looks fine, but what about using `findOrAddEntry` in `HashTable`?  Does that still show the same speed improvement?

l0rinc: The idea is very good, but we don't have a `value` in `HashTable`'s `Entry`, only `key` and `next`.

The other alternative would be to call `findOrAddEntry` directly, but then `defaultValue` would be evaluated twice, which should probably be avoided, right?

``` scala
override def getOrElseUpdate(key: A, defaultValue: => B): B = {
  val e = findOrAddEntry(key, defaultValue)
  if (e ne null) e.value else defaultValue
}
```

retronym: > which should probably be avoided, right?

Correct. Also, we mustn't evaluate `defaultValue` if the key already exists.

Ichoran: Oops, didn't read the signatures correctly.  It's not evaluating twice which is a problem; it's evaluating more than zero times when the entry exists.  You're right; we can't do it this way.

Ichoran:
@paplorinc - As an aside: I'm not sure why using JMH from Scala rather than Java is "counter-intuitive" for you.  A benchmark that depends on what language you use to embed it is almost by definition a bad benchmark because it means that the incidental stuff that you are not supposed to care about is not negligibly inexpensive compared to what you're actually interested in.
If there is a difference, I still wouldn't expect it to necessarily work out in favor of using Java for Scala (and vice-versa, I suppose): by removing the Scala code from a more normal context, you might make the microbenchmarking results even less relevant to normal usage than microbenchmarks usually are.  For instance, if javac uses a pattern around the benchmark that is really easy for the JIT compiler to optimize away, but almost all use cases are from Scala with scalac's not-exactly-the-same pattern, you might fail to see an effect that is bothering everyone.
But really, the point is if it matters, it's already being done wrong.  (The ops/s reported here are pretty well into the safe range where you could even call with interpreted Groovy and it'd be fine.)

src/library/scala/collection/mutable/HashTable.scala
Ichoran: If you inline findEntry0 and addEntry0 you'll end up with entirely protected methods, I think.

Ichoran:
I think the core question that I don't have an answer to is whether we're allowed to introduce an override to getOrElseUpdate and still stay binary compatible.  If yes, then I think the other issues can be worked around by duplicating code (inlining it manually) until you end up with a set of calls to protected or public methods.  If yes in both 2.12 and 2.11, maybe this patch could be backported!  Faster hash maps benefit almost everyone.

retronym:
> I think the core question that I don't have an answer to is whether we're allowed to introduce an override to getOrElseUpdate and still stay binary compatible

This is okay because the class already inherits a method with and identical signature from `AbstractMap`.  Client code of 2.12.1 may include `invokevirtual HashMap.getOrElseUpdate`, but this will not incur an `IncompatibleClassChangeError` if linked against the 2.12.0 class thanks to the override.

Here's an example of making this class change with Java code without linkage errors: https://gist.github.com/794c9f88771b210f5dab0b65793b4682

Here's the [MiMa test case](https://github.com/typesafehub/migration-manager/tree/master/reporter/functional-tests/src/test/class-method-pushed-up-ok) that shows that deleting an exact override is backwards compatible (meaning adding an exact override, like we do here, is forward compatible!)

Ichoran:
@retronym - Okay, then I _think_ we're clear once @paplorinc switches to use `table` and friends instead of `addEntry0` etc..  Hopefully that will still show a speedup.  The JIT compiler won't be able to take advantage of its efforts in already optimizing `addEntry0`, but hopefully it'll do just as well on the inlined version.

l0rinc:
@Ichoran, to be sure I understand what you're requesting, do you want me to keep copy-pasting things from `HashTable` until everything compiles? That would result in duplicating even some fields:

``` scala
override def getOrElseUpdate(key: A, defaultValue: => B): B = {
  val i = index(elemHashCode(key))
  val entry = findEntry0(key, i)
  if (entry != null)
    entry.value
  else {
    val newEntry = createNewEntry(key, defaultValue)
    addEntry0(newEntry, i)
    newEntry.value
  }
}
private[this] final def findEntry0(key: A, h: Int): Entry = {
  var e = table(h).asInstanceOf[Entry]
  while (e != null && !elemEquals(e.key, key)) e = e.next
  e
}
private[this] final def addEntry0(e: Entry, h: Int) {
  e.next = table(h).asInstanceOf[Entry]
  table(h) = e
  tableSize = tableSize + 1
  nnSizeMapAdd(h)
  if (tableSize > threshold)
    resize(2 * table.length)
}
private def resize(newSize: Int) {
  val oldTable = table
  table = new Array(newSize)
  nnSizeMapReset(table.length)
  var i = oldTable.length - 1
  while (i >= 0) {
    var e = oldTable(i)
    while (e != null) {
      val h = index(elemHashCode(e.key))
      val e1 = e.next
      e.next = table(h).asInstanceOf[Entry]
      table(h) = e
      e = e1
      nnSizeMapAdd(h)
    }
    i = i - 1
  }
  threshold = newThreshold(_loadFactor, newSize)
}
private[collection] final def newThreshold(_loadFactor: Int, size: Int) = ((size.toLong * _loadFactor) / loadFactorDenom).toInt
private[collection] final def loadFactorDenom = 1000
```

---

edit: about the Scala JMH tests (which I provided upon request), my concern is that (using a metaphor) you cannot trust a patient to self-diagnose an illness. i.e. you could use sbt-jmh for non-scala-core (i.e. when Scala is "healthy"), but for inside problems, you might be relying on the problem itself for diagnosing it (i.e if the compiler emits wrong bytecode, the benchmark itself will contain that error).

---

edit2: I could extract some logic to stateless methods in `HashTable` to DRY it, if you think that's better than exposing two stateful metohods :)

## 

edit3: I could annotate the two exposed methods with `@migration("exposed for internal optimizations, might change in the future", "2.12.1")`

src/library/scala/collection/mutable/HashMap.scala
l0rinc: Thanks!
In this particular case I think it cannot be evaluated once, only zero or two times (doesn't change the fact).

Ichoran:
@paplorinc - Which fields do you need?  You've got everything already there as protected, don't you?

It might be easier to intercept the size and just eat the extra lookup on an addEntry in the relatively rare case that the table needs to be enlarged.

I grant that it's ugly either way, but if we can avoid even a minor binary incompatibility, I think it's for the best.  If we'd caught this before 2.12 was out, I'd have much preferred your patch as it was where you just change the visibility of some things.

Also, I understand your point in general regarding using a flawed system to detect its own flaws, but in this particular case, do you also understand my point?

retronym:
The binary incompatibility wasn't reported because of a regression in the entry point into MiMa we use. Will be fixed tomorrow: https://github.com/typesafehub/migration-manager/pull/138

l0rinc:
@Ichoran, thanks for the idea, I've tested it from the new `Scala` benchmarks and they maintained their speed advantage (except for the case when it actually needs growing, but that's ok), and the code is also still maintainable - great balance!

Please wait with the merge until I validate the results from `Java` also (I get some strange results that I need to investigate)

@retronym, thank you for your investigations!

retronym:
Could you please add comments in both copies of each duplicated method to warn maintainers about the duplication? For bonus points/pints, you could submit a PR to the 2.13.x branch to make them protected and shared.

src/library/scala/collection/mutable/HashTable.scala
szeiger: @retronym I agree about the binary compatibility. I'll check MiMa for a test case. I'd be surprised if this isn't covered or doesn't work. Could indeed be a problem of the Scala build.

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -72,6 +72,37 @@ extends AbstractMap[A, B]
...
+
+  /* inlined HashTable.findEntry0 to preserve its visibility */
+  private[this] def findEntry(key: A, h: Int): Entry = {
+    var e = table(h).asInstanceOf[Entry]
+    while (notFound(key, e))
+      e = e.next
+    e
+  }
+  private[this] def notFound(key: A, e: Entry): Boolean = (e != null) && !elemEquals(e.key, key)
```
l0rinc, line 89: extracted to make `findEntry` inlinable (and the method more readable):
`scala.collection.mutable.HashMap::findEntry (32 bytes)   inline (hot)`

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -72,6 +72,37 @@ extends AbstractMap[A, B]
...
+
+  /* inlined HashTable.addEntry0 to preserve its visibility */
+  private[this] def addEntry(e: Entry, h: Int): B = {
```
l0rinc, line 92: `scala.collection.mutable.HashMap::addEntry (30 bytes)   inline (hot)`

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -72,6 +72,37 @@ extends AbstractMap[A, B]
     else Some(e.value)
   }
 
+  override def getOrElseUpdate(key: A, defaultValue: => B): B = {
```
l0rinc, line 75: `scala.collection.mutable.HashMap::getOrElseUpdate (48 bytes)   inline (hot)`

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -365,9 +365,8 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
   // this is of crucial importance when populating the table in parallel
   protected final def index(hcode: Int) = {
```
l0rinc, line 366: `scala.collection.mutable.HashTable::index (33 bytes)   inline (hot)`

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -365,9 +365,8 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
...
     val ones = table.length - 1
-    val improved = improve(hcode, seedvalue)
-    val shifted = (improved >> (32 - java.lang.Integer.bitCount(ones))) & ones
-    shifted
+    val exponent = Integer.numberOfLeadingZeros(ones)
```
l0rinc: since `table.length` is a power of two, `table.length - 1` is composed of ones. The bitCount of that is the number of ones (= the exponent of the original power of 2), subtracted from 32 is the number of trailing zeroes.

The benchmark indicates that this is smaller and faster:

``` java
public static class Test {
    final int[] squares = {1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, 33554432, 67108864, 134217728, 268435456, 536870912, 1073741824};

    @Benchmark
    public void bitCount(Blackhole bh) {
        for (int i = 0; i < squares.length; i++) {
            int leadingZeros = Integer.SIZE - Integer.bitCount(squares[i] - 1);
            bh.consume(leadingZeros);
        }
    }

    @Benchmark
    public void numberOfLeadingZeros(Blackhole bh) {
        for (int i = 0; i < squares.length; i++) {
            int leadingZeros = Integer.numberOfLeadingZeros(squares[i] - 1);
            bh.consume(leadingZeros);
        }
    }
}
```

``` java
Benchmark                        Score         Error  Units
Test.bitCount              8746665.311  ± 206703.549  ops/s
Test.numberOfLeadingZeros  9255031.968  ± 114967.712  ops/s
```

(`ops/s`, greater is better)

Note: this affect all hashing operations!

src/library/scala/runtime/BoxesRunTime.java
```diff
@@ -117,36 +117,29 @@ public static double unboxToDouble(Object d) {
...
-        if (x instanceof java.lang.Number)
-            return equalsNumObject((java.lang.Number)x, y);
-        if (x instanceof java.lang.Character)
-            return equalsCharObject((java.lang.Character)x, y);
-        if (x == null)
-            return y == null;
-
-        return x.equals(y);
+    private static boolean equalsNotSame(Object x, Object y) {
+        return x != null && y != null && equalsNotSameOrNull(x, y);
```
l0rinc: if they're not the same (`==`), but one of them is `null`, they're certainly not `equal`
(`scala.runtime.BoxesRunTime::equalsNotSame (22 bytes)   inline (hot)`)

src/library/scala/runtime/BoxesRunTime.java
```diff
@@ -117,36 +117,29 @@ public static double unboxToDouble(Object d) {
...
     }
 
-    public static boolean equalsNumObject(java.lang.Number xn, Object y) {
-        if (y instanceof java.lang.Number)
-            return equalsNumNum(xn, (java.lang.Number)y);
-        if (y instanceof java.lang.Character)
-            return equalsNumChar(xn, (java.lang.Character)y);
-        if (xn == null)
-            return y == null;
+    private static boolean equalsNotSameOrNull(Object x, Object y) {
```
l0rinc: `scala.runtime.BoxesRunTime::equalsNotSameOrNull (38 bytes)   inline (hot)`

test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,56 @@
...
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class HashMapBenchmark {
+  @Param(Array("10", "100", "1000"))
+  var size: Int = _
+
+  var keys: Vector[Any] = _
+
+  @Setup(Level.Trial) def initKeys(): Unit = {
+    keys = (0 to size).map(i => (i % 4) match {
```
l0rinc: @Ichoran, reduced the number of possibilities, hope it's ok now :)

src/library/scala/collection/mutable/HashMap.scala
szeiger: @Ichoran It should be binary compatible. A class gets a forwarder method that delegates to the static implementation method of the interface. Adding an override merely replaces the implementation of the class method without any changes to the signatures. I'll add a proper test case for this to MiMa.

src/library/scala/collection/mutable/HashTable.scala
Ichoran: Speed improvement or no, when the goal is to count the number of leading zeros it's better to clean up the method and say `numberOfLeadingZeros` unless there's a significant performance hit.  It's nice that it's faster too, though!

test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
Ichoran: This looks reasonable, thanks :)

Ichoran:
Looks like you accidentally got a couple of stray commits in here?

src/library/scala/runtime/BoxesRunTime.java
```diff
@@ -117,36 +117,29 @@ public static double unboxToDouble(Object d) {
...
+        if (x instanceof Number) return equalsNumObject((Number) x, y);
+        else if (x instanceof Character) return equalsCharObject((Character) x, y);
+        else return x.equals(y);
+    }
 
-        return xn.equals(y);
+    static boolean equalsNumObject(java.lang.Number xn, Object y) {
```
retronym: We can't change signatures or accessibility of these methods. Even after the #5532, our binary compatibility checker didn't pick up this problem. I've patched the tool in https://github.com/typesafehub/migration-manager/pull/142 to report these errors.

Beyond the constraint that we can't change signatures, we also can't change semantics of of these "building blocks" of boxed equality, because the compiler emits direct calls to a variety of them, depending on how much static knowledge we have about the types and nullability of the arguments.

retronym: It is safe to introduce new private methods, and to refactor existing methods to use those private methods, so long at the semantics of existing non-private methods is unchanged.

I'd suggest to split out changes to `BoxesRuntime` into a separate commit and present benchmark results. We need to weigh up the performance wins against the risk of introducing regressions or subtle compatibility problems when mixing library from 2.12.1 with code compiled with 2.12.0 and vice versa.

retronym: My patched version of MiMa reports:

```
Found 5 binary incompatibilities
================================
 * method equalsNumObject(java.lang.Number,java.lang.Object)Boolean in class
   scala.runtime.BoxesRunTime is inaccessible in current version, it must be
   public.
 * method equals2(java.lang.Object,java.lang.Object)Boolean in class
   scala.runtime.BoxesRunTime does not have a correspondent in current version
 * method equalsCharObject(java.lang.Character,java.lang.Object)Boolean in
   class scala.runtime.BoxesRunTime does not have a correspondent in current
   version
 * method equalsNumNum(java.lang.Number,java.lang.Number)Boolean in class
   scala.runtime.BoxesRunTime is inaccessible in current version, it must be
   public.
 * method equalsNumChar(java.lang.Number,java.lang.Character)Boolean in
   class scala.runtime.BoxesRunTime does not have a correspondent in current
   version
```

l0rinc: Thanks @retronym, will split it :)

test/benchmarks/README.md
```diff
@@ -18,8 +16,7 @@ Using this example, one would simply run
 
     jmh:runMain scala.collection.mutable.OpenHashMapRunner
```
l0rinc, line 17: not sure how to run only a single benchmark, `sbt jmh:run` runs everything, even if I provide a pattern after it :/

test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,70 @@
...
+  @Benchmark def getOrElseUpdate(bh: Blackhole): Unit = {
+    var i = 0;
+    while (i < size) {
+      bh.consume(map.getOrElseUpdate(existingKeys(i), -1))
+      bh.consume(map.getOrElseUpdate(missingKeys(i), -1))
+      i += 1
+    }
+  }
+
+  @Benchmark def get(bh: Blackhole): Unit = {
```
l0rinc, line 50: added `get` and `put` also, even though they weren't affected by this commit

test/benchmarks/src/main/scala/scala/collection/mutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,70 @@
...
+      bh.consume(map.get(missingKeys(i), -1))
+      i += 1
+    }
+  }
+
+  @Benchmark def put(bh: Blackhole): Any = {
+    var map = new mutable.HashMap[Any, Any]
+
+    var i = 0;
+    while (i < size) {
```
l0rinc, line 63: changed `for` comprehension to `while` for better SNR

l0rinc:
@Ichoran, not sure why the build is failing do I still have a `MiMa` violation?
[`library/mima failed: java.lang.RuntimeException: MiMa failed with exit code 1`](https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/3547/console)

szeiger:
```
[info] Checking backward binary compatibility
[info] prev = /home/jenkins/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.12.0.jar, curr = /home/jenkins/workspace/scala-2.12.x-validate-test@2/build/pack/lib/scala-library.jar
Found 1 binary incompatibilities
================================
 * method getExitValue()scala.Tuple2 in class
   scala.sys.process.ProcessImpl#CompoundProcess does not have a correspondent
   in current version

Generated backward filter config definition
============================================

    filter {
        problems=[
            {
                matchName="scala.sys.process.ProcessImpl#CompoundProcess.getExitValue"
                problemName=DirectMissingMethodProblem
            }
        ]
    }


Generated forward filter config definition
===========================================

    filter {
        problems=[]
    }
```

szeiger:
This looks like an unrelated failure due to MiMa having been accidentally disabled for a while.

l0rinc:
@szeiger, what should I do to fix it (tried rebasing and retriggering the build)?

szeiger:
/nothingtoseehere

The failure is due to https://github.com/scala/scala/pull/5481, which was based on an older commit before https://github.com/scala/scala/pull/5532, so the problem was not detected. There is nothing to do here. Since the binary incompatibility is in a private implementation class I would argue that we should whitelist it. @retronym, do you agree?

The other failure in the new build is a spurious error that we've seen before.

Ichoran (approved):
LGTM

retronym:
LGTM, too. This is a fantastic contribution, keep them coming!

l0rinc:
Thanks for your help @Ichoran, @retronym and @szeiger!

PR #5529, opened 2016-11-15
Better inliner support for 2.12 trait encoding

lrytz:
Some changes to the trait encoding came late in the 2.12 cycle, and the
inliner was not adapted to support it in the best possible way.

In 2.12.0 concrete trait methods are encoded as

    interface T {
      default int m() { return 1 }
      static int m$(T $this) { <invokespecial $this.m()> }
    }
    class C implements T {
      public int m() { return T.m$(this) }
    }

If a trait method is selected for inlining, the 2.12.0 inliner would
copy its body into the static super accessor `T.m$`, and from there into
the mixin forwarder `C.m`.

This commit special-cases the inliner:
  - We don't inline into static super accessors and mixin forwarders.
  - Insted, when inlining an invocation of a mixin forwarder, the
    inliner also follows through the two forwarders and inlines the
    trait method body.

There was a difficulty implementing this: inlining the static static
super accessor would copy an `invokespecial` instruction into a
different classfile, which is not legal / may change semantics. That
`invokespecial` is supposed to disappear when inlining the actual
default method body. However, this last step may fail, for example
because the trait method body itself contains instructions that are not
legal in a different classfile.

It is very difficult to perform all necessary checks ahead of time. So
instead, this commit implements the ability to speculatively inline a
callsite and roll back if necessary.

The commit also cleans up the implementation of inliner warnings a
little. The previous code would always emit a warning when a method
annotated `@inline` was not picked by the heuristics - this was a
problem when the callsite in the static super accessor was no longer
chosen.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
@@ -25,8 +25,99 @@ class Inliner[BT <: BTypes](val btypes: BT) {
   import inlinerHeuristics._
   import backendUtils._
 
-  case class InlineLog(request: InlineRequest, sizeBefore: Int, sizeAfter: Int, sizeInlined: Int, warning: Option[CannotInlineWarning])
-  var inlineLog: List[InlineLog] = Nil
+  sealed trait InlineLog {
+    def request: InlineRequest
+  }
+  final case class InlineLogSuccess(request: InlineRequest, sizeBefore: Int, sizeInlined: Int) extends InlineLog {
+    var downstreamLog: mutable.Buffer[InlineLog] = mutable.ListBuffer.empty
+  }
+  final case class InlineLogFail(request: InlineRequest, warning: CannotInlineWarning) extends InlineLog
+  final case class InlineLogRollback(request: InlineRequest, warnings: List[CannotInlineWarning]) extends InlineLog
+
+  object InlineLog {
+    private def shouldLog(request: InlineRequest): Boolean = {
+      def logEnabled = compilerSettings.YoptLogInline.isSetByUser
+      def matchesName = {
+        val prefix = compilerSettings.YoptLogInline.value match {
+          case "_" => ""
+          case p => p
+        }
+        val name: String = request.callsite.callsiteClass.internalName + "." + request.callsite.callsiteMethod.name
+        name startsWith prefix
+      }
+      logEnabled && (upstream != null || (isTopLevel && matchesName))
+    }
+
+    // indexed by callsite method
+    private val logs = mutable.Map.empty[MethodNode, mutable.LinkedHashSet[InlineLog]]
+
+    private var upstream: InlineLogSuccess = _
+    private var isTopLevel = true
+
+    def withInlineLogging[T](request: InlineRequest)(inlineRequest: => Unit)(inlinePost: => T): T = {
+      def doInlinePost(): T = {
+        val savedIsTopLevel = isTopLevel
+        isTopLevel = false
+        try inlinePost
+        finally isTopLevel = savedIsTopLevel
+      }
+      if (shouldLog(request)) {
+        val sizeBefore = request.callsite.callsiteMethod.instructions.size
+        inlineRequest
+        val log = InlineLogSuccess(request, sizeBefore, request.callsite.callee.get.callee.instructions.size)
+        apply(log)
+
+        val savedUpstream = upstream
+        upstream = log
+        try doInlinePost()
+        finally upstream = savedUpstream
+      } else {
+        inlineRequest
+        doInlinePost()
+      }
+    }
+
+    def apply(log: => InlineLog): Unit = if (shouldLog(log.request)) {
+      if (upstream != null) upstream.downstreamLog += log
+      else {
+        val methodLogs = logs.getOrElseUpdate(log.request.callsite.callsiteMethod, mutable.LinkedHashSet.empty)
+        methodLogs += log
+      }
+    }
+
+    def entryString(log: InlineLog, indent: Int = 0): String = {
+      val callee = log.request.callsite.callee.get
+      val calleeString = callee.calleeDeclarationClass.internalName + "." + callee.callee.name
+      val indentString = " " * indent
+      log match {
+        case s @ InlineLogSuccess(_, sizeBefore, sizeInlined) =>
+          val self = s"${indentString}inlined $calleeString. Before: $sizeBefore ins, inlined: $sizeInlined ins."
+          if (s.downstreamLog.isEmpty) self
+          else s.downstreamLog.iterator.map(entryString(_, indent + 2)).mkString(self + "\n", "\n", "")
+
+        case InlineLogFail(_, w) =>
+          s"${indentString}failed $calleeString. ${w.toString.replace('\n', ' ')}"
+
+        case InlineLogRollback(_, _) =>
+          s"${indentString}rolling back, nested inline failed."
+      }
+    }
+
+    def print(): Unit = if (compilerSettings.YoptLogInline.isSetByUser) {
+      val byClassAndMethod: List[(InternalName, mutable.Map[MethodNode, mutable.LinkedHashSet[InlineLog]])] = {
+        logs.
+          groupBy(_._2.head.request.callsite.callsiteClass.internalName).
+          toList.sortBy(_._1)
+      }
+      for {
+        (c, methodLogs) <- byClassAndMethod
+        (m, mLogs) <- methodLogs.toList.sortBy(_._1.name)
+        mLog <- mLogs // insertion order
+      } {
+        println(s"Inline into $c.${m.name}: ${entryString(mLog)}")
+      }
+    }
+  }
 
   def runInliner(): Unit = {
     for (request <- collectAndOrderInlineRequests) {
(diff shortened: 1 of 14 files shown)


lrytz:
review by @retronym

lrytz:
file sizes, 2.12.0:

```
5.2M    /Users/luc/scala/scala-2.12.0/lib/scala-library.jar
3.4M    /Users/luc/scala/scala-2.12.0/lib/scala-reflect.jar
9.6M    /Users/luc/scala/scala-2.12.0/lib/scala-compiler.jar
```

after this change:

```
5.0M    pack-after/lib/scala-library.jar
3.4M    pack-after/lib/scala-reflect.jar
9.5M    pack-after/lib/scala-compiler.jar
```

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -221,26 +213,82 @@ class Inliner[BT <: BTypes](val btypes: BT) {
...
+  class UndoLog(active: Boolean = true) {
+    private var actions = List.empty[() => Unit]
+    private var methodStateSaved = false
+
+    def apply(a: => Unit): Unit = if (active) actions = (() => a) :: actions
+    def run(): Unit = if (active) actions.foreach(_.apply())
+
+    private def arr[T: reflect.ClassTag](l: java.util.List[T]): Array[T] = {
+      val a: Array[T] = new Array[T](l.size)
+      l.toArray(a.asInstanceOf[Array[T with Object]]).asInstanceOf[Array[T]]
```
retronym: I'm guessing this is performance motivated. Warrants a comment? You could special case the empty list (a common case for `tryCatchBlocks`) to squeeze further perf.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -221,26 +213,82 @@ class Inliner[BT <: BTypes](val btypes: BT) {
...
+      methodStateSaved = true
+      val currentInstructions = methodNode.instructions.toArray
+      val currentLocalVariables = arr(methodNode.localVariables)
+      val currentTryCatchBlocks = arr(methodNode.tryCatchBlocks)
+      val currentMaxLocals = methodNode.maxLocals
+      val currentMaxStack = methodNode.maxStack
+
+      apply {
+        // this doesn't work: it doesn't reset the `prev` / `next` / `index` of individual instruction nodes
+        // methodNode.instructions.clear()
```
retronym: Are we after the same behaviour as [`InsnList#removeAll(true)`](https://github.com/scala/scala-asm/blob/master/src/main/java/scala/tools/asm/tree/InsnList.java#L488-L510)? We could make that public in a new release of scala-asm.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -221,26 +213,82 @@ class Inliner[BT <: BTypes](val btypes: BT) {
...
+      postRequests.flatMap(inline(_, undo))
+    }
+
+    canInlineCallsite(request.callsite) match {
+      case None =>
+        doInline(undo)
+
+      case Some((w, illegalAccessInsns)) if illegalAccessInsns.nonEmpty && illegalAccessInsns.forall(ins => request.post.exists(_.callsite.callsiteInstruction == ins)) =>
+        // speculatively inline, roll back if an illegalAccessInsn cannot be eliminated
+        if (undo == NoUndoLogging) {
```
retronym: Is this part preventing recursive speculation?

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -221,26 +213,82 @@ class Inliner[BT <: BTypes](val btypes: BT) {
     impl(post, mainCallsite)
   }
 
+  class UndoLog(active: Boolean = true) {
+    private var actions = List.empty[() => Unit]
+    private var methodStateSaved = false
+
+    def apply(a: => Unit): Unit = if (active) actions = (() => a) :: actions
+    def run(): Unit = if (active) actions.foreach(_.apply())
```
retronym: Perhaps rename to `rollback` or `undo` ?

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -221,26 +213,82 @@ class Inliner[BT <: BTypes](val btypes: BT) {
...
+        methodNode.tryCatchBlocks.clear()
+        methodNode.tryCatchBlocks.addAll(lst(currentTryCatchBlocks))
+
+        methodNode.maxLocals = currentMaxLocals
+        methodNode.maxStack = currentMaxStack
+      }
+    }
+  }
+
+  val NoUndoLogging = new UndoLog(active = false)
```
retronym, line 320: I think it would be a little cleaner to use subclasses rathert than `if (active)` in each method.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
```diff
@@ -221,26 +213,82 @@ class Inliner[BT <: BTypes](val btypes: BT) {
...
+    canInlineCallsite(request.callsite) match {
+      case None =>
+        doInline(undo)
+
+      case Some((w, illegalAccessInsns)) if illegalAccessInsns.nonEmpty && illegalAccessInsns.forall(ins => request.post.exists(_.callsite.callsiteInstruction == ins)) =>
+        // speculatively inline, roll back if an illegalAccessInsn cannot be eliminated
+        if (undo == NoUndoLogging) {
+          val undoLog = new UndoLog()
+          val warnings = doInline(undoLog)
+          if (warnings.nonEmpty) undoLog.run()
```
retronym: Some debug logging of failed speculation would be useful.

retronym:
Would it be possible to cleanly restore the 2.12.0 behaviour (which was imperfect but harmless enough) with a `-Y` option to help us guard against bugs in the new rollback mechanism?

retronym:
Overall this facility looks like it could be useful to build on in the future. For instance, if we inline based on the heuristic of "higher order method with a literal function argument" but we can't actually eliminate the function, we could rollback.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz: I basically need to save all elements into some new collection, an Array seemed most appropriate because `j.u.List` provides `toArray`. I couldn't get it to compile without casts, not sure if that's a Scala compiler bug. IRC, Scala interprets all java-defined type parameters with `<: Object`.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz: Yes, that would be the same behavior. I'll add a comment.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz: Yes. If we start a speculative inline and fail somewhere downstream, it the warning will be pushed up and will cause a rollback.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz: sounds good

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz: I tried the override version, but thought was verbose to repeat all the signatures :) but I can change it around.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
lrytz: Right, we have inline logging already. Will do.

lrytz:
> Would it be possible to cleanly restore the 2.12.0 behaviour (which was imperfect but harmless enough) with a -Y option to help us guard against bugs in the new rollback mechanism?

The problem is that this cannot be done on the user's side: in 2.12.0, when compiling a library, the trait method bodies are duplicated already. So when we ship scala-library, we have to decide whether we do that duplication or not.

> if we inline based on the heuristic of "higher order method with a literal function argument" but we can't actually eliminate the function, we could rollback.

Yes, that's very much the plan, and we can go beyond "eliminate the function" and also work towards eliminating various kinds of boxes. Scala.js does this.

retronym:
> The problem is that this cannot be done on the user's side: in 2.12.0, when compiling a library, the trait method bodies are duplicated already. So when we ship scala-library, we have to decide whether we do that duplication or not.

Makes sense. It might be worth doing a bytecode diff before and after this change on the library to make sure the only changes are the ones we expect.

src/compiler/scala/tools/nsc/backend/jvm/opt/Inliner.scala
retronym: Might be simplest to use `val savedLocalVariables = new ArrayList<LocalVariableNode>(methodNode.localVariables)`. Then you won't need the `lst` or `arr` helper methods at all.

lrytz: thanks, will take a look

lrytz:
Added some cleanup commits to address review feedback. `Yopt-log-inline` now supports nested inlines. Output for Scala library: https://gist.github.com/lrytz/019943fc5802467a3ae3974c72c84e03.

Note:
```
trait T {
  @inline final def foo {
    def bar = 1
    bar
  }
}
class C extends T
class D {
  def baz(c: C) = c.foo
}
```

Here, `C.foo` is inlined, and we get a call to `T.foo$`. Inlining `T.foo$` will fail, as shown by the log:

```
Inline into D.baz: inlined C.foo. Before: 7 ins, inlined: 7 ins.
  inlined T.foo$. Before: 20 ins, inlined: 7 ins.
    failed T.foo. T::foo()V is annotated @inline but could not be inlined: The callee T::foo()V contains the instruction INVOKESTATIC T.bar$1 ()I that would cause an IllegalAccessError when inlined into class D.
    rolling back, nested inline failed.
```

Note that

lrytz:
Added some cleanup commits to address review feedback. `Yopt-log-inline` now supports nested inlines. Output for Scala library: https://gist.github.com/lrytz/019943fc5802467a3ae3974c72c84e03.

Note:
```
trait T {
  @inline final def foo {
    def bar = 1
    bar
  }
}
class C extends T
class D {
  def baz(c: C) = c.foo
}
```

Here, `C.foo` is inlined, and we get a call to `T.foo$`. Inlining `T.foo$` will fail, as shown by the log:

```
Inline into D.baz: inlined C.foo. Before: 7 ins, inlined: 7 ins.
  inlined T.foo$. Before: 20 ins, inlined: 7 ins.
    failed T.foo. T::foo()V is annotated @inline but could not be inlined: The callee T::foo()V contains the instruction INVOKESTATIC T.bar$1 ()I that would cause an IllegalAccessError when inlined into class D.
    rolling back, nested inline failed.
```

Note that the current implementation only rolls back the inlining of `T.foo$`, but not all the way. So in the final result, we inlined the first forwarder. I could change this to roll back to the beginning.

I think with the current state, where inlining chains are only for forwarders, we can keep it as-is. Longer term, when we start doing more speculative inlines, we might want to roll back to the beginning to avoid unnecessary code duplication. But since we only have chains of forwarders for now, we're not increasing code size.

retronym:
LGTM

PR #5531, opened 2016-11-15
SI-10060 Fixes NumericRange.max bug on empty ranges

tabdulradi:
As described in [SI-10060](https://issues.scala-lang.org/browse/SI-10060)
`Range.Int.inclusive(1, 9, -1).max` returns `1` instead of throwing an exception.

### DIFF
--- a/test/junit/scala/collection/immutable/RangeTest.scala
+++ b/test/junit/scala/collection/immutable/RangeTest.scala
@@ -11,17 +11,29 @@ class RangeTest {
   import AssertUtil._
 
   @Test
-  def test_SI10060_long_max(): Unit = {
+  def test_SI10060_numeric_range_min_max(): Unit = {
+    assertEquals(Range.Long.inclusive(1, 9, 1).min, 1)
+    assertEquals(Range.Long.inclusive(1, 9, 1).max, 9)
+    assertEquals(Range.Long.inclusive(9, 1, -1).min, 1)
+    assertEquals(Range.Long.inclusive(9, 1, -1).max, 9)
+    assertThrows[java.util.NoSuchElementException](Range.Long.inclusive(1, 9, -1).min)
+    assertThrows[java.util.NoSuchElementException](Range.Long.inclusive(1, 9, -1).max)
+    assertThrows[java.util.NoSuchElementException](Range.Long.inclusive(9, 1, 1).min)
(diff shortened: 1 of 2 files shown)


test/junit/scala/collection/immutable/RangeTest.scala
```diff
@@ -0,0 +1,30 @@
...
+
+@RunWith(classOf[JUnit4])
+class RangeTest {
+  import Assert._
+  import AssertUtil._
+
+  @Test
+  def test_SI10060_long_max(): Unit = {
+    assertEquals(Range.Int.inclusive(1, 9, 1).min, 1)
+    assertEquals(Range.Int.inclusive(1, 9, 1).max, 9)
```
Blaisorblade, line 25: min and max should also work on `Range.Int.inclusive(9, 1, -1)`, so what about testing those too?

Blaisorblade: Also these tests should probably use `Range.Long`, since that's where the bug was and for consistency with `long` in the method name `test_SI10060_long_max`. EDIT: `Range.Int` currently also uses `NumericRange`, but it's not clear how much the test should rely on that.

Blaisorblade:
At a first look LGTM, apart from the nitpicks on the test. BTW why is this not targeted for 2.12.1 but for 2.12.2?

lrytz:
2.12.2 is already the default target, as we're working to get the 2.12.1 queue down and have a release ASAP. But if something low-risk is ready, it can still go in.

test/junit/scala/collection/immutable/RangeTest.scala
tabdulradi: I renamed the Jira ticket when I noticed the issue affects Range.Int as well, but I forgot to rename the test case. 
I'll fix the method name, and add tests for Long as well.

Blaisorblade:
LGTM, but maybe inclusion in 2.12.1 would call for extra tests? I lack time for double-checking this in more detail.

felixt-cake (approved):

Ichoran (approved):
LGTM - the fix is simple and brings NumericRange into line with Range.  Hopefully no-one was relying upon this behavior; it's asymmetric anyway, so it's less likely.

PR #5532, opened 2016-11-16
Reinstate MiMa and address problems

retronym:


retronym:
The ant and sbt builds on the 2.11.x branch use a version of MiMa that predates the regression.

szeiger:
D'Oh!

adriaanm:
Could you expand on why the scala.reflect.runtime changes are safe? Is there no way these methods could be called from user code? I guess we could classify this under the experimental API exemption, but would still be good to rubber duck debug this and see if there's a worse worst case scenario.

adriaanm:
(I'm on phone, didn't look at their scala source visibility)

retronym:
@adriaanm The only public API in `scala.reflect.runtime` is `.{universe,currentMirror}` [defined](https://github.com/scala/scala/blob/86999ed76b5d9df51c337c5e33fe265e04d8f39c/src/reflect/scala/reflect/runtime/package.scala#L17-L24) in the package object.

The rest should really be in `.runtime.internal` so we could exclude the entire package as implementation, like we do for `reflect.internal._`.

The particular methods excluded here are nested [within `private[scala] JavaMirrors`](https://github.com/scala/scala/blob/795d59a8f600e45ee9b05b483a4d80d2d8ce6de5/src/reflect/scala/reflect/runtime/JavaMirrors.scala#L25-L25).

PR #5534, opened 2016-11-16
SI-10059 reset the `DEFERRED` flag for Java varargs forwarders

lrytz:
When an abstract method is annotated `@varargs`, make sure that the
generated synthetic Java varargs method does not have the `DEFERRED`
flag (`ACC_ABSTRACT` in bytecode).

The flag lead to an NPE in the code generator, because the ASM framework
leaves certain fields `null` for abstract methods (`localVariables` in
this case).

Interestingly this did not crash in 2.11.x: the reason is that the test
whether to emit a method body or not has changed in the 2.12 backend
(in c8e6050).

    val isAbstractMethod = [..] methSymbol.isDeferred [..] // 2.11
    val isAbstractMethod = rhs == EmptyTree                // 2.12

So in 2.11, the varargs forwarder method was actually left abstract in
bytecode, leading to an `AbstractMethodError: T.m([I)I` at run-time.


retronym:
LGTM

PR #5536, opened 2016-11-17
Fix more compiler crashes with fields, refinement types

retronym:
In the same manner as scala/scala-dev#219, the placement of the fields
phase after uncurry is presenting some challenges in keeping our trees
type correct.

This commit whacks a few more moles by adding a casts in the body of
synthetic methods.

Fixes scala/scala-dev#268

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
+++ b/src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
@@ -332,7 +332,7 @@ trait AccessorSynthesis extends Transform with ast.TreeDSL {
 
         val isUnit    = isUnitGetter(lazyAccessor)
         val selectVar = if (isUnit) UNIT         else Select(thisRef, lazyVar)
-        val storeRes  = if (isUnit) rhsAtSlowDef else Assign(selectVar, rhsAtSlowDef)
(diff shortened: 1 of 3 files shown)


retronym:
Review by @adriaanm, 2.12.1 ?

src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
```diff
@@ -332,7 +332,7 @@ trait AccessorSynthesis extends Transform with ast.TreeDSL {
 
         val isUnit    = isUnitGetter(lazyAccessor)
         val selectVar = if (isUnit) UNIT         else Select(thisRef, lazyVar)
-        val storeRes  = if (isUnit) rhsAtSlowDef else Assign(selectVar, rhsAtSlowDef)
+        val storeRes  = if (isUnit) rhsAtSlowDef else Assign(selectVar, fields.castHack(rhsAtSlowDef, lazyVar.info.resultType))
```
adriaanm: Do we need resultType here? I think the lazyVar symbol is synthesized by us (which means we don't need ASF, and straight up info should be good)

retronym: Removed. I probably wasn't sure if this the symbol was a method (with a `NullaryType` to be stripped with `resultType`) or a field when I wrote that...

adriaanm:
LGTM 

s/mole/pancake/

PR #5537, opened 2016-11-17
Optimized HashTable.index

l0rinc:
Added `BitManipulationBenchmark` to measure alternative solutions (`ns/op`, smaller is better):
```java
Benchmark                                                 Mode  Cnt    Score    Error  Units
BitManipulationBenchmark.with2DeBruijn                    avgt   20  252.940 ±  1.968  ns/op
BitManipulationBenchmark.withSumBinSearch                 avgt   20  295.604 ±  2.026  ns/op
BitManipulationBenchmark.withIntegerNumberOfLeadingZeros  avgt   20  298.597 ±  1.155  ns/op
BitManipulationBenchmark.withIntegerBitCount*             avgt   20  323.815 ±  1.351  ns/op
BitManipulationBenchmark.withBinSearch                    avgt   20  330.712 ± 12.232  ns/op
BitManipulationBenchmark.withMatch                        avgt   20  360.977 ±  3.469  ns/op
BitManipulationBenchmark.withLoop                         avgt   20  613.530 ±  9.326  ns/op

```
(*) previous impl

`Integer.numberOfLeadingZeros` was chosen, as it's optimized to a single `LZCNT` assembly call.

---
Benchmarks (`ops/s`, smaller is better)
---

`Before (9c5d3f8)`:
```scala
Benchmark                                     (size)  Mode  Cnt      Score      Error  Units
s.c.immutable.VectorMapBenchmark.groupBy          10  avgt   20    645.594 ±    9.435  ns/op
s.c.immutable.VectorMapBenchmark.groupBy         100  avgt   20   2084.216 ±   37.814  ns/op
s.c.immutable.VectorMapBenchmark.groupBy        1000  avgt   20  19878.481 ±  262.404  ns/op
s.c.mutable.HashMapBenchmark.get                  10  avgt   20    689.941 ±    5.850  ns/op
s.c.mutable.HashMapBenchmark.get                 100  avgt   20   7357.330 ±   45.956  ns/op
s.c.mutable.HashMapBenchmark.get                1000  avgt   20  95767.200 ± 1550.771  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate      10  avgt   20    509.181 ±    2.683  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate     100  avgt   20   5563.301 ±   32.335  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate    1000  avgt   20  71965.365 ± 1809.738  ns/op
s.c.mutable.HashMapBenchmark.put                  10  avgt   20    247.270 ±    3.972  ns/op
s.c.mutable.HashMapBenchmark.put                 100  avgt   20   5646.185 ±  106.172  ns/op
s.c.mutable.HashMapBenchmark.put                1000  avgt   20  81303.663 ±  954.938  ns/op
```

`Changed modulo to bitwise and in hash calculation (4c729fe)`:
```scala
Benchmark                                     (size)  Mode  Cnt      Score      Error  Units
s.c.immutable.VectorMapBenchmark.groupBy          10  avgt   20    631.291 ±    9.269  ns/op
s.c.immutable.VectorMapBenchmark.groupBy         100  avgt   20   2077.885 ±   59.737  ns/op
s.c.immutable.VectorMapBenchmark.groupBy        1000  avgt   20  15458.278 ±  317.347  ns/op
s.c.mutable.HashMapBenchmark.get                  10  avgt   20    678.013 ±    4.453  ns/op
s.c.mutable.HashMapBenchmark.get                 100  avgt   20   7258.522 ±   76.088  ns/op
s.c.mutable.HashMapBenchmark.get                1000  avgt   20  94748.845 ± 1226.120  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate      10  avgt   20    498.042 ±    5.006  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate     100  avgt   20   5243.154 ±  110.372  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate    1000  avgt   20  68194.752 ±  655.436  ns/op
s.c.mutable.HashMapBenchmark.put                  10  avgt   20    257.275 ±    1.411  ns/op
s.c.mutable.HashMapBenchmark.put                 100  avgt   20   5318.532 ±  152.923  ns/op
s.c.mutable.HashMapBenchmark.put                1000  avgt   20  79607.160 ±  651.779  ns/op

```

`Optimized HashTable.index (6cc1504)`:
```scala
Benchmark                                     (size)  Mode  Cnt      Score      Error  Units
s.c.immutable.VectorMapBenchmark.groupBy          10  avgt   20    616.164 ±    4.712  ns/op
s.c.immutable.VectorMapBenchmark.groupBy         100  avgt   20   2034.447 ±   14.495  ns/op
s.c.immutable.VectorMapBenchmark.groupBy        1000  avgt   20  14712.164 ±  119.983  ns/op
s.c.mutable.HashMapBenchmark.get                  10  avgt   20    679.046 ±    6.872  ns/op
s.c.mutable.HashMapBenchmark.get                 100  avgt   20   7242.097 ±   41.244  ns/op
s.c.mutable.HashMapBenchmark.get                1000  avgt   20  95342.919 ± 1521.328  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate      10  avgt   20    488.034 ±    4.554  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate     100  avgt   20   4883.123 ±   59.268  ns/op
s.c.mutable.HashMapBenchmark.getOrElseUpdate    1000  avgt   20  65174.034 ±  496.759  ns/op
s.c.mutable.HashMapBenchmark.put                  10  avgt   20    267.983 ±    1.797  ns/op
s.c.mutable.HashMapBenchmark.put                 100  avgt   20   5097.351 ±  104.538  ns/op
s.c.mutable.HashMapBenchmark.put                1000  avgt   20  78772.540 ±  543.935  ns/op
```

Summary, i.e. the effect of this PR, according to the benchmarks:
* `groupBy` has a `~35%` speedup
* `get` didn't change
* `getOrElseUpdate` has a `~10%` speedup
* `put` has a `~3%` speedup

Note: caching the `exponent` to a local private field (`Byte` or `Int`) didn't have any performance advantage (only a minor slowdown was measured, possibly because it's accessed via an interface now)

---
comparing it to `2.11.8` via
```java
@Benchmark public Object scala_persistent_some() { return scalaPersistent.groupBy(Integer::bitCount); }
```
we get the following (`ops/s`, greater is better):
```java
Benchmark      (CONTAINER_SIZE)   Mode  Cnt       Score      Error  Units
VectorBenchmar               10  thrpt   30  613353.461 ± 5432.843  ops/s
VectorBenchmar              100  thrpt   30  213606.187 ± 5014.126  ops/s
VectorBenchmar             1000  thrpt   30   32443.292 ±  328.302  ops/s
```

`after this PR`:
```java
Benchmark      (CONTAINER_SIZE)   Mode  Cnt       Score      Error  Units
VectorBenchmark              10  thrpt   30  658523.151 ± 6456.227  ops/s
VectorBenchmark             100  thrpt   30  241050.196 ± 1416.711  ops/s
VectorBenchmark            1000  thrpt   30   40219.643 ±  268.688  ops/s
```

i.e. `~20%` faster `Vector.groupBy` than in `2.11.8`

### DIFF
--- a/src/library/scala/collection/mutable/HashTable.scala
+++ b/src/library/scala/collection/mutable/HashTable.scala
@@ -411,58 +411,23 @@ private[collection] object HashTable {
 
     protected def elemHashCode(key: KeyType) = key.##
 
-    protected final def improve(hcode: Int, seed: Int) = {
-      /* Murmur hash
-       *  m = 0x5bd1e995
-       *  r = 24
-       *  note: h = seed = 0 in mmix
-       *  mmix(h,k) = k *= m; k ^= k >> r; k *= m; h *= m; h ^= k; */
-      // var k = hcode * 0x5bd1e995
-      // k ^= k >> 24
-      // k *= 0x5bd1e995
-      // k
-
-      /* Another fast multiplicative hash
-       * by Phil Bagwell
-       *
-       * Comment:
-       * Multiplication doesn't affect all the bits in the same way, so we want to
-       * multiply twice, "once from each side".
-       * It would be ideal to reverse all the bits after the first multiplication,
-       * however, this is more costly. We therefore restrict ourselves only to
-       * reversing the bytes before final multiplication. This yields a slightly
-       * worse entropy in the lower 8 bits, but that can be improved by adding:
-       *
-       * `i ^= i >> 6`
-       *
-       * For performance reasons, we avoid this improvement.
-       * */
-      val i= scala.util.hashing.byteswap32(hcode)
-
-      /* Jenkins hash
-       * for range 0-10000, output has the msb set to zero */
-      // var h = hcode + (hcode << 12)
-      // h ^= (h >> 22)
-      // h += (h << 4)
-      // h ^= (h >> 9)
-      // h += (h << 10)
-      // h ^= (h >> 2)
-      // h += (h << 7)
-      // h ^= (h >> 12)
-      // h
-
-      /* OLD VERSION
-       * quick, but bad for sequence 0-10000 - little entropy in higher bits
-       * since 2003 */
-      // var h: Int = hcode + ~(hcode << 9)
-      // h = h ^ (h >>> 14)
-      // h = h + (h << 4)
-      // h ^ (h >>> 10)
-
-      // the rest of the computation is due to SI-5293
-      val rotation = seed % 32
-      val rotated = (i >>> rotation) | (i << (32 - rotation))
-      rotated
+    /**
+      * Defer to a high-quality hash in [[scala.util.hashing]].
+      * The goal is to distribute across bins as well as possible even if a hash code has low entropy at some bits.
+      * <p/>
+      * OLD VERSION - quick, but bad for sequence 0-10000 - little entropy in higher bits - since 2003
+      * {{{
+      * var h: Int = hcode + ~(hcode << 9)
+      * h = h ^ (h >>> 14)
+      * h = h + (h << 4)
+      * h ^ (h >>> 10)
+      * }}}
+      * the rest of the computation is due to SI-5293
+      */
+    protected final def improve(hcode: Int, seed: Int): Int = {
+      val hash = scala.util.hashing.byteswap32(hcode)
+      val shift = seed & ((1 << 5) - 1)
+      (hash >>> shift) | (hash << (32 - shift))
     }
   }
 
--- a/src/library/scala/collection/mutable/HashTable.scala
+++ b/src/library/scala/collection/mutable/HashTable.scala
@@ -360,14 +360,14 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
 
   protected def elemEquals(key1: A, key2: A): Boolean = (key1 == key2)
 
-  // Note:
-  // we take the most significant bits of the hashcode, not the lower ones
-  // this is of crucial importance when populating the table in parallel
-  protected final def index(hcode: Int) = {
+  /**
+    * Note: we take the most significant bits of the hashcode, not the lower ones
+    * this is of crucial importance when populating the table in parallel
+    */
+  protected final def index(hcode: Int): Int = {
     val ones = table.length - 1
-    val improved = improve(hcode, seedvalue)
-    val shifted = (improved >> (32 - java.lang.Integer.bitCount(ones))) & ones
-    shifted
+    val exponent = Integer.numberOfLeadingZeros(ones)
+    (improve(hcode, seedvalue) >>> exponent) & ones
   }
 
   protected def initWithContents(c: HashTable.Contents[A, Entry]) = {
--- a/test/benchmarks/src/main/scala/scala/BitManipulationBenchmark.scala
+++ b/test/benchmarks/src/main/scala/scala/BitManipulationBenchmark.scala
@@ -1,0 +1,170 @@
+package scala.collection
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra._
+import org.openjdk.jmh.runner.IterationType
+import benchmark._
+import java.util.concurrent.TimeUnit
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(2)
+@Threads(1)
+@Warmup(iterations = 10)
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class BitManipulationBenchmark {
+  val powersOfTwo = Array(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, 33554432, 67108864, 134217728, 268435456, 536870912, 1073741824)
+
+  //////////////////////////////////////////////
+
+  @Benchmark def withIntegerBitCount(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withIntegerBitCount(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  private def withIntegerBitCount(v: Int) = Integer.SIZE - Integer.bitCount(v - 1)
+
+  //////////////////////////////////////////////
+
+  @Benchmark def withIntegerNumberOfLeadingZeros(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withIntegerNumberOfLeadingZeros(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  private def withIntegerNumberOfLeadingZeros(v: Int) = Integer.numberOfLeadingZeros(v - 1)
+
+  //////////////////////////////////////////////
+
+  @Benchmark def withLoop(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withLoop(v)
+      bh.consume(leadingZeros)
+    }
+  }
+
+  private def withLoop(v: Int): Int = {
+    var r = Integer.SIZE
+    var copy = v >> 1
+    while (copy != 0) {
+      r -= 1
+      copy = copy >> 1
+    }
+    r
+  }
+
+  //////////////////////////////////////////////
+
+  @Benchmark def withMatch(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withMatch(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  private def withMatch(i: Int) = i match {
+    case 1 => 32
+    case 2 => 31
+    case 4 => 30
+    case 8 => 29
+    case 16 => 28
+    case 32 => 27
+    case 64 => 26
+    case 128 => 25
+    case 256 => 24
+    case 512 => 23
+    case 1024 => 22
+    case 2048 => 21
+    case 4096 => 20
+    case 8192 => 19
+    case 16384 => 18
+    case 32768 => 17
+    case 65536 => 16
+    case 131072 => 15
+    case 262144 => 14
+    case 524288 => 13
+    case 1048576 => 12
+    case 2097152 => 11
+    case 4194304 => 10
+    case 8388608 => 9
+    case 16777216 => 8
+    case 33554432 => 7
+    case 67108864 => 6
+    case 134217728 => 5
+    case 268435456 => 4
+    case 536870912 => 3
+    case 1073741824 => 2
+  }
+
+
+  //////////////////////////////////////////////
+
+  @Benchmark def with2DeBruijn(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = with2DeBruijn(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  // https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn
+  private val multiplyDeBruijnBitPosition2 = Array(32, 31, 4, 30, 3, 18, 8, 29, 2, 10, 12, 17, 7, 15, 28, 24, 1, 5, 19, 9, 11, 13, 16, 25, 6, 20, 14, 26, 21, 27, 22, 23)
+
+  private def with2DeBruijn(v: Int) = multiplyDeBruijnBitPosition2((v * 0x077CB531) >>> 27)
+
+
+  //////////////////////////////////////////////
+
+  @Benchmark def withBinSearch(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withBinSearch(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  private def withBinSearch(v: Int) =
+    if (v < 65536) if (v < 256) if (v < 16) if (v < 4) if (v == 1) 32 else 31
+    else if (v == 4) 30 else 29
+    else if (v < 64) if (v == 16) 28 else 27
+    else if (v == 64) 26 else 25
+    else if (v < 4096) if (v < 1024) if (v == 256) 24 else 23
+    else if (v == 1024) 22 else 21
+    else if (v < 16384) if (v == 4096) 20 else 19
+    else if (v == 16384) 18 else 17
+    else if (v < 16777216) if (v < 1048576) if (v < 262144) if (v == 65536) 16 else 15
+    else if (v == 262144) 14 else 13
+    else if (v < 4194304) if (v == 1048576) 12 else 11
+    else if (v == 4194304) 10 else 9
+    else if (v < 268435456) if (v < 67108864) if (v == 16777216) 8 else 7
+    else if (v == 67108864) 6 else 5
+    else if (v < 1073741824) if (v == 268435456) 4 else 3
+    else if (v == 1073741824) 2 else 1
+
+  //////////////////////////////////////////////
+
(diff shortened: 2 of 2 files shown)


test/benchmarks/build.sbt
```diff
@@ -1,11 +1,11 @@
 scalaHome := Some(file("../../build/pack"))
-scalaVersion := "2.12.0-dev"
+scalaVersion := "2.12.1-dev"
```
l0rinc, line 2: these will disappear once https://github.com/scala/scala/pull/5528 is merged

test/benchmarks/src/main/scala/scala/BitManipulationBenchmark.scala
```diff
@@ -0,0 +1,170 @@
...
+@State(Scope.Benchmark)
+class BitManipulationBenchmark {
+  val powersOfTwo = Array(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, 33554432, 67108864, 134217728, 268435456, 536870912, 1073741824)
+
+  //////////////////////////////////////////////
+
+  def withIntegerBitCount(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withIntegerBitCount(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
```
l0rinc, line 24: asserts are meant for correctness checking, but there is no easy way to disable assertions yet

test/benchmarks/src/main/scala/scala/BitManipulationBenchmark.scala
```diff
@@ -0,0 +1,170 @@
...
+      val leadingZeros = with2DeBruijn(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  // https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn
+  val multiplyDeBruijnBitPosition2 = Array(32, 31, 4, 30, 3, 18, 8, 29, 2, 10, 12, 17, 7, 15, 28, 24, 1, 5, 19, 9, 11, 13, 16, 25, 6, 20, 14, 26, 21, 27, 22, 23)
+
+  def with2DeBruijn(v: Int) = multiplyDeBruijnBitPosition2((v * 0x077CB531) >>> 27)
```
l0rinc: this turned out to be the fastest impl

test/benchmarks/src/main/scala/scala/BitManipulationBenchmark.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  @Benchmark def withSumBinSearch(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withSumBinSearch(v)
+      // assert(leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  def withSumBinSearch(v: Int): Int = {
```
l0rinc: tried this with the last two lines replaced with a lookup table -> no change

test/benchmarks/src/main/scala/scala/BitManipulationBenchmark.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  def withBinSearch(bh: Blackhole) {
+    for (v <- powersOfTwo) {
+      val leadingZeros = withBinSearch(v)
+      // assert (leadingZeros == withLoop(v), s"$leadingZeros != ${withLoop(v)} ($v)")
+      bh.consume(leadingZeros)
+    }
+  }
+
+  def withBinSearch(v: Int) =
```
l0rinc: I'm very proud of this contraption :D

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -363,11 +363,11 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
...
   // this is of crucial importance when populating the table in parallel
+  // See: https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn for explanation and BitManipulationBenchmark for alternatives
+  private val exponents = Array(32, 31, 4, 30, 3, 18, 8, 29, 2, 10, 12, 17, 7, 15, 28, 24, 1, 5, 19, 9, 11, 13, 16, 25, 6, 20, 14, 26, 21, 27, 22, 23)
   protected final def index(hcode: Int) = {
-    val ones = table.length - 1
-    val improved = improve(hcode, seedvalue)
-    val shifted = (improved >> (32 - java.lang.Integer.bitCount(ones))) & ones
-    shifted
+    val exponent = exponents((table.length * 0x077CB531) >>> 27)
+    (improve(hcode, seedvalue) >>> exponent) & (table.length - 1)
```
sjrd: Hum ... this code (before _and_ after) seems to assume that `table.length` is an exact power of 2. I believe this is assumed elsewhere in the implementation.

If that is true, then `exponent` is in fact equal to `java.lang.Integer.numberOfTrailingZeros(table.length)`. Have you tried that? Or also `31 - java.lang.Integer.numberOfLeadingZeros(table.length)`?

l0rinc: > assume that table.length is an exact power of 2

Indeed, https://github.com/scala/scala/pull/5537/commits/d6778238fe91d8dbb9d13a42afb5113ed6c96526#diff-a24c1e91e1a44ff65f3da12201465cebR154 resizes this to be a power of two always (please verify).

> numberOfTrailingZeros(table.length). Have you tried that? 

Yes, please check out the benchmarks below:
https://github.com/scala/scala/pull/5537/commits/d6778238fe91d8dbb9d13a42afb5113ed6c96526#diff-9ee4c1bdb2df6f45b8c03d48dd2b0b0dR41

sjrd: > Yes, please check out the benchmarks below:

Oops, I had actually had a look at the benchmarks, but somehow my eyes did not see the one I was looking for ^^

Ichoran:
This is fun--and I can't imagine the array lookup is actually faster than a proper LZCNT assembly instruction, so it indicates that the JIT compiler isn't actually emitting ideal assembly.  That suggests to me that we shouldn't bake in a table lookup because (1) if the JIT compiler improves the table will be slower than the alternatives, and (2) if the loop is less hot, the table lookup will be more expensive since it won't be hanging out in L1 cache the entire time.

But why not store tableLengthLeadingZeros and update it only when the table size changes (i.e. rarely)?  That should be faster yet.  Yes, it takes an extra 4 bytes of storage, but HashMap isn't exactly careful with how much it allocates as it is.

l0rinc:
Valid concerns, @Ichoran!
I pushed a version with the built-in leading zeros counter, and will verify whether `HashMap`'s object layout permits an additional `byte` for free, and post the benchmark results here, so that we can take a informed decision :).
Thanks for all your reviews! :)

---

Edit: The good news: adding an `exponent: Byte` doesn't modify the object size, displayed by JOL: new HashMap -> 120 bytes.
The bad news: somehow most methods got slightly slower (except for `groupBy`). Will try to investigate tomorrow :/

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -80,7 +80,7 @@ sealed class HashMap[A, +B] extends AbstractMap[A, B]
 
   protected def elemHashCode(key: A) = key.##
 
-  protected final def improve(hcode: Int) = {
+  protected final def improve(hcode: Int) = { // TODO unify with HashUtils.improve
```
l0rinc: @Ichoran, @retronym, should I unify the two, leave the comment or delete it?

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +411,10 @@ private[collection] object HashTable {
...
-      // h = h + (h << 4)
-      // h ^ (h >>> 10)
-
-      // the rest of the computation is due to SI-5293
-      val rotation = seed % 32
-      val rotated = (i >>> rotation) | (i << (32 - rotation))
-      rotated
+    protected final def improve(hcode: Int, seed: Int): Int = {
+      val i = scala.util.hashing.byteswap32(hcode)
+      val rotation = seed & ((1 << 5) - 1)
```
l0rinc: `x % 32` is the same as `x & 31` for positive numbers (which should be the case since we're using it for bit shifts below), just faster

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -363,11 +363,10 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
   // Note:
   // we take the most significant bits of the hashcode, not the lower ones
   // this is of crucial importance when populating the table in parallel
-  protected final def index(hcode: Int) = {
+  protected final def index(hcode: Int): Int = {
     val ones = table.length - 1
-    val improved = improve(hcode, seedvalue)
-    val shifted = (improved >> (32 - java.lang.Integer.bitCount(ones))) & ones
-    shifted
+    val exponent = Integer.numberOfLeadingZeros(ones)
```
paplorinc: * `numberOfLeadingZeros` might be optimized in the future and is already fast enough for now.
* for some reason extracting it as a variable didn't improve the speed at all :/

src/library/scala/collection/immutable/HashMap.scala
Ichoran: I don't think we should change the hashing scheme in point releases for 2.12.  Leave it alone until 2.13.

src/library/scala/collection/mutable/HashTable.scala
Ichoran: You've deleted all the discussion about hash code refinement.  Could you preserve enough so we have some idea of the thought process behind it (record of what worked & didn't)?  This is an important choice, so it's nice to have some documentation of it.

src/library/scala/collection/immutable/HashMap.scala
l0rinc: Sure, is it ok if I leave the comment here?

src/library/scala/collection/mutable/HashTable.scala
l0rinc: In other cases the history of a given method was stored in git, the method documentation or in the issue, corresponding to the change.

Here it seems like there are leftover comments that have nothing to do with the current state, which I found very confusing - especially since the code itself is not very complicated, though a documentation link would be welcome.

Could you please help me out in deciding which parts to keep and which to bury in git history instead?

Ichoran: The warning about the old bad algorithm seems particularly pertinent.  Anything that says, "We already did this but it was bad", and it isn't obvious from inspection that it is bad, should be preserved.  Everything else can go.

l0rinc: Thank you @Ichoran, reformatted the old comments as method comment, with separated code parts, hope you like it this way :)

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +410,61 @@ private[collection] object HashTable {
...
-      // h = h ^ (h >>> 14)
-      // h = h + (h << 4)
-      // h ^ (h >>> 10)
-
-      // the rest of the computation is due to SI-5293
-      val rotation = seed % 32
-      val rotated = (i >>> rotation) | (i << (32 - rotation))
-      rotated
+    /** Murmur hash
+     * {{{
```
Ichoran: I really don't think we need most of this.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +410,61 @@ private[collection] object HashTable {
...
+     *  m = 0x5bd1e995
+     *  r = 24
+     *  note: h = seed = 0 in mmix
+     *  mmix(h,k) = k *= m; k ^= k >> r; k *= m; h *= m; h ^= k;
+     *
+     *  var k = hcode * 0x5bd1e995
+     *  k ^= k >> 24
+     *  k * 0x5bd1e995
+     * }}}
```
Ichoran: This code doesn't explain whether this hash is good or bad.  Can discard it.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +410,61 @@ private[collection] object HashTable {
...
+     * Multiplication doesn't affect all the bits in the same way, so we want to
+     * multiply twice, "once from each side".
+     * It would be ideal to reverse all the bits after the first multiplication,
+     * however, this is more costly. We therefore restrict ourselves only to
+     * reversing the bytes before final multiplication. This yields a slightly
+     * worse entropy in the lower 8 bits, but that can be improved by adding:
+     * {{{
+     *  i ^= i >> 6
+     * }}}
+     * For performance reasons, we avoid this improvement.
```
Ichoran: We're not using a multiplicative hash any more, so this lengthy explanation is obsolete.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +410,61 @@ private[collection] object HashTable {
...
+     * {{{
+     *  var h = hcode + (hcode << 12)
+     *  h ^= (h >> 22)
+     *  h += (h << 4)
+     *  h ^= (h >> 9)
+     *  h += (h << 10)
+     *  h ^= (h >> 2)
+     *  h += (h << 7)
+     *  h ^  (h >> 12)
+     * }}}
```
Ichoran: Again, doesn't say anything about whether the Jenkins Hash is good or bad.  Can get rid of it.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +410,61 @@ private[collection] object HashTable {
...
+     * OLD VERSION
+     * quick, but bad for sequence 0-10000 - little entropy in higher bits
+     * since 2003
+     * {{{
+     * var h: Int = hcode + ~(hcode << 9)
+     * h = h ^ (h >>> 14)
+     * h = h + (h << 4)
+     * h ^ (h >>> 10)
+     * }}}
```
Ichoran: This part is important: a fast and small algorithm that we actually used but in practice works poorly.  "Don't do this!"  Leave this in.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -411,58 +410,61 @@ private[collection] object HashTable {
...
+     *
+     * the rest of the computation is due to SI-5293
```
Ichoran: Have you checked if this is true?  In any case, might just say, "Defer to a high-quality hash in `scala.util.hashing`.  The goal is to distribute across bins as well as possible even if a hash code has low entropy at some bits."

l0rinc: It wasn't part of my PR to come up with a new hashing, I haven't questioned the validity of the code or the comments.
Would gladly do it, if you think I can be of any service in this area :)

Applied your comments, thank you for your insight!

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -360,14 +360,14 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
...
+  /**
+    * Note: we take the most significant bits of the hashcode, not the lower ones
+    * this is of crucial importance when populating the table in parallel
+    */
+  protected final def index(hcode: Int): Int = {
     val ones = table.length - 1
-    val improved = improve(hcode, seedvalue)
-    val shifted = (improved >> (32 - java.lang.Integer.bitCount(ones))) & ones
-    shifted
+    val exponent = Integer.numberOfLeadingZeros(ones)
```
l0rinc, line 369: @Ichoran, `numberOfLeadingZeros` call is optimized to `LZCNT` assembly call (see http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8045398):
```asm
0x00007f65791cdbc5: mov    0x10(%r13,%rbp,4),%r10d
0x00007f65791cdbca: lzcnt  %r10d,%r10d        ;*invokestatic numberOfLeadingZeros
                                              ; - javaslang.collection.VectorBenchmark$Test::scala_withIntegerNumberOfLeadingZeros@25 (line 33)
```

It remains an open question why `DeBruijn` is faster in the benchmarks, though.
BTW, extracting it as a variable didn't improve the speed at all

Ichoran (approved):
LGTM

PR #5538, opened 2016-11-17
Providing minimalistic & controlled kind-polymorphism [ci: last-only]

mandubian:
> This is a PoC for supporting minimalistic & very controlled KindPolymorphism in Scala based on original work by @milessabin.

## Motivations
The idea is NOT to provide universal kind-polymorphism for all possible kinds which would be impossible and a bad idea.
We want to provide a limited kind-polymorphism controlled through typeclass implicits defining the accepted level of kind-polymorphism in a determined scope.

Our first motivations is to be able to manipulate kind of types (comparing, extracting etc...) and naturally to build structures that can abstract on the kind. This is meant to build higher-level tooling "à la shapeless".

## Example

A basic example is:

```scala
trait Foo[T <: AnyKind] { type Out ; def id(t: Out): Out = t }

  object Foo {
    implicit def foo0[T] = new Foo[T] { type Out = T }
    implicit def foo1[T[_]] = new Foo[T] { type Out = T[Any] }
    implicit def foo2[T[_, _]] = new Foo[T] { type Out = T[Any, Any] }
  }

  def foo[T <: AnyKind](implicit f: Foo[T]): f.type = f
  foo[Int].id(23)
  foo[List].id(List[Any](1, 2, 3))
  foo[Map].id(Map[Any, Any](1 -> "toto", 2 -> "tata", 3 -> "tutu"))
```

## Requirements

To make it possible, we need to provide:
- A flag `-Ykind-polymorphism` to activate the feature in the compiler
- A new technical type bound `<: AnyKind` to indicate a type is _kind polymorphic_

With some conditions:
- When `-Ykind-polymorphism` isn't activated, it shouldn't bother other features of scala at all.
- `AnyKind` shouldn't be used as a real type.

### `AnyKind` for type parameter

```scala
// T is <: AnyKind here because this is a kind-polymorphic structure
trait Foo[T <: AnyKind] { ... }
object Foo {
    // T and T[_] here aren't <: AnyKind but just types used to build a specific kinded instance of
    // a kind-polymorphic structure.
    implicit def foo0[T] = new Foo[T] { ... }
    implicit def foo1[T[_]] = new Foo[T] { ... }
}
// T is <: AnyKind because this is a kind-polymorphic function
def foo[T <: AnyKind](implicit f: Foo[T]) = ???
```

### `AnyKind` for path-dependent types

```scala
trait Kinder[MA] { type M <: AnyKind }
  object Kinder extends KinderLowerImplicits {
    type Aux[MA, M0 <: AnyKind] = Kinder[MA] { type M = M0 }

    implicit def kinder2[M0[_, _], A0, B0]: Kinder.Aux[M0[A0, B0], M0] = new Kinder[M0[A0, B0]] {  
        // M becomes a M[_, _] here
        type M[t, u] = M0[t, u]
    }

    implicit def kinder1[M0[_], A0]: Kinder.Aux[M0[A0], M0] = new Kinder[M0[A0]] {
        // M becomes a M[_] here
        type M[t] = M0[t]
    }
  }

  trait KinderLowerImplicits {
    // defines in a lower implicit priority to prevent implicit collisions
    implicit def kinder0[A]: Kinder.Aux[A, A] = new Kinder[A] {
        type M = A
    }    
  }
```

### global mechanism

When (and only when) `Ykind-polymorphism` is activated, if the typer encounters a type with `<: AnyKind`, it doesn't compare type parameters length as strictly as default Scala behavior but defers a bit the strictness checking.
The idea is to lose NO strictness in typing but to delay this strictness a bit later in the compiling chain.

Naturally, the `<: Anykind` is purely technical and is completely eliminated at erasure.

### the `isAnyKind` condition
We have added a condition to be checked in a very few places  of the coed: https://github.com/mandubian/scala/blob/topic/kind-poly-v1/src/reflect/scala/reflect/internal/Types.scala#L4379-L4382.

The main difficulty was to check that our type is subclassing `AnyKind`. But it appears that the call to `isNonBottomSubClass` on an _incomplete_ type can trigger `CyclicError` or `TypeError`.

> in `isAnyKind`, we have added checks only meant to prevent those exceptions to happen. If a better solution is possible, don't hesitate to propose it.

### Forbid using AnyKind as a type

We want the following cases to be forbidden:
https://github.com/mandubian/scala/blob/topic/kind-poly-v1/test/files/neg/kind-poly.scala#L20-L26

To strictly forbid those cases and provide user-friendly error messages, 2 new error cases have been added.

> If you have better ways to check that, don't hesitate to propose once again.

### More samples

Check this test file : https://github.com/mandubian/scala/blob/topic/kind-poly-v1/test/files/pos/kind-poly.scala
(with a kind-polymorphic list implementation)

### Next steps

- Explore what can be done & not with this feature as it is quite promising and already working so well.
- Check more edge-cases
- Study potentially `typeOf` taking into account kind polymorphism.
- More ideas

### DIFF
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -385,6 +385,9 @@ trait Definitions extends api.StandardDefinitions {
     lazy val JavaRepeatedParamClass = specialPolyClass(tpnme.JAVA_REPEATED_PARAM_CLASS_NAME, COVARIANT)(tparam => arrayType(tparam.tpe))
     lazy val RepeatedParamClass     = specialPolyClass(tpnme.REPEATED_PARAM_CLASS_NAME, COVARIANT)(tparam => seqType(tparam.tpe))
 
+    // TODO : requires ABSTRACT | TRAIT | FINAL ????
+    lazy val AnyKindClass   = enterNewClass(ScalaPackageClass, tpnme.AnyKind, AnyTpe :: Nil, ABSTRACT | TRAIT) markAllCompleted
+
     def isByNameParamType(tp: Type)        = tp.typeSymbol == ByNameParamClass
     def isScalaRepeatedParamType(tp: Type) = tp.typeSymbol == RepeatedParamClass
     def isJavaRepeatedParamType(tp: Type)  = tp.typeSymbol == JavaRepeatedParamClass
@@ -756,7 +759,7 @@ trait Definitions extends api.StandardDefinitions {
     def isStable(tp: Type): Boolean = tp match {
       case _: SingletonType                             => true
       case NoPrefix                                     => true
-      case TypeRef(_, NothingClass | SingletonClass, _) => true
+      case TypeRef(_, NothingClass | SingletonClass | AnyKindClass, _) => true
       case TypeRef(_, sym, _) if sym.isAbstractType     => tp.bounds.hi.typeSymbol isSubClass SingletonClass
       case TypeRef(pre, sym, _) if sym.isModuleClass    => isStable(pre)
       case TypeRef(_, _, _) if tp ne tp.dealias         => isStable(tp.dealias)
@@ -1360,7 +1363,8 @@ trait Definitions extends api.StandardDefinitions {
       AnyValClass,
       NullClass,
       NothingClass,
-      SingletonClass
+      SingletonClass,
+      AnyKindClass
     )
     /** Lists core methods that don't have underlying bytecode, but are synthesized on-the-fly in every reflection universe */
     lazy val syntheticCoreMethods = List(
--- a/bincompat-forward.whitelist.conf
+++ b/bincompat-forward.whitelist.conf
@@ -42,6 +42,10 @@ filter {
     {
       matchName="scala.util.hashing.MurmurHash3.wrappedArrayHash"
       problemName=DirectMissingMethodProblem
+    },
+    {
+      matchName="scala.reflect.runtime.Settings.YkindPolymorphism"
+      problemName=DirectMissingMethodProblem
     }
   ]
 }
--- a/src/compiler/scala/reflect/reify/Reifier.scala
+++ b/src/compiler/scala/reflect/reify/Reifier.scala
@@ -110,7 +110,7 @@ abstract class Reifier extends States
       // needs to be solved some day
       // upd. a new hope: https://groups.google.com/forum/#!topic/scala-internals/TtCTPlj_qcQ
       var importantSymbols = Set[Symbol](
-        NothingClass, AnyClass, SingletonClass, PredefModule, ScalaRunTimeModule, TypeCreatorClass, TreeCreatorClass, MirrorClass,
+        NothingClass, AnyClass, SingletonClass, AnyKindClass, PredefModule, ScalaRunTimeModule, TypeCreatorClass, TreeCreatorClass, MirrorClass,
         ApiUniverseClass, JavaUniverseClass, ReflectRuntimePackage, runDefinitions.ReflectRuntimeCurrentMirror)
       importantSymbols ++= importantSymbols map (_.companionSymbol)
       importantSymbols ++= importantSymbols map (_.moduleClass)
--- a/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
@@ -30,7 +30,7 @@ trait ScalaSettings extends AbsScalaSettings
   protected def defaultClasspath = sys.env.getOrElse("CLASSPATH", ".")
 
   /** Enabled under -Xexperimental. */
-  protected def experimentalSettings = List[BooleanSetting](YpartialUnification)
+  protected def experimentalSettings = List[BooleanSetting](YpartialUnification, YkindPolymorphism)
 
   /** Enabled under -Xfuture. */
   protected def futureSettings = List[BooleanSetting]()
@@ -217,6 +217,7 @@ trait ScalaSettings extends AbsScalaSettings
   val YdisableFlatCpCaching  = BooleanSetting    ("-YdisableFlatCpCaching", "Do not cache flat classpath representation of classpath elements from jars across compiler instances.")
   val YpartialUnification = BooleanSetting ("-Ypartial-unification", "Enable partial unification in type constructor inference")
   val Yvirtpatmat     = BooleanSetting    ("-Yvirtpatmat", "Enable pattern matcher virtualization")
+  val YkindPolymorphism = BooleanSetting ("-Ykind-polymorphism", "Enable kind polymorphism")
 
   val exposeEmptyPackage = BooleanSetting ("-Yexpose-empty-package", "Internal only: expose the empty package.").internalOnly()
   val Ydelambdafy        = ChoiceSetting  ("-Ydelambdafy", "strategy", "Strategy used for translating lambdas into JVM code.", List("inline", "method"), "method")
--- a/src/compiler/scala/tools/nsc/transform/Erasure.scala
+++ b/src/compiler/scala/tools/nsc/transform/Erasure.scala
@@ -146,7 +146,7 @@ abstract class Erasure extends InfoTransform
         if (sym == ArrayClass && args.nonEmpty)
           if (unboundedGenericArrayLevel(tp1) == 1) ObjectTpe
           else mapOver(tp1)
-        else if (sym == AnyClass || sym == AnyValClass || sym == SingletonClass)
+        else if (sym == AnyClass || sym == AnyValClass || sym == SingletonClass || sym == AnyKindClass)
           ObjectTpe
         else if (sym == UnitClass)
           BoxedUnitTpe
@@ -308,7 +308,7 @@ abstract class Erasure extends InfoTransform
             assert(!sym.isAliasType, "Unexpected alias type: " + sym)
             "" + TVAR_TAG + sym.name + ";"
           }
-          else if (sym == AnyClass || sym == AnyValClass || sym == SingletonClass)
+          else if (sym == AnyClass || sym == AnyValClass || sym == SingletonClass || sym == AnyKindClass)
             jsig(ObjectTpe)
           else if (sym == UnitClass)
             jsig(BoxedUnitTpe)
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -215,7 +215,6 @@ trait ContextErrors {
         }
         assert(!foundType.isErroneous, s"AdaptTypeError - foundType is Erroneous: $foundType")
         assert(!req.isErroneous, s"AdaptTypeError - req is Erroneous: $req")
-
         issueNormalTypeError(callee, withAddendum(callee.pos)(typeErrorMsg(foundType, req)))
         infer.explainTypes(foundType, req)
       }
@@ -681,6 +680,38 @@ trait ContextErrors {
 
       def MissingTypeParametersError(tree: Tree) = {
         issueNormalTypeError(tree, tree.symbol+" takes type parameters")
+        setError(tree)
+      }
+
+      def InferredReturnTypeError(tree: Tree, pt: Type) = {
+        issueNormalTypeError(tree, s"inferred result type ($pt) takes type parameters")
+        setError(tree)
+      }
+
+      def AnyKindTypeError(tree: Tree) = {
+        issueNormalTypeError(tree, s"AnyKind can't be used as a real type")
+        setError(tree)
+      }
+
+      def KindPolymorphicKindArityMismatchError(tree: Tree, pt: Type, targs: List[Type], badTParamArgs: List[(Symbol, Type)], appliedType: Type) = {
+        def typeKindStr(tp: Type): String = {
+          if(tp.typeParams.isEmpty) "*"
+          else tp.typeParams.map(p => symKindStr(p)+"->").mkString("(", "", "*)")
+        }
+
+        def symKindStr(tp: Symbol): String = {
+          if(tp.typeParams.isEmpty) "*"
+          else tp.typeParams.map(p => symKindStr(p)+"->").mkString("(", "", "*)")
+        }
+
+        issueNormalTypeError(tree, {
+          val kindStrs = badTParamArgs.map { case (tparam, targ) =>
+            val kind1 = symKindStr(tparam)
+            val kind2 = typeKindStr(targ)
+            s"${tparam}[$kind1]<=>$targ[$kind2]"
+          }
+          s"[Kind-Polymorphic Error] $tree inferred to type ${pt} with illegally kinded type arguments ${targs.mkString(", ")} applied to type ${appliedType} ${if(kindStrs.nonEmpty) kindStrs.mkString("(kind errors: ", ", ", ")") else ""}"
+        })
         setError(tree)
       }
 
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -468,7 +468,10 @@ abstract class RefChecks extends Transform {
             overrideError("cannot be used here - term macros cannot override abstract methods")
           } else if (other.isTermMacro && !member.isTermMacro) { // (1.10)
             overrideError("cannot be used here - only term macros can override term macros")
-          } else {
+          } else if (other.tpe.isComplete && other.tpe.typeSymbol.isNonBottomSubClass(definitions.AnyKindClass)) {
+            // a AnyKind bounded type can be overriden by a type not bounded by AnyKind (this is the purpose of it)
+          }
+            else {
             checkOverrideTypes()
             checkOverrideDeprecated()
             if (settings.warnNullaryOverride) {
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -170,7 +170,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
           val res =
             if (paramFailed || (paramTp.isErroneous && {paramFailed = true; true})) SearchFailure
-            else inferImplicitFor(paramTp, fun, context, reportAmbiguous = context.reportErrors)
+            else {
+              inferImplicitFor(paramTp, fun, context, reportAmbiguous = context.reportErrors)
+            }
+          
           argResultsBuff += res
 
           if (res.isSuccess) {
@@ -590,7 +593,6 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
      *  4. Give getClass calls a more precise type based on the type of the target of the call.
      */
     protected def stabilize(tree: Tree, pre: Type, mode: Mode, pt: Type): Tree = {
-
       // Side effect time! Don't be an idiot like me and think you
       // can move "val sym = tree.symbol" before this line, because
       // inferExprAlternative side-effects the tree's symbol.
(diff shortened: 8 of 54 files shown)


adriaanm:
So, this is more like dynamically kinded types than polymorphic kinds, right?

mandubian:
Yes if you prefer to call it that way...
or `ad-hoc kind polymorphism`...
The idea being that we control exactly where and how it can be "kind"-polymorphic!

milessabin:
@adriaanm a type variable which is `<: AnyKind` can be instantiated by a type of any kind. But any given type will only ever have a single kind.

adriaanm:
I don't think that's how I understand polymorphism in this context, which I'd assume to mean parametric polymorphism. I'm not against having some kind of dynamic kind, but we should be clear that it basically turns off kind checking.

adriaanm:
If we make the type-level analogy, this would basically be the `Dynamic` type, not `Any` -- you can't use `Any` to allow applying a value to an arbitrary number of arguments.

mandubian:
It turns off kind checking but in a very scoped & restricted way. That's why I prefer speaking about "delaying kind checking".
But I completely agree on your ideas and using a vocabulary that fits better (specially in the context of Scala) sounds good to me. (I wouldn't say I'm an expert in type theory or compiler, anyway)

mandubian:
About `this would basically be the Dynamic type, not Any`, do you mean it would be better to call it `DynamicKind` than `AnyKind` for example?

adriaanm:
Yes, something other than `AnyKind` would be my preference. `DynKind` ?

mandubian:
to be complete, it was `KindPolymorphic` at beginning which seemed a bit too much ;)
`DynKind` would be OK to me...
@miles WDYT?

milessabin:
I think there's a fairly clearcut sense in which this is polymorphism. In,

``` scala
def foo[T <: AnyKind] ...
```

the type variable `T` can be instantiated at any type irrespective of kind, ie. `Int`, `List`, `Map` etc. The instantiating kind is fixed at the call site where `T` is instantiated. This fits pretty well with my understanding of polymorphism. I don't really see this as being "dynamic" any more than I would see,

``` scala
def foo[T[_]] ...
```

as being "dynamic" because it can be instantiated at any type of kind `* -> *`.

But I'm not fussy about the name. We started out with `KindPolymorphic`, and I suggested `AnyKind` because it was terser but also because it seemed usefully consistent with `Any`, `AnyVal`, `AnyRef` etc. but I wouldn't want that to get in the way of the essentials of the feature.

retronym:
"Kind" in scala encompasses variance and bounds in addition to arity. How does this change (and your motivating example) deal with variability along those axes?

milessabin:
I believe that we have to abstract over bounds and variance as well as arity. As with arity, additional constraints would be picked up contextually via implicits.

In case it's not clear, all that can be done with an `<: AnyKind` type variable is propagate it to another `<: AnyKind` type variable. Typically the end of the line will be a type class instance which determines arity, bounds, variance etc. at the point of use.

mandubian:
About bounds & variance, I have not usecase for now but as @milessabin said, I don't see why additional constraints wouldn't be taken into account.
Yet, there are certainly a few cases to test because I haven't tested them for now...

retronym:
> Typically the end of the line will be a type class instance which determines arity, bounds, variance etc. at the point of use.

In the example above, the author of typeclass instances (`foo0`, ...) is responsible for enumerating the set of supported kinds. You can present a single method `foo` to the user (which is good), but you still need some level of prescience to enumerate a big enough set of the kinds (which is bad). I suppose the user can provide an instance if it is missing, though.

Could you build on the motivating example to make it, erm, more motivating? What is some problem that is klunky/impossible today that is elegant/possible tomorrow?

mandubian:
@retronym My first idea was to study if I can use this feature to build data structures that are abstracting on the kind (typically a Product, a Coproduct that work for _, *->_, _->_->\* etc...) because I need that to manipulate DSL in Free monads in more generic way. @milessabin is studying these abstraction ideas also on his side too.

Thus, this is a step forward to enable studying these concepts. But I feel like it might lead to something more too exactly like Shapeless has done (you know, serendipity ;)).
For example, the `SameKind[F <: AnyKind, G <: AnyKind]` is already interesting and gave me a few ideas to dig...

The other aspect to keep in mind is that we do not try to provide a universal kind-polymorphism in Scala because it wouldn't work & would lead to crazy cases (exactly as SI2712 patch would have lead if we hadn't limited it). This approach based on implicits allows to control/constrain/scope precisely the way we can apply "kind-polymorphism/dynamics" in our code. At same time, you can mix those implicits with other proofs/behaviors/rules (see the kind polymorphic list sample for example)...

IMHO, the level of prescience is the same as any implicit use-case, it doesn't appear in your function call but it doesn't mean you don't know what you put in your scope (or you're worrying)...
Finally, I think this feature will be first useful to those building libraries through abstractions and as a library builder, it opens new horizons that I'm eager to explore...

milessabin:
@retronym immediate motivating examples would be `=:=`, `<:<`, `typeOf`, `TypeTag`, `WeakTypeTag`, `ClassTag` and shapeless's `Typeable`.

As @mandubian suggests there are a lot of ideas floating around in a lot of Typelevel projects which could potentially benefit from this ... we'd like to be see just how far we can get with this minimalistic approach combined with well understood type level programming techniques.

sellout:
Cats has already started creating type classes, like `MonoidK`, etc. that attempt to provide higher-kinded versions of other type classes. In slamdata/matryoshka, we’ve started to duplicate all of our type classes as well, which has led to needing even more duplication of Cats’ type classes – CategoryK, etc. (in fact, both @mandubian and I had independently started on this work, and when @milessabin discovered it, he was like “we need to avoid doing this – can we generalize this in the compiler?”)

I plan to clone @mandubian’s branch tonight, and see what I can get can achieve with this in Matryoshka.

src/reflect/scala/reflect/internal/Definitions.scala
```diff
@@ -385,6 +385,9 @@ trait Definitions extends api.StandardDefinitions {
     lazy val JavaRepeatedParamClass = specialPolyClass(tpnme.JAVA_REPEATED_PARAM_CLASS_NAME, COVARIANT)(tparam => arrayType(tparam.tpe))
     lazy val RepeatedParamClass     = specialPolyClass(tpnme.REPEATED_PARAM_CLASS_NAME, COVARIANT)(tparam => seqType(tparam.tpe))
 
+    // TODO : requires ABSTRACT | TRAIT | FINAL ????
+    lazy val AnyKindClass   = enterNewClass(ScalaPackageClass, tpnme.AnyKind, AnyTpe :: Nil, ABSTRACT | TRAIT) markAllCompleted
```
retronym, line 389: The existence of the fictional class `scala.AnyKind` could be made conditional on the experimental flag to avoid source incompatibilities with programs already using that name.

mandubian: Interesting: is it done somewhere else because I don't know how to make it?

mandubian:
After struggling quite a lot against scalac implicit solver, I've found a minimal patch that enables "Kind-Polymorphic" type to be used in implicit type propagation (à la shapeless Aux) such as:

```
  def append[MA, M <: AnyKind, Self](m1: MA, m2: MA)(
    implicit kinder: Kinder.Aux[MA, M], semiGroup: SemiGroup.Aux[M, Self], appender: SemiGroup.AppendFunction[Self, MA, M]
  ): MA = semiGroup.append(m1, m2)
```

cf https://github.com/mandubian/scala/blob/topic/kind-poly-v1/test/files/pos/kind-poly2.scala

SethTisue:
`PK` leaves a sort of ghost of the original PKD reference (`Ubik`)

milessabin:
Rebase?

milessabin:
FYI: this has been [merged](https://github.com/typelevel/scala/blob/typelevel-readme/notes/2.12.1.md#minimal-kind-polymorphism-pull5538-mandubian) in Typelevel Scala 2.12.1.

TomasMikula:
@mandubian

> The idea is NOT to provide universal kind-polymorphism for all possible kinds which would be impossible and a bad idea.

> we do not try to provide a universal kind-polymorphism in Scala because it wouldn't work & would lead to crazy cases

My intuition is that only kind _inference_ would be problematic, whereas kind _checking_ should be quite simple. For starters, I would take having to explicitly specify kind arguments (i.e. no kind inference) over the implicit approach.

fommil:
@mandubian does this PR make the scalaz `Coproduct` data type redundant?

```scala
final case class Coproduct[F[_], G[_], A](run: F[A] \/ G[A])
```

mandubian:
@fommil no because PolyKinded structures aren't as straightforward as you would think because of limitations of Scalac (you have to write intermediate structures to help it which makes everything quite ugly) ... Moreover, for now, AFAIK there are issues in sharing polykinded code between compiling units... Actually, for now, I think current Polykind feature is too limited to be more than a toy and won't be more than that without drastically modifying scalac (true partially applied types, kind representation in the type system etc...)

SethTisue:
where does this stand...? should it remain in the PR queue?

milessabin:
I'm going to do some more work on kind polymorphism, but I think it's right to close this PR now.

adriaanm:
After discussion with Martin today, a crucial insight clicked for me: an any-kinded type cannot be applied to type arguments! That makes this much more palatable to me.

mandubian:
Just to be sure, what do you mean by "palatable"?
To me, after writing this poc 2 years ago & playing with it, my main own
blocking point is that it can compile & reuse kind-poly code in one single
compiling unit with known limitations linked to its ignorance of what is
the kind of the type... But in different compiling units, kind-poly
information is erased by my agressive rules so building a library on it was
impossible anyway...
This would require deeper thoughts in the compiler about kinds IMHO...

Le 12 mars 2018 9:07 PM, "Adriaan Moors" <notifications@github.com> a
écrit :

After discussion with Martin today, a crucial insight clicked for me: an
any-kinded type cannot be applied to type arguments! That makes this much
more palatable to me.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/scala/scala/pull/5538#issuecomment-372445442>, or mute
the thread

adriaanm:
I mean that I misunderstood the original proposal to mean "turning off kind checks, so that you can apply the same type parameter to any number of type arguments". Now that I understand it better, I think it's fine (== palatable == acceptable). 

I've always loved the idea of level polymorphism (first encountered it in Tim Sheard's Omega), which is kind of like collapsing all "classifiers" into one space, but indexing them with a level, then abstracting over that level. Maybe we don't even need that index, as hinted at by Miles in the other PR. That's probably a little too ambitious for Scala, though.

SethTisue:
meanwhile over in Dotty-land: https://github.com/lampepfl/dotty/pull/4108

TomasMikula:
I haven't gone through all of the test cases, but I conjecture that what can be achieved with this `<: AnyKind` kind polymorpism can also be achieved without it at a marginal increase in syntactic noise. Should this be right, I don't find the `AnyKind` proposal sufficiently justified.

The idea is simple:
 - Instead of parameterizing by `F <: AnyKind`, parameterize by plain monotype `F`.
 - When you need to pass a type constructor, say `G[_]`, for a type parameter `F`, pass `κ[G]` instead, where `type κ[_[_]]`. (We will need multiple shapes of these, one for each shape of type constructor.)
 - Adjust the rest of the code. 😎

The last point might be hiding great complexity, but it was really straightforward in the two cases I tried: semigroup and polymorphic function.

The following are test files from this PR adjusted in this way. They compile with plain Scala 2.12.4.

#### kind-poly2.scala
```scala

object Test {

  type κ1[_[_]]
  type κ2[_[_, _]]

  //////
  ////////////////////////////////////////////////
  // Extract Kind from a type
  trait Kinder[MA] { type κM }
  object Kinder extends KinderLowerImplicits {
    type Aux[MA, κM0] = Kinder[MA] { type κM = κM0 }
  
    implicit def kinder1[M0[_], A0]: Kinder.Aux[M0[A0], κ1[M0]] = new Kinder[M0[A0]] { type κM = κ1[M0] }
    implicit def kinder2[M0[_, _], A0, B0]: Kinder.Aux[M0[A0, B0], κ2[M0]] = new Kinder[M0[A0, B0]] { type κM = κ2[M0] }
  }
  trait KinderLowerImplicits {
    implicit def kinder0[A]: Kinder.Aux[A, A] = new Kinder[A] { type κM = A }    
  }

  trait SemiGroup[M] {
    // Just a mirror type of itself to ensure the owning of AppendFunction...
    type Self
    def append[MA](m1: MA, m2: MA)(implicit appender: SemiGroup.AppendFunction[Self, MA, M]) = appender(m1, m2)
  }

  object SemiGroup {
    type Aux[M, Self0] = SemiGroup[M] { type Self = Self0 }

    trait AppendFunction[P, FA, F] {
      def apply(m1: FA, m2: FA): FA
    }
  }

  implicit object SemiGroupInt extends SemiGroup[Int] {
    type Self = this.type
    implicit val appender = new SemiGroup.AppendFunction[Self, Int, Int] {
      def apply(m1: Int, m2: Int) = m1 + m2
    }
  }

  implicit object SemiGroupList extends SemiGroup[κ1[List]] {
    type Self = this.type
    implicit def appender[A] = new SemiGroup.AppendFunction[Self, List[A], κ1[List]] {
      def apply(m1: List[A], m2: List[A]) = m1 ++ m2
    }
  }

  implicit object SemiGroupMap extends SemiGroup[κ2[Map]] {
    type Self = this.type
    implicit def appender[A, B] = new SemiGroup.AppendFunction[Self, Map[A, B], κ2[Map]] {
      def apply(m1: Map[A, B], m2: Map[A, B]) = m1 ++ m2
    }
  }

  // Searching a semigroup and using it
  def semiGroup[M](implicit sg: SemiGroup[M]): SemiGroup.Aux[M, sg.Self] = sg

  semiGroup[Int].append(5, 8)
  semiGroup[κ1[List]].append(List(1), List(3))
  semiGroup[κ2[Map]].append(Map("toto" -> 1L), Map("tata" -> 3L))

  // higher level append function
  def append[MA, M, Self](m1: MA, m2: MA)(
    implicit kinder: Kinder.Aux[MA, M], semiGroup: SemiGroup.Aux[M, Self], appender: SemiGroup.AppendFunction[Self, MA, M]
  ): MA = semiGroup.append(m1, m2)

  val r1: Int = append(5, 8)
  val r2: List[Int] = append(List(1), List(3))
  val r3: Map[String, Long] = append(Map("toto" -> 1L), Map("tata" -> 3L))
}
```

#### kind-poly4.scala
```scala

object Test {
  type κ1[_[_]]

  case class Bar[A](a: A)

  /** A Typeclass indexed by a kind of morphism and able to:
    * 1/ extract a F[_*] from FA
    * 2/ apply M on it
    * 3/ return a GA which is a G[_*]
    */
  trait UnapplyApply[FA, F, G, GA, M <: Morphism[F, G]] {
    def apply(m: M)(g: FA): GA
  }

  /** a kind-polymorphic morphism */
  sealed trait Morphism[F, G] {
    def apply[FApp, GApp](v: FApp)(implicit unapplyApply: UnapplyApply[FApp, F, G, GApp, this.type]): GApp = unapplyApply(this)(v)
  }

  /** order-1 morphism AKA function */
  class Morphism1[A, B](val f: A => B) extends Morphism[A, B]

  object Morphism1 {
    /** the unapplyApply for morphism1 */
    implicit def unapplyApply1[A, B, M <: Morphism1[A, B]] = new UnapplyApply[A, A, B, B, M] {
      def apply(m: M)(a: A): B = m.f(a)
    }
  }

  /** order-2 morphism AKA FunctionK */
  trait Morphism2[F[_], G[_]] extends Morphism[κ1[F], κ1[G]] {
    def apply[A](fa: F[A]): G[A]
  }

  object Morphism2 {
    /** the unapplyApply for morphism2 */
    implicit def unapplyApply2[F[_], G[_], A, M <: Morphism2[F, G]] = new UnapplyApply[F[A], κ1[F], κ1[G], G[A], M] {
      def apply(m: M)(fa: F[A]): G[A] = m.apply(fa)
    }
  }

  // Morphism sample
  val m1 = new Morphism1( (i:Int) => i.toString )
  m1(1)

  val m2 = new Morphism2[Bar, List] {
    def apply[A](fa: Bar[A]): List[A] = List(fa.a)
  }
  m2(Bar(5))

  /** A simple Product (X1, X2) indexed by the type of morphism and respecting the categorical product representation
    * described on Wikipedia https://en.wikipedia.org/wiki/Product_(category_theory):
    * "Let C be a category with some objects X1 and X2. A product of X1 and X2 is an object X (often denoted X1 × X2)
    *  together with a pair of morphisms π1 : X → X1, π2 : X → X2 that satisfy the following universal property:
    *     -for every object Y and pair of morphisms f1 : Y → X1, f2 : Y → X2 there exists a unique morphism f : Y → X
    * "
    */
  trait Prod[M[a, b] <: Morphism[a , b]] {
    // synthetic poly-kinded representation of own type for be used in π1 & π2 morphisms
    type Self

    // the types in the product
    type X1
    type X2

    // the projection morphism
    def π1 : M[Self, X1]
    def π2 : M[Self, X2]
  }

  /** the Product[Morphism1] AKA a simple tuple */
  case class Prod1[A, B](a: A, b: B) extends Prod[Morphism1] {
    type Self = Prod1[A, B]
    type X1 = A
    type X2 = B

    val π1 = new Morphism1[Self, A](_ => a)
    val π2 = new Morphism1[Self, B](_ => b)

    val prj1: A = π1(this)
    val prj2: B = π2(this)
  }

  /** The Builder represents the unique morphism from categorical representation:
    * "for every object Y and pair of morphisms f1 : Y → X1, f2 : Y → X2 there exists a unique morphism f : Y → X"
    */
  class ProdBuilder1[Y, A, B](f1: Morphism1[Y, A], f2: Morphism1[Y, B]) extends Morphism1[Y, Prod1[A, B]](
    (y:Y) => new Prod1(f1(y), f2(y))
  )

  /** the Product[Morphism2] AKA a [A] => (F[A], G[A]) */
  case class Prod2[F[_], G[_], A](fa: F[A], ga: G[A]) extends Prod[Morphism] {
    type Self_[A] = Prod2[F, G, A]
    type Self = κ1[Self_]
    type X1 = κ1[F]
    type X2 = κ1[G]
    val π1 = new Morphism2[Self_, F] {
      def apply[A](x: Self_[A]): F[A] = x.fa
    }
    val π2 = new Morphism2[Self_, G] {
      def apply[A](x: Self_[A]): G[A] = x.ga
    }

    val prj1: F[A] = π1(this)
    val prj2: G[A] = π2(this)
  }

  /** The Builder represents the unique morphism from categorical representation:
    * "for every object Y and pair of morphisms f1 : Y → X1, f2 : Y → X2 there exists a unique morphism f : Y → X"
    */
  class ProdBuilder2[Y[_], F[_], G[_]](f1: Morphism2[Y, F], f2: Morphism2[Y, G])
    extends Morphism2[Y, ({ type l[t] = Prod2[F, G, t] })#l] {
    def apply[A](ya: Y[A]): Prod2[F, G, A] = new Prod2(f1(ya), f2(ya))
  }


  // Prod1 Sample (ok it's a bit less simple to write :D)
  // Y is here Unit for convenience but it could be anything
  val prodb1 = new ProdBuilder1(
    new Morphism1((y:Unit) => 5)
  , new Morphism1((y:Unit) => "toto")
  )
  val prod1 = prodb1()
  val i: Int = prod1.π1(prod1)
  val s: String = prod1.π2(prod1)


  // Prod2 Sample (ok it's a bit less simple to write :D)
  // ID is used for having a Y in order-2
  type Id[A] = A
  val prodb2 = new ProdBuilder2(
    new Morphism2[Id, List] {
      def apply[A](x: Id[A]): List[A] = List(x)
    }
  , new Morphism2[Id, Option] {
      def apply[A](x: Id[A]): Option[A] = Option(x)
    }
  )

  val prod2Int = prodb2(5)
  val l1: List[Int] = prod2Int.π1(prod2Int)
  val l11: List[Int] = prod2Int.prj1
  val o1: Option[Int] = prod2Int.π2(prod2Int)
  val o11: Option[Int] = prod2Int.prj2

  val prod2String = prodb2("toto")
  val l2: List[String] = prod2String.π1(prod2String)
  val l21: List[String] = prod2String.prj1
  val o2: Option[String] = prod2String.π2(prod2String)
  val o21: Option[String] = prod2String.prj2
}
```

milessabin:
@TomasMikula shapeless has had [an example](https://github.com/milessabin/shapeless/blob/master/examples/src/main/scala/shapeless/examples/kindpoly.scala) of a similar encoding since 2012. It's too cumbersome to be useful in practice.

TomasMikula:
> It's too cumbersome to be useful in practice.

On that we agree. My claim here is that it is only marginally more cumbersome than using `AnyKind`.

PR #5540, opened 2016-11-18
SI-9814 Fix synchronized in specialized overrides

retronym:
Specialization creates a subclasses of a specializd class for each
type parameter combination. These contains copies of the methods from
the superclass.

However, before this transform, the pattern of self-synchronization
in a method body had been replace by flag Flag.SYNCHRONIZED on the
method symbol. This was not being propagated to the override, and
hence no locking occured.

This commit modifies the transform to copy the SYNCHRONIZED flag.
I've also added tests for some other potentially tricky variations
of synchronization to confirm they are working.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
+++ b/src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
@@ -1049,7 +1049,8 @@ abstract class SpecializeTypes extends InfoTransform with TypingTransformers {
             }
           }
           debuglog(s"specialized overload $om for ${overriding.name.decode} in ${pp(env)}: ${om.info}")
-          if (overriding.isAbstractOverride) om.setFlag(ABSOVERRIDE)
+          om.setFlag(overriding.flags & (ABSOVERRIDE | SYNCHRONIZED))
+          om.withAnnotations(overriding.annotations.filter(_.symbol == ScalaStrictFPAttr))
           typeEnv(om) = env
           addConcreteSpecMethod(overriding)
           if (overriding.isDeferred) { // abstract override
--- a/test/files/run/t9814.scala
+++ b/test/files/run/t9814.scala
@@ -1,0 +1,28 @@
+import java.lang.reflect.Modifier
+
+import scala.annotation.strictfp
+
+class Foo extends (() => Unit) {
+  def apply(): Unit = synchronized {
+    // we're in a specialized subclass
+    assert(Thread.currentThread.getStackTrace.apply(1).getMethodName == "apply$mcV$sp")
+    assert(Thread.holdsLock(this))
+  }
+}
+
+class Bar extends (() => Unit) {
+  @strictfp def apply(): Unit = synchronized {
+    // we're in a specialized subclass
+    assert(Thread.currentThread.getStackTrace.apply(1).getMethodName == "apply$mcV$sp")
+    assert(Thread.holdsLock(this))
+  }
+}
+
+object Test {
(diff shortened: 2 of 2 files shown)


retronym:
Review by @lrytz

test/files/run/t9814b.scala
```diff
@@ -0,0 +1,54 @@
...
+    assert(Thread.holdsLock(this))
+  }
+  def method() = synchronized {
+    assert(Thread.holdsLock(this))
+  }
+}
+
+object Test {
+  // All of these were working, just double checking after the absense of synchronized in t9814.scala
+  def main(args: Array[String]): Unit = {
```
dragos: These tests have nothing to do with specialization, right? Maybe a different test file name would be better. People reading the bug report would come back more confused

dragos: There's also `test/files/run/synchronized.scala` which seems to cover at least some part of this suite.

dragos:
I have some remarks regarding the second part of the test, otherwise the fix LGTM

src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
```diff
@@ -1669,6 +1669,7 @@ abstract class SpecializeTypes extends InfoTransform with TypingTransformers {
               val tree1 = addBody(ddef, target)
               (new ChangeOwnerTraverser(target, tree1.symbol))(tree1.rhs)
               debuglog("changed owners, now: " + tree1)
+              if (target.hasFlag(Flags.SYNCHRONIZED)) ddef.symbol.setFlag(Flags.SYNCHRONIZED)
```
lrytz: I never looked at the implementation of `SpecializeTypes`, and I couldn't see the answer immediately, so let me just ask: why is the flag not copied when the `ddef.symbol` is created?

retronym: Good question. I've just poked around and found that the `Function0.apply` is used as the prototype from with `Foo.apply$mc...` is cloned. I've found a better place to do this and force pushed a new version.

test/files/run/t9814b.scala
retronym: I'm happy with the coverage in the other file so I just deleted this test.

src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
```diff
@@ -1049,7 +1049,8 @@ abstract class SpecializeTypes extends InfoTransform with TypingTransformers {
             }
           }
           debuglog(s"specialized overload $om for ${overriding.name.decode} in ${pp(env)}: ${om.info}")
-          if (overriding.isAbstractOverride) om.setFlag(ABSOVERRIDE)
+          om.setFlag(overriding.flags & (ABSOVERRIDE | SYNCHRONIZED))
+          om.withAnnotations(overriding.annotations.filter(_.symbol == ScalaStrictFPAttr))
```
lrytz, line 1053: OK, I see now.

```
class Foo extends (() => Unit) {
  def apply(): Unit = synchronized(println())
}
```

the symbol for `apply$mcV$sp` in `Foo` is cloned from the symbol of `apply` in `Function0`, not in `Foo`. I.e., a few lines above, we have `val om = specializedOverload(clazz, overridden, env)`. We cannot easily clone `overriding` instead, as this one no longer refers to type params..

lrytz:
LGTM

lrytz:
will you back-port this to 2.11.9? you added it as a fix version in jira.

PR #5542, opened 2016-11-18
SI-10066 Fix crash in erroneous code with implicits, dynamic

retronym:
The compiler support in the typechecker for `scala.Dynamic` is
very particular about the `Context` in which it is typechecked.
It looks at the `tree` in the enclosing context to find the expression
immediately enclosing the dynamic selection. See the logic in
`dyna::mkInvoke` for the details.

This commit substitutes the result of `resetAttrs` into the tree
of the typer context before continuing with typechecking.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -863,11 +863,24 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                   case _ =>
                 }
                 debuglog(s"fallback on implicits: ${tree}/$resetTree")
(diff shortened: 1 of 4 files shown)


src/reflect/scala/reflect/internal/Trees.scala
```diff
@@ -528,6 +528,7 @@ trait Trees extends api.Trees {
 
   case class Select(qualifier: Tree, name: Name)
        extends RefTree with SelectApi {
+    private val debug = Thread.currentThread().getStackTrace.mkString("\n")
```
som-snytt: FYI.

retronym: My secrets revealed!

PR #30, opened 2011-12-05
Add link to collections API overview to scaladoc of collection classes (SI-5280)

joshmarcus:
See https://issues.scala-lang.org/browse/SI-5280

Recently, on scala-debate, Martin Odersky wrote: "I think two of the things we need to do on the docs side is hide implementation details better in the ScalaDoc and put a link to the collections API overview

http://www.scala-lang.org/docu/files/collections-api/collections.html

into the scaladoc of each collection class."

I've implemented the second suggestion. For the classes that have their own pages in the collections API overview, I've added @see ("See also") links to their specific pages in the collections API overview. I used the existing @see from Array.scala as a model.

PR #5543, opened 2016-11-18
SI-10069 Fix code gen errors with array updates, Nothing

retronym:
Crashes in ASM or VerifyErrors used to occur when assigning
an expression of type Nothing to an element of a primitive array.

This commit adapts the RHS of the assignment to the element
type to correct this. `adapt` contains logic to insert an
`ATHROW` of the slot of type `Nothing$`, which makes everything
line up. The subsequent array stores become dead code and are
dropped later on in code gen, so the test case compiles to:

  public void foo0(double[]);
    Code:
       0: bipush        42
       2: istore_2
       3: aload_1
       4: iconst_0
       5: aload_0
       6: invokevirtual #30                 // Method throwExpected:()Lscala/runtime/Nothing$;
       9: athrow

Review by @lrytz. Does `adapt` ever do more than we want here?

JIRA: https://issues.scala-lang.org/browse/SI-10069

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -168,7 +168,7 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
       } else if (scalaPrimitives.isArraySet(code)) {
         val List(a1, a2) = args
         genLoad(a1, INT)
-        genLoad(a2)
+        genLoad(a2, elementType)
         generatedType = UNIT
         bc.astore(elementType)
       } else {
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -169,6 +169,7 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
         val List(a1, a2) = args
         genLoad(a1, INT)
         genLoad(a2)
+        adapt(tpeTK(a2), elementType)
```
lrytz: You could just do `genLaod(a2, elementType)` on the previous line.

lrytz:
LGTM after that :)

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
retronym: Much simpler, thanks!

retronym:
@lrytz I've found another instance of the bug in `{Int, ...}.unbox` and fixed that too.

lrytz:
LGTM! Feel free to merge this for 2.12.1.

PR #5544, opened 2016-11-18
SI-8779 Enable inlining of code within a REPL session

retronym:
The REPL has a long running instance of Global which outputs
classfiles by default to a VirtualDirectory. The inliner did not find
any of these class files when compiling calls to methods defined in
previous runs (ie, previous lines of input.)

This commit:

  - Adds a hook to augment the classpath that the optimizer searches,
    and uses this in the REPL to add the output directory
  - Fixes the implementation of `findClassFile` in VirtualDirectory,
    which doesn't seem to have been used in anger before. I've factored out
    some common code into a new method on `AbstractFile`.
  - Fixes a similar problem getSubDir reported by Li Haoyi
  - Adds missing unit test coverage.

This also fixes a bug in REPL autocompletion for types defined
in packages >= 2 level deep (with the `:paste -raw` command).
I've added a test for this case.

Review by @lrytz

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
@@ -6,6 +6,9 @@
 package scala.tools.nsc
 package interpreter
 
+import scala.tools.nsc.backend.JavaPlatform
+import scala.tools.nsc.classpath.{AggregateClassPath, ClassPathFactory}
+import scala.tools.nsc.util.ClassPath
 import typechecker.Analyzer
 
 /** A layer on top of Global so I can guarantee some extra
@@ -29,6 +32,16 @@ trait ReplGlobal extends Global {
       macroLogVerbose("macro classloader: initializing from a REPL classloader: %s".format(global.classPath.asURLs))
       val virtualDirectory = globalSettings.outputDirs.getSingleOutput.get
       new util.AbstractFileClassLoader(virtualDirectory, loader) {}
+    }
+  }
+
+  override def optimizerClassPath(base: ClassPath): ClassPath = {
+    settings.outputDirs.getSingleOutput match {
+      case None => base
+      case Some(out) =>
+        // Make bytecode of previous lines available to the inliner
+        val replOutClasspath = ClassPathFactory.newClassPath(settings.outputDirs.getSingleOutput.get, settings)
+        AggregateClassPath.createAggregate(platform.classPath, replOutClasspath)
     }
   }
 }
--- a/test/files/run/repl-inline.scala
+++ b/test/files/run/repl-inline.scala
@@ -1,0 +1,21 @@
+import scala.tools.nsc._
+
+object Test {
+  val testCode = """
+def callerOfCaller = Thread.currentThread.getStackTrace.drop(2).head.getMethodName
+def g = callerOfCaller
+def h = g
+assert(h == "g", h)
+@inline def g = callerOfCaller
+def h = g
+assert(h == "h", h)
+  """
+
+  def main(args: Array[String]) {
+    val settings = new Settings()
+    settings.processArgumentString("-opt:l:classpath")
+    settings.usejavacp.value = true
+    val repl = new interpreter.IMain(settings)
+    testCode.linesIterator.foreach(repl.interpret(_))
+  }
+}
--- a/test/files/run/repl-inline.scala
+++ b/test/files/run/repl-inline.scala
@@ -1,7 +1,8 @@
 import scala.tools.nsc._
 
 object Test {
-  val testCode = """
+  val testCode =
+    """
 def callerOfCaller = Thread.currentThread.getStackTrace.drop(2).head.getMethodName
 def g = callerOfCaller
 def h = g
@@ -12,10 +13,15 @@ assert(h == "h", h)
   """
 
   def main(args: Array[String]) {
-    val settings = new Settings()
-    settings.processArgumentString("-opt:l:classpath")
-    settings.usejavacp.value = true
-    val repl = new interpreter.IMain(settings)
-    testCode.linesIterator.foreach(repl.interpret(_))
+    def test(f: Settings => Unit): Unit = {
+      val settings = new Settings()
+      settings.processArgumentString("-opt:l:classpath")
+      f(settings)
+      settings.usejavacp.value = true
+      val repl = new interpreter.IMain(settings)
+      testCode.linesIterator.foreach(repl.interpret(_))
+    }
+    test(_ => ())
+    test(_.Yreplclassbased.value = true)
   }
 }
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -95,6 +95,9 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
 
   type ThisPlatform = JavaPlatform { val global: Global.this.type }
   lazy val platform: ThisPlatform  = new GlobalPlatform
+  /* A hook for the REPL to add a classpath entry containing products of previous runs to inliner's bytecode repository*/
+  // Fixes SI-8779
+  def optimizerClassPath(base: ClassPath): ClassPath = base
 
   def classPath: ClassPath = platform.classPath
 
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
@@ -37,7 +37,7 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
   val coreBTypes = new CoreBTypesProxy[this.type](this)
   import coreBTypes._
 
-  val byteCodeRepository: ByteCodeRepository[this.type] = new ByteCodeRepository(global.classPath, this)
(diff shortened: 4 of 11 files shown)


som-snytt:
The sun rises early down under.

Missed my shot at this; and Li H. just complained about findClassFile; the lesson is, you can't wait for the weekend.

I once had a repl that specialized the platform classpath, with the old classpath impl; I guess putting the hook on global is much simpler.

src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
```diff
@@ -6,12 +6,16 @@
...
+import scala.tools.nsc.backend.JavaPlatform
+import scala.tools.nsc.classpath.{AggregateClassPath, ClassPathFactory}
+import scala.tools.nsc.util.ClassPath
 import typechecker.Analyzer
 
 /** A layer on top of Global so I can guarantee some extra
  *  functionality for the repl.  It doesn't do much yet.
  */
 trait ReplGlobal extends Global {
+  self =>
```
som-snytt: To thine own `self =>` be true.

lihaoyi:
FWIW, I'm pretty sure `getSubDir` on the `VirtualDirectoryClassPath` is borked too. It suffers from the same problem: doing a shallow lookup of the immediate children, rather than recursively resolving each path segment to find the file it wants.

This is what I had to do to get 2.12's `VirtualDirectoryClassPath` working with Ammonite:

``` scala
    val dynamicCP = new VirtualDirectoryClassPath(dynamicClasspath){

      override def getSubDir(packageDirName: String): Option[AbstractFile] = {
        val pathParts = packageDirName.split('/')
        var file: AbstractFile = dir
        for (dirPart <- pathParts) {
          file = file.lookupName(dirPart, directory = true)
          if (file == null) return None
        }
        Some(file)

      }
      override def findClassFile(className: String): Option[AbstractFile] = {
        val relativePath = FileUtils.dirPath(className)
        val pathParts = relativePath.split('/')
        var file: AbstractFile = dir
        for (dirPart <- pathParts.init) {
          file = file.lookupName(dirPart, directory = true)
          if (file == null) return None
        }

        file.lookupName(pathParts.last + ".class", directory = false) match {
          case null => None
          case file => Some(file)
        }
      }

    }
```

There may be other issues with it, but all my tests pass, so that's good enough for me for now.

(`findClassFile` doesn't seem to be called at all in the normal flow of compilation, but i included the fix anyway because the original code was clearly incorrect and maybe some future version will start using it)

retronym:
@lihaoyi Thanks for the heads up. I've fixed that method as well. Autocompletion in the REPL was actually broken because of this for classes defined with `:paste` in nested packages.

som-snytt:
Edit: this is an existing thing with existing things.

With an existing `Thing` on the class path:
```
$ ./build/pack/bin/scala -cp /tmp
Welcome to Scala 2.12.1-20161121-143919-1d2d383 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_111).

scala> stuff.Thing
   object Thing

scala> stuff.Thing.f
res0: Int = 42

scala> :pa

package stuff { object Thing { def g = "hi" } }

scala> stuff.Thing.g    // autocompletes!
java.lang.NoSuchMethodError: stuff.Thing$.g()Ljava/lang/String;
  ... 31 elided


```

src/compiler/scala/tools/nsc/backend/JavaPlatform.scala
```diff
@@ -23,6 +23,7 @@ trait JavaPlatform extends Platform {
     if (currentClassPath.isEmpty) currentClassPath = Some(new PathResolver(settings).result)
     currentClassPath.get
   }
+  private[nsc] def previousRunClassPath: Option[ClassPath]= None
```
som-snytt: ?

src/compiler/scala/tools/nsc/backend/jvm/opt/ByteCodeRepository.scala
```diff
@@ -290,7 +290,9 @@ class ByteCodeRepository[BT <: BTypes](val classPath: ClassPath, val btypes: BT)
       classNode
     } match {
       case Some(node) => Right(node)
-      case None       => Left(ClassNotFound(internalName, javaDefinedClasses(internalName)))
+      case None       =>
+
+        Left(ClassNotFound(internalName, javaDefinedClasses(internalName)))
```
som-snytt: `Left(some extra space)`.

som-snytt:
I hadn't noticed the nested package quirk.
```
scala> :pa

package p { package q { object x { def f = 42 } } }

scala> p.q.x
   object x

scala> p.q.x.f   // completed
res1: Int = 42

```

src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
```diff
@@ -6,6 +6,9 @@
 package scala.tools.nsc
 package interpreter
 
+import scala.tools.nsc.backend.JavaPlatform
```
som-snytt, line 9: This is the one you obviate.

src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
```diff
@@ -31,4 +34,14 @@ trait ReplGlobal extends Global {
...
     }
   }
+
+  override def optimizerClassPath(base: ClassPath): ClassPath = {
+    settings.outputDirs.getSingleOutput match {
+      case None => base
+      case Some(out) =>
+        // Make bytecode of previous lines available to the inliner
+        val replOutClasspath = ClassPathFactory.newClassPath(settings.outputDirs.getSingleOutput.get, settings)
+        AggregateClassPath(platform.classPath :: replOutClasspath :: Nil)
```
som-snytt: This order made me think twice. The REPL class loader delegates to parent; it would be a feature to prefer REPL output to whatever is on the class path. But it's a little late in the day (local time) to start pondering that.

som-snytt:
LGTM

src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
retronym: A more conservative feature might be to warn that the generated classfiles from `:paste package ...` clash with existing classpath entries and won't be picked up.

src/compiler/scala/tools/nsc/backend/JavaPlatform.scala
retronym: removed.

src/compiler/scala/tools/nsc/backend/jvm/opt/ByteCodeRepository.scala
retronym: :)

`Right(back to where we started)`

retronym:
I'll let @lrytz take a look at this from the inliner perspective before merging.

som-snytt:
This isn't surprising, but I noticed that module loads are not optimized away; I think it was considered important to keep side-effects of module init for case class apply rewriting.

And I notice that `-Yrepl-class-based` class wrappers are not final, as it will warn for inlining a non-final method.

retronym:
> And I notice that -Yrepl-class-based class wrappers are not final, as it will warn for inlining a non-final method.

Just pushed a commit to make this work, by marking them `sealed`.

retronym:
> This isn't surprising, but I noticed that module loads are not optimized away; I think it was considered important to keep side-effects of module init for case class apply rewriting.

Yeah, in the REPL context we need to take another swing at getting rid of those pesky wrappers altogether. As it stands, the REPL gets a little slower every line by needing ever deeper wrappers.

som-snytt:
@retronym My PR for your unwrapped REPL is dormant but I'll revive it. After a PR is several months old, it's real work just to take it up again.

lrytz:
needs a check file update for `test/files/run/t7747-repl.check` - i tried pushing to your branch, but permission denied

src/repl/scala/tools/nsc/interpreter/ReplGlobal.scala
lrytz: Could use `AggregateClassPath.createAggregate` on this line to prevent nesting `AggregateClassPath`s

test/files/run/repl-inline.scala
```diff
@@ -0,0 +1,27 @@
...
+def h = g
+assert(h == "h", h)
+  """
+
+  def main(args: Array[String]) {
+    def test(f: Settings => Unit): Unit = {
+      val settings = new Settings()
+      settings.processArgumentString("-opt:l:classpath")
+      f(settings)
+      settings.classpath.value = System.getProperty("java.class.path")
```
lrytz: could we instead pass `"-usejavacp"` above?

test/junit/scala/reflect/internal/util/AbstractFileClassLoaderTest.scala
```diff
@@ -5,6 +5,9 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
 
+import scala.tools.nsc.classpath.VirtualDirectoryClassPath
+import scala.tools.nsc.util.ClassPath
+
```
lrytz: no-op change

lrytz:
looks all good aside from these details.

PR #5545, opened 2016-11-18
Minor Vector cleanup

l0rinc:
My original intention didn't work out, but since I created these simple style unification commits, I though I'll share them. :)

### DIFF
--- a/src/library/scala/collection/immutable/Vector.scala
+++ b/src/library/scala/collection/immutable/Vector.scala
@@ -1020,29 +1020,29 @@ private[immutable] trait VectorPointer[T] {
     private[immutable] final def gotoPosWritable0(newIndex: Int, xor: Int): Unit = (depth - 1) match {
       case 5 =>
         display5 = copyOf(display5)
-        display4 = nullSlotAndCopy(display5, (newIndex >> 25) & 31).asInstanceOf[Array[AnyRef]]
-        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31).asInstanceOf[Array[AnyRef]]
-        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display4 = nullSlotAndCopy(display5, (newIndex >> 25) & 31)
+        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31)
+        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31)
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       case 4 =>
         display4 = copyOf(display4)
-        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31).asInstanceOf[Array[AnyRef]]
-        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31)
+        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31)
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       case 3 =>
         display3 = copyOf(display3)
-        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31)
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       case 2 =>
         display2 = copyOf(display2)
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       case 1 =>
         display1 = copyOf(display1)
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       case 0 =>
         display0 = copyOf(display0)
     }
@@ -1064,8 +1064,8 @@ private[immutable] trait VectorPointer[T] {
         display2 = copyOf(display2)
         display1((oldIndex >>  5) & 31) = display0
         display2((oldIndex >> 10) & 31) = display1
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       } else
       if (xor < (1 << 20)) { // level = 3
         display1 = copyOf(display1)
@@ -1074,9 +1074,9 @@ private[immutable] trait VectorPointer[T] {
         display1((oldIndex >>  5) & 31) = display0
         display2((oldIndex >> 10) & 31) = display1
         display3((oldIndex >> 15) & 31) = display2
-        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31)
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       } else
       if (xor < (1 << 25)) { // level = 4
         display1 = copyOf(display1)
@@ -1087,10 +1087,10 @@ private[immutable] trait VectorPointer[T] {
         display2((oldIndex >> 10) & 31) = display1
         display3((oldIndex >> 15) & 31) = display2
         display4((oldIndex >> 20) & 31) = display3
-        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31).asInstanceOf[Array[AnyRef]]
-        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31)
+        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31)
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       } else
       if (xor < (1 << 30)) { // level = 5
         display1 = copyOf(display1)
@@ -1103,11 +1103,11 @@ private[immutable] trait VectorPointer[T] {
         display3((oldIndex >> 15) & 31) = display2
         display4((oldIndex >> 20) & 31) = display3
         display5((oldIndex >> 25) & 31) = display4
-        display4 = nullSlotAndCopy(display5, (newIndex >> 25) & 31).asInstanceOf[Array[AnyRef]]
-        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31).asInstanceOf[Array[AnyRef]]
-        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
-        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31).asInstanceOf[Array[AnyRef]]
-        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31).asInstanceOf[Array[AnyRef]]
+        display4 = nullSlotAndCopy(display5, (newIndex >> 25) & 31)
+        display3 = nullSlotAndCopy(display4, (newIndex >> 20) & 31)
+        display2 = nullSlotAndCopy(display3, (newIndex >> 15) & 31)
+        display1 = nullSlotAndCopy(display2, (newIndex >> 10) & 31)
+        display0 = nullSlotAndCopy(display1, (newIndex >>  5) & 31)
       } else { // level = 6
         throw new IllegalArgumentException()
       }
--- a/src/library/scala/collection/immutable/Vector.scala
+++ b/src/library/scala/collection/immutable/Vector.scala
@@ -23,7 +23,7 @@ object Vector extends IndexedSeqFactory[Vector] {
     ReusableCBF.asInstanceOf[GenericCanBuildFrom[A]]
   private[immutable] val NIL = new Vector[Nothing](0, 0, 0)
   override def empty[A]: Vector[A] = NIL
-  
+
   // Constants governing concat strategy for performance
   private final val Log2ConcatFaster = 5
   private final val TinyAppendFaster = 2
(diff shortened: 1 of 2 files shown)


src/library/scala/collection/immutable/Vector.scala
```diff
@@ -88,7 +83,7 @@ override def companion: GenericCompanion[Vector] = Vector
 
   override def lengthCompare(len: Int): Int = length - len
 
-  private[collection] final def initIterator[B >: A](s: VectorIterator[B]) {
+  private[collection] def initIterator[B >: A](s: VectorIterator[B]) {
```
Ichoran: If we can get away with finality, we should leave it in.  In general, immutable data structures do not lend themselves well to subclassing.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -100,10 +95,7 @@ override def companion: GenericCompanion[Vector] = Vector
     s
   }
 
-
-  // can still be improved
-  override /*SeqLike*/
-  def reverseIterator: Iterator[A] = new AbstractIterator[A] {
+  override /*SeqLike*/ def reverseIterator: Iterator[A] = new AbstractIterator[A] {
```
Ichoran: This is harder to read.  It's better the original way.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -286,61 +271,46 @@ override def companion: GenericCompanion[Vector] = Vector
         s
       } else {
 
-        val freeSpace = ((1<<5*(depth)) - endIndex) // free space at the right given the current tree-structure depth
-        val shift = freeSpace & ~((1<<5*(depth-1))-1) // number of elements by which we'll shift right (only move at top level)
-        val shiftBlocks = freeSpace >>> 5*(depth-1) // number of top-level blocks
+        val freeSpace = ((1 << (5 * depth)) - endIndex) // free space at the right given the current tree-structure depth
+        val shift = (freeSpace & ~((1 << (5 * (depth - 1))) - 1)) // number of elements by which we'll shift right (only move at top level)
+        val shiftBlocks = (freeSpace >>> (5 * (depth - 1))) // number of top-level blocks
```
Ichoran: Spaces are good.  Gratuitious parens around the entire expression are not good--just visual clutter that doesn't aid anything.  Internal parens that make one not rely upon knowing operator precedence are good.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -368,56 +336,41 @@ override def companion: GenericCompanion[Vector] = Vector
...
         s.dirty = dirty
         s.gotoPosWritable(focus, blockIndex, focus ^ blockIndex)
         s.display0(lo) = value.asInstanceOf[AnyRef]
         s
       } else {
-        val shift = startIndex & ~((1<<5*(depth-1))-1)
-        val shiftBlocks = startIndex >>> 5*(depth-1)
-
-        //println("----- appendBack " + value + " at " + endIndex + " reached block end")
+        val shift = startIndex & ~((1 << 5 * (depth - 1))-1)
```
Ichoran: Missed spaces around the last `-`

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -368,56 +336,41 @@ override def companion: GenericCompanion[Vector] = Vector
...
+        val shiftBlocks = startIndex >>> 5 * (depth - 1)
```
Ichoran: Should surround multiplication with parens for clarity

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -609,20 +555,7 @@ override def companion: GenericCompanion[Vector] = Vector
...
-    val s = new Vector(cutIndex-shift, endIndex-shift, blockIndex-shift)
-    s.initFrom(this)
-    if (s.depth > 1)
-      s.gotoPos(blockIndex, focus ^ blockIndex)
-    s.depth = d
-    s.stabilize(blockIndex-shift)
-    s.cleanLeftEdge(cutIndex-shift)
-    s
-*/
+    val shift = (cutIndex & ~((1 << (5 * d)) - 1))
```
Ichoran: Don't need outer parens

src/library/scala/collection/immutable/Vector.scala
l0rinc: the whole class is [final](https://github.com/scala/scala/pull/5545/files/228da21af702a51e8dc59956fcf61a0de3957726#diff-59f3462485b74027de4fd5e9febcc81bR64), does it make sense to keep final methods also?

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -639,13 +572,8 @@ override def companion: GenericCompanion[Vector] = Vector
     val blockIndex = (cutIndex - 1) & ~31
     val xor = startIndex ^ (cutIndex - 1)
     val d = requiredDepth(xor)
-    val shift = (startIndex & ~((1 << (5*d))-1))
+    val shift = (startIndex & ~((1 << (5 * d)) - 1))
```
Ichoran: Don't need outer parens

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -832,7 +757,7 @@ private[immutable] trait VectorPointer[T] {
       if (xor < (1 << 5)) { // level = 0 (could maybe removed)
       } else
       if (xor < (1 << 10)) { // level = 1
-        display0 = display1((index >> 5) & 31).asInstanceOf[Array[AnyRef]]
+        display0 = display1((index >>  5) & 31).asInstanceOf[Array[AnyRef]]
```
Ichoran: Added an extra space by mistake?

src/library/scala/collection/immutable/Vector.scala
l0rinc: in other places it was formatted like this. should I change the rest also, or just revert this?

src/library/scala/collection/immutable/Vector.scala
l0rinc: understood, will remove external ones in other places also :)

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -1123,16 +1041,13 @@ private[immutable] trait VectorPointer[T] {
...
 
-
-
     // USED IN APPEND
     // create a new block at the bottom level (and possibly nodes on its path) and prepares for writing
 
     // requires structure is clean and at pos oldIndex,
     // ensures structure is dirty and at pos newIndex and writable at level 0
     private[immutable] final def gotoFreshPosWritable0(oldIndex: Int, newIndex: Int, xor: Int): Unit = { // goto block start pos
       if (xor < (1 << 5)) { // level = 0
```
Ichoran: Should leave a comment that we're already at the block start.

Ichoran:
This looks like a welcome improvement overall; just a few aspects that could be even tidier if you're tidying anyway.

src/library/scala/collection/immutable/Vector.scala
Ichoran: Ah.  No.  Sorry, I missed that.  I had somehow assumed, without any evidence, that it was in VectorPointer.

src/library/scala/collection/immutable/Vector.scala
Ichoran: If you don't mind changing the others, may as well change them.  The `override /* SeqLike */` thing is arguably the least important part of the signature, so having it push other stuff out of view on the right, potentially, is a bad idea.  Two lines solve the problem nicely.

l0rinc:
Thanks @Ichoran!
I've applied your suggestions and rebased the commit, hoping the build will pass with this base.
I've also changed all signed left shifts to unsigned ones, since all indices are inherently positive.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -888,35 +888,35 @@ private[immutable] trait VectorPointer[T] {
         display3 = copyOf(display3)
         display2 = copyOf(display2)
         display1 = copyOf(display1)
-        display5((index >> 25) & 31) = display4
-        display4((index >> 20) & 31) = display3
-        display3((index >> 15) & 31) = display2
-        display2((index >> 10) & 31) = display1
-        display1((index >>  5) & 31) = display0
+        display5((index >>> 25) & 31) = display4
```
l0rinc, line 898: indices are inherently positive or zero

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -414,22 +420,30 @@ override def companion: GenericCompanion[Vector] = Vector
...
+      array(i) = null
+      i += 1
+    }
   }
 
-  private def copyLeft(array: Array[AnyRef], right: Int): Array[AnyRef] = {
-    val a2 = new Array[AnyRef](array.length)
-    java.lang.System.arraycopy(array, 0, a2, 0, right)
-    a2
+  private[immutable] def copyLeft(array: Array[AnyRef], right: Int): Array[AnyRef] = {
```
l0rinc: to unify with `copyOf`

l0rinc:
@szeiger, not sure how to fix this unrelated [MiMa](https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/3561/console) failure, what should I rebase againts?

src/library/scala/collection/immutable/Vector.scala
Ichoran: If there is error-checking to make sure they're positive or zero, then there is no reason to bother changing. If there is not error-checking, an exception could be replaced with a silently wrong result.  I think they should stay signed!

l0rinc: The error checking is in [checkRangeConvert](https://github.com/paplorinc/scala/blob/4279bbc9f467ef6d8a0ddf4190f2dd947f13d580/src/library/scala/collection/immutable/Vector.scala#L120) and `>>>` was already used in many other places, it's the reason why I unified it: https://github.com/paplorinc/scala/blob/4279bbc9f467ef6d8a0ddf4190f2dd947f13d580/src/library/scala/collection/immutable/Vector.scala#L482

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -113,22 +106,18 @@ override def companion: GenericCompanion[Vector] = Vector
...
   def apply(index: Int): A = {
     val idx = checkRangeConvert(index)
-    //println("get elem: "+index + "/"+idx + "(focus:" +focus+" xor:"+(idx^focus)+" depth:"+depth+")")
     getElem(idx, idx ^ focus)
   }
 
   private def checkRangeConvert(index: Int) = {
     val idx = index + startIndex
-    if (0 <= index && idx < endIndex)
+    if (index >= 0 && idx < endIndex)
```
l0rinc, line 120: @Ichoran, this makes sure all indices are strictly non-negative.
(switched the comparison order to make it more intuitive - it's also easier to see that different variables are compared for boundaries)

szeiger:
There's nothing to rebase against yet, all 2.12.x builds are currently broken. I've created https://github.com/scala/scala/pull/5549 to fix this.

l0rinc:
@szeiger, I still have a MiMa failure (not sure where to view the failure's reason).
Is it because of the deleted [`final` method modifier in the `final` class](https://github.com/scala/scala/pull/5545#pullrequestreview-9292947)?

szeiger:
@paplorinc Go to the details of `validate-test`, select `Console Output` and `Full Log`, search for "binary compatibility":

```
/home/jenkins/workspace/scala-2.12.x-validate-test/build/pack/lib/scala-library.jar
Found 4 binary incompatibilities (1 were filtered out)
 * method debug()Unit in class scala.collection.immutable.VectorBuilder does
   not have a correspondent in current version
 * method debug()Unit in class scala.collection.immutable.VectorIterator
   does not have a correspondent in current version
 * method debug()Unit in class scala.collection.immutable.Vector does not
   have a correspondent in current version
 * method debug()Unit in interface scala.collection.immutable.VectorPointer
   does not have a correspondent in current version
```

Unlike most projects we require both backwards and forwards binary compatibility for Scala itself, so you can never add or remove a method.

Some cases are really not binary compatible but we're OK with it because it only affects internal code that could not be accessed from outside the standard library without subverting Scala's access restrictions. Package-private elements generally fall into this category. Once you have convinced yourself that the binary-incompatible changes are harmless you can whitelist them. See the log for the ready-to-copy-and-paste whitelisting configuration. Note that it's always shown under backwards compatibility even when it really affects forwards compatibility (and therefore has to go into `bincompat-forward-whitelist.conf`).

l0rinc:
Thank you @szeiger, had to remove some things, but now it's finally green :/.
@Ichoran, please review :)

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -351,73 +327,56 @@ override def companion: GenericCompanion[Vector] = Vector
...
         s.dirty = dirty
         s.gotoPosWritable(focus, blockIndex, focus ^ blockIndex)
         s.display0(lo) = value.asInstanceOf[AnyRef]
         s
       } else {
-        val shift = startIndex & ~((1<<5*(depth-1))-1)
-        val shiftBlocks = startIndex >>> 5*(depth-1)
-
-        //println("----- appendBack " + value + " at " + endIndex + " reached block end")
+        val shift       = startIndex & ~((1 << (5 * (depth - 1))) - 1)
```
Ichoran: This is not sufficiently similar to the next computation for alignment to be a good idea.  It suggests that the two are closely related, when in fact they are in detail rather different computations.  So I would just go `shift = startIndex` without the extra spaces.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -639,29 +575,22 @@ override def companion: GenericCompanion[Vector] = Vector
...
 }
 
-
 class VectorIterator[+A](_startIndex: Int, endIndex: Int)
-extends AbstractIterator[A]
-   with Iterator[A]
-   with VectorPointer[A @uncheckedVariance] {
+  extends AbstractIterator[A]
+    with Iterator[A]
+    with VectorPointer[A @uncheckedVariance] {
```
Ichoran: This style makes it hard to tell at a glance where the class signature ends and where the body begins.  The style is all over the place in practice, but I recommend left-justifying everything; vertical whitespace should always be used to separate classes so there is no need to indent them to indicate that they all belong to `VectorIterator`.  Failing that, the previous version was preferable because at least all the extended classes/traits are lined up.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -799,98 +725,81 @@ private[immutable] trait VectorPointer[T] {
...
+        display1((index >>>  5) & 31).asInstanceOf[Array[AnyRef]](index & 31).asInstanceOf[T]
+      } else if (xor < (1 << 15)) { // level = 2
+        display2((index >>> 10) & 31).asInstanceOf[Array[AnyRef]]((index >>>  5) & 31).asInstanceOf[Array[AnyRef]](index & 31).asInstanceOf[T]
+      } else if (xor < (1 << 20)) { // level = 3
+        display3((index >>> 15) & 31).asInstanceOf[Array[AnyRef]]((index >>> 10) & 31).asInstanceOf[Array[AnyRef]]((index >>>  5) & 31).asInstanceOf[Array[AnyRef]](index & 31).asInstanceOf[T]
+      } else if (xor < (1 << 25)) { // level = 4
+        display4((index >>> 20) & 31).asInstanceOf[Array[AnyRef]]((index >>> 15) & 31).asInstanceOf[Array[AnyRef]]((index >>> 10) & 31).asInstanceOf[Array[AnyRef]]((index >>>  5) & 31).asInstanceOf[Array[AnyRef]](index & 31).asInstanceOf[T]
+      } else if (xor < (1 << 30)) { // level = 5
+        display5((index >>> 25) & 31).asInstanceOf[Array[AnyRef]]((index >>> 20) & 31).asInstanceOf[Array[AnyRef]]((index >>> 15) & 31).asInstanceOf[Array[AnyRef]]((index >>> 10) & 31).asInstanceOf[Array[AnyRef]]((index >>> 5) & 31).asInstanceOf[Array[AnyRef]](index & 31).asInstanceOf[T]
```
Ichoran: This is an awful mess even with the improvements unless you have a very wide screen.  If you're up for it maybe you can rewrite it in the style
```
display5((index >>> 25) & 31).asInstanceOf[Array[AnyRef]].
   apply((index >>> 20) & 31).asInstanceOf[Array[AnyRef]].
   apply((index >>> 15) & 31) . . .
```
The alternative--lining up the matching part of the displays--pushes the earlier ones so far right that they disappear, so that's probably not a good idea.

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -1118,117 +1023,83 @@ private[immutable] trait VectorPointer[T] {
...
-      //println("DISPLAY 3: " + (if (display3 ne null) display3.map(x=> if (x eq null) "." else x.asInstanceOf[Array[AnyRef]].deepMkString("[","","]")).mkString(" ") else "null"))
-      //println("DISPLAY 2: " + (if (display2 ne null) display2.map(x=> if (x eq null) "." else x.asInstanceOf[Array[AnyRef]].deepMkString("[","","]")).mkString(" ") else "null"))
-      //println("DISPLAY 1: " + (if (display1 ne null) display1.map(x=> if (x eq null) "." else x.asInstanceOf[Array[AnyRef]].deepMkString("[","","]")).mkString(" ") else "null"))
-      //println("DISPLAY 0: " + (if (display0 ne null) display0.map(x=> if (x eq null) "." else x.toString).mkString(" ") else "null"))
-    }
-
-
-}
-
+}
```
Ichoran: Add a newline at the end to keep things happy that like newlines at the end!

l0rinc: :)) sir, yes sir!

src/library/scala/collection/immutable/Vector.scala
l0rinc: how about
```scala
var array = display5 
array = array((index >>> 25) & 31).asInstanceOf[Array[AnyRef]]
array = array((index >>> 20) & 31).asInstanceOf[Array[AnyRef]]
array = array((index >>> 15) & 31).asInstanceOf[Array[AnyRef]]
array = array((index >>> 10) & 31).asInstanceOf[Array[AnyRef]]
array = array((index >>>  5) & 31).asInstanceOf[Array[AnyRef]]
array(index & 31).asInstanceOf[T]
```
or would that be slower than the direct reference?

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -799,98 +727,102 @@ private[immutable] trait VectorPointer[T] {
...
+        array(index & 31).asInstanceOf[T]
+      } else if (xor < (1 << 30)) { // level = 5
+        var array = display5
+        array = array((index >>> 25) & 31).asInstanceOf[Array[AnyRef]]
+        array = array((index >>> 20) & 31).asInstanceOf[Array[AnyRef]]
+        array = array((index >>> 15) & 31).asInstanceOf[Array[AnyRef]]
+        array = array((index >>> 10) & 31).asInstanceOf[Array[AnyRef]]
+        array = array((index >>>  5) & 31).asInstanceOf[Array[AnyRef]]
+        array(index & 31).asInstanceOf[T]
+      } else {                      // level = 6
```
Ichoran, line 759: I have witnessed cases where allocating a temporary var produces worse performance than chaining methods, and at this point in time I think chaining is slightly more idiomatic than repeated assignment to vars.  So I'd just chain them as I indicated before.  If you don't like the `apply`, you can do it this way:
```
array((index >>> 25) & 31).
  asInstanceOf[Array[AnyRef]]((index >>> 20) & 31).
  asInstanceOf[Array[AnyRef]]((index >>> 15) & 31).
  ...
```
You miss the alignment on the first shift, but I guess you could add whitespace until it was aligned.

l0rinc: Done, resolved it via parenthesizing everything to make it multilinable, making alignment trivial also :)

Ichoran (approved):
LGTM

l0rinc:
@retronym, could you please review this also, it's related to https://issues.scala-lang.org/browse/SI-10049 as well?

l0rinc:
@Ichoran, @retronym, rebased to fix conflict in whitelist.

szeiger:
/nothingtoseehere

The CLA checker had some hickups recently. CLA confirmed.

SethTisue:
thanks @paplorinc! you're killing it lately

l0rinc:
Thank you @Ichoran, @szeiger and @SethTisue! :)

PR #5546, opened 2016-11-19
SI-9636 More precise error pos on apply inference

som-snytt:
If a method type arg is inferred Any, warn about the
function and not the innocent arg.

The arg-specific code still backs up the `isApplicable` path.

As suggested by retronym on jira.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -495,21 +495,22 @@ trait Infer extends Checkable {
     }
 
     /** Return inferred type arguments, given type parameters, formal parameters,
-    *  argument types, result type and expected result type.
-    *  If this is not possible, throw a `NoInstance` exception.
-    *  Undetermined type arguments are represented by `definitions.NothingTpe`.
-    *  No check that inferred parameters conform to their bounds is made here.
-    *
-    *  @param   tparams         the type parameters of the method
-    *  @param   formals         the value parameter types of the method
-    *  @param   restpe          the result type of the method
-    *  @param   argtpes         the argument types of the application
-    *  @param   pt              the expected return type of the application
-    *  @return  @see adjustTypeArgs
-    *
-    *  @throws                  NoInstance
-    */
-    def methTypeArgs(tparams: List[Symbol], formals: List[Type], restpe: Type,
+     *  argument types, result type and expected result type.
+     *  If this is not possible, throw a `NoInstance` exception.
+     *  Undetermined type arguments are represented by `definitions.NothingTpe`.
+     *  No check that inferred parameters conform to their bounds is made here.
+     *
+     *  @param   fn              the function for reporting, may be empty
+     *  @param   tparams         the type parameters of the method
+     *  @param   formals         the value parameter types of the method
+     *  @param   restpe          the result type of the method
+     *  @param   argtpes         the argument types of the application
+     *  @param   pt              the expected return type of the application
+     *  @return  @see adjustTypeArgs
+     *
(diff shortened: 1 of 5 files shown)


som-snytt:
So much water under the bridge, and now Jenkins challenges me on my CLA status. Is this what a Trump administration will look like?

SethTisue:
I poked around trying to figure out why the CLA checker isn't working here and didn't come up with anything. We can just `/nothingtoseehere` it at merge time and hope the problem doesn't recur too often :-\

lrytz:
This needs a rebase by now..

lrytz:
Wait, the second commit (which causes the conflict) doesn't belong to this PR, right?

lrytz:
LGTM (the first commit)

som-snytt:
The second commit was to placate mima after she woke from her extended sleep in a bad mood.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -559,21 +560,23 @@ trait Infer extends Checkable {
...
-          case Apply(fn, args) if idx < args.size => args(idx).pos
-          case _                                  => context.tree.pos
-        }
-        case _ => context.tree.pos
-      }
       if (settings.warnInferAny && context.reportErrors && canWarnAboutAny) {
-        foreachWithIndex(targs) ((targ, idx) =>
+        def argumentPosition(idx: Int): Position = context.tree match {
+          case x: ValOrDefDef => x.rhs match {
+            case Apply(fn, args) if idx < args.size => args(idx).pos
```
retronym: I'm struggling to see how this is ever a useful position to report the warning (why is the n-th value parameter of an application related to the n-th type argument?), I'd suggest to drop this altogether.

retronym:
> The arg-specific code still backs up the isApplicable path.

Let's try to disable this warning logic in the `isApplicable` path. Those cases are probably running in a silent mode reporter. If a `Tree` falls afoul of the linter, but there isn't a reporter to hear it, does it make a sound?

som-snytt:
I had to retweet that.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
som-snytt: I don't recall where I was on Nov 18. Possibly I committed this from a dive bar.

retronym: The part that puzzles me dates back to eb2375cc53, but I can't say what sort of bar it was written in.

som-snytt: A `foo bar`, of course.

som-snytt:
OK, I refreshed my memory and reconstituted retronym's last remark about silent mode.

I see I'd already linked SI-9211 which retronym asked for a long time ago now.

The goal would be for the args which contribute to the undesired inference to be highlighted in red in the REPL.

Also, there's Nothing to see during overload resolution.
```
scala> def f[A](x: A, y: A, z: Int) = x.toString + y.toString + z.toString
f: [A](x: A, y: A, z: Int)String

scala> f(17, "42", 0)
<console>:13: warning: a type was inferred to be `Any`; this may indicate a programming error.
       f(17, "42", 0)
       ^

scala> f(17, "42", "0")
 found   : String("0")
 required: Int
       f(17, "42", "0")
                   ^

scala> object X {def f[A](x: A, y: A) = x.toString + y.toString ; def f[A](i: Int, j: String) = j.toInt + i }
defined object X

scala> X.f _
res2: (Int, String) => Int = $$Lambda$1940/2144962543@7209ffb5
```

PR #5547, opened 2016-11-19
Use equal() instead of identity in read-only TNode

rovarga:
Fix consistency of what equality function is used in case
of a tombed node with read-only snapshot, useful only in
the case the equality is customized.

### DIFF
--- a/src/library/scala/collection/concurrent/TrieMap.scala
+++ b/src/library/scala/collection/concurrent/TrieMap.scala
@@ -261,7 +261,7 @@ private[collection] final class INode[K, V](bn: MainNode[K, V], g: Gen) extends 
           clean(parent, ct, lev - 5)
           RESTART // used to be throw RestartException
         } else {
-          if (tn.hc == hc && tn.k == k) tn.v.asInstanceOf[AnyRef]
+          if (tn.hc == hc && equal(tn.k, k, ct)) tn.v.asInstanceOf[AnyRef]
           else null
         }
         cleanReadOnly(tn)


rovarga:
I just signed the CLA.

szeiger:
Can you add a regression test to show what bug is fixed by this change?

SethTisue:
also note that rebasing on top of current 2.12.x will make the MiMa error go away

rovarga:
@szeiger That may prove a bit problematic. It requires the equality function to be overridden, which I do not know how to make happen. I think I can then reduce the operations executed to get a reproducer.

I found this indirectly by tracking down https://bugs.opendaylight.org/show_bug.cgi?id=6577, where it (due to Java port) boiled down to identity vs. equality and was reproducible reliably under load (both in terms of size and operation rate).

lrytz:
CC @axel22, for review, and maybe an idea for testing.

axel22:
Yep, @rovarga had already pinged me. The changes look correct to me!

As for testing - I suggest adding a ScalaCheck test in which P threads concurrently remove entries from an initially full `TrieMap` (1-2 millions of elements), and Q threads concurrently read entries from this map. The key type `K` must be a class whose `equals` is overridden to throw an exception, and the custom equality should work correctly. That way, if any part of the `TrieMap` implementation still uses normal `equals` instead of the equality typeclass, then a thread will very likely get crashed by the exception in the `equals` implementation.

axel22:
@rovarga Something along these lines:

```
case class CustomKey(id: Int) {
  override def equals(that: AnyRef) = sys.error("hasta la vista!")
}

class CustomKeyEquiv extends Equiv[CustomKey] {
  def equiv(k1: CustomKey, k2: CustomKey) = k1.id == k2.id
}

property("no vanilla equals") = forAll { (n: Int, p: Int, q: Int) =>
  val count = new AtomicInteger(p + q)
  val trie = new TrieMap[CustomKey, Int](new CustomKeyEquiv)
  for (i <- 0 until n) trie.put(i, i)
  val removers = for (i <- 0 until p) yield new Thread {
    override def run() = {
      for (j <- 0 until n) trie.remove(CustomKey(j))
      count.decrementAndGet()
    }
  }
  val lookers = for (i <- 0 until q) yield new Thread {
    override def run() = {
      for (j <- 0 until n) trie.get(CustomKey(j))
      count.decrementAndGet()
    }
  }
  removers.foreach(_.start())
  lookers.foreach(_.start())
  removes.foreach(_.join())
  lookers.foreach(_.join())
  count.get == 0
}
```

Of course, you should validate that the exception is triggered above without your change.

rovarga:
@axel22 Thanks, I'll see if I can find some time to add it. One thing I found while dissecting the Java port, though, is that LNodes are afflicted by this, too -- unless I am missing how TrieMap.ef is passed down to ListMap...

SethTisue:
@rovarga where do we stand, here?

SethTisue:
@rovarga ping?

> LNodes are afflicted by this, too

does this mean further changes are needed...?

rovarga:
I am sorry I cannot devote time to this issue due to churn in my primary project :( realistic ETA for getting out of that is somewhere around mid-April, I am afraid.

This LNode part can be treated separately, as it occurs only in case:
a) a custom equality function is used, and
b) there is a hash collision, and
c) colliding keys compare as equal using default equality

If all of these hold, LNode's ListMap can hold only one of the keys, hence the TrieMap will loose a mapping.

SethTisue:
closing for inactivity. can always be reopened, or resurrected in a new PR

PR #5549, opened 2016-11-21
Whitelist the remaining changes since 2.12.0 that break all builds

szeiger:
The changes were made in https://github.com/scala/scala/pull/5481,
subsequently breaking binary compatibility checks after
https://github.com/scala/scala/pull/5532 was merged, too. The affected
methods are part of an internal implementation class. Whitelisting
should be safe.

### DIFF
--- a/bincompat-backward.whitelist.conf
+++ b/bincompat-backward.whitelist.conf
@@ -12,7 +12,10 @@ filter {
     {
       matchName="scala.reflect.runtime.SymbolLoaders#TopClassCompleter.this"
       problemName=IncompatibleMethTypeProblem
+    },
+    {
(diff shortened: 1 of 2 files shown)


som-snytt:
I put them in https://github.com/scala/scala/pull/5546 but FSR after two shots put them in both listings.

SethTisue:
LGTM

szeiger:
Merging to make our builds succeed again, just in case #5546 will take a bit longer before it gets merged. @som-snytt The "generated filters" output of MiMa also tripped me up. Our forwards check is done by running a backwards check with swapped versions, so the filters will always be listed for "backwards" even when they really need to be used for "forwards". When I wrote that code MiMa didn't have a forwards mode yet.

som-snytt:
@szeiger Thanks!

PR #5550, opened 2016-11-21
SI-3772 Fix detection of term-owned companions

retronym:
Companion detection consults the scopes of enclosing Contexts during
typechecking to avoid the cycles that would ensue if we had to look
at into the info of enclosing class symbols. For example, this used
to typecheck:

    object CC { val outer = 42 }
    if ("".isEmpty) {
      case class CC(c: Int)
      CC.outer
    }

This logic was not suitably hardened to find companions in exactly
the same nesting level.

After fixing this problem, a similar problem in `Namer::inCurrentScope`
could be solved to be more selective about synthesizing a companion
object. In particular, if a manually defined companion trails after
the case class, don't create an addiotional synthetic comanpanion object.

JIRA: https://issues.scala-lang.org/browse/SI-3772


retronym:
@som-snytt Could you think of any additional test cases for this?

som-snytt:
This progresses to a VerifyError. (https://issues.scala-lang.org/browse/SI-8733)

I was thinking of those odd constructor contexts. Maybe there's a better corner case; I've forgotten the context.

```
scala> class X(x: Int) { def f = x }
defined class X

scala> class Y extends X({ case class C(c: Int) ; object C { val x = 42 } ; C(17).c + C.x })
defined class Y

scala> new Y().f
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    Y.<init>()V @11: invokespecial
  Reason:
    Type uninitializedThis (current frame, stack[1]) is not assignable to 'Y'
  Current Frame:
    bci: @11
    flags: { flagThisUninit }
    locals: { uninitializedThis, 'scala/runtime/LazyRef' }
    stack: { uninitializedThis, uninitializedThis, 'scala/runtime/LazyRef' }
  Bytecode:
    0x0000000: 2abb 001f 59b7 0038 4c2a 2bb7 003a 1011
    0x0000010: b600 3eb6 0042 2a2b b700 3ab6 0045 60b7
    0x0000020: 0048 b1

  ... 31 elided

scala> class C extends { val x = { case class C(c: Int) ; object C { val x = 42 } ; C(17).c + C.x } } with AnyRef { def f = x }
defined class C

scala> new C().f
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    C.<init>()V @10: invokespecial
  Reason:
    Type uninitializedThis (current frame, stack[0]) is not assignable to 'C'
  Current Frame:
    bci: @10
    flags: { flagThisUninit }
    locals: { uninitializedThis, top, 'scala/runtime/LazyRef' }
    stack: { uninitializedThis, 'scala/runtime/LazyRef' }
  Bytecode:
    0x0000000: bb00 2259 b700 394d 2a2c b700 3b10 11b6
    0x0000010: 003f b600 422a 2cb7 003b b600 4360 3c2a
    0x0000020: 1bb5 0018 2ab7 0044 b1

  ... 31 elided

```

som-snytt:
There was a test for your other case that fails now:

test/files/pos/t8002-nested-scope.scala

som-snytt:
What if `partest --update-check` also shuttled tests between `pos` and `neg`.

src/reflect/scala/reflect/internal/Scopes.scala
```diff
@@ -282,6 +282,30 @@ trait Scopes extends api.Scopes { self: SymbolTable =>
...
+      null
+    }
+
+    final def lookupCompanion(original: Symbol): Symbol = {
+      lookupSymbolEntry(original) match {
+        case null =>
+        case entry if entry.sym == original =>
+          entry.owner.lookupAllEntries(original.name.companionName)
+            .filter(_.owner eq entry.owner)
+            .filter(x => original.isTerm || x.sym.hasModuleFlag).toList match {
```
lrytz: A comment on this filter would help - i guess `original` is either a class symbol or a module symbol? I see it was copied from the original code, but it's not very clear. Also, the original code had a check for `isCoDefinedWith` - is this no longer needed?

retronym: I've re-added the `isCodefinedWith` check, moved some code around, and added a comment for the `hasModuleFlag` check.

src/reflect/scala/reflect/internal/Scopes.scala
```diff
@@ -282,6 +282,35 @@ trait Scopes extends api.Scopes { self: SymbolTable =>
...
+        if (e.sym == sym) return e
+        e = lookupNextEntry(e)
+      }
+      null
+    }
+
+    final def lookupCompanion(original: Symbol): Symbol = {
+      lookupSymbolEntry(original) match {
+        case null =>
+        case entry if entry.sym == original =>
```
lrytz: Looking at the implementation of `lookupSymbolEntry`, the guard here seems redundant.

src/reflect/scala/reflect/internal/Scopes.scala
```diff
@@ -282,6 +282,35 @@ trait Scopes extends api.Scopes { self: SymbolTable =>
...
+          var e = lookupEntry(original.name.companionName)
+          while (e != null) {
+            // 1) Must be owned by the same Scope, to ensure that in
+            //   `{ class C; { ...; object C } }`, the class is not seen as a comaniopn of the object.
+            // 2) Must be in exactly the same scope, so that `{ class C; def C }` are not companions
+            // TODO Exclude type aliases as companions of terms?
```
lrytz: yes, we should. `sym.companionClass` also does since https://github.com/scala/scala/pull/5482/commits

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -1192,6 +1192,27 @@ trait Contexts { self: Analyzer =>
       }
       res
     }
+
+    final def lookupCompanionOf(original: Symbol): Symbol = {
+      lookupScopeEntry(original) match {
+        case null => NoSymbol
+        case entry => entry.owner.lookupCompanion(original).filter(_.isCoDefinedWith(original))
```
lrytz: What is the reason to put the filter here, instead of inside `lookupCompanion`?

retronym:
Pushed a commit with your suggestions. We could go a bit further and change `Symbol#companion{Class,Module}` to internally delegate to `flatOwnerInfo.decls.lookupCompanion(this)`, but I wouldn't make that change without benchmarking.

lrytz:
LGTM, thanks!

PR #5551, opened 2016-11-21
Override equals and hashCode for WrappedArray

alexeyr:
Avoid unnecessary boxing when calling `equals` or `hashCode` on `WrappedArray`s.

### DIFF
--- a/src/library/scala/collection/mutable/WrappedArray.scala
+++ b/src/library/scala/collection/mutable/WrappedArray.scala
@@ -13,8 +13,12 @@ package collection
 package mutable
 
 import scala.reflect.ClassTag
+import scala.runtime.BoxedUnit
 import scala.collection.generic._
 import scala.collection.parallel.mutable.ParArray
+import scala.util.hashing.MurmurHash3
+
+import java.util.Arrays
 
 /**
  *  A class representing `Array[T]`.
@@ -125,6 +129,11 @@ object WrappedArray {
     def length: Int = array.length
     def apply(index: Int): T = array(index).asInstanceOf[T]
     def update(index: Int, elem: T) { array(index) = elem }
+    override def hashCode = MurmurHash3.wrappedArrayHash(array)
+    override def equals(that: Any) = that match {
+      case that: ofRef[_] => Arrays.equals(array.asInstanceOf[Array[AnyRef]], that.array.asInstanceOf[Array[AnyRef]])
+      case _ => super.equals(that)
+    }
   }
 
   final class ofByte(val array: Array[Byte]) extends WrappedArray[Byte] with Serializable {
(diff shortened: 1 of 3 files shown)


alexeyr:
The error is `- library/mima failed: java.lang.RuntimeException: MiMa failed with exit code 2`. Is it something I need to handle?

szeiger:
@alexeyr See the log of `validate-test` for details. The `private[scala]` methods are not binary compatible at the JVM level. Should be OK to whitelist though.

src/library/scala/collection/mutable/WrappedArray.scala
```diff
@@ -125,68 +129,118 @@ object WrappedArray {
...
   }
 
   final class ofUnit(val array: Array[Unit]) extends WrappedArray[Unit] with Serializable {
     def elemTag = ClassTag.Unit
     def length: Int = array.length
     def apply(index: Int): Unit = array(index)
     def update(index: Int, elem: Unit) { array(index) = elem }
+    override def hashCode = MurmurHash3.wrappedArrayHash(array)
+    override def equals(that: Any) = that match {
+      case that: ofUnit => Arrays.equals(array.asInstanceOf[Array[AnyRef]], that.array.asInstanceOf[Array[AnyRef]])
```
xuwei-k: just compare by `length`?

```scala
  case that: ofUnit => array.length == that.array.length
```

alexeyr: That was my initial thought, but we could somehow end up with an `Array[Unit]` containing `null` at runtime and I decided to err in favor of safety.

szeiger: If you're hoping for any kind of correctness, `null` would have to be treated as being identical to `()` in this case. This is true for all primitive types. You can get `null` instead of the type's proper default value due to the way in which Scala abstracts over primitive types (and practically atavistic types such as `void`).

Just some of the weirdness:

```
scala> class A[T] { var x: T = _ } // a convenient way to get nulls of primitive types
defined class A

scala> new A[Unit]
res0: A[Unit] = A@359b650b

scala> res0.x // looks like we have ()

scala> println(res0.x) // null leaking when "converting" back to AnyRef
null

scala> res0.x == ((): Unit) // even scalac believes that this has to be true
<console>:13: warning: comparing values of types Unit and Unit using `==' will always yield true
       res0.x == ((): Unit)
              ^
res4: Boolean = false

scala> Array[Unit](res0.x)
res6: Array[Unit] = Array(())

scala> res6(0)

scala> println(res6(0)) // whoa, this one's actually correct
()

scala> new Array[Unit](1)
res9: Array[Unit] = Array(())

scala> println(res9(0)) // default is null again
null

scala> res9(0) = res0.x

scala> println(res9(0)) // still null
null
```

lrytz: See also https://issues.scala-lang.org/browse/SI-602

alexeyr: Changed.

alexeyr:
There is also the question of whether overrides for `ArrayOps` should be added too. On one hand, `ArrayOps` objects should only be a result of implicit conversion and immediately disappear, on the other I don't think there are any drawbacks to providing the overrides.

szeiger (approved):

szeiger:
I wouldn't bother with `ArrayOps`. It's supposed to be only implicitly created. Unless you perform the conversion explicitly or use an explicit type ascription (neither of which I would want to encourage), you cannot call `equals` or `hashCode` on it.

szeiger:
Ah, I should have merged this before merge conflicts start creeping in. The Conflict will probably resolve itself after #5590 though.

adriaanm:
Closing in favor of the rebased one.

PR #5552, opened 2016-11-21
added implicit conversion from try to iterable (like option)

timvw:


szeiger:
I think implicitly discarding errors would be a bad idea. Note that there is also no implicit conversion from `Either` to `Option` even though it is right-biased in 2.12. Unlike `Option`, both `Try` and `Either` contain a failure value.

SethTisue:
yeah, I don't think we should move in this direction. the conversion from `Option` to `Iterable` is pretty questionable already. (and as Stefan points out, this one is actually lossy, Option to Iterable isn't.) in general in the standard library we are moving towards explicit conversions and away from implicit ones.

(is `toList` currently the only/best way to get an Iterable?)

szeiger:
`toList` is also added in this PR. You could use `toOption` and then rely on the implicit conversion from `Option` to `Iterable`.

SethTisue:
mm hmm, or explicit `.toOption.toIterable`

timvw:
Try already has a toOption (which is biased). From that point of view it makes sense to me to have an implicit conversion to Iterable as well (shortcutting the .toOption.toList path).

When the argument is that we don't like biased functions, the implicit conversion should be removed from Option as well. Right?

timvw:
+ Agree on favoring explicit over implicit 
+ Agree on Option.toIterable not being lossy, unlike Try/Either

PR #5553, opened 2016-11-22
Improve performance of REPL autocompletion

retronym:
The code used to fuzzily match, e.g, `declasses` with `getDeclaredClasses`
was exploring fruitless parts of the search space. The enclosed test
case was hanging the REPL.

This commit improves this by performing a prefix match of the unconsumed input
against the current chunk of the candidate before exploring the `inits`.

Fixes scala/scala-dev#271

Review by @som-snytt. Looks like this was just a dumb bug rather than
something calling for a brilliant algorithm. But there might be pathological
cases lurking that still could be improved.

### DIFF
--- a/src/interactive/scala/tools/nsc/interactive/Global.scala
+++ b/src/interactive/scala/tools/nsc/interactive/Global.scala
@@ -1202,7 +1202,8 @@ class Global(settings: Settings, _reporter: Reporter, projectName: String = "") 
             case Nil => entered.isEmpty && matchCount > 0
             case head :: tail =>
               val enteredAlternatives = Set(entered, entered.capitalize)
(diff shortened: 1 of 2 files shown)


retronym:
I'm going to merge this one without review as I'm happy the test covers against regressions fairly well and the performance bug is so glaring that I'd like to get this into 2.11.x and forward merged to 2.12.x in time for 2.12.1.

som-snytt:
LGTM

post-facto. The glare makes me avert my eyes. Sorry, just off a three-week flu, holiday and now back to the office. Plus, I'm lazy. And it's choral concert week. I hadn't looked at this code last year, but every opportunity to review code is a future opportunity to chip in. I'll try to stay no more than a step or two behind.

PR #5554, opened 2016-11-23
SI-10009 Fields survive untypecheck/retypecheck

retronym:
Some places in the compiler, and many places in macros, use
`untypecheck` (aka `resetAttrs`) to strip types and local symbols
from a tree before retypechecking it under some different context.

The refactoring of the desugaring of vals and vars in Scala 2.12.0
broke an assumption in this facility.

When a ValDef must be split into multiple members (e.g. a field and
a getter, or a perhaps also a setter), the ValDef that was parsed
assumes the role of the `field`, and the trees for other members are
stached by `Namer` to the `synthetics` map of the compilation unit,
in order to spliced into the right statement list by typechecking.
See `enterGetterSetter` for more details.

However, the parsed ValDef is now used verbatim, carrying the meaning
(ie, the symbol) of the `private[this]` field. This tree now had
an inconsistency between the flags in `tree.mods.flags` and
`tree.symbol.flags`. `tree.name` also differed from `tree.symbol.name`
(the latter was renamed to be a local name, ie one with a trailing space.)

When `ResetAttrs` stripped off the symbol and we retypechecked, we'd
end up with two symbols in scope with the same name.

In the first from the `run` test:

```
================================================================================
{
  class a extends scala.AnyRef {
    def <init>(): a = {
      a.super.<init>();
      ()
    };
    private[this] val x: Int = 42;
    <stable> <accessor> def x: Int = a.this.x
  };
  new a()
}

{
  class a extends scala.AnyRef {
    def <init>() = {
      super.<init>();
      ()
    };
    val x = 42; // oops, the name is "x" rather than "x " and we've missing `private[this]`!
    <stable> <accessor> def x: Int = a.this.x
  };
  new a()
}

scala.tools.reflect.ToolBoxError: reflective typecheck has failed: x is already defined as value x
```

This commit uses the flags and name of the symbol in `typedValDef`.
I've also had to modify the internals of `CodePrinter` to use the
implicit, override, and deferred flags from the modifiers of an
accessor when recovering pre-typer tree for a ValDef.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/MethodSynthesis.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/MethodSynthesis.scala
@@ -132,7 +132,11 @@ trait MethodSynthesis {
 
       // only one symbol can have `tree.pos`, the others must focus their position
       // normally the field gets the range position, but if there is none, give it to the getter
+      //
+      // SI-10009 the tree's modifiers can be temporarily out of sync with the new symbol's flags.
+      //          typedValDef corrects this later on.
       tree.symbol = fieldSym orElse (getterSym setPos tree.pos)
+
       val namer = namerOf(tree.symbol)
 
       // the valdef gets the accessor symbol for a lazy val (too much going on in its RHS)
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -13,11 +13,12 @@ package scala
 package tools.nsc
 package typechecker
 
-import scala.collection.{mutable, immutable}
-import scala.reflect.internal.util.{ Statistics, ListOfNil }
+import scala.collection.{immutable, mutable}
+import scala.reflect.internal.util.{ListOfNil, Statistics}
 import mutable.ListBuffer
 import symtab.Flags._
 import Mode._
+import scala.reflect.macros.whitebox
 
 // Suggestion check whether we can do without priming scopes with symbols of outer scopes,
 // like the IDE does.
@@ -2020,7 +2021,12 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
     // use typedValDef instead. this version is called after creating a new context for the ValDef
     private def typedValDefImpl(vdef: ValDef) = {
       val sym = vdef.symbol.initialize
-      val typedMods = typedModifiers(vdef.mods)
+      val typedMods = if (nme.isLocalName(sym.name) && sym.isPrivateThis && !vdef.mods.isPrivateLocal) {
+        // SI-10009 This tree has been given a field symbol by `enterGetterSetter`, patch up the
+        // modifiers accordingly so that we can survive resetAttrs and retypechecking.
+        // Similarly, we use `sym.name` rather than `vdef.name` below to use the local name.
+        typedModifiers(vdef.mods.copy(flags = sym.flags, privateWithin = tpnme.EMPTY))
+      } else typedModifiers(vdef.mods)
 
       sym.annotations.map(_.completeInfo())
       val tpt1 = checkNoEscaping.privates(sym, typedType(vdef.tpt))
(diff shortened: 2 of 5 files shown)


src/compiler/scala/tools/nsc/typechecker/MethodSynthesis.scala
```diff
@@ -125,14 +125,21 @@ trait MethodSynthesis {
     def enterGetterSetter(tree: ValDef): Unit = {
       val fieldSym =
         if (noFieldFor(tree, owner)) NoSymbol
-        else owner.newValue(tree.name append NameTransformer.LOCAL_SUFFIX_STRING, tree.pos, tree.mods.flags & FieldFlags | PrivateLocal)
+        else owner.newValue(tree.name append NameTransformer.LOCAL_SUFFIX_STRING, tree.pos, tree.mods.flags & (FieldFlags | DEFERRED) | PrivateLocal)
```
adriaanm: why add `DEFERRED`?

src/compiler/scala/tools/nsc/typechecker/MethodSynthesis.scala
```diff
@@ -125,14 +125,21 @@ trait MethodSynthesis {
...
       val getter = Getter(tree)
       val getterSym = getter.createSym
 
       // only one symbol can have `tree.pos`, the others must focus their position
       // normally the field gets the range position, but if there is none, give it to the getter
-      tree.symbol = fieldSym orElse (getterSym setPos tree.pos)
+      if (fieldSym != NoSymbol) {
+        tree.symbol = fieldSym
+        val fieldValDef: global.ValDef = treeCopy.ValDef(tree, tree.mods.copy(flags = fieldSym.flags, privateWithin = tpnme.EMPTY), fieldSym.name, tree.tpt, tree.rhs)
+        context.unit.transformed(tree) = fieldValDef
```
adriaanm: Do we need to store this here, only to retrieve it in `typedValDef`? Isn't the needed info still available there?

src/compiler/scala/tools/nsc/typechecker/MethodSynthesis.scala
retronym: Oh, I forgot to explain that change in the commit comment. 

This isn't adding the `DEFERRED` flag, but rather adding it to the set of flags that are copied from the tree to the symbol. Abstract vals in traits were not being emitted in code gen without this change. I'll collect a trace of what was going wrong to justify this.

src/compiler/scala/tools/nsc/typechecker/MethodSynthesis.scala
retronym: Yes, we can also solve this with a special case in `typedValDef` to fix up the modifiers. I thought I'd start with this proposal as it reuses an established namer/typer comms channel. But perhaps for 2.12.1 the special case is less risky.

adriaanm: I like your suggestion of using typedValDef better than this approach, I think. I'm not wild about that comms channel of transformed trees.

retronym: Okay, I'll try that. That comms channel is a necessary evil to avoid double-typechecking. We should probably replace the current `Map[Tree, Tree]` with something more localized (add the typed tree to the untyped tree as a tree or symbol attachment and have the typer pick it up from there.

adriaanm:
LGTM! Thank you.

SethTisue:
could this have caused https://issues.scala-lang.org/browse/SI-10163 ?

benmccann:
@retronym @adriaanm would you be able to take a look at the issue Seth linked to?

adriaanm:
I believe we had already followed up at https://github.com/scala-ide/scala-refactoring/issues/180#issuecomment-274313163. Could you ping us there if there's still an issue with this?

PR #4395, opened 2015-03-24
SI-9208 Phase out ad-hoc, incorrect implicit shadowing detection

retronym:
Paraphrasing Greenspun:

> Any sufficiently complicated Implicit Search program contains an
> ad hoc, informally-specified, bug-ridden, slow implementation of
> half of `typedIdent`.

The shadowing detection code that is being phased out was exactly
this. We have since heeded the aphorism above and perform
shadowing detection after typechecking a candidate implicit
(in 9129cfe911).

But, we didn't remove the pre-filtering of candidated based on
the old shadowing checks. This commit does exactly that. As this
introduces a minor source incompatibility, it is predicated on
-Xsource:2.12.

Review by @som-snytt, @adriaanm

PR #5555, opened 2016-11-24
SI-9208 Retire implicit shadower

som-snytt:
Retronym's fix. See what breaks.

https://github.com/scala/scala/pull/4395


som-snytt:
```
apm@mara:~/projects/snytt$ skala
Welcome to Scala 2.12.1-20161123-094816-d5f21af (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_111).

scala> object X { implicit val postfixOps = language.postfixOps }
defined object X

scala> { import language.postfixOps ; import X._ ; "42" toInt }
res0: Int = 42

scala> :quit
apm@mara:~/projects/snytt$ scala

scala> object X { implicit val postfixOps = language.postfixOps }
defined object X

scala> { import language.postfixOps ; import X._ ; "42" toInt }
<console>:13: warning: postfix operator toInt should be enabled
by making the implicit value scala.language.postfixOps visible.
This can be achieved by adding the import clause 'import scala.language.postfixOps'
or by setting the compiler option -language:postfixOps.
See the Scaladoc for value scala.language.postfixOps for a discussion
why the feature should be explicitly enabled.
       { import language.postfixOps ; import X._ ; "42" toInt }
                                                        ^
res0: Int = 42
```

som-snytt:
```
[error] /home/jenkins/workspace/scala-2.12.x-validate-test/src/library/scala/collection/mutable/TreeMap.scala:145: ambiguous implicit values:
[error]  both value ordering in class TreeMap of type => Ordering[A]
[error]  and value ordering in class TreeMap of type => Ordering[A]
[error]  match expected type Ordering[A]
[error]     override def get(key: A) = if (isInsideViewBounds(key)) RB.get(tree, key) else None
```

milessabin:
I wonder if we should consider eliminating implicit shadowing altogether?

som-snytt:
note to self: that last change was an attempt to say, collect implicits naively from contexts, then filter for "accessible using simple name" rule using ordinary lookup. Maybe draconian or the type test is breaking.

propensive:
@milessabin Definitely worth *considering*, but I would miss it if there weren't some other way of unimporting implicits.

som-snytt:
I cherry-picked the commits fixing binding precedence, for convenience. A shadow of the shadower remains to prefilter eligible implicits by name lookup. Probably the few lines that verify that the lookup result and the implicit are the same thing can be improved. One nuance is that the lookup must be non-recording for purposes of `-Ywarn-unused`. One ancient test was promoted to pos.

som-snytt:
There are two tests to address, one is scaladoc, the other is reflect.

In `test/files/run/virtpatmat_typetag.scala`,
```
trait X { type T ; implicit val t: ClassTag[T] }
trait E extends X { type T = Something ; val t = classTag[T] }
```
suffers some type-forcing. Annotating `t` makes it work. But if the PR to have `t` inferred from `super`, it would probably work as-is. This doesn't seem like a pernicious cycle to me, but I don't know much. I wonder what the community build would say.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -864,12 +842,39 @@ trait Implicits {
...
+        }
+        !duplicates(info.name) && lookedup
+      }
+
       /** True if a given ImplicitInfo (already known isValid) is eligible.
        */
       def survives(info: ImplicitInfo) = (
-           !isIneligible(info)                      // cyclic, erroneous, shadowed, or specially excluded
+           !isIneligible(info)                      // cyclic, erroneous, or specially excluded
+        && (!isLocalToCallsite || looksupOK(info))  // name is accessible
```
retronym, line 876: `looksupOK` seems too expensive to do at this point. It is effectively the same amount of work as `typedIdent`, we don't want to have to pay that cost (execution time and potential for spurious cycles) for every in-scope implicit.

som-snytt: yes, I realized you wanted to just do it once and ignore shadowed implicits either in `context.implicitss` or after when typechecking. I'll try to come up with something.

lrytz:
Labelled WIP; this needs to be reviewed to get in 2.12.5.

som-snytt:
@lrytz Thanks I'll try to get back to this ASAP. How much time do I have for 2.12.5?

SethTisue:
> How much time do I have for 2.12.5?

https://contributors.scala-lang.org/t/getting-ready-for-2-12-5/1618

som-snytt:
@SethTisue Thanks! You're a help!

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -1109,22 +1103,30 @@ trait Contexts { self: Analyzer =>
...
+       *  3) Wildcard imports have next highest precedence.
+       *  4) Definitions made available by a package clause not in the compilation unit where
+       *     the reference occurs have lowest precedence.
+       */
+      def depthOk(imp: ImportInfo) = imp.depth > symbolDepth || (
+           imp.isExplicitImport(name)
+        && imp.depth == symbolDepth
+        && (
+              unit.isJava
+           || (defSym != NoSymbol && isPackageOwnedInDifferentUnit(defSym))  // SI-2458
```
retronym: I think we need to stay bug-compatible with the long standing behaviour under 2.12. I'd suggest predicating all the fixes in the PR on `-Xsource:2.13`.

som-snytt: "Copy that."

adriaanm:
When you get a chance to apply another round of polish, please resubmit as a PR for 2.13.0-M5.

retronym:
I'm very keen to see the implicit shadower retire into the shadows, as it is a major source of allocation pressure during typechecking. I feel like I've already asked you, but I can't remember, so I'll ask again: what is the relationship between that part of the change and the binding precendence change? Can we get the former without the latter?

I know the hairy part about the removing the implicit shadower, is that it seems to currently obscure (shadow?) a bug in `Context.implicitss` that duplicates decl implicits (once for the member in scope, and again for the members of the self type.) That should be fixable.

som-snytt:
The binding precedence fix is a no-brainer in its current form. FWIW. I'll get back on the on-ramp for this fix. Contra Adriaan, I'll follow up on both 2.1[23].x.

retronym:
My latest take on the implicit shadower: https://github.com/scala/scala/pull/7757

diesalbla:
@propensive Regarding [your previous comment](https://github.com/scala/scala/pull/5555#issuecomment-354740994), what kind of things would not be possible or easy without the implicit shadowing?

som-snytt:
Renaming to "Fix implicit shadowing before I retire".

Implicit shadowing is removed from Dotty in favor of preferring the definition at greater context depth.

PR #5556, opened 2016-11-24
SI-10071 Separate compilation for varargs methods

dragos:
Make sure that methods annotated with varargs are properly mixed-in. This commit
splits the transformation into an info transformer (that works on all symbols, whether
they come from source or binary) and a tree transformer.

The gist of this is that the symbol-creation part of the code was moved to the UnCurry
info transformer, while tree operations remained in the tree transformer. The newly
created symbol is attached to the original method so that the tree transformer can still
retrieve the symbol.

A few fall outs:

- I removed a local map that was identical to TypeParamsVarargsAttachment
- moved the said attachment to StdAttachments so it’s visible between reflect.internal
  and nsc.transform
- a couple more comments in UnCurry to honour the boy-scout rule

review by @lrytz


src/reflect/scala/reflect/internal/transform/UnCurry.scala
```diff
@@ -67,8 +74,25 @@ trait UnCurry {
       tp match {
         case ClassInfoType(parents, decls, clazz) =>
           val parents1 = parents mapConserve uncurry
-          if (parents1 eq parents) tp
-          else ClassInfoType(parents1, decls, clazz) // @MAT normalize in decls??
+          val varargOverloads = mutable.ListBuffer.empty[Symbol]
```
retronym, line 77: You could skip the new logic here if `clazz.isJavaDefined`.

retronym: In fact, I think we could skip the rest of the info transform in that case. Worth an experiment.

test/files/jvm/varargs-separate-bytecode/Props_1.flags
```diff
@@ -0,0 +1 @@
+-sourcepath test/files/jvm/varargs-separate-bytecode/
```
retronym: I haven't seen this done before in partest. What's the intent? Is there a simpler way?

src/reflect/scala/reflect/internal/transform/UnCurry.scala
dragos: Good idea, I'll give it a go.

test/files/jvm/varargs-separate-bytecode/Props_1.flags
dragos: You're right, it's being used in presentation compiler tests only. What I need is to make sure that `AbstractProps` is not compiled together with `Props.scala`. Is there a way to compile first one set of files and afterwards another set of files?

dragos: Ok, it looks like appending _1 and _2 to source files makes them compile in separate compiler runs, so my last commit simplifies the test to not use `-sourcepath`. I can squash the commits later, for the moment I'll keep them as separate, so reviewers can track what's new.

src/compiler/scala/tools/nsc/transform/UnCurry.scala
```diff
@@ -739,68 +749,32 @@ abstract class UnCurry extends InfoTransform
...
+    /**
+     * Called during post transform, after the method argument lists have been flattened.
+     * It looks for the forwarder symbol in the symbol attachments and generates a Java-style
      * varargs forwarder.
+     *
+     * @note The Java-style varargs method symbol is generated in the Uncurry info transformer. If the
+     *       symbol can't be found this method reports a warning and carries on.
+     * @see  [[scala.reflect.internal.transform.UnCurry]]
      */
     private def addJavaVarargsForwarders(dd: DefDef, flatdd: DefDef): DefDef = {
```
dragos, line 761: This method signature is confusing, but I chose to keep things simple and I didn't touch it.

`flatdd` is not required since almost all the logic is based on Symbols and the two trees always have the same symbol. Additionally, the only legitimate use of a tree is in the call to `newTyper(dd,..)`, though I'm not sure that there's no other way. If you have any suggestions here I'm happy to at least remove the extra parameter.

src/reflect/scala/reflect/internal/transform/UnCurry.scala
```diff
@@ -67,8 +74,29 @@ trait UnCurry {
       tp match {
         case ClassInfoType(parents, decls, clazz) =>
           val parents1 = parents mapConserve uncurry
```
lrytz, line 76: move this declaration into the `else` branch below?

src/reflect/scala/reflect/internal/transform/UnCurry.scala
```diff
@@ -67,8 +74,29 @@ trait UnCurry {
...
+            val varargOverloads = mutable.ListBuffer.empty[Symbol]
+
+            // Not using `hasAnnotation` here because of dreaded cyclic reference errors:
+            // it may happen that VarargsClass has not been initialized yet and we get here
+            // while processing one of its superclasses (such as java.lang.Object). Since we
+            // don't need the more precise `matches` semantics, we only check the symbol, which
+            // is anyway faster and safer
+            for (decl <- decls if decl.annotations.exists(_.symbol == VarargsClass)) {
+              if (mexists(decl.paramss)(sym => definitions.isRepeatedParamType(sym.tpe))) {
+                val forwarderSym = varargForwarderSym(clazz, decl, exitingPhase(phase)(decl.info))
```
lrytz: nit, but this local isn't necessary IMO, the code would be clear without

dragos:
Two things still pending:

- do we do anything about the parameters of `def addJavaVarargsForwarders`? I'm fine to leave it as-is, just making sure you've seen my comment
- squash, and how much?

Once this is in I will open a PR to backport the fix to 2.11.

src/reflect/scala/reflect/internal/transform/UnCurry.scala
```diff
@@ -65,18 +72,90 @@ trait UnCurry {
...
+
+          // Not using `hasAnnotation` here because of dreaded cyclic reference errors:
+          // it may happen that VarargsClass has not been initialized yet and we get here
+          // while processing one of its superclasses (such as java.lang.Object). Since we
+          // don't need the more precise `matches` semantics, we only check the symbol, which
+          // is anyway faster and safer
+          for (decl <- decls if decl.annotations.exists(_.symbol == VarargsClass)) {
+            if (mexists(decl.paramss)(sym => definitions.isRepeatedParamType(sym.tpe))) {
+              varargOverloads += varargForwarderSym(clazz, decl, exitingPhase(phase)(decl.info))
+            }
```
lrytz, line 87: could use `exitingUncurry` to be more explicit

lrytz:
LGTM, thanks a lot! In terms of squashing I think one commit for the `if (clazz.isJavaDefined) tp` change (this one could go first), and a second with the rest.

src/reflect/scala/reflect/internal/transform/UnCurry.scala
dragos: Unfortunately there is no such thing in `scala.reflect.internal`, only in the compiler cake.

lrytz: ah yes :)

src/reflect/scala/reflect/internal/transform/UnCurry.scala
```diff
@@ -65,18 +72,90 @@ trait UnCurry {
     def apply(tp0: Type): Type = {
       val tp = expandAlias(tp0)
       tp match {
-        case ClassInfoType(parents, decls, clazz) =>
+        case ClassInfoType(parents, decls, clazz) if !clazz.isJavaDefined =>
```
retronym, line 75: This oversight has come up twice in a week, I've opened a [ticket](https://github.com/scala/scala-dev/issues/276) to audit the other info transformers for similar opportunities to do less work!

src/reflect/scala/reflect/internal/transform/UnCurry.scala
```diff
@@ -65,18 +72,90 @@ trait UnCurry {
...
+        val newParam = oldParam.cloneSymbol(forwSym)
+        val tp = if (isRep) toArrayType(oldParam.tpe, newParam) else oldParam.tpe
+        newParam.setInfo(tp)
+      })
+
+      val resTp = newInfo.finalResultType.substSym(oldPs, ps)
+      val mt = MethodType(ps, resTp)
+      val r = if (tps.isEmpty) mt else PolyType(tps, mt)
+      r.substSym(oldTps, tps)
+    }
```
retronym: This is old code, but it is using very low level means to build up the new method symbol. I believe we can just clone the original symbol to get this for free.

https://github.com/retronym/scala/compare/1a21cc667a6f33a576900ea024c3bb21df84638e...retronym:review/5556?expand=11a21cc667a6f33a576900ea024c3bb21df84638e

Passes at least `sbt partest --grep varargs`.

@lrytz Do you remember any gotchas in this code that justify the current approach?

lrytz: no, your patch LGTM. @dragos you want to include it in this PR, maybe as a separate commit?

dragos: Ok, I'll add it as another commit and let the bot tell us what's going on

dragos:
If this is fine, I'd like to open the backport for 2.11.9. Can you please let me know if there's anything else here, or I can merge?

lrytz:
:ship: it!

dragos:
Wohoo! Thanks for the speedy review!

PR #5559, opened 2016-11-28
Use https links to JS and CSS in the spec when serving from https

szeiger:
The spec is published on a server that supports https
(https://www.scala-lang.org/files/archive/spec/2.11/) and this comes up
as the default in search results (as it should) but the link to
MathJAX is hardcoded to http, which prevents any web browser that cares
about security from loading it.

This commit changes the links to MathJAX and to the Highlight.js
stylesheet to be scheme-relative (like the link to JQuery already was).

### DIFF
--- a/spec/_layouts/default.yml
+++ b/spec/_layouts/default.yml
@@ -15,9 +15,9 @@
     }
   });
   </script>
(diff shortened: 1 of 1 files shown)


SethTisue:
LGTM. nice catch!

PR #5562, opened 2016-11-29
SD-275 Further harden against refs to absentee classes

retronym:
- Rework previous fixes for SI-8502 to move the crea
    a term or type stub symbol during unpickling to the
    point of stub creation, based on the tag.
  - Just set the PACKAGE flag on class stub symbols cre
    unpickling `ThisType`, rather than bothering with a
    subclass of `StubSymbol` for (assumed) packages.
  - Limit the strategy of unpickling an external reference to a
    module class to a lookup of the module var to non-stub owners
    in order to enable fall through to stub symbol creation.

Fixes scala/scala-dev#275

Review by @lrytz

### DIFF
--- a/src/reflect/scala/reflect/internal/pickling/UnPickler.scala
+++ b/src/reflect/scala/reflect/internal/pickling/UnPickler.scala
@@ -216,18 +216,6 @@ abstract class UnPickler {
             }
             adjust(decl)
         }
-        def nestedObjectSymbol: Symbol = {
-          // If the owner is overloaded (i.e. a method), it's not possible to select the
-          // right member, so return NoSymbol. This can only happen when unpickling a tree.
-          // the "case Apply" in readTree() takes care of selecting the correct alternative
-          //  after parsing the arguments.
-          if (owner.isOverloaded)
-            return NoSymbol
-
-          if (tag == EXTMODCLASSref && !owner.isInstanceOf[StubSymbol])
-            owner.info.decl(nme.moduleVarName(name.toTermName))
-          else NoSymbol
-        }
 
         def moduleAdvice(missing: String): String = {
           val module =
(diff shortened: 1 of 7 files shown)


retronym:
@lrytz I still think we should investigate ways to reduce unnecessary eager traversal of the symbol graph in the new backend, but until then this should provide some extra defence.

holograph:
Fast work, thanks @retronym (and hopefully @lrytz shortly :-))

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -224,10 +224,9 @@ abstract class UnPickler {
           if (owner.isOverloaded)
             return NoSymbol
 
-          if (tag == EXTMODCLASSref) {
+          if (tag == EXTMODCLASSref && !owner.isInstanceOf[StubSymbol])
             owner.info.decl(nme.moduleVarName(name.toTermName))
-          }
-          NoSymbol
+          else NoSymbol
```
lrytz: huh, `nestedObjectSymbol` consistently returned `NoSymbol` before - could we come up with an example that exploits this as a bug?

lrytz:
LGTM!

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
retronym: Oh right, I didn't even see that. Guess I was blinded by the early return above!

It's been a no-op since https://github.com/scala/scala/commit/df3689139c4d4bcd2933364d13b8195c3433eb43#diff-7bf0456105421b3217bd1d63b8c864d4L229, the original code was added in 020053c3215579e8aeb871a4ad0078516994270d.

I'll poke at it and see if there is a bug. Otherwise we can drop this tactic from unpickler altogether.

retronym: Looks like it was only needed to be able to unpickle nested module references compiled before a5d47fb693d9b88ea9ed414762f16e027be64ada where the refered-to class was compiled under the new scheme. I'll remove the code.

https://issues.scala-lang.org/browse/SI-1591 and https://issues.scala-lang.org/browse/SI-3879.

lrytz:
L even better TM!

holograph:
Any chance this can be backported to 2.11.x?

holograph:
Pinging @lrytz @retronym (sorry, I don't know if this is the right channel to ask for this sort of thing, please let me know if I'm out of line)

lrytz:
It look like it could be, but it's probably not going to make it to our priority list.. a PR would be welcome!

holograph:
I'm certain it would, but I wouldn't even know where to start :-)  I might give it a go if I find a free day or two.. no promises where I can't necessarily keep them though.

PR #5563, opened 2016-11-29
Don't exclude super calls to trait methods from inlining

lrytz:
In 8020cd6, the inliner was changed to make sure trait methods bodies
are not duplicated into the static super accessors, and from there into
mixin forwarders.

The check for mixin forwarders was too wide. In `def t = super.m`, where
`m` is a trait method annotated `@inline`, we want to inline `m`. Note
that `super.m` is translated to an `invokestatic T.m$`. The current
check incorrectly identifies `t` as a mixin forwarder, and skip
inlining.


lrytz:
Follow-up for #5529. Review by @retronym.

retronym:
LGTM

PR #5566, opened 2016-11-30
Hash table getOrElseUpdate and indexing 2.12 backport [ci: last-only]

l0rinc:
Backported by recommendation from @Ichoran and @retronym: https://github.com/scala/scala/pull/5528#issuecomment-260799361

### DIFF
--- a/src/library/scala/collection/mutable/HashTable.scala
+++ b/src/library/scala/collection/mutable/HashTable.scala
@@ -183,6 +183,7 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
         table(h) = e.next
         tableSize = tableSize - 1
         nnSizeMapRemove(h)
+        e.next = null
         return e
       } else {
         var e1 = e.next
@@ -194,6 +195,7 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
           e.next = e1.next
           tableSize = tableSize - 1
           nnSizeMapRemove(h)
+          e1.next = null
           return e1
         }
       }
@@ -227,8 +229,9 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
     var es        = iterTable(idx)
 
     while (es != null) {
+      val next = es.next // Cache next in case f removes es.
       f(es.asInstanceOf[Entry])
-      es = es.next
+      es = next
 
       while (es == null && idx > 0) {
         idx -= 1
--- a/src/library/scala/collection/mutable/HashTable.scala
+++ b/src/library/scala/collection/mutable/HashTable.scala
@@ -12,7 +12,7 @@ package scala
 package collection
 package mutable
 
-import java.lang.Integer.rotateRight
+import java.lang.Integer.{numberOfLeadingZeros, rotateRight}
 import scala.util.hashing.byteswap32
 
 /** This class can be used to construct data structures that are based
@@ -367,11 +367,7 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
     * Note: we take the most significant bits of the hashcode, not the lower ones
     * this is of crucial importance when populating the table in parallel
     */
-  protected final def index(hcode: Int): Int = {
-    val ones = table.length - 1
-    val exponent = Integer.numberOfLeadingZeros(ones)
-    (improve(hcode, seedvalue) >>> exponent) & ones
-  }
+  protected final def index(hcode: Int): Int = if (table.length == 1) 0 else improve(hcode, seedvalue) >>> numberOfLeadingZeros(table.length - 1)
 
   protected def initWithContents(c: HashTable.Contents[A, Entry]) = {
     if (c != null) {
--- a/src/library/scala/collection/mutable/HashTable.scala
+++ b/src/library/scala/collection/mutable/HashTable.scala
@@ -411,58 +411,23 @@ private[collection] object HashTable {
 
     protected def elemHashCode(key: KeyType) = key.##
 
-    protected final def improve(hcode: Int, seed: Int) = {
-      /* Murmur hash
-       *  m = 0x5bd1e995
-       *  r = 24
-       *  note: h = seed = 0 in mmix
-       *  mmix(h,k) = k *= m; k ^= k >> r; k *= m; h *= m; h ^= k; */
-      // var k = hcode * 0x5bd1e995
-      // k ^= k >> 24
-      // k *= 0x5bd1e995
-      // k
-
-      /* Another fast multiplicative hash
-       * by Phil Bagwell
-       *
-       * Comment:
-       * Multiplication doesn't affect all the bits in the same way, so we want to
-       * multiply twice, "once from each side".
-       * It would be ideal to reverse all the bits after the first multiplication,
-       * however, this is more costly. We therefore restrict ourselves only to
-       * reversing the bytes before final multiplication. This yields a slightly
-       * worse entropy in the lower 8 bits, but that can be improved by adding:
-       *
-       * `i ^= i >> 6`
-       *
-       * For performance reasons, we avoid this improvement.
-       * */
-      val i= scala.util.hashing.byteswap32(hcode)
-
-      /* Jenkins hash
-       * for range 0-10000, output has the msb set to zero */
-      // var h = hcode + (hcode << 12)
-      // h ^= (h >> 22)
-      // h += (h << 4)
-      // h ^= (h >> 9)
-      // h += (h << 10)
-      // h ^= (h >> 2)
-      // h += (h << 7)
-      // h ^= (h >> 12)
-      // h
-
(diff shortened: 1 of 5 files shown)


src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -408,58 +408,23 @@ private[collection] object HashTable {
...
+      * h = h ^ (h >>> 14)
+      * h = h + (h << 4)
+      * h ^ (h >>> 10)
+      * }}}
+      * the rest of the computation is due to SI-5293
+      */
+    protected final def improve(hcode: Int, seed: Int): Int = {
+      val hash = scala.util.hashing.byteswap32(hcode)
+      val shift = seed & ((1 << 5) - 1)
+      (hash >>> shift) | (hash << (32 - shift))
```
l0rinc: @Ichoran, @retronym, I just noticed this can be changed to `Integer.rotateRight`.

~~Let's not merge it yet, let's see if that's more optimal (should be replaced with a single assembly instruction).~~

l0rinc: Fixed :)

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -357,14 +360,14 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
...
+    * Note: we take the most significant bits of the hashcode, not the lower ones
+    * this is of crucial importance when populating the table in parallel
+    */
+  protected final def index(hcode: Int): Int = {
     val ones = table.length - 1
-    val improved = improve(hcode, seedvalue)
-    val shifted = (improved >> (32 - java.lang.Integer.bitCount(ones))) & ones
-    shifted
+    val exponent = Integer.numberOfLeadingZeros(ones)
+    (improve(hcode, seedvalue) >>> exponent) & ones
```
sjrd: Now I realize: since `ones` is of the form `00...0011...11`, I do not think there is any value of `x` for which `(x >>> exponent) & ones` is different from `(x >>> exponent)`. It seems to me that `& ones` is redundant and can be dropped. That's because `& ones` masks off the `exponent` most significant bits of `(x >>> exponent)`, which by construction already has `0`'s in all its `exponent` most significant bits.

l0rinc: I thought about that also (though I thought I still have to zero out the highest bit), but haven't validated my find yet.
Thanks, lemme' recheck :)

sjrd: > though I still have to zero out the highest bit

No, you don't, because `>>>` is a *logical* shift right, so it does not sign-extent.

Unless I completely misunderstood what you're saying here?

l0rinc:
```scala
(Integer.MAX_VALUE >>> 32 & 0) != (Integer.MAX_VALUE >>> 32)
```
i.e. if the length is `1`, `ones` will be `0`, `exponent` will be `32`, which would produce the same shift as `0` (i.e. `123 >>> 32` == `123`) ... other than that (i.e. if `length > 1`) this seems to be true indeed.
Should I simply assert that the `length >= 2` at the beginning?

Edit:
```scala
protected final def index(hcode: Int): Int = {
  assert(table.length >= 2)
  improve(hcode, seedvalue) >>> numberOfLeadingZeros(table.length - 1)
}
```
or simply
```scala
protected final def index(hcode: Int): Int = improve(hcode, seedvalue) >>> numberOfLeadingZeros(table.length - 1)
```
?

sjrd: Very good observation!

But I don't think we need to care. If `length == 1`, we do not care about the quality of hashes by definition, since everything enters the same bucket anyway. So this does not seem to be an issue. It probably deserves a comment, though.

> Should I simply assert that the length >= 2 at the beginning?

No, `assert` would kill all the performance of this performance-sensitive method.

sjrd: Oh I'm not paying enough attention tonight, it seems. This is the `index` method, which *needs* to return `0` for correctness when `length == 1` (assuming that ever happens).

I don't know whether it is reasonable to assume that `length` is never 1. If it is, then we can drop `& ones`. If it isn't, we need to keep the code as it currently is, but then it would probably be a good thing to add a comment that `& ones` is there specifically to handle `length == 1`, in case someone wonders in the future.

src/library/scala/collection/mutable/HashTable.scala
```diff
@@ -361,11 +364,7 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
     * Note: we take the most significant bits of the hashcode, not the lower ones
     * this is of crucial importance when populating the table in parallel
     */
-  protected final def index(hcode: Int): Int = {
-    val ones = table.length - 1
-    val exponent = Integer.numberOfLeadingZeros(ones)
-    (improve(hcode, seedvalue) >>> exponent) & ones
-  }
+  protected final def index(hcode: Int): Int = if (table.length == 1) 0 else improve(hcode, seedvalue) >>> numberOfLeadingZeros(table.length - 1)
```
l0rinc, line 370: @sjrd, thanks for the comments, is this too awkward?
Branch prediction should eliminate this fork.

sjrd: I don't know. At this point we're past my understanding of JVM performance. I would expect `& ones` (one CPU cycle with operands already loaded in registers so no stalling due to memory) to always be faster than a branch, even if correctly predicted and/or using a conditional move, but I cannot assert that.

retronym: Instruction Level Parallelism can achieve > 1 instruction/clock (IPC), so a predictable branch guarding an operation might in fact be faster than a branchless algorithm with a data dependency.

l0rinc: Thanks @retronym, and also, the JVM can detect that the first branch is never taken and eliminate it completely :)

l0rinc:
@szeiger, what should I do to avoid all these build failures, I can't even check the source of the `combined` :/

szeiger:
@paplorinc `combined` just means that there are earlier commits for which the build failed. I don't think we have a way to selectively rebuild those. Maybe @SethTisue knows more. Do you intend to squash this before merging? (i.o.w. is all of this still a backport or are there new improvements that should be merged forward?) This would trigger a build of the new commits anyway, so the problem should go away. Otherwise we can just ignore the failures.

lrytz:
> selectively rebuild those

You can open the corresponding "validate-main" build on jenkins, log in, and click "Rebuild"

l0rinc:
Thanks @szeiger, squashed it :)
@lrytz, logged in to @scala-jenkins, now I get `404` and can only access it in incognito mode :)

lrytz:
> @lrytz, logged in to @scala-jenkins, now I get 404 and can only access it in incognito mode :)

uh, that's not very nice :) i knew that jenkins control is only [available to a few](https://github.com/scala/scala-jenkins-infra/blob/master/attributes/default.rb#L102), but causing 404 for others is rather unfriendly.. does it help to access https://scala-ci.typesafe.com/logout ?

adriaanm:
After cherry-picking 9a24860 on 2.11.x (which could be good to include in this PR?), and cherry picking 26c87f1 from #5600 onto my local 2.12.x, the diff with 2.12.x is empty for the three files affected by this PR.

l0rinc:
Added 9a24860 to this commit an rebased it.
Not sure I understood the rest, but I've rebased https://github.com/scala/scala/pull/5600 also.

adriaanm:
Great, thanks! My comment was meant to indicate I did a quick check on this PR, making sure that the backport indeed got these files up to date with 2.12.x.

adriaanm:
For your next backport, I'd suggest using git cherry-pick. For this PR, the command would have been `git cherry-pick -x 9a24860 a501444 7952525 bc91223 b67ca7d 26c87f1` (the -x includes the original commit sha in the commit message for future reference)

l0rinc:
You're right @adriaanm, thank you for your suggestion, force pushed it (excluding the benchmark modifications and https://github.com/scala/scala/pull/5600/commits/635b8afd7071620e38af6642a3ebd01f0fef8485).
I squashed it previously as it kept failing.

adriaanm:
Thanks! Also appreciate your patience with that sneaky genLoadIdent bug that's been plaguing the 2.11.x PR queue :-/ We've spent a lot of time chasing it already. I'm sure @retronym will corner it soon.

adriaanm:
Since these are all backports, I'll add the [ci: last-only] marker so that only the last commit needs to pass the CI.

l0rinc:
@SethTisue, thanks for your patience, this build seems to fail again for some weird reason too: https://scala-ci.typesafe.com/job/scala-2.11.x-validate-publish-core/2978/console

adriaanm:
(The weird compiler crash is tracked as https://github.com/scala/scala-dev/issues/251)

adriaanm:
Rebased in #5626

PR #5567, opened 2016-11-30
Changed HashMap.getOrElseUpdate to only calculate the index once

l0rinc:
Special forward port, as requested by https://github.com/scala/scala/pull/5528#issuecomment-260927187


retronym:
Thanks for remembering! I've amended your commit and added it to #5568. Could you please review it in that context?

l0rinc:
Thanks :)

PR #5570, opened 2016-11-30
SI-10075 propagate annotations to lazy val's underlying field

adriaanm:
This likely regressed in #5294.

Fixes scala/scala-dev#213

Review by @retronym

### DIFF
--- a/test/files/run/t10075.scala
+++ b/test/files/run/t10075.scala
@@ -1,0 +1,35 @@
+class NotSerializable
+
+trait SerializableActually {
+  @transient
+  lazy val notSerializedTLV: NotSerializable = new NotSerializable
+
+  @transient
+  val notSerializedTL: NotSerializable = new NotSerializable
+
+  @transient
+  var notSerializedTR: NotSerializable = new NotSerializable
+}
+
+class SerializableBecauseTransient extends Serializable with SerializableActually {
+  @transient
+  lazy val notSerializedLV: NotSerializable = new NotSerializable
+
+  @transient
+  val notSerializedL: NotSerializable = new NotSerializable
+
+  @transient
+  var notSerializedR: NotSerializable = new NotSerializable
+}
+
+// Indirectly check that the @transient annotation on `notSerialized` made it to the underyling field in bytecode.
+// If it doesn't, `writeObject` will fail to serialize the field `notSerialized`, because `NotSerializable` is not serializable
+object Test {
+  def main(args: Array[String]): Unit = {
+    val obj = new SerializableBecauseTransient
+    // must force, since `null` valued field is serialized regardless of its type
+    val forceTLV = obj.notSerializedTLV
+    val forceLV  = obj.notSerializedLV
+    new java.io.ObjectOutputStream(new java.io.ByteArrayOutputStream) writeObject obj
+  }
+}
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
@@ -494,7 +494,7 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
               genDefDef(statified)
             } else {
               val forwarderDefDef = {
-                val dd1 = global.gen.mkStatic(deriveDefDef(dd)(_ => EmptyTree), traitSuperAccessorName(sym), _.cloneSymbol)
+                val dd1 = global.gen.mkStatic(deriveDefDef(dd)(_ => EmptyTree), traitSuperAccessorName(sym), _.cloneSymbol.withoutAnnotations)
                 dd1.symbol.setFlag(Flags.ARTIFACT).resetFlag(Flags.OVERRIDE)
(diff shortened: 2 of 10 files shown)


retronym:
Not sure we've got this sorted yet. Here's a [test case](https://gist.github.com/retronym/c46af6372d325a8bdd1d8cf1315eccf5) for combinations of trait defined/class defined/trait-mixed in members; val/var/lazy val; field/getter/setter meta annotations that shows the difference between 2.11.8, 2.12.0, and 2.12.1 (as proposed in this PR).

There are still some problems:

 - [X] A `field` targeted annotation on a class member lazy val ends up on both the field and the getter.
 - [x] A `field` targeted annotation on a trait member lazy val is lost (not propagated to the field in the class that mixes in the trait)
 - [x] The static accessor method of a getter in a trait contain the same list of annotations as the method it accesses. The same probably occurs for regular concrete trait methods.

adriaanm:
Thanks for the test case -- where did it go, though?

adriaanm:
I think I have fixes for the first two. Not sure about the static accessors yet

retronym:
https://gist.github.com/retronym/c46af6372d325a8bdd1d8cf1315eccf5

Static accessors are spawned in the backend using https://github.com/scala/scala/blob/d9974be376e23f2a41373cd85977732aee6761a1/src/compiler/scala/tools/nsc/ast/TreeGen.scala#L329-L329

adriaanm:
Thanks. So you're saying we should just drop all annotations from the static accessor?

adriaanm:
I'll incorporate your test tomorrow as well as the logic for the static accessor annotation triage

retronym:
Yes, I think we should drop them.

test/files/run/t10075.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  @transient
+  var notSerializedR: NotSerializable = new NotSerializable
+}
+
+// Indirectly check that the @transient annotation on `notSerialized` made it to the underyling field in bytecode.
+// If it doesn't, `writeObject` will fail to serialize the field `notSerialized`, because `NotSerializable` is not serializable
+object Test {
+  def main(args: Array[String]): Unit =
+    new java.io.ObjectOutputStream(new java.io.ByteArrayOutputStream) writeObject (new SerializableBecauseTransient)
```
durban: Shouldn't the lazy vals be forced before trying to serialize? Otherwise the backing field will be null, and serialization will succeed even if it is not transient, right?

adriaanm: Good point. Thanks!

adriaanm:
One test failure:
```
!!  201 - run/junitForwarders                       [non-zero exit code]
##### Log file '/home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/run/junitForwarders-run.log' from failed test #####

java.lang.AssertionError: assertion failed: found: $init$ - ;foo - @org.junit.Test();foo$ - 
expected: $init$ - ;foo - @org.junit.Test();foo$ - @org.junit.Test()
	at Test$.check(C_1.scala:10)
```

adriaanm:
Updated test, since it had a nice TODO message that confirmed it needed updating :-) thanks, @lrytz

dwijnand:
It's green now

retronym:
LGTM, tricky stuff!

adriaanm:
If I understand @lrytz's comment at https://github.com/scala/scala-dev/issues/213#issuecomment-264645365 correctly, I should drop the `withoutAnnotations` for static accessors. I think we should either drop all annotations or we keep them all. If we have to keep `@inline`, someone may be able to make a similar case for other annotations.

Alternatively, could we avoid looking at annotations of static accessors in the inliner? Could the inliner see through the static accessor and decide inlining based on the target of the accessor? `inlineRequest` is already special cased on `isTraitSuperAccessor` to use `findSingleCall`, but I didn't look deeper than that.

lrytz:
It turns out the inliner does not rely on the `@inline` annotation on static super accessors.

For every member method, we store some metadata in the `ClassBType` (The `MethodInlineInfo`, it contains is information that can only be obtained from a method symbol, but, without unpickling, not from the classfile signature. Example: is the method annotated `@inline`. This information is added to the class as a custom attribute and used in case the inliner pulls in a new classfile.)

When building a `ClassBType`, the member method symbols are obtained from the `classSym.info.decls`. However, the static accessors are not entered into the class scope, so they don't show up. That's why there's a special case that creates additional `MethodInlineInfo`s for static accessors (https://github.com/scala/scala/blob/ee1c02b374a4b8a053e9a8b14af5e205afa67e14/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala#L598-L608). The `annotatedInline` bit is taken from the corresponding member method.

adriaanm:
Aha! Thanks for explaining. So this is ready to go as far as you're concerned?

lrytz:
yes, LGTM

adriaanm:
Cool. When I split up my old commit in two, I missed one test case. The last commit is green, so I'm going to merge. One broken test isn't going to break future git bisecting.

PR #5573, opened 2016-12-02
Remove existential infer hack enabled by #5376

adriaanm:
It wasn't a good idea after all.

Also removed some tracing code that I cannot imagine
was ever used in a production compiler. It's still just
a recompile away.

Fixes scala/scala-dev#280

/cc @milessabin


adriaanm:
Review by @retronym

milessabin:
I disagree here. @retronym, can you elaborate on what you meant by "too much breakage" ... as I said, I ran a successful community build with the inference improvements enabled.

Also, with @retronym's new example I can't see any difference in behaviour in 2.12.x with the inference improvements enabled or disabled.

I'd like to see at least one test which illustrates any problem there might be.

retronym:
I was referring to a known bug in static overload resolution. As you were diagnosing a change in that space,  I thought I'd mention it, but I didn't mean to imply it was related to the experimental feature being demoted here. 

I agree a test case extracted from the community build failure would help here, at least to defend against regression and hopefully to guide a fix to the implementation. I don't think this ought to block 2.12.1, though, we could spin out a ticket.

milessabin:
@retronym OK, but @adriaanm seems to have interpreted this as you saying that the change being made in #5376 is problematic and/or too risky, and I don't see that at all.

milessabin:
Oh, and there was _no_ community build failure!

milessabin:
I've reproduced the issue with scalameter built with 2.12.0 and `-Xexperimental -Ypartial-unification:false`, so I agree there's an issue here.

However, this raises the problem of individual disablement of features under `-Xexperimental` again. If this feature is too risky to make default then it should be possible to individually disable it via a `-Y` flag, just as I did for `-Ypartial-unification` above. Otherwise I won't be able to recommend that people use `-Xexperimental` in preference to individually enabling features.

adriaanm:
This PR is just about removing what turns out to be a bad experimental type inference feature. It was never enabled by default in a release (just briefly by #5376), and if we bring it back, the community build will now fail (since it affects scalameter, which is included, although it may not have been part of it when we tried the original PR #5376).

milessabin:
Understood. However, can you address the part of my comment relating to individual disablement?

adriaanm:
Could you clarify how it applies to this PR? All this PR does is remove my hack *entirely* for repacking existential during type inference.

milessabin:
Oh! I read "Revert existential infer part of #5376" as meaning that you'd reverted the commit in #5376 which removed this change from under `-Xexperimental`.

retronym:
LGTM

PR #5575, opened 2016-12-02
SI-10086 NumericRange.min|max with custom Integral

gzm0:


som-snytt:
I didn't see this. You can copy the unit test. It passed locally.

https://github.com/scala/scala/compare/2.12.x...som-snytt:issue/10086-max-order?expand=1

som-snytt:
It's not obvious to me that the following `min` should be slow-path. If the ordering passed to min/max is the constructor Integral, `ord eq num`, can it just return start/last. (I noticed Numeric was not always an Ordering.)

```
scala> import scala.collection.immutable.NumericRange, math._
import scala.collection.immutable.NumericRange
import math._

scala> val sz = 1000000000
sz: Int = 1000000000

scala> implicit val customIntegral = new Numeric.IntIsIntegral with Ordering.IntOrdering { }
customIntegral: scala.math.Numeric.IntIsIntegral with scala.math.Ordering.IntOrdering = $anon$1@127a7272

scala> val nr = NumericRange(1, sz, 1)
nr: collection.immutable.NumericRange.Exclusive[Int] = NumericRange 1 until 1000000000

scala> nr.min
res0: Int = 1

```

gzm0:
This is ready for review.

som-snytt:
Probably review by @Ichoran "Ichorange".

test/junit/scala/collection/immutable/RangeConsistencyTest.scala
```diff
@@ -172,4 +172,43 @@ class RangeConsistencyTest {
...
+    assert(nr.max == 9)
+
+    // Also test with custom ordering.
+    assert(nr.min(customIntegral.reverse) == 9)
+    assert(nr.max(customIntegral.reverse) == 1)
+
+    case class A(v: Int)
+
+    implicit object aIsIntegral extends scala.math.Integral[A] {
+      def compare(x: A, y: A): Int = x.v - y.v
```
Ichoran: This is inconsistent.  Orderings must be transitive with respect to the sign of `compare` but this is not true with `A(Int.MinValue)`, `A(0)`, and `A(Int.MaxValue)`.  Though extreme values are not used in this test, it is nonobvious that this method fails, and there is no advantage to using subtraction rather than the `compare` method on the underlying Ints.

Ichoran:
It would be good to have something--at least a commit message--that explains why `(ord eq num)` could be true.  It's not utterly obvious.

gzm0:
Updated. Please let me know when you would like me to squash the commits.

lrytz:
LGTM, squashing on merge.

PR #5576, opened 2016-12-03
add util.Random.nextLong(n)

sbromberger:
Added `nextLong(n)` based on http://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator

### DIFF
--- a/src/library/scala/util/Random.scala
+++ b/src/library/scala/util/Random.scala
@@ -69,6 +69,16 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    *  from this random number generator's sequence.
    */
   def nextLong(): Long = self.nextLong()
+  
+  def nextLong(n: Long): Long = {
+    if (n <= 0) throw new java.lang.IllegalArgumentException("bound must be positive")
+    if (n <= Int.MaxValue) nextInt(n.toInt).toLong
+    else {
+      val RAND_MAX = Long.MaxValue     // 2^63-1
+      val x = nextLong() >>> 1         // [0, 2^63), strip the sign bit
+      if (x < (RAND_MAX - (RAND_MAX % n))) x % n else nextLong(n)
+    }
+  }
 
   /** Returns a pseudorandomly generated String.  This routine does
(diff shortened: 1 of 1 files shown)


sbromberger:
> Thank you for signing the Scala CLA.
> If you have already sent a pull request please add a comment to the pull request saying you did sign the CLA.

(I signed the CLA.)

sbromberger:
PS: this probably benefits from tests, but I've never written scala tests before so I'm not quite sure how to proceed.

sjrd:
Pointing out the implementation we have in Scala.js: https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala#L48

Not sure what the trade-offs are. One sure thing is that our implementation will be immensely faster in JS, because it avoids division and modulo on `Long`s, which are extremely expensive operations.

sbromberger:
@sjrd that's an interesting algorithm, but it runs the risk of entropy depletion (you're hitting the pool up to 31 times, where the proposed algorithm is nondeterministic but is at most 2-3 times in real-world use). 

As to performance, I'd love to see benchmarks against it, since the div/mod happens at most once in the proposed code.

Anyway - I don't know why the checks are failing, and I'm afraid I'm not well-versed enough in this to figure out how to test it locally before submission/correction. Advice appreciated.

som-snytt:
As someone mentioned on gitter, MiMa errors because you're adding public API, which would have to happen on 2.13.x.

Click the Details link next to validate-test comment, then console output and the little full log link. MiMa runs after the tests.

SethTisue:
GitHub lets you retarget an existing PR to a different branch (in this case, you want `2.13.x`), but you have to do the rebasing part yourself. (An alternative is just to close this PR and make a new one.)

sbromberger:
I'll close this and retry. Thanks for the advice.

sbromberger:
See #5577 - still getting MiMa failures. Not sure what to do.

PR #5577, opened 2016-12-04
nextLong(n)

sbromberger:
Ref #5576 (take 2)

### DIFF
--- a/src/library/scala/util/Random.scala
+++ b/src/library/scala/util/Random.scala
@@ -13,6 +13,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.collection.generic.CanBuildFrom
 import scala.collection.immutable.{ List, Stream }
 import scala.language.{implicitConversions, higherKinds}
+import scala.annotation.tailrec
 
 /**
  *  @author Stephane Micheloud
@@ -69,6 +70,24 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    *  from this random number generator's sequence.
    */
   def nextLong(): Long = self.nextLong()
+
+  /** Returns the next pseudorandom, uniformly distributed long value
+   *  between 0 (inclusive) and the specified value (exclusive), drawn
+   *  from this random number generator's sequence.
+   */
+  def nextLong(n: Long): Long = {
+    if (n <= 0) throw new java.lang.IllegalArgumentException("bound must be positive")
+    if (n <= Int.MaxValue) nextInt(n.toInt).toLong
+    else {
+      val RandMax = Long.MaxValue
+      val bound = RandMax - (RandMax % n)
+      @inline @tailrec def genValid(): Long = {
+        val x = nextLong() >>> 1
+        if (x < bound) x else genValid()
+      }
+      genValid() % n
+    }
+  }
 
   /** Returns a pseudorandomly generated String.  This routine does
    *  not take any measures to preserve the randomness of the distribution
--- a/test/junit/scala/util/RandomTest.scala
+++ b/test/junit/scala/util/RandomTest.scala
@@ -12,4 +12,18 @@ class RandomTest {
       Assert.assertTrue(s"$c should be alphanumeric", isAlphaNum(c))
     }
   }
+
+  @Test def testNextLongN: Unit = {
+    val testsPerN = 1000
+    val nCats = 100
+    def testOneN(n:Long): Unit = {
+      for (_ <- 0.until(testsPerN)) {
+        val rN = Random.nextLong(n)
+        val inRange = (0 <= rN) && (rN < n)
+        Assert.assertTrue(s"$rN should be within [0, $n)", inRange)
+      }
+    }
+
+    for (_ <- 0.until(nCats)) testOneN(Math.abs(Random.nextLong()))
+  }
 }


sbromberger:
Still getting MiMa failures:
> [error]   - library/mima failed: java.lang.RuntimeException: MiMa failed with exit code 13
java.lang.RuntimeException

SethTisue:
recent 2.13 nightly builds have also been failing with MiMa errors. I'll look into it this week unless somebody beats me to it

src/library/scala/util/Random.scala
```diff
@@ -70,6 +70,16 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    */
   def nextLong(): Long = self.nextLong()
 
+  def nextLong(n: Long): Long = {
+    if (n <= 0) throw new java.lang.IllegalArgumentException("bound must be positive")
+    if (n <= Int.MaxValue) nextInt(n.toInt).toLong
+    else {
+      val RAND_MAX = Long.MaxValue     // 2^63-1
+      val x = nextLong() >>> 1         // [0, 2^63), strip the sign bit
+      if (x < (RAND_MAX - (RAND_MAX % n))) x % n else nextLong(n)
```
sjrd: The apparently tail-recursive call here is not actually tail-recursive, because `nextLong` is neither final nor private. Moreover, it will re-do the first two checks every time for no reason. Consider an explicit loop (which also avoids redoing `RAND_MAX % n` on every iteration:
```scala
else {
  val RandMax = Long.MaxValue
  val bound = RandMax - (RandMax % n)
  var x = nextLong() >>> 1
  while (x >= bound)
    x = nextLong() >>> 1
  x % n
}
```

sbromberger: I'm trying to make this purely functional (that is, no `var` statements). Let me see if there's an alternate way of doing it.

Also, how inefficient is two compares like that? We really should only be going back through this function a maximum of two times. That's four Long comparisons plus a possibly less efficient recursion.

sjrd: Here is a purely functional version with tail-recursion:
```scala
else {
  val RandMax = Long.MaxValue
  val bound = RandMax - (RandMax % n)
  @inline @tailrec def genValid(): Long = {
    val x = nextLong() >>> 1
    if (x < bound) x else genValid()
  }
  genValid() % n
}
```
The comparisons are not inefficient. Modulo operations are. Even on the JVM, they are relatively very slow compared to other arithmetic operations. On JS they are *terribly* slow, especially when the left-hand-side is > 2^53, which is true for the first `%` and virtually always true for the second `%`.

sbromberger: I like that proposed modification. As to efficiency of div/mod, I'm afraid I don't know a thing about scala.js architecture. I wrote this for my general scala use. Since scala.js already has a method for this, would it still be ok to pursue this for inclusion into the scala base itself?

Again, we're doing, realistically, *at most* 3 modulos, and I'd say that for the vast vast majority of cases, we're doing 1. It would be interesting to benchmark in scala.js.

sjrd: > for the vast vast majority of cases, we're doing 1

You're doing at least 2 modulos: `RandMax % n` and `x % n`. With my suggestion there are always *exactly* 2 modulos.

> Since scala.js already has a method for this

It has one, but it's not the same method. People writing cross-compiling libraries will want to use the method you're adding (and not switch between this one and the other one depending on the target platform -- that would be a nightmare). IMO, if a change significantly improves Scala.js without affecting the JVM side, it should be applied at the source, in scala/scala. There are a few precedents for this rationale.

> As to efficiency of div/mod, I'm afraid I don't know a thing about scala.js architecture.

`Long`s are implemented in user-space in Scala.js. Most operations are fast (within 3x of their `Int` equivalents) but `/` and `%` can be as bad as 30x slower than addition, taking as long as 1 millisecond per `/` or `%`. Each `/` or `%` that is avoided is a noticeable improvements even at the scale of a larger algorithm. For example, in the implementation of `BigInteger.toString()`, which needs to do both `a / b` and `a % b` in the inner loop, deriving `a % b` as `(a - (a_div_b)*b)` (trading a `%` for one `*` and one `-`) improved the performance of the entire `toString()` by almost 2x, showing that `toString()` basically spends all its time in the division and modulo.

On the JVM, I have also measured division to be around 40x slower than addition on `Long`s, although I trust these numbers less because this a very narrow micro-benchmark, and that is tricky to get right on the JVM.

sbromberger: Interesting about the div/mod - thank you for the explanation.

I'm wondering whether it's worth even pursuing this PR, given your statement about this not working well with scala.js, which is, I guess, a soft requirement for any scala code. The main issue that I have with the scala.js code is that it doesn't look like any code I've seen that is designed to provide a uniform random distribution. That's not to say it *doesn't*, but the code I've proposed is based off / inspired by Schneier and other sources I trust. For me, it's more important to have a uniform distribution with this function than to eke out every single optimization. Perhaps that's not the most important thing for others, which is why I'm now questioning the appropriateness of the PR.

Let me know - I'm happy to close this out and just use the code locally.

sjrd: IMO the contribution is valuable! I was just trying to nudge you into making it as efficient as possible on Scala.js given the algorithm. :-)

sbromberger: Thanks :)

I've made your suggested changes to the function. Does this represent the "best for scala.js" given the known issues with div/mod? (Does it pass muster?)

I'd still love to benchmark this in both scala.js and scala but I'm not quite sure how to do the former.

sjrd: Yep, that's the best of Scala.js ;)

Benchmarking in Scala.js is not trivial. You can have a look at https://github.com/sjrd/scalajs-benchmarks for inspiration. The easiest for quick tests is to modify an existing benchmark (like Richards, which is pretty self-contained) and replace it with what you want to test instead. Then run `richardsJS/run`.

sbromberger: Heh, ok. I'll try to find some time to learn about it more. I don't use scala.js so it'll be a series of new experiences.

Anyway - if you're comfortable with this code, awesome. I don't know why it's failing an additional test (I know @SethTisue is looking into the MiMa failures) but hopefully it's not related to any code breakage with the new commit.

Edit: ahh, it's `@tailrec`:

```
[error] /home/jenkins/workspace/scala-2.13.x-validate-publish-core/src/library/scala/util/Random.scala:79: not found: type tailrec
[error]       @inline @tailrec def genValid(): Long = {
[error]                ^
[error] one error found
```

sbromberger:
@sjrd I incorporated your changes and squashed the commit. Thanks for all your feedback.

sjrd:
:+1: I'll let a maintainer of scala/scala take it from here.

SethTisue:
after https://github.com/scala/scala/pull/5578 is merged and you rebase on top of it, the MiMa error will go away

sbromberger:
Thanks, @SethTisue - I'll wait for that. In the meantime, could you point me to where my tests should go? I've found `test/junit/scala/util/RandomTest.scala` but it doesn't seem to be a comprehensive test suite for `util.Random`. How should I proceed?

SethTisue:
there is a lot of stuff in the stdlib that doesn't have proper testing, sadly, often because it's been there a decade-plus. we try to hold new additions (like this) to a higher standard

you can either add tests to RandomTest.scala, or if you're into property-based testing, add something to test/files/scalacheck . (for historical reasons, ScalaCheck based tests are run via the intermediary of partest, rather than being run by sbt directly. I've opened https://github.com/scala/scala-dev/issues/282 on that)

l0rinc:
I think we could do this without division or multiplication with something like:
```scala
def nextLong(limit: Long): Long = {
  val mask = -1 >>> java.lang.Long.numberOfLeadingZeros(limit)
  var result = 0L
  do {
    result = nextLong() & mask
  } while (result >= limit)
  result
}
```
I have only checked that apparently this results in uniform distribution 
(i.e
```scala
val v = Iterator.continually{ rnd.nextLong(10) }.take(100000).toVector.groupBy(x => x).mapValues(_.length)
```
returns
```scala
Map(0 -> 10078, 5 -> 9884, 1 -> 9944, 6 -> 10010, 9 -> 10032, 2 -> 10017, 7 -> 9984, 3 -> 10067, 8 -> 9999, 4 -> 9985)
```
, https://www.wolframalpha.com/input/?i=number+line+10078+9944+10017+10067+9985+9884+10010+9984+9999+10032#Statistics
) and that `numberOfLeadingZeros` is intrinsically compiled to `LZCNT` assembly.
Not sure whether it's correct or not, it's just a thought that I will test thoroughly tomorrow (`JMH` and `ScalaCheck`), comparing against `nextInt`'s performance also.

sbromberger:
There are lots of other ways to do this. However, closely approxmiating a uniform distribution is the tricky part and an effort that should not be undertaken casually. This is why, IMO, it's better to rely on existing vetted code to provide uniform distribution rather than try to come up with something novel that hasn't had the exposure. 

That's not to say it *can't* be done, just that when the feature is important, it's probably best to stand on the shoulders of giants. That's why I prefer the current implementation.

(As an example: I attempted to shortcut the `val x = nextLong() >>> 1` line using `Math.abs`. This introduced a very subtle "half-zero" sampling bug. Randomness is difficult.)

sbromberger:
OK. I've been doing some digging here and there are a few problems with random numbers based on the Java `Random` class.

It turns out that Java's `nextLong()` function [is not truly uniform](http://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextLong--). This is bad for a huge number of reasons and really calls into question whether Scala should be using it.

There are alternatives (specifically the upcoming [Apache Commons RNG](http://commons.apache.org/proper/commons-rng/)) that appear to be better for uniform randomness and provide the `nextLong(n)` functionality I need.

As a result, I'm going to close this out. Even if it's convenient, it's not something that I'd recommend anyone use. (In fact, I think all of the `scala.util.Random` methods should be deprecated in favor of their more secure alternatives, but I understand that this may not be a popular opinion.)

Thanks to everyone for their suggestions and comments. This was a great experience.

retronym:
I'll also point out the java standard library's `SecureRandom` (which uses entropy sources from the hardware rather than a pseudorandom sequence.)

The requirements for random generators are indeed quite dependent on the application, so it makes sense to defer to specialist libraries like Commons RNG.

Ichoran:
@sbromberger - There are a variety of incredibly fast highly random non-cryptographic RNGs.  The Pcg family from Melissa O'Niell is my favorite overall--really fast, really high quality, mathematically interesting, and with variants specialized for a wide variety of uses.  Sebastiano Vigna's ShiftMix is a bit faster with equivalent quality.

I have Scala implementations of these in KSE: https://github.com/Ichoran/kse/blob/master/src/main/scala/maths/Stochastic.scala

I haven't tuned them for performance, but in casual benchmarks they perform as well as I hoped (Pcg64 is 6-12x faster than scala.util.Random; ShiftMix64 is at best 50% faster yet (so 18x faster than scala.util.Random for nextLong).

If someone needs random numbers, feel free to use those.

Ichoran:
@retronym - Actually, I disagree that one should _automatically_ defer to specialist libraries.  Almost every non-cryptographic need for pseudorandom numbers can be met by, say, the RXS M XS 64 variant of Pcg64 that I implemented (or a similar one with a bit more state and longer period).  You only need to defer to specialist libraries if your base library is lousy (or hasn't been rethought since Numerical Recipes in C version _two_ came out) or if you have specialized needs for unusual distributions or cryptographic properties.

PR #5579, opened 2016-12-05
Improve Queue.++ when building another Queue

edmundnoble:
Adds in the Okasaki `concat` version in case we are constructing a `Queue` from two `Queue`s. You can see the impact of the default implementation (and inheritance hierarchy etc.) here: https://ethercalc.org/kknm1xz3mb5s_naakfyfum4p3/app in all of the concat benchmarks, especially the right-associated tree being concatenated. The performance difference is a few orders of magnitude at my benchmark size. Let me know if this is more appropriate for 2.11.x.

### DIFF
--- a/src/library/scala/collection/immutable/Queue.scala
+++ b/src/library/scala/collection/immutable/Queue.scala
@@ -100,6 +100,15 @@ class Queue[+A] protected(protected val in: List[A], protected val out: List[A])
     case _                               => super.:+(elem)(bf)
   }
 
+  override def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Queue[A], B, That]): That = {
+    if (bf eq Queue.ReusableCBF) {
+      val thatQueue = that.asInstanceOf[Queue[B]]
+      new Queue[B](thatQueue.in ++ (thatQueue.out reverse_::: this.in), this.out).asInstanceOf[That]
+    } else {
+      super.++(that)(bf)
+    }
+  }
+
   /** Creates a new queue with element added at the end
    *  of the old queue.
    *
--- a/test/files/run/iq.scala
+++ b/test/files/run/iq.scala
@@ -25,12 +25,18 @@ object iq {
     assert(q2 == qb)
     val qc = 42 +: q :+ 0
     assert(q2 == qc)
+    assert(q ++ qa == qa)
+    val qdr =  1 +: 2 +: 3 +: 4 +: q
+    val qcon1 = 1 +: 2 +: q
+    val qcon2 = q :+ 3 :+ 4
+    val qd = qcon1 ++ qcon2
+    assert(qd == qdr)
 
     Console.println("q2: " + q2)
     Console.println("qa: " + qa)
     Console.println("qb: " + qb)
     Console.println("qc: " + qc)
-    
+
     /* Test is empty and dequeue.
      * Expected: Head: 42
      */


src/library/scala/collection/immutable/Queue.scala
```diff
@@ -100,6 +100,15 @@ class Queue[+A] protected(protected val in: List[A], protected val out: List[A])
     case _                               => super.:+(elem)(bf)
   }
 
+  override def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Queue[A], B, That]): That = {
+    if (bf eq Queue.ReusableCBF) {
```
szeiger, line 104: This looks wrong. The `ReusableCBFInstance` is shared by all GenTraversables. There's no guarantee that either `that` or `That` is of type `Queue`.

edmundnoble: The assumption that `That =:= Self` seems to be made in List.scala:
```scala
  override def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[List[A], B, That]): That =
    if (bf eq List.ReusableCBF) (this ::: that.seq.toList).asInstanceOf[That]
    else super.++(that)

  override def +:[B >: A, That](elem: B)(implicit bf: CanBuildFrom[List[A], B, That]): That = bf match {
    case _: List.GenericCanBuildFrom[_] => (elem :: this).asInstanceOf[That]
    case _ => super.+:(elem)(bf)
  }
```

Ichoran: `.asInstanceOf[Queue[B]]` is absolutely not a parallel to `.seq.toList`.

edmundnoble: @Ichoran When is it that `That =:= Self; val that: That = _; that.isInstanceOf[Self]` is not the case? Please provide an example.

retronym: `ReusableCBFInstance` is a field in `class GenTraversableFactory`. The companions of each collection extend this, and thus have distinct instances.

```
scala> Queue.ReusableCBF eq Seq.ReusableCBF
res8: Boolean = false
```

So this code isn't obviously incorrect to me.

szeiger: Hm, you're right, `ReusableCBFInstance` is a field of the class. I must have read that wrong the first time and assumed it's in the object.

src/library/scala/collection/immutable/Queue.scala
```diff
@@ -100,6 +100,15 @@ class Queue[+A] protected(protected val in: List[A], protected val out: List[A])
...
+      val thatQueue = that.asInstanceOf[Queue[B]]
+      new Queue[B](thatQueue.in ++ (thatQueue.out.reverse ++ this.in), this.out).asInstanceOf[That]
```
SethTisue: `thatQueue.out.reverse ++ this.in` copies `thatQueue.out` twice — I think you can get it down to one copy by using `List#reverse_:::`, a trick you'll see used in the body of `def enqueue[B >: A](iter: Iterable[B])`

SethTisue: `thatQueue.out.reverse ++ this.in` copies `thatQueue.out` twice — I think you can get it down to one copy by using `List#reverse_:::`, a trick you'll see used in the body of `def enqueue[B >: A](iter: Iterable[B])`

SethTisue (changes requested):

SethTisue (approved):

xuwei-k:
https://github.com/scala/bug/issues/10298

src/library/scala/collection/immutable/Queue.scala
Ichoran: @edmundnoble - Ugh, I'm sorry I overlooked this after the first comment.  As I guess we're all aware now, it's the `that: That` assumption that is wrong.  `That` is the type being produced, while `that` is an instance that is not parameterized by type.  It's a very unfortunate naming convention.

edmundnoble: Ughhhhh. Crap. My bad.

PR #5581, opened 2016-12-05
2.12 -> 2.13 in the most important places

SethTisue:
there are a lot of other places that could be updated, especially in
comments/docs, but the sooner we update them all, the more annoying
merge conflicts we'll have to suffer through. so at the moment let's
just hit the main ones

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,7 +87,7 @@ lazy val publishSettings : Seq[Setting[_]] = Seq(
 // should not be set directly. It is the same as the Maven version and derived automatically from `baseVersion` and
 // `baseVersionSuffix`.
 globalVersionSettings
-baseVersion in Global := "2.12.1"
+baseVersion in Global := "2.13.0"
 baseVersionSuffix in Global := "SNAPSHOT"
 
 // to be locked down sometime around the time of 2.13.0-RC1
--- a/project/VersionUtil.scala
+++ b/project/VersionUtil.scala
@@ -136,7 +136,10 @@ object VersionUtil {
   /** Build a dependency to a Scala module with the given group and artifact ID */
   def scalaDep(group: String, artifact: String, versionProp: String = null, scope: String = null, compatibility: String = "binary") = {
     val vp = if(versionProp eq null) artifact else versionProp
-    val m = group % (artifact + "_" + versionProps(s"scala.$compatibility.version")) % versionNumber(vp)
+    // TODO for 2.13.0-M1 -- eventually it will no longer work to use 2.12 modules,
+    // but at the moment, there aren't any 2.13.0-M1 modules to use yet, so hardcode 2.12 for now.
+    // val m = group % (artifact + "_" + versionProps(s"scala.$compatibility.version")) % versionNumber(vp)
+    val m = group % (artifact + "_2.12") % versionNumber(vp)
     val m2 = if(scope eq null) m else m % scope
     // exclusion of the scala-library transitive dependency avoids eviction warnings during `update`:
     m2.exclude("org.scala-lang", "*")
--- a/versions.properties
+++ b/versions.properties
@@ -1,7 +1,7 @@
 # Scala version used for bootstrapping. (This has no impact on the
 # final classfiles, since compiler and library are built first using
 # starr, then rebuilt using themselves.)
-starr.version=2.12.0
+starr.version=2.12.1
 
 # These are the versions of the modules that go with this release.
 # These properties are used during PR validation and in dbuild builds.
@@ -12,7 +12,7 @@ starr.version=2.12.0
 #  - After 2.x.0 is released, the binary version is 2.x.
 #  - During milestones and RCs, modules are cross-built against the full version.
 #    So the value is the full version (e.g. 2.12.0-M2).
-scala.binary.version=2.12
+scala.binary.version=2.13.0-M1
 
 # external modules shipped with distribution, as specified by scala-library-all's pom
 scala-xml.version.number=1.0.6


SethTisue:
the other reason I didn't update every conceivable location "2.12" appears is that I'm interested in merging this on the fast track, in the interest of avoiding some bizarre accident where a 2.13.x Jenkins job wipes out 2.12.1 stuff somehow

SethTisue:
I made a 2.13.x branch in scala-dist-smoketest and merged changes in the scala-dist repo, and the 2.13 community build is green, so with a little luck, this might actually result in our first green 2.13 release-main job

SethTisue:
hmm you know what though, I bet this will fail because of modules(/ScalaCheck). I might have to override it to use 2.12 modules

SethTisue:
modules overridden to 2.12.

but now: `java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Java heap space` reproducible locally with `;setupPublishCore ;dist/mkPack`. sigh

retronym:
We could go back to embedding scalacheck for now on the 2.13 branch.

SethTisue:
We'll have to at some point, but I think we can get away without that for now.

The `OutOfMemoryError`, I think it's trying to construct an OSGi version string, or maybe trying to construct an error message describing a failure to construct an OSGi version string, some gross thing. I need to set this aside for now

SethTisue:
for now I've submitted just the spec-generation part of this as https://github.com/scala/scala/pull/5582, since that's the only part where I'm aware of an actual concrete danger of 2.13 stuff overwriting 2.12 stuff.

SethTisue:
ha. same failure at 5582, so probably the cause is actually some other recently merged PR? maybe a 2.12->2.13 merge PR? maybe the one we `/nothingtoseehere`d?

SethTisue:
rebased

lrytz:
got a heap dump, which shows some strings of 146, one with 184 MB :)

Looks like this over and over:
```
Warnings
000: No translation found for macro: range;[==,=+);2.13.0-M1.v20161206-112152-30ef5ad, in -----------------
Errors
000: Invalid syntax for version: 2.13.0-M1.v20161206-112152-30ef5ad, for cmd: range, arguments; [range, [==,=+), 2.13.0-M1.v20161206-112152-30ef5ad]
```

cc @dwijnand - maybe you already know where sbt's Scala version parser needs to be updated?

dwijnand:
I've never seen that warning and error (I actually don't really understand what's going on in general in this PR). So I'm not sure what Scala version parser you're looking for.

However one that I do know about is https://github.com/sbt/sbt/blob/v0.13.13/util/cross/src/main/input_sources/CrossVersionUtil.scala . I hope that helps.

lrytz:
OK, thank you. Then the warnings are generated by an osgi tool.

lrytz:
someone else: https://github.com/bndtools/bnd/issues/724. so there's a problem with those warnings piling up, seems to be fixed (https://github.com/bndtools/bnd/pull/818, not sure in what release, but we're on 2.4.1).

however, something is wrong on our side, these warnings shouldn't be generated. i think during our 2.12 milestones we were still on ant. @szeiger maybe you know have an idea why this triggers a warning:

```
val reflect = configureAsSubproject(project)
  .settings(Osgi.settings)  // headers := {Seq(..., ""Import-Package" -> "scala.*;version=\"${range;[==,=+);${ver}}\",*", ...)
  .settings(
...
    Osgi.headers +=
      "Import-Package" -> ("scala.*;version=\"${range;[==,=+);${ver}}\","+
                           "scala.tools.nsc;resolution:=optional;version=\"${range;[==,=+);${ver}}\","+
```
(https://github.com/scala/scala/blob/2.13.x/build.sbt#L377-L378) for osgi `ver` `2.13.0-M1.v20161206-112152-30ef5ad`?

szeiger:
Milestone and RC version numbers should be pretty much the same, so we must have gone through this already. I'll have to take a closer look.

build.sbt
```diff
@@ -87,7 +87,7 @@ lazy val publishSettings : Seq[Setting[_]] = Seq(
 // should not be set directly. It is the same as the Maven version and derived automatically from `baseVersion` and
 // `baseVersionSuffix`.
 globalVersionSettings
-baseVersion in Global := "2.12.1"
+baseVersion in Global := "2.13.0-M1"
```
szeiger: Here's the problem. `baseVersion` needs to be `2.13.0`. `M` or `RC` suffixes are not part of SHA-versioned snapshots. An actual M1 release will get `M1` as the `baseVersionSuffix`. See https://github.com/scala/scala/blob/2.12.x/project/VersionUtil.scala#L49-L63 for the allowed combinations of values.

SethTisue:
thanks Stefan. revised version pushed

lrytz:
thanks! we could sanity-check some of these settings / values in our build..

lrytz:
(or simply a comment at the declaration)

dwijnand:
Yeah imho version shouldn't be a String. I'm not sure if it was a design goal of sbt to prefer primitive types.

Also, it's sadly not possible to customise the type of keys in sbt (eg in scala/scala redefine type Version to instead of be the default tyoe SemVer, to be EpicMajorMinorVer)

dwijnand:
Oh it's your own `baseVersion`, so you could make it a more restrictive type than plain String.

SethTisue:
```
% sbt       
> set resolvers += "nightlies" at "https://scala-ci.typesafe.com/artifactory/scala-release-temp/"
> set scalaVersion := "2.13.0-87c75a7-nightly"
> console
Welcome to Scala 2.13.0-87c75a7-nightly (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> 2 + 2
res0: Int = 4
```

🎉

PR #5582, opened 2016-12-05
in spec generation, 2.12 -> 2.13 so we don't collide

SethTisue:


SethTisue:
this is probably not sufficient to prevent the 2.13 spec from overwriting the 2.12 one, I think we will have to deal with the ssh forced-command stuff from https://github.com/scala/scala-lang/issues/479

SethTisue:
some server work is probably needed too. ticket on that is https://github.com/scala/scala-lang/issues/567

review by @adriaanm

adriaanm:
Server work done:
```
ssh scalatest@chara.epfl.ch -i ~/.ssh/for_chara
cd ~/.ssh/
cat authorized_keys2 | tail -n -2 | perl -pe 's/2.12/2.13/g' >> authorized_keys2 # or some manual approximation of that :-)
```

adriaanm:
LGTM

SethTisue:
fixed this up to only include the changes I had intended.

merging, since I'm still worried the 2.13 spec will overwrite the 2.12 spec on the server. Adriaan took care of the ssh forced-command bit already

SethTisue:
as feared, the 2.12 spec was overwritten by the 2.13 one (reported by @som-snytt)

som-snytt:
Sometimes it's nice to mix it up a little. That would be "Som" times.

SethTisue:
don't worry, be A.P.

PR #5583, opened 2016-12-06
SI-10093 don't move member traits to constructor body in constructors

lrytz:
Fixes a regression introduced in c8e6050. Member traits with only
abstract definitions (`isInterface`) were moved into the primary
constructor by mistake. (Flatten moved the classes back.)

The member trait was duplicated into the constructor of specialized
subclasses, causing it to be generated multiple times.

Also removes some unnecessary `isMixinConstructor` checks: the mixin
constructor is always the primary constructor.

This commit also clarifies (and tests) what `isInterface` means: for
scala-defined traits, it means there are only abstract members. For
java-defined interfaces, it is always true.


adriaanm:
LGTM!

PR #5587, opened 2016-12-07
SI-10072 cast Function nodes to environment in specialization

lrytz:
This commit basically make sure the fix for SI-5284 works correctly when
a Function node reaches specialization (`-Ydealmbdafy:method` and
IndyLambda are default in 2.12.x). To understand it, best read the
excellent description in b29c01b.

The code that's removed in this commit was added in 13ea590. It
prevented `castType` from being invoked on the `Function` node, which
is exactly what is needed here. It's also what happens under
`-Ydelambdafy:inline`, the `new anonfun()` tree is being casted from
`(Int, Int) => Int` to `(Int, A) => Int`.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Duplicators.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Duplicators.scala
@@ -240,10 +240,6 @@ abstract class Duplicators extends Analyzer {
             result.symbol.updateAttachment(DelambdafyTarget)
           result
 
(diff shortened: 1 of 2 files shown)


lrytz:
@dragos, could you maybe take a look? I'm really quite unfamiliar with this part of the compiler.

adriaanm:
This is not my specialty 😏, but LGTM!

PR #5588, opened 2016-12-07
SI-10098 Fix regression in Unix runner script with JAVA_HOME unset

retronym:
Rework bfa7ade0 to unconditionally set the system property with the
contents of the bootclasspath, rather than trying to do this only
for JVM 9+.

The attempted JVM version detection code assumed JAVA_HOME was set,
which isn't always the case.

### DIFF
--- a/src/compiler/scala/tools/ant/templates/tool-unix.tmpl
+++ b/src/compiler/scala/tools/ant/templates/tool-unix.tmpl
@@ -197,14 +197,11 @@ fi
 # to java to suppress "." from materializing.
 if [[ "$usebootcp" == "true" ]]; then
   classpath_args=("-Xbootclasspath/a:$TOOL_CLASSPATH" -classpath "\"\"")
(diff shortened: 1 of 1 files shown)


paulp:
May I suggest the bash scripts should have `set -euo pipefail` (or something like it) at the top. Maybe it's innocuous in this instance that it will run commands like `cat /release` when an expected variable is unset, but it does not provide much confidence about what it will do in other scenarios.

Attempting to use an uninitialized variable in a shell command ought to be a fatal error.

lrytz (approved):
@paulp's suggestion sounds good to me too, feel free to add it here.

retronym:
Adding `-u` will need to be done with some care, idioms like:

```
if [[ -n ${SCALA_RUNNER_DEBUG:} ]]; then
```

will fail under that mode. Let's address that in a separate PR.

SethTisue:
~~someone might like to backport this to 2.11.x for a possible eventual 2.11.13 release someday~~

EDIT: oh, that already happened (scala/scala#6205)

PR #5589, opened 2016-12-08
SI-4700 Types with symbolic names print in infix by default

ghost:
before:

```scala
scala> implicitly[Int =:= Int]
res0: =:=[Int,Int] = <function1>
```

after:

```scala
scala> implicitly[Int =:= Int]
res0: Int =:= Int = <function1>
```

Building off of @VladUreche's work in #5401, here's a PR that addresses some of the comments there.

- makes two-parameter types with symbolic names print infixly by default, and
- adds a `@showAsInfix` annotation to enable/disable override that for types.

Review by @lrytz, @retronym, and others?

--allison

### DIFF
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -307,6 +307,9 @@ trait Types
     /** Is this type completed (i.e. not a lazy type)? */
     def isComplete: Boolean = true
 
+    /** Should this be printed as an infix type (@showAsInfix class &&[T, U])? */
+    def isShowAsInfixType: Boolean = false
+
     /** If this is a lazy type, assign a new type to `sym`. */
     def complete(sym: Symbol) {}
 
@@ -2097,6 +2100,9 @@ trait Types
         trivial = fromBoolean(!sym.isTypeParameter && pre.isTrivial && areTrivialTypes(args))
       toBoolean(trivial)
     }
+
+    override def isShowAsInfixType: Boolean = sym.hasAnnotation(ShowAsInfixAnnotationClass)
+
     private[Types] def invalidateTypeRefCaches(): Unit = {
       parentsCache = null
       parentsPeriod = NoPeriod
@@ -2345,6 +2351,14 @@ trait Types
               xs.init.mkString("(", ", ", ")") + " => " + xs.last
           }
         }
+        else if (isShowAsInfixType && args.length == 2)
+          args(0) + " " + sym.decodedName + " " +
+            (
+              if (args(1).isShowAsInfixType)
+                "(" + args(1) + ")"
+              else
+                args(1)
+            )
         else if (isTupleTypeDirect(this))
           tupleTypeString
         else if (sym.isAliasType && prefixChain.exists(_.termSymbol.isSynthetic) && (this ne dealias))
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -2101,7 +2101,9 @@ trait Types
       toBoolean(trivial)
     }
 
-    override def isShowAsInfixType: Boolean = sym.hasAnnotation(ShowAsInfixAnnotationClass)
+    /* It only makes sense to show 2-ary type constructors infix. */
+    override def isShowAsInfixType: Boolean =
+      sym.hasAnnotation(ShowAsInfixAnnotationClass) && hasLength(args, 2)
 
     private[Types] def invalidateTypeRefCaches(): Unit = {
       parentsCache = null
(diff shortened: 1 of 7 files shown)


lrytz:
Thanks @allisonhb for picking this up, it looks great in the current form! CC @adriaanm 

However, it adds a public class to the standard library, which is not forwards binary compatible. @SethTisue do you know why this is not flagged by MiMa? Could we ignore this case?

retronym:
I we fixed a MiMa bug recently in our build. If this PR were rebased on 2.12.x, the compat violation ought to be reported. 

I'm not totally against the idea of allowing annotations in a minor release. Need to give it some thought. 

Last time we considered this was with `@compileTimeOnly`, which we decided had to wait until 2.11.0 to move from reflect.internal to annotation.

ghost:
So for what little it's worth, I can't think of any places where someone would rather keep the prefix notation for symbolic type names, and only a few ([refined](https://github.com/fthomas/refined) comes to mind) where infix notation is preferred for non-symbolic names.

On the other hand, I know that at least in my workplace, types such as `shapeless.::[Int, shapeless.::[(Long, Long), shapeless.::[String, shapeless.HNil]]]` are a major cause for complaint. If you like I can rebase this onto 2.12.x, drop the annotation, and reinstate it in a 2.13.x PR.

--allison

edit: woops accidentally pushed "close + commit" button...

milessabin:
@allisonhb I'd be very keen to merge this in Typelevel Scala for the upcoming 2.12.1 release ... please let me know if you'd be happy with that.

ghost:
@milessabin that sounds great! I'll make a PR against TLS as soon as I get home.

milessabin:
@allisonhb no need ... I can merge this PR to TLS as it stands. Thanks for working on this!

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -2097,6 +2100,15 @@ trait Types
         trivial = fromBoolean(!sym.isTypeParameter && pre.isTrivial && areTrivialTypes(args))
       toBoolean(trivial)
     }
+
+    /* It only makes sense to show 2-ary type constructors infix.
+     * By default we do only if it's a symbolic name. */
+    override def isShowAsInfixType: Boolean =
+      args.size == 2 &&
```
adriaanm: It's best to avoid calling `size` on a list. We have a more performant convenience method for this: `hasLength(args, 2)`

lrytz:
I did some testing what will happen if a class annotated `@showAsInfix` is used (on the classpath) by an older compiler, where the annotation doesn't exist.

```
scala> cls.annotations
res15: List[$r.intp.global.AnnotationInfo] = List(showAsInfix(showAsInfix.<init>$default$1))

scala> cls.annotations.head.atp.typeSymbol
res16: $r.intp.global.Symbol = class showAsInfix

scala> cls.annotations.head.atp.typeSymbol.getClass
res17: Class[_ <: $r.intp.global.Symbol] = class scala.reflect.internal.Symbols$StubClassSymbol
```

so this seems to be safe.

lrytz:
LGTM, veto by @retronym?

SethTisue:
Not sure about the details of the code, but I definitely support this design change.

milessabin:
FYI: this has been [merged](https://github.com/typelevel/scala/blob/typelevel-readme/notes/2.12.1.md#infix-type-printing-pull5589-allisonhb) in Typelevel Scala 2.12.1.

adriaanm:
Based on Lukas's testing, I think this is ready to go.

fommil:
we get failures in the `ensime-server` tests that I think are because of this... looks like our inferred scaladoc URLs are changing, e.g. having to add to our CI

```
        equal(Some("docs/scala-library-" + c.scalaVersion + "-javadoc.jar/scala/Some.html#flatten[B](implicitev:<:<[A,Option[B]]):Option[B]")) or
        equal(Some("docs/scala-library-" + c.scalaVersion + "-javadoc.jar/scala/Some.html#flatten[B](implicitev:A<:<Option[B]):Option[B]"))
```

fommil:
looks like the tags of the actual scaladocs did change with this. Was that expected?

adriaanm:
I hadn't expected that, no! Could you file a bug at scala/bug?

fommil:
@adriaanm I'm not sure if it is a bug, it's a regression if anything, but flip flopping on regressions is not great either

SethTisue:
@allisonhb since this PR is linked directly from the 2.12.2 release notes, I've edited your PR description to include a "before/after" pair

eparejatobes:
Would be really nice if types would display accordingly in the scaladoc output.

fommil:
I'd settle for scalac providing utility methods that converted between all the different kinds of names that a thing can have. This is the bulk of our work interfacing with the interactive compiler in ensime.

SethTisue:
> Would be really nice if types would display accordingly in the scaladoc output

that would make a great scala/bug ticket (if there isn't one already?), and a nice manageably sized issue, I expect, for a volunteer contributor to tackle

was thinking of this just now because I'm watching a talk where the speaker is demoing `HList` from Shapeless, and those types sure do look nice in infix.

PR #5596, opened 2016-12-12
Removed procedure syntax

Arneball:


lrytz:
Hi! The e-mail address used in the commit is not associated with your (or any) GitHub account, could you register it so that the patch is linked to your CLA signature? Thanks!

image

Ichoran:
I appreciate the interest in fixing these but if there's going to be an automated way to do it, starting the process piecemeal isn't necessarily an advantage.  If this is done with a rewriting tool, can you apply it to the entire library?  If not, do you have a plan for how to catch essentially everything so we don't have to keep revisiting this?  (Example of something missed: sys/process/ProcessImpl.scala `def destroy()`.)

adriaanm:
I agree with Rex. The downstream cost of making sweeping changes like this is quite high (need to also merge between different branches), so let's do this in one fell swoop once we have scalafix up and running.

Thank you for taking the initiative to fix this, but I'm going to close this until we can do this automatically.

PR #5597, opened 2016-12-12
Unix script template: get java version through java command

jodersky:
The `release` file does not exist in every java distribution (at least not in OpenJDK for Debian Stretch).
This change parses the output `java -version` to determine the version instead.

cc @retronym


jodersky:
nvm, just saw #5588

PR #5601, opened 2016-12-13
Consider a val/var to match a nullary method with the same name

szeiger:
The current definition of “matching” in the spec does not take the case
into account where a method matches a non-method member. I believe this
is necessary to make 5.1.4 (Overriding) correctly explain the case of
a val or var implementing an abstract method or overriding a concrete
method, both of which are allowed.

### DIFF
--- a/spec/05-classes-and-objects.md
+++ b/spec/05-classes-and-objects.md
@@ -263,6 +263,8 @@ and $M'$ bind the same name, and one of following holds.
    equal number of argument types $\overline T$, $\overline T'$
    and equal numbers of type parameters
    $\overline t$, $\overline t'$, say, and  $\overline T' = [\overline t'/\overline t]\overline T$.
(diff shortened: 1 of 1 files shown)


adriaanm:
Not sure about this one -- could you elaborate? My understanding is that the accessors are considered when deciding overriding, and they have a `(Nullary)MethodType`. The underlying field does have a non-method type, but it's private (and thus cannot override anything).

szeiger:
I see. Accessors are indeed mentioned in 4.2 (although they are called "getter and setter functions" instead of "methods" in some places, which is probably a bug).

I originally understood "Neither M nor M′ is a method definition" in the definition of "matching" in 5.1.3 as referring to vars and vals, which does not cover the case where only one of M and M' is a var/val and the other is a method. If you take equivalent setters and getters into account, this problem goes away.

This leaves the case of type members for "Neither M nor M′ is a method definition" where my proposed additional clause is actually wrong. I'm therefore closing this PR.

PR #5602, opened 2016-12-13
Modules w. serializable type alias "companions" are not serializable

retronym:
MiMa alerted us to a change in the parentage of two objects in the forkjoin
package object.

In Scala 2.12.0/1, they implemented `scala.Serializable`. Recently, this
(synthetically added) parent was absent. This appears to be due to a bug
fix in `companionSymbolOf`, which no longer treats objects and same-named
type aliases to be companions.

This commit manually adds the formerly-synthetic parents to these objects,
and documents the change in compiler behaviour with a test.

## Compatibility Note

Library authors with code like:

```scala
package object p1 {
  type A = SomeSerializableType
  object A
}
```

Will need to change this to:

```scala
package object p1 {
  type A = SomeSerializableType
  object A extends scala.Serializable
}
```

To remain technically binary compatible with the version compiled with Scala 2.12.0 or 2.12.1.

### DIFF
--- a/src/library/scala/concurrent/forkjoin/package.scala
+++ b/src/library/scala/concurrent/forkjoin/package.scala
@@ -25,7 +25,7 @@ package object forkjoin {
   @deprecated("use java.util.concurrent.ForkJoinTask directly, instead of this alias",         "2.12.0")
   type ForkJoinTask[T] = juc.ForkJoinTask[T]
   @deprecated("use java.util.concurrent.ForkJoinTask directly, instead of this alias",         "2.12.0")
-  object ForkJoinTask {
+  object ForkJoinTask extends scala.Serializable {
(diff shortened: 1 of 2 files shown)


retronym:
This [bootstrap build](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/696/) will tell us if we've appeased MiMa.

retronym:
image

sjrd:
Does that mean that [SI-9668](https://issues.scala-lang.org/browse/SI-9668) is fixed?

Jasper-M:
I thought type aliases could never have companion objects?

retronym:
@sjrd I didn't see that bug when you opened it, but, yes, that's exactly the issue fixed here.

lrytz:
LGTM - merging this to unblock CI failures for other PRs

PR #5603, opened 2016-12-16
[ci:last-only] Remove parallel collections from scala-library

szeiger:
They will be moved into a separate module as part of the ongoing
modularization of the Scala standard library.


SethTisue:
see also https://github.com/scala/scala-parallel-collections

adriaanm:
Some ideas (probably a bit too ambitious, but hey, it's the new year!) for the next steps in modularization: https://github.com/scala/scala/compare/a5864ad...adriaanm:modularize-v2?expand=1

szeiger:
Rebased on current 2.13.x & added a commit to remove all tests for parallel collections

SethTisue:
@jvican fyi on modularization

szeiger:
Scabot didn't care about my `ci:last-only` after an earlier build had already failed, but the new one is green, so this is good to go. Removing `wip` label.

szeiger:
Here are the removed tests in scala-parallel-colletions: https://github.com/scala/scala-parallel-collections/pull/6

adriaanm:
Proposed next steps in modularization: https://github.com/scala/scala/compare/eb9fdcd...adriaanm:modularize-213-step1?expand=1. 

https://github.com/scala/scala/pull/5677  lays the groundwork in 2.12.x

lrytz:
Opinions on trying to get this merged before M1? Is the new module (https://github.com/scala/scala-parallel-collections) already in a state where it could be published against M1?

szeiger:
Everything is ready except for the publishing. I haven't yet found the time to work on this again. What's the best way to bootstrap? The code in the module overlaps with what we remove here, so we can't publish the module first. We need a version of Scala without parallel collections to publish the module against.

lrytz:
Should the module be part of scala-library-all? Then we need to build it as part of our bootstrap job, I can take care of that. But if scala-library-all goes away in 2.13, we can skip this step. In this case, we could merge this PR, release M1, and then publish the new module afterwards.

Maybe we better run the community build before merging. Could you rebase to fix the merge conflict?

SethTisue:
I would support eliminating scala-library-all. (https://github.com/search?p=4&q=scala-library-all&type=Code&utf8=✓ only shows 258 results GitHub-wide, which is not zero but seems like a fairly negligible number to me. And I've never seen anyone recommend using it — books, blogs, chat rooms, etc.)

szeiger:
If we keep scala-library-all then parallel-collections should be a part of it IMHO, but I don't know why we'd want to keep scala-library-all, either.

I'll rebase this PR to get it ready to merge again and port any changes that have been done in the meantime to the parallel-collections repo.

szeiger:
Rebased. There was one merge conflict in the imports of `WrappedArray` and a bunch of scalacheck tests were moved around in https://github.com/scala/scala/pull/5661. Nothing new that needs to be ported to the separate module.

lrytz:
scalacheck tests for parallel collections got restored accidentally, probably caused by https://github.com/scala/scala/pull/5661
```
[info] Compiling 62 Scala sources to /home/jenkins/workspace/scala-2.13.x-validate-test/target/scalacheck/test-classes...
[error] /home/jenkins/workspace/scala-2.13.x-validate-test/test/scalacheck/scala/collection/parallel/ParallelHashTrieCheck.scala:21: not found: type ParHashMap
[error]   type CollType = ParHashMap[K, V]
[error]                   ^
```

dwijnand:
Does the decision for what's in or not scala-library-all determine what's present in the `scala` REPL?

Or is that a separate decision?

Should `scala` 2.13 include the parallel collections?

lrytz:
> Does the decision for what's in or not scala-library-all determine what's present in the scala REPL?

Good point, we do ship these modules with the downloadable distribution. If we remove scala-library-all, we have to decide what to do on this front. It would be more elegant to get them from maven-central, have a list of default dependencies, and allow users to add more.

> Should scala 2.13 include the parallel collections?

The plan is to move them to a module, out of scala-library.

szeiger:
Updated and squashed.

lrytz:
```
[info] Compiling 44 Scala sources to /home/jenkins/workspace/scala-2.13.x-validate-test/target/scalacheck/test-classes...
[error] /home/jenkins/workspace/scala-2.13.x-validate-test/test/scalacheck/scala/pc.scala:5: object parallel is not a member of package collection
[error] import scala.collection.parallel._
[error]                         ^
```

I guess you missed a few.. In the meantime, I started a community build:
https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/271/console

szeiger:
Bah, I thought it was only JUnit tests. I didn't run partest locally.

szeiger:
Oh, I see. I ran the wrong set of JUnit tests.

lrytz:
There will be some things to handle, as shown by the community build:
```
[scala-js] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.7-RC1/project-builds/scala-js-7a00b322d46fa67b6a790f73709ab9e9984ef334/tools/jvm/src/main/scala/org/scalajs/core/tools/linker/frontend/optimizer/ParIncOptimizer.scala:14: object parallel is not a member of package collection
[scala-js] [error] import scala.collection.parallel.mutable.{ParTrieMap, ParArray}
[scala-js] [error]                         ^
```

szeiger:
As expected. All projects that use parallel collections will need the new project as a dependency. If they use `.par` they also need an additional import because that is now an extension method.

lrytz:
If we want to merge this and get a green community build, we need to add https://github.com/scala/scala-parallel-collections to the community build. Projects (or our forks) need to add a dependency, probably in a way that it only applies if the Scala version is 2.13*.

SethTisue:
from https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/271/consoleFull

```
[info] >>> The dbuild result is------------: FAILED (failed: scala-js, twitter-util, fs2, specs2)
```

lots of projects depend on scala-js and specs2 especially, so those are probably masking additional downstream failures

I think we should bite the bullet here, even if it initially causes the 2.13 community build to contract substantially. we can gradually chip away at growing it again, one project at a time.

SethTisue:
Note that that test run still includes 40 `SUCCESS`ful projects (of 94 total)

lrytz:
Ah, you're suggesting to shrink the community build for releasing M1 to the projects that don't depend on parallel collections? That would be easier than trying to fix those projects beforehand, because with M1 they have a 2.13 version that they can use themselves, and there will also have a release of the new module. +1

SethTisue:
> Ah, you're suggesting to shrink the community build for releasing M1 to the projects that don't depend on parallel collections

yes...

...except that before M1 I intend to fix 2 or 3 key projects like scala-js and specs2 myself, in forks, so we know that this all works for at least a few projects, and because that will allow more downstream projects to be built

I've made a ticket on it (https://github.com/scala/community-builds/issues/495) and added a checkbox to https://github.com/scala/scala-dev/issues/332, too

SethTisue:
I wonder how much this knocks off our PR validation time.

retronym:
I noticed that this PR has improved the performance of our compiler benchmarks. I haven't investigated the root cause the improvement, but I'd wager that reducing the complexity of base type sequences of collections (which are computed any time someone compiles against the collections, which is pretty much every compilation) is a key factor.

https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=ef63cb91bf%7Ce2a2cbacd0&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=now-6M&to=now

image

jvican:
Wow... Do you guys have benchmarks with other projects to check if it's reproducible outside of better-files?

retronym:
We also see it in `Vector.scala` (the third chart), and to a lesser degree in `scalap` (the second chart). 

Given that base type sequences are cached, it would seem large compilation batches would not see the improvement (the one-time extra cost of the more complex BTS computation is amortized against lots of places that use the type).

But, reducing overhead for compiling a single file is still really helpful, as that's what people experience in incremental compilation.

retronym:
We also have a benchmark series for compiling `scala-{library,reflect,compiler}` jointly to test "big projects", but we that is slower to execute, so we don't collect fine-grained results for it.

PR #5604, opened 2016-12-17
Use ClassTag instead of Manifest in AssertUtil.assertThrows.

sjrd:
This allows it to work in Scala.js, which supports `ClassTag`s but not `Manifest`s.


lrytz (approved):

lrytz:
Thanks!

sjrd:
That was quick, thanks!

PR #5605, opened 2016-12-18
Fix spec links to `07-implicits.html`

szeiger:
The file referenced by the current links actually exists under
http://www.scala-lang.org/files/archive/spec/2.11/ but with a different
design than the rest. It seems to be left over from an older version.


SethTisue:
the `rsync` command in our `.travis.yml` doesn't use `--delete`, so that's probably why there would be old files left around

SethTisue (approved):

PR #5607, opened 2016-12-20
Override equals and hashCode for WrappedArray

adriaanm:
Rebase of #5551 (also removed mima filter that should no longer be needed)


lrytz:
The whitelist change moved from `bincompat-forward.whitelist.conf` to `bincompat-backward.whitelist.conf`, which seems to be wrong

lrytz:
I `git push -f`'d to your branch, how evil! 🙈

adriaanm:
🙉

alexeyr:
@lrytz It seemed wrong to me to, but that's what I had to do to avoid MiMa errors. Maybe MiMa had a bug fixed since then?

PR #5609, opened 2016-12-20
scala-xml dependency should be provided

ashawley:
Nothing should transitively depend on the scala-xml module except for
scaladoc and library-all.

Compiler needs scala-xml as a build dependency, but not at runtime.
The user is supposed to provide scala-xml as a dependency at runtime.
This fixes an issue where the classpath in SBT shadows the scala-xml
module, see scala/scala-module-dependency-sample#14.  The only
consequence of dropping the dependency at runtime is that scala-xml
always needs to be binary compatible.  That's the point of MiMa and
dbuild, no?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -423,7 +423,7 @@ lazy val compiler = configureAsSubproject(project)
     description := "Scala Compiler",
     libraryDependencies ++= Seq(antDep, asmDep),
     // These are only needed for the POM:
-    libraryDependencies ++= Seq(scalaXmlDep, scalaParserCombinatorsDep, jlineDep % "optional"),
+    libraryDependencies ++= Seq(scalaXmlDep % "provided", scalaParserCombinatorsDep, jlineDep % "optional"),
     // this a way to make sure that classes from interactive and scaladoc projects
     // end up in compiler jar (that's what Ant build does)
     // we need to use LocalProject references (with strings) to deal with mutual recursion


ashawley:
Seems the dependency in scaladoc on scala-xml is broken:

```
[error] (scalap/compile:doc) java.lang.NoClassDefFoundError: scala/xml/NamespaceBinding
[error] (reflect/compile:doc) java.lang.NoClassDefFoundError: scala/xml/NamespaceBinding
[error] (actors/compile:doc) java.lang.NoClassDefFoundError: scala/xml/NamespaceBinding
[error] (library/compile:doc) java.lang.NoClassDefFoundError: scala/xml/NamespaceBinding
[error] (compiler/compile:doc) java.lang.NoClassDefFoundError: scala/xml/NamespaceBinding
```

adriaanm:
Thanks for looking into this! I think you're on the right track with your analysis at https://github.com/scala/scala-module-dependency-sample/issues/14. I'm a bit concerned with changing the current setup in 2.11.9, though. Even if it's suboptimal, I'd rather not change this in the last 2.11 release.

We should absolutely take this up in 2.13 to make sure we have a cleaner dependency graph. Maybe finally cut the scaladoc dependency's on scala-xml.

adriaanm:
I'm going to close this for 2.11. Tracking the 2.13 issue at https://github.com/scala/scala-dev/issues/291. Please consider resubmitting for 2.13.x!

ashawley:
Thanks @adriaanm.  Indeed, there should definitely be a clearer dependency graph between the scala-compiler, scaladoc, and scala-xml.  It's a bit haphazard at the moment, but I can understand why you would want to leave the issue behind in 2.11 and 2.12.

Now that the scala-compiler build is defined in SBT instead of Ant, this should be easier to resolve.

I'll try to find time to resubmit against 2.13.

PR #5610, opened 2016-12-21
Fix how "sbt" is written

dwijnand:
"sbt" is not an acronym (it used to be, but it isn't any longer).
It's a proper name, like "iPhone" or "eBay".

So, just like you wouldn't write "Get Started With EBay" or
"How To Reset Your IPhone", we don't write "Using the Sbt Build".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,7 +50,7 @@
 # Standard symbolic link to build/quick/bin
 /qbin
 
-# Sbt's target directories
+# sbt's target directories
 /target/
 /project/target/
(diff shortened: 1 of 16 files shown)


smarter:
> "sbt" is not an acronym

I thought everyone knew that sbt stands for "sbt build tool" :)

lrytz (approved):

dwijnand:
That was fast :) Thanks Lukas.

lrytz:
Thanks to you :)

PR #5611, opened 2016-12-21
Common usage of finding random number in a range [ci: last-only]

fhalde:
Is it a good idea to have a function like `between` within the `scala.util.Random` class?
It is quite often used while developing games and stuff. Also I feel like it is a functionality closer to `Random` class

### DIFF
--- a/src/library/scala/util/Random.scala
+++ b/src/library/scala/util/Random.scala
@@ -43,10 +43,32 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    */
   def nextDouble(): Double = self.nextDouble()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
+   *  between min (inclusive) and max (exclusive) from this random number generator's sequence.
+   */
+  def between(minInclusive: Double, maxExclusive: Double): Double = {
+    require(minInclusive < maxExclusive, "Invalid bounds")
+
+    val next = nextDouble() * (maxExclusive - minInclusive) + minInclusive
+    if (next < maxExclusive) next
+    else Math.nextAfter(maxExclusive, Double.NegativeInfinity)
+  }
+
   /** Returns the next pseudorandom, uniformly distributed float value
    *  between 0.0 and 1.0 from this random number generator's sequence.
    */
   def nextFloat(): Float = self.nextFloat()
+
+  /** Returns the next pseudorandom, uniformly distributed float value
+   *  between min (inclusive) and max (exclusive) from this random number generator's sequence.
+   */
+  def between(minInclusive: Float, maxExclusive: Float): Float = {
+    require(minInclusive < maxExclusive, "Invalid bounds")
+
+    val next = nextFloat() * (maxExclusive - minInclusive) + minInclusive
+    if (next < maxExclusive) next
+    else Math.nextAfter(maxExclusive, Float.NegativeInfinity)
+  }
 
   /** Returns the next pseudorandom, Gaussian ("normally") distributed
    *  double value with mean 0.0 and standard deviation 1.0 from this
@@ -65,10 +87,87 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    */
   def nextInt(n: Int): Int = self.nextInt(n)
 
+  /** Returns a pseudorandom, uniformly distributed int value between min
+   *  (inclusive) and the specified value max (exclusive), drawn from this
+   *  random number generator's sequence.
+   */
+  def between(minInclusive: Int, maxExclusive: Int): Int = {
+    require(minInclusive < maxExclusive, "Invalid bounds")
+
+    val difference = maxExclusive - minInclusive
+    if (difference >= 0) {
+      nextInt(difference) + minInclusive
+    } else {
+      /* The interval size here is greater than Int.MaxValue,
+       * so the loop will exit with a probability of at least 1/2.
+       */
+      def loop(): Int = {
+        val n = nextInt()
+        if (n >= minInclusive && n < maxExclusive) n
+        else loop()
+      }
+      loop()
+    }
+  }
+
   /** Returns the next pseudorandom, uniformly distributed long value
    *  from this random number generator's sequence.
    */
   def nextLong(): Long = self.nextLong()
+
+  /** Returns a pseudorandom, uniformly distributed long value between 0
+   *  (inclusive) and the specified value (exclusive), drawn from this
+   *  random number generator's sequence.
+   */
+  def nextLong(n: Long): Long = {
+    require(n > 0, "n must be positive")
+
+    /*
+     * Divide n by two until small enough for nextInt. On each
+     * iteration (at most 31 of them but usually much less),
+     * randomly choose both whether to include high bit in result
+     * (offset) and whether to continue with the lower vs upper
+     * half (which makes a difference only if odd).
+     */
+
+    var offset = 0L
+    var _n = n
+
+    while (_n >= Integer.MAX_VALUE) {
+      val bits = nextInt(2)
+      val halfn = _n >>> 1
+      val nextn =
+        if ((bits & 2) == 0) halfn
+        else _n - halfn
+      if ((bits & 1) == 0)
+        offset += _n - nextn
+      _n = nextn
+    }
+    offset + nextInt(_n.toInt)
+  }
+
+  /** Returns a pseudorandom, uniformly distributed long value between min
+   *  (inclusive) and the specified value max (exclusive), drawn from this
+   *  random number generator's sequence.
+   */
+  def between(minInclusive: Long, maxExclusive: Long): Long = {
+    require(minInclusive < maxExclusive, "Invalid bounds")
+
+    val difference = maxExclusive - minInclusive
+    if (difference >= 0) {
+      nextLong(difference) + minInclusive
+    } else {
+      /* The interval size here is greater than Long.MaxValue,
+       * so the loop will exit with a probability of at least 1/2.
+       */
+      def loop(): Long = {
+        val n = nextLong()
+        if (n >= minInclusive && n < maxExclusive) n
+        else loop()
+      }
+      loop()
+    }
+  }
 
   /** Returns a pseudorandomly generated String.  This routine does
    *  not take any measures to preserve the randomness of the distribution
--- a/test/junit/scala/util/RandomUtilTest.scala
+++ b/test/junit/scala/util/RandomUtilTest.scala
@@ -1,0 +1,39 @@
+package scala.util
+
+import org.junit.{ Assert, Test }
+import scala.util.Random
+
+class RandomUtilTest {
+  @Test def testBetween: Unit = {
+	val rand = new Random()
+	Assert.assertTrue("Random between Int must be inclusive-exclusive", rand.between(0, 1).equals(0))
+	Assert.assertTrue("Random between Long must be inclusive-exclusive", rand.between(0L, 1L).equals(0L))
+	Assert.assertTrue("Random nextLong must be inclusive-exclusive", rand.nextLong(1L).equals(0L))
+	val float: Float = rand.between(0.0f, 1.0f)
+	Assert.assertTrue("Float Random should be inclusive-exclusive", 0.0f <= float && float < 1.0f)
+	val double: Double = rand.between(0.0f.toDouble, 1.0f.toDouble)
+	Assert.assertTrue("Random between Double should be inclusive-exclusive", 0.0f.toDouble <= double && double < 1.0f.toDouble)
+  }
+
+  @Test(expected = classOf[IllegalArgumentException]) def testBetweenIntException: Unit = {
+	val rand = new Random()
+	rand.between(1, 0)
+  }
+
+
+  @Test(expected = classOf[IllegalArgumentException]) def testBetweenLongException: Unit = {
+	val rand = new Random()
+	rand.between(1L, 0L)
+  }
+
+  @Test(expected = classOf[IllegalArgumentException]) def testBetweenFloatException: Unit = {
+	val rand = new Random()
+	rand.between(1.0f, 0.0f)
+  }
+
+  @Test(expected = classOf[IllegalArgumentException]) def testBetweenDoubleException: Unit = {
+	val rand = new Random()
+	rand.between(1.0f.toDouble, 0.0f.toDouble)
+  }
+
+}


adriaanm:
Thanks! Our binary compatibility requirements preclude including this in a 2.12.x release. Please retarget your PR to 2.13.x (using the Edit button at the top of the page). I've assigned a reviewer to decide the API question itself.

src/library/scala/util/Random.scala
```diff
@@ -65,11 +75,23 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
   /** Returns the next pseudorandom, uniformly distributed long value
    *  from this random number generator's sequence.
    */
   def nextLong(): Long = self.nextLong()
 
+   /** Returns a pseudorandom, uniformly distributed long value between min
+   *  (inclusive) and the specified value max (exclusive), drawn from this
+   *  random number generator's sequence.
+   */
+  def between(min: Long, max: Long): Long = min + self.nextLong() % (max - min)
```
Ichoran: This does not sample uniformly and is buggy since the modulus of a negative number is negative.

src/library/scala/util/Random.scala
```diff
@@ -43,11 +43,21 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    */
   def nextDouble(): Double = self.nextDouble()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
+   *  between min and max from this random number generator's sequence.
+   */
+  def between(min: Double, max: Double): Double = min + self.nextDouble() % (max - min + 1.0)
```
Ichoran: This is the wrong formula.  It should be `min + (max - min)*self.nextDouble()`, or a variant that is guaranteed to not overflow the bounds and has minimal rounding error.

Ichoran:
I agree with the sentiment, but the implementation should be carefully done and there should be tests of the functionality.

fhalde:
@Ichoran Yes I was going to write tests. I just wanted to confirm its usability before writing the tests

fhalde:
@adriaanm Can someone explain me this binary compatibility? Is my change breaking anything and why?

src/library/scala/util/Random.scala
fhalde: I guess I just got carried away to generate a random number between two numbers without thinking of the uniformity of the random numbers generated. Not an easy task. Looking into some theory before I get back to this

src/library/scala/util/Random.scala
fhalde: overflow the bounds? Of double or the max value of the range?

src/library/scala/util/Random.scala
fhalde: Also I did a test to see the uniformity of `min + Math.abs(r.nextInt()) % (max - min) and the values between min and max were uniform. Coincidence? Or am I missing something

SethTisue:
> Can someone explain me this binary compatibility

http://docs.scala-lang.org/overviews/core/binary-compatibility-of-scala-releases.html

fhalde:
@adriaanm Seems like I did changes in `scala:2.12.x` which is causing problems in the pull request. Should I close this one?

SethTisue:
@fhalde you can rebase your changes onto the 2.13.x branch and force-push, or start over in a new PR, whatever's easier for you

it's a bit confusing/surprising that changing the target branch on GitHub doesn't offer to do the rebase for you.

fhalde:
@SethTisue rebase it is. Better to continue the conversation here itself. Thanks

adriaanm:
I'm not sure what happened, but your branch still has a bunch of commits from 2.12.x. When you did the rebase, did you use `git rebase -i` and retained only your commit on the list?

retronym:
I manually aborted the Jenkins jobs for the commits in this PR, hence the build status failures above.

retronym:
Please run `git reset --hard origin/2.13.x; git cherry-pick f6a8f58; git push -f` to fixup this branch.

fhalde:
@adriaanm phew done. forgot to remove other commits while rebasing

SethTisue:
you leveled up your GitHub fu! 🏆

sjrd:
I suggest taking a look at https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/ThreadLocalRandom.scala for sampling algorithms between two values.

src/library/scala/util/Random.scala
Ichoran: You are missing something.  Try min = 0L, max = (1L << 62) + (1L << 61) and check how many values fall in the ranges 0..(1L << 61), (1L << 61)..(2L << 61), and (2L << 61)..(3L << 61).  Alternatively, think more carefully about what happens when the last bit of the range is wrapped.  Also, (max - min) may overflow.

fhalde:
@sjrd looks like what I'm trying to achieve exists. My PR cannot compete against Doug Lea's implementation obviously

szeiger:
@fhalde The fact that it exists in Java API doesn't mean that it wouldn't be a worthwhile addition to Scala (since we already have our own `Random` class). But you could call the Java implementation or use the same algorithm.

src/library/scala/util/Random.scala
```diff
@@ -43,11 +43,21 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
+
   /** Returns the next pseudorandom, uniformly distributed float value
    *  between 0.0 and 1.0 from this random number generator's sequence.
    */
   def nextFloat(): Float = self.nextFloat()
 
+   /** Returns the next pseudorandom, uniformly distributed float value
+   *  between min and max from this random number generator's sequence.
+   */
+  def between(min: Float, max: Float): Float = min + self.nextFloat() % (max - min + 1.0f)
```
szeiger: The most confusing part for me when working with ranges is inclusive vs exclusive bounds. This should be made clear in every case (like this doc comment). I'd go a step further and add it to the parameter names (e.g. `minInclusive: Float, maxExclusive: Float`). You're not going to use these as named parameters anyway and the descriptive names will be shown by IDE auto-complete.

szeiger (changes requested):
(Using a proper github review to satisfy the review request:)

As explained in my other comments, I think this needs to be made clearer wrt include/exclusive bounds, and use the correct algorithms (either by calling the corresponding Java version or copying it)

fhalde:
@szeiger Thanks Stefan. I'll probably look into the implementation of the same in Java. Yes, I too was confused regarding the bounds while implementing it. Will write up the docs!

fhalde:
I've made the specified changes.
For the `between float`, I wrote an implementation following the `nextDouble` from `ThreadLocalRandom`
On a side note, I had a discussion with one of my friend. If a number has to be randomly generated, then why should we be concerned about it being uniformly distributed ? Uniform distribution doesn't make sense because it's a Random number after all. It can be anything!

SethTisue:
> Uniform distribution doesn't make sense because it's a Random number after all. It can be anything

non-uniformly-distributed random number generator: http://dilbert.com/strip/2001-10-25

fhalde:
@SethTisue while working on this, `min + nextInt(max - min)` was not considered uniformly random according to many internet resource :confused:

Uniformly random, what the irony!

src/library/scala/util/Random.scala
```diff
@@ -43,11 +44,27 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
    */
   def nextDouble(): Double = self.nextDouble()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
+   *  between min (inclusive) and max (exclusive) from this random number generator's sequence.
+   */
+  def between(minInclusive: Double, maxExclusive: Double): Double = ThreadLocalRandom.current().nextDouble(minInclusive, maxExclusive)
```
szeiger: Using `ThreadLocalRandom.current()` ignores the seed value, thus making it impossible to get reproducible sequences of values. Since `scala.util.Random` can be constructed with a specific seed value, I would expect all methods to honor it (and the scaladoc comment even states it explicitly).

We could replace the wrapped `java.util.Random` by a `java.util.SplittableRandom` (Java 8+ only). Unfortunately, `scala.util.Random` exposes the wrapped `self` value and for some reason `SplittableRandom` is not a subclass of `Random`. (`ThreadLocalRandom` OTOH is a subclass of `Random` but cannot be instantiated).

The backwards-compatible solution would be to reimplement all of the `between` methods the same way it is done for the `Float` version. Not particularly elegant but the implementaion is simple enough after all. Or we bite the bullet and break backwards compatibility by switching to `SplittableRandom`. Opinions?

src/library/scala/util/Random.scala
```diff
@@ -43,11 +44,27 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
   def nextFloat(): Float = self.nextFloat()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
+   *  between min (inclusive) and max (exclusive) from this random number generator's sequence.
+   */
+  def between(minInclusive: Float, maxExclusive: Float): Float = {
+		require(minInclusive >= maxExclusive, "Invalid bounds")
+
+		val next = nextFloat() * (maxExclusive - minInclusive) + minInclusive
+    	if (next < maxExclusive) next
```
szeiger: The formatting of the lines above looks messed up. Did some tabs sneak in?

src/library/scala/util/Random.scala
```diff
@@ -43,11 +44,27 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
   /** Returns the next pseudorandom, uniformly distributed float value
    *  between 0.0 and 1.0 from this random number generator's sequence.
    */
   def nextFloat(): Float = self.nextFloat()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
+   *  between min (inclusive) and max (exclusive) from this random number generator's sequence.
+   */
+  def between(minInclusive: Float, maxExclusive: Float): Float = {
+		require(minInclusive >= maxExclusive, "Invalid bounds")
```
szeiger: Shouldn't this be `minInclusive < maxExclusive`?

szeiger (changes requested):

src/library/scala/util/Random.scala
fhalde: yup. on it

src/library/scala/util/Random.scala
fhalde: looks fine in my vim. gotta check

src/library/scala/util/Random.scala
fhalde: umm, _**making it impossible to get reproducible sequences of value**_ is this for testing purpose ?
To me it looks reasonable to maintain backward compatibility considering the little effort required.

Speaking of testing, can't one always make a proxy object ?

SethTisue: one of the whole points of `java.util.Random` and `scala.util.Random` is to always give the same series of [pseudorandom numbers](https://en.wikipedia.org/wiki/Pseudorandom_number_generator) when seeded with the same seed. they are PRNGs, not true RNGs. and yes, testing (not just of the RNG itself, either, but of anything that uses it) is one of the reasons you might want this. another is that sometimes when doing scientific work it's important that results be reproducible.

fhalde: Hey @SethTisue the solution works for Float, Double and thankfully Int (since it has a `nextInt(n)`)

Using the formula `nextLong * (max - min) + min` would cause overflow if i'm not wrong !!
For `Int`, I've used `nextInt(max - min) + min`

src/library/scala/util/Random.scala
```diff
@@ -65,11 +88,27 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
   /** Returns the next pseudorandom, uniformly distributed long value
    *  from this random number generator's sequence.
    */
   def nextLong(): Long = self.nextLong()
 
+   /** Returns a pseudorandom, uniformly distributed long value between min
+   *  (inclusive) and the specified value max (exclusive), drawn from this
+   *  random number generator's sequence.
+   */
+  def between(minInclusive: Long, maxExclusive: Long): Long = ThreadLocalRandom.current.nextLong(minInclusive, maxExclusive)
```
fhalde: I remember `nextLong() % (max - min) + min` not being accepted previously. Sounds like a solution ?

src/library/scala/util/Random.scala
```diff
@@ -43,11 +44,33 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
    */
   def nextFloat(): Float = self.nextFloat()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
+   *  between min (inclusive) and max (exclusive) from this random number generator's sequence.
+   */
+  def between(minInclusive: Float, maxExclusive: Float): Float = {
+    require(minInclusive < maxExclusive, "Invalid bounds")
+
+    val next = nextFloat() * (maxExclusive - minInclusive) + minInclusive
```
Ichoran, line 68: This should use `nextDouble` for a more accurate calculation.

src/library/scala/util/Random.scala
Ichoran: As discussed, this doesn't use the seed of this random number generator.  Modulus is a poor solution because for large ranges you can be off by a factor of 2 in probability of getting a value (!).  There are methods that either reroll if you end up off the end, or which use two Longs and do long division by hand to get acceptably small error (2^-64) in probability.

fhalde:
What does one think of the following solution for `nextLong(n)`? 

Having this will make `nextLong(min, max)` easier.

long has 8 bytes. how about we set individual bits using `nextInt(2)` ? Obviously whilst taking care of the boundary condition. Whenever we are setting a bit to 1, we do the bound check. If it is exceeds the limit then we don't set that bit.

SethTisue:
> how about we set individual bits using nextInt(2)

two problems:

1) I don't think the algorithm you suggest gives uniformly distributed results. you would need to cite a published source saying it does, but I believe you won't find one, by the following argument. consider generating a number between 0 and 4, inclusive. the chance of getting each number should be exactly 1/5. but by this method, 4 will be the answer 1/2 of the time if we set the bits from the high bits down. if we set the bits from the low bit up instead, 4 will only be chosen 1/8 of the time.
2) it sounds extremely like a very slow/expensive method (since it takes so many draws from the underlying generator)

fhalde:
You recently agreed that `random` function should not be concerned with `uniform distribution` :stuck_out_tongue:

Ichoran:
I don't understand why we're having this discussion.  @sjrd already linked to a reasonable implementation.  @fhalde - Why not just implement that one instead of trying to guess at a new one?

SethTisue:
Perhaps you misunderstood a joke, or perhaps I'm misunderstanding your joke, but all joking aside: random numbers should be uniformly distributed, it's a very basic requirement for this. The most basic requirement.

src/library/scala/util/Random.scala
```diff
@@ -95,19 +95,80 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
   def between(minInclusive: Int, maxExclusive: Int): Int = {
     require(minInclusive < maxExclusive, "Invalid bounds")
 
-    return nextInt(maxExclusive - minInclusive) + minInclusive
```
fhalde: Had a bug if `maxExclusive - minInclusive` overflowed, then `nextInt` threw a `illegal argument exception`

src/library/scala/util/Random.scala
```diff
@@ -95,19 +95,80 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
    *  from this random number generator's sequence.
    */
   def nextLong(): Long = self.nextLong()
 
-   /** Returns a pseudorandom, uniformly distributed long value between min
+  /** Returns a pseudorandom, uniformly distributed long value between 0
+   *  (inclusive) and the specified value (exclusive), drawn from this
+   *  random number generator's sequence.
+   */
+  def nextLong(n: Long): Long = {
```
fhalde, line 122: As per the discussion, picked the implementation from thread local random

src/library/scala/util/Random.scala
```diff
@@ -95,19 +95,80 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
+    }
+    offset + nextInt(_n.toInt)
+  }
+
+  /** Returns a pseudorandom, uniformly distributed long value between min
    *  (inclusive) and the specified value max (exclusive), drawn from this
    *  random number generator's sequence.
    */
-  def between(minInclusive: Long, maxExclusive: Long): Long = ThreadLocalRandom.current.nextLong(minInclusive, maxExclusive)
+  def between(minInclusive: Long, maxExclusive: Long): Long = {
```
fhalde: as per the discussion, picked up from thread local random

fhalde:
All the review comments have been met here.

src/library/scala/util/Random.scala
```diff
@@ -43,11 +44,33 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
+    if (next < maxExclusive) next
+    else Math.nextAfter(maxExclusive, Double.NegativeInfinity)
+  }
+
   /** Returns the next pseudorandom, uniformly distributed float value
    *  between 0.0 and 1.0 from this random number generator's sequence.
    */
   def nextFloat(): Float = self.nextFloat()
 
+  /** Returns the next pseudorandom, uniformly distributed double value
```
szeiger: this should say "float" instead of "double"

src/library/scala/util/Random.scala
```diff
@@ -65,11 +88,88 @@ class Random(val self: java.util.Random) extends AnyRef with Serializable {
...
+   */
+  def between(minInclusive: Int, maxExclusive: Int): Int = {
+    require(minInclusive < maxExclusive, "Invalid bounds")
+
+    val difference = maxExclusive - minInclusive
+    if (difference >= 0) {
+      nextInt(difference) + minInclusive
+    } else {
+      /* The interval size here is greater than Int.MaxValue,
+       * so the loop will exit with a probability of at least 1/2.
```
szeiger, line 102: The algorithm looks reasonable to me. WDYT, @Ichoran?

src/library/scala/util/Random.scala
fhalde: fixed

fhalde:
do you want me to squash all the commits into one ?

szeiger:
Yes, please.

SethTisue:
@szeiger have your review comments been addressed?

szeiger (approved):

SethTisue:
@fhalde can you rebase onto current 2.13.x? after that this should be ready to merge.

fhalde:
yup

SethTisue:
thanks!

@Ichoran I don't think we have your sign-off on this...? not sure if all your review comments were addressed or not.

src/library/scala/util/Random.scala
Ichoran: By eye it looks fine, @szeiger

src/library/scala/util/Random.scala
```diff
@@ -13,6 +13,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.collection.generic.CanBuildFrom
 import scala.collection.immutable.{ List, Stream }
 import scala.language.{implicitConversions, higherKinds}
+import java.util.concurrent.ThreadLocalRandom
```
Ichoran: This shouldn't be imported.  It's not used.

Ichoran (changes requested):
There's an extra import and the float calculation should be done in doubles to reduce error.

Also, a basic test of the new functionality should be written.

fhalde:
There doesn't seem to be a test case for the `scala` version of Random class. Could I know in which path should I place my test cases ? Does `test/junit/*` look fine ?

szeiger:
Yes, JUnit test are preferred for library code.

SethTisue:
(sorry for the endless rebasing, but we did a big 2.11->2.12->2.13 merge recently)

fhalde:
So I'll rebase it with 2.13.x by tomorrow

lrytz:
@fhalde this seems to be almost ready, just a basic JUnit test and the stale import that @Ichoran mentioned.

I'm moving it to 2.13.0-M2 for now, as 2.13.0-M1 is going to be released soon and it's not blocking that release. But if it's ready in time we can still merge it.

fhalde:
All done. Sorry for the tremendous delay !! @lrytz

lrytz:
I think most delays were on our end :)

PR #1900, opened 2013-01-15
Modifies "maybeRewrap" to focus more on the maybe.

paulp:
Existential types are rewrapped under a bunch of conditions
unless the operation performed on the underlying type returns
the same type by reference equality. That depends on a
foundation of predictability which doesn't exist. The upshot is
that existential types were rewrapped with abandon, even when
the type were identical.

This had both performance and correctness implications.
Note where the test case output changes like so:

  -scala.collection.immutable.List[Any]
  +scala.collection.immutable.List[<?>]

That's correctness.

PR #5613, opened 2016-12-23
SI-10016 preserve annotation on existential

adriaanm:
We'll just have to live with manifests not being
able to resolve the difference between `scala.List[_]` and `scala.List[Any]`
(dealiasing causes rewrapping causes existential extrapolation).
They are equivalent types after all (and manifests are deprecated).

A type tag will be more precise.

This commit reverts #1900, which caused this regression.
Recently, #5263 improved the situation a bit, but showed that
the original tweak was not a good idea (too many exceptions).
Also, I doubt using `=:=` will improve performance, even if it
reduces allocation a bit, as it's far more expensive than `eq`...


SethTisue:
@adriaanm should this remain in the PR queue...?

PR #5614, opened 2016-12-23
SI-9264 Disable JIT for a method failure [no-merge] [:pray:]

retronym:
I found that after disabling JIT for this method the "genLoadIdent"
crash went away. It was intermittend beforehand, so I don't know for
sure if this will work, but let's try this out for a while.


lrytz:
`Unrecognized VM option 'UnlockDiagnosticVMOptions'`

PR #5615, opened 2016-12-24
[Backport] Modify ArrayBuilder and WrappedArrayBuilder to be reusable

monkey-mas:
# Problem
In 2.11.x, ArrayBuilder should be reusable after a call of `result` followed by `clear`, and it MUST not mutate any arrays it has created.

This is not the case under the condition below.
```
scala> import scala.collection.mutable.ArrayBuilder
import scala.collection.mutable.ArrayBuilder

scala> val builder = new ArrayBuilder.ofInt
builder: scala.collection.mutable.ArrayBuilder.ofInt = ArrayBuilder.ofInt

scala> builder ++= Vector.range(1, 17)
res0: builder.type = ArrayBuilder.ofInt

scala> val arr = builder.result()
arr: Array[Int] = Array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16)

scala> builder.clear()

scala> builder += 100
res2: builder.type = ArrayBuilder.ofInt

scala> val arr2 = builder.result()
arr2: Array[Int] = Array(100)

scala> arr
res3: Array[Int] = Array(100, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16)
// arr should be Array(1, .., 16) but was unexpectedly mutated by `+=` operation
```

`arr` was mutated as follows;
1. `result` & `clear`
    - `arr = elems`
    - `size = 0`
2. `+=(100)`
    - `ensureSize(0 + 1)`
     => `capacity < size || capacity == 0` is `false` as `capacity == 16` and `size == 1`
    -  `elems(0) = 100` this is where `arr(0) = 100` was done because we did not reallocate a new array for `elems` when calling `ensureSize`, which should have happened. 
    - `size = 1`
3. `result`
    - `mkArray(1)` gives us `arr2 = Array(100)`

# Solution
We should reset `capacity` to `0` when calling `result` so that `ensureSize(1)` calls `resize(16)`, which satisfies the property of Builder reusability. Besides mutation of previously created arrays does not happen.

### DIFF
--- a/src/library/scala/collection/mutable/ArrayBuilder.scala
+++ b/src/library/scala/collection/mutable/ArrayBuilder.scala
@@ -104,7 +104,10 @@ object ArrayBuilder {
     }
 
     def result() = {
-      if (capacity != 0 && capacity == size) elems
+      if (capacity != 0 && capacity == size) {
+        capacity = 0
+        elems
+      }
       else mkArray(size)
     }
 
@@ -169,7 +172,10 @@ object ArrayBuilder {
     }
 
     def result() = {
-      if (capacity != 0 && capacity == size) elems
+      if (capacity != 0 && capacity == size) {
+        capacity = 0
+        elems
+      }
       else mkArray(size)
     }
 
@@ -234,7 +240,10 @@ object ArrayBuilder {
     }
 
     def result() = {
-      if (capacity != 0 && capacity == size) elems
+      if (capacity != 0 && capacity == size) {
+        capacity = 0
+        elems
+      }
       else mkArray(size)
     }
 
@@ -299,7 +308,10 @@ object ArrayBuilder {
     }
 
     def result() = {
-      if (capacity != 0 && capacity == size) elems
+      if (capacity != 0 && capacity == size) {
+        capacity = 0
+        elems
+      }
       else mkArray(size)
     }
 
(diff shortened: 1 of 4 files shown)


Ichoran:
Note that the behavior was undefined in 2.11 and was redefined in 2.12 to do what you state.  So this isn't precisely a bug fix.  In 2.11, the API says that behavior after a call to result is undefined.

This backport (of functionality, not the ReusableBuilder trait) might be a good idea anyway, though.

monkey-mas:
> Note that the behavior was undefined in 2.11 and was redefined in 2.12 to do what you state. So this isn't precisely a bug fix. In 2.11, the API says that behavior after a call to result is undefined.

Oh I see, thanks for your comment! I'd love to backport if this change is okay as the current behaviour is a little bit confusing even though it's not a bug :)

Ichoran:
You might want to change WrappedArray's builder also (I think that was the only other behavioral change in https://github.com/scala/scala/commit/6eaae1b969b68ed3dc65a40613a8168b09246256).  I'm not sure whether the backport is a good idea; I tend to think yes, but I don't get the same kind of feedback as others like @SethTisue probably do.

monkey-mas:
> You might want to change WrappedArray's builder also

Yes, you’re right `WrappedArrayBuilder` has the same behavior. I’ve done the same change as I did to `ArrayBuilder`. Hope the PR is now ready to be reviewed. (it’s possibly not a good idea though) :)

Ichoran (approved):
Looks good save for someone deciding it's a good idea, which I don't feel qualified to do.

monkey-mas:
Thanks for your review :) I've squashed and updated the commit (log).

szeiger (approved):

PR #5616, opened 2016-12-26
Change the signature of WrappedArray.newBuilder[T] to return WrappedArrayBuilder[T]

monkey-mas:
(I'm not quite sure if this is a good idea but) As we have already defined [WrappedArrayBuilder](https://github.com/scala/scala/blob/v2.12.1/src/library/scala/collection/mutable/WrappedArrayBuilder.scala), we can return `WrappedArrayBuilder` instead of `ArrayBuffer`.

The return type has not been changed since it was first added to `WrappedArray` in this commit.[1]

[1] https://github.com/scala/scala/commit/d5b02c8652d7edbdfb0b5a02570d370d3bad299f#diff-5999d408677554483bfed5cdce592b1fR73

### DIFF
--- a/src/library/scala/collection/mutable/WrappedArray.scala
+++ b/src/library/scala/collection/mutable/WrappedArray.scala
@@ -132,7 +132,7 @@ object WrappedArray {
         ArrayBuilder.make[T]()(m) mapResult WrappedArray.make[T]
   }
 
-  def newBuilder[A]: Builder[A, IndexedSeq[A]] = new ArrayBuffer
+  def newBuilder[T](implicit t: ClassTag[T]): Builder[T, WrappedArray[T]] = new WrappedArrayBuilder[T](t)
 
   private val emptyWrappedByte = new ofByte(new Array[Byte](0))
   private val emptyWrappedShort = new ofShort(new Array[Short](0))


src/library/scala/collection/mutable/WrappedArray.scala
```diff
@@ -118,7 +118,7 @@ object WrappedArray {
         ArrayBuilder.make[T]()(m) mapResult WrappedArray.make[T]
   }
 
-  def newBuilder[A]: Builder[A, IndexedSeq[A]] = new ArrayBuffer
+  def newBuilder[T](implicit t: ClassTag[T]): WrappedArrayBuilder[T] = new WrappedArrayBuilder[T](t)
```
sjrd: The least would be for this function to return a `Builder[A, WrappedArrayBuilder[A]]`. `newBuilder` methods in the existing API always return a `Builder` in their public API, not a more specific type.

monkey-mas: Right, thanks! Fixed the return type to `Builder[T, WrappedArray[T]]` :)

lrytz:
@monkey-mas could you squash the two commits?

monkey-mas:
@lrytz sorry about it, fixed it :)

lrytz:
there's nothing to be sorry about :)

szeiger (approved):
LGTM. This method doesn't seem to be used at all but I agree in principle that `WrappedArray.newBuilder` should give you a builder for a `WrappedArray` (whereas `ArrayBuffer` is a builder for `ArrayBuffer`)

PR #4400, opened 2015-03-24
Can catch any expression

som-snytt:
`catch` accepts an arbitrary expression, which must conform
to `Function[Throwable, ?]`.

The previous transform was name-based:
```
scala> try 42 catch new {
     | def isDefinedAt(x: Any) = false
     | def apply(x: Any) = ()
     | }
warning: there was one feature warning; re-run with -feature for details
res0: AnyVal = 42

scala> try 42 catch 22
<console>:8: error: value isDefinedAt is not a member of Int
              try 42 catch 22
                           ^
```
Now:
```
scala> try 42 catch 22
                    ^
       error: type mismatch;
        found   : Int(22)
        required: Throwable => ?

scala> def npe: Int = throw null
def npe: Int

scala> val pf: PartialFunction[Throwable, Int] = { case _: NullPointerException => 42 }
val pf: PartialFunction[Throwable,Int] = <function1>

scala> try npe catch pf
val res0: Int = 42

scala> def err: Int = throw new Error()
def err: Int

scala> try err catch pf
java.lang.Error
  at err(<console>:1)
  at liftedTree1$1(<console>:1)
  ... 33 elided
```
As shown, if the handler is a `PartialFunction`, it is invoked
selectively, as though by calling:
```
pf.applyOrElse(e, (t: Throwable) => throw t)
```
Otherwise, the handler is applied to all thrown exceptions:
```
scala> def f(t: Throwable) = if (t.isInstanceOf[NullPointerException]) 27 else ???
def f(t: Throwable): Int

scala> try err catch f
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:345)
  at f(<console>:1)
  at $anonfun$res6$1(<console>:1)
  at $anonfun$res6$1$adapted(<console>:1)
  at liftedTree1$1(<console>:1)
  ... 33 elided
```
Note that applying a partial function literal in the form of a pattern-matching
anonymous function will use `applyOrElse`. The following example invokes
the `apply` method because the static type of the handler is not a
`PartialFunction`:
```
scala> val g: Throwable => Int = pf
val g: Throwable => Int = <function1>

scala> try err catch g
scala.MatchError: java.lang.Error (of class java.lang.Error)
  at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:344)
  at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:342)
  at $anonfun$1.applyOrElse(<console>:1)
[...]

PR #5617, opened 2016-12-27
SI-5365 Exhaustivity of extractors, guards, and unsealed traits.

sellout:
This builds on @retronym’s #4929.

It handles the unsealed trait case (via an additional `-Xlint:strict-unsealed-patmat` flag) and the extractor-only case (both raised by @durban).

It still defaults to lax checking by default. I don’t know what’s involved in testing it against the community build to see if we can do strict-by-default.


src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
```diff
@@ -888,7 +893,7 @@ trait MatchAnalysis extends MatchApproximation {
               // if uniqueEqualTo contains more than one symbol of the same domain
               // then we can safely ignore these counter examples since we will eventually encounter
               // both counter examples separately
-              case _ if inSameDomain => None
+              case _ if inSameDomain => Some(WildcardExample)
```
sellout: There is probably a better way to handle this case. I’m not sure what should ever hit the previous version of this – no tests failed, but I’m sure I’m overlooking something.

sellout:
I signed the CLA.

retronym:
Some tests have changed the output:

From https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/3948/consoleFull, you can run this to update the expectations ("checkfiles" in partest parlance), and then see if this is as you expect.

```
partest --update-check \
  test/files/neg/t1503.scala \
  test/files/neg/t5365.scala \
  test/files/neg/t5365b.scala
```

sellout:
@retronym Thanks. I must have been running the wrong test command locally, since I missed the failures. Hopefully one of the failures will point me in the right direction for fixing my confusion in `toCounterExample`.

sellout:
As @durban pointed out on typelevel/scala#129, these changes affect `catch` blocks as well (if `-Xlint:strict-unsealed-patmat` is used), but they don’t affect `PartialFunction`. IMO, a `catch` block should behave more like a `PartialFunction` than like an exhaustive pattern match. I’m working to make this change, but if anyone wants to jump in with a pointer, I wouldn’t mind.

som-snytt:
I'm not following the convo, but I had an old PR to treat catches exactly as PartialFunctions.

https://github.com/scala/scala/pull/4400

sellout:
@som-snytt Yeah, that sounds like exactly what I want. Should that PR be revived, or should I steal the part I need? (The parsing/hygiene stuff sounds a bit intimidating, so I doubt I’d pick up the whole PR, and @retronym suggested splitting it up anyway).

milessabin:
FYI: this has been [merged](https://github.com/typelevel/scala/blob/typelevel-readme/notes/2.12.1.md#exhaustivity-of-extractors-guards-and-unsealed-traits-pull5617-sellout) in Typelevel Scala 2.12.1.

sellout:
I see there is a failure on the “combined” check. Does that mean I should squash failing commits so that everything in the history builds cleanly?

SethTisue:
> Does that mean I should squash failing commits so that everything in the history builds cleanly?

yes please. as a rule, before final merge we ask that every commit be green (and that unnecessary merge commits be avoided)

durban:
Another strangeness (tested with Scalatl 2.12.1, `-Xlint:_` and `-Xstrict-patmat-analysis`):

```scala
scala> import scala.reflect.runtime.universe._
import scala.reflect.runtime.universe._

scala> (NoSymbol : Symbol) match {
     |   case NoSymbol => "nosym"
     |   case sym: Symbol if sym.isMethod => "meth"
     |   case x: Symbol => "fallback"
     | }
<console>:15: warning: match may not be exhaustive.
It would fail on the following input: (x: reflect.runtime.universe.Symbol forSome x not in NoSymbol)
       (NoSymbol : Symbol) match {
                 ^
```

However with a simple trait, it seems to work fine:

```scala
scala> trait Foo
defined trait Foo

scala> val x: Foo = new Foo {}
x: Foo = $anon$1@10bd5fb8

scala> (x : Foo) match {
     |   case `x` => "x"
     |   case foo: Foo if scala.util.Random.nextBoolean() => "rnd"
     |   case foo: Foo => "fallback"
     | }
res6: String = x

scala>
```

sellout:
@durban Thanks. This is an interesting case. It only happens with _both_ flags enabled, unlike the other ones that only required `-Xlint:strict-unsealed-patmat`.

sellout:
@durban `: Symbol` uses the `universe.SymbolTag.unapply` so it’s an extractor, which means it doesn’t help exhastivity. If I swap in `: SymbolApi` the warning goes away. But why does a type pattern ever use an extractor?

durban:
@sellout Ok, thanks, that makes (some) sense. So the behavior (while somewhat surprising) seems to be correct. Those cases indeed aren't exhaustive, so the compiler is right.

As to why a (syntactically) type pattern uses an extractor, this comment on `ClassTag#unapply` seems to be the answer:
```scala
  /** A ClassTag[T] can serve as an extractor that matches only objects of type T.
   *
   * The compiler tries to turn unchecked type tests in pattern matches into checked ones
   * by wrapping a `(_: T)` type pattern as `ct(_: T)`, where `ct` is the `ClassTag[T]` instance.
   * Type tests necessary before calling other extractors are treated similarly.
   * `SomeExtractor(...)` is turned into `ct(SomeExtractor(...))` if `T` in `SomeExtractor.unapply(x: T)`
   * is uncheckable, but we have an instance of `ClassTag[T]`.
   */
```

So, if I understand correctly, this is to work around type erasure, and make more type patterns work (somewhat) safely. And it seems, that the reflection/macro API uses this feature extensively. Fair enough.

sellout:
@durban Well, your comment just helped me understand it better than the docs I’ve been reading all morning 😆

However, I wonder if we can avoid using the extractor in the case where the type annotation is the same as the scrutinee’s type? Since then you don’t really need to use the unapply. The type doesn’t change – i.e., `(x: Q) match { ...; case y: Q => ??? }` could be rewritten to `(x: Q) match { …; case y => ??? }` safely, no?

sellout:
In b4 @paulp explains why that doesn’t work.

paulp:
@sellout 🤔
screen shot 2017-02-27 at 12 37 30 pm

sellout:
@paulp http://www.urbandictionary.com/define.php?term=inb4

paulp:
TIL.

retronym:
@sellout The typechecker [rewrites](https://github.com/scala/scala/blob/44dac961782aa1193493c181e8423d7751c013ee/src/compiler/scala/tools/nsc/typechecker/PatternTypers.scala#L332-L376) uncheckable type tests as class tag extractor calls before the pattern matcher reasons about whether a type test is redundant (due to the type of the scrutinee and also the knowledge that previous cases haven't matched). So the pattern matcher just sees an opaque extractor pattern and emits the `unapply` call.

In other words (inb4 @paulp again; TIAL), "an artifact of the implementation"

src/compiler/scala/tools/nsc/transform/patmat/MatchAnalysis.scala
```diff
@@ -761,6 +765,7 @@ trait MatchAnalysis extends MatchApproximation {
       def chop(path: Tree): List[Symbol] = path match {
         case Ident(_) => List(path.symbol)
         case Select(pre, name) => chop(pre) :+ path.symbol
+        case Apply(fun, args) => chop(fun) :+ path.symbol
```
retronym, line 770: Could you please rebase this change into a separate commit to help understand its impact (ie, is a no-op without the new flag enabled?) and comment this line? I'm assuming we know must account for getting an extractor call in the tree `path` under the new option.

test/files/neg/patmatexhaust.scala
```diff
@@ -56,7 +56,7 @@ class TestSealedExhaustive { // compile only
       case Ga =>
     }
 
-  def ma6()  = List(1,2) match { // give up
```
retronym, line 59: What change caused this test case to progress?

som-snytt:
Isn't B4 a beginner classification in the Scala Center's update of the Odersky skill levels? The number indicates how many times you've retaken the Coursera.

durban:
As a sidenote/question: pattern exhaustivity of destructuring `val`s (e.g., `val Some(x) = Option.empty[Int]`) doesn't seem to be checked (either with or without this PR). I guess this is because the `match` this code desugars to, seems to contain an `@unchecked` annotation. What's the reason for inserting this `@unchecked`?

milessabin:
@sellout were you able to make any progress on @durban's issue?

adriaanm:
Happy to reschedule for 2.12.4 once feel we can push this forward in time.

SethTisue:
closing for inactivity. anyone want to try and carry this work forward...?

milessabin:
I'll pick this one up too.

adriaanm:

SethTisue:
happy ending at https://github.com/scala/scala/pull/9140

som-snytt:
Two follow-ups to the thread are -Xlint enables check on "destructuring vals" and the "catch takes a function" is almost in https://github.com/scala/scala/pull/4400 but I haven't looked at whether checks are applied.

PR #5618, opened 2016-12-28
Fix indent and comment of ArrayOps.scala

monkey-mas:
1. There should be space at the beginning of the `final class` definition.
  2. ofXxx is a subclass of ArrayOps (of a particular type, such as `Byte`s)
  3. Unnecessary empty lines are removed

### DIFF
--- a/src/library/scala/collection/mutable/ArrayOps.scala
+++ b/src/library/scala/collection/mutable/ArrayOps.scala
@@ -177,9 +177,7 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
     (a1, a2, a3)
   }
 
-
   def seq = thisCollection
-
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/library/scala/collection/mutable/ArrayOps.scala
```diff
@@ -202,7 +202,7 @@ object ArrayOps {
   }
 
   /** A class of `ArrayOps` for arrays containing `byte`s. */
-final class ofByte(override val repr: Array[Byte]) extends AnyVal with ArrayOps[Byte] with ArrayLike[Byte, Array[Byte]] {
+  final class ofByte(override val repr: Array[Byte]) extends AnyVal with ArrayOps[Byte] with ArrayLike[Byte, Array[Byte]] {
```
janekdb: In my experience whitespace edits are accepted when accompanied by more significant changes and not otherwise.

som-snytt: The class comments should read "A subclass...", and the element type should use the correct name such as "containing `Byte`s." Also there are extraneous empty lines around line 180.

adriaanm: I'd be happy to merge this once these suggested changes are in.

monkey-mas: Thanks for your feedback!! :-) Fixed my commit, please review again!

adriaanm:
Thanks!

PR #5619, opened 2016-12-29
[WIP] draft of switch annotation for Strings

martijnhoekstra:
Generate a lookupswitch for matching over strings

This is a bit faster for most reasonable cases. I've benchmarked the throughput of this version against the current version for short strings and longer strings:

```

short strings:
      ||   1  |   2  |   4  |   8  |  16  |  32  |  64  |  128 |  256 |  512 | 1024 |
_____________________________________________________________________________________
   1  ||  1,38|      |      |      |      |      |      |      |      |      |      |
   2  ||  1,14|  2,36|      |      |      |      |      |      |      |      |      |
   4  ||  1,25|  2,09|  3,62|      |      |      |      |      |      |      |      |
   8  ||  1,16|  2,18|  2,90|  6,50|      |      |      |      |      |      |      |
  16  ||  0,98|  1,82|  3,70|  6,06|  7,47|      |      |      |      |      |      |
  32  ||  0,86|  1,48|  3,10|  4,88|  6,14| 19,12|      |      |      |      |      |
  64  ||  0,81|  1,32|  2,51|  4,46|  5,44| 16,41| 48,69|      |      |      |      |
  128 ||  0,56|  1,14|  2,24|  4,91|  7,03| 22,27| 44,33| 49,57|      |      |      |
  256 ||  0,85|  1,24|  2,18|  4,03|  5,18| 14,48| 45,58| 30,71|136,45|      |      |
  512 ||  0,57|  0,83|  1,48|  2,58|  3,10|  9,60| 25,53| 20,35| 96,54|188,05|      |
 1024 ||  0,55|  0,78|  1,38|  2,45|  3,09|  9,35| 23,69| 19,56| 96,49|188,98|326,71|
long strings:
      ||   1  |   2  |   4  |   8  |  16  |  32  |  64  |  128 |  256 |  512 | 1024 |
_____________________________________________________________________________________
   1  ||  1,39|      |      |      |      |      |      |      |      |      |      |
   2  ||  0,99|  2,38|      |      |      |      |      |      |      |      |      |
   4  ||  0,98|  2,17|  2,82|      |      |      |      |      |      |      |      |
   8  ||  0,92|  1,84|  2,85|  7,02|      |      |      |      |      |      |      |
  16  ||  1,08|  1,41|  2,50|  4,56|  8,95|      |      |      |      |      |      |
  32  ||  1,08|  1,26|  2,37|  4,11|  8,22| 19,64|      |      |      |      |      |
  64  ||  0,68|  1,15|  2,04|  3,68|  7,35| 16,32| 39,08|      |      |      |      |
  128 ||  0,76|  1,06|  2,03|  3,82|  6,87| 17,56| 33,01| 67,32|      |      |      |
  256 ||  0,73|  1,03|  1,51|  3,04|  5,49| 11,24| 34,52| 68,76|153,68|      |      |
  512 ||  0,52|  0,66|  1,14|  2,12|  4,81|  9,43| 24,84| 45,60| 90,80|187,70|      |
 1024 ||  0,41|  0,65|  0,99|  2,19|  3,56|  8,62| 18,14| 36,38| 92,09|168,74|457,05|
```

this show the relative throughput of the switch version vs the match version. The rows are the number of cases in the switch, and the columns the nth branch taken. That shows a modest increase in throughput for cases that seem common (between 4 and 16 cases).

There is more significant performance gain in the "more exotic but not bonkers" cases, e.g. a 4 to 5 times throughput increase in taking the 8th branch of 32 branches.

If branch prediction has any effect on the artificial workload, it can be expected to increase the speed up.

I've gotten a little lost in the tree makers, symbols and trees, so the way I did this is probably far from ideal: this is not boyscout rule-proof. 

This also doesn't implement hash-collision handling yet.


sjrd:
Repeating what I said somewhere else about this issue (probably scala-dev?): I would very much see this tackled as a backend-specific optimization, i.e., in the backend. That's because it *is* very much backend-specific. I haven't measured but I strongly expect this new translation to be significantly slower in Scala.js.

martijnhoekstra:
I understood that earlier comment to be about the case class encoding only @sjrd, sorry for the misunderstanding.

I'll try to move it to the backend.

I'm entirely unfamiliar with benchmarking js. What js engines should I test on, and what benchmark tools are commonly used?

sjrd:
> I'm entirely unfamiliar with benchmarking js. What js engines should I test on, and what benchmark tools are commonly used?

Node.js (the default runner of Scala.js) is a reasonable platform to benchmark on. Regarding benchmarking framework, unfortunately there isn't anything standard in the JS ecosystem (unlike JMH or ScalaMeter on the JVM). The normal thing to do is to use a stupid loop, and discard the first few iterations as warmup. You can reuse the setup in https://github.com/sjrd/scalajs-benchmarks, which uses the following looping logic: https://github.com/sjrd/scalajs-benchmarks/blob/master/common/js/src/main/scala/org/scalajs/benchmark/Benchmark.scala

martijnhoekstra:
@sjrd I'm in a bit of a pickle how to move this to the backend. By the time the compiler gets to the backend, I can't find any way to relate it back to the original match. It seems unwise to try to detect eligible if chains and make them in to a switch as an optimization. Do you have any suggestions on what to try?

sjrd:
Would it perhaps make sense to do something similar to what happens for `Int` switches? I.e., the `patmat` phase identifies that something is an optimizable match on strings, and when that is the case, *leaves it* as a `Match` node. It would then trickle down the rest of the pipeline until the back-end unchanged. The back-end can then trivially recognize the `Match` and compile it down in the most efficient way, given the platform.

martijnhoekstra:
Thanks, I'll give that a try.

PR #5621, opened 2016-12-30
SI-9936 LinearSeqOptimized.indexWhere

som-snytt:
Also suffered from the negative `from` bug.

Prefer `math.max` to avoid `RichInt`.

Follows up https://github.com/scala/scala/pull/5421


som-snytt:
Would be nice if `i max 0` were rewritten to `math.max(i, 0)` rewritten to `if (i < 0) 0 else i`. Also if the test framework checked boundary inputs, etc.

sjrd:
> Would be nice if `i max 0` were rewritten to `math.max(i, 0)` rewritten to `if (i < 0) 0 else i`.

It's not impossible. The Scala.js optimizer does it. But that kind of assumes whole program knowledge, or quite close to it.

lrytz (approved):

lrytz:
Thanks!

lrytz:
Happy new year!

PR #5622, opened 2016-12-31
Improved error messages for identically named, differently prefixed types

edmundnoble:
I'm betting this needs a lot of feedback; first compiler PR. The purpose of this PR is to make disambiguating concrete types easier; for example, from type-diagnostics.check:
```
- found   : scala.collection.Set[String]
- required: scala.collection.immutable.Set[String]
+ found   : Set[String] (in scala.collection)
+ required: Set[String] (in scala.collection.immutable)
```

This PR does not change the error message if either `found` or `required` is an abstract type or type parameter, because I'm not sure it helps as much; in that case it is more likely to be a unification error and not a misplaced import, but if replacing the standard error mechanisms for both cases would help as well I can amend it.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
@@ -11,6 +11,7 @@ import scala.collection.mutable.ListBuffer
 import scala.util.control.Exception.ultimately
 import symtab.Flags._
 import PartialFunction._
+import scala.annotation.tailrec
 
 /** An interface to enable higher configurability of diagnostic messages
  *  regarding type errors.  This is barely a beginning as error messages are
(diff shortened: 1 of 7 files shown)


test/files/neg/t2102.check
```diff
@@ -1,6 +1,6 @@
 t2102.scala:2: error: type mismatch;
- found   : java.util.Iterator[Int]
- required: scala.collection.Iterator[_]
+ found   : Iterator[Int] (in java.util)
+ required: Iterator[_] (in scala.collection)
```
ktoso: Firstly: very cool idea!
Could we also align the `(` segments to easier visually identify where the mismatch is? E.g. what I often find:

```scala
Thing[Bla]    (in akka.stream.scaladsl)
Thing[Blabla] (in akka.stream.javadsl)
```

Which makes it even easier to spot "aha! the diff is the javadsl/scaladsl" and not that the type parameter was the problem (it could be fine, e.g. because Bla extended Blabla etc).

ktoso: Additional idea would be colour coding the different part of the path... No idea if this is something Scala team would want or not though. The aligning seems already very useful :)

fommil:
this looks pretty cool. Can we be sure that it won't introduce any regressions in name reporter for the interactive compiler?

test/files/neg/t2102.check
edmundnoble: Love the first idea at least for sure. I think it definitely deserves including.

edmundnoble:
@fommil I'm not sure what exactly that would entail; this should only affect callsites of `foundReqMsg` which is only called during error reporting as far as I know.

test/files/neg/t2102.check
adriaanm: Based on this idea, I think a small tweak to the current format could be as effective, and more compact:
```
found   :        java.util.Iterator[Int]
required: scala.collection.Iterator[_]
```

dwijnand: For @ktoso's example that would mean

```scala
found:   akka.stream.scaladsl.Thing[Bla]
required: akka.stream.javadsl.Thing[Blabla]
```

ktoso: That works well too 👍  Actually probably even better :)

edmundnoble: I think that's less readable. Most of the point of this PR is to bring the identical parts of the name all the way to the left, so that it's easily recognizable what the problem is. This way, the identical parts of the name are pushed a varying amount to the right.

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -274,19 +275,51 @@ trait TypeDiagnostics {
     if (AnyRefTpe <:< req) notAnyRefMessage(found) else ""
   }
 
+  def finalOwners(tpe: Type): Boolean = (tpe.prefix == NoPrefix) || recursivelyFinal(tpe)
+
+  @tailrec
+  final def recursivelyFinal(tpe: Type): Boolean = {
```
lrytz, line 281: could be a local method within `finalOwners`

lrytz (commented):
I'm starting to like it.

SethTisue:
thanks Edmund, nice work & nice improvement.

PR #5623, opened 2017-01-02
Add substitute methods to <:< =:= making these Liskov/Leibniz

johnynek:
the `=:=` and `<:<` classes lack a substitute method which works on `F[A]` to `F[B]`. This is easy to add, but without it, we can't do basic things like:

```
A =:= B => Map[A, V] =:= Map[B, V]
A <:< B => List[A] <:< List[B]
```
without these we can resort to casts, but we shouldn't have to. Scalaz has a solution to this, but can't interop fully with the standard library since the standard library lacks `substitute`. Similarly a PR to cats was proposed to address the issue:

https://github.com/typelevel/cats/pull/1178

Since these classes are abstract, adding a few methods to remedy this seems like a very minimal change to address this (minor) defect.

### DIFF
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -464,8 +464,27 @@ object Predef extends LowPriorityImplicits {
    * @group type-constraints
    */
   @implicitNotFound(msg = "Cannot prove that ${From} <:< ${To}.")
-  sealed abstract class <:<[-From, +To] extends (From => To) with Serializable
-  private[this] final val singleton_<:< = new <:<[Any,Any] { def apply(x: Any): Any = x }
+  sealed abstract class <:<[-From, +To] extends (From => To) with Serializable {
+    /** substitute the To in the type F[To] to From
+     */
+    def substitute[F[-_]](ft: F[To]): F[From]
+    /** create a new evidence for a covariant type F[_]
+     */
+    def liftCo[F[+_]]: <:<[F[From], F[To]] = {
+      type G[-T] = F[T] <:< F[To]
+      substitute[G](implicitly[F[To] <:< F[To]])
+    }
+    /** create a new evidence for a contravariant type F[_]
+     */
+    def liftContra[F[-_]]: <:<[F[To], F[From]] = {
+      type G[-T] = F[To] <:< F[T]
+      substitute[G](implicitly[F[To] <:< F[To]])
+    }
+  }
+  private[this] final val singleton_<:< = new <:<[Any,Any] {
+    def apply(x: Any): Any = x
+    def substitute[F[-_]](ft: F[Any]): F[Any] = ft
+  }
   // The dollar prefix is to dodge accidental shadowing of this method
   // by a user-defined method of the same name (scala/bug#7788).
   // The collections rely on this method.
@@ -478,8 +497,27 @@ object Predef extends LowPriorityImplicits {
    * @group type-constraints
    */
   @implicitNotFound(msg = "Cannot prove that ${From} =:= ${To}.")
-  sealed abstract class =:=[From, To] extends (From => To) with Serializable
-  private[this] final val singleton_=:= = new =:=[Any,Any] { def apply(x: Any): Any = x }
+  sealed abstract class =:=[From, To] extends (From => To) with Serializable {
+    /** substitute the From in the type F[From] to To
+     */
+    def substitute[F[_]](ff: F[From]): F[To]
+    /** create an instance for a type constructor F[_]
+     */
+    def liftTo[F[_]]: =:=[F[From], F[To]] = {
+      type G[T] = F[From] =:= F[T]
+      substitute[G](implicitly[F[From] =:= F[From]])
+    }
+    /** flip the order of the type paraemeters
+     */
+    def flip: =:=[To, From] = {
+      type G[T] = T =:= From
+      substitute[G](implicitly[From =:= From])
+    }
+  }
+  private[this] final val singleton_=:= = new =:=[Any,Any] {
+    def apply(x: Any): Any = x
+    def substitute[F[_]](ff: F[Any]): F[Any] = ff
+  }
   /** @group type-constraints */
   object =:= {
      implicit def tpEquals[A]: A =:= A = singleton_=:=.asInstanceOf[A =:= A]
--- a/test/files/pos/leibniz_liskov.scala
+++ b/test/files/pos/leibniz_liskov.scala
@@ -1,0 +1,33 @@
+object LeibnizLiskov {
+  val isInt = implicitly[Int =:= Int]
+
+  isInt.substitute(List(1, 2, 3))
+
+  def convertInt1[T, U](l: List[T])(ev: T =:= U): List[U] =
+    ev.substitute(l)
+
+  def convertInt2[T, U](l: List[T])(ev: T =:= U): List[U] =
+    ev.liftTo[List](l)
+
+  val isIntSub = implicitly[Int <:< Int]
+
+  isIntSub.liftCo(List(1, 2, 3))
+
+  def convertIntSub[T, U](l: List[T])(ev: T <:< U): List[U] =
+    ev.liftCo[List](l)
+
+  trait Consumes[-T] {
+    def consume(t: T): Unit = ()
+  }
+
+  isIntSub.liftContra(new Consumes[Int] { }): Consumes[Int]
+  isIntSub.substitute(new Consumes[Int] { }): Consumes[Int]
+
+  def convertConsume1[U, T](c: Consumes[T])(ev: U <:< T): Consumes[U] =
+    ev.substitute(c)
+
+  def convertConsume2[U, T](c: Consumes[T])(ev: U <:< T): Consumes[U] = {
+    val ev2 = ev.liftContra[Consumes]
+    ev2(c)
+  }
+}


S11001001:
See also #3992 for further discussion.

e: And #5359.

SethTisue:
this must target 2.13, not 2.12, because binary compatibility

johnynek:
@S11001001 thanks for pointing that out. I hadn't seen it.

The difference here:

1. no casts, so it is a bit easier to see that nothing funky is happening.
2. no overloading of andThen and compose to avoid any chance of breaking any source compatibility with the name collision with functions. Those can be generated from subst without casts in the case that it is needed.
3. I believe this is binary compatible.

johnynek:
@SethTisue how is this not binary compatible? Also, are there not mima tests for compatibility in the 2.12 series? Shouldn't they fail if this is not compatible?

adriaanm:
Yes, mima not failing seems to be a bug :-/

We can't add e.g. `def substitute[F[-_]](ft: F[To]): F[From]` to `class <:<` in, say, 2.12.2, because code that calls that method would crash when running on 2.12.1. The goal of our compatibility policy is that you don't need to worry about which 2.12.x you're using to compile/run your code. To be precise, we promise both forwards and backwards compatibility. You're right that this is backwards compatible.

milessabin:
This seems like a good candidate for TLS 2.12.x if it can't be merged immediately here.

smarter:
> This seems like a good candidate for TLS 2.12.x if it can't be merged immediately here.

I thought TLS was supposed to be binary-compatible with upstream Scala?

milessabin:
Could `substitute` be added in a binary compatible way as an extension method?

dwijnand:
> Could `substitute` be added in a binary compatible way as an extension method?

Yes.

But the issue is forward compatibility. All code/libraries built with 2.12.x must work with scala-library 2.12.0.

milessabin:
Sure, but an extension method could be provided via an additional library for earlier 2.12.x releases. I think that would be sufficient to be in the spirit of TLS's binary compatibility goals.

dwijnand:
What would provide the additional library? Would it a be a manual step? If it's a manual step, then they could also just update to the minor version that includes it in the standard library.

I think the purpose/spirit of the forward compatibility restriction is that it removes the need to worry about which minor version one is using when including a Scala library - no additional, manual steps required.

milessabin:
> What would provide the additional library? Would it a be a manual step?

A separate compat library added as a manual step.

> If it's a manual step, then they could also just update to the minor version that includes it in the standard library.

Am I understanding you correctly here? That can't happen until 2.13.0 is released.

dwijnand:
I mean if a manual step is to be required, then why not just stick it in a standard library and make that manual step updating the minor version of scala instead.

As an example, if you ship TLS 2.12.2 with this, and cats 0.10 uses it, then anything using cats (including transitively!) must also add your compat library. Either because they know or because it bombed at runtime (hopefully in their tests as opposed to in production!).

milessabin:
Gotcha. I was thinking along the lines of "available by default in TLS 2.12.1 or greater, available via a compat library in LBS or TLS 2.12.0".

> As an example, if you ship TLS 2.12.2 with this, and cats 0.10 uses it, then anything using cats (including transitively!) must also add your compat library. Either because they know or because it bombed at runtime (hopefully in their tests as opposed to in production!).

Agreed. That's the trade off which would have to be made. I can see that maybe this feature is too minor to warrant that.

dwijnand:
Btw, I think the restriction is too archaic - legacy of the days of `libs` directories, defining classpaths with ant, and such.

If this hypothetically shipped in scala-library-2.12.2.jar, and cats 0.10 depended on it, in my opinion in sbt you should transitively get scala-library 2.12.2 (*note*, I found out this isn't currently true: https://github.com/sbt/sbt/issues/2894).

So, what I'm saying is, perhaps it's time to lift this restriction. I would consider creating a SIP to do so.

sjrd:
@dwijnand I agree. I have always been skeptic about the forward binary compatibility requirement. I do not really understand its value in an era where virtually all classpaths are constructed through Ivy dependency resolution.

Note that, using a different hat (Scala.js author instead of Scala user), I have to disagree with myself ^^ It is actually significantly valuable to Scala.js that the Scala stdlib is forward binary compatible. Thanks to this property, we can easily publish *old* Scala.js versions for new minor versions of Scala. Because an old version of Scala.js is published with exactly one implementation of the stdlib, which is that of the old Scala version that was available when it was published. Thanks to forward binary compatibility, that old stdlib from 2.12.1 can readily be used with 2.12.2 as soon as we publish the Scala.js 0.6.14 compiler for 2.12.2. If it were not forward binary compatible, the old 2.12.1 stdlib compiled and published with Scala.js 0.6.14 would not be able to (completely) support 2.12.2. We would have to wait for 0.6.15 to get full support for 2.12.2.

All of that rambling to say that, @dwijnand, if you start working on a SIP about this, I would like to be involved, to clearly expose the trade-offs wrt. Scala.js versioning.

johnynek:
Just to chime in, I'm actually +1 on the forward and backward compat story if we want to encourage many scalas (scalajs, scala-native, typesafe, dotty?).

The only bummer is that the cycle is so long, fixing the standard library takes much patience. It's a real bummer the previous attempt was closed (#3992) when it was the same outcome as this.

adriaanm:
Agreed on shortening the cycle. 2.13 should only have 4 quarterly
milestones this year, with RC1 in Q1 2018. I'd also like to shrink
scala-library (ideally to classes directly in the scala package,
annotations and the collections -- everything else would be an optional
dependency with only backwards compat).

adriaanm:
To make that more concrete, here's a sketch of what I have in mind: https://github.com/scala/scala/compare/a5864ad...adriaanm:modularize-v2?expand=1

dwijnand:
@milessabin Thinking more about this, I think the solution is to ship this in the Predef of Scala 2.13.0+. And if you want it sooner then that create that mini compat library (using extension methods or whatever) for 2.12/2.11. Avoids forward compatibility both here and in typelevel/scala.

johnynek:
@adriaanm huge +1 to more modularization. The compiler + stdlib itself should be as small as conveniently possible in my opinion. And then we can use backwards-only compatibility for libraries.

adriaanm:
Gah, changing the base doesn't rebase, so I'm going to leave that to you.

johnynek:
@adriaanm changes the PR base seems to still cleanly apply.

Do I need to rebase to get the a test of the resulting PR?

SethTisue:
> Do I need to rebase to get the a test of the resulting PR

yes

lrytz:
Could you add some tests to this PR? This is a good area for testing implicit search. Ideally both `pos` (or maybe `run`?) and `neg` tests.

johnynek:
@lrytz I will try to find some time to update the PR. I don't think this changes implicit resolution to add methods does it? Do you feel the current tests on `<:<` and `=:=` are not sufficient?

This code is actually a rare case where if it compiles it is actually correct, since all of these instances are `identity` under the hood, and there is no unsafe code (unless of course, this touches on an area where scala is unsound, but I don't think it does).

lrytz:
@johnynek I haven't looked at existing tests for `<:<` / `=:=`, let's assume they're fine :) It would be good to have some tests for type inference with the new methods added in this PR.

sjrd:
> This code is actually a rare case where if it compiles it is actually correct

Which means it's probably useful to test that what *shouldn't* compile indeed doesn't, i.e., `neg` tests. It's also still valuable to assert that desirable snippet that ought to compile indeed do.

lrytz:
```
scala> def f[T, U](ev: T =:= U) = ev.lift
f: [T, U](ev: T =:= U)Nothing =:= Nothing

scala> def f[T, U](ev: T =:= U) = ev.lift[List]
f: [T, U](ev: T =:= U)List[T] =:= List[U]

scala> def f[T, U](ev: T =:= U) = ev.flip
f: [T, U](ev: T =:= U)U =:= T
```
etc..

johnynek:
Okay, I added some compilation tests.

If anyone has any specific tests they want to see added, please let me know. Note, this does not change implicit resolution at all and ultimately all the methods are identity, so I guess any tests will be checking for compiler bugs so perhaps someone else can suggest good tests there.

I tried to search for tests for `=:=` and `<:<` but I didn't find anything focused on them.

@adriaanm how does this look now?

szeiger:
Design and impl LGTM but I'd prefer JUnit tests over partest. partest is usually only needed for testing the compiler, library features can be tested more efficiently with JUnit.

johnynek:
@szeiger can you suggest some tests you'd like to see?

I feel a bit silly testing identity functions, which these things are.

Should I write something like:
```scala
val  ev = implicitly[Int =:= Int]
assert(ev(1) == 1) // this method already exists in =:= but I guess we should add more tests?
assert(ev.flip(1) == 1)
assert(ev.substitute(List(1, 2, 3)) == List(1, 2, 3)) // there is only one instance and substitute is identity
```

S11001001:
@szeiger The interesting tests are all does-compile/does-not-compile tests, as @sjrd suggests. Can `neg` tests be implemented with JUnit?

szeiger:
@johnynek I can't think of any additional test, my suggestion is to port the existing tests over to JUnit. If they are only supposed to compile but not actually be run, just put them into a test class but don't annotate the test method with `@Test`. Negative tests are possible in JUnit with something like https://github.com/slick/slick/blob/master/slick-testkit/src/main/scala/com/typesafe/slick/testkit/util/ShouldNotTypecheck.scala but I don't think we currently have any such tests in scala/scala.

szeiger:
@johnynek Do you have any more changes here. We're planning to close the queue for M2 until the end of next week.

johnynek:
I don't have any changes. I just haven't had the time to rework the tests.

Are there any changes you want to see other than moving the tests to junit?

szeiger:
The commits should be squashed but other than that I don't have any changes to request. Since it has `neg` tests and we don't have a precedent for doing those with JUnit I am willing to merge this as-is if you don't have the time to change it but JUnit would be preferred.

johnynek:
Great. I'll try to find some time today or tomorrow to do the squash or is
GitHub's ability to do a squash merge acceptable? (Just change the option
when you click merge).

johnynek:
@szeiger I've squashed. Can this be merged?

lrytz (approved):

fommil:
Heh, I found myself looking at <:< yesterday wondering why the hell this don't exist, and doing a cast. It finally clicked with me today that you'd done this, thank you!

milessabin:
This closes https://github.com/scala/bug/issues/4040 as well I think ...

PR #5624, opened 2017-01-03
Fix documentation of immutable.Queue

Jasper-M:
`enqueue` appends elements to the `Queue`, it doesn't prepend them.


lrytz:
Thanks!

PR #5625, opened 2017-01-04
Bundle newer scala-parser-combinators (1.0.5)

gourlaysama:


SethTisue (approved):

SethTisue:
Because 2.11.9 is coming up very rapidly, and since it is the last planned 2.11.x release, and since 1.0.5 is new and relatively untried, I agree with the choice to leave 2.11 alone and just target 2.12.

PR #5627, opened 2017-01-05
Do not instantiate TrieMap root updater for each instance

rovarga:
AtomicReferenceFieldUpdater is costly to instantiate, involving
call stack introspection, reflection and privileged operations.

It is recommended to pay this price only once, and then use an
updater for all objects.

Let's create a field in TrieMap companion object to hold an updater
and use that in the constructor. Unfortunately the companion object
is barred from instanting the updater by itself, as the updater
references a TrieMap private field. To overcome this, we define
a method by which the companion can request an updater from an
otherwise-defunct TrieMap instance.

This brings down the time to instantiate an empty TrieMap in local
testing from ~580ns to ~60ns, i.e. an improvement of one order
of magnitude.

Signed-off-by: Robert Varga <nite@hq.sk>

### DIFF
--- a/src/library/scala/collection/concurrent/TrieMap.scala
+++ b/src/library/scala/collection/concurrent/TrieMap.scala
@@ -628,7 +628,7 @@ private[concurrent] case class RDCSS_Descriptor[K, V](old: INode[K, V], expected
  *  @since 2.10
  */
 @SerialVersionUID(0L - 6402774413839597105L)
-final class TrieMap[K, V] private (r: AnyRef, rtupd: AtomicReferenceFieldUpdater[TrieMap[K, V], AnyRef], hashf: Hashing[K], ef: Equiv[K])
+final class TrieMap[K, V] private (r: AnyRef, rtupd: AtomicReferenceFieldUpdater[TrieMap[_, _], AnyRef], hashf: Hashing[K], ef: Equiv[K])
 extends scala.collection.concurrent.Map[K, V]
    with scala.collection.mutable.MapLike[K, V, TrieMap[K, V]]
    with CustomParallelizable[(K, V), ParTrieMap[K, V]]
@@ -644,7 +644,7 @@ extends scala.collection.concurrent.Map[K, V]
 
   def this(hashf: Hashing[K], ef: Equiv[K]) = this(
     INode.newRootNode,
-    AtomicReferenceFieldUpdater.newUpdater(classOf[TrieMap[K, V]], classOf[AnyRef], "root"),
+    TrieMap.rootupdater,
     hashf,
     ef
   )
@@ -668,7 +668,7 @@ extends scala.collection.concurrent.Map[K, V]
 
   private def readObject(in: java.io.ObjectInputStream) {
     root = INode.newRootNode
-    rootupdater = AtomicReferenceFieldUpdater.newUpdater(classOf[TrieMap[K, V]], classOf[AnyRef], "root")
+    rootupdater = TrieMap.rootupdater
 
     hashingobj = in.readObject().asInstanceOf[Hashing[K]]
     equalityobj = in.readObject().asInstanceOf[Equiv[K]]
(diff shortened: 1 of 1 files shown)


rovarga:
@axel22 I think you'll like this one

Ichoran:
I'm not sure allocating a bunch of empty trie maps is the case we should be optimizing.  How does it fare in "normal" use (i.e. putting some things into the map)?

rovarga:
I don't have a benchmark on actual operations, but they should not be affected in any way, or maybe improved by JIT seeing a single object only.

I agree that just instantiating maps is not really useful benchmark, but it shows how much time in a simple create&fill operation can be spent in the create part -- 0.5ms is a lot of time.

The cost can be side-stepped by getting hold of an initial empty map and taking snapshots from it -- but that is not something casual users would do.

src/library/scala/collection/concurrent/TrieMap.scala
```diff
@@ -943,12 +943,19 @@ extends scala.collection.concurrent.Map[K, V]
 
   override def stringPrefix = "TrieMap"
 
+  // This is needed to instantiate the updater in this class, because it is accessing
+  // a private field, and then leak it to the companion object for storage
+  private def updater(guard: TrieMap.UpdaterGuard): AtomicReferenceFieldUpdater[TrieMap[_, _], AnyRef] = {
```
Ichoran: Why is the guard needed?  It prevents anyone but the companion from calling this method, but on the other hand why complicate things just to prevent it?

rovarga: That is the only reason. I did not leave a door open for anyone to muck with the field, intentionally or otherwise.

Ichoran: But it's already private.  Why is that not enough?

rovarga: Right, but I think it still can be accessed from Java -- anyway I can remove it, if leaking the updater inside the JVM is not a problem.

Ichoran: This doesn't work anyway because TrieMap.UpdaterGuard is public when viewed from Java.

rovarga: I have not checked that part. Fair enough, I will remove it.

axel22:
Apparently, the test fails due to binary compatibility - I guess this should target 2.13.

The change itself looks good to me. As Rex said, it is maybe not a common case, but if someone is creating a lot of empty maps, definitely helps a lot.

adriaanm:
```
Found 1 binary incompatibilities (4 were filtered out)
 * method
   this(java.lang.Object,java.util.concurrent.atomic.AtomicReferenceFieldUpdater,scala.util.hashing.Hashing,scala.math.Equiv)Unit
   in class scala.collection.concurrent.TrieMap does not have a correspondent
   in current version
```

adriaanm:
Weird, that mima failure seems like a bug. /cc @szeiger

szeiger:
Looks harmless judging by the source code. I'll take a closer look.

src/library/scala/collection/concurrent/TrieMap.scala
```diff
@@ -943,12 +943,17 @@ extends scala.collection.concurrent.Map[K, V]
...
+  // a private field, and then leak it to the companion object for storage
+  private def updater(): AtomicReferenceFieldUpdater[TrieMap[_, _], AnyRef] =
+    AtomicReferenceFieldUpdater.newUpdater(classOf[TrieMap[_, _]], classOf[AnyRef], "root")
 }
 
-
 object TrieMap extends MutableMapFactory[TrieMap] {
   val inodeupdater = AtomicReferenceFieldUpdater.newUpdater(classOf[INodeBase[_, _]], classOf[MainNode[_, _]], "mainnode")
 
+  private val rootupdater = new TrieMap[Any, Any](null, null, null, null).updater()
```
szeiger, line 955: The binary incompatibility is real:

```
$ javap -cp /Users/szeiger/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.12.0.jar scala.collection.concurrent.TrieMap | grep "TrieMap("
  public scala.collection.concurrent.TrieMap(scala.util.hashing.Hashing<K>, scala.math.Equiv<K>);
  public scala.collection.concurrent.TrieMap();

$ javap -cp /Users/szeiger/code/scala/build/pack/lib/scala-library.jar scala.collection.concurrent.TrieMap | grep "TrieMap("
  public scala.collection.concurrent.TrieMap(java.lang.Object, java.util.concurrent.atomic.AtomicReferenceFieldUpdater<scala.collection.concurrent.TrieMap<?, ?>, java.lang.Object>, scala.util.hashing.Hashing<K>, scala.math.Equiv<K>);
  public scala.collection.concurrent.TrieMap(scala.util.hashing.Hashing<K>, scala.math.Equiv<K>);
  public scala.collection.concurrent.TrieMap();
```

This here is the culprit. This constructor used to be JVM-private but calling it from the companion object makes it JVM-public.

SethTisue:
closing for inactivity. can be reopened if activity resumes

PR #5629, opened 2017-01-07
SI-10133 Require escaped single quote char lit

som-snytt:
The spec specifically requires `'\''` and not `'''`.
    
The error consumes all consecutive single quotes.

SI-10120 Extra advice on unclosed char literal
    
Folks from other languages might mistakenly enclose
a string in single quotes. Since this presents as
a symbol literal followed by the unpaired single
quote, we can add a syntax reminder.


som-snytt:
/rebuild

adriaanm (approved):
So much for scala being quote/unquote unfriendly to newcomers.

adriaanm:
lets hope our 'communal' build reveals any accidental usages of `'''`

som-snytt:
"quote/unquote unfriendly!" That was a good one.

adriaanm:
Sadly, I'm going to have to revert this. Looks like `'''` exists in the wild, again illustrating the rule that any syntactically valid program exists in the wild.

adriaanm:
Lets bring the spec-abidance back under -Xsource:2.13

som-snytt:
Hold on, let's make triple-single-quotes analogous to triple-double-quotes, where it can hold 0xA etc! Multiline char literal.

som-snytt:
Oh, this has a couple of commits; I'll resubmit separately.

PR #5631, opened 2017-01-10
SI-9013 SI-1459 Inherit @varargs annotations

adriaanm:
This is needed if an abstract method is annotated this way.
In addition to the abstract varargs forwarder that we already emit,
we have to emit a concrete forwarder in the subclass.


adriaanm:
/nothingtoseehere 

spurious: `run/reflection-mem-typecheck.scala`

adriaanm:
Review by @retronym (for the part he didn't author ;-))

adriaanm:
One test failure because I accidentally removed trailing whitespace from check file. Oh computer, why you so OCD.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -422,7 +422,13 @@ abstract class Erasure extends AddInterfaces
...
       new overridingPairs.Cursor(root) {
         override def parents              = List(root.info.firstParent)
-        override def exclude(sym: Symbol) = !sym.isMethod || super.exclude(sym)
+        // Varargs bridges may need generic bridges due to the non-repeated part of the signature of the involved methods.
+        // The vararg bridge is generated during refchecks (probably to simplify override checking),
+        // but then the resulting varargs "bridge" method is considered an artifact (why???),
+        // causing `super.exclude` to veto it from our bridge generation here, unless we explicitly include it (hence `!sym.hasFlag(VBRIDGE) &&`).
+        // TODO: could we generate just one bridge method that does both the java-vararg-array <-> scala-repeated-param-seq conversion
+        //       as well as the casts from the erased types of the overridden generic method (regular bridge method)?
+        override def exclude(sym: Symbol) = !sym.isMethod || (!sym.hasFlag(VBRIDGE) && super.exclude(sym))
```
lrytz: The overloaded meaning of the `ARTIFACT` flag (`ACC_SYNTHETIC` in bytecode, affects overriding pairs) is surprising, I wasn't aware of that. Without knowing the details, it seems that more explicit checks would be better in places where `isArtifact` impacts type-checking (the doc comment on `ARTIFACT` says "should be ignored when typechecking" - what ever that means).

Maybe instead of this fix here, we could skip the `ARTIFACT` flag for `VBRIDGE`s? But that's probably a bad idea in 2.11.x, as it changes the bytecode (`ACC_SYNTHETIC`).

src/compiler/scala/tools/nsc/transform/UnCurry.scala
```diff
@@ -811,7 +811,7 @@ abstract class UnCurry extends InfoTransform
       }
 
       // check if the method with that name and those arguments already exists in the template
-      enteringUncurry(currentClass.info.member(forwSym.name).alternatives.find(s => s != forwSym && s.tpe.matches(forwSym.tpe))) match {
+      enteringUncurry(currentClass.info.member(forwSym.name).filter(s => s != forwSym && s.tpe.matches(forwSym.tpe)).alternatives.headOption) match {
```
lrytz: I don't understand how this fixes the bug - why are the two versions not the same?

retronym:
I wonder if we need a corresponding change (special case for `ARTIFACT` / `VBRIDGE`) in [`checkNoDoubleDefs`](https://github.com/scala/scala/blob/83b08269e670ee12a0cecf4b72449a40ceb509f6/src/compiler/scala/tools/nsc/transform/Erasure.scala#L832-L836).

What behaviour changes if we remove the `|| sym.isArtifact` condition from `OverridingPairs#Cursor.exclude` altogether?

src/compiler/scala/tools/nsc/transform/Erasure.scala
adriaanm: I think artifacts are intended to be skipped while type checking (for performance? because they may have incorrect types? because their types are trivially correct?)

src/compiler/scala/tools/nsc/transform/UnCurry.scala
adriaanm: You're right -- the tests pass with both variations. I'll revert the code change.

adriaanm:
> What behaviour changes if we remove the || sym.isArtifact condition from OverridingPairs#Cursor.exclude altogether?

I'd prefer to delay this experiment to 2.12 or 2.13. For 2.11, I've refactored a bit to make the choice a more explicit one.

retronym:
>I think artifacts are intended to be skipped while type checking (for performance? because they may have incorrect types? because their types are trivially correct?)

`ACC_SYNTHETIC` is described a little vaguely as "does not exist in source code". More precisely, it is used to mark a member that does not exist in source code and cannot be called from source code. 

Examples in javac:

```
  public static java.lang.annotation.RetentionPolicy valueOf(java.lang.String);
    descriptor: (Ljava/lang/String;)Ljava/lang/annotation/RetentionPolicy;
    flags: ACC_PUBLIC, ACC_STATIC
```

This method is synthesized by javac because of the `enum` keyword, but it can be called from user code, so is not marked `ACC_SYNTHETIC`.

```java
class Test {
	private Object x;
	class Inner {
		{ x.toString(); }
	}
}
```

```java
  static java.lang.Object access$000(Test);
    descriptor: (LTest;)Ljava/lang/Object;
    flags: ACC_STATIC, ACC_SYNTHETIC
```

Static accessors are only called by the compiler on behalf of the user, so are marked `ACC_SYNTHETIC`.

Another example:

```java
abstract class Test implements java.util.function.Function<String, String> {
	abstract public String apply(String s);
}
```

```java
  public abstract java.lang.String apply(java.lang.String);
    descriptor: (Ljava/lang/String;)Ljava/lang/String;
    flags: ACC_PUBLIC, ACC_ABSTRACT

  public java.lang.Object apply(java.lang.Object);
    descriptor: (Ljava/lang/Object;)Ljava/lang/Object;
    flags: ACC_PUBLIC, ACC_BRIDGE, ACC_SYNTHETIC
```

So javac marks generic bridges are `ACC_SYNTHETIC`, too. Failure to hide this method from client code would make this look like an overload.

scalac emits this flag for the benefit of Java clients. AFAIK, it has no semantics to the VM itself.

When typechecking calls to Scala-defined classes, we use the `ARTIFACT` flag (as stored in the `ScalaSignature`) rather than the `ACC_SYNTHETIC` flag in the classfile. This would also serve to hide such methods from user code. In the case of generic bridges, however, we hide them from user code in a different way: the class info is pickled at a phase before it has them added.

retronym:
Finally, we also have the `SYNTHETIC` flag, which means "was not defined in source code". 

```

scala> case class C()
defined class C

scala> typeOf[C].decls.filter(_.isSynthetic)
res6: $r.intp.global.Scope = Scopes.Scope(method copy, method productPrefix, method productArity, method productElement, method productIterator, method canEqual, method hashCode, method toString, method equals)
```

This does not prevent calls to these members.

retronym:
Finally (part deux)

I remembered [SI-9482](https://issues.scala-lang.org/browse/SI-9482) that our typechecker actually does not honour `ACC_SYNTHETIC` for Java defined classes. I wasn't able to fix that on my [first attempt](https://github.com/scala/scala/compare/scala:2.12.x...retronym:ticket/9482?expand=1).

adriaanm:
I've pushed a different solution, which simply does not add the ARTIFACT flag, just VBRIDGE

adriaanm:
@retronym, @lrytz, does [this approach](https://github.com/scala/scala/pull/5631/files#diff-e2cfc09e4d7c753232666c6e31bd1926L196) look better to you?

lrytz:
LGTM, I like this approach much more.

PR #4125, opened 2014-11-10
SI-8944 A more resiliant naming scheme for case accessors

retronym:
Case class parameters that are less that public have an extra
accessor method created to ensure universal pattern matchability.
See #4081 for more background.

Currently, this is given a fresh name based on the parameter name.
However, this is fragile and the name can change based on unrelated
edits higher up in the source file.

This commit switches to a stable naming scheme for these methods.

A non-public case field `foo` has a corresponding accessor
`foo$access$N`, where `N` is the index of the parameter within
the constructor parameter list.

The enclosed tests show a case that used to trigger a linkage
error under separate compilation that now works; shows that by
choosing the `foo$access$1` rather than `foo$1` we don't clash with
lambda lifted methods in the class; and shows the names of the
accessor methods as seen via Java reflection.

Just submitting for a test run for now.

/cc @gkossakowski who is usually has some good insights when it
comes to accessor naming schemes. This is one of the cross
compilation unit use cases that we have to deal with that Java
doesn't.

PR #5634, opened 2017-01-10
SI-8831 Fix mapping case accessor lookup with backticked idents

adriaanm:
Case class parameters may be declared as non-public. If this is the
case, the case accessor method is still synthesized as public, but
it has a mangled name of the form `origName$N`.

See also SI-8944 / #4125.

The compiler maps from case fields to the corresponding accessors
with this naming convention in mind. However, it is too lenient, and
assumes that only one accessor will be of the form "origName\$.*".

If, however, a regular, public case parameter is named `a b`, it will
have a name `a$u0020b`, which spurously matches that overly permissive
pattern.

This commit changes tightens up the logic to expect "origName\$\d+".


adriaanm:
I'm working on a fix/cleanup, but I'm going to target 2.12. Too risky for 2.11.9.

PR #5636, opened 2017-01-10
Upgrade to jline2 2.14.3

adriaanm:
Already "backported" to 2.11.x in #5635


adriaanm:
See also https://issues.scala-lang.org/browse/SI-1416

PR #5638, opened 2017-01-10
SI-10137 Correct Queue.enqueue(collection.Iterable)

som-snytt:
Correct Queue.enqueue(collection.Iterable). Otherwise it required immutable.Iterable.

Compare with replacing with enqueue(x,y,zs):
https://github.com/scala/scala/compare/2.13.x...som-snytt:issue/10137-new-api?expand=1

Better improvements are left to the strawman.

Review by @SethTisue


SethTisue:
Sorry for the long review delay here. We are trying to do better on this (again... sigh).

JIRA: https://issues.scala-lang.org/browse/SI-10137

SethTisue (approved):

SethTisue:
actually... isn't SI-10137 just a duplicate of SI-6166? if so, we should close SI-10137 and change the commit message here to refer to SI-6166?

@Ichoran usually likes to take a quick look at this kind of thing.

som-snytt:
Actually, since both SI-6166 and SI-6981 were reported by Daniel Sobral, they are subject to the nullification rule, which was instituted to prevent JIRA issue inflation. Also, SI-6166 falls victim to the fix-by rule, whereby an issue is marked as "deferred to 2.11" (in this case) but then ignored. Indefinite deferral is grounds for closure. Obviously, the current issue has the additional advantage of both an astute analytical comment and a PR.

In addition, the example in SI-10137 is actually puzzler-worthy. What a missed opportunity! Is it possible to defer this PR until after the next edition of the puzzlers book? Or never mind, let's just squash the puzzlers.

SethTisue:
pretty sure if you enter "10137" into an old TI calculator and hold it upside down, it spells something rude. (have I made this joke before?)

SethTisue:
review by @Ichoran 

hey let's see if this new Scabot feature works where merely saying that makes it so

SethTisue:
livin’ in the 1001-Arabian-Nights future

Ichoran (approved):
Looks fine to me!  (Scabot, too, @SethTisue!)

SethTisue:
thanks AP, RK, SB!

som-snytt:
Pretty sure if you type `1001` into an old `TI` calculator and invert it, it spells the same number of nights in arabic numerals.

PR #5639, opened 2017-01-10
[nomerge] SI-10137 Queue.enqueue(Iterable) is old

som-snytt:


som-snytt:
Decided that tweaking API before the strawman storm isn't helpful.

For the record, I'd opted for adding `enqueue(x: B, y: B, zs: B*)` which is awkward to use. Efficiently enqueueing a single item is desirable.

PR #5643, opened 2017-01-16
SI-8417 Check adapts of each param section

som-snytt:
Previously, adaptations like auto-tupling were checked
only on the last param section of an application.

This commit runs the sanity check always.

JIRA: https://issues.scala-lang.org/browse/SI-8417

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3426,29 +3426,29 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           // repeat vararg as often as needed, remove by-name
           val formals = formalTypes(paramTypes, argslen)
 
-          /* Try packing all arguments into a Tuple and apply `fun`
-           * to that. This is the last thing which is tried (after
-           * default arguments)
+          /* Try packing all arguments into a Tuple and apply `fun` to that.
+           * This is the last thing which is tried (after default arguments).
            */
-          def tryTupleApply: Tree = {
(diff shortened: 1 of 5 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3413,29 +3413,31 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
-              silent(_.doTypedApply(tree, fun, tupleArgs, mode, pt)) map { t =>
-                // Depending on user options, may warn or error here if
-                // a Unit or tuple was inserted.
-                val keepTree = (
-                     !mode.typingExprNotFun // why? introduced in 4e488a60, doc welcome
-                  || t.symbol == null       // ditto
-                  || checkValidAdaptation(t, args)
-                )
+              // May warn or error here if a Unit or tuple was inserted.
+              def validate(t: Tree): Tree = {
```
adriaanm: This goes in the right direction in making the check more explicit. Can we push it a bit more and avoid future reorderings of the conditions?
```scala
// May warn or error if a Unit or tuple was inserted.
def validate(t: Tree): Tree = {
  val invalidAdaptation = t.symbol != null && !checkValidAdaptation(t, args) // regardless of typer's mode
  if (invalidAdaptation && mode.typingExprNotFun) EmptyTree // only bail if we're typing an expression (and not inside another application)
  else t
}
```

adriaanm (changes requested):

adriaanm (approved):

PR #5648, opened 2017-01-17
SI-10148 Follow Java for float literals

som-snytt:
Use `Float.parseFloat` instead of converting from Double.
Error when a value rounds to zero.

JIRA: https://issues.scala-lang.org/browse/SI-10148


test/files/run/literals.scala
```diff
@@ -84,6 +84,10 @@ object Test {
     check_success("3.14f == 3.14f", 3.14f, 3.14f)
     check_success("6.022e23f == 6.022e23f", 6.022e23f, 6.022e23f)
     check_success("09f == 9.0f", 09f, 9.0f)
+    check_success("1.00000017881393421514957253748434595763683319091796875001f == 1.0000001f",
+      1.00000017881393421514957253748434595763683319091796875001f,
+      1.0000001f)
+    check_success("3.4028235E38f == Float.MaxValue", 3.4028235E38f, Float.MaxValue)
```
janekdb, line 90: Why no check using Float.MinValue? Probably demonstrating my ignorance of something here!

som-snytt: My name isn't ScalaCheck. Also, the difference is just a sign bit which is turned on by a leading midlevel horizontal bar symbol. Of course, the test wouldn't hurt.

janekdb: Ignorance of IEEE 754 fully demoed (by me)! Java MIN_VALUE is not the negative float with the greatest magnitude but the smallest positive nonzero value. 

https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#MIN_VALUE

Having said that why not add a check for the effect of a leading `hyphen-minus`?

som-snytt: OK, if I have time to switch my repo back to this branch on my old home machine.

There's a puzzler (BTW) that turns on the fact that `-1.0f` is a literal. [See here.](http://scalapuzzlers.com/#pzzlr-041)

SethTisue:
thx AP!

SethTisue:
this turned up in the community build:

```
[breeze] [error] /Users/tisue/community.212/target-0.9.7-RC1/project-builds/breeze-ef73d8d5e54fa4ff81b6f6abfe9e238739cc5d8a/math/src/test/scala/breeze/signal/filter/DesignFilterTest.scala:53: double precision floating point number too small
[breeze] [error]   val testFirwin4 = DenseVector( -0.000000000000000000e+00, 3.794040820411270776e-01, 3.794040820411272441e-01, -0.000000000000000000e+00)
[breeze] [error]                                   ^
[breeze] [error] /Users/tisue/community.212/target-0.9.7-RC1/project-builds/breeze-ef73d8d5e54fa4ff81b6f6abfe9e238739cc5d8a/math/src/test/scala/breeze/signal/filter/DesignFilterTest.scala:53: double precision floating point number too small
[breeze] [error]   val testFirwin4 = DenseVector( -0.000000000000000000e+00, 3.794040820411270776e-01, 3.794040820411272441e-01, -0.000000000000000000e+00)
[breeze] [error]                                                                                                                  ^
[breeze] [error] two errors found
```

is this a regression, or is the code in breeze erroneous...?

som-snytt:
Zero is pretty small. It looks like the little heuristic for testing whether you said zero is not right.

src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
```diff
@@ -972,23 +972,45 @@ trait Scanners extends ScannersCommon {
...
-    def floatVal(negated: Boolean): Double = {
-      val limit: Double = if (token == DOUBLELIT) Double.MaxValue else Float.MaxValue
+    def floatVal(negated: Boolean): Float = {
       try {
-        val value: Double = java.lang.Double.valueOf(strVal).doubleValue()
-        if (value > limit)
+        val value: Float = java.lang.Float.parseFloat(strVal)
+        if (value > Float.MaxValue)
           syntaxError("floating point number too large")
+        val zeroly = "0.fF"
```
som-snytt, line 982: This leaves out the scientists, who use scientific notation.

PR #5649, opened 2017-01-18
Faster compilation of inductive implicits.

milessabin:
This PR provides much faster (35 x in the provided benchmark case) compilation of the kind of inductive implicit resolution that is found in shapeless and its uses for type class derivation in libraries such as Circe, Doobie and Scodec. I will follow up with a PR for a backport of this to 2.11.x later today.

The method is to spot patterns in the set of eligible implicit definitions for an implicit argument of the form `F[A, B, C ...]` corresponding to a singly recursive, convergent (to either success or failure) induction. When these hold the induction is solved for directly by unfolding the recursive steps until one of the base cases is reached. The patterns are a little conservative (in the sense that some inductions might be missed), but they cover many important cases, in particular most of the interesting inductions in shapeless are handled.

The strategy is enabled by the `-Yinduction-heuristics` scalac flag. If it isn't enabled then normal implicit resolution is always performed. If it is enabled, then the inductive strategy is applied where possible. If it cannot be applied implicit resolution falls back to the normal mechanism. In all cases it is intended that the semantics are identical to normal implicit resolution: the only effect of enabling this option should be a reduction in compile times for applicable implicit resolutions.

An `@inductive` annotation for types is provided with a similar intent to `@tailrec`. It asserts that implicit instances of types with the annotation are expected to be solved for by the new induction mechanism. It is an error if they are not and they are instead solved by the normal mechanism. Note that the annotation does not change semantics in the success case (it does change semantics in the failure case because an error is reported). Nor does it affect which algorithm is used to resolve implicits: it is simply a post hoc check that the expected solver was used successfully, in the same way that `@tailrec` is a post hoc check that an expected tail call optimization was performed.

The use of `@inductive` allows more informative error messages to be generated for inductions which fail due to missing implicit instances of types which are auxiliary to the main induction. For example, in the case of shapeless's `Mapper` type class (which recurses through an `HList` applying an auxiliary type class instance specific to the type of each element as it goes), with normal implicit search if any instance of the auxiliary type class is missing the error reported will be just that the overall `Mapper` instance can't be resolved. With this algorithm, and in the presence of the `@inductive` annotation, the compiler knows that an induction is being performed and is able to report that an induction has failed due to the absence of the specific auxiliary type class instance. This is a much requested feature by the authors and users of libraries based on shapeless type class derivation. See `test/files/neg/inductive-implicts.{scala,flags,check}` for an example.

One limitation that people using inductive implicits now where the induction step is wrapped in shapeless's `Lazy` will notice is that these will not take advantage of this algorithm. In some cases it may be possible to drop the use of `Lazy`, though not all. I intend to follow this PR with another which provides an interpretation of byname implicit arguments which is equivalent to shapeless's `Lazy` and which will get along nicely with this induction heuristic.

All (par)tests pass with this option forcibly enabled. In addition shapeless, Circe, Doobie and Scodec, which make extensive use of these forms of induction, all compile and test successfully. A benchmark is provided in `test/files/induction` which validates my performance claims.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -82,14 +82,25 @@ trait ContextErrors {
   //    (pt at the point of divergence gives less information to the user)
   // Note: it is safe to delay error message generation in this case
   // because we don't modify implicits' infos.
+  sealed abstract class ImplicitTypeError extends TreeTypeError
+
   case class DivergentImplicitTypeError(underlyingTree: Tree, pt0: Type, sym: Symbol)
-    extends TreeTypeError {
+    extends ImplicitTypeError {
     def errMsg: String   = errMsgForPt(pt0)
     def withPt(pt: Type): AbsTypeError = this.copy(pt0 = pt)
     private def errMsgForPt(pt: Type) =
       s"diverging implicit expansion for type ${pt}\nstarting with ${sym.fullLocationString}"
   }
 
+  case class NoninductiveImplicitTypeError(underlyingTree: Tree, pt: Type)
+    extends ImplicitTypeError {
+    def errMsg: String = s"Noninductive implicit expansion for type ${pt}"
+  }
+
+  case class IncompleteInductionImplicitTypeError(underlyingTree: Tree, pt: Type, aux: Type)
+    extends ImplicitTypeError {
+    def errMsg: String = s"Inductive implicit expansion for type ${pt} failed due to missing auxiliary implicit ${aux}"
+  }
 
   case class PosAndMsgTypeError(errPos: Position, errMsg: String)
     extends AbsTypeError
@@ -1306,6 +1317,12 @@ trait ContextErrors {
 
     def DivergingImplicitExpansionError(tree: Tree, pt: Type, sym: Symbol)(implicit context0: Context) =
       issueTypeError(DivergentImplicitTypeError(tree, pt, sym))
+
+    def NoninductiveImplicitExpansionError(tree: Tree, pt: Type)(implicit context0: Context) =
+      issueTypeError(NoninductiveImplicitTypeError(tree, pt))
+
+    def IncompleteInductionImplicitExpansionError(tree: Tree, pt: Type, aux: Type)(implicit context0: Context) =
+      issueTypeError(IncompleteInductionImplicitTypeError(tree, pt, aux))
   }
 
   object NamesDefaultsErrorsGen {
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -1321,15 +1321,23 @@ trait Contexts { self: Analyzer =>
     // have to pass in context because multiple contexts may share the same ReportBuffer
     def reportFirstDivergentError(fun: Tree, param: Symbol, paramTp: Type)(implicit context: Context): Unit =
       errors.collectFirst {
-        case dte: DivergentImplicitTypeError => dte
+        case ite: ImplicitTypeError => ite
       } match {
-        case Some(divergent) =>
+        case Some(divergent: DivergentImplicitTypeError) =>
           // DivergentImplicit error has higher priority than "no implicit found"
           // no need to issue the problem again if we are still in silent mode
           if (context.reportErrors) {
             context.issue(divergent.withPt(paramTp))
             errorBuffer.retain {
-              case dte: DivergentImplicitTypeError => false
+              case ite: ImplicitTypeError => false
+              case _ => true
+            }
+          }
+        case Some(ite) =>
+          if (context.reportErrors) {
+            context.issue(ite)
+            errorBuffer.retain {
+              case ite: ImplicitTypeError => false
               case _ => true
             }
           }
@@ -1345,7 +1353,7 @@ trait Contexts { self: Analyzer =>
 
     def propagateImplicitTypeErrorsTo(target: ContextReporter) = {
       errors foreach {
-        case err@(_: DivergentImplicitTypeError | _: AmbiguousImplicitTypeError) =>
+        case err@(_: ImplicitTypeError | _: AmbiguousImplicitTypeError) =>
           target.errorBuffer += err
         case _ =>
       }
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -183,6 +183,7 @@ trait Implicits {
     def isFailure          = false
     def isAmbiguousFailure = false
     def isDivergent        = false
+    def isNoninductive     = false
     final def isSuccess    = !isFailure
   }
 
@@ -198,6 +199,11 @@ trait Implicits {
   lazy val AmbiguousSearchFailure = new SearchResult(EmptyTree, EmptyTreeTypeSubstituter, Nil) {
     override def isFailure          = true
     override def isAmbiguousFailure = true
+  }
+
+  lazy val NoninductiveSearchFailure = new SearchResult(EmptyTree, EmptyTreeTypeSubstituter, Nil) {
+    override def isFailure      = true
+    override def isNoninductive = true
   }
 
   /** A class that records an available implicit
@@ -220,7 +226,7 @@ trait Implicits {
       isCyclicOrErroneousCache.booleanValue
     }
 
-    private[this] final def computeIsCyclicOrErroneous =
+    def computeIsCyclicOrErroneous =
       try sym.hasFlag(LOCKED) || containsError(tpe)
       catch { case _: CyclicReference => true }
 
@@ -825,12 +831,70 @@ trait Implicits {
      *                             shadowing. The head of the list `iss` must represent implicits from the closest
      *                             enclosing scope, and so on.
      */
-    class ImplicitComputation(iss: Infoss, isLocalToCallsite: Boolean) {
+    class ImplicitComputation(iss: Infoss, isLocalToCallsite: Boolean) extends AbstractImplicitComputation(iss, isLocalToCallsite) {
+      /** Sorted list of eligible implicits.
+       */
+      val eligible: List[ImplicitInfo]  = {
+        val matches = iss flatMap { is =>
+          val result = is filter (info => checkValid(info.sym) && survives(info))
+          shadower addInfos is
+          result
+        }
+
+        // most frequent one first
+        matches sortBy (x => if (isView) -x.useCountView else -x.useCountArg)
+      }
+      if (eligible.nonEmpty)
+        printTyping(tree, eligible.size + s" eligible for pt=$pt at ${fullSiteString(context)}")
+    }
+
+    // MS: The split between the original and the new inductive computation could be more
+    // cleanly factored, but deferred to make it easier to see the difference between the
+    // cases.
+    class InductiveImplicitComputation(iss: Infoss, isLocalToCallsite: Boolean) extends AbstractImplicitComputation(iss, isLocalToCallsite) {
+      def inductionType(tpe: Type): Type = {
+        tpe.dealias match {
+          case TypeRef(pre, sym, args) => typeRef(pre, sym, args.map(_ => WildcardType))
+          case RefinedType(List(t), _) => inductionType(t)
+          case RefinedType(List(a, t), _) if a =:= AnyRefClass.tpe => inductionType(t)
+          case t => t
+        }
+      }
+
+      val inductionPt = inductionType(pt)
+
+      def matchesInductionPt(info: ImplicitInfo): Boolean = (
+        info.isStablePrefix && matchesPt(depoly(info.tpe), inductionPt, Nil)
+      )
+
+      def isIneligibleForInduction(info: ImplicitInfo) = (
+           info.computeIsCyclicOrErroneous // avoid side effects
+        || isView && (info.sym eq Predef_conforms) // as an implicit conversion, Predef.$conforms is a no-op, so exclude it
+        || (!context.macrosEnabled && info.sym.isTermMacro)
+      )
+
+      def survivesForInduction(info: ImplicitInfo) = (
+           !isIneligibleForInduction(info)               // cyclic, erroneous, shadowed, or specially excluded
+        && isPlausiblyCompatible(info.tpe, inductionPt)  // optimization to avoid matchesPt
+        && !shadower.isShadowed(info.name)          // OPT rare, only check for plausible candidates
+        && matchesInductionPt(info)                          // stable and matches expected type
+      )
+
+      val eligible: List[ImplicitInfo] = {
+        iss flatMap { is =>
+          val result = is filter (info => checkValid(info.sym) && survivesForInduction(info))
+          shadower addInfos is
+          result
+        }
+      }
+    }
+
+    abstract class AbstractImplicitComputation(iss: Infoss, isLocalToCallsite: Boolean) {
       abstract class Shadower {
         def addInfos(infos: Infos)
         def isShadowed(name: Name): Boolean
       }
-      private val shadower: Shadower = {
+      val shadower: Shadower = {
         /** Used for exclude implicits from outer scopes that are shadowed by same-named implicits */
         final class LocalShadower extends Shadower {
           val shadowed = util.HashSet[Name](512)
@@ -870,7 +934,7 @@ trait Implicits {
 
       /** Tests for validity and updates invalidImplicits by side effect when false.
        */
-      private def checkValid(sym: Symbol) = isValid(sym) || { invalidImplicits += sym ; false }
+      def checkValid(sym: Symbol) = isValid(sym) || { invalidImplicits += sym ; false }
 
       /** Preventing a divergent implicit from terminating implicit search,
        *  so that if there is a best candidate it can still be selected.
@@ -911,20 +975,7 @@ trait Implicits {
         }
       }
 
-      /** Sorted list of eligible implicits.
-       */
-      val eligible = {
-        val matches = iss flatMap { is =>
-          val result = is filter (info => checkValid(info.sym) && survives(info))
-          shadower addInfos is
-          result
-        }
-
-        // most frequent one first
-        matches sortBy (x => if (isView) -x.useCountView else -x.useCountArg)
-      }
-      if (eligible.nonEmpty)
-        printTyping(tree, eligible.size + s" eligible for pt=$pt at ${fullSiteString(context)}")
+      def eligible: List[ImplicitInfo]
 
       /** Faster implicit search.  Overall idea:
        *   - prune aggressively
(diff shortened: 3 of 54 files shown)


src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -1302,15 +1302,23 @@ trait Contexts { self: Analyzer =>
...
+            }
+          }
+        case Some(ite) =>
+          if (context.reportErrors) {
+            context.issue(ite)
+            errorBuffer.retain {
+              case ite: ImplicitTypeError => false
               case _ => true
             }
           }
```
xeno-by, line 1343: Maybe it would be possible to merge thes two cases into one and instead have a smaller pattern match to determine what error to issue?

milessabin: Possibly. I found the logic about which error actually gets reported pretty impenetrable here and tried multiple variations in the hope of doing what you're suggesting ... none of them did the right thing. Advice on the right way to do it would be very much appreciated.

src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
```diff
@@ -82,14 +82,25 @@ trait ContextErrors {
...
   }
 
+  case class NoninductiveImplicitTypeError(underlyingTree: Tree, pt: Type)
+    extends ImplicitTypeError {
+    def errMsg: String = s"Noninductive implicit expansion for type ${pt}"
+  }
+
+  case class IncompleteInductionImplicitTypeError(underlyingTree: Tree, pt: Type, aux: Type)
+    extends ImplicitTypeError {
+    def errMsg: String = s"Inductive implicit expansion for type ${pt} failed due to missing auxilliary implicit ${aux}"
```
xeno-by: Is double L in `auxiliary` a typo?

milessabin: Yes!

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1011,6 +1062,271 @@ trait Implicits {
...
+          def loop(pt: Type): SearchResult = {
+            ranked.foldLeft(nonInductive: SearchResult) {
+              case (acc, info) if acc.isFailure =>
+                val savedUndetparams = context.undetparams
+                val solvedContext = solve(pt, info.tpe) // Is there duplication between solve and the subsequent type
+                                                        // checking which could be eliminated?
+                val result =
+                  if(solvedContext.isFailure) acc
+                  else {
+                    val allUndetparams = (context.undetparams ++ context.outer.undetparams ++ solvedContext.undetparams).distinct
```
jvican: s/context.undetparams/savedUndetparams/g?

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
xeno-by: Hmm I just noticed that the cases are exactly the same except for the small `divergent.withPt(paramTp)` vs `ite` diff. I'd write something like:

```
case Some(ite) =>
  if (context.reportErrors) {
    val err = if (ite.isInstanceOf[DivergentImplicitTypeError]) divergent.withPt(paramTp) else ite
    context.issue(err)
    errorBuffer.retain {
      case ite: ImplicitTypeError => false
      case _ => true
    }
  }
```

src/library/scala/annotation/inductive.scala
```diff
@@ -0,0 +1,19 @@
...
+
+package scala.annotation
+
+/** A type annotation which verifies that the implicit values of the
+ *  annotated type will be resolved inductively.
+ *
+ *  If it is present, the compiler will issue an error if implicit values
+ *  cannot be resolved inductively.
+ *
+ *  @since 2.12.0
```
fommil, line 17: shouldn't this be 2.12.2?

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: No, I don't think so.

src/library/scala/annotation/inductive.scala
milessabin: Yes it should.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -805,12 +811,70 @@ trait Implicits {
...
+        // most frequent one first
+        matches sortBy (x => if (isView) -x.useCountView else -x.useCountArg)
+      }
+      if (eligible.nonEmpty)
+        printTyping(tree, eligible.size + s" eligible for pt=$pt at ${fullSiteString(context)}")
+    }
+
+    // MS: The split between the original and the new inductive computation could be more
+    // cleanly factored, but deferred to make it easier to see the difference between the
+    // cases.
```
xeno-by, line 853: I think it would be beneficial if there was a step-by-step example in comments explaining inductive implicit lookup as per newly implemented logic. The test suite is extensive and even has approximate performance numbers (!), but personally for me a detailed example would be useful too.

milessabin: Agreed.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1011,6 +1062,271 @@ trait Implicits {
...
+                    val tree1 =
+                      if(isImplicitMethodType(mt)) {
+                        applyImplicitArgs(fun3, mt, solvedContext)
+                      } else {
+                        solvedContext.subst traverse fun3
+                        fun3
+                      }
+
+                    val tree2 = typed(tree1, EXPRmode, wildPt)
+                    if(tree2.isErroneous) acc
```
xeno-by: I think the `if(...)` style is much less common in scalac than `if (...)`.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1011,6 +1062,271 @@ trait Implicits {
...
+        val targs = solvedTypes(tvars, allUndetparams, allUndetparams map varianceInType(pt), upper = false, lubDepth(wildPt :: tp :: Nil))
+        if(!checkBounds(EmptyTree, NoPrefix, NoSymbol, allUndetparams, targs, "inferred ")) SearchFailure
+        else {
+          val AdjustedTypeArgs(okParams, okArgs) = adjustTypeArgs(allUndetparams, tvars, targs)
+          val undetparams = allUndetparams diff okParams
+          val tpSubst = tp0.instantiateTypeParams(okParams, okArgs)
+          if(!matchesPt(tpSubst, wildPt, undetparams)) SearchFailure
+          else {
+            val subst: TreeTypeSubstituter =
+              if (okParams.isEmpty) EmptyTreeTypeSubstituter
```
mpociecha: I know it's a trifle but you write `if`s inconsistently. Personally I would always add one space between `if` and `(`.

test/files/neg/inductive-implicits3.scala
```diff
@@ -0,0 +1,196 @@
+object shapeless {
```
fommil, line 1: (puts on RMS halo) you may wish to include the shapeless LICENSE / NOTICE, unless this is a subset of which you are the sole author, or consider it trivial.

smarter:
Great work! I have two remarks:
- Would it be possible to add a  mode where both the induction search and regular search are used to verify that induction does not come up with something different? It would be enabled by a flag like `-Ycheck-induction-heuristics` and used in tests and when debugging.
- This stuff is not trivial at all, I think the code would benefits from more comments explaining how the different pieces of the algorithm work in details, what are the assumption, etc.

fommil:
does the scalac build produce a mima report ? I'd like to be sure that this won't break ensime or scala-ide without a recompile (which is more important for the 2.11.x branch)

dwijnand:
@fommil yes, to verify both backwards and forwards compatibility actually.

mpociecha:
@fommil Also there's the integration build for scala-ide.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: True. I'll fix.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: Agreed. My normal style is no space, the scalac style is a space. I try to conform to local customs but sometimes slip.

milessabin:
@smarter agreed on both points.

milessabin:
For some reason I'm having trouble finding the Jenkins log for the failure ... am I looking in the wrong place or is Jenkins borked?

test/files/neg/inductive-implicits3.scala
milessabin: This is a subset of which I am the sole author.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: Agreed.

adriaanm:
```
[info] prev = /home/jenkins/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.12.0.jar, curr = /home/jenkins/workspace/scala-2.12.x-validate-test/build/pack/lib/scala-library.jar
Found 1 binary incompatibilities (4 were filtered out)
 * class scala.annotation.inductive does not have a correspondent in current
   version

Generated backward filter config definition
============================================

    filter {
        problems=[
            {
                matchName="scala.annotation.inductive"
                problemName=MissingClassProblem
            }
        ]
    }


Generated forward filter config definition
===========================================

    filter {
        problems=[]
    }
```

available by clicking on "Details" next to the validate-test failure, then on Console Output. (https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/3997/console)

adriaanm:
Before you backport to 2.11, let's first complete the review/merge cycle in 2.12. This is not a trivial change, so it's going to take a while to digest, review and iterate. Because the 2.11.9 deadline is so close, I expect this will not make it into 2.11.9, though a fine candidate for the TLS 2.11.x series.

milessabin:
I'll make a PR against 2.11.x and update as this PR evolves. I don't expect you to merge for LBS 2.11.9, but it will be included in TLS 2.11.9.

milessabin:
Gah, why do I always seem to get those whitelists wrong :-(

lrytz:
> Also there's the integration build for scala-ide.

@mpociecha that build is actually not running on 2.12.x, there's no scala-ide for 2.12 yet.

retronym:
I'm having some trouble compiling the pos test with and without the new option:

```
⚡ time  qscalac -nowarn  test/files/pos/inductive-implicits.scala
test/files/pos/inductive-implicits.scala:583: error: could not find implicit value for parameter toTraversable: shapeless.ToTraversable[shapeless.HNil,List]
  val ttnil = ToTraversable[HNil, List]
                           ^
test/files/pos/inductive-implicits.scala:586: error: could not find implicit value for parameter e: shapeless.ToTraversable.Aux[shapeless.HNil,List,Int]
  implicitly[ToTraversable.Aux[HNil, List, Int]]
            ^
test/files/pos/inductive-implicits.scala:588: error: could not find implicit value for parameter toTraversable: shapeless.ToTraversable[shapeless.::[Int,shapeless.HNil],List]
  val tti = ToTraversable[Int :: HNil, List]
                         ^
test/files/pos/inductive-implicits.scala:591: error: could not find implicit value for parameter toTraversable: shapeless.ToTraversable[shapeless.::[Test.Apple,shapeless.::[Test.Pear,shapeless.HNil]],List]
  val ttap = ToTraversable[Apple :: Pear :: HNil, List]
                          ^
four errors found

real	0m26.175s
user	0m45.837s
sys	0m0.796s

~/code/scala on topic/inductive-implicits-2.12.x*
⚡ time  qscalac -Yinduction-heuristics -nowarn  test/files/pos/inductive-implicits.scala
test/files/pos/inductive-implicits.scala:583: error: Inductive implicit expansion for type shapeless.ToTraversable[shapeless.HNil,List] failed due to missing auxilliary implicit scala.collection.generic.CanBuildFrom[List[Nothing],Nothing,List[Nothing]]
  val ttnil = ToTraversable[HNil, List]
                           ^
test/files/pos/inductive-implicits.scala:586: error: Inductive implicit expansion for type shapeless.ToTraversable.Aux[shapeless.HNil,List,Int] failed due to missing auxilliary implicit scala.collection.generic.CanBuildFrom[List[Int],Int,List[Int]]
  implicitly[ToTraversable.Aux[HNil, List, Int]]
            ^
test/files/pos/inductive-implicits.scala:588: error: Inductive implicit expansion for type shapeless.ToTraversable[shapeless.::[Int,shapeless.HNil],List] failed due to missing auxilliary implicit scala.collection.generic.CanBuildFrom[Nothing,Int,List[Int]]
  val tti = ToTraversable[Int :: HNil, List]
                         ^
test/files/pos/inductive-implicits.scala:591: error: Inductive implicit expansion for type shapeless.ToTraversable[shapeless.::[Test.Apple,shapeless.::[Test.Pear,shapeless.HNil]],List] failed due to missing auxilliary implicit scala.collection.generic.CanBuildFrom[Nothing,Test.Pear,List[Test.Pear]]
  val ttap = ToTraversable[Apple :: Pear :: HNil, List]
                          ^
four errors found

real	0m31.445s
user	0m44.614s
sys	0m0.633s
```

I was hoping to produce and diff logs of the old and new tree of implicit searches to help get my intuition for how the trick works.

dwijnand:
@retronym Maybe just needs a clean rebuild of the compiler (for whatever reason)?

```bash
11:23:02 $ time qscalac -nowarn test/files/pos/inductive-implicits.scala

real	0m21.548s
user	0m43.258s
sys	0m0.970s

11:23:35 $

11:26:43 $ rm -r *.class shapeless/

11:26:50 $ time qscalac -nowarn -Yinduction-heuristics test/files/pos/inductive-implicits.scala

real	0m31.359s
user	0m50.577s
sys	0m1.076s
```

milessabin:
@retronym I'm not able to reproduce that failure running `partest test/files/pos/inductive-implicits.scala` from the SBT prompt. What does your `qscala` wrapper look like?

retronym:
Things are working after a clean build and `rm *.class`.

Next question: I'm seeing that compilation is slower for `pos/inductive-implicits.scala` with `-Yinduction-heuristics` enabled.

```
% sbt setupPublishCore dist/mkQuick && time qscalac -nowarn test/files/pos/inductive-implicits.scala && time qscalac -nowarn -Yinduction-heuristics test/files/pos/inductive-implicits.scala
...
real	0m25.708s
user	0m54.443s
sys	0m1.089s

real	0m34.328s
user	0m55.813s
sys	0m0.823s
```

milessabin:
@retronym there is an overhead due to performing the test that the inductive heuristic is applicable. The breakeven point would be for inductions which are larger than the ones in `pos/inductive-implicits.scala` ... see the benchmark in `test/files/induction` for an example which shows how times scale with induction size for compilations with and without `-Yinduction-heurisitics`. The large sizes seen there are not atypical for people attempting to use shapeless for type class derivation for medium to large application data models.

I think there is probably scope for improving performance even for small instances, but the big wins are the ones we have here for medium to large instances.

clhodapp:
I'll chime in and note that in its current state, this patch cuts the compile time for my inductive-implicit-heavy work project by slightly more than half (258 seconds for a clean compile becomes 121). If @milessabin implements the replacement for `shapeless.Lazy` described in the description, that should drop even more. I can't share the code though, sorry... I just wanted to note that there definitely do exist real-world projects that benefit a *lot* from this even without further optimization (not that further optimization couldn't make the picture even better).

fommil:
ensime is heavy on typeclass derivation, any volunteers willing to take that on? I spent today getting a 2.12 release out..

milessabin:
FYI: this has been [merged](https://github.com/typelevel/scala/blob/typelevel-readme/notes/2.12.1.md#faster-compilation-of-inductive-implicits-pull5649-milessabin) in Typelevel Scala 2.12.1.

dotta:
I've cherry-picked the commits in this PR on top of the v2.12.1 tag to check if a codebase (can't reveal more) that extensively uses shapeless would benefit from the new inductive heuristics. Our (non-extensive) tests showed that on the specific codebase there was a significant (almost 30%) compile time degradation with `-Yinduction-heuristics` enabled. The Typelevel Scala 2.12.1 release was also tested, and it yields similar results. I'm sharing this just as a data point.

milessabin:
@dotta Are you able to give an indication of the sizes of the inductions you're seeing in that codebase? 

Also, have you verified that the heuristic is actually being applied? You can check that by adding the `@inductive` annotation to the type class trait. If that results in a warning the the instances aren't being solved for inductively then you'll be experiencing all the overhead of the mechanism without gaining any of the benefits.

dotta:
@milessabin Unfortunately, I can't. All I've tried was turning on the flag and check if compilation time would drop. No further investigation was done.

fommil:
that flag sounds very useful. A no-op annotation for 2.10/2.11 would be very useful to use it in cross builds. I may do that (but I still await your Lazy fix...)

milessabin:
@dotta I'll ping you offline and we'll see if we can get a bit more information out of this.

milessabin:
I've followed up with @dotta and it appears that the induction heuristics weren't applicable (due to use of shapeless's `Lazy` as discussed at the top of this thread) so all he's seeing is the overhead of the applicability test.

There's quite a bit of scope for improving the performance of the applicability test (paging @rorygraves). I was hoping for some feedback on correctness before digging in to tuning ... @adriaanm, @retronym if you're happy on that score I'll starting looking at reducing the cost of the test.

fommil:
is there any hope at all of a version of this that works with current usages of `Lazy`? I know the new language feature you were discussing is far superior... but it doesn't help with existing codebases and will introduce both source and binary incompatible changes to libraries like spray-json-shapeless and circe.

milessabin:
@fommil that would require explicit knowledge of (though not a dependency on) shapeless in the compiler. If you're willing to lobby for it then I would be more than happy to implement it.

fommil:
@milessabin just point my cage where it needs to point, and tell me when to start speaking.

milessabin:
@fommil here is fine :-)

fommil:
In that case I'll motivate why it would be so useful.

Miles' work here is most beneficial for speeding up compilation when there are complex implicit derivations matching a particular pattern, such as typeclass derivation. There is a problem in that resolution diverges and fails for recursive types (I assume you all know what that means) and there is a hack in shapeless which introduces `Lazy` and `Strict` containers to give hints such that the developer can trick the compiler into implicit derivation of recursive types (caveat emptor).

But unfortunately, Miles' improvements in this PR do not work for `Lazy` - he says it needs the compiler to be shapeless-aware. This is even more unfortunate, because many typeclass derivation libraries such as spray-json-shapeless, circe, scodec, etc etc, are all designed to handle recursive types even if the developer doesn't have any. Therefore, none of this great work is being passed on to users of those libraries.

So, please let Miles add the hack! There is a real fix coming soon with an associated language change (for by-name implicits), but that is source incompatible with existing code that will likely require making all typeclass derivation libraries 2.12.x+

Today, we actually need something that is source compatible all the way back to 2.10 - library authors still has to cross compile for 2.10, because sbt.

Please somebody ask Eugene very nicely to release sbt 0.13.13 built with scala 2.12 and call it 0.14.0 :smiley:

milessabin:
Ambiguous implicits now handled correctly. Squashed.

fommil:
Any news on Lazy?

milessabin:
That's not part of this PR.

Atry:
Does this PR perform the same algorithm as https://github.com/lampepfl/dotty/issues/1998 ?

milessabin:
@Atry that's what @fommil was referring to earlier. I'm working on that as a separate PR.

adriaanm:
There are a few pending TODO's in the review comments left by Eugene. The key to making this PR reviewable is more docs: the commit message should motivate the change and explain how it works in some detail, and the code also needs more comments and examples. Accepting this PR means taking on responsibility for maintaining this code in the future, which crucially relies on a design doc / spec.

jvican:
I've just discovered a funny behaviour with implicits. [In this inductive example](https://github.com/milessabin/scala/blob/d3dd25e1886d0d5b5bedfce62ce7a693efd5b900/test/induction/inductive-implicits-bench.scala),  turning the type parameter `L` contravariant slightly speeds up implicit lookup.
  
These are the results I get for N=30 (**EDIT**: N=300). I've reproduced for smaller N's too, and I've also made sure to warm up the compiler before getting these behaviours. They're all captured from within sbt.
```
1. Waiting for source changes... (press enter to interrupt)
[info] Packaging /data/rw/code/scala/scala-profiling/plugin/target/scala-2.12/scalac-profiling_2.12-0.1-SNAPSHOT.jar ...
[info] Done packaging.
[info] Compiling 2 Scala sources to /data/rw/code/scala/scala-profiling/plugin/target/scala-2.12/test-classes...
[success] Total time: 125 s, completed Aug 3, 2017 1:29:56 PM
2. Waiting for source changes... (press enter to interrupt)
[info] Packaging /data/rw/code/scala/scala-profiling/plugin/target/scala-2.12/scalac-profiling_2.12-0.1-SNAPSHOT.jar ...
[info] Done packaging.
[info] Compiling 2 Scala sources to /data/rw/code/scala/scala-profiling/plugin/target/scala-2.12/test-classes...
[success] Total time: 108 s, completed Aug 3, 2017 1:31:55 PM
3. Waiting for source changes... (press enter to interrupt)
```
  
I've tried this with Typelevel Scala too and there's no observable difference between the two, both compile in 6 seconds (!).
  
I considered important to let everyone in this ticket know about the results of my experiment. Is this a well-known behaviour? It's the first time I see it, and it doesn't make sense that turning `L` contravariant speeds up implicit search -- should be the other way around. I wonder, is there something in the implicit search algorithm that we're not considering?

milessabin:
N=30? Your numbers don't seem quite right to me ... for Lightbend Scala (ie. without `-Yinductive-implicits`) I don't see compile times like that until about N=300.

Ahh ... except that my numbers are from compilng with bare scalac ... is it possible that you're seeing some SBT related overhead?

jvican:
Yes, I suppose it's because of an overhead of the incremental compiler. I'm using 0.13.16. Also, my sbt build for the project I'm doing is quite bloated because I have an sbt source dependency on Circe for integration tests (which could be increasing memory consumption and forcing GC to kick in more times).

jvican:
Oh, no, sorry -- I meant N=*300*. I was only counting from 10 to 10.

milessabin:
That's more like it :-)

fommil:
@milessabin did this [separate PR](https://github.com/scala/scala/pull/5649#issuecomment-296151148) ever happen? Or has this effort been shelved for now?

milessabin:
I'm working on it now.

milessabin:
Rebased ...

threeseed:
What's the latest on this ? .. Seems like such an amazing feature to have landed.

milessabin:
@adriaanm I'm going to rebase on top of https://github.com/scala/scala/pull/6050 so that I can make some progress on this.

milessabin:
Moved to https://github.com/scala/scala/pull/6481.

PR #5652, opened 2017-01-18
Array slice optimisations

mkeskells:


Ichoran:
This looks like a good approach in principle but there seem to be issues with return types and presence or absence of a `ClassTag` that makes me think it could not work as written.  Have you compiled this locally?

mkeskells:
Hi

I thought so but apparently not – have fixed the code now

 

Please note the handling of Unit is different from the other cases  as detailed in the comment

Otherwise there is a compile error 

[error] C:\Users\dev\Desktop\code\scalac\scalac_perf\src\library\scala\collection\mutable\WrappedArray.scala:235: type mismatch;

[error]  found   : Array[Unit]

[error]  required: Array[Unit]

[error] Note: Unit >: Unit, but class Array is invariant in type T.

[error] You may wish to investigate a wildcard type such as `_ >: Unit`. (SLS 3.2.10)

[error]       new ofUnit(util.Arrays.copyOfRange[Unit](array, from, until))

[error]

 

So assuming that Array[Unit] always contains (java) null only 

Should I change the path to needlessly fill Unit onto the result?

 

I will attach some benchmark results later

Prob next week will go through all of the PR and do this 

 

Paid work gets in the way …..

 

Regards

Mike

src/library/scala/collection/mutable/ArrayOps.scala
```diff
@@ -296,6 +335,12 @@ final class ofUnit(override val repr: Array[Unit]) extends AnyVal with ArrayOps[
     override protected[this] def thisCollection: WrappedArray[Unit] = new WrappedArray.ofUnit(repr)
     override protected[this] def toCollection(repr: Array[Unit]): WrappedArray[Unit] = new WrappedArray.ofUnit(repr)
     override protected[this] def newBuilder = new ArrayBuilder.ofUnit
+    protected override def emptyImpl = emptyUnitArray
+    protected override def sliceImpl(from: Int, until: Int) = {
+      // cant use util.Arrays.copyOfRange[Unit](repr, from, until) - Unit is special and doesnt compile
+      // but Unit and null are analogous so values are not important to copy across
+      Array[Unit](until-from)
```
Ichoran: I think you missed the `new`.  As written, this discards the value `until-from` to produce a single unit, then creates an array of length one.  There should be tests of all the slices for all the different implementations to catch stuff like this!

Ichoran: If you wanted to actually copy the array (to preserve whether the contents are `null` or instances of `scala.runtime.BoxedUnit`--perversely, either one is okay--you could
```
java.util.Arrays.copyOfRange(
  repr.asInstanceOf[Array[scala.runtime.BoxedUnit]],
  from,
  until
).asInstanceOf[Array[Unit]]
```
(maybe on a single line--I'm just spreading it out here to make it easier to see).

src/library/scala/collection/mutable/WrappedArray.scala
```diff
@@ -121,73 +131,108 @@ object WrappedArray {
...
     def elemTag = ClassTag.Unit
     def length: Int = array.length
     def apply(index: Int): Unit = array(index)
     def update(index: Int, elem: Unit) { array(index) = elem }
+    protected override def emptyImpl = emptyWrappedUnit
+    protected override def sliceImpl(from: Int, until: Int) = {
+      // cant use
+      // new ofUnit(util.Arrays.copyOfRange[Unit](array, from, until)) - Unit is special and doesnt compile
+      // but Unit and null are analogous so values are not important to copy across
+      new ofUnit(Array[Unit](until-from))
```
Ichoran: Same mistake here as with the ArrayOps version.

Ichoran:
This looks pretty good!  Benchmarks for this one are less important, as it is super-obvious that this will be a big win in many cases.  However, unit tests are very important to make sure that now, and in the future, no indices have been messed up.

src/library/scala/collection/mutable/WrappedArray.scala
mkeskells: doh

mkeskells:
HiI had assumed that the existing unit tests for the collections api would cover itIf not then I presume that we should pr a set of tests
Several of these products that we have and have not yet submitted are collected turn issueso and tweaks
What is the expectation for exception behaviour it is isn't in the scala docEg myseq(-1)Is it documented elsewhereRegardsMike

Sent from Yahoo Mail on Android 
 
  On Thu, 19 Jan 2017 at 1:39, Ichoran<notifications@github.com> wrote:   
This looks pretty good! Benchmarks for this one are less important, as it is super-obvious that this will be a big win in many cases. However, unit tests are very important to make sure that now, and in the future, no indices have been messed up.

Ichoran:
You should verify that a proper unit test exists, or write a new one.  You can't automatically assume that when you split one implementation into ten that the unit tests will properly cover all ten cases.  When everything shared an implementation, a single unit test with a single data type would have been enough.

I'm not sure understand what you were asking about "expectation for exception behavior".  `slice` is generally considered a "safe" operation; it won't throw an exception but rather will clip things to sensible ranges.  So for instance `a.slice(-8, -3)` will return an empty array rather than throwing an exception.  Is this what you meant?

mkeskells:
That was what I meant

I have added some initial unit test. I could not see any existing test to extend

Please can you review these tests to see if they look right

I am not convinced that the Array[Unit] test will work

Not sure how to check that BoxedUnit vs null in the arrays though, without resorting to java! 

Tests pass locally against 2.11.8 without my mods. Ran out of time to test with the mods, but thought that it was better to give some sight of these and see if they were in the right direction/sufficient etc

mkeskells:
I will remove the isLenient etc checks in next push

Ichoran:
Looks very extensive!  That's good.  I can't eyeball whether you've successfully gotten around generic code making everything look like Object,  but it looks like you probably have.  (It would be more idiomatic to use ClassTag, but I think it's okay anyway.)

I don't have time right now to think about binary compatibility (which is why the test failed).  That could be a problem.  Maybe someone else can pick this up?  @szeiger ?

szeiger:
```
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofChar is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofShort is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofUnit is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofInt is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofBoolean is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofDouble is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofRef is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofByte is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofLong is declared final in new version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.ArrayOps#ofFloat is declared final in new version
```

These look harmless and could simply be whitelisted. They are members of final classes, so the methods are effectively final anyway. MiMa shouldn't report them in the first place.

```
 * method sliceImpl(Int,Int)java.lang.Object in trait
   scala.collection.mutable.ArrayOps does not have a correspondent in old
   version
 * method emptyImpl()java.lang.Object in trait
   scala.collection.mutable.ArrayOps does not have a correspondent in old
   version
 * method slice(Int,Int)java.lang.Object in class
   scala.collection.mutable.WrappedArray is declared final in new version
 * abstract method sliceImpl(Int,Int)scala.collection.mutable.WrappedArray
   in class scala.collection.mutable.WrappedArray does not have a correspondent
   in old version
 * abstract method emptyImpl()scala.collection.mutable.WrappedArray in class
   scala.collection.mutable.WrappedArray does not have a correspondent in old
   version
```

These changes break binary compatibility in a fundamental way. You can never add a new abstract method or make a non-final method final. For 2.11 & 2.12 you could do these changes in a new private implementation class inside the `ArrayOps` and `WrappedArrray` companion objects (where the concrete subclasses that use the new implementation are also defined). It can be refactored back to the current version in 2.13.

mkeskells:
I have updated the patch to, hopefully, comply with binary compatibility needs

I have taken the liberty of marking WrappedArray as @deprecatedInheritance like ArrayOps. Happy to remove this if they should not be in step with ArrayOps, but it seemed sane to me

What is the appropriate way to provide a markup for the changes at 2.13 as proposed above? For the moment I have just added comments //at 2.13 ....

I have also moved the regression test to scala.collections, and added some support ofor other classes that follow a similar pattern
It seems sensible to be that this regression test should be the subject of a separate PR, otherwise it proves nothing, just that the code with my chnages matches some new test
If this is a separate PR then it shows that this is a valid representation of the current state and acts as a regression test to my changes. Would approciate your thoughts of if this is needed of it is easier for you to validate the current functionallity in some other way

retronym:
> I have taken the liberty of marking WrappedArray as @deprecatedInheritance like ArrayOps. Happy to remove this if they should not be in step with ArrayOps, but it seemed sane to me

Typically we don't introduce deprrcations in minor releases, so I'd leave this until 2.13.x.

> What is the appropriate way to provide a markup for the changes at 2.13 as proposed above? For the moment I have just added comments //at 2.13 ....

You can just keep the extra commit that will eventually target 2.13.x on a separate branch in your repository and link to it from here.

> It seems sensible to be that this regression test should be the subject of a separate PR, otherwise it proves nothing, just that the code with my chnages matches some new test

You can add the test for existing functionality in the first commit of this PR, then change the implementation in the second. Our pull request validation checks that each commit passes the test suite.

mkeskells:
HiWhen we can agree that the problem contains the correct set of changes to the runtime and the tests I can either remove all the commits and apply in this order or generate a new pr with just these commitsMike

Sent from Yahoo Mail on Android 
 
  On Tue, 24 Jan 2017 at 23:06, Jason Zaugg<notifications@github.com> wrote:   

It seems sensible to be that this regression test should be the subject of a separate PR, otherwise it proves nothing, just that the code with my chnages matches some new test


You can add the test for existing functionality in the first commit of this PR, then change the implementation in the second. Our pull request validation checks that each commit passes the test suite.

szeiger:
The changes for binary compatibility are not enough. Adding a protected concrete trait method is not even backward compatible in 2.11 (but it is in 2.12). It is not forward compatible in either version. You really need to move all of that stuff down into a private implementation class and keep the trait as it is.

mkeskells:
restructured the commits as first one with the test code and second with the changes as @retronym suggested
 updated the trait based chnages to includes a seperate trait for the extension as @szeiger suggested. Cant use a class as I see it because of the issues of extending from AnyVal

adriaanm:
The test failure is likely due to your branch being based on an old version of 2.11.x. To get up to date, could you do a `git pull --rebase https://github.com/scala/scala.git 2.11.x` in the `arraySlice` branch and `git push -f https://github.com/rorygraves/scala.git arraySlice`? Thanks!

mkeskells:
applied the git foo suggested by @adriaanm

szeiger:
Looks good. Please add the remaining problem filters from the log to `bincompat-backward.whitelist.conf` to make CI happy.

mkeskells:
Is there anything that I need to do here?
@szeiger I presume that the filters as part of the build system, not part of this commit

adriaanm:
The filters are part of the code tree: https://github.com/scala/scala/blob/2.11.x/bincompat-backward.whitelist.conf

the items to add to `problems` are:
```
            {
                matchName="scala.collection.mutable.ArrayOps#ofChar.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofShort.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofUnit.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofInt.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofBoolean.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofDouble.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofRef.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofByte.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofLong.slice"
                problemName=FinalMethodProblem
            },
            {
                matchName="scala.collection.mutable.ArrayOps#ofFloat.slice"
                problemName=FinalMethodProblem
            }
```

adriaanm:
See #5664 for a consolidated version with test & mima fixes

adriaanm:
Commits moved to #5664. Thanks!

PR #5214, opened 2016-06-04
SI-7898 Preserve reader against subversion

som-snytt:
SBT tries to install its own SimpleReader (for some reason)
if it needs to create its own IMain.

Because Rube Goldberg needs to execute some postinit hooks,
don't let SBT do that.

I tested this on locally built sbt 13.12 and scala 2.12.

```
apm@mara:~/projects/sample$ sbt console
[info] Set current project to Sample (in build file:/home/apm/projects/sample/)
[info] Compiling 1 Scala source to /home/apm/projects/sample/target/scala-2.12.0-M4/classes...
[info] Starting scala interpreter...
[info] 
Welcome to Scala 2.12.0-20160604-133738-e33ff9e (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_60).
Type in expressions for evaluation. Or try :help.

scala> 42
res0: Int = 42

scala> 43
res1: Int = 43

scala> 44
res2: Int = 44

scala> 45
res3: Int = 45

scala> 

scala> 

scala> println("hello")
hello

scala> 
[success] Total time: 379 s, completed Jun 4, 2016 3:07:28 PM
apm@mara:~/projects/sample$ cat build.properties 
sbt.version=0.13.12-SNAPSHOT
```

PR #5655, opened 2017-01-19
SI-10102 REPL ignores sbt init errors

som-snytt:
sbt runs user init code early, while creating its
custom `IMain`. Failures in this (arbitrary) code
keep the console from starting, including (silent)
errors due to bugs. Ignore silent errors emitted
on startup when sbt is running.

### DIFF
--- a/src/repl/scala/tools/nsc/InterpreterLoop.scala
+++ b/src/repl/scala/tools/nsc/InterpreterLoop.scala
@@ -3,10 +3,15 @@ package scala.tools.nsc
 import interpreter._
 import java.io._
 
-/** A compatibility stub.
+/** A compatibility stub for sbt.
  */
-@deprecated("Use a class in the scala.tools.nsc.interpreter package.", "2.9.0")
+@deprecated("Use scala.tools.nsc.interpreter.ILoop.", "2.9.0")
 class InterpreterLoop(in0: Option[BufferedReader], out: PrintWriter) extends ILoop(in0, out) {
   def this(in0: BufferedReader, out: PrintWriter) = this(Some(in0), out)
   def this() = this(None, new PrintWriter(scala.Console.out))
+
+  override protected final val isSbt = true
+
+  @deprecated("use `process` instead", "2.9.0")
+  def main(settings: Settings): Unit = process(settings) //used by sbt
 }
--- a/src/repl/scala/tools/nsc/interpreter/ILoop.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ILoop.scala
@@ -41,10 +41,7 @@ import Completion._
  *  @author  Lex Spoon
  *  @version 1.2
  */
-class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
-                extends AnyRef
-                   with LoopCommands
-{
+class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter) extends LoopCommands {
   def this(in0: BufferedReader, out: JPrintWriter) = this(Some(in0), out)
   def this() = this(None, new JPrintWriter(Console.out, true))
 
@@ -56,6 +53,9 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
   var intp: IMain = _
 
   private var globalFuture: Future[Boolean] = _
+
+  // ignore silent sbt errors on init
+  protected def isSbt: Boolean = false
 
   /** Print a welcome message! */
   def printWelcome(): Unit = {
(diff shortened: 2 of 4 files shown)


SethTisue:
can you suggest a reviewer who could consider this from the sbt side...?

som-snytt:
@dwijnand still owes me one for https://github.com/scala/scala/pull/5214

som-snytt:
Also, instead of deprecating `InterpreterLoop`, it should be considered the canonical place to workaround the sbt code. (I intended to donate upgraded `ConsoleInterface` to sbt.)

dwijnand:
I'm not sure. On all front to be honest. On (a) what broke it, (b) what exactly happens, and (c) fixing it like that. Sorry :-/

som-snytt:
(a) previously, I had refactored how REPL startup happens to be more sane. (b) it checks whether the reporter saw errors during startup. Since sbt executes arbitrary code during startup, that is not unlikely. (c) How about we not fail to start in that case. The reporter still knows it saw an error. Reporters like elephants never forget.

But wait, at this point, I want to refactor some imports to use trailing commas.

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -29,8 +29,22 @@ class ReplReporter(intp: IMain) extends ConsoleReporter(intp.settings, Console.i
...
   override def warning(pos: Position, msg: String): Unit = withoutTruncating(super.warning(pos, msg))
-  override def error(pos: Position, msg: String): Unit   = withoutTruncating(super.error(pos, msg))
+  override def error(pos: Position, msg: String): Unit   = {
+    val count0 = errorCount
+    withoutTruncating(super.error(pos, msg))
+    val count1 = errorCount
+    if (count1 > count0 && intp.totalSilence)
+      silentErrors += 1
+  }
+  def reportableErrorCount = errorCount - silentErrors
```
adriaanm: Let's make these additions `private[interpreter]`.

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -29,8 +29,22 @@ class ReplReporter(intp: IMain) extends ConsoleReporter(intp.settings, Console.i
...
   }
 
+  private var silentErrors = 0
   override def warning(pos: Position, msg: String): Unit = withoutTruncating(super.warning(pos, msg))
-  override def error(pos: Position, msg: String): Unit   = withoutTruncating(super.error(pos, msg))
+  override def error(pos: Position, msg: String): Unit   = {
+    val count0 = errorCount
+    withoutTruncating(super.error(pos, msg))
+    val count1 = errorCount
+    if (count1 > count0 && intp.totalSilence)
```
adriaanm, line 38: Does the reporter contract guarantee `0 <= count1 - count0 <= 1`? I guess so.

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
adriaanm: Let's make these additions `private[interpreter]`.

adriaanm (changes requested):
You've unlocked the pragmatic fix badge! Thanks for moving us in the right direction here. Eventually, closer cooperation of sbt & scalac is the ticket. A few small requested tweaks, if I may. Also, in future, please consider separate boy scout and bug fix commits.

som-snytt:
Made `isSbt` a member overridden in `InterpreterLoop` and also pushed down `main`, which may trigger mima.

Members of `ReplReporter` are more private.

Rebased.

src/repl/scala/tools/nsc/InterpreterLoop.scala
```diff
@@ -3,10 +3,15 @@ package scala.tools.nsc
...
-/** A compatibility stub.
+/** A compatibility stub for sbt.
  */
-@deprecated("Use a class in the scala.tools.nsc.interpreter package.", "2.9.0")
+@deprecated("Use scala.tools.nsc.interpreter.ILoop.", "2.9.0")
 class InterpreterLoop(in0: Option[BufferedReader], out: PrintWriter) extends ILoop(in0, out) {
   def this(in0: BufferedReader, out: PrintWriter) = this(Some(in0), out)
   def this() = this(None, new PrintWriter(scala.Console.out))
+
+  override protected final val isSbt = true
```
dwijnand, line 13: Given Adriaan's previous comment, at a guess I would think this should be `protected[nsc]` (or tools or scala).

nsc is principle of least power and scala is principle of maximal flexibility?

dwijnand (approved):
Thanks Som for your explanation and sorry for the late review. This needs rebasing.

I've noted a little concern, if you agree then I guess you could fix it up while rebasing.

It's not a blocker for me though. LGTM.

dwijnand:
It's a shame we can't create a test case to reproduce the original bug and stand as a regression test.

retronym:
This PR needs a rebase to resolve a conflict with #5727

som-snytt:
Probably I'll rebase atop `Adriaanite`?

SethTisue:
@som-snytt rebase?

lrytz:
Adriaanite is 2.13, so it cannot bite here. I think it would be good to get this fix in for 2.12.3, as the error feedback for affected users is non-obvious.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -988,15 +988,15 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
           createInterpreter()
         }
         intp.initializeSynchronous()
-        globalFuture = Future successful true
-        if (intp.reporter.hasErrors) {
+        globalFuture = Future.successful(true)
+        if (intp.reporter.hasErrors && (!isSbt || intp.reporter.hasReportableErrors)) {
           echo("Interpreter encountered errors during initialization!")
```
lrytz, line 993: Should we actually show (some of) the errors? How can I trigger this line?

lrytz (approved):

PR #5656, opened 2017-01-22
SI-10159: Don't report cyclic reference errors found in `isCyclicOrErroneous`.

ghost:
Previously, while the implicit would rightly be discarded, `typedInternal` would catch the CyclicReference error and call `reportTypeError` before rethrowing, which would cause compilation to fail.

This should help fix https://github.com/milessabin/shapeless/issues/679.

Review by @adriaanm, @retronym

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -226,8 +226,9 @@ trait Implicits {
     }
 
     private[this] final def computeIsCyclicOrErroneous =
-      try sym.hasFlag(LOCKED) || containsError(tpe)
-      catch { case _: CyclicReference => true }
+      try global.withPropagateCyclicReferences {
+        sym.hasFlag(LOCKED) || containsError(tpe)
+      } catch { case _: CyclicReference => true }
 
     var useCountArg: Int = 0
     var useCountView: Int = 0
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -221,14 +221,21 @@ trait Implicits {
     }
 
     def isCyclicOrErroneous: Boolean = {
-      if (!isCyclicOrErroneousCache.isKnown) isCyclicOrErroneousCache = computeIsCyclicOrErroneous
-      isCyclicOrErroneousCache.booleanValue
+      if (!isCyclicOrErroneousCache.isKnown)
+        try global.withPropagateCyclicReferences {
+          val cyclicOrErroneous = sym.hasFlag(LOCKED) || containsError(tpe)
+          isCyclicOrErroneousCache = cyclicOrErroneous
+          cyclicOrErroneous
+        } catch {
+          case _: CyclicReference =>
+            /* if the symbol was not locked but we got a cyclic reference error,
+             * it was likely from a different symbol and does not disqualify this one
+             * That said, it's still not good this time around */
+            true
+        }
+      else
+        isCyclicOrErroneousCache.booleanValue
     }
-
-    private[this] final def computeIsCyclicOrErroneous =
-      try global.withPropagateCyclicReferences {
-        sym.hasFlag(LOCKED) || containsError(tpe)
-      } catch { case _: CyclicReference => true }
 
     var useCountArg: Int = 0
     var useCountView: Int = 0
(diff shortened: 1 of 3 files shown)


retronym:
I believe this is an analogous test case that doesn't involve macros and Dynamic.

```scala
class D[T <: Bound](t: T) { type TT = T}
object Test {
  object C extends D(1)
  implicit val unrelated: List[C.TT] = null // error: illegal cyclic reference involving object C
}
class Bound
object Bound {
  implicit def AnyToBound[T](t: T): Bound = new Bound
}
```

Could you check whether this PR fixes it, too, and add it as an extra test case?

retronym:
My understanding of `ImplicitInfo#computeIsCyclicOrErroneous` is that even after you've suppressed the unwanted reporting of the cyclic error, the implicit will still be out of service as the `ImplicitInfo` is cached. I think there are different instances of `ImplicitInfo` to represent the in-scope implicits (`Context#implicitss`) vs the companion scope implicits (`Implicits#implicitsOfExpectedType`), so in- or out-of-service might vary depending on how you consume the implicit.

ghost:
That's a good point, about the caching. Currently, adding `val i = implicitly[List[K]]` to my test gives a "could not find implicit" error, but `val i: List[K] = implicitly[List[K]]` does not... it appears that the implicit cache gets cleared at some point.

I'm not quite sure what can be done about that: the most direct fix I can think of would be to not cache `isCyclicOrErroneous` for locked symbols, but that seems like it would come across as somewhat arcane. What do you think?

Also, I cannot get your test case to compile. On 2.12.1, the cyclic reference error appears, but with other errors as well:
```
<pastie>:13: error: inferred type arguments [Int] do not conform to class D's type parameter bounds [T <: Bound]
         object C extends D(1)
                          ^
<pastie>:13: error: type mismatch;
 found   : Int(1)
 required: T
         object C extends D(1)
```

However, with this change the cyclic reference error vanishes (but those two remain).

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -221,14 +221,22 @@ trait Implicits {
...
     def isCyclicOrErroneous: Boolean = {
-      if (!isCyclicOrErroneousCache.isKnown) isCyclicOrErroneousCache = computeIsCyclicOrErroneous
-      isCyclicOrErroneousCache.booleanValue
+      if (!isCyclicOrErroneousCache.isKnown)
+        try global.withPropagateCyclicReferences {
+          val cyclicOrErroneous = sym.hasFlag(LOCKED) || containsError(tpe)
+          isCyclicOrErroneousCache = cyclicOrErroneous
+          cyclicOrErroneous
+        } catch {
+          case _: CyclicReference =>
```
ghost, line 230: I believe that this is a correct test. If it's unavoidably a cyclic reference, then `sym` should be locked... otherwise we probably should consider it next time around.

I'm going to compile shapeless w/ and w/o this patch to see if this will cause a major performance regression. I believe it won't, however.

ghost: Hm. `validate-test` took somewhat longer on this commit than the previous... some others take longer... I can't tell if that's in normal variation or whatnot.

adriaanm: There's a lot of noise on the duration of tests. We do have benchmarking infra coming up soon /cc @retronym

adriaanm:
I've been digging around a bit (apologies for taking so long!), and it looks like your PR reveals some much deeper issues! When changing the type alias to

```
  type K = Record.bip.type#T // was `Record.bip.T`
```

the example compiles without error, even though that should be an equivalent formulation (it works because implicit search is correctly disabled when typing `Record.bip.type`). That's the first issue (code duplication ftw): typing a `Select(qual, name)`, where `name.isTypeName`, should behave the same as `SelectFromTypeTree(SingletonTypeTree(qual), name)`.

Secondly, and sadly, I think neither version should compile, as the `[select|apply]Dynamic` expansion was never intended to happen in types... Without macros, it wasn't possible to escape the stable identifier requirement, so it didn't occur to me to check that explicitly.

/cc @odersky for advice on wwdd (what would dotty do)

adriaanm:
> Secondly, and sadly, I think neither version should compile,

To clarify, for backwards compatibility we won't disallow this in 2.12.x, but I expect this loophole to be closed. I have a PR in the making that implements your initial observation that implicit search should not be used while type checking the qualifier of a type selection. (Which makes your test case compile)

ghost:
@adriaanm has looked deeper and found [a much nicer solution](https://github.com/scala/scala/pull/5762)... closing for now

PR #5657, opened 2017-01-23
SI-8521 ScalaClassLoader.asContext saves context

som-snytt:
It saves the actual current thread context loader,
not the loader it might biasedly prefer to have.

REPL uses asContext to preserve context, which broke
if the current loader didn't happen to be a
ScalaClassLoader.

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/IMain.scala
+++ b/src/repl/scala/tools/nsc/interpreter/IMain.scala
@@ -357,8 +357,7 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends 
       _runtimeClassLoader
     })
 
-  // Set the current Java "context" class loader to this interpreter's class loader
-  def setContextClassLoader() = classLoader.setAsContext()
+  def setContextClassLoader() = ()
 
   def allDefinedNames: List[Name]  = exitingTyper(replScope.toList.map(_.name).sorted)
   def unqualifiedIds: List[String] = allDefinedNames map (_.decode) sorted
--- a/src/repl/scala/tools/nsc/interpreter/IMain.scala
+++ b/src/repl/scala/tools/nsc/interpreter/IMain.scala
@@ -156,9 +156,9 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends 
   }
 
   import global._
-  import definitions.{ ObjectClass, termMember, dropNullaryMethod}
+  import definitions.{ObjectClass, termMember, dropNullaryMethod}
 
-  lazy val runtimeMirror = ru.runtimeMirror(classLoader)
+  def runtimeMirror = ru.runtimeMirror(classLoader)
 
   private def noFatal(body: => Symbol): Symbol = try body catch { case _: FatalError => NoSymbol }
 
@@ -357,7 +357,8 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends 
       _runtimeClassLoader
     })
 
-  def setContextClassLoader() = ()
+  @deprecated("The thread context classloader is now set and restored around execution of REPL line, this method is now a no-op.", since = "2.12.0")
+  def setContextClassLoader() = () // Called from sbt-interface/0.12.4/src/ConsoleInterface.scala:39
 
   def allDefinedNames: List[Name]  = exitingTyper(replScope.toList.map(_.name).sorted)
   def unqualifiedIds: List[String] = allDefinedNames map (_.decode) sorted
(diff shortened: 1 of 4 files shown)


som-snytt:
/rebuild

retronym (approved):
LGTM

retronym (dismissed):
Oh, wait, do we have the same problem in:

```scala
object ScalaClassLoader {
  def savingContextLoader[T](body: => T): T = {
    val saved = contextLoader
    try body
    finally setContext(saved)
  }
```

som-snytt:
I don't know why life can be so evil, but I'll give that API a haircut.

som-snytt:
Rebased, and I think I removed the offending method, which was a stop-gap. Since REPL now restores context class loader in a principled way, the stop-gap is not required.

I don't really know what "principled" means, except that retronym uses that word approvingly, so I hope it helps my case. It might mean, We promise to restore the context class loader. Right brace means "rewind" as in "Please be kind, rewind" from the days of VHS.

src/repl/scala/tools/nsc/interpreter/IMain.scala
```diff
@@ -359,8 +359,7 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends
       _runtimeClassLoader
     })
 
-  // Set the current Java "context" class loader to this interpreter's class loader
-  def setContextClassLoader() = classLoader.setAsContext()
+  def setContextClassLoader() = ()
```
lrytz: mark `@deprecated`?

lrytz (commented):
> REPL now restores context class loader in a principled way

Can you provide a reference?

src/repl/scala/tools/nsc/interpreter/IMain.scala
som-snytt: I went too far; I'll back off the changes. Comment on another PR says sbt uses this. (Or maybe that's OK.)

som-snytt:
Saving context class loader was part of the original contribution for scripting:

https://github.com/scala/scala/blob/v2.12.1/src/repl/scala/tools/nsc/interpreter/IMain.scala#L563

I found the previous topic:

https://github.com/scala/scala/pull/4841

I'll refresh my memory and also maybe take up that old PR.

som-snytt:
As seen in https://github.com/scala/scala/pull/5655 sbt runs `initialCommands` before `process` so prophylactic `savingContextClassLoader` doesn't provide extra security.

I'm not sure if leaving the API in `object ScalaClassLoader` is nice to do, even though it is nominally internal API.

som-snytt:
Just tacking on retronym's old fix, cherry not picked. I haven't tested yet.

SethTisue:
@som-snytt @cnestor you pinged me on Gitter about this. yes, let's make sure this makes 2.12.4. Som, can you rebase, and comment on whether this is otherwise ready for merge...?

som-snytt:
:notes: When you wish upon a star... :musical_note: :zzz:

som-snytt:
@SethTisue Thanks, I pushed something. As summary, `IMain` uses `asContext` to save/restore context class loader, and this is deemed correct. After fixing `asContext` to work correctly, remove the similar utility method and don't use it from `ILoop.process`.

som-snytt:
The failure mode is
```
Error: publish-core did not publish "jenkins.properties" artifact, unable to proceed.
ERROR: Failed to run DSL Script
groovy.lang.MissingPropertyException: No such property: Result for class: Script1
```

lrytz:
```
[error] /home/jenkins/workspace/scala-2.12.x-validate-publish-core/src/repl/scala/tools/nsc/interpreter/IMain.scala:1041: stable identifier required, but IMain.this.runtimeMirror.universe found.
[error]       import runtimeMirror.universe.{Symbol, InstanceMirror, TermName}
[error]                            ^
```

som-snytt:
I swear I did a `testOnly ScriptedTest` yesterday, but I just realized I was probably in the middle of the rebase and the last commit hadn't been applied yet. I only had a few minutes to spare; just to confirm, testOnly does actually compile repl. I think I need a tool to tell me if I'm pushing without at least compiling and running tests that are in the commit. In fact, to run those tests because I don't want to look them up. (Edit: yes, the window was still open with the testOnly success. Definitely need that tool. But what to call it? `gitcha` for git gotchas?)

som-snytt:
@retronym 's requested change is in. I think I'm _not_ supposed to "dismiss review", but the ball is in his court?

adriaanm:
Let's see what this dismiss thing can do for our workflow.

adriaanm:
I rerequested a rereview.

retronym (approved):

adriaanm:
This won't merge forward to 2.13.x -- will need a separate port if we need these changes there.

PR #5661, opened 2017-01-25
run ScalaCheck tests directly, not through partest

SethTisue:
ScalaCheck ever being under partest in the first place is ancient
history, from back in the Ant build days (shudder)

also upgrades ScalaCheck from 1.11.6 to 1.13.4, since we might
as well.  no source changes were necessary.

upgrades partest to 1.1.0, from which ScalaCheck support was
removed.


SethTisue:
fyi @rickynils

retronym:
@SethTisue I've taken this a bit further in https://github.com/retronym/scala/tree/review/5661

 - [x] The new scalacheck module inadvertently depends on partest via the scaladoc module. I shuffled some test glue code to a new home to fix this once and for all.
 - [x] Update the IntelliJ template config (both from this change and from some previous changes that weren't reflected). This is still a semi-manual process that needs some IntellIJ nous.
 - [x] Moves some tests into suitable packages, and then organized packaged tests into corresponding folders.

SethTisue:
very nice! I hadn't noticed the dependency thing. (the other things I knew about but figured could be future work...)

is this ready for me to pull in, or are you still tinkering? your commit e3d15e138aae3b853971e8bffac5249bb8881a05 lacks a commit message, perhaps you intended a squash or rebase

retronym:
> is this ready for me to pull in, or are you still tinkering?

I've just squashed that commit. Could you generated the IntelliJ config from clean to sanity check that it works on a machine other than mine? (`sbt intellijFromSample intellij; idea src/intellij/scala.ipr`, check that `test/scalacheck` is an IntelliJ module, and the sources there in have IDE smarts working.)

build.sbt
```diff
@@ -567,7 +567,20 @@ lazy val junit = project.in(file("test") / "junit")
     javaOptions in Test += "-Xss1M",
     libraryDependencies ++= Seq(junitDep, junitInterfaceDep, jolDep),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-v"),
-    // testFrameworks -= new TestFramework("org.scalacheck.ScalaCheckFramework"),
+    unmanagedSourceDirectories in Compile := Nil,
+    unmanagedSourceDirectories in Test := List(baseDirectory.value)
```
szeiger, line 571: What's the purpose of the new settings here?

test/files/scalacheck/parallel-collections/pc.scala
```diff
@@ -1,71 +0,0 @@
-/*
```
szeiger, line 1: This gets removed in https://github.com/scala/scala/pull/5603 (and reinstated as a pure ScalaCheck test in https://github.com/scala/scala-parallel-collections/blob/master/scalacheck/src/test/scala/pc.scala)

build.sbt
SethTisue: it's copied from the `junit` subproject. it's because the sources are all test sources and are all at the root level (of the `scalacheck` subproject's `baseDirectory`)

szeiger (commented):
n/a (github broke reviews without requested changes)

build.sbt
szeiger: This *is* the `junit` subproject

sjrd: Yes but if you look at the new `scalacheck` project, it also contains these two lines, in white. It's a diffing artifact that those in `junit` appear green and those in `scalacheck` appear white. Logically speaking, the ones in `scalacheck` should be green.
In other words, this looks good to me.

build.sbt
```diff
@@ -39,11 +39,11 @@ val scalaSwingDep                = scalaDep("org.scala-lang.modules", "scala-swi
 val scalaXmlDep                  = scalaDep("org.scala-lang.modules", "scala-xml")
 val scalaParserCombinatorsDep    = scalaDep("org.scala-lang.modules", "scala-parser-combinators")
 val partestDep                   = scalaDep("org.scala-lang.modules", "scala-partest",              versionProp = "partest")
-val scalacheckDep                = scalaDep("org.scalacheck",         "scalacheck",                 scope = "it")
 
 // Non-Scala dependencies:
 val junitDep          = "junit"                  % "junit"           % "4.11"
 val junitInterfaceDep = "com.novocode"           % "junit-interface" % "0.11"                            % "test"
+val scalacheckDep     = "org.scalacheck"         % "scalacheck_2.12" % "1.13.4"                          % "test"
```
sjrd, line 46: Would this be a problem in the 2.13 build?

build.sbt
szeiger: Oh, right, after looking closely at the diff for the 3rd time, I see it, too.

build.sbt
SethTisue: at the moment it will almost certainly work. when it does break, we will likely address it by re-insourcing ScalaCheck for a while again

sjrd (approved):

SethTisue:
> Could you generated the IntelliJ config from clean to sanity check that it works on a machine other than mine?

@retronym I've pulled in your commits, and yes, the IntellIiJ stuff works on my machine too

SethTisue:
I'll look into the Jenkins failures.

SethTisue:
@retronym does the discrepancy between "sources" and "test sources" here indicate a misconfiguration?

screen shot 2017-01-25 at 11 37 34 am

SethTisue:
The test failures are reproducible locally. The cause is that the new partest version also includes https://github.com/scala/scala-partest/pull/69, so our test cases here that depended on triple compilation need updating to explicitly request it. ~I'll take care of it.~ EDIT: Lukas took care of it already a while back elsewhere, I pulled in his commit.

retronym:
@SethTisue Yes, that's unintentional. It is a benign mistake, but you could fix it by hand-editing the `scalacheck.iml.SAMPLE` file after looking at the difference to `junit.iml.SAMPLE`.

SethTisue:
the test failure in Duration may be intermittent — I wasn't able to reproduce it locally. ~if the current run passes, I'll squash before merging.~

SethTisue:
rebased into what is hopefully a series of coherent commits that will all pass Jenkins, knock on wood

SethTisue:
@retronym Jenkins likes it, want to do the merge honors?

PR #4473, opened 2015-04-29
SI-9285 Don't warn about non-sensible equals in synthetic methods

retronym:
Notably, in the synthetic equals method of a case class. Otherwise,
we get an unsuppressable warning when defining a case class with a
`Unit`-typed parameter, which some folks use a placeholder for
real type while evolving a design.

PR #5662, opened 2017-01-26
Address SI-9675 Add warning about non-sensible equals in anonymous functions

teldosas:
The problem was that in an attempt to address an issue with #4473, warnings about non-sensible equals were removed for synthetic entities. But an anonymous function is also a synthetic entity, so warnings were removed for these too by mistake.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -1130,7 +1130,7 @@ abstract class RefChecks extends Transform {
     }
     /** Sensibility check examines flavors of equals. */
     def checkSensible(pos: Position, fn: Tree, args: List[Tree]) = fn match {
-      case Select(qual, name @ (nme.EQ | nme.NE | nme.eq | nme.ne)) if args.length == 1 && isObjectOrAnyComparisonMethod(fn.symbol) && !currentOwner.isSynthetic =>
+      case Select(qual, name @ (nme.EQ | nme.NE | nme.eq | nme.ne)) if args.length == 1 && isObjectOrAnyComparisonMethod(fn.symbol) && (!currentOwner.isSynthetic || currentOwner.isAnonymousFunction) =>
         checkSensibleEquals(pos, qual, name, fn.symbol, args.head)
       case _ =>
     }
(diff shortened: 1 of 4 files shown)


teldosas:
I signed the CLA

teldosas:
review @adriaanm 
(Is this how I'm supposed to ask for review?)

adriaanm:
Good catch! Looks like a good example of where we could sharpen the distinction between artifact and synthetic.

adriaanm (changes requested):
The fix looks good, but the test could be rewritten as a neg/ test instead of a run/ one, right? You'll need to create a neg/t9675.flags file that contains `-Xfatal-warnings`, but that should be it.

teldosas:
@adriaanm But that does not guarantee that we get a warning for every single function. It would be enough to get a warning for a single function for it to succeed.

adriaanm:
It does, as far as I can see -- neg tests will check that exactly the
errors in the check file are emitted. There's no need to run your app to
get the errors, just compile it

teldosas:
Ok I made the changes. Now it is more clear to me what "run" means. Thanks @adriaanm

adriaanm:
Sure! Thanks for the quick follow up.

adriaanm:
How is this test related to this PR?

```
[[syntax trees at end of                    parser]] // client_2.scala
package <empty> {
  object Test extends App {
    def <init>() = {
      super.<init>();
      ()
    };
    val wit = Random.foo("1");
    var t = 1;
    var one: wit.T = 1
  }
}

[[syntax trees at end of                     namer]] // client_2.scala: tree is unchanged since parser
!!  409 - run/random                                [compilation failed]
```

teldosas:
@adriaanm it doesn't it's about another bug that I was working on, I pushed it by mistake. I will remove it

teldosas:
@adriaanm I removed it. It should be fine now

adriaanm (approved):

adriaanm:
Thanks, @teldosas! For your next fix, please use a commit title more like "SI-9675 warn about non-sensible equals in anonymous functions" (start with ticket number, use active voice)

som-snytt:
There was no adequate emoji to say, "oh, artifact vs synthetic, I was just wondering about that."

PR #5663, opened 2017-01-27
SD-256 enable colored output by default on unix

gourlaysama:
This enables `-Dscala.color` by default when running via the shell
script and when stdout is a terminal.

Can be overriden on the CLI or via `JAVA_OPTS`:

```sh
$ qbin/scala -nc -e 'println(util.Properties.propOrFalse("scala.color"))'
true
$ qbin/scala -nc -e 'println(util.Properties.propOrFalse("scala.color"))' | cat
false
$ qbin/scala -nc -Dscala.color=false -e 'println(util.Properties.propOrFalse("scala.color"))'
false
$ JAVA_OPTS="-Dscala.color=false" qbin/scala -nc -e 'println(util.Properties.propOrFalse("scala.color"))'
false
```

### DIFF
--- a/bincompat-forward.whitelist.conf
+++ b/bincompat-forward.whitelist.conf
@@ -423,6 +423,14 @@ filter {
     {
       matchName="scala.annotation.showAsInfix"
       problemName=MissingClassProblem
+    },
+    {
+      matchName="scala.util.PropertiesTrait.coloredOutputEnabled"
+      problemName=DirectMissingMethodProblem
+    },
+    {
+      matchName="scala.util.Properties.coloredOutputEnabled"
+      problemName=DirectMissingMethodProblem
     }
   ]
 }
--- a/src/library/scala/util/Properties.scala
+++ b/src/library/scala/util/Properties.scala
@@ -154,6 +154,12 @@ private[scala] trait PropertiesTrait {
   /* Some runtime values. */
   private[scala] def isAvian = javaVmName contains "Avian"
 
+  private[scala] def coloredOutputEnabled: Boolean = propOrElse("scala.color", "auto") match {
+    case "auto" => System.console() != null && !isWin
+    case a if a.toLowerCase() == "true" => true
+    case _ => false
+  }
+
   // This is looking for javac, tools.jar, etc.
   // Tries JDK_HOME first, then the more common but likely jre JAVA_HOME,
   // and finally the system property based javaHome.
--- a/src/reflect/scala/reflect/internal/TypeDebugging.scala
+++ b/src/reflect/scala/reflect/internal/TypeDebugging.scala
@@ -59,7 +59,7 @@ trait TypeDebugging {
   object typeDebug {
     import scala.Console._
 
-    private val colorsOk = sys.props contains "scala.color"
+    private val colorsOk = scala.util.Properties.coloredOutputEnabled
     private def inColor(s: String, color: String) = if (colorsOk && s != "") color +        s + RESET else s
     private def inBold(s: String, color: String)  = if (colorsOk && s != "") color + BOLD + s + RESET else s
 
--- a/src/repl/scala/tools/nsc/interpreter/ReplProps.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ReplProps.scala
@@ -17,7 +17,7 @@ class ReplProps {
   private def int(name: String)  = Prop[Int](name)
 
   // This property is used in TypeDebugging. Let's recycle it.
-  val colorOk = bool("scala.color")
+  val colorOk = Properties.coloredOutputEnabled
 
   val info  = bool("scala.repl.info")
   val debug = bool("scala.repl.debug")


adriaanm:
Sweet! Thanks! Anyone out there familiar with unix gotchas like this? I'm not qualified to review, but would love to get this in!

src/compiler/scala/tools/ant/templates/tool-unix.tmpl
```diff
@@ -127,6 +127,11 @@ fi
 
 [[ -n "$JAVA_OPTS" ]] || JAVA_OPTS="@javaflags@"
 
+# enable colored output by default
+# we prepend it to JAVA_OPTS so users can override it on the
+# command line or via JAVA_OPTS
+[[ -t 1 ]] && JAVA_OPTS="-Dscala.color=true $JAVA_OPTS"
```
dwijnand: Rather than mess directly with JAVA_OPTS I would follow the precendence set by `-Dscala.usejavacp`/`OVERRIDE_USEJAVACP`

gourlaysama: I originally didn't want to clutter the command line to run with one more variable, but you're right, it's cleaner that way. Done.

smarter:
That only solves part of the problem, it would also be nice to get colors when running `sbt console` for example, for this the default value of `scala.color` should be `auto` which would behave like `grep --color=auto` and many other tools, but this means that the terminal detection must be done on the JVM.

smarter:
`System.console` seems to be the way to do that: https://stackoverflow.com/questions/1403772/how-can-i-check-if-a-java-programs-input-output-streams-are-connected-to-a-term

gourlaysama:
The original reason I did it in the bash script was that we only want to activate it by default on unix-like things.

Hmm, having an `auto` mode that tests `System.console` would indeed make it work all on its own in sbt *and* with the bash runner. Then I can add a `-Dscala.color=false` default in the bat runner for windows.
Maybe that's a better solution?

smarter:
I think the Windows console supports color now, though I have no idea how to test for that: https://blogs.msdn.microsoft.com/commandline/2016/09/22/24-bit-color-in-the-windows-console/ Simply disabling color in the batch script might be enough.

smarter:
Or always disable if the platform is windows: http://www.scala-lang.org/api/2.12.0/scala-compiler/scala/tools/nsc/Properties$.html#isWin:Boolean

gourlaysama:
> Or always disable if the platform is windows

Oh, right, I'll just do that.

> I think the Windows console supports color now, though I have no idea how to test for that

After [a bit of digging][2]: it is enabled by default when used from the Windows Subsystem for Linux, but not for normal windows console programs run directly or from `cmd` or powershell. `cmd.exe` does enable it for itself, so it works within a `.bat` script, but `java.exe` doesn't, and that property isn't inherited by child processes for compatibility reasons. So no color out of the box for the normal usecase. We could enable it, but it is [a native API][1], so JNI and all that... let's just keep it disabled on Windows.

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/ms686033%28v=vs.85%29.aspx
[2]: https://github.com/Microsoft/BashOnWindows/issues/1173#issuecomment-254250445

gourlaysama:
@smarter done. Now it works in SBT too :)

I moved the property somewhere more central since it is already used in two places. The next step would be to use it for a few more things, like the default `ConsoleReporter`, and maybe more things in the repl.

gourlaysama:
```
library/mima failed: java.lang.RuntimeException: MiMa failed with exit code 2
```
😦

adriaanm:
Awesome! Yeah, scala's package-private can't be represented in byte code. Will need whitelisting :-/

lrytz:
LGTM - needs a rebase unfortunately

gourlaysama:
Rebased.

lrytz:
/rebuild - interesting failure

```
Caused by: sbt.ForkMain$ForkError: java.lang.ArrayIndexOutOfBoundsException: 15
	at java.util.ArrayList.add(ArrayList.java:459)
	at sbt.ForkMain$Run$2$1.handle(ForkMain.java:294)
	at scala.tools.partest.sbt.SBTRunner$$anon$1.onFinishTest(SBTRunner.scala:70)
	at scala.tools.partest.nest.SuiteRunner.runTest(Runner.scala:781)
	at scala.tools.partest.nest.SuiteRunner.$anonfun$runTestsForFiles$2(Runner.scala:788)
	at scala.tools.partest.package$$anon$2.call(package.scala:135)
```

adriaanm:
🎉

SethTisue:
> Now it works in SBT too

confirmed. 🙌

som-snytt:
Just noticed my scripts with `-Dscala.color` stopped working. That is, it's no longer taken as `-Dscala.color=true` or even `auto`.

PR #5664, opened 2017-01-28
Optimise common operations on Array and List

adriaanm:
Consolidates the following PRs:
- List.mapConserve avoids ListBuffer creation (#5650)
- Array slice optimisations (#5652)
- Optimised implementation of List.filter/filterNot (#5653)


adriaanm:
I suggest we either close the other PRs, merge this one, and add the benchmarks, or incorporate these commits into one of the other PRs and close this one.

rorygraves:
@adriaanm  The List.filter/filterNot benchmark already exists on the https://github.com/scala/scala/pull/5653 PR ListBenchmark.scala

adriaanm:
Ok, I wasn't clear whether that was the last word on that. I cherry-picked the commits, so then the benchmark is in here too.

adriaanm:
Mima should be happy with these new commits. I haven't looked into whether each whitelist item is good to go. Leaving that up to @szeiger

rorygraves:
Looks good.

adriaanm:
Thanks, sounds like I can safely close the other PRs, since all their commits have been included here? Will get this one merged asap.

rorygraves:
Yes they can all be safely closed. 
:)

adriaanm:
@szeiger, are you ready to sign off on this? /cc @retronym

bincompat-forward.whitelist.conf
```diff
@@ -628,6 +628,319 @@ filter {
...
+        matchName="scala.collection.mutable.WrappedArray#ofShort.sliceImpl"
+        problemName=DirectMissingMethodProblem
+    },
+    {
+        matchName="scala.collection.mutable.WrappedArray#ofShort.emptyImpl"
+        problemName=DirectMissingMethodProblem
+    },
+    {
+        matchName="scala.collection.mutable.WrappedArray.sliceImpl"
+        problemName=DirectMissingMethodProblem
```
szeiger: This is a breaking change. WrappedArrray is public and the new method is protected. As discussed before, all these changes need to be moved down into a private implementation class.

bincompat-forward.whitelist.conf
```diff
@@ -628,6 +628,319 @@ filter {
...
+    },
+    {
+        matchName="scala.collection.mutable.WrappedArray.emptyImpl"
+        problemName=DirectMissingMethodProblem
```
szeiger: Same here.

bincompat-forward.whitelist.conf
```diff
@@ -628,6 +628,319 @@ filter {
...
+    },
+    {
+        matchName="scala.collection.mutable.WrappedArray.slice"
+        problemName=IncompatibleResultTypeProblem
```
szeiger: This is also a problem.

szeiger (changes requested):

szeiger:
I'll try to fix these remaining issues after lunch. Should be ready in a few hours.

szeiger:
https://github.com/adriaanm/scala/pull/26

adriaanm:
@szeiger I merged your fixes for your own review -- thanks!

szeiger (approved):

adriaanm:
Looks like this is causing LinkageErrors in user code.  Should we revert, declare 2.11.9 a dud and cut 2.11.10?

# Detailed diagnosis by @xuwei-k:
https://github.com/xuwei-k/Scala-2.11.9-FilteredTraversableInternal

# Explanation by @lrytz (hinted at by @sjrd)
The reason why a reference to `FilteredTraversableInternal` appears in @xuwei-k's example:
```scala
List(1).map(identity).filter(_ => true)
```
In erasure type checking, `List(1).map(identity)` has type `Object` (the result type of `map` is `That` which is a type parameter, therefore erases to `Object`). Selecting the `filter` method requires a cast (`List(1).map(identity).asInstanceOf[FilteredTraversableInternal].filter`). This is done by [erasure's `adaptMember`](https://github.com/scala/scala/blob/v2.11.9/src/compiler/scala/tools/nsc/transform/Erasure.scala#L674-L676), which casts to the filter method symbol's `owner`, which is now `FilteredTraversableInternal`.

szeiger:
@lrytz So we now have a situation that looks binary compatible in bytecode but leads to incompatibilities due to additional information in ScalaSig?

lrytz:
it's not in the scala signature, but bytecode instructions (cast, call
receiver type). i'm not sure it's a binary incompatibility by our defnition
(no linkage error), as you get a ClassNotFound. but the end result is an
incompatibility we don't want to have in a minor release.

lrytz:
oh, or maybe it's a NoClassDefFoundErro, which is a linkage error, i'm on
the phone :)

szeiger:
"in the bytecode instructions" is the effect, not the cause. Binary compatibility only matters when you have separate compilation, in which case the compiler must be getting the idea of casting to `FilteredTraversableInternal` from somewhere in the class file (ScalaSig?). If I understand you correctly the erased return type of `filter` is `Object` but somehow the compiler knows that the return type is really a type parameter that was instantiated as `FilteredTraversableInternal` and inserts a cast to that type. This implies that looking at the erased types is not sufficient for checking binary compatibility in MiMa. OTOH, why isn't there a bridge method for `filter` if it's overridden in `FilteredTraversableInternal`? Shouldn't we see two copies of this method (one returning `Object` and the other one returning `FilteredTraversableInternal` in all subclasses? Or maybe we do but MiMa doesn't recognize it correctly? Or is it missing because they type is private and therefore must not leak (but still does due to a compiler bug)?

(I'm only speculating here. I'd need more time to look into this but I don't expect to find any until after ScalaDays. For now the only safe solution appears to be never to add a new superclass, even if it's declared private and not supposed to leak.)

lrytz:
> If I understand you correctly the erased return type of filter is Object but somehow the compiler knows that the return type is really a type parameter that was instantiated as FilteredTraversableInternal and inserts a cast to that type

My explanation was unclear, sorry: the return type of `map` is its `That` type parameter. When type-checking `List(1).map(identity).filter` during erasure, the qualifier `List(1).map(identity)` gets type `Object`, so the selection `qual.filter` needs a cast: `qual.asInstanceOf[<owner-of-filter>].filter`. This happens here: https://github.com/scala/scala/blob/v2.11.9/src/compiler/scala/tools/nsc/transform/Erasure.scala#L674-L676

> This implies that looking at the erased types is not sufficient for checking binary compatibility in MiMa

I would say that MiMa did its job correctly: it identified the class being added, it was whitelisted. Any reference to a new class is a binary incompatibility.

> Shouldn't we see two copies of this method (one returning Object and the other one returning FilteredTraversableInternal) in all subclasses?

The return type of `FilteredTraversableInternal.filter` in bytecode is `Object`.

szeiger:
Thanks, that makes it clearer. So we'd have to change this to cast to the actual type of the qualifier instead of the method's owner (which can be a supertype) to prevent these leaks?

> I would say that MiMa did its job correctly: it identified the class being added, it was whitelisted. Any reference to a new class is a binary incompatibility.

I was hoping we could change MiMa to ignore new non-accessible superclasses and treat this case as binary compatible but with these owner casts leaking superclasses the only safe situation would be one where all methods implemented by the new superclass are overridden in all subclasses.

lrytz:
> Thanks, that makes it clearer. So we'd have to change this to cast to the actual type of the qualifier instead of the method's owner (which can be a supertype) to prevent these leaks?

This cannot easily be done, unfortunately. At the beginning of erasure, all types of trees are set to `null`, and then the type checker is run again, this time assigning only erased types. So `List(1).map(identity)` really gets type `Object`, I don't think there's a way to get `List` at this point.

We could remember the type that the tree had after typer (`List[Int]`) and use an erased version of that for the cast, but that sounds like a hack. Incidentally, we need to do exactly that in order to avoid `IllegalAccessError`s for selections of Java-defined methods: the method may be defined in a package-protected class that is not accessible. I very recently worked on this: https://github.com/scala/scala/pull/5792. I'm not sure we should use this hack in general, for Scala-only code.

szeiger:
After further consideration 

> the only safe situation would be one where all methods implemented by the new superclass are overridden in all subclasses

even this wouldn't be safe because it's not transitive. Adding a new superclass with overrides of all methods in all subclasses would be considered safe, adding a new override to an existing private class is safe, but the combination of both breaks binary compatibility.

PR #5665, opened 2017-01-29
SI-8662 Repl completions

som-snytt:
Support command completions with a completer that
delegates to a colon command completer, which
either completes the command or completes a filename.

Forgot to check if this small refactor collides with the backlog of PRs.
Some "command" code is moved over to "LoopCommands.scala".

Forgot to add, it also attempts to complete settings by prefix.

Needs tests. Like `:se -Ywarn<tab>`.

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/ILoop.scala
+++ b/src/repl/scala/tools/nsc/interpreter/ILoop.scala
@@ -9,6 +9,7 @@ package interpreter
 import scala.language.{ implicitConversions, existentials }
 import scala.annotation.tailrec
 import Predef.{ println => _, _ }
+import PartialFunction.{cond => when}
 import interpreter.session._
 import StdReplTags._
 import scala.tools.asm.ClassReader
@@ -17,7 +18,7 @@ import scala.tools.nsc.util.{ ClassPath, stringFromStream }
 import scala.reflect.classTag
 import scala.reflect.internal.util.{ BatchSourceFile, ScalaClassLoader, NoPosition }
 import ScalaClassLoader._
-import scala.reflect.io.File
+import scala.reflect.io.{Directory, File, Path}
 import scala.tools.util._
 import io.AbstractFile
 import scala.concurrent.{ ExecutionContext, Await, Future }
@@ -25,6 +26,8 @@ import ExecutionContext.Implicits._
 import java.io.BufferedReader
 
 import scala.util.{ Try, Success, Failure }
+
+import Completion._
 
 /** The Scala interactive shell.  It provides a read-eval-print loop
  *  around the Interpreter class.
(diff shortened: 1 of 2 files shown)


src/repl/scala/tools/nsc/interpreter/LoopCommands.scala
```diff
@@ -24,40 +24,110 @@ class ProcessResult(val line: String) {
   override def toString = "`%s` (%d lines, exit %d)".format(line, buffer.size, exitCode)
 }
 
-trait LoopCommands {
+trait LoopCommands { self: { def echo(msg: String): Unit } =>
```
adriaanm, line 27: Is there a nominal type we could use instead? The current type means `self.echo` is a reflective call.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -214,16 +186,16 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
     cmd("implicits", "[-v]", "show the implicits in scope", intp.implicitsCommand),
     cmd("javap", "<path|class>", "disassemble a file or class name", javapCommand),
     cmd("line", "<id>|<line>", "place line(s) at the end of history", lineCommand),
-    cmd("load", "<path>", "interpret lines in a file", loadCommand),
-    cmd("paste", "[-raw] [path]", "enter paste mode or paste a file", pasteCommand),
+    cmd("load", "<path>", "interpret lines in a file", loadCommand, fileCompletion),
```
adriaanm, line 189: Could the kind of completion be a member of the command?

adriaanm:
Very complete & wholly excellent! I left a few suggestions, but am also happy to already get this in as-is. One more thing: how about encapsulating an algorithm for finding the closest word to a user's input (levinshtein? CmlCs?). We must have a few different implementations accumulating around here by now.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
som-snytt: A `LoopCommand` is only `String => Result` so all the config is externalized here. The list of commands here could be rewritten as a list of `SomeCommand`, or pairs of `String -> SomeCommand`. That is worth doing when REPL is rewritten.

src/repl/scala/tools/nsc/interpreter/LoopCommands.scala
som-snytt: Nominally, yes. I don't remember coding that, but it's probably not a hot path. If computers don't let you be lazy when it doesn't matter, then what are computers good for? I mean besides porn, online bullying and the surveillance state? Somewhere I just read the phrase, "surveillance capitalism" and it all made sense.

adriaanm: :shipit:

PR #5666, opened 2017-01-30
SI-10030 - added quotes in case the path contains a space and an 8.3-…

Philippus:
…compliant path is not available


Philippus:
I signed the CLA

lrytz (approved):

lrytz:
Thank you!

PR #5667, opened 2017-01-31
SI-9729 -Xmaxerrs to limit messages

som-snytt:
Since 2.7.2, console reporter has limited error messages to
a fixed 100. Use `-Xmaxerrs -Xmaxwarns` as from `javac` to configure.


lrytz (approved):

som-snytt:
@lrytz Thanks!

PR #5668, opened 2017-02-01
Finish resolving conflicts from rebasing commits within #5661

dwijnand:

PR #5670, opened 2017-02-01
Fix intellij base file

rorygraves:
Fix the intellij base file (for running ```sbt intellij```)


rorygraves:
Beaten to it by @dwijnand  - https://github.com/scala/scala/pull/5668

PR #5671, opened 2017-02-02
Avoid compiler crash with missing transitive dependencies

retronym:
When we unpickle a reference to `a.b.C` and we don't have a package symbol for `a.b`, we synthesize a stub symbol. This code was modifed in b47aaf6445a to mutate the flags of a stub symbol rather than trying to create a particular subclass of `StubSymbol` when we figure out that a stub represents a package. However, this neglected to set the `MODULE` flag, which can lead to an assertion error later in the compiler.

This PR includes a direct fix for this flag error.

It also includes a few improvements to avoid forcing symbol info transformers in the compiler backend that I identified during the investigation.
I have a few more changes of this nature in [another branch](https://github.com/scala/scala/compare/2.12.x...retronym:topic/stubby-3?expand=1) that I'll revisit.

### DIFF
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -1197,7 +1197,6 @@ trait Types
       else super.safeToString
     override def narrow: Type = this
     override def kind = "ThisType"
-    override def boundSyms = if (sym.isInstanceOf[StubSymbol]) emptySymbolSet else super.boundSyms
   }
 
   final class UniqueThisType(sym: Symbol) extends ThisType(sym) { }
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -1197,6 +1197,7 @@ trait Types
       else super.safeToString
     override def narrow: Type = this
     override def kind = "ThisType"
+    override def boundSyms = if (sym.isInstanceOf[StubSymbol]) emptySymbolSet else super.boundSyms
   }
 
   final class UniqueThisType(sym: Symbol) extends ThisType(sym) { }
--- a/src/reflect/scala/reflect/internal/pickling/UnPickler.scala
+++ b/src/reflect/scala/reflect/internal/pickling/UnPickler.scala
@@ -377,7 +377,7 @@ abstract class UnPickler {
 
       def readThisType(): Type = {
         val sym = readSymbolRef() match {
-          case stub: StubSymbol => stub.setFlag(PACKAGE)
+          case stub: StubSymbol => stub.setFlag(PACKAGE | MODULE)
           case sym => sym
         }
         ThisType(sym)
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -953,10 +953,11 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
        definitions.isDefinitionsInitialized
     && rootMirror.isMirrorInitialized
   )
-  override def isPastTyper = (
+  override def isPastTyper = isPast(currentRun.typerPhase)
+  def isPast(phase: Phase) = (
        (curRun ne null)
     && isGlobalInitialized // defense against init order issues
-    && (globalPhase.id > currentRun.typerPhase.id)
+    && (globalPhase.id > phase.id)
   )
 
   // TODO - trim these to the absolute minimum.
--- a/src/compiler/scala/tools/nsc/backend/ScalaPrimitives.scala
+++ b/src/compiler/scala/tools/nsc/backend/ScalaPrimitives.scala
@@ -446,9 +446,10 @@ abstract class ScalaPrimitives {
       inform(s"Unknown primitive method $cls.$method")
     else alts foreach (s =>
       addPrimitive(s,
-        s.info.paramTypes match {
-          case tp :: _ if code == ADD && tp =:= StringTpe => CONCAT
-          case _                                          => code
+        if (code != ADD) code
+        else exitingTyper(s.info).paramTypes match {
+          case tp :: _ if tp =:= StringTpe => CONCAT
+          case _                           => code
         }
       )
     )
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -809,7 +809,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
     final def isDerivedValueClass =
       isClass && !hasFlag(PACKAGE | TRAIT) &&
-      info.firstParent.typeSymbol == AnyValClass && !isPrimitiveValueClass
+      !phase.erasedTypes && info.firstParent.typeSymbol == AnyValClass && !isPrimitiveValueClass
 
     final def isMethodWithExtension =
       isMethod && owner.isDerivedValueClass && !isParamAccessor && !isConstructor && !hasFlag(SUPERACCESSOR) && !isMacro && !isSpecialized
--- a/test/files/run/t10171/Test.scala
+++ b/test/files/run/t10171/Test.scala
@@ -1,0 +1,59 @@
+import scala.tools.partest._
+import java.io.File
+
+object Test extends StoreReporterDirectTest {
+  def code = ???
+
+  def compileCode(code: String) = {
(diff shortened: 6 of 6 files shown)


retronym:
TODO:
 - [x] Does this fix [SI-10171](https://issues.scala-lang.org/browse/SI-10171) ?

retronym:
@lrytz Could you please take a look at these commits to see if they are on the right track?

The first one avoids tripping an assertion with a stub symbol during symbol substitution that was happening in Uncurry's info transformer, which had dealiased a method parameter from `Predef.String` to `java.lang.String`, but crashed because the method return type was a stub. Special casing `boundSyms` avoids the problem there.

The second one independentally fixes the problem by stopping the propagation of info transformers sooner, by disabling the specialization transform when the compiler has reached the backend. I believe this is safe, even for multi-run compilation (in that case we discard all but the first element of the info history on the next run).

The specialization info transform is the worst offender in forcing method infos. We could potentially disable parts of other info transforms under the same condition, e.g. the [addition of outer accessor methods](https://github.com/scala/scala/blob/d2d33ddf8ce4dff503c8830958bac0e449eb7d34/src/compiler/scala/tools/nsc/transform/ExplicitOuter.scala#L178-L190) in `ExplicitOuter`.

The final commit attempts to lazify part of the conversion from Symbols to BTypes. I haven't validated whether or how much this actually limits the traversal of symbol infos.

A test project is attached to https://issues.scala-lang.org/browse/SI-10171, I'm going to try to extract a `run` test from that.

retronym:
@lrytz I've added a test case and another simple fix (give the stub symbol the `MODULE` flag). However, I think the other commits still have value: avoiding needless work limits our surface area to future bugs and helps performance.

src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
```diff
@@ -1199,7 +1199,7 @@ abstract class SpecializeTypes extends InfoTransform with TypingTransformers {
    *  If it is a 'no-specialization' run, it is applied only to loaded symbols.
    */
   override def transformInfo(sym: Symbol, tpe: Type): Type = {
-    if (settings.nospecialization && currentRun.compiles(sym)) tpe
```
retronym: The backend needs to specialize `TupleN` to populate its list of tuple constructors to make tests for box-elimination to work. This seems to be the only case covered by our test suite that requires specialization of a type after the specialization phase itself has run.

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -1194,6 +1194,7 @@ trait Types
       else super.safeToString
     override def narrow: Type = this
     override def kind = "ThisType"
+    override def boundSyms = if (sym.isInstanceOf[StubSymbol]) emptySymbolSet else super.boundSyms
```
lrytz: Looks OK, but looking around for a while, it seems that allowing to access `tpe` on stub symbols is a feature. So I prefer the fix in your next commit (`MODULE` flag for stub package symbols), and we could live without the special-case here.

```
scala> import scala.reflect.internal.Flags._
scala> val c = definitions.ScalaPackageClass.newStubSymbol(newTypeName("C"), "not here")
scala> c.tpe
res7: $r.intp.global.Type = C

scala> val c = definitions.ScalaPackageClass.newStubSymbol(newTypeName("C"), "not here")
scala> c.setFlag(PACKAGE)
scala> c.tpe
java.lang.IllegalArgumentException: requirement failed: package C
  at scala.reflect.internal.Types$ModuleTypeRef.<init>(Types.scala:1879)

scala> val c = definitions.ScalaPackageClass.newStubSymbol(newTypeName("C"), "not here")
scala> c.setFlag(PACKAGE | MODULE)
scala> c.tpe
res12: $r.intp.global.Type = C.type
```

While looking at `boundSyms`, do you know why [`ExistentialType.boundSyms`](https://github.com/scala/scala/blob/v2.12.1/src/reflect/scala/reflect/internal/Types.scala#L2619) doesn't `++ underlying.boundSyms`?

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -377,7 +377,7 @@ abstract class UnPickler {
 
       def readThisType(): Type = {
         val sym = readSymbolRef() match {
-          case stub: StubSymbol => stub.setFlag(PACKAGE)
+          case stub: StubSymbol => stub.setFlag(PACKAGE | MODULE)
```
lrytz, line 380: This looks good.

src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
```diff
@@ -1199,22 +1204,34 @@ abstract class SpecializeTypes extends InfoTransform with TypingTransformers {
...
-        val parents1 = parents mapConserve specializedType
-        if (parents ne parents1) {
-          debuglog("specialization transforms %s%s parents to %s".format(
-            if (tparams.nonEmpty) "(poly) " else "", clazz, parents1)
-          )
+        if (isPast(ownPhase)) {
+          assert(!currentRun.compiles(sym), sym)
+          // Skip specialization info transform for third party classes that aren't referenced directly
+          // from the tree or by the specialization info transform itself that are run up to the end of
+          // the specialization phase.
```
lrytz: this comment is a bit hard to read

src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala
```diff
@@ -198,6 +198,11 @@ abstract class SpecializeTypes extends InfoTransform with TypingTransformers {
   override def newPhase(prev: scala.tools.nsc.Phase): StdPhase = new SpecializationPhase(prev)
   class SpecializationPhase(prev: scala.tools.nsc.Phase) extends super.Phase(prev) {
     override def checkable = false
+    override def apply(unit: CompilationUnit): Unit = {
+      super.apply(unit)
+      FunctionClass.seq.map(_.info)
+      TupleClass.seq.map(_.info)
```
lrytz: there should be a comment here

lrytz (commented):
This looks good! Also the commit introducing lazy completion, though I only took a quick look.

src/reflect/scala/reflect/internal/Types.scala
retronym: > While looking at `boundSyms`, do you know why `ExistentialType.boundSyms` doesn't `++ underlying.boundSyms`?

Looks like it is called repeatedly as the [substituition `TypeMap`](https://github.com/scala/scala/blob/1eb331b4196492f8f8084a2c102272c3a9c0cd2d/src/reflect/scala/reflect/internal/tpe/TypeMaps.scala#L696-L759) recurses, so its doing the right thing by being shallow. 

Come to think of it, I suspect we could actually just return null (in `SimpleTypeProxy.boundSyms`) and have the same results as we do today.

retronym:
@lrytz I've trimmed this down to the essential commit and some low-risk passengers.

lrytz (approved):

src/reflect/scala/reflect/internal/Types.scala
lrytz: Would you like to keep this override, given the other more fundamental fix? I'd remove it, but either way is fine.

dwijnand:
super green

src/reflect/scala/reflect/internal/Types.scala
lrytz: just a quick ping here, then we can merge.

adriaanm: Does this mean the review is currently still pending? Lets make that clear with "request changes" review. I was about to merge this PR because the review status was "approved".

lrytz: ok, yes, i just wanted @retronym to consider this once more.

lrytz (changes requested):
()

src/reflect/scala/reflect/internal/Types.scala
retronym: I've removed the override for now. I think we could actually generalize it to:

```scala
class SubType {
  override def boundSyms = emptySymbolSet
}
```

Along with overrides to pick that:

```
abstract case class ThisType {
  override def boundSyms = super[SubType].boundSyms
}
...
```

But let's rely on `underlying.boundSyms` for now.

lrytz (approved):

PR #5672, opened 2017-02-02
Improve performance of immutable Map.contains

rorygraves:
Map.contains defaults to ``` get(key).isDefined ``` generating an extra Option wrapper on each use (which is immediately discarded).    Override the base implementation to reduce memory churn.

Performance benchmark.
```
before 
[info] Benchmark              (size)  Mode  Cnt      Score      Error  Units
[info] MapBenchmark.contains       1  avgt   20      2.610 ±    0.138  ns/op
[info] MapBenchmark.contains      10  avgt   20     30.473 ±    1.233  ns/op
[info] MapBenchmark.contains     100  avgt   20   1549.097 ±   65.146  ns/op
[info] MapBenchmark.contains    1000  avgt   20  21638.244 ± 1021.994  ns/op

after
[info] Benchmark              (size)  Mode  Cnt      Score     Error  Units
[info] MapBenchmark.contains       1  avgt   20      2.682 ±   0.116  ns/op
[info] MapBenchmark.contains      10  avgt   20     31.703 ±   1.563  ns/op
[info] MapBenchmark.contains     100  avgt   20   1068.583 ±  53.095  ns/op
[info] MapBenchmark.contains    1000  avgt   20  14310.988 ± 758.181  ns/op
```

### DIFF
--- a/src/library/scala/collection/immutable/HashMap.scala
+++ b/src/library/scala/collection/immutable/HashMap.scala
@@ -49,7 +49,10 @@ sealed class HashMap[A, +B] extends AbstractMap[A, B]
 
   override def foreach[U](f: ((A, B)) => U): Unit = ()
 
-  def get(key: A): Option[B] =
+  override def contains(key: A): Boolean =
+    contains0(key, computeHash(key), 0)
+
+  override def get(key: A): Option[B] =
     get0(key, computeHash(key), 0)
 
   override def updated [B1 >: B] (key: A, value: B1): HashMap[A, B1] =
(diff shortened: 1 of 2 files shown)


rorygraves:
Great minds think alike @retronym? #5673

retronym:
> Great minds think alike @retronym?

Either that, or simple minds seldom differ... 😉 

We'll need the whitelist changes from my PR, so if you don't mind, I'll close this PR and we can continue the review and compatibilty discussions over there.

rorygraves:
Lol. Sounds good to me.

PR #5600, opened 2016-12-13
Further small HashTable optimizations

l0rinc:
These are rather refactorings, as I cannot measure any real difference with the existing `JMH` benchmarks.

PR #5673, opened 2017-02-02
Optimizations in immutable.Map.{get, contains}

retronym:
- Avoid allocation of Some in get
  - defer integer left shift until needed
  - avoid redundantly masking an integer

before:

    Benchmark                  (size)  Mode  Cnt      Score      Error  Units
    HashMapBenchmark.contains      10  avgt   20    284.624 ±   18.985  ns/op
    HashMapBenchmark.contains     100  avgt   20   3190.580 ±   33.622  ns/op
    HashMapBenchmark.contains    1000  avgt   20  52967.171 ± 1524.834  ns/op
    HashMapBenchmark.get           10  avgt   20    248.168 ±    2.612  ns/op
    HashMapBenchmark.get          100  avgt   20   2795.469 ±   54.458  ns/op
    HashMapBenchmark.get         1000  avgt   20  52238.773 ± 1268.764  ns/op

after:

    Benchmark                  (size)  Mode  Cnt      Score     Error  Units
    HashMapBenchmark.contains      10  avgt   20    195.107 ±   2.442  ns/op
    HashMapBenchmark.contains     100  avgt   20   2454.151 ±  24.392  ns/op
    HashMapBenchmark.contains    1000  avgt   20  40722.993 ± 520.473  ns/op
    HashMapBenchmark.get           10  avgt   20    245.282 ±   3.547  ns/op
    HashMapBenchmark.get          100  avgt   20   2729.669 ±  32.767  ns/op
    HashMapBenchmark.get         1000  avgt   20  49568.410 ± 794.565  ns/op

### DIFF
--- a/src/library/scala/collection/immutable/HashMap.scala
+++ b/src/library/scala/collection/immutable/HashMap.scala
@@ -52,6 +52,9 @@ sealed class HashMap[A, +B] extends AbstractMap[A, B]
   def get(key: A): Option[B] =
     get0(key, computeHash(key), 0)
 
+  override final def contains(key: A): Boolean =
+    contains0(key, computeHash(key), 0)
+
   override def updated [B1 >: B] (key: A, value: B1): HashMap[A, B1] =
     updated0(key, computeHash(key), 0, value, null, null)
 
@@ -92,7 +95,7 @@ sealed class HashMap[A, +B] extends AbstractMap[A, B]
   import HashMap.{Merger, MergeFunction, liftMerger}
 
   private[collection] def get0(key: A, hash: Int, level: Int): Option[B] = None
-
+  protected def contains0(key: A, hash: Int, level: Int): Boolean = false
   private[collection] def updated0[B1 >: B](key: A, hash: Int, level: Int, value: B1, kv: (A, B1), merger: Merger[A, B1]): HashMap[A, B1] =
     new HashMap.HashMap1(key, hash, value, kv)
 
@@ -185,6 +188,7 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
     }
   }
 
+  @deprecatedInheritance("This class will be made final in a future release.", "2.12.2")
   class HashMap1[A,+B](private[collection] val key: A, private[collection] val hash: Int, private[collection] val value: (B @uV), private[collection] var kv: (A,B @uV)) extends HashMap[A,B] {
     override def size = 1
 
@@ -195,6 +199,8 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
     override def get0(key: A, hash: Int, level: Int): Option[B] =
       if (hash == this.hash && key == this.key) Some(value) else None
 
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean =
+      hash == this.hash && key == this.key
     private[collection] override def updated0[B1 >: B](key: A, hash: Int, level: Int, value: B1, kv: (A, B1), merger: Merger[A, B1]): HashMap[A, B1] =
       if (hash == this.hash && key == this.key ) {
         if (merger eq null) {
@@ -238,6 +244,9 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
 
     override def get0(key: A, hash: Int, level: Int): Option[B] =
       if (hash == this.hash) kvs.get(key) else None
+
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean =
+      hash == this.hash && kvs.contains(key)
 
     private[collection] override def updated0[B1 >: B](key: A, hash: Int, level: Int, value: B1, kv: (A, B1), merger: Merger[A, B1]): HashMap[A, B1] =
       if (hash == this.hash) {
@@ -294,6 +303,7 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
     }
   }
 
+  @deprecatedInheritance("This class will be made final in a future release.", "2.12.2")
   class HashTrieMap[A, +B](
     private[collection] val bitmap: Int,
     private[collection] val elems: Array[HashMap[A, B @uV]],
@@ -306,21 +316,41 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
     override def size = size0
 
     override def get0(key: A, hash: Int, level: Int): Option[B] = {
+      // Note: this code is duplicated with `contains0`
       val index = (hash >>> level) & 0x1f
-      val mask = (1 << index)
       if (bitmap == - 1) {
-        elems(index & 0x1f).get0(key, hash, level + 5)
-      } else if ((bitmap & mask) != 0) {
-        val offset = Integer.bitCount(bitmap & (mask-1))
-        elems(offset).get0(key, hash, level + 5)
-      } else
-        None
+        elems(index).get0(key, hash, level + 5)
+      } else {
+        val mask = (1 << index)
+        if ((bitmap & mask) != 0) {
+          val offset = Integer.bitCount(bitmap & (mask - 1))
+          elems(offset).get0(key, hash, level + 5)
+        } else {
+          None
+        }
+      }
+    }
+
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean = {
+      // Note: this code is duplicated from `get0`
+      val index = (hash >>> level) & 0x1f
+      if (bitmap == - 1) {
+        elems(index).contains0(key, hash, level + 5)
+      } else {
+        val mask = (1 << index)
+        if ((bitmap & mask) != 0) {
+          val offset = Integer.bitCount(bitmap & (mask - 1))
+          elems(offset).contains0(key, hash, level + 5)
+        } else {
+          false
+        }
+      }
     }
 
     private[collection] override def updated0[B1 >: B](key: A, hash: Int, level: Int, value: B1, kv: (A, B1), merger: Merger[A, B1]): HashMap[A, B1] = {
       val index = (hash >>> level) & 0x1f
       val mask = (1 << index)
-      val offset = Integer.bitCount(bitmap & (mask-1))
+      val offset = Integer.bitCount(bitmap & (mask - 1))
       if ((bitmap & mask) != 0) {
         val sub = elems(offset)
         val subNew = sub.updated0(key, hash, level + 5, value, kv, merger)
@@ -342,7 +372,7 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
     override def removed0(key: A, hash: Int, level: Int): HashMap[A, B] = {
       val index = (hash >>> level) & 0x1f
       val mask = (1 << index)
-      val offset = Integer.bitCount(bitmap & (mask-1))
+      val offset = Integer.bitCount(bitmap & (mask - 1))
       if ((bitmap & mask) != 0) {
         val sub = elems(offset)
         val subNew = sub.removed0(key, hash, level + 5)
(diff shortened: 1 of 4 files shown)


l0rinc:
Thanks, will review it tomorrow :)

Ichoran (approved):
Looks good. I prefer to write comments about duplicated code as `// Note: this code is duplicated in bar` and `// Note: this code is duplicated from foo`, but it's clear enough what's going on as written.

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -195,6 +199,8 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
     override def get0(key: A, hash: Int, level: Int): Option[B] =
       if (hash == this.hash && key == this.key) Some(value) else None
 
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean =
+      hash == this.hash && key == this.key
```
l0rinc, line 203: minor: on line [249](https://github.com/scala/scala/pull/5673/files#diff-ef59ae1920818cddd42a380fc636af4fR249) you're using parens around binary operators

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -317,6 +328,20 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
         None
     }
 
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean = {
+      // Note: this code is duplicated with `get0`
+      val index = (hash >>> level) & 0x1f
+      val mask = (1 << index)
+      if (bitmap == - 1) {
+        elems(index & 0x1f).contains0(key, hash, level + 5)
```
l0rinc: it seems to me `index` was already masked, can probably be replaced by:
```scala
elems(index)
```
in most instances in this class

test/benchmarks/src/main/scala/scala/collection/immutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,49 @@
...
+      case 1 => i.toChar
+      case 2 => i.toDouble
+      case 3 => i.toInt
+    }).toArray
+    missingKeys = (size to 2 * size).toArray
+  }
+
+  var map: collection.immutable.HashMap[Any, Any] = null
+
+  @Setup(Level.Invocation) def initializeMutable = {
```
l0rinc: you don't need invocation level for immutables, please rename and use
```scala
@Setup(Level.Trial) def initialize = {
```

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -317,6 +328,20 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
         None
     }
 
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean = {
+      // Note: this code is duplicated with `get0`
+      val index = (hash >>> level) & 0x1f
+      val mask = (1 << index)
+      if (bitmap == - 1) {
```
l0rinc: minor: formatting

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -317,6 +328,20 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
...
+        elems(index & 0x1f).contains0(key, hash, level + 5)
+      } else if ((bitmap & mask) != 0) {
+        val offset = Integer.bitCount(bitmap & (mask-1))
```
l0rinc: minor: formatting

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -317,6 +328,20 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
         None
     }
 
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean = {
+      // Note: this code is duplicated with `get0`
+      val index = (hash >>> level) & 0x1f
+      val mask = (1 << index)
```
l0rinc: minor: `mask` is not used in the first branch, consider moving it to the `else` branch

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -317,6 +328,20 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
...
+      if (bitmap == - 1) {
+        elems(index & 0x1f).contains0(key, hash, level + 5)
+      } else if ((bitmap & mask) != 0) {
+        val offset = Integer.bitCount(bitmap & (mask-1))
+        elems(offset).contains0(key, hash, level + 5)
+      } else {
+        false
```
l0rinc: this line was the reason for the PR, right?

test/benchmarks/src/main/scala/scala/collection/immutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,49 @@
...
+    map = collection.immutable.HashMap(existingKeys.map(x => (x, x)) : _*)
```
l0rinc: I think `contains` needs to test collisions also, maybe we should add some redundancy to the values

src/library/scala/collection/immutable/HashMap.scala
l0rinc: `+ 5`, as in `log2 0x1f`, right?
Trying to understand the algorithm...

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -317,6 +328,20 @@ object HashMap extends ImmutableMapFactory[HashMap] with BitOperations.Int {
...
 
+    override protected def contains0(key: A, hash: Int, level: Int): Boolean = {
+      // Note: this code is duplicated with `get0`
+      val index = (hash >>> level) & 0x1f
+      val mask = (1 << index)
+      if (bitmap == - 1) {
+        elems(index & 0x1f).contains0(key, hash, level + 5)
+      } else if ((bitmap & mask) != 0) {
+        val offset = Integer.bitCount(bitmap & (mask-1))
+        elems(offset).contains0(key, hash, level + 5)
```
l0rinc: Could this possibly cause a stack overflow in case of very heavy collision? Or is it bound by the height of the tree (log32 size)?

src/library/scala/collection/immutable/HashMap.scala
```diff
@@ -52,6 +52,9 @@ sealed class HashMap[A, +B] extends AbstractMap[A, B]
   def get(key: A): Option[B] =
     get0(key, computeHash(key), 0)
 
+  override final def contains(key: A): Boolean =
```
l0rinc, line 55: It looks like `HashSet` is not affected by this change. Do you think it would be worth applying it there too?

src/library/scala/collection/immutable/HashMap.scala
l0rinc: Is it even worth verifying the hash in this case?

l0rinc (changes requested):

src/library/scala/collection/immutable/HashMap.scala
l0rinc: If I understood it correctly, this traverses down the tree, calculating the 32 bit digits of the hash.
In Vector this path is unrolled to avoid all those method calls. While I think it's a very excessive optimization, it might be worth trying it here also.

src/library/scala/collection/immutable/HashMap.scala
Ichoran: In this code, it was certainly masked and the extra masking does nothing, and the JIT compiler will almost surely elide it, but why make it do the extra work?  Good catch.

src/library/scala/collection/immutable/HashMap.scala
Ichoran: It's bound by the height of the tree.  The unrolling will make the code *very* ugly, and might even be a net loss by expanding the bytecode a lot (which can make the JIT compiler try less hard for a good optimization).  If you've still got your own branch around it might be worth a try, but I'd really try to avoid doing this.  Manually making it tail-recursive might be worth a try too.  (Loop as long as isInstanceOf gives the right answer.)

src/library/scala/collection/immutable/HashMap.scala
Ichoran: No.  The equivalent line is `None`, which also is zero-allocation.  It's the success branch that is the reason.

src/library/scala/collection/immutable/HashMap.scala
jvican: I think `mask` cannot be removed given the fact that it's used twice in the else branch, first in the predicate and then in the body.

src/library/scala/collection/immutable/HashMap.scala
l0rinc: ~~Thanks, I thought it's the additional `Option` wrapping(s).~~ misread your post :)

src/library/scala/collection/immutable/HashMap.scala
l0rinc: I meant something like:
```scala
val index = (hash >>> level) & 0x1f
if (bitmap == -1) {
  elems(index).contains0(key, hash, level + 5)
} else {
  val mask = (1 << index)
  if ((bitmap & mask) != 0) {
    val offset = Integer.bitCount(bitmap & (mask - 1))
    elems(offset).contains0(key, hash, level + 5)
  } else {
    false
  }
}
```

jvican: Right, in that case it would work. Thanks for elaboration.

src/library/scala/collection/immutable/HashMap.scala
l0rinc: @Ichoran, I fully agree, even if it would be faster, it made `Vector` very obfuscated, in the end you guys will have to decide whether it's worth the trouble.
@retronym, would you like me to try this out, or would you like to? I'm not very familiar with the code, it would probably take me some time...

retronym:
@paplorinc I've incorporated your suggestions and re-run the benchmark.

test/benchmarks/src/main/scala/scala/collection/immutable/HashMapBenchmark.scala
```diff
@@ -0,0 +1,58 @@
...
+      case 1 => i.toChar
+      case 2 => i.toDouble
+      case 3 => i.toInt
+    }).toArray
+    missingKeys = (size to 2 * size).toArray
+  }
+
+  var map: collection.immutable.HashMap[Any, Any] = null
+
+  @Setup(Level.Trial) def initializeMutable = {
```
l0rinc: sorry for nitpicking, but this isn't mutable :)

l0rinc (approved):

l0rinc:
Related to: https://github.com/scala/scala/pull/5600

rorygraves:
@retronym @paplorinc 
Looking at the test failure - its unrelated
```
[info] ! Division of Duration by Long.on overflow edge cases: Exception raised on property evaluation.
[info] > Exception: org.scalacheck.Gen$Choose$IllegalBoundsError: invalid bounds: low=9223372036854775799, high=-9223372036854775797
```
I did some investigation and the failure occurs here:
https://github.com/scala/scala/blob/2.12.x/test/scalacheck/duration.scala#L32
A value of 1 or -1 fails the choose on line 35.
I don't have time to raise the proper bug report right now - but thought I would highlight it.

retronym:
Noted the unrelated test failure as https://github.com/scala/scala-dev/issues/296

adriaanm:
Looking over the bytecode diff between 2.12.1 and 2.12.2 (see https://github.com/scala/scala-dev/issues/331), this added a (byte code) public method to a public class. Just noting for the record that this method should not be called from user (Java) code (it's not visible to Scala code outside of the package).

```diff
--- a/library/scala/collection/immutable/HashMap.class
+++ b/library/scala/collection/immutable/HashMap.class
@@ -251,6 +251,34 @@ public class scala/collection/immutable/HashMap extends scala/collection/immutab
<snip>
+  // access flags 0x1
+  // signature (TA;II)Z
+  // declaration: boolean contains0(A, int, int)
+  public contains0(Ljava/lang/Object;II)Z
```

PR #5674, opened 2017-02-04
SI-9519 removed the usecase section of the ++-method

Philippus:
JIRA: https://issues.scala-lang.org/browse/SI-9519


adriaanm:
Thanks, @Philippus!

PR #5675, opened 2017-02-04
Add tests for ConsoleReporter.

piyush-jaiswal:
This commit adds tests for the ConsoleReporter which includes tests for the feature added by [#5667](https://github.com/scala/scala/pull/5667).

### DIFF
--- a/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
@@ -77,10 +77,10 @@ class ConsoleReporter(val settings: Settings, reader: BufferedReader, writer: Pr
     if (reader != null) {
       reader.read match {
         case 'a' | 'A' =>
-          new Throwable().printStackTrace()
+          new Throwable().printStackTrace(writer)
           System.exit(1)
         case 's' | 'S' =>
-          new Throwable().printStackTrace()
+          new Throwable().printStackTrace(writer)
           writer.println()
           writer.flush()
         case _ =>
--- a/test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
+++ b/test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
@@ -1,0 +1,173 @@
+package scala
+package tools.nsc
+package reporters
+
+import java.io.{ByteArrayOutputStream, StringReader, BufferedReader, PrintStream, PrintWriter}
+import org.junit.Assert._
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+import scala.reflect.internal.util._
+
+
+@RunWith(classOf[JUnit4])
+class ConsoleReporterTest {
+  val source = "Test_ConsoleReporter"
+  val batchFile = new BatchSourceFile(source, "For testing".toList)
+  val posWithSource = new OffsetPosition(batchFile, 4)
+  val content = posWithSource.lineContent
+  val writerOut = new ByteArrayOutputStream()
+  val echoWriterOut = new ByteArrayOutputStream()
+
+
+  def createConsoleReporter(inputForReader: String, errOut: ByteArrayOutputStream, echoOut: ByteArrayOutputStream = null): ConsoleReporter = {
+    val reader = new BufferedReader(new StringReader(inputForReader))
+      
+    /** Create reporter with the same writer and echoWriter if echoOut is null */
+    echoOut match {
+      case null => new ConsoleReporter(new Settings(), reader, new PrintWriter(errOut))
+      case _ => new ConsoleReporter(new Settings(), reader, new PrintWriter(errOut), new PrintWriter(echoWriterOut))
+    }
+  }
+
+
+  def testHelper(pos: Position = NoPosition, msg: String, severity: String = "")(test: Position => Unit) = {
+    test(pos)
+    if (msg.isEmpty && severity.isEmpty) assertTrue(writerOut.toString.isEmpty)
+    else {
+      if (!pos.isDefined) assertEquals(severity + msg, writerOut.toString.lines.next)
+      else {
+        val it = writerOut.toString.lines
+        assertEquals(source + ":1: " + severity + msg, it.next)
+        assertEquals(content, it.next)
+        assertEquals("    ^", it.next)
+      }
+    }
+    writerOut.reset
+  }
+
+  
+  @Test
+  def printMessageTest(): Unit = {
+    val reporter = createConsoleReporter("r", writerOut)
+    testHelper(msg = "Hello World!")(_ => reporter.printMessage("Hello World!"))
+    testHelper(msg = "Testing with NoPosition")(reporter.printMessage(_, "Testing with NoPosition"))
+    testHelper(posWithSource, "Testing with Defined Position")(reporter.printMessage(_, "Testing with Defined Position"))
+  }
+
+
+  @Test
+  def echoTest(): Unit = {
+    val reporter = createConsoleReporter("r", writerOut, echoWriterOut)
+    reporter.echo("Hello World!")
+    assertEquals("Hello World!", echoWriterOut.toString.lines.next)
+
+    /** Check with constructor which has the same writer and echoWriter */
+    val reporter2 = createConsoleReporter("r", writerOut)
+    testHelper(msg = "Hello World!")(_ => reporter2.echo("Hello World!"))
+  }
+
+
+  @Test
+  def printTest(): Unit = {
+    val reporter = createConsoleReporter("r", writerOut)
+    testHelper(msg = "test")(reporter.print(_, "test", reporter.INFO))
+    testHelper(msg = "test", severity = "warning: ")(reporter.print(_, "test", reporter.WARNING))
+    testHelper(msg = "test", severity = "error: ")(reporter.print(_, "test", reporter.ERROR))
+    testHelper(posWithSource, msg = "test")(reporter.print(_, "test", reporter.INFO))
+    testHelper(posWithSource, msg = "test", severity = "warning: ")(reporter.print(_, "test", reporter.WARNING))
+    testHelper(posWithSource, msg = "test", severity = "error: ")(reporter.print(_, "test", reporter.ERROR))
+  }
+
+
+  @Test
+  def printColumnMarkerTest(): Unit = {
+    val reporter = createConsoleReporter("r", writerOut)
+    testHelper(msg = "")(reporter.printColumnMarker(_))
+
+    reporter.printColumnMarker(posWithSource)
+    assertEquals("    ^", writerOut.toString.lines.next)
+    writerOut.reset
+  }
+
+
+  @Test
+  def displayTest(): Unit = {
+    val reporter = createConsoleReporter("r", writerOut)
+
+    /** Change maxerrs and maxwarns from default */
+    reporter.settings.maxerrs.value = 1
+    reporter.settings.maxwarns.value = 1
+
+    testHelper(msg = "Testing display")(reporter.display(_, "Testing display", reporter.INFO))
+    testHelper(msg = "Testing display", severity = "warning: ")(reporter.display(_, "Testing display", reporter.WARNING))
+    testHelper(msg = "Testing display", severity = "error: ")(reporter.display(_, "Testing display", reporter.ERROR))
+    testHelper(posWithSource, msg = "Testing display")(reporter.display(_, "Testing display", reporter.INFO))
+    testHelper(posWithSource, msg = "Testing display", severity = "warning: ")(reporter.display(_, "Testing display", reporter.WARNING))
+    testHelper(posWithSource, msg = "Testing display", severity = "error: ")(reporter.display(_, "Testing display", reporter.ERROR))
+
+    reporter.resetCount(reporter.ERROR)
+    reporter.resetCount(reporter.WARNING)
+
+    reporter.ERROR.count += 1
+    testHelper(posWithSource, msg = "Testing display for maxerrs to pass", severity = "error: ")(reporter.display(_, "Testing display for maxerrs to pass", reporter.ERROR))
+    reporter.ERROR.count += 1
+    testHelper(msg = "")(reporter.display(_, "Testing display for maxerrs to fail", reporter.ERROR))
+
+    reporter.WARNING.count += 1
+    testHelper(posWithSource, msg = "Testing display for maxwarns to pass", severity = "warning: ")(reporter.display(_, "Testing display for maxwarns to pass", reporter.WARNING))
+    reporter.WARNING.count += 1
+    testHelper(msg = "")(reporter.display(_, "Testing display for maxwarns to fail", reporter.WARNING))
+  }
+
+
+  @Test
+  def finishTest(): Unit = {
+    val reporter = createConsoleReporter("r", writerOut)
+
+    reporter.resetCount(reporter.ERROR)
+    reporter.resetCount(reporter.WARNING)
+    testHelper(msg = "")(_ => reporter.finish())
+
+    reporter.ERROR.count = 10
+    reporter.WARNING.count = 3
+    reporter.finish()
+    val it = writerOut.toString.lines
+    assertEquals("three warnings found", it.next)
+    assertEquals("10 errors found", it.next)
+    writerOut.reset
+  }
+
+
+  @Test
+  def displayPromptTest(): Unit = {
+    val output = "a)bort, s)tack, r)esume: "
+
+    /** Check for stack trace */
+    val reporter = createConsoleReporter("s", writerOut, echoWriterOut)
+    reporter.displayPrompt()
+    val it = writerOut.toString.lines
+    assertTrue(it.next.isEmpty)
+    assertEquals(output + "java.lang.Throwable", it.next)
+    assertTrue(it.hasNext)
+    
+    /** Check for no stack trace */
+    val writerOut2 = new ByteArrayOutputStream()
+    val reporter2 = createConsoleReporter("w", writerOut2)
+    reporter2.displayPrompt()
+    val it2 = writerOut2.toString.lines
+    assertTrue(it2.next.isEmpty)
+    assertEquals(output, it2.next)
+    assertFalse(it2.hasNext)
+
+    /** Check for no stack trace */
+    val writerOut3 = new ByteArrayOutputStream()
+    val reporter3 = createConsoleReporter("r", writerOut3)
+    reporter3.displayPrompt()
+    val it3 = writerOut3.toString.lines
+    assertTrue(it3.next.isEmpty)
+    assertEquals(output, it3.next)
+    assertFalse(it3.hasNext)
+  }
+}


piyush-jaiswal:
Added 2 files, am not certain about the coding style.

som-snytt:
I think it's the git style. You didn't intend to commit all the extra files.

piyush-jaiswal:
Yeah, exactly. Anyways, just to clarify, I have signed the CLA.

piyush-jaiswal:
@adriaanm, could you review it?

som-snytt:
Since `ConsoleReporter` takes a reader and writer, the unit test should supply them and not use `Console`.

(Using `Console` in particular reduces test isolation, a problem if you want to run many tests without forking.)

This test could also subsume the new test at https://github.com/scala/scala/pull/5667 where probably an end-to-end test is overkill.

Note that maxerr isn't fixed at 100 now, so setting it explicitly for the test is useful (instead of testing the default).

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      reporter.ERROR.count = 101
+      reporter display(posWithFile, "Testing display2", reporter.ERROR)
+      assertEquals("", err_out.toString)
+      err_out.reset
+
+
+
+      // finish test
+      reporter.ERROR.count = 0
+      reporter finish;    // don't know why but I'm requiring ';' here, otherwise the file does not compile
```
som-snytt: I'd suggest not enabling postfixOps at all.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,171 @@
...
+    val batch_file = new BatchSourceFile(file)
+    val posWithFile = new OffsetPosition(batch_file, 4)
+    val content = posWithFile.lineContent
+    val err_out = new ByteArrayOutputStream   //for storing the output printed in the console
+
+    Console.withErr(err_out) {
+      val reporter = new ConsoleReporter(new Settings())
+
+      // printMessage test
+      reporter printMessage("Hello World!")
```
som-snytt: I'd suggest using dot-selection, especially with parens.
```
reporter.printMessage("hello, world.")
reporter printMessage "hello, world."    // Recently, I use this style less, but community hasn't outlawed it.
```

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,171 @@
...
+
+
+      // finish test
+      reporter.ERROR.count = 0
+      reporter finish;    // don't know why but I'm requiring ';' here, otherwise the file does not compile
+      assertEquals("", err_out.toString)
+      err_out.reset
+
+      reporter.WARNING.count = 3
+      reporter.ERROR.count = 10
```
som-snytt: Probably when the reporter API is nailed down some more, this should be disallowed. Maybe Reporter should have `protected increment(severity)`.

piyush-jaiswal:
Pardon me @som-snytt , I didn't know you were my assigned reviewer. I will make the changes as suggested but in the meanwhile, my midterms are around the corner. Is it okay if this can wait for some days?

som-snytt:
Absolutely, reviewing and updating are on a volunteer basis!

piyush-jaiswal:
Hey @som-snytt, I made the changes. Could you review and suggest further changes if needed.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,185 @@
...
+import scala.reflect.internal.util._
+import scala.reflect.io.AbstractFile
+
+
+@RunWith(classOf[JUnit4])
+class ConsoleReporterTest {
+  @Test
+  def unitTest: Unit = {
+    val noPos = NoPosition
+    val path = "test/junit/scala/tools/nsc/reporters/file_ConsoleReporterTest.txt"
```
som-snytt: I'm not clear on why a file is needed. Source can be just a byte array.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,185 @@
...
+    //Changing maxerrs and maxwanrs
+    setting.maxerrs.tryToSet(List("50"))
+    setting.maxwarns.tryToSet(List("25"))
+    val reporter = new ConsoleReporter(setting, input_reader, output_writer)
+
+
+    // printMessage test
+    reporter.printMessage("Hello World!")
+    assertEquals("Hello World!\n", err_out.toString)
+    err_out.reset
```
som-snytt: Could refactor this pattern to a method.

piyush-jaiswal: I'm not sure that I get you correctly. Do you want me to group the related tests inside a method, for example the tests for printMessage inside a single method?

piyush-jaiswal:
Made the changes. Still not sure, if it's exactly what you wanted.

piyush-jaiswal:
Hey @som-snytt,  could you review the changes?

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,188 @@
...
+    val batch_file = new BatchSourceFile(source, "For testing".toList)
+    val posWithSource = new OffsetPosition(batch_file, 4)
+    val content = posWithSource.lineContent
+
+    val err_out = new ByteArrayOutputStream  //for storing the output printed in the ConsoleReporter
+    val output_writer = new PrintWriter(err_out)
+    val input = new StringReader("s\n")
+    val input_reader = new BufferedReader(input)
+    val setting = new Settings()
+    setting.maxerrs.tryToSet(List("50"))  //changing maxerrs
```
som-snytt: `settings.maxerrs.value = 50`

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,188 @@
...
+    setting.maxwarns.tryToSet(List("25")) //chaging maxwarns
+    val reporter = new ConsoleReporter(setting, input_reader, output_writer)
+
+    @Test
+    def printMessageTest(): Unit = {
+      reporter.printMessage("Hello World!")
+      assertEquals("Hello World!\n", err_out.toString)
+      err_out.reset
```
som-snytt: You could capture this repeated pattern in a routine that performs the assert and the reset. I often see simple helpers like that in unit tests.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,188 @@
...
+    @Test
+    def printColumnMarkerTest(): Unit = {
+      // printColumnMarkerTest test with NoPosition
+      reporter.printColumnMarker(noPos)
+      assertEquals("", err_out.toString)
+      err_out.reset
+
+      // printColumnMarkerTest test with DefinedPosition
+      reporter.printColumnMarker(posWithSource)
+      assertEquals("    ^\n", err_out.toString)
```
som-snytt: This assumes the platform line ending, perhaps. `text.lines` converts to lines.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,188 @@
...
+      reporter.ERROR.count = 10
+      reporter.finish
+      assertEquals("three warnings found\n10 errors found\n", err_out.toString)
+      err_out.reset
+    }
+
+    @Test
+    def displayPromptTest(): Unit = {
+      val stack = new ByteArrayOutputStream
+      System.setErr(new PrintStream(stack))  // used for checking the stack trace of displayPrompt()
```
som-snytt: `Console.withErr`. Again, it's easier to encapsulate with a method that does the setup. Allocating a BAOS for every test is easier than `stack.reset`. As a style matter, maybe the idea is that it's worth reusing a heavyweight object?

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,188 @@
...
+      
+      reporter.displayPrompt;
+      assertFalse(stack.toString.equals(""))
+      stack.reset
+
+      {
+        val input = new StringReader("r\n")
+        val input_reader = new BufferedReader(input)
+        val reporter = new ConsoleReporter(new Settings(), input_reader, output_writer)
+        reporter.displayPrompt
```
som-snytt: Not a huge deal, but I'd expect parens to demonstrate a side-effecting application. That's the standard advice. This looks like it just accesses a value.

som-snytt (commented):
I made a couple of small comments. I'm trying out the "review" button for the first time, which really complicates the workflow. (note to github)

You need to rebase and clean up the git history, which includes spurious merge commits. They want to see your one squashed commit before they merge. (as a rule)

som-snytt:
You have to fix the git history. Your PR should show your one commit.

piyush-jaiswal:
I didn't push form the '2.12.x' branch after rebasing it, think that was the root of the problem. Thanks @som-snytt !

I added the protected increment and reset severity functions in ConsoleReporter and also modified the tests. Hope you can take a look at it.

adriaanm:
@piyush-jaiswal, note that you seem to have accidentally committed a bunch of unrelated files. Please also review our [guidelines](https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#clean-commits-clean-history) on how to format a good commit message. As it stands, your commit title is too long, is not in the present tense, and there's no elaboration in the commit message itself (for example: what does this do? why? what was wrong before?)

Thanks for reviewing @som-snytt! When you feel this PR is ready, feel free to merge.

piyush-jaiswal:
Made the changes. Kindly point out other issues if present.

piyush-jaiswal:
Hey @som-snytt , how are you? I've made some changes since we last communicated. In particular I've added the increment and reset severity functions in the ConsoleReporter as you had earlier suggested.

 I still couldn't figure out how to use `Console.withErr` instead of `System.setErr` for the `displayPromptTest` as I am checking for the stack trace which I think is printed in the standard error stream System.err. Could you go ahead and have a look?

som-snytt:
Sorry, just a busy time, albeit not midterms. If I suggested adding API to ConsoleReporter just for a test, don't listen to that guy. Also, there was just a commit to let it take two writers, for error and echo. That might affect your test.

You can actually fix something by changing it to use `printStackTrace(writer)`. It becomes testable and correct. I guess you can't test the abort option.

piyush-jaiswal:
Okay, I've changed my tests according to the commit and yeah I don't think it's possible to test the abort option too.

piyush-jaiswal:
review @som-snytt ?

piyush-jaiswal:
Is there any further issue present? I am finding it difficult to spotting them on my own.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,211 @@
...
+
+  def createConsoleReporter(inputForReader: String, writerOut: ByteArrayOutputStream, echoWriterOut: ByteArrayOutputStream = null): ConsoleReporter = {
+      val writer = new PrintWriter(writerOut)
+      val reader = new BufferedReader(new StringReader(inputForReader))
+      
+      /** create reporter with the same writer and echoWriter */
+      if(echoWriterOut == null)
+        new ConsoleReporter(new Settings(), reader, writer) 
+      else
+      {
```
som-snytt: Repo style is space after if `if (` and open brace on same line `else {`.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,211 @@
...
+  val echoWriterOut = new ByteArrayOutputStream()
+  val reporter = createConsoleReporter("s", writerOut, echoWriterOut)
+
+  /** Change maxerrs and maxwarns */
+  reporter.settings.maxerrs.value = 50
+  reporter.settings.maxwarns.value = 25
+
+
+  def testHelper(checkBlank: Boolean, hasPos: Boolean, message: String, severity: String): Unit = {
+    if(checkBlank) assertEquals("", writerOut.toString)
```
som-snytt: the flag is redundant with `message = ""`. Possibly `message.isEmpty && severity.isEmpty`.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,211 @@
...
+      if(!hasPos) assertEquals(severity + message, writerOut.toString.lines.next)
+      else {
+        val it = writerOut.toString.lines
+        assertEquals(source + ":1: " + severity + message, it.next)
+        assertEquals(content, it.next)
+        assertEquals("    ^", it.next)
+      }
+    }
+    writerOut.reset
+  }
```
som-snytt, line 48: The test rig could be: `testHelper(pos: Position = NoPosition, msg: String, severity: String = "")(test: Position => Unit)`.

`testHelper(msg = "hello, world")(_ => reporter.printMessage("hello, world"))`.

`testHelper(myPos, "test")(reporter.print(_, "test", INFO))` or similar.

The `hasPos` flag is `pos.isDefined`.

I lose track of the `true, false` pairs in the tests.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,211 @@
...
+
+
+  @Test
+  def printTest(): Unit = {
+    /** print test with NoPosition */
+    reporter.print(noPos, "test", reporter.INFO)
+    testHelper(false, false, "test", "")
+
+    reporter.print(noPos, "test", reporter.WARNING)
+    testHelper(false, false, "test", "warning: ")
```
som-snytt: The `false, false`, which I've forgotten what it means, is noise.

Sometimes a little table of truth values in code is tidy, but not as args. It can help to name them, `f(hasPos = false, isBlank = false, x, y)`, and in non-test code the repo requires named boolean args for this reason.

Eliminating the noise would take this to another level.

src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
```diff
@@ -91,4 +91,4 @@ class ConsoleReporter(val settings: Settings, reader: BufferedReader, writer: Pr
   override def flush() = writer.flush()
 
   override def finish() = printSummary()
-}
+}
```
som-snytt: Probably restore terminal NL.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,211 @@
...
+    reporter.display(posWithSource, "Testing display for maxerrs to fail", reporter.ERROR)
+    testHelper(true, false, "", "")
+
+    reporter.WARNING.count = 25
+    reporter.display(posWithSource, "Testing display for maxwarns to pass", reporter.WARNING)
+    testHelper(false, true, "Testing display for maxwarns to pass", "warning: ")
+
+    reporter.WARNING.count = 26
+    reporter.display(posWithSource, "Testing display for maxwarns to fail", reporter.WARNING)
+    testHelper(true, false, "", "")
```
som-snytt: The one thing tested by `test/files/neg/maxerrs.scala` is that output is suppressed but the error or warning over the max to display is still counted, and not simply ignored. Here, the count is bumped.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
piyush-jaiswal: I'm not sure if we require `Position` as an argument of the function stored in the second parameter, `test` of `testHelper`.

We could just do:
`def testHelper(pos: Position = NoPosition, msg: String, severity: String = "")(test: () => Unit)`

som-snytt: I think I was trying to avoid repeating the pos info. My second example reads well; you'd only need arrow syntax to ignore the pos in the test function.

But yes, you could just:

`testHelper(myPos, "test")(reporter.print(myPos, "test", INFO))`

with by-name param instead of a function:

`def testHelper(...)(test: => Unit)`

Either way, I think it's worth the trouble, because they want to do more tricks with reporters, like filtering, so it's likely your test will get exercised and expanded.

piyush-jaiswal:
I've tried to make the changes. Please review. Thanks!

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  def createConsoleReporter(inputForReader: String, writerOut: ByteArrayOutputStream, echoWriterOut: ByteArrayOutputStream = null): ConsoleReporter = {
+      val writer = new PrintWriter(writerOut)
+      val reader = new BufferedReader(new StringReader(inputForReader))
+      
+      /** Create reporter with the same writer and echoWriter if None */
+      Option(echoWriterOut) match {
+        case None => new ConsoleReporter(new Settings(), reader, writer)
+        case Some(_) => new ConsoleReporter(new Settings(), reader, writer, new PrintWriter(echoWriterOut))
+      }
```
som-snytt: This is an idiom you'll see nowhere. I won't nitpick default to null, but you can just `case null =>` or just `val echo = if (out != null) new PrintWriter(out) else writer`. You might as well name the params `errOut` and `echoOut` since they aren't Writers.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    testHelper(msg = "Hello World!")(_ => reporter2.echo("Hello World!"))
+  }
+
+
+  @Test
+  def printTest(): Unit = {
+    testHelper(msg = "test")(_ => reporter.print(NoPosition, "test", reporter.INFO))
+    testHelper(msg = "test", severity = "warning: ")(_ => reporter.print(NoPosition, "test", reporter.WARNING))
+    testHelper(msg = "test", severity = "error: ")(_ => reporter.print(NoPosition, "test", reporter.ERROR))
+    testHelper(posWithSource, msg = "test")(_ => reporter.print(posWithSource, "test", reporter.INFO))
```
som-snytt: Since your helper takes `Pos => Unit`, this would be
```
testHelper(posWithSource, msg = "test")(reporter.print(_, "test", reporter.INFO))
```
If you find that finicky, then helper could just take `(body: => Unit)`, and maybe that's easier to read, too:
```
testHelper(posWithSource, msg = "test")(reporter.print(posWithSource, "test", reporter.INFO))
```

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    testHelper(msg = "Testing display", severity = "warning: ")(_ => reporter.display(NoPosition, "Testing display", reporter.WARNING))
+    testHelper(msg = "Testing display", severity = "error: ")(_ => reporter.display(NoPosition, "Testing display", reporter.ERROR))
+    testHelper(posWithSource, msg = "Testing display")(_ => reporter.display(posWithSource, "Testing display", reporter.INFO))
+    testHelper(posWithSource, msg = "Testing display", severity = "warning: ")(_ => reporter.display(posWithSource, "Testing display", reporter.WARNING))
+    testHelper(posWithSource, msg = "Testing display", severity = "error: ")(_ => reporter.display(posWithSource, "Testing display", reporter.ERROR))
+
+    reporter.ERROR.count += 1
+    testHelper(posWithSource, msg = "Testing display for maxerrs to pass", severity = "error: ")(_ => reporter.display(posWithSource, "Testing display for maxerrs to pass", reporter.ERROR))
+    reporter.ERROR.count += 1
+    testHelper(msg = "")(_ => reporter.display(posWithSource, "Testing display for maxerrs to fail", reporter.ERROR))
```
som-snytt: Since you're testing `display` (and not `reporter.error`), I'd set `maxerrs = 1` and `ERROR.count = 0` explicitly in the test so we can see the conditions of the test. I don't know what the current count is, does it depend on order tests are run? etc.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    testHelper(msg = "")(_ => reporter.finish())
+
+    reporter.ERROR.count = 10
+    reporter.WARNING.count = 3
+    reporter.finish()
+    val it = writerOut.toString.lines
+    assertEquals("three warnings found", it.next)
+    assertEquals("10 errors found", it.next)
+    writerOut.reset
+  }
```
som-snytt, line 140: The tests assume test isolation. Probably I wouldn't have used a shared reporter, since it's cheap to construct. Each test could use a throw-away reporter. Here, for instance, the counts are bumped up and other tests could see that, depending on how the test is run (one Test instance per test method or one instance for all, test order, etc).

som-snytt:
Just a couple of small comments to push it across. It's rude at this point to ask not to use a shared reporter across test methods, so I'll leave that to you, so long as the tests actually pass. For future reference, I avoid sharing the "system under test" unless it's expensive to construct.

adriaanm:
Good work, @piyush-jaiswal -- thanks. Thank you for the in-depth reviewing, @som-snytt!

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
piyush-jaiswal: I think I read somewhere that `null` was provided in scala for java interop and that `Option` is better to use in scala. As it may be, I think I'll use `null =>` as it utilizes one of the constructors in ConsoleReporter.

som-snytt:
I'm pretty sure this PR hit all of the hardest problems in computer science, or I might be off by one.

This test will be so useful when someone adds reporter features. :+1:

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
piyush-jaiswal: Should I use a separate `BAOS` for each test as well and pass it as an argument in `testHelper` or let it remain a global as it is now?

piyush-jaiswal:
Updated. I've used the same `BAOS` for constructing all the separate reporters ATM.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
```diff
@@ -0,0 +1,173 @@
...
+    else {
+      if (!pos.isDefined) assertEquals(severity + msg, writerOut.toString.lines.next)
+      else {
+        val it = writerOut.toString.lines
+        assertEquals(source + ":1: " + severity + msg, it.next)
+        assertEquals(content, it.next)
+        assertEquals("    ^", it.next)
+      }
+    }
+    writerOut.reset
```
som-snytt, line 47: Technically, we should have put the reset in a finally block. If it fails, it throws, so the next test to run could see the dirty buffer. Probably the test class is reinstantiated; I think junit does that.

som-snytt:
LGTM

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
som-snytt: To avoid both `null` and boxing, method overload: `f(out)` calls `f(out, out)`. Sometimes the default arg can be an empty value of the desired type.

piyush-jaiswal:
Thanks @adriaanm and thanks @som-snytt  for taking the time to review this PR. This definitely was a great learning experience for me!

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
dave-shawley: Good catch, fixed in 496feac3. This was a leftover from renaming the session to `<service>-<namespace>` — the fallback still tested the bare namespace, so a concurrent `just up <service>` winning the race would fail the recipe while the session it was waiting for existed.

test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala
dave-shawley: The analysis is right, and one worker is the intended design — `WEB_CONCURRENCY=1` is already set in both the `Dockerfile` and `okteto.yaml`. What was missing is any explanation of why it must stay there, which is the real risk: someone scales the service later and quietly breaks metrics.

769c9ba6 documents the constraint at each place it can be changed. The canonical explanation lives in the `fastapi_runner.metrics` docstring, where the limitation actually is; the `Dockerfile` and `okteto.yaml` carry short notes against the setting itself.

We are not adding cross-worker aggregation. This service scales with replicas, not workers — Prometheus scrapes each pod and aggregates across them, which is the standard pattern here and avoids `PROMETHEUS_MULTIPROC_DIR` and its shared-directory lifecycle entirely.

PR #5676, opened 2017-02-06
SI-10173 Crash output shows true line number

som-snytt:
Where true means 1-based, for whatever tree was last to typer.


SethTisue:
would adding a test be worth the effort...?

som-snytt:
It would be worth it to someone who knew how to implement a test that wouldn't fail as soon as either the crasher was fixed or the line numbers in the crash report changed arbitrarily.

Dark themed github actually makes me feel more dark!

lrytz:
I remember this problem :) https://twitter.com/lrytz/status/569105699550187520

lrytz (approved):

PR #5677, opened 2017-02-07
Prepare for 2.13 library modularization [ci: last-only]

adriaanm:
In 2.13, the following packages will move out of scala-library
(to their own module):
  - scala.concurrent
  - scala.util (Try and Either move to the scala package)
  - scala.io (see also better-files)
  - scala.ref
  - scala.sys (not recommended for use)
  - scala.compat (entirely deprecated)
  - scala.text (entirely deprecated)

This commit cuts some ties to prepare for this move.

To compensate for the associated deprecations in your own code:
  - `compat.Platform.EOL` --> add `import java.lang.System.{lineSeparator => EOL}`
  - `compat.Platform.currentTime` --> add `import java.lang.System.{currentTimeMillis => currentTime}`
  - `compat.Platform.arraycopy` --> add `import java.lang.System.arraycopy`
  - `sys.error` --> `throw new RuntimeException` (or subclass)
  - `sys.exit` --> `System.exit`
  - `sys.props` --> `System.getProperties().asScala` (with `import scala.collection.JavaConverters._`)
  - `sys.env` --> `System.getenv().asScala`
  - `sys.runtime` --> `Runtime.getRuntime`
  - `sys.addShutdownHook` --> `Runtime.getRuntime.addShutdownHook(new Thread(() => ...))`
  - `Console.BOLD` --> `AnsiColor.BOLD` (with `import scala.io.AnsiColor`)
  - `Console.readLine` --> `Console.in.readLine()`
  - `concurrent.Lock` --> `java.util.concurrent.locks.ReentrantLock` (lock/unlock)

# TODO
  - [ ] verify last commit preserves sys.process semantics.
  - [ ] deprecations in test
  - [ ] osgi
  - [ ] bincompat  `method this(java.lang.String)Unit in class scala.sys.ShutdownHookThread`
  - [ ] run/lazy-concurrent.scala                 [output differs]
  - [ ] run/t2318.scala `java.security.AccessControlException: access denied ("java.util.PropertyPermission" "scala.control.noTraceSuppression" "read")`


adriaanm:
Deprecations
  - run/constrained-types.scala 
  - run/noInlineUnknownIndy 
  - run/sd275-java 
  - run/sd275.scala 
  - run/shutdownhooks.scala 
  - run/t2464 
  - run/t4671.scala 
  - run/t5463.scala 
  - run/t5717.scala 
  - run/t5940.scala 
  - run/t6240-universe-code-gen.scala 
  - run/t6440.scala 
  - run/t6440b.scala 
  - run/t6502.scala 
  - run/t7439 
  - run/t7455 
  - run/t7582-private-within 
  - run/t7741a 
  - run/t7741b 
  - run/t8502.scala 
  - run/t8502b.scala 
  - run/t8549.scala 
  - run/t8907.scala 
  - run/t9268 
  - run/typetags_without_scala_reflect_manifest_lookup.scala 
  - run/typetags_without_scala_reflect_typetag_lookup.scala 
  - run/typetags_without_scala_reflect_typetag_manifest_interop.scala 
  - run/various-flat-classpath-types.scala   
  - jvm/innerClassEnclMethodJavaReflection.scala 
  - jvm/interpreter.scala

dwijnand:
For clarity, "to their own modules" means: moved out of the repo, switch groupId to org.scala-lang.modules and new versions, right? Like XML and parser-combinators? Or is it more conservative than that?

adriaanm:
Yes, for most of these, I'd want them to evolve faster than the core of the
library, or, if that's not possible, they should make way for community
modules of higher quality (such as better-files).

AlecZorab:
Do you expect the same treatment to be applied to the collections library, either for 2.13 or for some hypothetical later release?

adriaanm:
No. The collections are core to the language. We will keep reducing the
coupling between the spec/compiler and the collections, but they will
always be part of the core. Doing otherwise would fragment the language
because then you would have two implementations of, say, List and Iterable.

AlecZorab:
My recollection (which is likely wrong, I looked at this a long time ago)
was that the only "collection" types referred to specifically in the spec
itself were `Option` (for `unapply`) and `Seq` (`unapplySeq`, possibly
others? ) . Since the addition of name based pattern matching, it seems
like references to `Option` could be dropped, which would leave just `Seq`.

My extremely rough searching (on a phone through a pdf of the 2.9 spec)
found no references to `Iterable` nor `List` in the spec outside of
examples.
It would be a shame to forgo the potential  increase in modularity, given
that the collections library is being reworked. Indeed, if `Seq` lost its
place of privilege in the spec then it would be possible to offer both the
legacy collections library and the reworked one as individual, separate
modules which would no doubt substantially ease the lives of people
migrating to 2.13

adriaanm:
I agree that the coupling should be reduced, as I tried to acknowledge in my message.

However, the more important point is that, in my opinion, it is crucial for Scala adoption that there be only one Scala collections API. Library interop crucially relies on agreeing on many of the core collection types, even if the compiler and spec could be reworked not to rely on our specific implementations (using a name-based approach as with for comprehensions). 

I'm happy to see extensions built on top of the collections, but wholesale replacement is not feasible for a widely used library (again, in my opinion).

SethTisue:
can we discuss this elsewhere, e.g. on http://contributors.scala-lang.org? it isn't directly relevant to this pull request, which is more limited in scope.

AlecZorab:
I'd suggest that exactly that did happen for actors, but I'll certainly concede that multiple collection libraries might result in some level of inconvenience for people needing to go between libraries depending on different libs. 

As Seth suggests, I'll reserve further comment for another venue

adriaanm:
review by adriaanm

adriaanm:
review @adriaanm

adriaanm:
(Testing out https://github.com/scala/scabot/pull/68, but something seems off -- it did work at https://github.com/scala/scala/pull/5662!)

SethTisue:
review by @SethTisue

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -27,7 +27,7 @@ trait ScalaSettings extends AbsScalaSettings
    *  defaults to the value of CLASSPATH env var if it is set, as in Java,
    *  or else to `"."` for the current user directory.
    */
-  protected def defaultClasspath = sys.env.getOrElse("CLASSPATH", ".")
+  protected def defaultClasspath = Option(System.getenv("CLASSPATH")).getOrElse(".")
```
gourlaysama, line 30: There is `scala.util.Properties.{envOrElse, envOrNone}` for that :)

src/compiler/scala/reflect/macros/compiler/Errors.scala
```diff
@@ -1,7 +1,7 @@
 package scala.reflect.macros
 package compiler
 
-import scala.compat.Platform.EOL
+import java.lang.System.{lineSeparator => EOL}
```
gourlaysama, line 4: There already was `Properties.lineSeparator` too, though it doesn't really add anything at all.

src/compiler/scala/tools/nsc/transform/patmat/Logic.scala
```diff
@@ -309,10 +309,8 @@ trait Logic extends Debugging  {
 
     }
 
-    // TODO: remove since deprecated
-    val budgetProp = scala.sys.Prop[String]("scalac.patmat.analysisBudget")
-    if (budgetProp.isSet) {
-      reportWarning(s"Please remove -D${budgetProp.key}, it is ignored.")
+    if (System.getProperty("scalac.patmat.analysisBudget") != null) {
```
gourlaysama, line 312: That's where all the `Properties.{propOrXXX, propIsSet, ...}` methods are useful. I hadn't even realized it was all duplicated in `sys.props` until I saw it being used in the REPL.

sjrd:
@gourlaysama From the PR description, IIUC, the idea is also not to rely on `scala.util.Properties`.

gourlaysama:
Ah, never mind, I hadn't seen that `scala.util` was going away too.

I would argue for moving `Properties` into the `scala` package then. Some of the things in there are useful to have all in one place in the core, like the stuff read from `library.properties`.

sjrd:
The stuff read from `library.properties` is not portable. It doesn't work on Scala.js. Moving something non-portable to the top `scala._` package would be a serious mistake, IMO.

Besides, even though system properties overall do work in Scala.js, the only reason it is supported is because of JUnit. They shouldn't otherwise be used, IMO.

adriaanm:
> Moving something non-portable to the top scala._ package would be a serious mistake, IMO.

Agreed. This is very much the goal here: the core should be easily portable across platforms. (Of course, as in scala.js and scala-native, parts of the Java stdlib will have to be reimplemented on those platforms.)

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
adriaanm: As Sébastien said, this is one of the ties I'd like to cut.

src/compiler/scala/reflect/macros/compiler/Errors.scala
adriaanm: As Sébastien said, this is one of the ties I'd like to cut.

src/compiler/scala/tools/nsc/transform/patmat/Logic.scala
adriaanm: The scala.sys.Prop is a bit over-designed for my tastes.

nafg:
Why is scala.sys being deprecated? I think it makes scripting with scala a lot nicer.

For instance just now I entered `eval sys.env` into the sbt prompt, to help me figure out which database I'm currently running against. Conveniences are important. I don't remember seeing a discussion about it.

SethTisue:
> Why is scala.sys being deprecated

Eh? It isn't.

I've responded at some length over at https://contributors.scala-lang.org/t/modularization-of-scala-2-13-standard-library/635

lrytz:
> scala.util (Try and Either move to the scala package)

What about these:
  - `scala.util.matching.Regex`
  - `scala.util.hashing` (`MurmurHash3` has various uses, also in the compiler)
  - `scala.util.control`
  - `scala.util.Random`
  - `scala.util.Sorting`

jvican:
> scala.util.matching.Regex
> scala.util.Random
> scala.util.Sorting
> scala.util.matching.Regex
  
I think it makes sense to remove these.
  
> scala.util.hashing (MurmurHash3 has various uses, also in the compiler)
  
Honestly, I've used this so much, and I've seen so many people relying on it, that I'm a little bit reluctant on moving this away. I think it should stay.

lrytz:
One thing I never understood was the history of `scala.util.Properties` vs `scala.sys.props`, why do we have both / was the older not deprecated? Is `scala.util.Properties` going to stay? [edit: i see now that `scala.util.Properties` does more by reading the `library.properties` file]

About `scala.util.matching.Regex`, I don't think we can remove it, because it's used in collections (`StringLike.r`). The same is true for `scala.util.control.Breaks`. `ControlThrowable`, `NonFatal` are used in many places.

I think we really need to keep those core things in the core library.

Are we going to have an "util" module? Sounds a bit like a dumping ground :) But some util things are quite useful, and we want to provide the "batteries included" experience (with easy to get modules).

`scala.concurrent` is used in the repl. If we move it to a module we'll introduce a new bootstrap cycle. Maybe the repl moving out as well? Maybe we want to use futures in the compiler implementation at some point? I guess we'll have to use jdk apis. Also `scala.io` has a few uses in the compiler / repl, these should move to java.nio I guess.

jvican:
Keeping `control` and `hashing` makes sense since they have value on their own.
  
However, I'm not sure if we should leave these in the core even if Scalac depends on them:
  
> scala.util.matching.Regex
> scala.util.Random
> scala.util.Sorting
  
This functionality seems auxiliary. It's not clear if we should keep them because:
  
* These utils are easily portable to the Scala side. The burden to maintain these simple wrappers around Java libs is close to zero (they haven't been touched in a while).
* The contents of core should have a clear purpose. These seem to be just facilities but are not built around a common idea.
  
That being said, I think that a high-level discussion on the contents of the core would be beneficial. It's obvious to me that collections are necessary, but utilities that are not as visible as collections should be picked with care and a clear intention.
  
Should we raise this discussion in the next SIP meeting? We can probably identify common ground between the two compilers and find the most urgent needs that could benefit all Scala developers. I don't know if this makes sense, so I think it would be useful to bring it up to debate. /cc @odersky @adriaanm @SethTisue  @sjrd @xeno-by @heathermiller @dragos @jsuereth.

src/library/scala/sys/package.scala
```diff
@@ -45,6 +45,7 @@ package object sys {
    *
    *  @return   the result of `java.lang.Runtime.getRuntime()`
    */
+  @deprecated("use Runtime.getRuntime", "2.12.2")
```
lrytz, line 48: as discussed (half?) offline, i think this pr should not deprecate things that will move out to separate modules. it would be better to define those modules first, carve them out, and then evolve the codebase there, hopefully with new contributors and energy :)

I won't copy this comment to other deprecations introduced in this PR, they are easy to find.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
lrytz: I understand the goal of this PR is to cut dependencies, so this comment is out of scope: I'm in favor of shipping this kind of convenience (`envOrElse`) with Scala (in a form that makes it easily available, I understand we want to provide an opt-out).

lrytz (commented):
The changes for cutting ties look good to me, let me know what you think about my other remarks.

lrytz:
now in #5830

SethTisue:
@nafg pointed out [on the contributors forum](https://contributors.scala-lang.org/t/modularization-of-scala-2-13-standard-library/635/4?u=sethtisue) that if `Try` moves up to `scala.Try`, `Success` and `Failure` would move with them, which seems questionable

nafg:
A lot of ADTs follow the convention that the subtypes are put inside the
companion, so technically one option is to do Try.Success / Try.Failure...

floating-cat:
Current `Option`, `Try`, `Either`, `List`'s subtypes package are same to its parent, and `LazyList`'s subtypes are put inside the companion (There must be some other ADTs in standard lib I don't know). 

Should we apply the some rules to these ADTs (put their subtypes inside the companion), or just treat some special?

src/compiler/scala/tools/nsc/transform/patmat/Logic.scala
som-snytt: I would just re-read the source whenever I needed to use it.

som-snytt:
I'm not sure the fate of regex was fixed?

It would nice if support were in a form like xml, where the std lib doesn't depend on an impl, but usages of `"p".r` do. Then whatever impl on the classpath is used. I don't know off the cuff what that would look like.

I hope somebody actually takes a big pair of scissors and cuts Adriaan's tie when this is all complete.

SethTisue:
we've run out of time in the 2.13 cycle — let's pick this up again for 2.14

PR #5678, opened 2017-02-07
Bump Swing version for 2.0.0 release

benhutchison:


SethTisue:
I noticed that https://github.com/scala/scala-swing/releases wasn't listing 2.0.0 as latest — now fixed. Can you add some release notes there...? So that once this ships, if anyone wants to find out what's new, they can. (Even if what's new is actually nothing.)

I wasn't expecting that page to say “SethTisue released this” as a result and didn't mean to take credit, sorry :-\

adriaanm:
(Let's squash when merging this PR)

PR #5680, opened 2017-02-08
update and improve description of deprecation policy

SethTisue:
the old text said that if we (e.g.) deprecated something in 2.12.x, we
couldn't remove it until 2.14.0. that was never the intent, it isn't
what we have said elsewhere, and I don't think the community ever
understood that to be the policy, either.


SethTisue:
inspired by #5505, which was abandoned

adriaanm (approved):

PR #5681, opened 2017-02-08
SI-9704 don't add a closing HtmlTag if it is already closed

Philippus:


SethTisue:
review by @VladUreche ?

src/scaladoc/scala/tools/nsc/doc/base/comment/Body.scala
```diff
@@ -86,7 +86,9 @@ final case class HtmlTag(data: String) extends Inline {
   }
 
   private val TagsNotToClose = Set("br", "img")
-  def close = tagName collect { case name if !TagsNotToClose(name) => HtmlTag(s"</$name>") }
+  def close = tagName collect {
+    case name if !TagsNotToClose(name) && !data.endsWith(s"</$name>") => HtmlTag(s"</$name>")
+  }
```
lrytz: The change looks correct - the usage of `HtmlTag` a bit messy in general... Since you are in the area, would you mind pulling out `Pattern` and `TagsNotToClose` into the companion of `HtmlTag`? There's no reason to have them as fields in every instance.

Philippus: Done!

lrytz (approved):

lrytz:
Thanks a lot!

lrytz:
@Philippus here's the diff for the scala library before and after this change: https://github.com/lrytz/scaladoc-diff/commits/master

Would you mind checking if that all looks as expected?

Philippus:
LGTM
I did notice another problem with HtmlTag ending up in the title-tag, visible for instance on http://www.scala-lang.org/files/archive/api/current/scala/beans/BeanInfoSkip.html if you hover in the right menu over "BeanInfoSkip", but that's a separate issue.

PR #5683, opened 2017-02-08
Remove deprecated stuff from Predef, scala._, scala.runtime

SethTisue:
- bye Pair, Triple
- bye conforms
- bye Predef.readLine and friends
- bye DeprecatedConsole
- bye NotNull
- bye RichException
- ~~bye SeqCharSequence~~
- bye DeprecatedPredef
- bye Responder
- ~~bye `round` on integer types~~
- bye ScalaRuntime.hash
- bye runtime.StringFormat

and:
- `final`ed some @deprecatedOverriding things
- removed some @deprecatedNames

some notable things that were *not* touched in this round:
- Manifest/ClassManifest (sigh)
- DelayedInit (sigh)
- StringAdd (tangled w/ reflection, needs re-STARR)


src/library/scala/runtime/RichInt.scala
```diff
@@ -37,10 +37,6 @@ final class RichInt(val self: Int) extends AnyVal with ScalaNumberProxy[Int] wit
   override def min(that: Int): Int = math.min(self, that)
   override def signum: Int         = math.signum(self)
 
-  /** There is no reason to round an `Int`, but this method is provided to avoid accidental loss of precision from a detour through `Float`. */
-  @deprecated("this is an integer type; there is no reason to round it.  Perhaps you meant to call this on a floating-point value?", "2.11.0")
-  def round: Int = self
```
sjrd: The comment here suggests that this is a deprecation meant to stay forever. IIUC, without it the following silently loses precision:
```scala
val x: Int = ???
x.round
```
because it turns `x` into a `Float` to be able to call `RichFloat.round`, or something like that.

We have similar intended deprecated-forever methods in Scala.js: https://github.com/scala-js/scala-js/blob/v0.6.14/library/src/main/scala/scala/scalajs/js/JSNumberOps.scala#L90-L104

src/library/scala/runtime/RichLong.scala
```diff
@@ -33,10 +33,6 @@ final class RichLong(val self: Long) extends AnyVal with IntegralProxy[Long] {
   override def min(that: Long): Long = math.min(self, that)
   override def signum: Int           = math.signum(self).toInt
 
-  /** There is no reason to round a `Long`, but this method is provided to avoid accidental conversion to `Int` through `Float`. */
-  @deprecated("this is an integer type; there is no reason to round it.  Perhaps you meant to call this on a floating-point value?", "2.11.0")
-  def round: Long = self
```
sjrd: Same remark here. I think this is meant to stay.

sjrd (changes requested):

src/library/scala/Console.scala
```diff
@@ -122,7 +122,7 @@ import scala.util.DynamicVariable
  *             a body of code. Threadsafe by virtue of [[scala.util.DynamicVariable]].
  *
  */
-object Console extends DeprecatedConsole with AnsiColor {
+object Console extends AnsiColor {
```
adriaanm, line 125: Ideally, we'd also not inherit AnsiColor, as it's in `scala.io`, which shouldn't be a requirement of the core.

SethTisue: Yeah. But I figure it's #5677's province.

dwijnand: The scaladoc specifically calls out to getting your 22 bizarro ansi escape string literals by importing through `Console`, so I guess this has to go through a deprecation cycle?

>  import Console.{GREEN, RED, RESET, YELLOW_B, UNDERLINED}

SethTisue:
thanks @sjrd, good catch. I've pushed a revised commit that preserves those `round` methods, and also fixes some failed partests.

sjrd (approved):

src/library/scala/Console.scala
sjrd: I would suggest copying those `final val` into `Console` to cut ties with `scala.io`.

SethTisue:
@adriaanm shall we merge?

SethTisue:
will be interesting to see if anything breaks in the community build...

SethTisue:
> will be interesting to see if anything breaks in the community build.

very little broke — a few lines here and there in a few projects

PR #5684, opened 2017-02-09
run partest from sbt always, command line never

SethTisue:
these scripts are (I assume) unused these days and there's no reason
to maintain them. and it's risky to have two different ways of running
the same thing which could get out of sync with each other

sbt forks a JVM to run partest, so it's not like we need these scripts
in order to get a more isolated environment in that respect.


SethTisue:
@retronym has a knack for thinking up sadly-all-too-sensible reasons for not deleting things

dragos:
Not @retronym but I'm using the `partest` script all the time. The startup time is much better than Sbt and often I have just one test that I want to run (with neat TAB-completion without the annoying extra space that Sbt keeps adding). Minor annoyance but here's my vote to keep them

SethTisue:
any other opinions on this? I hear you Iulian, but I personally don't think it outweighs the issues I have noted. (sbt should be left running, and if the tab completion isn't right, better to actually fix it)

retronym:
I don't use `partest-ack` anymore, much of the functionality is rolled into `sbt "partest --grep ..."` now. I also don't use the command line runner often, and would be fine with this change.

If I'm working on a individual test and need fine grained control over, e.g., opening a debug port, I tend to:

```
$ scalac test/files/run/t1234.scala && scala -J-$DEBUG_OPTS Test
```

retronym:
The annoying space in SBT tab completion goes away with:

```diff
⚡ git diff head
diff --git a/project/PartestUtil.scala b/project/PartestUtil.scala
index ab7e62b3b4..23570a88ec 100644
--- a/project/PartestUtil.scala
+++ b/project/PartestUtil.scala
@@ -90,6 +90,6 @@ object PartestUtil {
     val ScalacOptsParser = (token("-Dpartest.scalac_opts=") ~ token(NotSpace)) map { case opt ~ v => opt + v }

     val P = oneOf(knownUnaryOptions.map(x => token(x))) | SrcPath | TestPathParser | Grep | ScalacOptsParser
-    (Space ~> repsep(P, oneOrMore(Space))).map(_.mkString(" ")).?.map(_.getOrElse("")) <~ OptSpace
+    (Space ~> repsep(P, oneOrMore(Space))).map(_.mkString(" ")).?.map(_.getOrElse(""))
   }
 }
```

SethTisue:
thx @retronym, I've added a commit with your fix. (I tested it and it seems to work.)

SethTisue:
Adriaan OK'ed merging

adriaanm:
sbt startup time is zero once you've started it once -- I always keep an sbt shell open and run the scalac/partest tasks from there

PR #5686, opened 2017-02-09
Correct the AbstractVar error message to be more correct

propensive:
The error currently reads "only classes can have declared but undefined
members", which isn't true on two counts: traits can have them, and
concrete classes cannot. This corrects the error message to read "only
traits and abstract classes can have declared but undefined members".


propensive:
Let me know if this needs a corresponding SI- issue opened...

SethTisue:
in general, for minor things, if there's a pull request, we don't need a JIRA issue too

propensive:
My only thought was that someone (ScalaClippy, IntelliJ, maybe?) might be depending on the exact text of the error message for something... Though I guess a JIRA issue probably wouldn't help with that.

SethTisue (approved):

SethTisue:
thank you @propensive!

PR #5687, opened 2017-02-09
SI-10177 Override lazy operations to preserve TrieMap's semantics

retronym:
Calling .iterator on a TrieMap gives a read-only snapshot. This then
extends to most inherited implementations written in terms of .iterator.

However, some inherited methods, such as .values, either defer the call
to .iterator or call it more than once. This results in subsequent
mutations to the original map being visible

I reviewed the inherited implementations from MapLike and found we
needed overrides of `values`, `keySet`, `filterKeys`, and `mapValues`.
Like `iterator`, these now create a read-only snapshot.

Review by @axel22

### DIFF
--- a/src/library/scala/collection/concurrent/TrieMap.scala
+++ b/src/library/scala/collection/concurrent/TrieMap.scala
@@ -932,6 +932,33 @@ extends scala.collection.concurrent.Map[K, V]
     if (nonReadOnly) readOnlySnapshot().iterator
     else new TrieMapIterator(0, this)
 
+  ////////////////////////////////////////////////////////////////////////////
+  //
+  // SI-10177 These methods need overrides as the inherited implementations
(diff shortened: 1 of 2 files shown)


src/library/scala/collection/concurrent/TrieMap.scala
```diff
@@ -912,6 +913,18 @@ extends scala.collection.concurrent.Map[K, V]
       }
     }
   }
+  def fail[K, B](self: TrieMap[K, V]): Option[V] = {
```
axel22: Looks good to me overall, but I failed to understand what `fail` is for. And why does it have a `B` type parameter?

retronym: Oh, that was me just hand-inlining some code to add to the bug report. Wasn't intended to be included here, will remove.

axel22:
LGTM

PR #5689, opened 2017-02-13
scala-xml dependency should be provided

ashawley:
Nothing should transitively depend on the scala-xml module except for
scaladoc and library-all.

Compiler needs scala-xml as a build dependency, but not at runtime.

The user is supposed to provide scala-xml as a dependency at runtime.
This fixes an issue where the classpath in SBT shadows the scala-xml
module, see scala/scala-module-dependency-sample#14.

Unfortunately, scaladoc needs to be pulled out of the compiler.  This
is a first attempt at mangling the build so that scaladoc is a
separate jar.  Following on, SBT will need to know how to run scaladoc
as a separate jar.

Previously #5609 for 2.11.  For 2.13 is tracked at scala/scala-dev#291

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -522,7 +522,7 @@ lazy val scaladoc = configureAsSubproject(project)
   .settings(
     name := "scala-compiler-doc",
     description := "Scala Documentation Generator",
-    libraryDependencies ++= Seq(scalaXmlDep, partestDep),
+    libraryDependencies ++= Seq(scalaXmlDep, partestDep), // Good here
     includeFilter in unmanagedResources in Compile := "*.html" | "*.css" | "*.gif" | "*.png" | "*.js" | "*.txt" | "*.svg" | "*.eot" | "*.woff" | "*.ttf"
   )
   .dependsOn(compiler)
--- a/build.sbt
+++ b/build.sbt
@@ -395,17 +395,16 @@ lazy val compiler = configureAsSubproject(project)
     description := "Scala Compiler",
     libraryDependencies ++= Seq(antDep, asmDep),
     // These are only needed for the POM:
-    libraryDependencies ++= Seq(scalaXmlDep, jlineDep % "optional"),
+    libraryDependencies ++= Seq(jlineDep % "optional"),
     buildCharacterPropertiesFile := (resourceManaged in Compile).value / "scala-buildcharacter.properties",
     resourceGenerators in Compile += generateBuildCharacterPropertiesFile.map(file => Seq(file)).taskValue,
-    // this a way to make sure that classes from interactive and scaladoc projects
+    // this a way to make sure that classes from interactive and repl projects
     // end up in compiler jar. note that we need to use LocalProject references
     // (with strings) to deal with mutual recursion
     products in Compile in packageBin :=
       (products in Compile in packageBin).value ++
         Seq((dependencyClasspath in Compile).value.find(_.get(moduleID.key) == Some(asmDep)).get.data) ++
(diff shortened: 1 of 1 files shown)


lrytz:
sbt's `compiler-interface` needs scaladoc on the classpath:

```
[info] 'compiler-interface' not yet compiled for Scala 2.13.0-20170212-224828-034ccfd. Compiling...
/home/jenkins/tmp/sbt_60a1acc5/xsbt/ScaladocInterface.scala:13: error: object doc is not a member of package scala.tools.nsc
  import scala.tools.nsc.{ doc, Global, reporters }
```

Not sure if there's a way to make this work without requiring a new, patched version of sbt.

ashawley:
> Not sure if there's a way to make this work without requiring a new, patched version of sbt.

Not in my opinion there isn't

https://github.com/scala/scala-module-dependency-sample/pull/14#issuecomment-270952028

lrytz:
Thanks, @ashawley. Since this requires a coordinated effort with sbt, I'm closing this for now to keep our PR queue in shape. We keep track of the issue at https://github.com/scala/scala-dev/issues/291.

build.sbt
```diff
@@ -416,15 +415,13 @@ lazy val compiler = configureAsSubproject(project)
     mappings in Compile in packageSrc ++= {
       val base = (unmanagedResourceDirectories in Compile).value ++
         (unmanagedResourceDirectories in Compile in LocalProject("interactive")).value ++
-        (unmanagedResourceDirectories in Compile in LocalProject("scaladoc")).value ++
         (unmanagedResourceDirectories in Compile in LocalProject("repl")).value
       base ** ((includeFilter in unmanagedResources in Compile).value || "*.scala" || "*.psd" || "*.ai" || "*.java") pair relativeTo(base)
     },
-    // Include the additional projects in the scaladoc JAR:
+    // Include the additional projects in the compiler JAR:
```
dwijnand, line 421: this was correct the way it was btw, it means the jar of scaladocs of the compiler

ashawley: Ok, I see, now.  You're right.  I had misunderstood.  Thanks for reviewing that.

ashawley:
I looked in to how sbt needed to be patched, but I didn't get far.

First, I tried to provide the separate scaladoc dependency to sbt's compiler interface.  It failed because sbt is still compiling itself with 2.10

```
(compileInterfaceProj/*:update) sbt.ResolveException: unresolved dependency:
 org.scala-lang#scala-compiler-doc;2.10.6: not found
```

My naive attempt is ashawley/sbt@255b228

So that suggests fixing the scala-xml dependency with scala-compiler-doc in 2.13 would actually require a new 2.10 release with a separate scala-compiler-doc jar just for sbt.  sbt would use that release of Scala 2.10 and the separate jar going forward.

In addition to making the separate jar, the new scala 2.10 release would still, for compatibility sake, provide a scala-compiler with the scala-compiler-doc code in it.

Will there be another 2.10 release of Scala to include this kind of a build change?

If there is, I can try to research making the 2.10 Ant build change and then look in to whether patching sbt is possible.

dwijnand:
I don't think there would be a need for a new release of 2.10/2.11. The compiler bridge is compiled at runtime of _using_ sbt (you might've seen before "compiling the compiler-interface for scala 2.11.8..."). So the place to add scala-compiler-doc is somewhere in the vicinity of compile/ivy/src/main/scala/sbt/compiler/ComponentCompiler.scala.

See https://github.com/sbt/sbt/commit/90a8423d8e20b90b327f5a8893a4fe7460ac67e7#diff-a1981f84fdd4c348d65e4a883adb3137

PR #5690, opened 2017-02-13
SI-10180 Scan generic signatures when populating InnerClass attribute

retronym:
Generic signatures, like descriptors, include class names directly into the
grammar, rather than referring to class constants in the constant pool.

For the benefit of other compilers and tools consuming generic signatures,
we should record the metadata for inner classes that the reference.

This commit includers a parser for generic signatures that is designed for
performance with the simplifying assumption that it need only read
internal names. I started with the existing parser in ClassfileParser
that builds up symbols/types based on generic signatures. I don't see
an easy way to share more code between the two parsers without compromising
performance.

The new parser is used in the backend to trawl for referenced names in
the signatures of class definitions, field types, and method types.

Review by @lrytz. This is an old bug, so I'm happy to defer it to 2.12.3 if we don't have time to review it for 2.12.2

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
@@ -275,7 +275,7 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
         visitInternalName(internalName.toString)
       }
       override def raiseError(msg: String): Unit = {
-        // don't crash on invalid generic signatures
+        throw new FatalError(msg)
       }
     }
 
--- a/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
@@ -14,6 +14,7 @@ import scala.tools.asm.{Handle, Type}
 import scala.tools.nsc.backend.jvm.BTypes._
 import scala.tools.nsc.backend.jvm.GenBCode._
 import scala.tools.nsc.backend.jvm.opt.BytecodeUtils._
+import scala.tools.nsc.symtab.classfile.JavaSignatureWalker
 
 /**
  * This component hosts tools and utilities used in the backend that require access to a `BTypes`
@@ -269,6 +270,15 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
       }
     }
 
+    object sigWalker extends JavaSignatureWalker {
+      override def visitName(internalName: CharSequence): Unit = {
+        visitInternalName(internalName.toString)
+      }
+      override def raiseError(msg: String): Unit = {
+        // don't crash on invalid generic signatures
+      }
+    }
+
     // either an internal/Name or [[Linternal/Name; -- there are certain references in classfiles
     // that are either an internal name (without the surrounding `L;`) or an array descriptor
     // `[Linternal/Name;`.
@@ -304,6 +314,9 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
       case _ => // skip over primitive types
     }
 
+    def visitSignature(sig: String) =
+      if (sig != null) sigWalker.walk(sig)
+
     def visitConstant(const: AnyRef): Unit = const match {
       case t: Type => visitDescriptor(t.getDescriptor)
       case _ =>
@@ -331,6 +344,7 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
     visitInternalName(classNode.superName)
     classNode.interfaces.asScala foreach visitInternalName
+    visitSignature(classNode.signature)
     visitInternalName(classNode.outerClass)
 
     visitAnnotations(classNode.visibleAnnotations)
@@ -340,6 +354,7 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
     for (f <- classNode.fields.asScala) {
       visitDescriptor(f.desc)
+      visitSignature(f.signature)
       visitAnnotations(f.visibleAnnotations)
       visitAnnotations(f.visibleTypeAnnotations)
       visitAnnotations(f.invisibleAnnotations)
@@ -348,7 +363,7 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
     for (m <- classNode.methods.asScala) {
       visitDescriptor(m.desc)
-
+      visitSignature(m.signature)
       visitAnnotations(m.visibleAnnotations)
       visitAnnotations(m.visibleTypeAnnotations)
       visitAnnotations(m.invisibleAnnotations)
--- a/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
@@ -263,7 +263,10 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
     def visitInternalName(internalName: InternalName): Unit = if (internalName != null) {
       val t = classBTypeFromParsedClassfile(internalName)
-      if (t.isNestedClass.get) innerClasses += t
+      t.isNestedClass match {
+        case Right(true) => innerClasses += t
+        case _ => // ignores absent classes
+      }
     }
 
     // either an internal/Name or [[Linternal/Name; -- there are certain references in classfiles
--- a/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
@@ -686,7 +686,7 @@ abstract class ClassfileParser {
           accept(';')
           tpe
         case ARRAY_TAG =>
-          while ('0' <= sig.charAt(index) && sig.charAt(index) <= '9') index += 1
+          while ('0' <= sig.charAt(index) && sig.charAt(index) <= '9') index += 1 // TODO this line doesn't seem to correspond to the spec, remove?
           var elemtp = sig2type(tparams, skiptvs)
           // make unbounded Array[T] where T is a type variable into Array[T with Object]
           // (this is necessary because such arrays have a representation which is incompatible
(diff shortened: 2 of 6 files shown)


lrytz:
```
[info] 'compiler-interface' not yet compiled for Scala 2.12.2-20170213-144915-2fcd948. Compiling...
java.lang.StringIndexOutOfBoundsException: String index out of range: -2147483648
	at java.lang.String.charAt(String.java:658)
	at scala.tools.nsc.symtab.classfile.JavaSignatureWalker.current$1(JavaSignatureWalker.scala:30)
	at scala.tools.nsc.symtab.classfile.JavaSignatureWalker.parseTypeArgs$1(JavaSignatureWalker.scala:41)
	at scala.tools.nsc.symtab.classfile.JavaSignatureWalker.parseMethodOrFieldType$1(JavaSignatureWalker.scala:55)
	at scala.tools.nsc.symtab.classfile.JavaSignatureWalker.walk(JavaSignatureWalker.scala:104)
	at scala.tools.nsc.backend.jvm.analysis.BackendUtils.visitSignature$1(BackendUtils.scala:315)
```

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -687,7 +687,7 @@ abstract class ClassfileParser {
           accept(';')
           tpe
         case ARRAY_TAG =>
-          while ('0' <= sig.charAt(index) && sig.charAt(index) <= '9') index += 1
+          while ('0' <= sig.charAt(index) && sig.charAt(index) <= '9') index += 1 // TODO this line doesn't seem to correspond to the spec, remove?
```
lrytz, line 689: Agreed. It goes back to the first version in 0b2d0a6.

src/compiler/scala/tools/nsc/symtab/classfile/JavaSignatureWalker.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+    var index = 0
+    val end = sig.length
+    def accept(ch: Char) {
+      assert(current() == ch, (current(), ch))
+      index += 1
+    }
+    def skipSubName(isDelimiter: Char => Boolean): Unit = {
+      val start = index
+      while (!isDelimiter(current())) { index += 1 }
```
lrytz: If `current()` doesn't stop / throw after `end`, it might be better to check for `\0` to prevent an infinite loop.

src/compiler/scala/tools/nsc/symtab/classfile/JavaSignatureWalker.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+    def appendSubName(builder: java.lang.StringBuilder, isDelimiter: Char => Boolean): CharSequence = {
+      val start = index
+      skipSubName(isDelimiter)
+      builder.append(sig, start, index)
+    }
+    def current(): Char = {
+      if (index < sig.length())
```
lrytz: `sig.length` / `end`. I would prefer if `current` throws when the index is out of bounds (fail fast).

lrytz (commented):
Looks good - I couldn't spot the reason for the crash just by looking at the code.

lrytz:
Oh, the `StringIndexOutOfBoundsException` is at `-2147483648`, which is `Int.MinValue`, so it's an overflow. The while loop in `parseTypeArgs` continues because `current` keeps returning `\0`.

lrytz:
I think `current` should throw and show the `sig` in the exception message if the index is out of bounds.

retronym:
```
[info] Compiling 293 Scala sources to /home/jenkins/workspace/scala-2.12.x-validate-test/build/quick/classes/compiler...
java.lang.AssertionError: assertion failed: The classfile for scala/tools/reflect/ToolBoxFactory$ToolBoxImpl$withCompilerApi$$api$ could not be found on the compilation classpath.
	at scala.tools.nsc.backend.jvm.BackendReporting$RightBiasedEither$.get$extension(BackendReporting.scala:50)
	at scala.tools.nsc.backend.jvm.analysis.BackendUtils.scala$tools$nsc$backend$jvm$analysis$BackendUtils$$visitInternalName$1(BackendUtils.scala:267)
	at scala.tools.nsc.backend.jvm.analysis.BackendUtils$sigWalker$2$.visitName(BackendUtils.scala:272)
	at scala.tools.nsc.symtab.classfile.JavaSignatureWalker.parseMethodOrFieldType$1(JavaSignatureWalker.scala:67)
```

```
⚡ javap -s -private -v -classpath build/quick/classes/compiler  'scala/tools/reflect/ToolBoxFactory$ToolBoxImpl$withCompilerApi$' 2>&1  | grep 'Signature.*\$api'
    Signature: #20                          // Lscala/tools/reflect/ToolBoxFactory<TU;>.ToolBoxImpl$withCompilerApi$$api$;
    Signature: #119                         // ()Lscala/tools/reflect/ToolBoxFactory<TU;>.ToolBoxImpl$withCompilerApi$$api$;
```

```
build/quick/classes/compiler/scala/tools/reflect/ToolBoxFactory$ToolBoxImpl$withCompilerApi$api$.class
```

Looks like an inconsistency between our generic signatures (`...withCompilerApi$$api$`) and class names (`...withCompilerApi$api$.class`). We should fix that, but also make this a non-fatal error in the backend.

retronym:
Pushed an extra WIP commit to fix the generic signatures and avoid crashing in the backend when referenced classfiles are missing during inner class table population.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -286,7 +286,18 @@ abstract class Erasure extends InfoTransform
               (
                 if (needsJavaSig(preRebound, Nil)) {
                   val s = jsig(preRebound, existentiallyBound)
-                  if (s.charAt(0) == 'L') s.substring(0, s.length - 1) + "." + sym.javaSimpleName
+                  if (s.charAt(0) == 'L') {
+                    val withoutSemicolon = s.substring(0, s.length - 1)
+                    val withoutOwingModuleDollar =
```
lrytz: typo `Owing`

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -286,7 +286,18 @@ abstract class Erasure extends InfoTransform
...
+                      if (preRebound.typeSymbolDirect.isModuleClass && sym.isModuleOrModuleClass)
+                        // mimic the (questionable) choice of `flatName` to use the owning module's
```
lrytz: `flatName` is not an obvious reference - can you give a more specific one?

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -91,7 +91,7 @@ abstract class Erasure extends InfoTransform
   // a close angle bracket '>'.  Any which do not are replaced with '$'.
   // This arises due to multiply nested classes in the face of the
   // rewriting explained at rebindInnerClass.   This should be done in a
-  // more rigorous way up front rather than catching it after the fact,
+  // more rigorous way up   front rather than catching it after the fact,
```
lrytz: whitespace

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -263,7 +263,10 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
 
     def visitInternalName(internalName: InternalName): Unit = if (internalName != null) {
       val t = classBTypeFromParsedClassfile(internalName)
-      if (t.isNestedClass.get) innerClasses += t
+      t.isNestedClass match {
+        case Right(true) => innerClasses += t
+        case _ => // ignores absent classes
+      }
```
lrytz, line 270: I would argue that compilation should require this class to be on the classpath and fail otherwise. If we just ignore it here, the resulting classfile is different whether or not something is on the classpath or not, which is problematic IMO.

src/compiler/scala/tools/nsc/symtab/classfile/JavaSignatureWalker.scala
lrytz: ping

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
retronym: I see your point. OTOH, I'm hesitant to turn remaining bugs in our generic signatures into fatal errors as the code that generates them is somewhat brittle. Maybe a warning is appropriate?

retronym: I pushed a commit to make it fatal, we can run that through the community build to get some more data.

src/compiler/scala/tools/nsc/transform/Erasure.scala
retronym: done

src/compiler/scala/tools/nsc/transform/Erasure.scala
retronym: fixed

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
lrytz: > turn remaining bugs in our generic signatures into fatal errors

My worry is that ignoring errors here applies not only to buggy generic signatures, but also ignores classfiles that are missing on the classpath. This should be an error I think.

src/compiler/scala/tools/nsc/symtab/classfile/JavaSignatureWalker.scala
```diff
@@ -0,0 +1,113 @@
...
+            }
+          }
+
+          val internalName = new java.lang.StringBuilder
+          appendSubName(internalName, c => c == ';' || c == '<')
+          parseTypeArgs()
+          while (current() == '.') {
+            accept('.')
+            internalName.append('$')
+            appendSubName(internalName, c => c == ';' || c == '<' || c == '.')
```
lrytz, line 64: We should call `visitName` for intermediate names here. 

```
class A<T> {
  class B<U> {
    class C { }
  }
}

class T {
  A<String>.B<Object>.C foo() { return null; }
}
```

should give two entries in A: `A$B` and `A$B$C`.

No need to update this PR, I'm creating a best blend of your PR and mine (#5925).

lrytz:
Closing in favor of #5925

PR #5691, opened 2017-02-13
EOL is System.lineSeparator

som-snytt:
The old EOL is EOL. Use `println()` when feasible and obvious.

Minor cleanup of surrounding code.

### DIFF
--- a/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
@@ -7,9 +7,10 @@ package scala
 package tools.nsc
 package reporters
 
-import java.io.{ BufferedReader, PrintWriter }
-import scala.reflect.internal.util._
-import StringOps._
+import java.io.{BufferedReader, PrintWriter}
+import scala.reflect.internal.util.{Position, StringOps}
+import Position.formatMessage
+import StringOps.{countElementsAsString => countAs, trimAllTrailingSpace => trimTrailing}
 
 /** This class implements a Reporter that displays messages on a text console.
  */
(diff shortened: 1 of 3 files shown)


lrytz:
There's a lot of new newlines in repl test transcripts, probably unintended?

som-snytt:
@lrytz Thanks for checking. I bet you thought this was a trivial review. I didn't have time to run tests on the laptop. The problem was `ILoop.runForTranscript` overrides `write(String)` to strip margin and it ignored empty lines instead of writing them. So empty lines were missing.

lrytz:
Thanks! What prompted this change, is `compat.Platform.EOL` deprecated / not recommended?

som-snytt:
Motivated by yesterday's gitter:

> that's actually vi. scalac emits a mix of line endings, so vim infers unix. This is the output from partest. The compiler output "one error found" with the wrong line ending. You'd think ConsoleReporter would be console-aware. https://github.com/scala/scala/blob/v2.12.1/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala#L43

And the move away from the traditional EOL.

lrytz:
Right, I understand `lineSeparator` vs `\n`, but was wondering why 

```
-import scala.compat.Platform.EOL
+import java.lang.System.{lineSeparator => EOL}
```

The former is an alias for the latter as far as I can tell.

SethTisue:
it's because of https://github.com/scala/scala/pull/5677 which gets rid of scala.compat

som-snytt:
Thanks @SethTisue , yes that's what I meant. Zero dependencies! I hesitate to use the `f`-interpolator. I suppose they can take away all our old half-broken toys because they're giving us shiny new enums.

som-snytt:
I just realized `old EOL is EOL` works on a couple of levels, because you could also say `old end-of-line is at the end of the line.`

Also eerie that this is my first contribution where I successfully reduced the line count. I guess mostly by making longer lines.

lrytz (approved):

lrytz:
Thank you!

adriaanm:
Poor old eol'er.

PR #5696, opened 2017-02-14
during bootstrap, don't build scala-parser-combinators or -swing

SethTisue:
I see no further reason to include them in the bootstrap process.
Adriaan agrees (in-person discussion).

note that the community build makes sure the modules still compile as
we make changes to Scala.

as for actually publishing scala-parser-combinators and scala-swing
when we make binary incompatible changes to Scala, I think it's fine
to consider that properly the job of the module maintainers, rather
than properly the job of this script.  (if we want more automation
on that, we could make some elsewhere, keeping that concern separate
from actual bootstrapping concerns.)

scala-xml, scala-partest, and scalacheck (commented out at the moment)
remain in this script since they are all actually still part of the
bootstrapping picture.

fixes https://github.com/scala/scala-dev/issues/302 , so that
integrate-bootstrap will pass again (currently it fails, since
scala-parser-combinators merged Scala.js support)

### DIFF
--- a/scripts/jobs/integrate/bootstrap
+++ b/scripts/jobs/integrate/bootstrap
@@ -178,8 +178,7 @@ sbtResolve() {
       'show update' >> $baseDir/logs/resolution 2>&1
 }
 
-# Oh boy... can't use scaladoc to document scala-xml/scala-parser-combinators
-# if scaladoc depends on the same version of scala-xml/scala-parser-combinators.
+# Oh boy... can't use scaladoc to document scala-xml if scaladoc depends on the same version of scala-xml.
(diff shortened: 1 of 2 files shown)


SethTisue:
review by @retronym, who straps more boo than the rest of us put together

SethTisue:
test CI run: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-bootstrap/784/consoleFull (404 while Jenkins runs to the store for tennis balls and mustache wax)

SethTisue:
Jenkins run was green.

adriaanm (approved):

PR #5704, opened 2017-02-15
SI-10190 Elide string to empty instead of null

som-snytt:
Avoid NPE when eliding string-valued functions.

For example, `log(s"$cheap$expensive")` needn't print null.

This is a natural and inexpensive way to elide strings.

Also avoid using assert when testing elide-below.


som-snytt:
You'd think any other ref type would be required to be `Option` and elide to `None`.

Ichoran:
I'm not very fond of this idea.  In particular, it doesn't match the output of `"%s".format(null: String)`

Ichoran:
(Sorry, I misunderstood what you meant.  Never mind!)

lrytz:
sounds good to me

lrytz (approved):

PR #5708, opened 2017-02-16
SI-10194: Fix abstract type resolution for overloaded HOFs

szeiger:
Types in the applicable overload alternatives need to be seen from
the respective owners of the individual alternative, not from the
target’s owner (which can be a subtype of the types that define the
methods).

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3333,15 +3333,15 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           def handleOverloaded = {
             val undetparams = context.undetparams
 
-            def funArgTypes(tps: List[Type]) = tps.map { tp =>
-              val relTp = tp.asSeenFrom(pre, fun.symbol.owner)
+            def funArgTypes(tpAlts: List[(Type, Symbol)]) = tpAlts.map { case (tp, alt) =>
+              val relTp = tp.asSeenFrom(pre, alt.owner)
               val argTps = functionOrSamArgTypes(relTp)
               //println(s"funArgTypes $argTps from $relTp")
               argTps.map(approximateAbstracts)
             }
 
-            def functionProto(argTps: List[Type]): Type =
-              try functionType(funArgTypes(argTps).transpose.map(lub), WildcardType)
+            def functionProto(argTpWithAlt: List[(Type, Symbol)]): Type =
+              try functionType(funArgTypes(argTpWithAlt).transpose.map(lub), WildcardType)
               catch { case _: IllegalArgumentException => WildcardType }
 
             // To propagate as much information as possible to typedFunction, which uses the expected type to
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3333,15 +3333,14 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+            def funArgTypes(relTps: List[Type]) = relTps.map { relTp =>
               val argTps = functionOrSamArgTypes(relTp)
               //println(s"funArgTypes $argTps from $relTp")
               argTps.map(approximateAbstracts)
             }
 
-            def functionProto(argTps: List[Type]): Type =
-              try functionType(funArgTypes(argTps).transpose.map(lub), WildcardType)
+            def functionProto(argTpWithAlt: List[(Type, Symbol)]): Type =
+              try functionType(funArgTypes(argTpWithAlt.map { case (tp, alt) => tp.asSeenFrom(pre, alt.owner) }).transpose.map(lub), WildcardType)
```
adriaanm: Why not keep the`asSeenFrom` in `funArgTypes`? There's a lot going on on this line already.

adriaanm (changes requested):
Good catch! Would you mind doing the refactoring I suggested if it makes sense?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
szeiger: I tried to simplify it but that didn't really work out.

szeiger:
Updated (moving `asSeenFrom` back down into `funArgTypes`) and rebased (to clear up the MiMa failure).

src/compiler/scala/tools/nsc/typechecker/Typers.scala
lrytz: @szeiger I don't understand what didn't work out - it seems it should be easy to move the `asSeenFrom` into `funArgTypes`?

szeiger: @lrytz That's where it used to be. Moving it out didn't really make the code any simpler.

lrytz: > Moving it out didn't really make the code any simpler.

Then why was it moved out? I feel like we're talking past each other.. :)

szeiger: To make it simpler, obviously :)

lrytz: It's such a detail that I'm sorry to go back and forth so many times.. I just don't understand.. IIUC adriaan suggests to pass `argTpWithAlt` as argument to `funArgTypes` directly and do the `asSeenFrom` there, which also looks better to me. You even seem to agree, as you say

  > Moving it out [of `funArgTypes`?] didn't really make the code any simpler

so why did you move it out?

szeiger: Because I expected it would. There really isn't any deeper or better explanation...

adriaanm: > Updated (moving asSeenFrom back down into funArgTypes) 

I'm not seeing this update in the diff.

szeiger: Huh, still had some unstaged changes lying around. Updated again.

lrytz: That explains my confusion :-)

lrytz (approved):

PR #5711, opened 2017-02-17
Faster and simpler Java 9 classpath implementation

retronym:
- Take advantage of the `/packages` index provided by the
    jrt file system to avoid (expensive) Files.exist for
    non-existent entries across the full list of modules.
  - Extends ClassPath directly which leads to a simpler
    implemnentation that using the base class.
  - Add a unit test that shows we can read classes and packages
    from the Java standard library.

Fixes scala/scala-dev#306

With this change bootstrap time under Java 9 was comparable to
Java 8. Before, it was about 40% slower.

Review by @lrytz. I'm getting close tp having a Java 9 build
on CI, but for now this is manually tested:

```
⚡ (java_use 9; qscala)
Welcome to Scala 2.12.2-20170217-114130-06f1c83 (Java HotSpot(TM) 64-Bit Server VM, Java 9-ea).

scala> new java.lang.Object
res0: Object = java.lang.Object@7219ac49
```

```
(java_use 9; sbt -sbt-version 0.13.14-b646662 -Dstarr.version=2.12.2-ec84a4f-SNAPSHOT -Dscala.ext.dirs=/Library/Java/JavaVirtualMachines/jdk-9.jdk/Contents/Home/scala-ext)

> junit/testOnly scala.tools.nsc.classpath.JrtClassPathTest
[info] Compiling 1 Scala source to /Users/jz/code/scala-java9-ci/target/junit/test-classes...
[info] Test run started
[info] Test scala.tools.nsc.classpath.JrtClassPathTest.lookupJavaClasses started
[info] Test run finished: 0 failed, 0 ignored, 1 total, 0.434s
[info] Passed: Total 1, Failed 0, Errors 0, Passed 
```

### DIFF
--- a/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
+++ b/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
@@ -10,9 +10,10 @@ import java.util.function.IntFunction
 import java.util
 import java.util.Comparator
 
-import scala.reflect.io.{AbstractFile, PlainFile}
+import scala.reflect.io.{AbstractFile, PlainFile, PlainNioFile}
 import scala.tools.nsc.util.{ClassPath, ClassRepresentation}
 import FileUtils._
+import scala.collection.JavaConverters._
 
 /**
  * A trait allowing to look for classpath entries in directories. It provides common logic for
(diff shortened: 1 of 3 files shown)


lrytz:
```
[info] Test scala.tools.nsc.classpath.JrtClassPathTest.lookupJavaClasses started
[error] Test scala.tools.nsc.classpath.JrtClassPathTest.lookupJavaClasses failed: java.lang.AssertionError: null, took 0.005 sec
[error]     at scala.tools.nsc.classpath.JrtClassPathTest.lookupJavaClasses(JrtClassPathTest.scala:32)
[error]     ...
```

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -121,51 +122,78 @@ trait JFileDirectoryLookup[FileEntryType <: ClassRepresentation] extends Directo
...
-  protected def isPackage(f: Path): Boolean = Files.isDirectory(f) && mayBeValidPackage(f.getFileName.toString)
+
+  override private[nsc] def list(inPackage: String): ClassPathEntries =
+    if (inPackage == "") ClassPathEntries(packages(inPackage), Nil)
+    else ClassPathEntries(packages(inPackage), classes(inPackage))
 
   def asURLs: Seq[URL] = Seq(dir.toUri.toURL)
-  def asClassPathStrings: Seq[String] = asURLs.map(_.toString)
+  // We don't yet have a scheme to represent the JDK modules in our `-classpath`.
+  // java models them as entries in the new "module path", we'll probably need to follow this.
```
lrytz, line 182: maybe return some dummy string to indicate the lack to clients?

src/compiler/scala/tools/util/PathResolver.scala
```diff
@@ -234,7 +234,7 @@ final class PathResolver(settings: Settings) {
 
     // Assemble the elements!
     def basis = List[Traversable[ClassPath]](
-      JImageDirectoryLookup.apply(),                // 0. The Java 9 classpath (backed by the jrt:/ virtual system)
+      JrtClassPath.apply(),                // 0. The Java 9 classpath (backed by the jrt:/ virtual system)
```
lrytz: (indentation)

lrytz (approved):
nice!

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: I thought it would be best to omit it for now in case the returned string is used to form an actual compiler classpath that is expected to work. No entry is need on `-classpath` to make the java standard library available.

retronym:
Test and indentation fixed.

dwijnand:
Woah. That flew in. Nice work.

PR #3930, opened 2014-08-19
make ConsoleReporter handle both stdout and stderr

gourlaysama:
`scala` and `scalac` always had a problem with deciding where to send
messages in the console and just sent everything to stderr, so it is
time to improve that:

`ConsoleReporter` now takes two `PrintWriter`, one for errors/warnings and
one for normal messages. The short constructor sets them to stderr and
stdout, respectively.

This finally makes scala(c) ouput like any normal program. For example:

``` sh
$ scalac -help > /dev/null # will now print nothing

$ scala -nc -e "val" > /dev/null # will print an error message

$ scala -nc -e "println(1)" > /dev/null # but this won't
```

People calling the old constructor (with a single writer) get no change
in behavior: everything goes to the writer.

---

The second commit uses that to fix a 6-year-old issue, [SI-1264](https://issues.scala-lang.org/browse/SI-1264).

PR #5714, opened 2017-02-17
SI-10178 Route reporter.echo to stdout

dragos:
`echo` is currently used only for usage information, and that makes a lot more sense to 
go to stdout instead of stderr. This allows grepping through the extensive list of
compiler options.

```
$ build/quick/bin/scalac -Y | grep macro
  -Ymacro-debug-lite                       Trace essential macro-related activities.
  -Ymacro-debug-verbose                    Trace all macro-related activities: compilation, generation of synthetics, classloading, expansion, exceptions.
  -Ymacro-expand:<policy>                  Control expansion of macros, useful for scaladoc and presentation compiler. Choices: (normal,none,discard), default: normal.
  -Ymacro-no-expand                        Don't expand macros. Might be useful for scaladoc and presentation compiler, but will crash anything which uses macros and gets past typer.
                                             deprecated: Use -Ymacro-expand:none
```

### DIFF
--- a/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala
@@ -14,8 +14,10 @@ import StringOps.{countElementsAsString => countAs, trimAllTrailingSpace => trim
 
 /** This class implements a Reporter that displays messages on a text console.
  */
-class ConsoleReporter(val settings: Settings, reader: BufferedReader, writer: PrintWriter) extends AbstractReporter {
-  def this(settings: Settings) = this(settings, Console.in, new PrintWriter(Console.err, true))
+class ConsoleReporter(val settings: Settings, reader: BufferedReader, writer: PrintWriter, echoWriter: PrintWriter) extends AbstractReporter {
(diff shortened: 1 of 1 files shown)


gourlaysama:
Ah, this brings back memories: #3930

👍 for moving to the GNU guidelines :)

dragos:
I wasn't aware of your PR, @gourlaysama, thanks for the pointer.

dragos:
It seems this was a binary incompatible change, I'll look into it.

dragos:
The failure is a scalacheck test on Duration. I don't see how that can have anything to do with my changes, though.

SethTisue:
yeah that's https://github.com/scala/scala-dev/issues/296

som-snytt:
LGTM

Taking echo output is cleaner than taking INFO or some slice of error output.

I hope there is no infrastructure that relies on the previous behavior. Spark on 2.11 does not, it seems.

I guess Partest may not care because https://issues.scala-lang.org/browse/SI-7003 or https://github.com/scala/scala-partest/commit/d4bd4c3ec07e531a4c744ed73e5946b69b9843d2

som-snytt:
BTW, my previous rationalization was: stdout meant I did my job (compiling source), stderr is anything to indicate why I didn't fulfill that primary mission (errors, they stopped me to ask for help).

PR #5701, opened 2017-02-15
[ci: last-only] merge 2.11.x onto 2.12.x (February 14, 2017)

SethTisue:
many of the PRs merged for 2.11.9 were backports, so we proceed
by considering each PR individually.

`git log --merges origin/2.11.x ^origin/2.12.x` shows:

* 0965028809 - Merge pull request #5695 from SethTisue/fix-indexedseqtest-under-ant (6 hours ago) <Seth Tisue>
* 214a158f67 - Merge pull request #5664 from adriaanm/rework-coll-perf (6 days ago) <Adriaan Moors>
* 6644017dad - Merge pull request #5631 from adriaanm/ticket/9013 (3 weeks ago) <Adriaan Moors>
* 90a764bcba - Merge pull request #5645 from sjrd/avoid-array-oob-in-rbtree-iterator-from (3 weeks ago) <Adriaan Moors>
* a5d38ea334 - Merge pull request #5630 from adriaanm/rebase-5557 (5 weeks ago) <Adriaan Moors>
* 36967321c7 - Merge pull request #5632 from adriaanm/ticket/9114 (5 weeks ago) <Adriaan Moors>
* fac487e397 - Merge pull request #5633 from adriaanm/ticket/9331 (5 weeks ago) <Adriaan Moors>
* 32a7461037 - Merge pull request #5635 from adriaanm/jline-2.14.3 (5 weeks ago) <Adriaan Moors>
* 4f5c30e10a - Merge pull request #5615 from monkey-mas/modify-ArrayBuilder-reusability-bug-2016-12-24 (5 weeks ago) <Stefan Zeiger>
* 1b6b19b701 - Merge pull request #5626 from adriaanm/rebase-5566 (6 weeks ago) <Adriaan Moors>
* 18269eabe1 - Merge pull request #5612 from adriaanm/patmat_cleanup (6 weeks ago) <Adriaan Moors>
* 8367bf68c1 - Merge pull request #5624 from Jasper-M/patch-1 (6 weeks ago) <Lukas Rytz>

5612 is a special case: part backport, part not.  I consulted with
Adriaan and he said to handle 3 commits as follows:

* 014ebc4 merge forward, resolving easy merge conflict manually
* 80ce231 do not merge, 2.12 already has the fix
* da98683 backport, do not merge

to avoid extra merge commits, we don't need to merge every PR individually,
we can chunk them according to whether they are backports or not.  so that yields:

```
git merge                 8367bf68c1   # 5624
[...]

PR #5715, opened 2017-02-17
Update spec build to Jekyll 3.3

janekdb:
Seth Tisue suggested updating the Jekyll dependency for the Scala spec
in scala/scala-lang#559 so that is consistent with the scala-lang.org
and docs.scala-lang.org Web sites.

Upgrading Jekll from 2.5.3 to 3.3.0 produces slight differences in the
HTML output.  The differences in the HTML are isolated to the markup
for source code highlighting.  In the spec, those are Scala code
listings and the blocks of extended-BNF. From what I saw looking at
all the pages, they render exactly the same, pixel-for-pixel, as
before.  I hit print preview and also didn't notice any anomalies.

I tested in Ruby 2.2.5 using rbenv 0.4.0, gem 2.4.5.1 and Bundler
1.13.6.  This is compared to the Travis build that uses Ruby 2.2.0p0
using rvm 1.26.10, gem 2.4.5 and Bundler 1.7.9.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 # To build the spec on Travis CI
 source "https://rubygems.org"
 
-gem "jekyll", "2.5.3"
+gem "jekyll", "3.3.0"
 gem "rouge"
 # gem 's3_website'
 gem "redcarpet", "3.3.2"


janekdb:
`cherry-pick` from https://github.com/scala/scala/commit/1f4cd50fa4088ed18db70e417711760518a1c8ba

janekdb:
Building with jeykyll 3.3.0 instead of 2.5.3 does not change the LaTeX markup.

It does however fatten up Scala code blocks with a slew of taggage. [diff for 07-implicits.html][diff].

[diff]: https://gist.github.com/janekdb/2c56beffb64ca85e373aac31c66cc922

The aim of this commit is to eliminate differences between 2.12.x and 2.13.x while preferring newer versions  of jekyll and friends. I did not see any difference in the rendered code blocks on 2.13.x with Jekyll 2.5.3 vs 3.3.0

SethTisue:
it looks to me like this isn't needed, because #5593 will be merged forward the next time we merge 2.12.x onto 2.13.x. (which I intend to do as soon as #5701 is merged, which I expect will be very soon.)

I apologize for having let 2.13.x fall so far behind 2.12.x. (I was waiting for #5654, which ended up dragging on waaaaaaaay longer than I anticipated.)

PR #5717, opened 2017-02-18
SI-10148 Accept verbose zero

som-snytt:
The test for non-zero must recognize `-0e+00f` and variants.

This is a follow-up to https://github.com/scala/scala/pull/5648

Community build, I think I love you.

You too, @SethTisue

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
@@ -983,6 +983,8 @@ trait Scanners extends ScannersCommon {
 
     def intVal: Long = intVal(negated = false)
 
+    private val zeroFloat = raw"[0.]+(?:[eE][+-]?[0-9]+)?[fFdD]?".r
+
     /** Convert current strVal, base to float value.
      */
     def floatVal(negated: Boolean): Float = {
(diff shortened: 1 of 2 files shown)


Ichoran (changes requested):
This isn't enough.  -0.0000000000000000e14f doesn't pass, but it is identically zero so it should.  The logic should be `strVal.find(c => c != '0' && c != '.').exists(_.isDigit)` or somesuch

som-snytt:
Thanks @Ichoran . I couldn't follow somesuch logic so I tried a regex.

I'm grateful they didn't run community build only to fail on your repo with zero to the nth degree.

som-snytt:
Am I supposed to "dismiss review" after making a change? I don't wish to appear dismissive.

Ichoran (approved):
This looks like it should work.  Not the most efficient way, but presumably we're not spending very much compilation time parsing floating point numbers in source code.

Ichoran:
I don't know how it's supposed to work either.  But my new review evicted the old one.

som-snytt:
More specifically, the test for "did the user write some variant of zero" is invoked rarely, so I focused on maintenance cost over strict efficiency. Moreover, we know the source text is a good float, and we're only checking for good significant digits. We could make it even more efficient if desired.

som-snytt:
Or Adriaan could just merge it.

Ichoran:
Adriaan is even more efficient.

SethTisue:
2.12 community build is green again

PR #5719, opened 2017-02-19
SI-10187 Support mutation of mutable.HashMap in getOrElseUpdate

retronym:
Scala 2.12.1 included optimizations to `HashMape.getOrElseUpdate`
to avoid recomputing the index in the hash table when adding an
the element.

However, this index could be stale if the callback added elements
to the map and triggered a resize.

This commit checks that the table is unchanged before reusing
the index, restoring the 2.12.0 behaviour.

JIRA: https://issues.scala-lang.org/browse/SI-10187

### DIFF
--- a/src/library/scala/collection/mutable/HashMap.scala
+++ b/src/library/scala/collection/mutable/HashMap.scala
@@ -73,10 +73,18 @@ extends AbstractMap[A, B]
   }
 
   override def getOrElseUpdate(key: A, defaultValue: => B): B = {
-    val i = index(elemHashCode(key))
+    val hash = elemHashCode(key)
+    val i = index(hash)
     val entry = findEntry(key, i)
     if (entry != null) entry.value
-    else addEntry(createNewEntry(key, defaultValue), i)
+    else {
+      val table0 = table
+      val default = defaultValue
+      // Avoid recomputing index if the `defaultValue()` hasn't triggered
+      // a table resize.
+      val newEntryIndex = if (table0 eq table) i else index(hash)
+      addEntry(createNewEntry(key, default), newEntryIndex)
+    }
   }
 
   /* inlined HashTable.findEntry0 to preserve its visibility */
--- a/test/junit/scala/collection/mutable/HashMapTest.scala
+++ b/test/junit/scala/collection/mutable/HashMapTest.scala
@@ -1,0 +1,38 @@
+package scala.collection
+package mutable
+
+import org.junit.Assert._
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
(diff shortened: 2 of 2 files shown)


src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -73,10 +73,18 @@ extends AbstractMap[A, B]
   }
 
   override def getOrElseUpdate(key: A, defaultValue: => B): B = {
-    val i = index(elemHashCode(key))
+    val hash = elemHashCode(key)
+    val table0 = table
```
Ichoran: This is too early.  You don't have to store table0 until right before the `default = defaultValue` inside the `else`.

src/library/scala/collection/mutable/HashMap.scala
```diff
@@ -73,10 +73,18 @@ extends AbstractMap[A, B]
...
+    val i = index(hash)
     val entry = findEntry(key, i)
     if (entry != null) entry.value
-    else addEntry(createNewEntry(key, defaultValue), i)
+    else {
+      val default = defaultValue
+      // Avoid recomputing index if the `defaultValue()` hasn't triggered
+      // a table resize.
+      val newEntryIndex = if (table0 eq table) i else index(hash)
+      addEntry(createNewEntry(key, defaultValue), newEntryIndex)
```
Ichoran: This should be `default` not `defaultValue`.

szeiger (changes requested):
Looks good apart from the two issues from Rex's comments.

retronym:
I've updated the code to address the problems @Ichoran pointed out, and also added a test to show we evaluate the `default` callback the right number of times (0 or 1).

l0rinc:
Sorry for the inconvenience, though I think Java 9 explicitly prohibited this behavior (and in Java 8 it just wasn't working correctly)
http://download.java.net/java/jdk9/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function- 
> The remapping function should not modify this map during computation.

retronym:
Yep, we considered just adopting the Java 9 approach of declaring the behaviour undefined, but it seemed easy enough to rework the implementation to keep the old behaviour, which seemed like the conservative thing to do in a minor release.

l0rinc:
I guess recursivity would make a lot of sense in case of memoization, e.g.
```scala
def factorial(num: Int, memo: mutable.HashMap[Int, Int] = mutable.HashMap(1 -> 1)): Int
  = memo.getOrElseUpdate(num, num * factorial(num - 1, memo))
```
(or could this be done with something like Groovy's [memoized](http://docs.groovy-lang.org/2.4.7/html/gapi/groovy/transform/Memoized.html)?)

szeiger (approved):

PR #5722, opened 2017-02-20
Move dealiasing from Uncurry to Erasure

retronym:
The motivation is to make the uncurry info tranform an identity
for more method signatures.

A few spots in erasure needed to be updated to account for
the possibility of aliases, by callng normalize and/or using
baseType rather than just assuming that the type args and class
type params line up.

Fixes scala/scala-dev#297

### DIFF
--- a/src/reflect/scala/reflect/internal/transform/Erasure.scala
+++ b/src/reflect/scala/reflect/internal/transform/Erasure.scala
@@ -111,7 +111,7 @@ trait Erasure {
 
     protected def eraseDerivedValueClassRef(tref: TypeRef): Type = erasedValueClassArg(tref)
 
-    def apply(tp: Type): Type = tp match {
+    def apply(tp: Type): Type = tp.dealias match {
       case ConstantType(ct) =>
         // erase classOf[List[_]] to classOf[List]. special case for classOf[Unit], avoid erasing to classOf[BoxedUnit].
         if (ct.tag == ClazzTag && ct.typeValue.typeSymbol != UnitClass) ConstantType(Constant(apply(ct.typeValue)))
@@ -127,7 +127,7 @@ trait Erasure {
           else typeRef(apply(pre), sym, args map applyInArray)
         else if (sym == AnyClass || sym == AnyValClass || sym == SingletonClass) ObjectTpe
         else if (sym == UnitClass) BoxedUnitTpe
-        else if (sym.isRefinementClass) apply(mergeParents(tp.parents))
+        else if (sym.isRefinementClass) apply(mergeParents(tp.normalize.parents))
         else if (sym.isDerivedValueClass) eraseDerivedValueClassRef(tref)
         else if (sym.isClass) eraseNormalClassRef(tref)
         else apply(sym.info asSeenFrom (pre, sym.owner)) // alias type or abstract type
@@ -142,8 +142,8 @@ trait Erasure {
           // this replaces each typeref that refers to an argument
           // by the type `p.tpe` of the actual argument p (p in params)
           else apply(mt.resultType(mt.paramTypes)))
-      case RefinedType(parents, decls) =>
-        apply(mergeParents(parents))
+      case rt : RefinedType =>
+        apply(mergeParents(rt.normalize.parents))
       case AnnotatedType(_, atp) =>
         apply(atp)
       case ClassInfoType(parents, decls, clazz) =>
@@ -332,7 +332,7 @@ trait Erasure {
         // treat arrays specially
         arrayType(
           intersectionDominator(
-            parents filter (_.typeSymbol == ArrayClass) map (_.typeArgs.head)))
+            parents filter (_.typeSymbol == ArrayClass) map (_.baseType(ArrayClass).typeArgs.head)))
       } else {
         // implement new spec for erasure of refined types.
         def isUnshadowed(psym: Symbol) =
--- a/src/reflect/scala/reflect/internal/transform/Erasure.scala
+++ b/src/reflect/scala/reflect/internal/transform/Erasure.scala
@@ -111,7 +111,7 @@ trait Erasure {
 
     protected def eraseDerivedValueClassRef(tref: TypeRef): Type = erasedValueClassArg(tref)
 
-    def apply(tp: Type): Type = tp.dealias match {
+    def apply(tp: Type): Type = tp match {
       case ConstantType(ct) =>
         // erase classOf[List[_]] to classOf[List]. special case for classOf[Unit], avoid erasing to classOf[BoxedUnit].
         if (ct.tag == ClazzTag && ct.typeValue.typeSymbol != UnitClass) ConstantType(Constant(apply(ct.typeValue)))
@@ -130,7 +130,11 @@ trait Erasure {
         else if (sym.isRefinementClass) apply(mergeParents(tp.normalize.parents))
         else if (sym.isDerivedValueClass) eraseDerivedValueClassRef(tref)
         else if (sym.isClass) eraseNormalClassRef(tref)
-        else apply(sym.info asSeenFrom (pre, sym.owner)) // alias type or abstract type
+        else {
+          val tp1 = if (sym.isAliasType) tp.dealias else tp
+          if (tp1 ne tp) apply(tp1)
+          else apply(sym.info asSeenFrom (pre, sym.owner))
+        } // alias type or abstract type
       case PolyType(tparams, restpe) =>
         apply(restpe)
       case ExistentialType(tparams, restpe) =>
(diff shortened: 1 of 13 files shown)


retronym:
```
[error] Uncaught exception when running partest: java.lang.NoSuchMethodError: scala.tools.cmd.Meta$StdOpts$Bash$.<init>(Lscala/tools/cmd/Meta$StdOpts;)V
sbt.ForkMain$ForkError: java.lang.NoSuchMethodError: scala.tools.cmd.Meta$StdOpts$Bash$.<init>(Lscala/tools/cmd/Meta$StdOpts;)V
	at scala.tools.partest.nest.RunnerSpec$.Bash$lzycompute$1(RunnerSpec.scala:50)
	at scala.tools.partest.nest.RunnerSpec$.Bash(RunnerSpec.scala:50)
	at scala.tools.cmd.Meta$StdOpts.$init$(Meta.scala:25)
	at scala.tools.partest.nest.RunnerSpec$.<init>(RunnerSpec.scala:50)
```

adriaanm:
```
partest --update-check \
  test/files/run/t6308.scala \
  test/files/run/t6555.scala \
  test/files/neg/t3653.scala \
```

progressions, seems like

src/reflect/scala/reflect/internal/transform/Erasure.scala
```diff
@@ -111,7 +111,7 @@ trait Erasure {
 
     protected def eraseDerivedValueClassRef(tref: TypeRef): Type = erasedValueClassArg(tref)
 
-    def apply(tp: Type): Type = tp match {
+    def apply(tp: Type): Type = tp.dealias match {
```
retronym: Probably should do `(if (tp.isRefinement) tp.normalize else tp.dealias)) match` upfront here, rather than the subsequent calls to `normalize` below.

szeiger:
@retronym Is this still on track for M2 or should we reschedule it to M3?

retronym:
@szeiger We can move it. This isn't really user visible, it is just an internal cleanup that should marginally help performance.

retronym:
Rebased, and re-focused this PR on the subject of dealiasing. The patch to change how uncurry/erasure treats the `IMPLICIT` flag is noted in https://github.com/scala/bug/issues/9848

src/reflect/scala/reflect/internal/transform/Erasure.scala
retronym: Or maybe not, `RefinedTypes` are not a common case, so it might be better to keep this code direct.

retronym:
Comparative benchmarks scheduled to [appear soon](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=25b9e2b7d3%7Cccbacdd0b3%7C6d297645f8&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1500851507539&to=1503270707539).

retronym:
The performance gains are modest but measurable, @adriaanm could you take another look at the patch from a correctness/simplicity perspective?

src/reflect/scala/reflect/internal/transform/Erasure.scala
```diff
@@ -142,8 +146,8 @@ trait Erasure {
           // this replaces each typeref that refers to an argument
           // by the type `p.tpe` of the actual argument p (p in params)
           else apply(mt.resultType(mt.paramTypes)))
-      case RefinedType(parents, decls) =>
-        apply(mergeParents(parents))
+      case rt : RefinedType =>
+        apply(mergeParents(rt.normalize.parents)) // RefinedType#normalize
```
adriaanm, line 150: Would it make sense to move RefinedType's `normalizeImpl` here? Also, could we do less work when doing Java-style erasure?

adriaanm (approved):
LGTM, one question that could also be addressed later.

src/reflect/scala/reflect/internal/transform/Erasure.scala
retronym: `JavaErasureMap` overrides `mergeParents`:

```scala
    /** In java, always take the first parent.
     *  An intersection such as `Object with Trait` erases to Object.
     */
    def mergeParents(parents: List[Type]): Type =
      if (parents.isEmpty) ObjectTpe
      else parents.head
```

Rather than:

```scala
    /** In scala, calculate a useful parent.
     *  An intersection such as `Object with Trait` erases to Trait.
     */
    def mergeParents(parents: List[Type]): Type =
      intersectionDominator(parents)
```

The erasure type map is run (just) prior to `phase.erasedTypes`, so `RefinedType.normalize` leans on an internal cache:

```scala
    final override def normalize: Type =
      if (phase.erasedTypes) normalizeImpl
      else {
        if (normalized eq null) normalized = normalizeImpl
        normalized
      }
```

If this cache were null at the time we called `normalize`. If `normalize` is not an identity (ie we need to flatten nested refinements), there is a small inefficiency in:
  - computing the flat list of normalized parents and creating a new `RefinedType` with them and the `decls`
  - ... only to extract the `parents` and discard the rest in `ErasureMap`
  - ... only to use the first parent in `JavaErasureMap`.

I believe most of the time `RefinedType.normalize` is an identity, so we're unlikely to see a performance win from this.

retronym:
Let me `jardiff` this before we merge.

retronym:
`NeedsSigCollector` is not aware of alias types, so we're missing some generic signatures after this patch.

retronym:
Even after my latest change, we're missing generic signatures on static forwarders for module methods that with refinement + typealias.

```diff
--- a/scala/util/control/Exception.class.asm
+++ b/scala/util/control/Exception.class.asm
@@ -123,8 +123,6 @@
     MAXLOCALS = 1

   // access flags 0x9
-  // signature <Ex:Ljava/lang/Throwable;T:Ljava/lang/Object;>(Lscala/Function1<TEx;Ljava/lang/Object;>;Lscala/Function1<TEx;TT;>;Lscala/reflect/ClassTag<TEx;>;)Lscala/PartialFunction<Ljava/lang/Throwable;TT;>;
-  // declaration: scala.PartialFunction<java.lang.Throwable, T> mkCatcher<Ex extends java.lang.Throwable, T>(scala.Function1<Ex, java.lang.Object>, scala.Function1<Ex, T>, scala.reflect.ClassTag<Ex>)
   public static mkCatcher(Lscala/Function1;Lscala/Function1;Lscala/reflect/ClassTag;)Lscala/PartialFunction;
     GETSTATIC scala/util/control/Exception$.MODULE$ : Lscala/util/control/Exception$;
     ALOAD 0
```

minimizes to:

```scala
object Exception {
  type Catcher[+T] = PartialFunction[Throwable, T]
  def mkCatcher[Ex <: Throwable, T](isDef: Ex => Boolean, f: Ex => T) = new Catcher[T] {
    def isDefinedAt(x: Throwable) = true
    def apply(x: Throwable): T = ???
  }
}
```

The difference is buried in:

```scala
      def staticForwarderGenericSignature: String = {
        // scala/bug#3452 Static forwarder generation uses the same erased signature as the method if forwards to.
        // By rights, it should use the signature as-seen-from the module class, and add suitable
        // primitive and value-class boxing/unboxing.
        // But for now, just like we did in mixin, we just avoid writing a wrong generic signature
        // (one that doesn't erase to the actual signature). See run/t3452b for a test case.
        val memberTpe = enteringErasure(moduleClass.thisType.memberInfo(m))
        val erasedMemberType = erasure.erasure(m).apply(memberTpe)
        if (erasedMemberType =:= m.info)
          getGenericSignature(m, moduleClass, memberTpe)
        else null
      }
```

Which ends up with:

```
erasedMemberType = (isDef: Function1, f: Function1)Object
m.info = (isDef: Function1, f: Function1)PartialFunction
```

Furthermore, specialization is damaged:

```diff
diff --git a/scala/tools/nsc/ast/parser/Scanners$ParensAnalyzer.class.asm b/scala/tools/nsc/ast/parser/Scanners$ParensAnalyzer.class.asm
index c653301..0db90df 100644
--- a/scala/tools/nsc/ast/parser/Scanners$ParensAnalyzer.class.asm
+++ b/scala/tools/nsc/ast/parser/Scanners$ParensAnalyzer.class.asm
@@ -1732,12 +1732,14 @@
     ISTORE 19
     ALOAD 0
     INVOKEVIRTUAL scala/tools/nsc/ast/parser/Scanners$ParensAnalyzer.nextToken ()V
-    NEW scala/Tuple2$mcII$sp
+    NEW scala/Tuple2
     DUP
     ILOAD 19
+    INVOKESTATIC scala/runtime/BoxesRunTime.boxToInteger (I)Ljava/lang/Integer;
     ALOAD 4
     GETFIELD scala/runtime/IntRef.elem : I
-    INVOKESPECIAL scala/Tuple2$mcII$sp.<init> (II)V
+    INVOKESTATIC scala/runtime/BoxesRunTime.boxToInteger (I)Ljava/lang/Integer;
+    INVOKESPECIAL scala/Tuple2.<init> (Ljava/lang/Object;Ljava/lang/Object;)V
     GOTO L22
```

I'm going to close this PR on account of the rapidly increasing worm population in the can I'm attempting to open.

retronym:
If we try to pick this up again, I think we should leave refined type normalization (which including dealiasing of its parents) in the `uncurry` info transform. We'll still need to adapt specialization (perhaps it is just the `specializedType` type map that needs alias awareness.

PR #5723, opened 2017-02-20
Fix regression introduced by 5751763

dragos:
enterClass/Module may return an existing symbol, but in 5751763 the return value was dropped leading to assertion failures. This shows up only in the presentation compiler, which explains why it went unnoticed.

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
+++ b/src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
@@ -122,11 +122,16 @@ abstract class SymbolLoaders {
    *  and give them `completer` as type.
    */
   def enterClassAndModule(root: Symbol, name: String, getCompleter: (ClassSymbol, ModuleSymbol) => SymbolLoader) {
-    val clazz = newClass(root, name)
-    val module = newModule(root, name)
-    val completer = getCompleter(clazz, module)
-    enterClass(root, clazz, completer)
-    enterModule(root, module, completer)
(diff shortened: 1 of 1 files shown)


lrytz:
Thank you! Can you think of a way to write a regression test for this?

dragos:
Unfortunately I don't think it's easy. Here's what needs to happen:

- a class with a companion is loaded by the IDE, but the class name is *different than the file name*. This is from source
- the same class and companion object exist as binary, and are loaded from *classfiles* when the package is completed (since they have different names than the source file, the classpath abstraction will only "know" that there is a classfile, and no corresponding source file)

It seems that `companionClass` always prefers to return the companion defined in a source file, but if this assertion is called from the code path that tries to load the binary version, the newly created module will not match.

The tricky part to write a test for this is the timing (the source one has to be loaded before), and the fact that there must be some binaries as well. It should be a presentation compiler test as well... If you have some suggestions on how to do it, I'd give it a try, but it seems pretty tough.

adriaanm:
Great description, and yes -- sounds tricky enough to test that we can skip it. Let's make sure your explanation is preserved by including it in the commit message, with maybe a short version added to the comment in the code.

dragos:
```
```

Should I rebuild, or we assume the build was successful (I only changed the comment and commit message)?

lrytz:
/rebuild - this happens from time to time..

SethTisue:
it's https://github.com/scala/scala-dev/issues/277

lrytz (approved):

PR #5724, opened 2017-02-20
SCP-009: Support direct-dependency-only compilation

jvican:
This work is the result of a [Scala Center Advisory Board proposal](https://github.com/scalacenter/advisoryboard/blob/master/proposals/009-improve-direct-dependency-experience.md) by Twitter,
written by Stu Hood and Eugene Burmako.

This PR implements small changes in spots like `Symbols`, `SymbolTable` and
`UnPickler` to greatly improve stub error messages and allow users to know
which classpath entries have not been used, which help them to narrow their
direct dependencies until everything compiles without a warning.

<s>The new phase `analyzeDeps` emits warnings when classpath entries are not used.
It is also enabled under a flag `-Xanalyze-deps`, along the same lines to
similar functionality like unused imports.</s>
  
EDIT: The analyze deps phase has been removed from this PR and will be implemented as a compiler plugin.

This PR also adds new testing infrastructure to check the correct behaviour of
stub errors and unpickling. For more information on any of the aforementioned
changes, please check the commit messages.

A preliminary report on why stub errors happen can be [found here](https://contributors.scala-lang.org/t/scp-009-analysis-of-transitive-dependencies-and-stub-errors/508), though its primary 
focus is to give a general idea of where and when these errors happen, it's not meant to
be an exhaustive guide on the topic. It also describes the strategy that Eugene and I
prepared before hacking on the project.


jvican:
I'll push a backport for the Scala 2.11.x series soon.

adriaanm:
> I'll push a backport for the Scala 2.11.x series soon.

Lets do the backport after this has been reviewed here.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -574,6 +574,34 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+    val runsAfter = List("jvm")
+    val runsRightAfter = None
+
+    def newPhase(prev: Phase): GlobalPhase = {
+      new AnalyzeDepsPhase(prev)
+    }
+    private class AnalyzeDepsPhase(prev: Phase) extends GlobalPhase(prev) {
+      def name = phaseName
+      def apply(unit: CompilationUnit) {
+        val entries = settings.classpath.value.split(":").map(s => new File(s))
```
retronym: This is unix specific. It also bypasses logic to expand wildcards in this setting, see calls to `expandPath` in `ClasspathFactory`.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -574,6 +574,34 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+  object analyzeDeps extends {
+    val global: Global.this.type = Global.this
+  } with SubComponent {
+    val phaseName = "analyzeDeps"
+    val runsAfter = List("jvm")
+    val runsRightAfter = None
+
+    def newPhase(prev: Phase): GlobalPhase = {
+      new AnalyzeDepsPhase(prev)
+    }
```
retronym: Perhaps `if (settings.analyzeDeps.value) new AnalyzeDepsPhase(prev) else prev` to totally bypass this phase in the common case.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -574,6 +574,34 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+    private class AnalyzeDepsPhase(prev: Phase) extends GlobalPhase(prev) {
+      def name = phaseName
+      def apply(unit: CompilationUnit) {
+        val entries = settings.classpath.value.split(":").map(s => new File(s))
+        def findEntry(p: File) = entries.find(e => p.getAbsolutePath.startsWith(e.getAbsolutePath))
+        val usedEntries = loaders.completedClassfiles.flatMap({
+          case f: PlainFile => findEntry(f.file)
+          case f: ZipArchive#Entry => findEntry(f.underlyingSource.get.file)
+          case _ => None
+        }).distinct
```
retronym: This looks like is has `O(N x M)` complexity, where 'N' is the number of complete classfiles, and `M` is the number of classpath entries (which can be high for people using this feature!).

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -574,6 +574,34 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+        val unusedEntries = entries.diff(usedEntries)
+        if (settings.analyzeDeps.value)
+          unusedEntries.foreach(e => warning("unused classpath entry " + e))
```
retronym: Should sort the list before issuing the errors for stable behaviour. I'd also suggest to issue a single, multiline warning with all unused entries, to avoid the noise of repeating `unused classpath entry`.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -281,6 +281,8 @@ abstract class SymbolLoaders {
     }
   }
 
+  val completedClassfiles = scala.collection.mutable.ListBuffer[AbstractFile]()
```
retronym: This should be cleared at the end of a `Run`, either by storing it as a field of `Run`, or by registering it with `perRunCaches`.

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -1488,6 +1488,11 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
      *  ensuring that symbol is initialized (i.e. type is completed).
      */
     def info: Type = try {
+      /* Register the position from the symbol we are completing iff it comes
+       * from a source file, which means that this symbol forces unpickling.
+       * It's useful to report on stub errors originating from this symbol. */
+      if (pos.isDefined)
+        self.lastKnownPosition = pos
```
retronym: This should be skipped if the symbol is already complete (which is the common case and must avoid any overhead).

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -701,6 +705,7 @@ abstract class UnPickler {
       private val definedAtRunId = currentRunId
       private val p = phase
       protected def completeInternal(sym: Symbol) : Unit = try {
+        lazyCompletingSymbol = sym
```
retronym: This assignment looks like it might be clobbered by a nested call to `completeInternal`. That nested call might not trigger a stub error at all, but subsequent code pertaining to this `sym` would still report the nested symbol.

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -3444,8 +3450,8 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     override def rawInfo         = fail(NoType)
     override def companionSymbol = fail(NoSymbol)
   }
-  class StubClassSymbol(owner0: Symbol, name0: TypeName, val missingMessage: String) extends ClassSymbol(owner0, owner0.pos, name0) with StubSymbol
-  class StubTermSymbol(owner0: Symbol, name0: TermName, val missingMessage: String) extends TermSymbol(owner0, owner0.pos, name0) with StubSymbol
+  class StubClassSymbol(owner0: Symbol, name0: TypeName, val missingMessage: String, override val pos: Position) extends ClassSymbol(owner0, owner0.pos, name0) with StubSymbol
+  class StubTermSymbol(owner0: Symbol, name0: TermName, val missingMessage: String, override val pos: Position) extends TermSymbol(owner0, owner0.pos, name0) with StubSymbol
```
retronym: Rather than adding a override for `pos`, we could just call `setPos` in `newStubSymbol`. That seems simpler, and avoids having to worry about performance interactions with JIT by adding overloads.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -307,6 +309,7 @@ abstract class SymbolLoaders {
     protected def description = "class file "+ classfile.toString
 
     protected def doComplete(root: Symbol) {
+      completedClassfiles += classfile
```
retronym: This should be avoided unless `-Xanalze-deps` is active.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -574,6 +574,34 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+    val global: Global.this.type = Global.this
+  } with SubComponent {
+    val phaseName = "analyzeDeps"
+    val runsAfter = List("jvm")
+    val runsRightAfter = None
+
+    def newPhase(prev: Phase): GlobalPhase = {
+      new AnalyzeDepsPhase(prev)
+    }
+    private class AnalyzeDepsPhase(prev: Phase) extends GlobalPhase(prev) {
```
retronym: I suspect that this phase could be implemented without needng to modify the compiler to keep track of `completedClassfiles`.  We seem to have all the information we need with a walk of the symbol table.

Here's a prototype: https://github.com/retronym/scalac-survival-guide/blob/master/src/main/scala/guide/_23_CompletedSymbols.scala

That outputs:

https://gist.github.com/retronym/0abbd4ab969f2743ddf2ef30054bf0d4

This would mean you could distribute it as a compiler plugin, which would help it evolve into a useful/mature tool before adding it to the compiler itself, and would help people on older compiler versions to use it.

retronym (changes requested):

src/compiler/scala/tools/nsc/Global.scala
jvican: Good point 👍

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
jvican: Agreed.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
jvican: Agreed.

src/reflect/scala/reflect/internal/Symbols.scala
jvican: Agreed.

src/reflect/scala/reflect/internal/Symbols.scala
jvican: Good point, agreed.

src/compiler/scala/tools/nsc/Global.scala
xeno-by: I tried that approach in my initial prototype, and it was significantly slower than the one based on `completedClassfiles`. Do you think we can make it faster?

JieGhost:
Jorge,

First thank you so much for working on this!

My use case with error message is as following:
I compile a scala library with incomplete classpaths, thus I expect to see error messages telling me that "cannot find some symbols blabla...". When that happens, I want to get the missing classname(s) from error message so I can figure out what classpaths should I add.

A typical error message can be like this:
```
[error] missing or invalid dependency detected while loading class file 'UserTweetGraphLHSHandler.class'.
[error] Could not access type Request in package com.twitter.finagle.http,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'UserTweetGraphLHSHandler.class' was compiled against an incompatible version of com.twitter.finagle.http.
```
In the above case, the missing classname is "com.twitter.finagle.http.Request".
Thus I can easily find the missing classpaths corresponding to that.

However, in some cases, error message are like this:
```
[error] missing or invalid dependency detected while loading class file 'RTBCreativeDetails.class'.
[error] Could not access term ads in package com.twitter,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'RTBCreativeDetails.class' was compiled against an incompatible version of com.twitter.
```
In the above example, I can only get "com.twitter.ads" but it is not a classname (more like a prefix), thus it is hard to locate the missing classpath from that.

I noticed if the error message starts with "Could not access **term** XXX", then it does not give me classname, but if the error message starts with "Could not access **type** XXX", then it usually carries more info.

JieGhost:
Update:
I tried the 2.11 backport of this branch. The new error message looks nicer.
Even though "term symbol" still only has the partial classname, I am able to use the current error message to get to the source file and eventually find the missing class. So my use case is satisfied.

Thanks @jvican !

src/compiler/scala/tools/nsc/Global.scala
retronym: What sort of performance do you get with: https://github.com/retronym/classpath-shrinker ?

That only adds 10ms when compiling the standard library, which has a reasonably sized symbol table to walk through.

xeno-by: I don't remember details since my experiments took place last December. I think my problem was in traversing the entire symbol table, which is much-much bigger than just the top-level part.

retronym:
I'd ask to remove this part of the PR in favour of the compiler plugin
(scala centre could fork my poc, test and cross-publish an artifact)

We can then focus this pr on the improved stub errors

adriaanm:
We'll hold the 2.11.9 release to get the required improvements to the compiler merged and backported to 2.11. Could you share an estimated timeline of when you think we can get this done, so I can adjust the 2.11.9 release date?

jvican:
> I'd ask to remove this part of the PR in favour of the compiler plugin (scala centre could fork my poc, test and cross-publish an artifact).
  
Eugene seemed to had performance problems with your proposed approach. I'll try both strategies out in a big repo and benchmark them. I see value in leaving the current implementation in the PR if it performs faster than other alternatives.
  
> We'll hold the 2.11.9 release to get the required improvements to the compiler merged and backported to 2.11. Could you share an estimated timeline of when you think we can get this done, so I can adjust the 2.11.9 release date?
  
I'm updating this PR on Monday. I think I'll be done by Wednesday, just in case something comes up.

adriaanm:
Ok, great! Thanks for the update.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
jvican: I actually tried to reproduce this several times on tests but I did not succeed. Can you provide an example where this would happen?

johnynek:
if we merge this, we will want to backport this:
https://github.com/scala/scala/pull/5297

in 2.12 we removed an unneeded stub warning. When minimizing classpaths, you will see that warning a lot in 2.11 unless we also remove the warning. Apologies is this has already been backported.

adriaanm:
Thanks for the pointer, @johnynek. Let's get that one and potentially other required backports into the 2.11 queue next week.

jvican:
> if we merge this, we will want to backport this: #5297
  
Sure, no problem. I'll take care of this one too.

jvican:
@JieGhost @johnynek With my latest push, *positions* of error messages are better. Now, we report on the use site and not the definition site. Here's an example:
  
**Before**: https://github.com/jvican/scala/blob/stub-errors/test/files/run/StubErrorComplexInnerClass.check.
  
**After**: https://github.com/jvican/scala/blob/f0d79457ad36c428a3cab6753b510268efae92b8/test/files/run/StubErrorComplexInnerClass.check.

jvican:
I have just updated this PR. I've removed the `analyzeDeps` phase in favor of the compiler plugin that @retronym has done: https://github.com/retronym/classpath-shrinker. I will defer people wanting the "unused classpath entries" warning to it.
  
Regarding the improvements to stub error messages, I've updated this PR with a new approach that makes the previous prototype more robust and improves error positions. If you guys can have a look at this, I'll happily do a PR for 2.11.9 including the small change that @johnynek mentioned.

jvican:
@retronym @adriaanm CI is happy now. Let me know if you want this backported to 2.11.8.

ittaiz:
@jvican w.r.t. the usage of the compiler plugin what about the performance implications @xeno-by mentioned? I think performance implications are substantial here since users of build tools like Pants and Bazel will probably need this running always to prune the graph.

jvican:
@ittaiz There is no performance penalty in the compiler plugin. Walking top-level trees are enough to get all the unused classpath entries, which means that the compiler plugin will provide the ~same performance than the previous prototyped phase, but it's more elegant because there is no performance hit whenever symbols are completed.

ittaiz:
Ok, sounds good. Thanks!
Wdyt about maybe adding a flag to get the messages in a way that is more
machine accessible? That would enable building a tool which will consume
them and try to automatically add the relevant entries to the build
descriptor if possible.
Seeing as Scala has a lot of cases where transitive entries are used
automation can be key when trying to prune the graph.

johnynek:
+1 to making the messages easily parsed, since with bazel rules we may want
to have a separate option to fail the build or translate the jars into
target names so it is easier for users to remove them.

jvican:
>+1 to making the messages easily parsed, since with bazel rules we may want to have a separate option to fail the build or translate the jars into target names so it is easier for users to remove them.
  
To my best understanding, Twitter is already parsing the format I have and it's working correctly. Perhaps you can reach out to them and share some code? I know they wanted to open source it in Pants.

adriaanm:
Let's backport as soon as the reviews are green.

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -376,7 +376,9 @@ abstract class ClassfileParser {
     //   - remove eager warning
     val msg = s"Class $name not found - continuing with a stub."
     if ((!settings.isScaladoc) && (settings.verbose || settings.developer)) warning(msg)
-    NoSymbol.newStubSymbol(name.toTypeName, msg)
+    val stub = NoSymbol.newStubSymbol(name.toTypeName, msg)
+    stub.pos = NoPosition
```
retronym: Could you add comment here to justify why `lastTreeToTyper.pos` isn't appropriate, perhaps linking to a related test case?

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -193,6 +193,16 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
   private[reflect] case class SymbolKind(accurate: String, sanitized: String, abbreviation: String)
 
+  protected def newStubSymbol(owner: Symbol,
+                              name: Name,
+                              missingMessage: String,
+                              isPackage: Boolean = false): Symbol = {
```
retronym: `isPackage` is actually unused after recent refactorings, could you remove it from this method and calling methods a setup commit for this one?

jvican: Sure.

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
jvican: We can use `lastTreeToTyper.pos` here, but I haven't found any example in which it is necessary.

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -3415,6 +3425,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
   trait StubSymbol extends Symbol {
     devWarning("creating stub symbol to defer error: " + missingMessage)
 
+    def pos: Position
```
retronym: Is this line actually needed? We already inherit a concrete method `pos` from `Symbol`.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -63,6 +63,9 @@ abstract class UnPickler {
     /** A map from symbols to their associated `decls` scopes */
     private val symScopes = mutable.HashMap[Symbol, Scope]()
 
+    /** Keeps track of the symbol that is being lazily completed */
+    private var lazyCompletingSymbol: Symbol = NoSymbol
```
retronym: Is this still used, or has the introduction of `completingStack` replaced it?

src/reflect/scala/reflect/internal/Symbols.scala
jvican: No, it's  not.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
jvican: Should be removed, it's an oversight.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -696,11 +701,18 @@ abstract class UnPickler {
       new TypeError(e.msg)
     }
 
+    /** Keep track of the symbols pending to be initialized.
+      *
+      * Useful for reporting on stub errors and cyclic errors.
+      */
+    private var completingStack = List.empty[Symbol]
```
retronym, line 704: To help with `cyclic` errors, I think we should also include source file based type completers in this stack.

Looks like we already have a little bit of related infrastructure in:

https://github.com/scala/scala/blob/37a0eb758e72af1249fa90859dc8a0e32ea42077/src/reflect/scala/reflect/internal/Symbols.scala#L29-L35

... to support an experimental option, `-Yrecursion`.

I'm happy to take on this generalization as a followup, though.

jvican: I'm happy to follow up on this one in another PR.

ittaiz:
@retronym will the classpath-shrinker be maintained? it feels like the [decision](https://github.com/scalacenter/advisoryboard/blob/master/minutes/003-2016-q4.md) of the scala center was geared towards the classpath-shrinker (option 3 in the [proposal](https://github.com/scalacenter/advisoryboard/blob/master/proposals/009-improve-direct-dependency-experience.md)) but this ended up focusing on option 1.
For those of us who were counting on option 3 this is less than ideal.

jvican:
@ittaiz The Scala Center will publish and maintain that plugin. I'll keep you posted.

I'm updating this PR tomorrow. I'm sorry for the delay.

ittaiz:
Thanks! Eagerly waiting for the plugin

jvican:
@retronym This is ready for review again.

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -1051,8 +1051,11 @@ abstract class ClassfileParser {
       val sflags      = jflags.toScalaFlags
       val owner       = ownerForFlags(jflags)
       val scope       = getScope(jflags)
-      def newStub(name: Name) =
-        owner.newStubSymbol(name, s"Class file for ${entry.externalName} not found").setFlag(JAVA)
+      def newStub(name: Name) = {
+        val stub = owner.newStubSymbol(name, s"Class file for ${entry.externalName} not found")
+        stub.pos = owner.pos
```
lrytz: there's also `setPos` that allows chaining.

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -1174,7 +1177,11 @@ abstract class ClassfileParser {
         if (enclosing == clazz) entry.scope lookup name
         else lookupMemberAtTyperPhaseIfPossible(enclosing, name)
       }
-      def newStub = enclosing.newStubSymbol(name, s"Unable to locate class corresponding to inner class entry for $name in owner ${entry.outerName}")
+      def newStub = {
+        val stub = enclosing.newStubSymbol(name, s"Unable to locate class corresponding to inner class entry for $name in owner ${entry.outerName}")
+        stub.pos = enclosing.pos
```
lrytz: `setPos`

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -246,14 +246,16 @@ abstract class UnPickler {
...
-                    |missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
+                s"""|Symbol '${name.nameKind} ${owner.fullName}.$name' is missing from the classpath.
+                    |This symbol is required by '${lazyCompletingSymbol.kindString} ${lazyCompletingSymbol.fullName}'.
+                    |
+                    |Make sure that ${name.longString} is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
                     |A full rebuild may help if '$filename' was compiled against an incompatible version of ${owner.fullName}.$advice""".stripMargin
               val stubName = if (tag == EXTref) name else name.toTypeName
-              owner.newStubSymbol(stubName, missingMessage)
+              // The position of the error message is set by `newStubSymbol`
+              NoSymbol.newStubSymbol(stubName, missingMessage)
```
lrytz, line 257: curious, why did the stub's owner change here? i guess the actual owner doesn't matter.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -246,14 +246,16 @@ abstract class UnPickler {
...
               val advice = moduleAdvice(s"${owner.fullName}.$name")
+              val lazyCompletingSymbol = completingStack.headOption.getOrElse(NoSymbol)
               val missingMessage =
-                s"""|missing or invalid dependency detected while loading class file '$filename'.
-                    |Could not access ${name.longString} in ${owner.kindString} ${owner.fullName},
-                    |because it (or its dependencies) are missing. Check your build definition for
-                    |missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
+                s"""|Symbol '${name.nameKind} ${owner.fullName}.$name' is missing from the classpath.
+                    |This symbol is required by '${lazyCompletingSymbol.kindString} ${lazyCompletingSymbol.fullName}'.
+                    |
```
lrytz: detail, but i think we should drop the empty line. if there are multiple errors in the log we don't have any newlines either.

lrytz (changes requested):
Could you also update the PR description now that you "removed the `analyzeDeps` phase"?

lrytz:
Oh, and also squash the commits, this can all go into one.

ittaiz:
Maybe the PR detail should also refer back to the scala center proposal and
say which option this advances since it changed

JieGhost:
@jvican Sorry for missing your comment! I am a bit confused on use site vs definition site difference. The examples you gave is about "type stuberrors.A" missing. Shouldn't use site be the place where stuberrors.A is used?

adriaanm:
Ping @jvican, as this PR is blocking the 2.12.2 and the 2.11.9 releases, which I'd very much like to have out this month.

jvican:
@adriaanm Sorry, I'm taking care of this today.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
jvican: I think having an empty line makes the error message nicer to read. However, I'll make this change.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
jvican: It changed to make it more explicit that we don't use anything from the owner in the error message. All the information comes from auxiliary places.

jvican:
Feedback addressed and everything is squashed.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
adriaanm: Could you create an issue at scala/scala-dev to track this?

adriaanm:
Great, thanks! Let's start the backport to 2.11.9. I'll let the reviewers give their final ok on Monday.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
jvican: Yes.

jvican: Done https://github.com/scala/scala-dev/issues/346.

lrytz (approved):

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -246,14 +246,15 @@ abstract class UnPickler {
             adjust(mirrorThatLoaded(owner).missingHook(owner, name)) orElse {
               // (4) Create a stub symbol to defer hard failure a little longer.
               val advice = moduleAdvice(s"${owner.fullName}.$name")
+              val lazyCompletingSymbol = completingStack.headOption.getOrElse(NoSymbol)
               val missingMessage =
-                s"""|missing or invalid dependency detected while loading class file '$filename'.
-                    |Could not access ${name.longString} in ${owner.kindString} ${owner.fullName},
-                    |because it (or its dependencies) are missing. Check your build definition for
```
jvican, line 252: @JieGhost Note that the new line has been removed as per @lrytz's suggestion. You may be interesting in updating your parser if you were counting on that one.

jvican:
@ittaiz @JieGhost @johnynek Version 1.0 of compiler plugin is out: https://index.scala-lang.org/scalacenter/classpath-shrinker.
  
Official repository: https://github.com/scalacenter/classpath-shrinker.
  
If you find any issue, don't doubt to report it in the issue tracker. I'll be releasing versions actively if you experience them.

ittaiz:
Awesome! Thanks.
A colleague should try to utilize it in the next few weeks. Does it support
2.11 as well?

jvican:
@ittaiz Yes, it is cross-published for both 2.11.x and 2.12.x.

JieGhost:
@jvican Thanks. The error message improvement will be released with compiler right? When is the target date?

adriaanm:
Thanks everyone for the fruitful collaboration! This will go into 2.12.2, which is slated for release this week.

ittaiz:
It was mentioned this change will be back ported to 2.11, any ETA on that?

jvican:
@ittaiz The backport for 2.11.x was done several days ago and has just been merged https://github.com/scala/scala/pull/5804#issuecomment-289623809.

ittaiz:
Thanks!

adriaanm:
2.11.9 is being cut today -- release steps are tracked at https://github.com/scala/scala-dev/issues/330

stuhood:
Thank you everyone.

PR #5727, opened 2017-02-22
SI-10133 Deprecate unescaped single quote char literal

som-snytt:
Deprecate `'''` syntax for single quote. Error if `-Xsource:2.13`.

Also try to suggest using double quotes for `'string'` but that's quite limited.

Let `ReplReporter` handle indentation of error text (instead of indenting source); that also handles syntax errors.


adriaanm:
(just reducing jenkins load until this is reworked)

src/compiler/scala/tools/nsc/ast/parser/Scanners.scala
```diff
@@ -543,24 +543,40 @@ trait Scanners extends ScannersCommon {
...
             else if (isOperatorPart(ch) && (ch != '\\'))
               charLitOr(getOperatorRest)
             else if (!isAtEnd && (ch != SU && ch != CR && ch != LF || isUnicodeEscape)) {
+              val isEmptyCharLit = (ch == '\'')
               getLitChar()
               if (ch == '\'') {
-                nextChar()
-                token = CHARLIT
-                setStrVal()
+                if (isEmptyCharLit && settings.isScala213)
```
adriaanm, line 564: Sadly, the `isScala213` check is expensive (`def isScala213: Boolean = source.value >= version213`). This keeps coming up, maybe it's time to do something about this (/cc performance team @retronym, @lrytz).

som-snytt: Luckily it's only invoked on two successive single quotes.

By expensive, you mean more than a boolean? A couple of calls, instanceof, int compares?

adriaanm: Fair point. I missed that it was in a rare error condition. We're talking lots of object allocation, regex matching, other kinds of matching and tons of method calls. I guess we could start by not creating a new ScalaVersion object every time from a string.

som-snytt: I don't think it does that any more.

adriaanm: Ah, I spent my week in 2.11.x land, maybe that's why.

som-snytt: Oh, is that where you were. I hope you found someone to water the plants while you were away.

adriaanm: I prefer succulents anyway.

SethTisue:
needs rebase

som-snytt:
Letting Jenkins tell me what to do, it's an annoying rebase. Le rebase ennuyeux.

lrytz:
Needs some `.check` file updates (hopefully that's all)

som-snytt:
There was a clever change for `SessionTest` to source its transcript from a `.check` file, like pasting a transcript into REPL.

retronym (approved):

som-snytt:
@adriaanm sorry, did that a51b9865f5bdc60dccf8ee36d2a8bf8126d48adc collide with adriaanite? I lost track.

adriaanm:
Yes, I think I managed to incorporate its improvements into 2.13.0-M2

PR #5728, opened 2017-02-22
inlineToStr is not exhaustive and does not remove html tags inside HtmlTag [ci: last-only]

Philippus:
as can be seen on f.e. this page:
http://www.scala-lang.org/files/archive/api/current/scala/beans/BeanInfoSkip.html
when hovering over "BeanInfoSkip" on the right the title tag contains some unwanted text.
HtmlTag(...) and the `<strong>`, `</strong>` should not be inside the title-tag.
This PR should fix that problem.

- added missing Inline matches to inlineToStr so it is now exhaustive
- scala.xml.XML.loadString(tag).text will remove all html tags inside the HtmlTag

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/Page.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/Page.scala
@@ -7,7 +7,7 @@ package scala
 package tools.nsc.doc.html
 
 import scala.tools.nsc.doc.model._
-import scala.tools.nsc.doc.base.comment
+import scala.tools.nsc.doc.base.comment._
 import java.io.{FileOutputStream, File}
 import scala.reflect.NameTransformer
 import java.nio.channels.Channels
@@ -106,16 +106,21 @@ abstract class Page {
     case dtpl: DocTemplateEntity => dtpl.companion.isDefined
     case _ => false
   }
+}
 
-  protected def inlineToStr(inl: comment.Inline): String = inl match {
-    case comment.Chain(items) => items flatMap (inlineToStr(_)) mkString ""
-    case comment.Italic(in) => inlineToStr(in)
-    case comment.Bold(in) => inlineToStr(in)
-    case comment.Underline(in) => inlineToStr(in)
-    case comment.Monospace(in) => inlineToStr(in)
-    case comment.Text(text) => text
-    case comment.Summary(in) => inlineToStr(in)
-    case comment.EntityLink(comment.Text(text), _) => text
-    case _ => inl.toString
+object Page {
+  def inlineToStr(inl: Inline): String = inl match {
+    case Chain(items) => items flatMap (inlineToStr(_)) mkString ""
+    case Italic(in) => inlineToStr(in)
+    case Bold(in) => inlineToStr(in)
+    case Underline(in) => inlineToStr(in)
+    case Superscript(in) => inlineToStr(in)
+    case Subscript(in) => inlineToStr(in)
+    case Link(raw, title) => inlineToStr(title)
+    case Monospace(in) => inlineToStr(in)
+    case Text(text) => text
+    case Summary(in) => inlineToStr(in)
+    case HtmlTag(tag) => "<[^>]*>".r.replaceAllIn(tag, "")
+    case EntityLink(Text(text), _) => text
   }
 }
--- a/test/scaladoc/run/inlineToStr-strips-unwanted-text.scala
+++ b/test/scaladoc/run/inlineToStr-strips-unwanted-text.scala
@@ -1,0 +1,51 @@
+import scala.tools.nsc.doc.html.Page
+import scala.tools.nsc.doc.model._
+import scala.tools.partest.ScaladocModelTest
+
+object Test extends ScaladocModelTest {
+
+  override def code = """
+  /** This comment contains ^superscript^ */
+  class Foo {
+    /** This comment contains ,,subscript,, */
+    def bar = ???
+
+    /** This comment contains a link [[https://scala.epfl.ch/]] */
+    def baz = ???
+
+    /** This comment contains an <strong>html tag</strong> */
+    def qux = ???
+
+    /** This comment contains a<br> single html tag */
+    def quux = ???
+
+    /** This comment contains nested <strong>html<br> tags</strong> */
+    def quuz = ???
+  }
+  """
+  def scaladocSettings = ""
+
+  def testModel(root: Package) = {
+    import scala.tools.nsc.doc.base.comment._
+    import access._
+
+    val foo = root._class("Foo")
+    val bar = foo._method("bar")
+    val baz = foo._method("baz")
+    val qux = foo._method("qux")
+    val quux = foo._method("quux")
+    val quuz = foo._method("quuz")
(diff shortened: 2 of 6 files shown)


SethTisue:
would it be possible (and not unreasonably hard) to add test coverage...?

lrytz:
I quickly looked at existing scaladoc tests, they test on the model. The PR here doesn't change the model. Maybe @VladUreche can give a hint if it's easy to test this fix?

src/scaladoc/scala/tools/nsc/doc/html/Page.scala
```diff
@@ -107,15 +107,18 @@ abstract class Page {
...
+    case Italic(in) => inlineToStr(in)
+    case Bold(in) => inlineToStr(in)
+    case Underline(in) => inlineToStr(in)
+    case Superscript(in) => inlineToStr(in)
+    case Subscript(in) => inlineToStr(in)
+    case Link(raw, title) => inlineToStr(title)
+    case Monospace(in) => inlineToStr(in)
+    case Text(text) => text
+    case Summary(in) => inlineToStr(in)
+    case HtmlTag(tag) => scala.xml.XML.loadString(tag).text
```
lrytz: The `tag` data could be invalid xml, no? For example:

```
import scala.tools.nsc.doc.model._
import scala.tools.partest.ScaladocModelTest

object Test extends ScaladocModelTest {
  override def code = """
  /** Hi<br>there. */
  class A
  """

  def scaladocSettings = ""

  def testModel(root: Package) = {
    import access._
    root._class("A").comment foreach println
  }
}
```

prints

```
Body(List(Paragraph(Chain(List(Summary(Chain(List(Chain(List(Text(Hi), HtmlTag(<br>), Text(there))), Text(.)))))))))
```

and

```
scala> scala.xml.XML.loadString("<br>")
org.xml.sax.SAXParseException: XML document structures must start and end within the same entity.
```

Philippus:
I would like to add test cases but it will be a little annoying to do so as the method is inside a larger structure. 
The initial tests for the PR where the method `inlineToStr` is introduced are in `/scala/test/scaladoc/run/shortDescription-annotation.scala`and those tests repeat the `inlineToStr`-method inside the test.

src/scaladoc/scala/tools/nsc/doc/html/Page.scala
Philippus: Indeed, I will look into it a little later. I'd prefer to not use a regex, but maybe that's the way to go.

SethTisue:
(in my experience, retroactive application of `[ci: last-only]` doesn't work as a way to make the "combined" red X disappear, you just have to `/nothingtoseehere` it)

lrytz:
ok, that "/synch" didn't help, the "combined" status is still red. anyway, it would be a good idea to squash all these commits into one and `push -f`.

test/scaladoc/run/inlineToStr-strips-unwanted-text.scala
```diff
@@ -0,0 +1,59 @@
...
+      case Underline(in) => inlineToStr(in)
+      case Superscript(in) => inlineToStr(in)
+      case Subscript(in) => inlineToStr(in)
+      case Link(raw, title) => inlineToStr(title)
+      case Monospace(in) => inlineToStr(in)
+      case Text(text) => text
+      case Summary(in) => inlineToStr(in)
+      case HtmlTag(tag) => "<[^>]*>".r.replaceAllIn(tag, "")
+      case EntityLink(Text(text), _) => text
+    }
```
lrytz: Copy-pasting is not ideal, obviously.. Given that this function doesn't depend on the class `Page` in which it is defined, how about pulling it out into the companion object? I don't see a problem with that function being public (vs. protected).

lrytz (changes requested):

test/scaladoc/run/inlineToStr-strips-unwanted-text.scala
```diff
@@ -0,0 +1,59 @@
...
+  override def code = """
+  /** This comment contains ^superscript^ */
+  class Foo {
+    /** This comment contains ,,subscript,, */
+    def bar = ???
+
+    /** This comment contains a link [[https://scala.epfl.ch/]] */
+    def baz = ???
+
+    /** This comment contains an <strong>html tag</strong> */
```
lrytz, line 16: could alo add a test with nested html tags

test/scaladoc/run/inlineToStr-strips-unwanted-text.scala
Philippus: ok!

test/scaladoc/run/inlineToStr-strips-unwanted-text.scala
Philippus: ok!

Philippus:
after the tests complete, I will try squashing the commits.

Philippus:
@lrytz when you have time, could you take another look?

lrytz (approved):

lrytz:
Thanks a lot!

SethTisue:
seen in the Scala community build at https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/1301/consoleFull:

```
[scalatest] scala.MatchError: scala.tools.nsc.doc.base.MemberLookupBase$$anon$1@1ee38638 (of class scala.tools.nsc.doc.base.MemberLookupBase$$anon$1)
[scalatest] 	at scala.tools.nsc.doc.html.Page$.inlineToStr(Page.scala:124)
[scalatest] 	at scala.tools.nsc.doc.html.Page$.$anonfun$inlineToStr$1(Page.scala:113)
[scalatest] 	at scala.tools.nsc.doc.html.Page$.$anonfun$inlineToStr$1$adapted(Page.scala:113)
[scalatest] 	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)
[scalatest] 	at scala.collection.immutable.List.foreach(List.scala:390)
[scalatest] 	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:241)
[scalatest] 	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:238)
[scalatest] 	at scala.collection.immutable.List.flatMap(List.scala:353)
```

looks like a regression as a result of this PR?

lrytz:
Indeed - the pattern here is too narrow: https://github.com/scala/scala/pull/5728/files#diff-53f51aedd9537e3ba6f849e248398165R124, should probably be

```
    case EntityLink(title, _) => inlineToStr(title)
```

@Philippus do you have time to take a look?

Philippus:
yes I have time. I've managed to reproduce it with
`/** This comment contains a [[link ,,with a subscript title,,]] */`
shall I commit a fix (and test) to this branch?

SethTisue:
> shall I commit a fix (and test) 

yes please

> to this branch?

hmm... well, we'll need a new PR. I guess you could reuse the same branch? I don't think it matters

you're in good company, the community build has been catching a lot of regressions lately.

Philippus:
ok, I'll make a new PR.

PR #5730, opened 2017-02-23
Allow custom `apply` and `unapply` methods in case class companions

adriaanm:
Don't emit a synthetic `apply` (or `unapply`) when it would
clash with an existing one. This allows e.g., a `private apply`,
along with a `case class` with a `private` constructor.

We have to retract the synthetic method in a pretty roundabout way,
as we need the other methods and the owner to be completed already.
Unless we have to complete the synthetic `apply` while completing
the user-defined one, this should not be a problem. If this does
happen, this implies there's a cycle in computing the user-defined
signature and the synthetic one, which is not allowed.

Community build: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/541/console

### DIFF
--- a/spec/05-classes-and-objects.md
+++ b/spec/05-classes-and-objects.md
@@ -873,10 +873,9 @@ If a type parameter section is missing in the class, it is also missing in the `
 
 If the companion object $c$ is already defined,
 the  `apply` and `unapply` methods are added to the existing object.
+If the object $c$ already has a [matching](#definition-matching)
+`apply` (or `unapply`) member, no new definition is added.
 The definition of `apply` is omitted if class $c$ is `abstract`.
-If the object $c$ already defines a [matching](#definition-matching) member of the
-same name as the synthetic member to be added, the synthetic member
-is not added (overloading or mutual recursion is allowed, however).
 
 If the case class definition contains an empty value parameter list, the
 `unapply` method returns a `Boolean` instead of an `Option` type and
--- a/spec/05-classes-and-objects.md
+++ b/spec/05-classes-and-objects.md
@@ -851,9 +851,8 @@ already a `val` or `var` modifier. Hence, an accessor
 definition for the parameter is [generated](#class-definitions).
 
 A case class definition of `$c$[$\mathit{tps}\,$]($\mathit{ps}_1\,$)$\ldots$($\mathit{ps}_n$)` with type
-parameters $\mathit{tps}$ and value parameters $\mathit{ps}$ implicitly
-generates an [extractor object](08-pattern-matching.html#extractor-patterns) which is
-defined as follows:
+parameters $\mathit{tps}$ and value parameters $\mathit{ps}$ implies
+the definition of a companion object, which serves as an [extractor object](08-pattern-matching.html#extractor-patterns). It has the following shape:
 
 ```scala
 object $c$ {
@@ -870,11 +869,14 @@ each $\mathit{xs}\_i$ denotes the parameter names of the parameter
 section $\mathit{ps}\_i$, and
 $\mathit{xs}\_{11}, \ldots , \mathit{xs}\_{1k}$ denote the names of all parameters
 in the first parameter section $\mathit{xs}\_1$.
-If a type parameter section is missing in the
-class, it is also missing in the `apply` and
-`unapply` methods.
-The definition of `apply` is omitted if class $c$ is
-`abstract`.
+If a type parameter section is missing in the class, it is also missing in the `apply` and `unapply` methods.
+
+If the companion object $c$ is already defined,
+the  `apply` and `unapply` methods are added to the existing object.
+The definition of `apply` is omitted if class $c$ is `abstract`.
+If the object $c$ already defines a [matching](#definition-matching) member of the
+same name as the synthetic member to be added, the synthetic member
+is not added (overloading or mutual recursion is allowed, however).
 
 If the case class definition contains an empty value parameter list, the
 `unapply` method returns a `Boolean` instead of an `Option` type and
@@ -887,9 +889,6 @@ def unapply[$\mathit{tps}\,$]($x$: $c$[$\mathit{tps}\,$]) = x ne null
 The name of the `unapply` method is changed to `unapplySeq` if the first
 parameter section $\mathit{ps}_1$ of $c$ ends in a
 [repeated parameter](04-basic-declarations-and-definitions.html#repeated-parameters).
-If a companion object $c$ exists already, no new object is created,
-but the `apply` and `unapply` methods are added to the existing
-object instead.
 
 A method named `copy` is implicitly added to every case class unless the
 class already has a member (directly defined or inherited) with that name, or the
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -647,25 +647,41 @@ trait Namers extends MethodSynthesis {
     def applyUnapplyMethodCompleter(un_applyDef: DefDef, companionContext: Context): TypeCompleter =
       new CompleterWrapper(completerOf(un_applyDef)) {
         override def complete(sym: Symbol): Unit = {
+          assert(sym hasAllFlags CASE | SYNTHETIC, sym.defString)
+
           super.complete(sym)
 
-          // If there's a same-named locked symbol, we're currently completing its signature.
-          // This means it (may) refer to us, and is thus either overloaded or recursive without a signature.
-          // rule out locked symbols from the owner.info.member call
-          val scopePartiallyCompleted =
-            companionContext.scope.lookupAll(sym.name).exists(existing => existing != sym && existing.hasFlag(LOCKED))
+          // owner won't be locked
+          val ownerInfo = companionContext.owner.info
 
-          val suppress =
-            scopePartiallyCompleted || {
-            val userDefined = companionContext.owner.info.member(sym.name).filter(_ != sym)
+          // If there's a same-named locked symbol, we're currently completing its signature.
+          // If `scopePartiallyCompleted`, the program is known to have a type error, since
+          // this means a user-defined method is missing a result type while its rhs refers to `sym` or an overload.
+          // This is an error because overloaded/recursive methods must have a result type.
+          // The method would be overloaded if its signature, once completed, would not match the synthetic method's,
+          // or recursive if it turned out we should unlink our synthetic method (matching sig).
+          // In any case, error out. We don't unlink the symbol so that `symWasOverloaded` says yes,
+          // which would be wrong if the method is in fact recursive, but it seems less confusing.
+          val scopePartiallyCompleted = new HasMember(ownerInfo, sym.name, BridgeFlags | SYNTHETIC, LOCKED).apply()
+
+          // Check `scopePartiallyCompleted` first to rule out locked symbols from the owner.info.member call,
+          // as FindMember will call info on a locked symbol (while checking type matching to assemble an overloaded type),
+          // and throw a TypeError, so that we are aborted.
+          // Do not consider deferred symbols, as suppressing our concrete implementation would be an error regardless
+          // of whether the signature matches (if it matches, we omitted a valid implementation, if it doesn't,
+          // we would get an error for the missing implementation it isn't implemented by some overload other than our synthetic one)
+          val suppress = scopePartiallyCompleted || {
+            // can't exclude deferred members using DEFERRED flag here (TODO: why?)
+            val userDefined = ownerInfo.memberBasedOnName(sym.name, BridgeFlags | SYNTHETIC)
+
             (userDefined != NoSymbol) && {
-              userDefined.info match {
-                // TODO: do we have something for this already? the synthetic symbol can't be overloaded, right?
-                case OverloadedType(pre, alternatives) =>
-                  // pre probably relevant because of inherited overloads?
-                  alternatives.exists(_.isErroneous) || alternatives.exists(alt => pre.memberInfo(alt) matches pre.memberInfo(sym))
-                case tp =>
-                  (tp eq ErrorType) || tp.matches(sym.info)
+              assert(userDefined != sym)
+              val alts = userDefined.alternatives // could be just the one, if this member isn't overloaded
+              // don't compute any further `memberInfo`s if there's an error somewhere
+              alts.exists(_.isErroneous) || {
+                val self = companionContext.owner.thisType
+                val memberInfo = self.memberInfo(sym)
+                alts.exists(alt => !alt.isDeferred && (self.memberInfo(alt) matches memberInfo))
               }
             }
           }
@@ -748,8 +764,6 @@ trait Namers extends MethodSynthesis {
         val bridgeFlag = if (mods hasAnnotationNamed tpnme.bridgeAnnot) BRIDGE | ARTIFACT else 0
         val sym = assignAndEnterSymbol(tree) setFlag bridgeFlag
 
-        // copy/apply/unapply synthetics are added using the addIfMissing mechanism,
-        // which ensures the owner has its preliminary info (we may add another decl here)
         val completer =
           if (sym hasFlag SYNTHETIC) {
             if (name == nme.copy) copyMethodCompleter(tree)
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -610,7 +610,15 @@ trait Namers extends MethodSynthesis {
       noDuplicates(selectors map (_.rename), AppearsTwice)
     }
 
-    def enterCopyMethod(copyDef: DefDef): Symbol = {
+    class CompleterWrapper(completer: TypeCompleter) extends TypeCompleter {
+      val tree = completer.tree
+
+      override def complete(sym: Symbol): Unit = {
+        completer.complete(sym)
+      }
+    }
+
+    def copyMethodCompleter(copyDef: DefDef): TypeCompleter = {
       val sym      = copyDef.symbol
       val lazyType = completerOf(copyDef)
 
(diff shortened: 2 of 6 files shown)


adriaanm:
Community build with forked paradise: https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-community-build/535/console. Ok except for play-twirl, which seems to fail for an unrelated reason (macro compiled with older version of scala?? /cc @SethTisue, does this sound familiar?)

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -946,48 +978,63 @@ trait Namers extends MethodSynthesis {
...
+      // (the case class definitions is identified by `clazz`'s `ClassForCaseCompanionAttachment`),
+      // enter the following unless a matching method exists:
+      //
+      //  1. if case class is not abstract, add
+      //   <synthetic> <case> def apply[Ts](ps: Us): C[Ts] = new C[Ts](ps)
+      //  2. add a method
+      //   <synthetic> <case> def unapply[Ts](x: C[Ts]) = <ret-val>
+      //  where <ret-val> is the caseClassUnapplyReturnValue of class C (see UnApplies.scala)
+      //
+      // This last condition requires special care, as we need to detect overloading without forcing too
```
adriaanm: TODO: finish my sentence

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -463,31 +434,35 @@ trait Namers extends MethodSynthesis {
...
-        tree.symbol = m
+        updatePosFlags(existingModule, tree.pos, moduleFlags)
+        setPrivateWithin(tree, existingModule)
+        existingModule.moduleClass andAlso (setPrivateWithin(tree, _))
+        context.unit.synthetics -= existingModule
+        tree.symbol = existingModule
       }
       else {
-        m = assignAndEnterSymbol(tree)
+        enterInScope(assignMemberSymbol(tree))
```
lrytz: This is another candidate for "enterInScope may return existing symbol? In interactive setting?"

If this happens (`enterInScope` returns a different symbol), it looks like `tree.symbol` no longer holds the correct symbol. Not sure if this causes issues. In any case, this commit doesn't make things worse.

test/files/neg/userdefined_apply.scala
```diff
@@ -0,0 +1,33 @@
...
+case class ClashNoSig private (x: Int)
+
+// needs sig because overloaded -- TODO: actual error is double def
+object ClashOverloadNoSig {
+  private def apply(x: Int) = if (x > 0) new ClashOverloadNoSig(x) else apply("")
+  def apply(x: String): ClashOverloadNoSig = ???
+}
+case class ClashOverloadNoSig private (x: Int)
+
+// needs sig because it's recursive
```
lrytz: error message would ideally also be more clear in this case ("recursive method needs result type"), but the cure is the same.

lrytz (commented):
Looks very good!

There's an empty commit (676ec1618a7e99c1874ba8cbb6b09fd0acaf1c77).

adriaanm:
I'll see about moving some of the more ambitious changes over to 2.12.2. Is there anything before 		09c35ae (incl) that you're concerned about?

adriaanm:
On second thought, I'll just make the minimal changes here and move the cleanups to 2.12

lrytz:
It LGTM, but agree it's a good idea to make minimal changes in 2.11.

adriaanm:
Ready for final 👀 .

adriaanm:
More ambitious wip for 2.12 moved to https://github.com/scala/scala/compare/2.12.x...adriaanm:userdefined-apply-212?expand=1

SethTisue:
Does this need a spec update?

adriaanm:
True:
  - [x] If a companion object cc exists already, no new object is created, but the apply and unapply methods are added to the existing object instead, unless the object already has a [matching](#definition-matching) definition.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -629,14 +637,57 @@ trait Namers extends MethodSynthesis {
...
 
+    // for apply/unapply, which may need to disappear when they clash with a user-defined method of matching signature
+    def applyUnapplyMethodCompleter(un_applyDef: DefDef, companionContext: Context): TypeCompleter =
+      new CompleterWrapper(completerOf(un_applyDef)) {
+        override def complete(sym: Symbol): Unit = {
+          super.complete(sym)
+
+          // If there's a same-named locked symbol, we're currently completing its signature.
+          // This means it (may) refer to us, and is thus either overloaded or recursive without a signature.
+          // rule out locked symbols from the owner.info.member call
```
lrytz: Tricky stuff.. so if there's a locked sym in the same scope, the current sym is marked error. Does that cause the error to be reported? Why do we want to avoid the `owner.info.member` member call in this case?

Could there be a similar situation when there's a locked symbol not in the same scope, but inherited?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -629,14 +637,57 @@ trait Namers extends MethodSynthesis {
...
+                  (tp eq ErrorType) || tp.matches(sym.info)
+              }
+            }
+          }
+
+          if (suppress) {
+            sym setInfo ErrorType
+            sym setFlag IS_ERROR
+
+            // Don't unlink in an error situation to generate less confusing error messages.
```
lrytz, line 693: It's not obvious to me why/how this impacts error reporting..

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -697,11 +748,17 @@ trait Namers extends MethodSynthesis {
         val bridgeFlag = if (mods hasAnnotationNamed tpnme.bridgeAnnot) BRIDGE | ARTIFACT else 0
         val sym = assignAndEnterSymbol(tree) setFlag bridgeFlag
 
-        if (name == nme.copy && sym.isSynthetic)
-          enterCopyMethod(tree)
-        else
-          sym setInfo completerOf(tree)
-    }
+        // copy/apply/unapply synthetics are added using the addIfMissing mechanism,
```
lrytz: `addIfMissing` is no longer in this PR

test/files/neg/userdefined_apply.scala
```diff
@@ -0,0 +1,31 @@
...
+  // error: overloaded method apply needs result type
+  private def apply(x: Int) = if (x > 0) new ClashOverloadNoSig(x) else apply("")
+
+  def apply(x: String): ClashOverloadNoSig = ???
+}
+
+case class ClashOverloadNoSig private(x: Int)
+
+object ClashRecNoSig {
+  // error: recursive method apply needs result type
```
lrytz: The actual report says `overloaded method apply needs result type`, as you explain in the lengthy comment in Namers. So the behavior is fine, just the comment here is out of synch.

lrytz (commented):
Looks good, just a few clarifications.

fommil:
I really like this, it's stung me a couple of times in the past.

fommil:
btw, somewhat related, a GSoC proposal I'm putting together for this year: https://github.com/scala/scala-lang/blob/master/gsoc/2017.md#case-classes-a-la-carte-with-scalameta

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: `member` will need the `info` of the symbols its collecting in an `OverloadedType` (to see if they match). Calling `info` on those locked symbols will yield a `TypeError` to indicate the cycle, causing our stack to be unwound.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: Unlinking when there actually isn't a clash would result in the errors talking about recursive methods (since `symWasOverloaded` is `false`) instead of overloaded methods, flipping the wrongness of the error message to the other side.
```
    private def symWasOverloaded(sym: Symbol) = sym.owner.isClass && sym.owner.info.member(sym.name).isOverloaded
```

adriaanm: If we could be more precise in determining whether there is a clash (by partially completing the method info in methodSig with a wildcard for the result type when it's inferred), we could make the error messages more accurate. Looking into that for 2.12

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: good catch

test/files/neg/userdefined_apply.scala
adriaanm: fixed

spec/05-classes-and-objects.md
```diff
@@ -870,11 +869,14 @@ each $\mathit{xs}\_i$ denotes the parameter names of the parameter
...
-The definition of `apply` is omitted if class $c$ is
-`abstract`.
+If a type parameter section is missing in the class, it is also missing in the `apply` and `unapply` methods.
+
+If the companion object $c$ is already defined,
+the  `apply` and `unapply` methods are added to the existing object.
+The definition of `apply` is omitted if class $c$ is `abstract`.
+If the object $c$ already defines a [matching](#definition-matching) member of the
+same name as the synthetic member to be added, the synthetic member
+is not added (overloading or mutual recursion is allowed, however).
```
retronym: "defines or inherits" ?

"mutual recursion" sounds to me like the synthetic apply might call the user-defined apply, whereas it only calls the constructor. Can we tighten up this description, or maybe just avoid speccing the fact that we allow the user-defined apply to call the synthetic one.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -629,13 +637,58 @@ trait Namers extends MethodSynthesis {
...
+            (userDefined != NoSymbol) && {
+              userDefined.info match {
+                // TODO: do we have something for this already? the synthetic symbol can't be overloaded, right?
+                case OverloadedType(pre, alternatives) =>
+                  // pre probably relevant because of inherited overloads?
+                  alternatives.exists(_.isErroneous) || alternatives.exists(alt => pre.memberInfo(alt) matches pre.memberInfo(sym))
+                case tp =>
+                  (tp eq ErrorType) || tp.matches(sym.info)
+              }
+            }
```
retronym: Perhaps something along these lines might read more cleanly?

```
scala> case class C(a: Int); object C { def apply(a: Int, b: Int) = 42; object apply }
defined class C
defined object C

scala> val synthApply = typeOf[C.type].decl(nme.apply).alternatives.find(_.isSynthetic).get // just `sym` in code above
synthApply: $r.intp.global.Symbol = method apply

scala> val userDefinedApplies = typeOf[C.type].memberBasedOnName(nme.apply, Flags.BridgeFlags | Flags.SYNTHETIC).alternatives
userDefinedApplies: List[$r.intp.global.Symbol] = List(object apply, method apply)                                                                       

scala> userDefinedApplies.map(x => (x, typeOf[C.type].memberInfo(x), typeOf[C.type].memberInfo(x) matches synthApply.info)).mkString("\n")
res20: String =
(object apply,C.apply.type,false)
(method apply,(a: Int, b: Int)Int,false)
```

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -629,13 +637,58 @@ trait Namers extends MethodSynthesis {
...
+    // for apply/unapply, which may need to disappear when they clash with a user-defined method of matching signature
+    def applyUnapplyMethodCompleter(un_applyDef: DefDef, companionContext: Context): TypeCompleter =
+      new CompleterWrapper(completerOf(un_applyDef)) {
+        override def complete(sym: Symbol): Unit = {
+          super.complete(sym)
+
+          // If there's a same-named locked symbol, we're currently completing its signature.
+          // This means it (may) refer to us, and is thus either overloaded or recursive without a signature.
+          val scopePartiallyCompleted =
+            companionContext.scope.lookupAll(sym.name).exists(existing => existing != sym && existing.hasFlag(LOCKED))
```
retronym: This won't find cycles involving inherited `apply`-s.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
retronym: As we vaguely discussed, it would be create to have less opaque type completers for method symbols that let us find out the parameter list arity first (which could filter out some members already), then the method type with a wildcard result to represent "not yet inferred", and then finally the full method type. It would still be a bit of a pain to use, as we couldn't just reuse `findMember` as it, we'd need to generalise it to allow us to specify a signature schema to narrow the search.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -629,13 +637,58 @@ trait Namers extends MethodSynthesis {
...
+
+          val suppress =
+            // Check `scopePartiallyCompleted` first to rule out locked symbols from the owner.info.member call,
+            // as FindMember will call info on a locked symbol (while checking type matching to assemble an overloaded type),
+            // and throw a TypeError, so that we are aborted.
+            scopePartiallyCompleted || {
+            val userDefined = companionContext.owner.info.member(sym.name).filter(_ != sym)
```
retronym: Is there a chance that `companionContext.owner` is locked at this point?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -629,13 +637,58 @@ trait Namers extends MethodSynthesis {
...
+            (userDefined != NoSymbol) && {
+              userDefined.info match {
+                // TODO: do we have something for this already? the synthetic symbol can't be overloaded, right?
+                case OverloadedType(pre, alternatives) =>
+                  // pre probably relevant because of inherited overloads?
+                  alternatives.exists(_.isErroneous) || alternatives.exists(alt => pre.memberInfo(alt) matches pre.memberInfo(sym))
+                case tp =>
+                  (tp eq ErrorType) || tp.matches(sym.info)
```
retronym: This line neglects the `memberInfo` on the (potentially inherited) user defined member.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
retronym: Maybe not. I was thinking about bugs like:

https://issues.scala-lang.org/browse/SI-6357

But here, `companionContext.owner` is the just the module class symbol, not its enclosing owner.

spec/05-classes-and-objects.md
adriaanm: fixed

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: Good point! Added a `HasMember` subclass of `FindMemberBase` for a more complete check (which does not force any infos). Also added test.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: ASF strikes again. Fixed, and test added.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: I played with various approaches to partial sigs. Would be really nice, and perhaps feasible in 2.12, but leaving out of scope here.

Could you elaborate on your first comment? Do you intend to avoid the `.filter` by excluding `SYNTHETIC`?

src/reflect/scala/reflect/internal/tpe/FindMembers.scala
```diff
@@ -285,4 +285,21 @@ trait FindMembers {
       initBaseClasses.head.newOverloaded(tpe, members)
     }
   }
+
+  private[scala] final class HasMember(tpe: Type, name: Name, excludedFlags: Long, requiredFlags: Long, stableOnly: Boolean)
```
lrytz: `stableOnly` is not used

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -649,23 +649,34 @@ trait Namers extends MethodSynthesis {
...
           // If there's a same-named locked symbol, we're currently completing its signature.
-          // This means it (may) refer to us, and is thus either overloaded or recursive without a signature.
-          // rule out locked symbols from the owner.info.member call
+          // If `scopePartiallyCompleted`, the program is known to have a type error, since
+          // this means a user-defined method is missing a result type while its rhs refers to `sym` or an overload.
+          // This is an error because overloaded/recursive methods must have a result type.
+          // The method would be overloaded if its signature, once completed, would not match the synthetic method's,
+          // or recursive if it turned out we should unlink our synthetic method (matching sig).
+          // In any case, error out. We don't unlink the symbol so that `symWasOverloaded` says yes,
+          // which would be wrong if the method is in fact recursive, but it seems less confusing.
```
lrytz, line 664: thanks, this much more clear now 👍

src/compiler/scala/tools/nsc/typechecker/Namers.scala
retronym: If we don't exclude the `SYNTHETIC` member in from `findMember`, I don't think we'll notice if the object would have inherited a matching member. (We'd need to separately check the override chain of the synthetic one.) Excluding it from `findMember` seems to kill two birds with one stone.

src/reflect/scala/reflect/internal/tpe/FindMembers.scala
```diff
@@ -285,4 +285,21 @@ trait FindMembers {
...
+    extends FindMemberBase[Symbol](tpe, name, excludedFlags, requiredFlags) {
+    private[this] var _result: Symbol = NoSymbol
+
+    override protected def result: Symbol = _result
+
+    protected def shortCircuit(sym: Symbol): Boolean = {
+      _result = sym
+      true // prevents call to addMemberIfNew
```
retronym: AFAICT this will only find the first member with `name`, which doesn't seem like what we want: we need to consider all members.

It would be possible to change `addMemberIfNew` to just to treat everything as overrides (which would basically devolve to `cls.baseClasses.flatMap(_.decl(name)).filter(x => !x.isPrivate || x.owner == cls)`. That might avoids the `lo matches hi` checks needs to distinguish overrides from overloads.

We still need to force the infos of each member to see if they matches the synthetic method type, so I'm not sure whether it is any lazier on type completion.

adriaanm: In this case, all we care about finding any member of said name with the locked flag.
I guess we should return a boolean to avoid confusion.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: Ah yes. Hmm, I wonder how my pos test didn't catch this.

retronym:
Maybe it flukes it with a fortuitous order of members?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: Also had to exclude DEFERRED, since otherwise we suppress our synthetic implementation of the apply inherited from FunctionN (as it's a matching member, albeit not a concrete one)

adriaanm:
Well then, looks like this should be the one 🤞

retronym:
When I reviewed the last iteration of this change, I mistakenly thought that the custom `FindMember` implementation replaced the entire call to `.members`, rather than just the search for the existence of a locked decl. That's why I thought it was a mistake to short circuit.

This is the code I'd imagined at the time that joins these two parts into a single member walk:

https://github.com/retronym/scala/tree/review/5730

It also avoids completing infos of members over trees that don't match the arity.

That said, I'm happy enough with this change as it is. Feel free to take some ideas from my branch if you further rework this on the 2.12.x branch, but I'm giving this the 👍 here.

retronym (approved):

adriaanm:
Thanks! For the next refinement, I'd prefer to enrich the type completer interface so that you can ask more directly for partial completions (e.g., just the types you can determine without looking at RHS, just the type params, ...)

retronym:
Sounds good. e.g. we could add a method like `schematicType` that gives us the shape of the method type with wildcards. To be useful, we've have to change a few other places to use them, notable the checks for overriden members in namers, such as:

```
      val overridden = {
        val isConstr   = meth.isConstructor
        if (isConstr || !methOwner.isClass) NoSymbol else overriddenSymbol(methResTp)
      }
      val hasDefaults = mexists(vparamss)(_.symbol.hasDefault) || mexists(overridden.paramss)(_.hasDefault)
      if (hasDefaults)
        addDefaultGetters(meth, ddef, vparamss, tparams, overridden)
```

adriaanm:
Right, having a way to trigger cycle-free completion (peeling off just one
layer) may also enable experimenting with more pro-active completion
schemes so that we can populate the symbol table in a tighter loop, without
jumping back and forth and going into deep lazy completer call chains.

PR #5733, opened 2017-02-23
Get commit SHA and date with JGit

szeiger:
In order to get the SHA and date we used to run shell scripts and parse
the output of `git` commands. On Windows we even ran a batch file that
looked for `bash.exe` and then ran the shell script in bash.

Using JGit should be more robust than the old Rube Goldberg
implementation. The values produced are the same, except for the time
zone. Previously the timestamp was formatted with the local timezone,
now we use UTC.


tools/get-scala-commit-sha
```diff
@@ -1,22 +0,0 @@
-#!/usr/bin/env bash
```
janekdb: I am guessing {{_site/build.xml}} is no longer important. Just noting it has a reference to this command,

```
    <exec osfamily="unix" executable="tools/get-scala-commit-sha" outputproperty="git.commit.sha" failifexecutionfails="false" />
    <exec osfamily="windows" executable="cmd.exe" outputproperty="git.commit.sha" failifexecutionfails="false">
      <arg value="/c"/>
      <arg value="tools\get-scala-commit-sha.bat"/>
      <arg value="-p"/>
    </exec>
    <exec osfamily="unix" executable="tools/get-scala-commit-date" outputproperty="git.commit.date" failifexecutionfails="false" />
```

mkeskells:
I cant verify this on my windows install as I cant access szeiger:wip/use-jgit. Is szeiger:wip private?

SethTisue:
@mkeskells the branch name is `wip/use-jgit` which has a slash character in it — perhaps that's confusing some tool you're using?

SethTisue:
@mkeskells the relevant commit is https://github.com/szeiger/scala/commit/acd69677684ea722981503c694951a28ce4f9502

mkeskells:
The tool wasnt confused - just me - I was looking for a repo called wip :-(
verified that it works for me, and that it works with a sensible heap size of 2Gb

SethTisue:
@szeiger rebase?

PR #5734, opened 2017-02-23
WIP SI-10081 avoid divergence in implicit scope search with f-bounds, annotated types

szeiger:
Fix by @retronym. LGTM. Let's run this through PR validation and get it merged. We keep running into this bug again and again in collections-strawman.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -1064,9 +1064,10 @@ trait Implicits {
        *                 in order to cache them in infoMapCache
        */
       def getParts(tp: Type)(implicit infoMap: InfoMap, seen: mutable.Set[Type], pending: Set[Symbol]) {
-        if (seen(tp))
+        val tpWithoutAnnotations = tp.map(_.withoutAnnotations)
+        if (seen(tpWithoutAnnotations))
           return
-        seen += tp
+        seen += tpWithoutAnnotations
         tp match {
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1064,9 +1064,10 @@ trait Implicits {
        *                 in order to cache them in infoMapCache
        */
       def getParts(tp: Type)(implicit infoMap: InfoMap, seen: mutable.Set[Type], pending: Set[Symbol]) {
-        if (seen(tp))
+        val tpWithoutAnnotations = tp.map(_.withoutAnnotations)
```
retronym, line 1067: I fear that this will lead to exponential performance because at every level of recursion on the type in `getParts`, we're recursing over the entire `tp` to drop the annotations. 

We could test this theory this by counting the number of times the `_.withoutAnnotations` is called with an implicit search for `pt = Some[Some[Some[....[Any]]]]]`, and plotting vs the nesting level of `pt`.

```
⚡ (for N in 1 2 4 8 16 32 64 128 256 512 1024; do (printf 'class C { implicitly['; (for i in $(seq 1 $N); do printf 'Seq['; done ; printf Int; for i in $(seq 1 $N); do printf ']'; done); printf ']}\n') > /tmp/a.scala; qscalac -J-Xss5M /tmp/a.scala 2>&1 ; done ) | grep -i counter
```

image

adriaanm:
This will target 2.13 only, at least initially.

PR #5735, opened 2017-02-23
increase timeouts on some sys.process tests

SethTisue:
these were occasionally timing out on our Windows CI

reference: https://github.com/scala/scala-dev/issues/313


lrytz:
/rebuild - `ERROR: Failed to run DSL Script`

PR #5736, opened 2017-02-23
SI-10206 tighten fix for SI-6899

adriaanm:
There are more supertypes of `AnyRef` than you might think:
`?{def clone: ?}` is one example...

scala/bug#6899

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -134,11 +134,6 @@ trait Implicits {
   private val improvesCache = perRunCaches.newMap[(ImplicitInfo, ImplicitInfo), Boolean]()
   private val implicitSearchId = { var id = 1 ; () => try id finally id += 1 }
 
-  private def isInvalidConversionSource(tpe: Type): Boolean = tpe match {
-    case Function1(in, _) => in <:< NullClass.tpe
-    case _                => false
-  }
-
   def resetImplicits() {
     implicitsCache.clear()
     infoMapCache.clear()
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1388,27 +1383,30 @@ trait Implicits {
...
-              result = SearchFailure
-          case _                 => false
-        }
-        if (settings.isScala211 && isInvalidConversionSource(pt)) {
-          maybeInvalidConversionError("an expression of type Null is ineligible for implicit conversion")
-          result = SearchFailure
+          // SI-10206 don't use subtyping to rule out AnyRef/AnyVal: there are several valid structural types that are supertypes of AnyRef (e.g., created by HasMember);
+          // typeSymbol will do the trick (AnyRef is a type alias for Object)
+          case Function1(in, out)  =>
+            val outSym = out.typeSymbol
```
retronym: I think I like this approach better than mine. Another thing I was thinking about in was whether we might accidentally be contributing unwanted constraints to type variables in `out` in the old approach using `<:<`. Your approach also avoids that.

Noting that if `out` is itself a refinement type of `AnyRef`, `.typeSymbol` will not give us `AnyRefClass` (which is a good thing here, as we don't want to disallow that.

One corner case might be `AnnotatedTypes` with an annotation checker that treats `AnyRef @foo`  as distinct from `AnyRef` and uses implicits to model conversions between the two?

adriaanm: Good catch on the annotated types! I'll add an exclusion for those.

SethTisue:
is there any risk to targeting 2.11.x here? 2.12.x being the safer choice generally

adriaanm:
I'm comfortable with it, but I'll let Jason decide in his review. I think this is both narrow and enough of an improvement to just include here.

adriaanm:
@retronym, could you enter your final review verdict using the System

retronym (changes requested):
I'd like to see the test case from SI-10206 included, otherwise LGTM.

https://github.com/retronym/scala/blob/a58d8e89beb748ad8a2208322a0cf539a6b9c85c/test/files/pos/t10206.scala

adriaanm:
Tests added

adriaanm:
ping @retronym

retronym (approved):

som-snytt:
It was issue 6899.

SethTisue:
title fixed, ticket link added to PR description

som-snytt:
(Paulp had made the error on the mailing list, so this might have been an in-joke which I ruined.)

PR #5737, opened 2017-02-23
SI-10206 Fix implicit views for inapplicable overloads in Any-{Ref,Val}

retronym:
Implicit search implements a restruction that the target type for an
implicit view should be more specific that AnyRef or AnyVal.

  scala> def foo(s: String): AnyVal = s
  <console>:12: error: the result type of an implicit conversion must be more specific than AnyVal
         def foo(s: String): AnyVal = s
                                      ^
Without this, an implicit value classes over `String` would be applied,
which is unlikely to be what was intended.

Implicit views are implemented as an implicit search for a function type
with a structural type as its result. This structural type is created
with:

    scala> val schema = analyzer.HasMethodMatching.apply(TermName("clone"), Nil, WildcardType)
    schema: $r.intp.global.analyzer.global.Type = ?{def clone(): ?}

The quirk arises when, as above, we're seeking a member with the same name
as a member of AnyRef. AnyRef is seen to be a subtype of the result type:

   scala> AnyRefClass.tpe <:< schema
   res23: Boolean = true

Which leads to the implicit in the test case being disqualified. The typer
opts to report the error about the inapplicability of the inherited clone
method, so we don't even know why the implicit was discarded.

This commit limits the check against AnyRef/AnyVal to fully defined
result types, which exclude wildcard and abstract types.

This change is limited to -Xsource:2.13 (although that will
be available in the 2.12 series of compilers.)

I've manually tested that we're bug compatible under older -Xsource settings.
I didn't add this as a test case.

```
⚡ qscalac -Xsource:2.13 test/files/pos/t10206.scala

⚡ qscalac -Xsource:2.12 test/files/pos/t10206.scala
test/files/pos/t10206.scala:11: error: method clone in class Object cannot be accessed in Foo
 Access to protected method clone not permitted because
 prefix type Foo does not conform to
 object Main where the access take place
  println(foo.clone(1, 2))    // <- does not compile
              ^
one error found

⚡ qscalac -Xsource:2.11 test/files/pos/t10206.scala
test/files/pos/t10206.scala:11: error: method clone in class Object cannot be accessed in Foo
 Access to protected method clone not permitted because
 prefix type Foo does not conform to
 object Main where the access take place
  println(foo.clone(1, 2))    // <- does not compile
              ^
one error found
```


retronym:
Let's consider this a review for #5736

SethTisue:
somebody owes somebody a coke

adriaanm:
How about some Talisker? Funny how this one proved so irresistible :-)

PR #5739, opened 2017-02-24
Lazy ClassBTypes

lrytz:


lrytz:
This took a bit of time to investigate..

So we have two ways of building a `ClassBType`
  - from a `Symbol`: necessary for classes being compiled
  - from an `asm.ClassNode`: for classes read from the classpath; sometimes there's no symbol available, e.g. in the Inliner when needing a `ClassBType` for an internal name that was not touched before (this is one of the major fixes in the new backend, we don't need anymore the logic to find the correct symbol given an internal name, which was a weak point in 2.11)

Whenever we build a `ClassBType`, we store it in a global map `classBTypeFromInternalName`, which allows building the cyclic graph of types.

In this PR we start delaying completion of types. Before, when building a type from a symbol, all its dependncies were also built from symbols eagerly. Now that we delay that, it can happen that a type is completed from a classfile instead.

This is potentially buggy: we have to ensure that types for symbols being compiled are always built from the symbol.

The failing test checks that types are always the same, no matter if they are built from a class node or a symbol. However, there are slight (benign) differneces: for example, a class node lists all its nested classes, including anonymous. A symbol only lists its anonymous nested classes if its being compiled, not unpickled. The tests failed because a `ClassBType` suddenly was completed from a class node and got one more nested class.

I'll work on a proper fix.

lrytz:
Should be ready for review by @retronym now.

I did a simple `println(fullName)` in `Symbol.info`, here is number of forced symbols for compiling `List.scala`:
```
➜  sandbox git:(lazyBTypes) ✗ wc -l before after
    6201 before
    4783 after
```

Diff: https://gist.github.com/lrytz/8d11e12a848cf0ea228849b9fd4553ac

So it seems to help.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -1066,9 +1068,46 @@ abstract class BTypes {
...
+        function = () => {
+          prev()
+          f(value)
+        }
+      }
+    }
+
+    def force: T = {
+      if (value != null) value
+      else bTypes.synchronized {
```
retronym: `btypes` was a placeholder for "thing we should use as the global lock" in lazy parts of `BTypes`, which might get called in bytecode generation, which we might try to parallelize later. Is there something more appropriate to lock on?

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -1066,9 +1068,46 @@ abstract class BTypes {
...
+    private var function = {
+      val tt = t // prevent allocating a field for t
+      () => { value = tt() }
+    }
+
+    override def toString = if (value == null) "<?>" else value.toString
+
+    def onForce(f: T => Unit): Unit = {
+      if (value != null) f(value)
+      else {
```
retronym: Should probably `synchronize` this on the same global lock as below.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -847,7 +849,7 @@ abstract class BTypes {
    * a missing info. In order not to crash the compiler unnecessarily, the inliner does not force
    * infos using `get`, but it reports inliner warnings for missing infos that prevent inlining.
    */
-  final case class ClassBType(internalName: InternalName) extends RefBType {
+  final case class ClassBType(internalName: InternalName)(cache: mutable.Map[InternalName, ClassBType]) extends RefBType {
     /**
      * Write-once variable allows initializing a cyclic graph of infos. This is required for
      * nested classes. Example: for the definition `class A { class B }` we have
```
retronym, line 860: Does the introduction of laziness in `nestedClasses` / `innerClasses` allow us to simplify this aspect of the implementation?

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
lrytz: `bTypes` sounds reasonable, there's a single instance per Global.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
lrytz: Good point - I'd keep it for a follow up, but it sounds like a useful experiment.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
retronym: Is this the only spot that uses a lock rather than a concurrent data structure?

lrytz: There's https://github.com/scala/scala/blob/v2.12.1/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala#L81-L83, I couldn't find any others. I can add a lock object for the `Lazy` class somewhere.

retronym: Well spotted! We'll need to use the same lock for both, as they wrap unknown callback code. This is similar the use of `gilSynchronized` in runtime reflection.

lrytz: Right.. thanks!

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -64,17 +69,19 @@ abstract class BTypes {
...
-   * on multiple classes concurrently.
+   * name. The method assumes that every class type that appears in the bytecode exists in the map
    */
-  val classBTypeFromInternalName: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(TrieMap.empty)
+  def cachedClassBType(internalName: InternalName): Option[ClassBType] =
+    classBTypeCacheFromSymbol.get(internalName).orElse(classBTypeCacheFromClassfile.get(internalName))
+
+  // Concurrent maps because stack map frames are computed when in the class writer, which
+  // might run on multiple classes concurrently.
+  val classBTypeCacheFromSymbol: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(TrieMap.empty)
```
retronym, line 83: As a followup, we should see if we could key this map on `Symbol`, rather than `String`. That will avoid needing to compute `javaBinaryName` when successfully looking up values in this map. In #5784, I cached `Symbol#javaBinaryName` to workaround this inefficiency, but we could compare that approach with a refactoring here.

lrytz: OK, maybe that's best looked at in the context of https://github.com/scala/scala/pull/5815, which works on other parts of backend performance?

retronym:
I'm seeing a positive effect on some of the benchmarks. Using #5907 as the baseline:

better-files: 308->301
vector: 177->168

retronym (approved):

PR #5740, opened 2017-02-24
Optimize VectorBuilder#++=(xs: TraversableOnce)

monkey-mas:
We can improve the performance by using `xs.foreach(+=(_))` instead of using `super.++=(xs)`

Benchmark results are as follows:

before
```
[info] Benchmark                      (size)  Mode  Cnt          Score        Error  Units
[info] VectorBenchmark.builderConcat       1  avgt   20      61304.834 ±   5649.373  ns/op
[info] VectorBenchmark.builderConcat      10  avgt   20     143109.259 ±  22050.573  ns/op
[info] VectorBenchmark.builderConcat     100  avgt   20    1087507.944 ±  15564.617  ns/op
[info] VectorBenchmark.builderConcat    1000  avgt   20    9975918.443 ±  52197.275  ns/op
[info] VectorBenchmark.builderConcat   10000  avgt   20  101155017.712 ± 988804.970  ns/op
```

after
```
[info] Benchmark                      (size)  Mode  Cnt         Score         Error  Units
[info] VectorBenchmark.builderConcat       1  avgt   20     38958.564 ±    1012.915  ns/op
[info] VectorBenchmark.builderConcat      10  avgt   20    116479.188 ±     916.548  ns/op
[info] VectorBenchmark.builderConcat     100  avgt   20    789017.215 ±    4176.121  ns/op
[info] VectorBenchmark.builderConcat    1000  avgt   20   7428021.120 ±   41471.026  ns/op
[info] VectorBenchmark.builderConcat   10000  avgt   20  74379030.621 ± 1083102.872  ns/op
```

### DIFF
--- a/src/library/scala/collection/immutable/Vector.scala
+++ b/src/library/scala/collection/immutable/Vector.scala
@@ -655,7 +655,10 @@ final class VectorBuilder[A]() extends ReusableBuilder[A, Vector[A]] with Vector
     this
   }
 
-  override def ++=(xs: TraversableOnce[A]): this.type = super.++=(xs)
+  override def ++=(xs: TraversableOnce[A]): this.type = {
+    xs.foreach(+=(_))
+    this
+  }
 
   def result: Vector[A] = {
     val size = blockIndex + lo
--- a/test/benchmarks/src/main/scala/scala/collection/immutable/VectorBenchmark.scala
+++ b/test/benchmarks/src/main/scala/scala/collection/immutable/VectorBenchmark.scala
@@ -1,0 +1,42 @@
+package scala.collection.immutable
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra._
+import org.openjdk.jmh.runner.IterationType
+import benchmark._
+import java.util.concurrent.TimeUnit
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(2)
+@Threads(1)
+@Warmup(iterations = 10)
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+class VectorBenchmark {
+  @Param(Array("1", "10", "100", "1000", "10000"))
+  var size: Int = _
+
+  var values: Vector[Any] = _
+
+  @Setup(Level.Trial) def initKeys(): Unit = {
+    values = (0 to size).map(i => (i % 4) match {
+      case 0 => i.toString
+      case 1 => i.toChar
+      case 2 => i.toDouble
+      case 3 => i.toInt
+    }).toVector
+  }
+
+  @Benchmark
+  def builderConcat() = {
+    val builder = Vector.newBuilder[Any]
+    val loopSize = 1000
+    var i = 0
+    while (i < loopSize) {
+      builder ++= values
+      i += 1
+    }
+    builder.result()
+  }
+}
--- a/test/junit/scala/collection/immutable/VectorTest.scala
+++ b/test/junit/scala/collection/immutable/VectorTest.scala
@@ -27,4 +27,21 @@ class VectorTest {
     assertEquals(v, v drop Int.MinValue)
     assertEquals(v, v dropRight Int.MinValue)
   }
+
+  @Test
+  def builderConcat() = {
+    val builder = Vector.newBuilder[Int]
+
+    builder ++= Vector()
+    assertEquals(builder.result(), Vector.empty[Int])
+
+    builder ++= Vector(0)
+    assertEquals(builder.result(), Vector(0))
+
+    builder ++= Vector(1, 2, 3)
+    assertEquals(builder.result(), Vector(0, 1, 2, 3))
+
+    builder ++= Vector()
+    assertEquals(builder.result(), Vector(0, 1, 2, 3))
+  }
 }


Ichoran:
Can you fix this in `Growable` instead?  The problem is that the special loop case is `LinearSeq` when it should be `LinearSeqOptimized`.

monkey-mas:
Thanks for your feedback! I might be wrong but it seems like I can't fix `Grawable#++=(...)` as I want to use [VectorBuilder#+=(...)](https://github.com/scala/scala/pull/5740/files#diff-59f3462485b74027de4fd5e9febcc81bL646) to optimize the code.

Ichoran:
@monkey-mas - Growable has `+=`.

monkey-mas:
> The problem is that the special loop case is LinearSeq when it should be LinearSeqOptimized

Can you please be more specific? What do I exactly have to do for `Growable#++=(...)` in this case?

Ichoran:
Just change `LinearSeq` to `LinearSeqOptimized` and benchmark again.  `Vector` is a `LinearSeq`, but it has a relatively slow head/tail decomposition.  It is _not_ a `LinearSeqOptimized`, so it will fall back to the `foreach(b += _)` branch.

monkey-mas:
> Vector is a LinearSeq

Vector extends [IndexedSeq](https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/Vector.scala#L66). am I correct?

Ichoran:
It extends both IndexedSeq, because it can index more efficiently than O(n), and LinearSeq because it can do a head/tail decomposition more efficiently than O(n). It does _not_ extends LinearSeqOptimized because head/tail decomposition is not among the fastest operations it can do.

xuwei-k:
```scala
$ scala -language:_

scala> :paste

def getSuperClasses(clazz: Class[_]): List[Class[_]] = {
    @scala.annotation.tailrec
    def loop(o: Class[_], result: List[Class[_]]): List[Class[_]] = {
      val superClass = o.getSuperclass
      if (superClass == null) {
        result
      } else {
        loop(superClass, superClass :: result)
      }
    }
    loop(clazz, List(clazz))
  }

  def allSuperClassAndTrait(clazz: Class[_]): List[Class[_]] = {
    def loop(c: Class[_], result: List[Class[_]]): List[Class[_]] = {
      val interfaces = c.getInterfaces.toList
      if (interfaces.size == 0) {
        c :: result
      } else {
        interfaces.flatMap{ i => loop(i, i :: result) }
      }
    }

    val classes = for {
      c <- getSuperClasses(clazz)
      result <- loop(c, Nil)
    } yield result

    (classes.toSet - clazz).toList.sortBy(_.toString)
  }


getSuperClasses: (clazz: Class[_])List[Class[_]]
allSuperClassAndTrait: (clazz: Class[_])List[Class[_]]

scala> allSuperClassAndTrait(classOf[Vector[_]]).foreach(println)
class java.lang.Object
interface java.io.Serializable
interface scala.Equals
interface scala.Function1
interface scala.Immutable
interface scala.PartialFunction
interface scala.Serializable
interface scala.collection.CustomParallelizable
interface scala.collection.GenIterable
interface scala.collection.GenIterableLike
interface scala.collection.GenSeq
interface scala.collection.GenSeqLike
interface scala.collection.GenTraversable
interface scala.collection.GenTraversableLike
interface scala.collection.GenTraversableOnce
interface scala.collection.IndexedSeq
interface scala.collection.IndexedSeqLike
interface scala.collection.Iterable
interface scala.collection.IterableLike
interface scala.collection.Parallelizable
interface scala.collection.Seq
interface scala.collection.SeqLike
interface scala.collection.Traversable
interface scala.collection.TraversableLike
interface scala.collection.TraversableOnce
interface scala.collection.generic.FilterMonadic
interface scala.collection.generic.GenericTraversableTemplate
interface scala.collection.generic.HasNewBuilder
interface scala.collection.immutable.IndexedSeq
interface scala.collection.immutable.Iterable
interface scala.collection.immutable.Seq
interface scala.collection.immutable.Traversable
interface scala.collection.immutable.VectorPointer
```

> It extends both IndexedSeq, because it can index more efficiently than O(n), and LinearSeq 

I don't think so. Vector extends only IndexedSeq in Scala 2.12.1

Ichoran:
@xuwei-k - Oh!  You're right.  My mistake!

Sorry, @monkey-mas!

Still, can you test the speed of adding a list?  That's what the special-case in `Growable` is for.  Vector-vector additions are not test enough of a speed improvement.

I really hate to have to copy superclass code to subclasses to get decent performance.  The algorithm isn't different at all.

lrytz:
Moving to M2 as this is still being discussed and doesn't change / add any functionality

test/benchmarks/src/main/scala/scala/collection/immutable/VectorBenchmark.scala
```diff
@@ -0,0 +1,42 @@
...
+      case 1 => i.toChar
+      case 2 => i.toDouble
+      case 3 => i.toInt
+    }).toVector
+  }
+
+  @Benchmark
+  def builderConcat() = {
+    val builder = Vector.newBuilder[Any]
+    val loopSize = 1000
```
l0rinc, line 34: I think you should let `JMH` do the looping instead:
```scala
@Benchmark
def builderConcatEmpty() = Vector.newBuilder[Any] ++= values

@Benchmark
def builderConcat() = values ++= values
```

src/library/scala/collection/immutable/Vector.scala
```diff
@@ -655,7 +655,10 @@ final class VectorBuilder[A]() extends ReusableBuilder[A, Vector[A]] with Vector
     this
   }
 
-  override def ++=(xs: TraversableOnce[A]): this.type = super.++=(xs)
+  override def ++=(xs: TraversableOnce[A]): this.type = {
```
l0rinc, line 658: Any reason for overriding `++=` instead of `++`?

l0rinc (changes requested):

szeiger:
@monkey-mas Do you have time to address the remaining comment? We're planning to close the queue for M2 until the end of next week.

adriaanm:
Closing since this has been inactive for a while. Please consider resubmitting!

monkey-mas:
I'm very sorry for blocking this. (Haven't got time to do currently.) Please let me reopen it once I'm ready to contribute to it.

PR #5741, opened 2017-02-24
Bump up sbt-jmh to 0.2.21

monkey-mas:
With this change, we'll use JMH 1.17.4.

ref. https://github.com/ktoso/sbt-jmh#versions


SethTisue:
thank you!

PR #5743, opened 2017-02-24
SI-10207 Error before update conversion

som-snytt:
Gaze deeper for errors before committing to conversion
of assignment to update. The error buried in the transformed
tree escapes notice of retypechecking and leaks to backend.


som-snytt:
The deep dive for errors and the position test in tryTypedApply don't look great to me. In the context of, there was an error, can I repair it somehow? maybe that's as disciplined as it gets.

lrytz (approved):

PR #5744, opened 2017-02-25
Fix pattern matching of higher-order GADTs.

sellout:
Fixes scala/bug#10208.


sellout:
I’m probably overoptimistic in thinking that t3346b is now a “pos” test … but I couldn’t figure out how to check what the compiler inferred for `M[_]`.

paulp:
@sellout You are not overoptimistic. It should absolutely be a pos test. You might want to comment the line you added to the effect that Any is kind-polymorphic.

milessabin:
@sellout It'd be great if you could backport this for 2.11.9.

sellout:
@milessabin Happily. Do I just make another PR against that branch, or what?

milessabin:
@sellout yup.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -351,6 +351,7 @@ trait TypeComparers {
   // @assume tp1.isHigherKinded || tp2.isHigherKinded
   def isHKSubType(tp1: Type, tp2: Type, depth: Depth): Boolean = {
     def isSub(ntp1: Type, ntp2: Type) = (ntp1.withoutAnnotations, ntp2.withoutAnnotations) match {
+      case (TypeRef(_, AnyClass, _), PolyType(_, _))                        => true                     // Any is kind-polymorphic
       case (TypeRef(_, AnyClass, _), _)                                     => false                    // avoid some warnings when Nothing/Any are on the other side
       case (_, TypeRef(_, NothingClass, _))                                 => false
```
retronym, line 356: `Nothing` is also kind polymorphic, do we need a symmetrical change here, too?

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -351,6 +351,7 @@ trait TypeComparers {
   // @assume tp1.isHigherKinded || tp2.isHigherKinded
   def isHKSubType(tp1: Type, tp2: Type, depth: Depth): Boolean = {
     def isSub(ntp1: Type, ntp2: Type) = (ntp1.withoutAnnotations, ntp2.withoutAnnotations) match {
+      case (TypeRef(_, AnyClass, _), PolyType(_, _))                        => true                     // Any is kind-polymorphic
```
retronym, line 354: Isn't this saying that `Any` is a subtype of any `PolyType`? Seems the wrong way around.

The caller of this method already checks that:

```
    (    (tp1.typeSymbol eq NothingClass)       // @M Nothing is subtype of every well-kinded type
      || (tp2.typeSymbol eq AnyClass)           // @M Any is supertype of every well-kinded type (@PP: is it? What about continuations plugin?)
      || isSub(tp1.normalize, tp2.normalize) && annotationsConform(tp1, tp2)  // @M! normalize reduces higher-kinded case to PolyType's
    )
```

retronym:
> Do I just make another PR against that branch, or what?

To save some busywork in the future, I'd suggest holding off on doing this until the primary pull request is reviewed.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
retronym: Under this PR, this answer has incorrectly flipped:

```

scala> typeOf[Any] <:< typeOf[List[Any]].typeConstructor
res1: Boolean = false
```

```
Welcome to Scala 2.12.2-20170227-150750-57cf711 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> typeOf[Any] <:< typeOf[List[Any]].typeConstructor
res0: Boolean = true
```

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
paulp: Yeah, I was going to suggest that, except that I doubt it ever gets here with Nothing on the left, because, well, Nothing <:< X forAll X.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
paulp: Those answers look right to me. `List[Any]` is a proper type to which `Any` does not conform. The only proper type `Any` conforms to is `Any`. But the type constructor `List` is unapplied - that's why Any "conforms" to it, because all it's really doing is allowing it to survive this check.

paulp: To elaborate: in principle you shouldn't be able to ask whether `Any <:< List` - it's not a well-formed question. So this is all about dealing with the implementation, which depends on all manner of weird subtyping checks and the side effects thereof.

retronym: As far as I can tell, the main problem to solve is over in `instantiateTypeVars` and `setInst`, which temporarily mutate the info of the enclosing type parameter while typechecking GADT refined cases. They end up assigning a `TypeBounds(lo, hi)` to the type param, which is never seen as being higher-kinded itself (even if `lo` and `hi` are). That ends up tripping up the kind conformance check.

I'd suggest trying something like:

https://github.com/scala/scala/compare/2.12.x...retronym:ticket/10208?expand=1

paulp: There you go @sellout, just clone the HK type params and mutate the new symbols. Welcome to scalac!

retronym: Here's a test case (added to my branch) that exercises a more interesting code path (refining `M` to `Option`. The spurious error under 2.12.1 shows that the `TypeBounds` of two polytypes isn't automatically seen to be a polytype itself, hence the wrapping.

```scala
class Base[M[_]] { def mint: M[Int] = ??? }
final case class Sub() extends Base[Option]

class Test {
  def test[M[_]](b: Base[M]): Option[Int] = {
    bar : M[Any]
    b match {
      case Sub() =>
        bar : M[Any] // incorrect error: M does not take type parameters
        b.mint // GADT refinement: M = Option

        // Fix: change GADT refinement for HK type params to assign M the type:
        // [A]( >: Option[A] <: Option[A]),
        // rather than:
        // >: Option <: Option
        //
    }
  }
  def bar[N[_]]: N[Any] = ???
}
```

We have the same sort of bug in `lub` calculations, too.

```
scala> val tps = ListClass.typeConstructor :: OptionClass.typeConstructor :: Nil; val thelub = lub(tps)
tps: List[$r.intp.global.Type] = List(List, Option)
thelub: $r.intp.global.Type = Product with Serializable

scala> tps.forall(_.isHigherKinded)
res7: Boolean = true

scala> thelub.isHigherKinded
res8: Boolean = false
```

retronym: > mutate the new symbols. 

If only GADT refinement were that principled! We're mutating the info of the actual method type parameter symbol.

retronym: > We have the same sort of bug in lub calculations, too.

I can imagine the correct thing to do for hk lubs would be along the lines of:

  - expand kind polymorphic `Any`/`Nothing` to, e.g, `[a, b]Any` to align with the arity of the fellow lubees
  - come up with a set of type params for the result of the lub. Each type param would have a kind (arity + bounds) derived from the corresponding tparm of the list of types. The individual bounds would themselves need to be lub-ed/glb-ed somehow.
  - substitute the result type params for the original ones in the result types of the list of types
  - lub them.

Working the example above (albeit without handling higher order type parameters yet)

```
scala> val polytps = tps.map(_.normalize)
polytps: List[$r.intp.global.Type] = List([+A]List[A], [+A]Option[A])

scala> val (tparamss, results) = polytps.map(x => (x.typeParams, x.resultType)).unzip
tparamss: List[List[$r.intp.global.Symbol]] = List(List(type A), List(type A))
results: List[$r.intp.global.Type] = List(List[A], Option[A])

scala> tparamss.map(_.map(_.typeParams))
res14: List[List[List[$r.intp.global.Symbol]]] = List(List(List()), List(List()))

scala> tparamss.map(_.map(_.info.bounds.hi))
res15: List[List[$r.intp.global.Type]] = List(List(Any), List(Any))

scala> tparamss.map(_.map(_.info.bounds.lo))
res16: List[List[$r.intp.global.Type]] = List(List(Nothing), List(Nothing))

scala> val newTparams = NoSymbol.newSyntheticTypeParams(tparamss.head.length)
newTparams: List[$r.intp.global.TypeSymbol] = List(type T0)

scala> foreachWithIndex(newTparams)((sym, i) => {val tparams = tparamss.map(_(i)); sym.setInfo(TypeBounds(lub(tparams.map(_.info.bounds.lo)), glb(tparams.map(_.info.bounds.hi))))})

scala> val results = mapWithIndex(polytps)((tp, i) => tp.resultType.instantiateTypeParams(tp.typeParams, newTparams.map(_.tpeHK)))
results: List[$r.intp.global.Type] = List(List[T0], Option[T0])

scala> PolyType(newTparams, lub(results))
res29: $r.intp.global.PolyType = [T0]Product with Serializable
```

paulp: > We're mutating the info of the actual method type parameter symbol.

No, don't take @sellout down with you! He's not lost like you and I! He can still live a good life and someday enter the kingdom of heaven!

retronym:
I'm going to close this one for now, but I'd encourage you to try out my branch on some more real world examples you're cooking up. It could do with a few more tests, but isn't too far away from something we could merge.

milessabin:
What's the justification for closing a work in progress? If you need to clear the decks for 2.12.2 then by all means move the milestone, or add a label, but I don't see how closing this now is either helpful or encouraging of external contributors.

retronym:
Reopened. Didn't mean to dissuade.

Jacoby6000:
I'm not so sure where I should post this, but it seems like this is the appropriate place.  

~~@retronym using your branch, I'm able to consistently get a compilation failure from [this commit](https://github.com/Jacoby6000/scoobie/tree/3a02f6c387a8b9b0f6c9cb735094478b0dc880b3), but only during incremental compilation. A clean compile works fine, but if I trigger a recompilation of [ProjectionTests.scala](https://github.com/Jacoby6000/scoobie/tree/3a02f6c387a8b9b0f6c9cb735094478b0dc880b3/core/src/test/scala/scoobie/ProjectionTests.scala), I get [this failure](https://gist.github.com/Jacoby6000/5ed033987a06f42591c5d4c7ec72ab15).  I'll see if I can get a minimal example.~~

Upon further inspection, that looks like an SBT failure.  Should I go post this there?

Jacoby6000:
My previous comment seemed to be an SBT issue, but this time, I've come back with a reproducable issue!

```scala
case class Bar[A[_]](list: List[A[String]])
def doThing[A[_]]: Bar[A] = Bar(List.empty)
```
The above compiles fine in `2.12.2`, but using @retronym's branch it hangs during the typer phase.

Manually passing the type parameter for `A[_]` fixes the issue, so it seems like an inference problem.

Jacoby6000:
I've found another failing case, when trying to create an algebra for a paramorphism.

```scala
object Test {


  trait ~>[A[_], B[_]] {
    def apply[I](fa: A[I]): B[I]
  }

  type :*:[F[_], G[_]] = {
    type λ[A] = (F[A], G[A])
  }

  type ParaShape[F[_[_[_], _], _], G[_[_], _], A[_]] = {
    type in[I] = G[(({type λ[α] = ({type Λ[β[_[_], _]] = F[β, α]})#Λ[G]})#λ :*: A)#λ, I]
    type λ = in ~> A
  }

  // HIGHER KINDED GADT
  sealed trait GADTK[A[_], I]
  final case class MemberK[A[_]](i: Int) extends GADTK[A, Int]

  case class Const[I](i: Int)

  case class FixH[G[_[_], _], I](unfix: G[({type λ[α] = FixH[G, α]})#λ, I])

  def doesNotCompile: ParaShape[FixH, GADTK, Const]#λ =
    new (ParaShape[FixH, GADTK, Const]#λ) {
      def apply[I](v: ParaShape[FixH, GADTK, Const]#in[I]): Const[I] = v match {
        case MemberK(i) => Const(i)
      }
    }
}
```

fails with 

```scala
t10208c.scala:28: error: constructor cannot be instantiated to expected type;
 found   : Test.MemberK[A(in class MemberK)]
 required: Test.GADTK[[A(in type λ)](Test.FixH[Test.GADTK,A(in type λ)], Test.Const[A(in type λ)]),I]
        case MemberK(i) => Const(i)
             ^
one error found
```

I'd like to start hacking on the compiler to help resolve this issue, but I'm not sure where my work/comments about progress should go.  This PR is for @sellout's branch, but @retronym's branch is the best working implementation currently, and I'll likely poke around where he left off.

SethTisue:
closing for inactivity. someone want to pick this up...?

milessabin:
I'll pick this one up.

adriaanm:

PR #5745, opened 2017-02-25
Sketch of Steque

edmundnoble:
Sketch of stack-ended queues for Scala 2.13 (`Steque`). A persistent, functional data structure with implementation and design courtesy of fs2.util, it provides:
- constant-time (*very* fast: one or two constructor calls) `:+` (append), `+:` (prepend), `++` (concat) and `fromSeq[A](seq: Seq[A]): Steque[A]`
- amortized-constant-time (tight loop; amortization depends on using `uncons` to walk down the entire length of a `Steque`) `uncons`, `head`, `tail` (and potentially `splitAt`)
- linear-time `foldLeft`, `foreach` and friends

In these tasks, it beats out `Vector` in speed by an order of magnitude - see benchmarks at edmundnoble/scabsbench with results at https://ethercalc.org/kknm1xz3mb5s_3ht2cokr07oy/app. Some things it does badly include:
- indexing (linear time)
- `last`/`init` (degenerates to indexing into the last element)
- it allocates more than `Vector` (because `Vector` uses `Array`s underneath)

In terms of implementation, `Steque` is an association structure built from a binary tree which may have entire `Seq`s or single values at its leaves. For this reason, the asymptotics of consuming a `Steque` will also depend on those of the underlying collection. The asymptotics of `uncons`, `head`, `tail` and `foldLeft` will require constant-time `head`, `tail` and `foldLeft` for all `Seq` values included in the `Steque`.

`Steque` is essentially a binary tree which will graft subtrees from the left to the right as needed (i.e. as `uncons` is called), eventually flattening out into the trivial right-associated binary tree (cons list). It *may* be possible to reverse the algorithm to graft subtrees to the left and create a left-associated binary tree to provide amortized-constant `unsnoc`, but alternating between `unsnoc` and `uncons` will be incredibly expensive.


Jasper-M:
Isn't it a bad thing that underlying Seqs can be any kind of Seq, even mutable ones?

edmundnoble:
@Jasper-M It's a trade-off. Either allow mutable `Seq`s or force users to defensively copy; IMO the open use of `Seq` already makes `Steque` unsuitable for serialization, so that's one potential downside out the window.

lrytz:
Changes to collections in 2.13 needs to coordinated with the work at https://github.com/scala/collection-strawman, so I'm closing it here. @edmundnoble can you open an issue or PR there?

SethTisue:
Some version of this could be submitted to https://github.com/scala/scala-collection-contrib
and perhaps eventually https://github.com/scala/scala-library-next

(neither repo existed in 2017)

PR #5747, opened 2017-02-26
SD-317 Fix package & compiler/package

dwijnand:
repl-jline-embedded contains everything that repl-jline contains, except
for scala.tools.nsc.interactive.jline being rewritten to jline_embedded,
which is what we want.

Defining both sets leads to the following mapping error:

    [error] (compiler/compile:packageBin) java.util.zip.ZipException: duplicate entry: repl-jline.properties

Fixes scala/scala-dev#317.

### DIFF
--- a/project/Osgi.scala
+++ b/project/Osgi.scala
@@ -50,7 +50,8 @@ object Osgi {
       "Bundle-SymbolicName" -> (bundleSymbolicName.value + ".source"),
       "Bundle-Version" -> versionProperties.value.osgiVersion,
       "Eclipse-SourceBundle" -> (bundleSymbolicName.value + ";version=\"" + versionProperties.value.osgiVersion + "\";roots:=\".\"")
-    )
+    ),
+    Keys.`package` := bundle.value
   )
 
   def bundleTask(headers: Map[String, String], jarlist: Boolean, fullClasspath: Seq[File], artifactPath: File,


retronym:
> Defining both sets

I'm not sure what this means. 😕

dwijnand:
Including both sets of `products`, jline-repl and jline-repl-embedded's, in the compiler project's `products` leads to the above error.

I'm simply removing the smaller of the two, duplicate sets of products (classfiles, properties, etc..).

dwijnand:
By "defining both sets" I meant "defining the compiler products set on both jline-repl and jline-repl-embedded products sets leads to an error". So I removed the duplicate subset, fixing the problem.

szeiger:
sbt's `compiler/package` should never be called. There is no expectation that it would do the right thing and after this fix it still doesn't. We have to produce OSGi-enabled JARs which is why we use bnd for packaging.

I think the proper fix is to call our own bundling task `package`, thus overriding sbt's default implementation. We currently do the same thing as `sbt-osgi`, which means replacing the packaging *result* but not the original packaging *task*:

```
packagedArtifact in (Compile, packageBin) := (((artifact in (Compile, packageBin)).value, bundle.value)),
```

This calls `bundle` instead of `package` to produce the result but obviously doesn't prevent you from calling `package` manually.

dwijnand:
Thanks for looking into this and sharing the detail you know, @szeiger.

What definition of "the right thing" should we use? We can either:

1. Follow sbt-osgi's decision (which you mentioned on Gitter you had specifically honoured) that `package` should continue to create the normal, non OSGi-enabled jar, leaving that task to `bundle`. In which case this fix is correct, imo.

2. Go against sbt-osgi's decision (which is what you're saying now, here) and say that the OSGi-enabled bundle _is_ "the main artifact". Therefore we should re-wire as you said.

That is, of course, if it was a conscious decision. https://github.com/sbt/sbt-osgi/pull/32 is proposing to change the behaviour, but no one's reviewed it in 2 months.

I think it's debatable both ways. I see you have more experience than me with OSGi so have a think and let me know. Happy to reimplement this fix.

szeiger:
I don't see the value in having 'package' build a non-OSGi-enabled JAR in addition to 'bundle' building the OSGI-enabled one. Both take bunch of files and put them into a zip archive, they only differ in the additional metadata they produce. Being able to produce a non-OSGi-enabled JAR will only add more confusion because you can't distinguish it from the proper OSGi bundle unless you look specifically for the OSGi metadata.

I didn't consider going against `sbt-osgi` by replacing the `packageBin` task because I assumed nobody would call `packageBin` or `package` manually. I certainly hadn't done that before on any other project.

dwijnand:
Rewired. @szeiger WDYT?

build.sbt
```diff
@@ -406,7 +406,6 @@ lazy val compiler = configureAsSubproject(project)
         (products in Compile in packageBin in LocalProject("interactive")).value ++
         (products in Compile in packageBin in LocalProject("scaladoc")).value ++
         (products in Compile in packageBin in LocalProject("repl")).value ++
-        (products in Compile in packageBin in LocalProject("repl-jline")).value ++
```
szeiger: Why remove `repl-jline`? We want to be able use both, shadowed and non-shadowed JLine versions, so we need both projects.

project/Osgi.scala
```diff
@@ -50,7 +50,8 @@ object Osgi {
       "Bundle-SymbolicName" -> (bundleSymbolicName.value + ".source"),
       "Bundle-Version" -> versionProperties.value.osgiVersion,
       "Eclipse-SourceBundle" -> (bundleSymbolicName.value + ";version=\"" + versionProperties.value.osgiVersion + "\";roots:=\".\"")
-    )
+    ),
+    Keys.`package` := bundle.value
```
szeiger, line 54: Do the doc and source packages still work correctly? We only want to change `packageBin`.

build.sbt
dwijnand: Where/how do you "want to be able to use both"?

Before rewiring the package key, I was taking the fact that it was failing as proof that this setting was wrong.

But post-rewiring I have no proof, so I'm happy to drop it if you'd like.

project/Osgi.scala
dwijnand: Yes. That's `pacakgeSrc` and `packageDoc`.

All we're doing is replacing the [default](https://github.com/sbt/sbt/blob/v0.13.13/main/src/main/scala/sbt/Defaults.scala#L647) 

```scala
`package` := packageBin.value
```

wiring with this one.

dwijnand:
Disk space issues.

/rebuild

lrytz:
/rebuild (needs to be first in the comment)

dwijnand:
@szeiger this one-line change should be good to go.

build.sbt
szeiger: `packageBin` doesn't prove anything, it was never intended to work.

The REPL tries multiple JLine options and picks the best one:

```
$ scala -nc -Ydebug
Trying to instantiate a InteractiveReader from scala.tools.nsc.interpreter.jline.InteractiveReader
Trying to instantiate a InteractiveReader from scala.tools.nsc.interpreter.jline_embedded.InteractiveReader
All InteractiveReaders tried:
  - scala.tools.nsc.interpreter.jline.InteractiveReader OK
  - scala.tools.nsc.interpreter.jline_embedded.InteractiveReader OK
```

Both `jline` and `jline_embedded` should show up here in a Scala distribution that you build.

project/Osgi.scala
szeiger: Hm, right, the only reverse dependency for `packageBin` is `package`, so it shouldn't matter which one you replace.

dwijnand: Yeah. I'm happy to go either way. I picked `package` because its description says "This is typically an alias for the task that actually does the packaging.". So it felt more natural to do it this way.

SethTisue:
thank you Dale!

dwijnand:
Cool. Thanks.

PR #5749, opened 2017-02-27
Fix pattern matching of higher-order GADTs.

sellout:
Fixes SI-10208. (Backport of #5744.)


SethTisue:
note that #5744 is in limbo currently

SethTisue:
closing for inactivity. we can always reopen if activity resumes

PR #5751, opened 2017-03-01
make Predef.*CharSequence implicits public again

SethTisue:
fixing a recent regression in 2.13

when I was removing deprecated stuff in #5683,
I was going too fast and misunderstood these deprecations.
only the names of the class parameters were supposed to become
private, not the entire class.

note that now that the parameters aren't even vals anymore, they don't
need to have funny `__` names anymore, either.

this was caught by the Scala community build.

### DIFF
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -314,19 +314,19 @@ object Predef extends LowPriorityImplicits {
     def +(other: String): String = String.valueOf(self) + other
   }
 
-  private[scala] implicit final class SeqCharSequence(val __sequenceOfChars: scala.collection.IndexedSeq[Char]) extends CharSequence {
-    def length: Int                                     = __sequenceOfChars.length
-    def charAt(index: Int): Char                        = __sequenceOfChars(index)
-    def subSequence(start: Int, end: Int): CharSequence = new SeqCharSequence(__sequenceOfChars.slice(start, end))
-    override def toString                               = __sequenceOfChars mkString ""
+  implicit final class SeqCharSequence(sequenceOfChars: scala.collection.IndexedSeq[Char]) extends CharSequence {
+    def length: Int                                     = sequenceOfChars.length
+    def charAt(index: Int): Char                        = sequenceOfChars(index)
+    def subSequence(start: Int, end: Int): CharSequence = new SeqCharSequence(sequenceOfChars.slice(start, end))
+    override def toString                               = sequenceOfChars mkString ""
   }
 
   /** @group implicit-classes-char */
-  private[scala] implicit final class ArrayCharSequence(val __arrayOfChars: Array[Char]) extends CharSequence {
-    def length: Int                                     = __arrayOfChars.length
-    def charAt(index: Int): Char                        = __arrayOfChars(index)
-    def subSequence(start: Int, end: Int): CharSequence = new runtime.ArrayCharSequence(__arrayOfChars, start, end)
-    override def toString                               = __arrayOfChars mkString ""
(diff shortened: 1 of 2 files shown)


test/files/pos/CharSequence-implicits.scala
```diff
@@ -0,0 +1,4 @@
+object Test extends App {
+  Array('a', 'b'): CharSequence
+  IndexedSeq('a', 'b'): CharSequence
+}
```
som-snytt: Since it's a library fix, it could be a junit test, just not in `scala` package.

SethTisue: All of our existing JUnit tests are in the `scala` package under `test/junit/scala`. But yeah, I _could_ be the pioneer here. We'd need to establish a convention. `test/junit/notscala`? `test/junit/scalauser`? And document the convention. And think about whether existing tests ought to be reclassified. I'm not sure whether I feel like tackling that right now. How about if I open a scala-dev ticket on it and add a comment to this one test...? Curious what the rest of the team thinks.

SethTisue: Note that `private[scala]` is relatively (though not extremely) rare: 93 occurrences in `src`, nearly half of which are in `src/reflect/scala/reflect/internal`. and in the case of this bug, `private[scala]` was only introduced by a mistake, then taken out again.

som-snytt: My comment was funnier last night: ironic yet bathetic.

SethTisue: I converted it to a JUnit test, and even though the test is in `package scala`, it actually doesn't compile if the implicits in question are declared `private[scala]`. So that's good, actually, the test tests everything. (But I don't know why.)

SethTisue: Adriaan was also surprised by this; I will investigate separately, but not right away. (In the meantime, it doesn't affect merging this.)

som-snytt: OK, I take it back, my comment is funny again. Thanks for the chuckle! I also appreciate the air of mystery. It's kind of like Penn and Teller, humor and magic.

SethTisue: Pretty sure @adriaanm is Penn and you're Teller. Not sure where I fit in here. (Maybe I'm Siegfried or Roy.)

som-snytt:
I guess it's too late now, but you called it `SomethingTests.scala`, plural.

I challenge you to say "Tests by Seth" ten times fast.

som-snytt:
I had a theory that it had to do with name binding and a nesting such as:

`import scala.Predef._ ; package scala { ArrayCharSequence }`

since wildcard imports are higher precedence (3) than defs made available by package clause  && not defined here (4). (Though I'm not sure the fix for (4) was ever merged.)

Is it possible that `private[scala]` triggers (4)? Or maybe it's only an ambiguous reference for implicit resolution.

`ArrayCharSequence` would be ambiguous because it is (4) and cannot shadow the import at (3).

I did try:

```
scala> :pa

package scala {
object X { 
val x = scala.Predef.ArrayCharSequence _  
val y = (Array('a', 'b'): CharSequence)
}}

<pastie>:4: error: type mismatch;
 found   : Array[Char]
 required: CharSequence
val y = (Array('a', 'b'): CharSequence)
              ^

scala> :pa

package scala {
object X {
def x = ArrayCharSequence _
def y = implicitly[Array[Char] => CharSequence]
}}

<pastie>:4: error: No implicit view available from Array[Char] => CharSequence.
def y = implicitly[Array[Char] => CharSequence]
                  ^
```

PR #5752, opened 2017-03-01
Document error in Regex

libratiger:
There are two errors in the ```Regex``` document.

The first one is lack of a }.
The other one is error for code example.

### DIFF
--- a/src/library/scala/util/matching/Regex.scala
+++ b/src/library/scala/util/matching/Regex.scala
@@ -116,6 +116,7 @@ import java.util.regex.{ Pattern, Matcher }
  *  while (mi.hasNext) {
  *    val d = mi.next
  *    if (mi.group(1).toInt < 1960) println(s"$d: An oldie but goodie.")
+ *  }
  *  }}}
  *
  *  Note that `findAllIn` finds matches that don't overlap. (See [[findAllIn]] for more examples.)
@@ -132,11 +133,13 @@ import java.util.regex.{ Pattern, Matcher }
  *  {{{
  *  val ns = num.findAllIn("1 2 3")
  *  ns.start    // 0
+ *  ns.next     // 1
  *  ns.hasNext  // true
  *  ns.start    // 2
  *  val ms = num.findAllMatchIn("1 2 3")
  *  val m  = ms.next()
  *  m.start     // 0
+ *  ms.next     // 2
  *  ms.hasNext  // true
  *  m.start     // still 0
  *  }}}


libratiger:
@janekdb can you take a look at?

som-snytt:
Thanks, it took me a while to recall what was supposed to happen in the example. It's not clear that advancing `start` without calling `next` is unconfusing enough. That is, `hasNext` could advance the underlying matcher, but `start` could report the previous. It is clear that advancing `start` to the first match without `next` is OK, because throwing was confusing.

libratiger:
@som-snytt 

The real problem is that the origin code example is wrong, it is not necessary call the `next` , but I just try to guess the intention of the example, and calling the `next` can output the right answer.

```scala
val ns = num.findAllIn("1 2 3")
ns.start    // 0
ns.hasNext  // true
ns.start    // 2   <---- This is wrong, the output is 0.
```

som-snytt:
Rephrasing previous comment: the example should demonstrate something. It's trying to show that calling MatchData methods immediately don't throw: `r.findAllIn(s).start`; and that calling `hasNext` will side-effect by advancing the underlying matcher: `r.findAllIn(s).hasNext` or when called after `next`.

So correctness is important, but if it takes me time to understand what the example is for, then it's a bad example.

libratiger:
In fact, it takes me a little time to understand the origin example, the example modified by me is still not understand quickly at first。

do you have any idea to improve this example?

som-snytt:
The junit test I linked to previously:

https://github.com/scala/scala/blob/2.12.x/test/junit/scala/util/matching/RegexTest.scala#L119

It's almost the same, showing `hasNext` bumping `start`. But the output is a bit more obvious.

The text for the example is confusing, too, because the API is confusing. `hasNext may advance the MatchIterator and change the current MatchData.` Then the second example could say: `use findAllMatchIn to produce Match objects that capture the current match.`

When I use this API, not often, I reach for `findAllMatchIn` first because it is not subtle.

libratiger:
How about change the text for this example to the following.

>Also, calling `hasNext` or `next` in `findAllIn` may advance the `MatchIterator` and change the current match, 
while current matched object  in `findAllMatchIn` remain valid after the iterator is advanced. 


We just need to  stand out the `hasNext` or `next` may change something.

som-snytt:
I applied the insights https://github.com/scala/scala/pull/5771 which may supersede this PR.

adriaanm:
Welcome, @djvulee -- thanks for getting the ball rolling on this improvement! Should we close this PR in favor of #5771? We are no longer accepting new PRs for 2.11.9, by the way, so this will have to be 2.12-only.

SethTisue:
I see a "I will close my pull request" over at the other PR, so I'll go ahead and close. thanks @djvulee for taking this on

PR #5753, opened 2017-03-02
bump copyright year to 2017 [ci: last-only]

Philippus:


SethTisue:
nice catch, thank you! would you mind re-targeting and rebasing onto 2.11.x, actually? we'll need this for 2.11.9.

Philippus:
@SethTisue I think I did it. :)
one thing to note is that 2.12.x branch has a LICENSE file in the root dir that should also be updated.

SethTisue:
thank you!

SethTisue:
> one thing to note is that 2.12.x branch has a LICENSE file in the root dir that should also be updated

hopefully we'll catch that at merge time, but I've also made a to-do item for myself to make sure it doesn't get dropped.

PR #5754, opened 2017-03-03
fix regression in scaladoc

Philippus:
pattern for entitylink was too narrow, cleaned up the tests
this should fix the regression introduced in https://github.com/scala/scala/pull/5728


lrytz (approved):
thank you!

PR #5755, opened 2017-03-04
Improve the performance of Map4 to HashMap and Set4 to HashSet transitions

rorygraves:
When using immutable Map there is specialisation for empty,1,2,3,4 sized maps - then it moves to an immutable HashMap - the transition from Map4 to HashMap was inefficient generating tuples and a var args array wrapper.  
A similar things happens with Set4 to HashSet
Bonus fix for compilation error in existing ListBenchmark

**Map Before**
[info] Benchmark                 Mode  Cnt    Score   Error  Units
[info] MapBenchmark.addElement  avgt   20  364.899 ± 5.038  ns/op
**Map After**
[info] Benchmark                 Mode  Cnt   Score   Error  Units
[info] MapBenchmark.addElement  avgt   20  91.522 ± 4.136  ns/op

**Set Before**
[info] Benchmark                    Mode  Cnt    Score   Error  Units
[info] SetBenchmark.set4AddElement  avgt   20  226.679 ± 6.856  ns/op
**Set After**
[info] Benchmark                    Mode  Cnt   Score   Error  Units
[info] SetBenchmark.set4AddElement  avgt   20  86.789 ± 2.938  ns/op

### DIFF
--- a/src/library/scala/collection/immutable/Map.scala
+++ b/src/library/scala/collection/immutable/Map.scala
@@ -198,7 +198,7 @@ object Map extends ImmutableMapFactory[Map] {
       else if (key == key2) new Map4(key1, value1, key2, value, key3, value3, key4, value4)
       else if (key == key3) new Map4(key1, value1, key2, value2, key3, value, key4, value4)
       else if (key == key4) new Map4(key1, value1, key2, value2, key3, value3, key4, value)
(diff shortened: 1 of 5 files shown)


retronym (approved):

retronym:
Thanks!

PR #5756, opened 2017-03-05
Reduce boxing of scala.reflect.internal.Depth

rorygraves:
A fairly trivial change to Depth to remove the boxing generated during the max function.

### DIFF
--- a/src/reflect/scala/reflect/internal/Depth.scala
+++ b/src/reflect/scala/reflect/internal/Depth.scala
@@ -5,7 +5,7 @@ package internal
 import Depth._
 
 final class Depth private (val depth: Int) extends AnyVal with Ordered[Depth] {
-  def max(that: Depth): Depth   = if (this < that) that else this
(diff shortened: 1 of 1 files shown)


retronym (approved):

nafg:
Just saw this in @rorygraves' talk, maybe there's a related microoptimization, the slide also has what I assume is from https://github.com/scala/scala/blob/2.13.x/src/reflect/scala/reflect/internal/BaseTypeSeqs.scala#L242 --

```scala
def maxDepthOfElems: Depth = elems.map(x => typeDepth(f(x))).max
```

Maybe insignificant, but you could probably save an allocation by using `maxBy` instead of `map` and `max`

rorygraves:
@nafg This was fixed in the base class - but the subclass overrides it (which I missed) and needs a similar optimisation (can't use the original one as it adds a mapping function).

see https://github.com/scala/scala/blob/2.13.x/src/reflect/scala/reflect/internal/BaseTypeSeqs.scala#L137

PR #5757, opened 2017-03-05
new version numbers for snapshot and integration builds, new repository

lrytz:
Integration builds now have version number like `2.12.2-bin-sha7` or `2.13.0-pre-sha7` and are published to scala-integration (no longer scala-release-temp).

scala-release-temp is still used in the bootstrap script for publishing intermediate artifacts (starr, locker).

Various cleanups in the scripts.

### DIFF
--- a/project/VersionUtil.scala
+++ b/project/VersionUtil.scala
@@ -53,16 +53,17 @@ object VersionUtil {
   /** Compute the canonical, Maven and OSGi version number from `baseVersion` and `baseVersionSuffix`.
     * Examples of the generated versions:
     *
-    * ("2.11.8", "SNAPSHOT"    ) -> ("2.11.8-20151215-133023-7559aed", "2.11.8-SNAPSHOT",         "2.11.8.v20151215-133023-7559aed")
-    * ("2.11.8", "SHA-SNAPSHOT") -> ("2.11.8-20151215-133023-7559aed", "2.11.8-7559aed-SNAPSHOT", "2.11.8.v20151215-133023-7559aed")
-    * ("2.11.8", "SHA-NIGHTLY" ) -> ("2.11.8-7559aed-nightly",         "2.11.8-7559aed-nightly",  "2.11.8.v20151215-133023-NIGHTLY-7559aed")
-    * ("2.11.8", ""            ) -> ("2.11.8",                         "2.11.8",                  "2.11.8.v20151215-133023-VFINAL-7559aed")
-    * ("2.11.8", "M3"          ) -> ("2.11.8-M3",                      "2.11.8-M3",               "2.11.8.v20151215-133023-M3-7559aed")
-    * ("2.11.8", "RC4"         ) -> ("2.11.8-RC4",                     "2.11.8-RC4",              "2.11.8.v20151215-133023-RC4-7559aed")
-    * ("2.11.8-RC4", "SPLIT"   ) -> ("2.11.8-RC4",                     "2.11.8-RC4",              "2.11.8.v20151215-133023-RC4-7559aed")
+    * ("2.11.8", "SNAPSHOT"    ) -> ("2.11.8-20151215-133023-7559aed", "2.11.8-bin-SNAPSHOT",         "2.11.8.v20151215-133023-7559aed")
+    * ("2.11.8", "SHA-SNAPSHOT") -> ("2.11.8-20151215-133023-7559aed", "2.11.8-bin-7559aed-SNAPSHOT", "2.11.8.v20151215-133023-7559aed")
+    * ("2.11.8", "SHA"         ) -> ("2.11.8-7559aed",                 "2.11.8-bin-7559aed",          "2.11.8.v20151215-133023-7559aed")
+    * ("2.11.0", "SHA"         ) -> ("2.11.0-7559aed",                 "2.11.0-pre-7559aed",          "2.11.0.v20151215-133023-7559aed")
+    * ("2.11.8", ""            ) -> ("2.11.8",                         "2.11.8",                      "2.11.8.v20151215-133023-VFINAL-7559aed")
+    * ("2.11.8", "M3"          ) -> ("2.11.8-M3",                      "2.11.8-M3",                   "2.11.8.v20151215-133023-M3-7559aed")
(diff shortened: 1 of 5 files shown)


scripts/jobs/integrate/bootstrap
```diff
@@ -109,9 +102,8 @@ rm -rf $baseDir/resolutionScratch_
 mkdir -p $baseDir/resolutionScratch_
 
 # repo used to publish "locker" scala to (to start the bootstrap)
-releaseTempRepoCred="private-repo"
```
adriaanm: no longer used?

lrytz: yes, it seems so

lrytz:
@adriaanm I pushed another commit that changes the version numbers as discussed. Still need to check if that breaks assumptions in downstream repos (scala-dist, etc), you mention some here https://github.com/scala/scala-dev/issues/252. There are some questions about details: https://github.com/scala/scala/pull/5757/commits/7ed95cd3f6acdc06ce2133b00f45ceff9414347d#diff-19e53133756351b63e653b6a14dfc5ccL89

Another question is whether this change here makes sense: https://github.com/scala/scala/pull/5757/commits/7ed95cd3f6acdc06ce2133b00f45ceff9414347d#diff-661d447ea0c178cb62b2e4e9b2c84f66L399

Another thing: instead of publishing locker (and possibly starr) to the same repo as the boostrapped version, we could use scala-release-temp for the starr and locker builds, and scala-integration only for the final results. scala-release-temp could discard builds older than a month or so.

project/VersionUtil.scala
```diff
@@ -85,12 +86,27 @@ object VersionUtil {
...
+    }
+
     val (canonicalV, mavenSuffix, osgiV, release) = suffix match {
-      case "SNAPSHOT"     => (s"$base-$date-$sha",   s"-SNAPSHOT",      s"$base.v$date-$sha",         false)
-      case "SHA-SNAPSHOT" => (s"$base-$date-$sha",   s"-$sha-SNAPSHOT", s"$base.v$date-$sha",         false)
-      case "SHA-NIGHTLY"  => (s"$base-$sha-nightly", s"-$sha-nightly",  s"$base.v$date-NIGHTLY-$sha", true)
-      case ""             => (s"$base",              "",                s"$base.v$date-VFINAL-$sha",  true)
-      case suffix         => (s"$base-$suffix",      s"-$suffix",       s"$base.v$date-$suffix-$sha", true)
+      // TODO: should this remain `-SNAPSHOT`? does `-SNAPSHOT` have a special meaning for maven / sbt? does `bin-SNAPSHOT` affect this?
+      // TODO: include `-bin` / `-pre` in the canonical version? in the osgi version?
```
adriaanm: /cc @dwijnand re: -bin- version numbering vs -SNAPSHOT

scripts/jobs/integrate/bootstrap
adriaanm: Ah, it was used before in 2.11.x, but seems like it stayed within the bootstrap script

adriaanm:
> Another thing: instead of publishing locker (and possibly starr) to the same repo as the boostrapped version, we could use scala-release-temp for the starr and locker builds, and scala-integration only for the final results. scala-release-temp could discard builds older than a month or so.

Good idea! I have a vague memory that bootstrapping required the version used for locker and quick to coincide (to tie the recursive knot of scala-library and scala-$module deps), but not sure anymore.

project/VersionUtil.scala
dwijnand: Happy to help. But I need a little info here as I don't know or grok the context here.

Could someone give me some context please? Either here, or a chat room or a voice chat.

adriaanm: I'll let Lukas provide more detail tomorrow, but the gist as I understand it is how to design our version numbering to have the binary version derived correctly, using the `-bin-` delimiter. Specifically, how does it interact with the -SNAPSHOT suffix?

dwijnand: It's hard to tell, because there are so many systems that try to infer meaning from versions. Specifically: Maven, Ivy, the parts of Ivy that sbt injects, and sbt itself. And then there's OSGI which I _really_ don't know.

@cunei for instance (IIRC) was particularly concerned with using non-alphabetic sections in versions (I forgot the details).

In absence of some other, better informed, authority I would suggest experimentation and testing. :-/

With regards to strictly sbt's logic, I believe 2.12.2-bin-SNAPSHOT should work fine. The `-SNAPSHOT` would declare it as a non-stable version, and the `-bin` part would make it be considered binary compatible with the rest of the 2.12 series:

https://github.com/sbt/sbt/blob/v0.13.13/util/cross/src/main/input_sources/CrossVersionUtil.scala#L42

adriaanm: Ok, sounds good! I should've been more explicitly in that I was not expecting you to investigate this beyond your perspective on how sbt will interpret this.

lrytz: @dwijnand thanks. Just a detail: IUC, `2.12.1-SNAPSHOT` would also get the short binary version `2.12`, it fits the `NonReleaseV` pattern. However, `2.12.1-sha-nightly` would get the long binary version, as the second `-` doesn't match `\w+`.

Anyway, as you see in this PR, we don't want to rely on this detail, but instead use versions like `2.12.1-bin-.....` and `2.13.0-pre-...`.

dwijnand: You're right. I'm blaming lack of vertical alignment making it harder to comprehend.

I see what you're doing and it looks good.

lrytz:
I merged https://github.com/scala/scala-jenkins-infra/pull/223, wich means that the bootstrap jobs now publish to the new repo (https://scala-ci.typesafe.com/artifactory/scala-integration/).

This is also true for existing and new commits in 2.11.x, 2.12.x and 2.13.x, even before the PR here is merged, as there's a new indirection through this script: https://github.com/scala/scala-jenkins-infra/blob/master/externalScripts/jobs/integrate/bootstrap

So I'll wait one more day before merging this one to have fewer moving parts.

cc @SethTisue @adriaanm @retronym

PR #5758, opened 2017-03-05
Report per-phase JVM statistics (e.g cpu time, user time, allocated bytes)

mkeskells:
add some profiling support to scalac

This adds  a few -Yprofile... options that assist in profiling scalac

This has proved useful so far in https://github.com/rorygraves/scalac_perf for identification of opportunities for optimisation

The output can be generated to a console, or a CSV.
This could enable the generation of CI tasks that error if some indicator goes out of bounds, but that is outside the scope of this PR.

console output is in the form 
```
Profiler compile 1 after phase  4:               typer wallClockTime:   13088.2065ms, cpuTime   12609.3750ms, userTime   12500.0000ms, allocatedBytes     651.6394MB, retainedHeapBytes     681.0260MB, gcTime      0ms
Profiler compile 1 after phase  5:              patmat wallClockTime:    2128.9528ms, cpuTime    2062.5000ms, userTime    2062.5000ms, allocatedBytes     132.4393MB, retainedHeapBytes     153.0780MB, gcTime      0ms
Profiler compile 1 after phase  6:      superaccessors wallClockTime:     349.0044ms, cpuTime     328.1250ms, userTime     328.1250ms, allocatedBytes      27.4235MB, retainedHeapBytes      43.8591MB, gcTime      0ms
Profiler compile 1 after phase  7:          extmethods wallClockTime:     130.9848ms, cpuTime      93.7500ms, userTime      93.7500ms, allocatedBytes       5.7003MB, retainedHeapBytes       0.0000MB, gcTime      0ms
Profiler compile 1 after phase  8:             pickler wallClockTime:     371.8387ms, cpuTime     375.0000ms, userTime     375.0000ms, allocatedBytes      23.6769MB, retainedHeapBytes      21.9295MB, gcTime      0ms
Profiler compile 1 after phase  9:            xsbt-api wallClockTime:    2040.5242ms, cpuTime    2031.2500ms, userTime    2031.2500ms, allocatedBytes     314.7573MB, retainedHeapBytes     307.0130MB, gcTime      0ms
Profiler compile 1 after phase 10:     xsbt-dependency wallClockTime:    1410.9184ms, cpuTime    1406.2500ms, userTime    1078.1250ms, allocatedBytes     436.6854MB, retainedHeapBytes     438.5903MB, gcTime      0ms
```

PR #5760, opened 2017-03-07
Add more diagnostics for compiler performance analysis

retronym:
- `-Yprofile` will output the difference between snapshots of GC and CPU
    snapshot data, sourced from platform MBeans
  - This output can be sent to a file with `-Yprofile-destination <filename>`
    By default, output is to console
  - Use `-Yprofile-external-tool` will generate a call to the static method
    `before` and `after` around each of the the given phases. This can be
    used to communicate with an external profiler, such as YourKit, to generate a
    profile for a subset of the compiler.
  - `-Yprofile-run-gc` runs the GC after each phase to help more accurately
     attribute retained heap to a given phase.

### DIFF
--- a/src/compiler/scala/tools/nsc/profile/Profiler.scala
+++ b/src/compiler/scala/tools/nsc/profile/Profiler.scala
@@ -1,0 +1,153 @@
+package scala.tools.nsc.profile
+
+import java.io.{FileWriter, PrintWriter}
+import java.lang.management.ManagementFactory
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.tools.nsc.{Phase, Settings}
+
+object Profiler {
+  def apply(settings: Settings): Profiler =
+    if (!settings.YprofileEnabled) NoOpProfiler
+    else {
+      val reporter = if (settings.YprofileDestination.isSetByUser)
+        new StreamProfileReporter(new PrintWriter(new FileWriter(settings.YprofileDestination.value, true)))
+      else ConsoleProfileReporter
+      new RealProfiler(reporter, settings)
+    }
+}
+
+
+case class ProfileCounters(wallClockTimeNanos: Long, cpuTimeNanos: Long, userTimeNanos: Long, allocatedBytes: Long,
+                           retainedHeapBytes: Long, gcTimeMillis: Long) {
+  def -(that: ProfileCounters): ProfileCounters = {
+    ProfileCounters(
+      this.wallClockTimeNanos - that.wallClockTimeNanos,
+      this.cpuTimeNanos - that.cpuTimeNanos,
+      this.userTimeNanos - that.userTimeNanos,
+      this.allocatedBytes - that.allocatedBytes,
+      this.retainedHeapBytes - that.retainedHeapBytes,
+      this.gcTimeMillis - that.gcTimeMillis)
+  }
+  def updateHeap(heapDetails: ProfileCounters): ProfileCounters = {
+    copy(retainedHeapBytes = heapDetails.retainedHeapBytes)
+  }
+  private def toMillis(ns: Long) = ns / 1000000.0D
+  private def toMegaBytes(bytes: Long) = bytes / 1000000.0D
+
+  def wallClockTimeMillis: Double = toMillis(wallClockTimeNanos)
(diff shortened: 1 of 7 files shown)


retronym:
Rebase of #5758

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -338,6 +338,16 @@ trait ScalaSettings extends AbsScalaSettings
 
   def YstatisticsEnabled = Ystatistics.value.nonEmpty
 
+  val YprofileEnabled = BooleanSetting("-Yprofile-enabled", "Enable profiling.")
+  val YprofileDestination = StringSetting("-Yprofile-destination", "file", "where to send profiling output - specify a file, default is to the console.", "").
+    withPostSetHook( _ => YprofileEnabled.value = true )
+  val YprofileExternalTool = PhasesSetting("-Yprofile-external-tool", "Enable profiling for a phase using an external tool hook. Generally only useful for a single phase", "typer").
+    withPostSetHook( _ => YprofileEnabled.value = true )
+  val YprofileRunGcBetweenPhases = PhasesSetting("-Yprofile-run-gc", "Run a GC between phases - this allows heap size to be accurate at the expense of more time. Specify a list of phases, or *", "_").
```
lrytz, line 346: both `_` and `*` are not valid, should use `all` instead

src/compiler/scala/tools/nsc/profile/Profiler.scala
```diff
@@ -0,0 +1,153 @@
...
+      threadMx.getThreadAllocatedBytes(Thread.currentThread().getId), memoryMx.getHeapMemoryUsage.getUsed,
+      gcMx.foldLeft(0L) { case (sum, bean) => bean.getCollectionTime + sum })
+  }
+  private val overhead = {
+    val s1 = snap
+    val s2 = snap
+    s2 - s1
+  }
+  private def doGC(): Unit = {
+    System.gc()
```
lrytz, line 91: Javadoc says: "When control returns from the method call, the Java Virtual Machine has made a best effort to reclaim space from all discarded objects", I guess that's still true for concurrent GC? Anyway, we just have to be aware that `System.gc` is probably not the most reliable tool.

lrytz (commented):
I started reviewing this today and got sucked into getting per-phase profiling with JFR. This is my WIP branch: https://github.com/lrytz/scala/commit/9d16bb313cac9aef4db36368a9abb3994b2d6479
  - Run the compiler with `-Yprofile-enabled -J-XX:+UnlockCommercialFeatures`
  - In JMC, for the compiler process, open the "MBean Server" console and go to the "Triggers" tab
  - Import the config https://gist.github.com/lrytz/aa0e72e00a3f1fadee92d62f9708dfbf, adjust hard coded paths
  - enable the triggers, recording starts and stops automatically during `typer`

I haven't figured out how to do that from the command line, unfortunately..

SethTisue:
tentatively retargeted for 2.12.3, change it back if that's wrong

src/compiler/scala/tools/nsc/profile/Profiler.scala
mkeskells: the GC before/after was intended to provide some indication of the ratio of allocation vs retained sizes. Generally the information that this tool provide is indicative, and not 100 % reproducible, but with sufficient iteration and post processing of the data can provide a high confidence that a particular PR affected a certain metric

mkeskells: for the record - it also has inaccuracies related to finalization and object graphs that require multiple GC/finalization cycles to reclaim, and also with soft references

mkeskells:
Hi @retronym we have updated the base profiler to support capturing of stats from background threads (needed for https://github.com/scala/scala/pull/5815), like thread CPU time and allocations

This also includes better output control and formatting in https://github.com/rorygraves/scalac_perf/tree/2.12.x_profile2

is it best to take these additonal commits to a new PR or adjust the base of this one?

retronym:
I merged the original PR instead. I'll salvage any useful changes in this PR in a new one.

PR #5761, opened 2017-03-07
Don't use `equals` for comparing java.lang.Double/Float

lrytz:
Fixes https://github.com/scala/scala-dev/issues/329

The `equals` method for java.lang.Double/Float behaves differently than comparing
the `doubleValue`s / `floatValues` for `-0.0`/`0.0`/`NaN`.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -1264,14 +1264,22 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
     def genEqEqPrimitive(l: Tree, r: Tree, success: asm.Label, failure: asm.Label, targetIfNoJump: asm.Label, pos: Position) {
 
       /* True if the equality comparison is between values that require the use of the rich equality
-       * comparator (scala.runtime.Comparator.equals). This is the case when either side of the
+       * comparator (scala.runtime.BoxesRunTime.equals). This is the case when either side of the
        * comparison might have a run-time type subtype of java.lang.Number or java.lang.Character.
-       * When it is statically known that both sides are equal and subtypes of Number of Character,
-       * not using the rich equality is possible (their own equals method will do ok.)
+       *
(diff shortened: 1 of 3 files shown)


sjrd (approved):

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1264,14 +1264,18 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
+       * When it is statically known that both sides are equal and subtypes of Number or Character,
+       * not using the rich equality is possible (their own equals method will do ok), except for
+       * java.lang.Float and java.lang.Double: their `equals` have different behavior around `NaN`
+       * and `-0.0`, see Javadoc (scala-dev#329).
        */
       val mustUseAnyComparator: Boolean = {
-        val areSameFinals = l.tpe.isFinalType && r.tpe.isFinalType && (l.tpe =:= r.tpe)
-        !areSameFinals && platform.isMaybeBoxed(l.tpe.typeSymbol) && platform.isMaybeBoxed(r.tpe.typeSymbol)
+        def isBoxedFloatOrDouble(t: Type) = t.typeSymbol == BoxedFloatClass || t.typeSymbol == BoxedDoubleClass
+        def areSameFinals = l.tpe.isFinalType && r.tpe.isFinalType && (l.tpe =:= r.tpe) && !isBoxedFloatOrDouble(l.tpe)
```
Ichoran: I'm not sure adding extra defs here is advisable.  The first one is only used once (is that intentional?).  The second one is certainly only used once, presumably to get lazy evaluation.  But you can do the same thing by putting a code block after the `&&` below.

Ichoran: (Not that you even need a code block; I'm assuming you wanted a variable name for documentation?)

lrytz: OK, removed the `def`s

PR #5762, opened 2017-03-08
spurious cyclic error in type selection 

adriaanm:
Based on @allisonhb's excellent analysis in https://github.com/milessabin/shapeless/issues/679 (original fix in https://github.com/scala/scala/pull/5656), here's my proposed fix. 

The bug report analysis correctly identified an unexpected cyclic error, which was triggered by implicit search. Implicit search is not expected while type checking the qualifier of a type selection node. Fixing this also removes the cyclic error.

Thus, align `typedSelect` to be more consistent in dealing with `Select` and `SelectFromTypeTree` nodes. The spec is clear on the equivalence between `p.T` (`Select(p, T)`) and `p.type#T` (`SelectFromTypeTree(SingletonTypeTree(p), T)`).

This bug gets even more intriguing, in that it shows that you can sneak a macro call into a path using `selectDynamic`. (See the test in next commit.)

Eventually, we should disable applyDynamic in paths. It wasn't explicitly disallowed, since we assumed the stability check would rule out method calls. However, a macro application will dissolve into its rhs before stability is checked...

Fixes scala/bug#10159

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4869,155 +4869,149 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
        * `qual` is already attributed.
        */
       def typedSelect(tree: Tree, qual: Tree, name: Name): Tree = {
-        val t = typedSelectInternal(tree, qual, name)
-        // Checking for OverloadedTypes being handed out after overloading
-        // resolution has already happened.
-        if (isPastTyper) t.tpe match {
-          case OverloadedType(pre, alts) =>
-            if (alts forall (s => (s.owner == ObjectClass) || (s.owner == AnyClass) || isPrimitiveValueClass(s.owner))) ()
-            else if (settings.debug) printCaller(
-              s"""|Select received overloaded type during $phase, but typer is over.
-                  |If this type reaches the backend, we are likely doomed to crash.
-                  |$t has these overloads:
-                  |${alts map (s => "  " + s.defStringSeenAs(pre memberType s)) mkString "\n"}
-                  |""".stripMargin
-            )("")
-          case _ =>
-        }
-        t
-      }
-      def typedSelectInternal(tree: Tree, qual: Tree, name: Name): Tree = {
-        def asDynamicCall = dyna.mkInvoke(context, tree, qual, name) map { t =>
-          dyna.wrapErrors(t, (_.typed1(t, mode, pt)))
-        }
-
-        val sym = tree.symbol orElse member(qual, name) orElse inCompanionForJavaStatic(qual.tpe.prefix, qual.symbol, name)
-        if ((sym eq NoSymbol) && name != nme.CONSTRUCTOR && mode.inAny(EXPRmode | PATTERNmode)) {
-          // symbol not found? --> try to convert implicitly to a type that does have the required
-          // member.  Added `| PATTERNmode` to allow enrichment in patterns (so we can add e.g., an
-          // xml member to StringContext, which in turn has an unapply[Seq] method)
-          val qual1 = adaptToMemberWithArgs(tree, qual, name, mode)
-          if ((qual1 ne qual) && !qual1.isErrorTyped)
-            return typed(treeCopy.Select(tree, qual1, name), mode, pt)
-        }
-
-        // This special-case complements the logic in `adaptMember` in erasure, it handles selections
-        // from `Super`. In `adaptMember`, if the erased type of a qualifier doesn't conform to the
-        // owner of the selected member, a cast is inserted, e.g., (foo: Option[String]).get.trim).
-        // Similarly, for `super.m`, typing `super` during erasure assigns the superclass. If `m`
-        // is defined in a trait, this is incorrect, we need to assign a type to `super` that conforms
-        // to the owner of `m`. Adding a cast (as in `adaptMember`) would not work, `super.asInstanceOf`
-        // is not a valid tree.
-        if (phase.erasedTypes && qual.isInstanceOf[Super]) {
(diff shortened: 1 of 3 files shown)


ghost:
Cool! I think this is way more sensible than my fix; the equivalence with `SelectFromTypeTree(SingletonType(...), ...)` makes a lot of sense. `ApplyDynamic` ever producing a stable path is very suspicious but seems to have worked for a lot of "tricks" in shapeless and elsewhere... potentially you will be making @milessabin sad. 

Looks a lot nicer though; thanks! I'll close mine out for now.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4849,143 +4849,136 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
-        def asDynamicCall = dyna.mkInvoke(context, tree, qual, name) map { t =>
-          dyna.wrapErrors(t, (_.typed1(t, mode, pt)))
-        }
+        // note: on error, we discard the work we did in type checking tree.qualifier into qual
+        // (tree is either Select or SelectFromTypeTree, and qual may be different from tree.qualifier because it has been type checked)
+        val qualTp = qual.tpe
+        if ((qualTp eq null) || qualTp.isError) setError(tree)
+        else if (name.isTypeName && qualTp.isVolatile)  // TODO: use same error message for volatileType#T and volatilePath.T?
+          if (tree.isInstanceOf[SelectFromTypeTree]) TypeSelectionFromVolatileTypeError(tree, qual)
+          else UnstableTreeError(qual)
```
som-snytt, line 4878: Nested braceless style made me instinctively reach out for something to hold onto.

milessabin:
To clarify, is this fix going to break shapeless's _current_ use of this mechanism?

ghost:
As I understand it, no... @adriaanm mentioned considering that we could stop considering `SelectDynamic` to ever be part of a stable path, even during macro expansion. Obviously that wouldn't come around until 2.13.x... plenty of time to think of a workaround :P

adriaanm:
Indeed. At the *earliest* we would consider this change for 2.13. I can see its utility, though -- not saying we definitely will plug the hole. (From a soundness/IDE perspective, I think we should, but tbd.)

src/compiler/scala/tools/nsc/typechecker/Typers.scala
adriaanm: I really hated my braces as a teenager. I guess it still shows.

adriaanm:
Rebased.

adriaanm:
Re-rebased.

adriaanm:
no idea why cla check is failing, others are green

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4854,142 +4854,137 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
-        t
-      }
-      def typedSelectInternal(tree: Tree, qual: Tree, name: Name): Tree = {
-        def asDynamicCall = dyna.mkInvoke(context, tree, qual, name) map { t =>
-          dyna.wrapErrors(t, (_.typed1(t, mode, pt)))
-        }
+        // note: on error, we discard the work we did in type checking tree.qualifier into qual
+        // (tree is either Select or SelectFromTypeTree, and qual may be different from tree.qualifier because it has been type checked)
+        val qualTp = qual.tpe
+        if ((qualTp eq null) || qualTp.isError) setError(tree)
```
retronym, line 4875: Checking `qual.tpe.isError` for selections from term trees appears to be new. I haven't figured out how that case was handled before. The new formulation (checking early and bailing out) appears sane, but I need to connect a few more dots to feel comfortable.

I haven't found any example where the compiler before/after this change differs in behaviour, so I'm probably just being paranoid. I focussed on error cases.

retronym:
Comparative benchmark results coming soon [here](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=c9ebb13%7Cf9f2cd7&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40&from=1497526056626&to=1498735656626)

retronym:
Reassigning to 2.12.4. While I think the change looks good, the bug is pretty esoteric, and this touches core part of the typechecker, so I'd rather we merge it after the impending 2.12.3 release.

adriaanm:
Rebased. Ping @retronym. Let's decide this week whether we feel this is ok for 2.12.4

retronym:
merge error in the rebase, it would seem:

```
[error] /home/jenkins/workspace/scala-2.12.x-validate-publish-core/src/compiler/scala/tools/nsc/typechecker/Typers.scala:4884: ';' expected but 'if' found.
[error]         val sym = tree.symbol orElse member(qual, name) orElse inCompanionForJavaStatic(qual.tpe.prefix, qual.symbol, name) if ((sym eq NoSymbol) && name != nme.CONSTRUCTOR && mode.inAny(EXPRmode | PATTERNmode)) {
[error]                                                                                                                             ^
[warn] one warning found
[error] one error found
```

adriaanm:
ugh, i've gotten too lazy with IJ's magic "resolve all"

retronym:
[Benchmark](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=0dd574e%7Cbfa2f8b5d%7Caea61e3d&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40&from=1505307538610&to=1505790539307) scheduled.

milessabin:
Let me know if there's anything I can do to help with this ...

retronym (approved):

adriaanm:
As far as I can tell, benchmarks look unaffected -- right, @retronym?

PR #5767, opened 2017-03-09
SI-5621 Missing implicits are supplied by defaults

som-snytt:
Specify that a method parameter that doesn't have
a corresponding argument, whether supplied by name,
position or implicitly, must have a default argument,
which by implication supplies the missing value.

Edit: moved the note to section on default args.

### DIFF
--- a/spec/06-expressions.md
+++ b/spec/06-expressions.md
@@ -320,7 +320,7 @@ would not typecheck.
 
 ### Named and Default Arguments
 
-If an application might uses named arguments $p = e$ or default
+If an application is to use named arguments $p = e$ or default
 arguments, the following conditions must hold.
 
 - For every named argument $p_i = e_i$ which appears left of a positional argument
@@ -330,7 +330,7 @@ arguments, the following conditions must hold.
   argument defines a parameter which is already specified by a
   positional argument.
 - Every formal parameter $p_j:T_j$ which is not specified by either a positional
-  or a named argument has a default argument.
+  or named argument has a default argument.
 
 If the application uses named or default
 arguments the following transformation is applied to convert it into
--- a/spec/04-basic-declarations-and-definitions.md
+++ b/spec/04-basic-declarations-and-definitions.md
@@ -669,6 +669,15 @@ def f(a: Int = 0)(b: Int = a + 1) = b // OK
 f(10)()                               // returns 11 (not 1)
 ```
 
+If an [implicit argument](07-implicits.html#implicit-parameters)
+is not found by implicit search, it may be supplied using a default argument.
+
+```scala
+implicit val i: Int = 2
+def f(implicit x: Int, s: String = "hi") = s * x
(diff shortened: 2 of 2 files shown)


spec/06-expressions.md
```diff
@@ -329,8 +329,8 @@ arguments, the following conditions must hold.
 - The names $x_i$ of all named arguments are pairwise distinct and no named
   argument defines a parameter which is already specified by a
   positional argument.
-- Every formal parameter $p_j:T_j$ which is not specified by either a positional
```
adriaanm: Wouldn't an implicit lacking a default be supplied by implicit search?

som-snytt: I was hoping the language here expresses: each parameter has to have a corresponding argument. That argument could be supplied explicitly (for which there are two flavors of syntax, positional and named args) or implicitly (because the parameter list is marked implicit and an implicit was found). If an argument was not supplied in this way, it must have a default arg.

That's how I read, "every param which is not specified [in these ways] [must have] a default argument."

This is just to clarify that defaults are used only if an (explicit or implicit) arg is not specified. Also, trying to use the minimum of words.

lrytz:
While you're at it, could you fix the typo in line 323: "If an application might uses named arguments"?

spec/06-expressions.md
lrytz: Maybe (I'm not sure) it's wrong to talk about implicit arguments here: the spec here speaks about an application, which means `f(args)`. In this case, no implicits are inserted by the compiler.

So your clarification could be a note below the list of conditions.

som-snytt:
@lrytz In fact, I was about to.

spec/06-expressions.md
som-snytt: Yes, there is no notion of implicit arg here. For another ticket, I was just rereading 6.6 because I'd forgotten what a "function application" is. :1st_place_medal: It is so many things.

Maybe the note should go in 4.6.1 after all; I only just noticed that there is another clarification there about scoping (which bears on another ticket).

Basically, that can be the "what happens if I leave out this arg?" FAQ.

For some reason, during an insomniac night, I thought over here in func apps was the natural place to describe semantics of `f()`.

adriaanm:
I find myself losing the wittiness arms race, so I'll just go with 💯

adriaanm:
PS: 
> Go, oxford comma, go! 

🤣

spec/06-expressions.md
```diff
@@ -330,7 +330,7 @@ arguments, the following conditions must hold.
   argument defines a parameter which is already specified by a
   positional argument.
 - Every formal parameter $p_j:T_j$ which is not specified by either a positional
-  or a named argument has a default argument.
+  or named argument has a default argument.
```
som-snytt, line 333: Sorry, @lrytz haste made waste.

PR #5769, opened 2017-03-09
SI-8969 Accept poly+implicit for assignment syntax

som-snytt:
Follow-up to fb061f22d4c35df626d9651e017820a11f8fe56e
which allowed the type param only.

Reported:
```
scala> object Test {
     |   def a[R](implicit s: List[R]):Int = 0
     |   def a_=[R](v: Int)(implicit s: List[R]) = ()
     | }
```


adriaanm (approved):
Robust software, one case at a time!

som-snytt:
@adriaanm I always suspected you ran the world. In your unassuming way.

PR #5770, opened 2017-03-10
SI-6714 Update assignment preserves apply args

som-snytt:
Previously, extra arg lists to apply were ignored
when transforming to update. In particular, apply
could have an implicit param list, which might be
supplied explicitly. Just keep any and all params
in the expression passed to update. Indices are
always the first arg list.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4754,7 +4754,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
               Select(vble.duplicate, prefix) setPos fun.pos.focus, args) setPos tree.pos.makeTransparent
           ) setPos tree.pos
 
-        def mkUpdate(table: Tree, indices: List[Tree]) =
+        def mkUpdate(table: Tree, indices: List[Tree], argss: List[List[Tree]]) =
           gen.evalOnceAll(table :: indices, context.owner, context.unit) {
             case tab :: is =>
               def mkCall(name: Name, extraArgs: Tree*) = (
@@ -4763,9 +4763,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                   is.map(i => i()) ++ extraArgs
                 ) setPos tree.pos
               )
+              def mkApplies(core: Tree) = argss.foldLeft(core)((x, args) => Apply(x, args))
               mkCall(
                 nme.update,
-                Apply(Select(mkCall(nme.apply), prefix) setPos fun.pos, args) setPos tree.pos
+                Apply(Select(mkApplies(mkCall(nme.apply)), prefix) setPos fun.pos, args) setPos tree.pos
               )
             case _ => EmptyTree
           }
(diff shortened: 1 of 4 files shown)


som-snytt:
I can't wait to see the community build.

test/files/run/t6714.scala
```diff
@@ -0,0 +1,29 @@
...
+  def +(j: Int) = a + j
+}
+
+object Test extends App {
+  def checkA(x: A, y: Int, z: Int) = assert(x == A(y, z))
+  def checkB(x: B, y: Int, z: Int) = assert(x == B(y, z))
+  implicit def ev: Int = 8000
+  val a = A(5, 1)
+  checkA(a(1) = 10, 1, 10)
+  checkA(a(1) += 20, 1, 8020)
```
lrytz: OK, think I can understand this one:
```
a(1) += 20
a(1) = a(1) + 20
a.update(1, a(1) + 20)
a.update(1, a(1)(ev) + 20)
...
```

test/files/run/t6714.scala
```diff
@@ -0,0 +1,29 @@
...
+  val b = B(5, 1)
+  checkB(b(1)(5) += 20, 1, 8025)
```
lrytz: But here I'm lost

```
b(1)(5) += 20
b(1)(5) = b(1)(5) + 20
```
does this become `b.update(x, b(1)(5) + 20)`? is `x` 1 or 5? what happens to the other value?

As far as I can see this is not covered by the spec http://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#assignments. If anything, I think it would be `b(1).update(5, b(1)(5) + 20)` according to the spec (which would fail to compile)?

som-snytt: It's not allowed by the spec. I'd be comfortable failing in the new edge cases, too, per Seth's reaction on the second ticket.

Regular update for `v(1) = rhs` is done by parser. For `v(1) += rhs`, with an implicit arg, so we see `v(1)(imp) += rhs`, do we allow only if the arg was implicitly supplied, in order to enforce a syntactic rewrite? This PR just takes the indices to update from the first arg list and takes the whole LHS for the apply expression. `v.update(1, v(1)(anything)(imp) + rhs)` if there are more arglists.

The current just drops args, which is wrong.

som-snytt: On the [second ticket](https://issues.scala-lang.org/browse/SI-10021?focusedCommentId=75756&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-75756), I said it should just error.

som-snytt:
Pushed a second commit to allow only an implicitly-supplied arglist. I had forgotten that we can detect that condition, which I had noticed once when answering a question on stackoverflow.  I'll squash if that is the right answer.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4754,7 +4754,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
               Select(vble.duplicate, prefix) setPos fun.pos.focus, args) setPos tree.pos.makeTransparent
           ) setPos tree.pos
 
-        def mkUpdate(table: Tree, indices: List[Tree]) =
+        def mkUpdate(table: Tree, indices: List[Tree], argss: List[List[Tree]]) =
```
som-snytt, line 4757: `argss` is one optional arglist in second commit.

lrytz (approved):

PR #5771, opened 2017-03-11
Fix and improve Regex doc

som-snytt:
As reported by @djvulee, there was an error in
the example explaining the confusing behavior
of `MatchIterator`. It's really confusing.

### DIFF
--- a/src/library/scala/util/matching/Regex.scala
+++ b/src/library/scala/util/matching/Regex.scala
@@ -41,11 +41,13 @@ import java.util.regex.{ Pattern, Matcher }
  *  implicitly for strings:
  *
  *  {{{
- *  val date = """(\d\d\d\d)-(\d\d)-(\d\d)""".r
+ *  val date = raw"(\d{4})-(\d{2})-(\d{2})".r
  *  }}}
  *
  *  Since escapes are not processed in multi-line string literals, using triple quotes
  *  avoids having to escape the backslash character, so that `"\\d"` can be written `"""\d"""`.
+ *  The same result is achieved with certain interpolators, such as `raw"\d".r` or
+ *  a custom interpolator `r"\d"` that also compiles the `Regex`.
  *
  *  === Extraction ===
  *  To extract the capturing groups when a `Regex` is matched, use it as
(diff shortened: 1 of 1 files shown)


src/library/scala/util/matching/Regex.scala
```diff
@@ -116,29 +118,41 @@ import java.util.regex.{ Pattern, Matcher }
...
+ *  with alternating calls to `hasNext` and `next`, `hasNext` has the additional
+ *  side effect of advancing the underlying matcher to the next unconsumed match.
+ *  This effect is visible in the `MatchData` representing the "current match".
  *
  *  {{{
- *  val num = """(\d+)""".r
- *  val all = num.findAllIn("123").toList  // List("123"), not List("123", "23", "3")
+ *  val r = "(ab+c)".r
+ *  val s = "xxxabcyyyabbczzz"
+ *  r.findAllIn(s).start    // 3
```
libratiger, line 132: This is a little weird, it works under scala2.12.1, but output the following error under scala2.11.8


>scala> r.findAllIn(s).start
java.lang.IllegalStateException: No match available
  at java.util.regex.Matcher.start(Matcher.java:343)
  at scala.util.matching.Regex$MatchIterator.start(Regex.scala:761)
  ... 32 elided

libratiger:
Looks better now, but there is a bug under 2.11.8 as I commented.

src/library/scala/util/matching/Regex.scala
som-snytt: Yes this is definitely 2.12 only. The fix was last summer.

BTW, I considered mentioning that everyone was bitten by the old behavior, but scaladoc doesn't usually do that. That's for release notes.

libratiger: OK.

I am new to the Scala community, it looks good to me.
I will close my pull request.

src/library/scala/util/matching/Regex.scala
```diff
@@ -116,29 +118,41 @@ import java.util.regex.{ Pattern, Matcher }
  *  while (mi.hasNext) {
  *    val d = mi.next
  *    if (mi.group(1).toInt < 1960) println(s"$d: An oldie but goodie.")
```
SethTisue, line 120: a reference to SI- numbers, I'm sure

SethTisue:
thanks Andrew and Djvu!

libratiger:
Thanks for your merge! @SethTisue

PR #5772, opened 2017-03-12
SI-9745 typedFunction keeps x => f(x)

som-snytt:
To type `x => f(x)`, `typedFunction` uses the expected
type and `f` to type `x`. Previously, it discarded the
result of typechecking `f`, which was a problem in the
case of a `g(e)` where `e` is "erroneous" but rewriteable
to an expr that checks, e.g., `g(i += 1)`.

This commit keeps the rewritten `g(i = i + 1)` instead
of trying to recover the erroneous tree.


som-snytt:
Two tests to add:

```
  // pos, crashes now
  def f(i: Int)(z: Any): Int = ???
  val g = (x: Int, y: Int) => f(42)(x, y)

  // neg, adds arity error which is currently silent
  def f(i: Int)(z: Any): Int = ???
  val g = (x, y) => f(42)(x, y)
```

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -254,7 +254,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
     var context = context0
     def context1 = context
 
-    // for use with silent type checking to when we can't have results with undetermined type params
+    // for use with silent type checking for when we can't have results with undetermined type params
     // note that this captures the context var
     val isMonoContext = (_: Any) => context.undetparams.isEmpty
```
lrytz: unused now

test/files/neg/t9745.check
```diff
@@ -0,0 +1,27 @@
...
+  val g = x => f(x += 1)(x)
+                 ^
+t9745.scala:15: error: missing parameter type
+  val g = x => f(x += 1)(x)
+          ^
+t9745.scala:20: error: type mismatch;
+ found   : (z: Any)Int
+ required: (x$1: ?, x$2: ?)?
+  val g = (x, y) => f(42)(x, y)
+                     ^
```
lrytz, line 20: This one is a bit unfortunate..

lrytz (commented):
The new additional error is unfortunate, but I don't see a way out. Is it right that this version doesn't fix more / other issues than #5187, but just avoids type-checking the same tree twice?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
som-snytt: yup.

test/files/neg/t9745.check
som-snytt: I'm on the fence.

retronym:
I'm well impressed that we've found the spot to fix this. But I'm hesitant to touch this area in 2.12.x. Let's target at 2.13.x and re-rereview. @som-snytt I'm happy to do the busywork if you've expended your exasperation budget on this one...

som-snytt:
Since I have no memory of this change, I'm willing to start from scratch. There's a previous comment where I was on the fence about something, I wonder if it matters which side one comes down on.

som-snytt:
Superseded https://github.com/scala/scala/pull/6007

PR #5774, opened 2017-03-13
SI-884, allow type arguments on patterns.

paulp:
Also offers a way to achieve the goal in SI-6517, seen in the test.

The mechanism is to rewrite casedefs which contain type applications
so that they no longer contain type applications, by pulling the
unapply call into a lazy val which precedes the match. Example using
an extractor which parses a string based on a specified type:
```scala
  expr match {
    case FromString[Int](n)                   => n
    case FromString[(Double, Double)]((a, b)) => (a + b).toInt
  }
```
becomes (with the generated names unique of course)
```scala
  {
    val Scrutinee = expr;
    lazy val TypeArgsPattern1 = FromString.unapply[Int](Scrutinee);
    lazy val TypeArgsPattern2 = FromString.unapply[(Double, Double)](Scrutinee);

    Scrutinee match {
      case _ if !TypeArgsPattern1.isEmpty =>
        val n = TypeArgsPattern1.get
        n
      case _ if !TypeArgsPattern2.isEmpty =>
        val x$1 = TypeArgsPattern2.get
        val a = x$1._1
        val b = x$1._2
        (a + b).toInt
    }
  }
```
This approach means that the various machinery which swirls around
type application should work (e.g. and especially, implicit selection.)

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
@@ -1,0 +1,113 @@
+package scala.tools.nsc
+package ast.parser
+
+class MatchRewriter[G <: scala.tools.nsc.Global](val global: G) {
+  import global._
+
+  /** Entry point. */
+  object rewrite extends Transformer {
+    def apply(t: Tree): Tree = {
+      logResult("Transforming: " + t)(transform(t))
+    }
+    override def transform(t: Tree): Tree = t match {
+      case m: Match => new TranslateMatch(super.transform(m).asInstanceOf[Match]) resultTree
+      case _        => super.transform(t)
+    }
+  }
+
+  object CaseDecompose {
+    /** That's right, the compiler generates a wrong AppliedTypeTree
+     *  instead of a TypeApply tree when there are type args in a pattern.
+     *  "We can rebuild it... we have the technology."
+     *
+     *  Unfortunately fixing it at the source revealed many other
+     *  points in the compiler which depend on things happening as
+     *  they presently do. You, dear reader, are welcome to try it.
+     */
+    def makeTermTree(t: Tree): Tree = t match {
+      case AppliedTypeTree(fun, args) => TypeApply(makeTermTree(fun), args)
+      case Select(qual, name)         => Select(makeTermTree(qual), name.toTermName)
+      case Ident(name)                => Ident(name.toTermName)
+    }
+    def unapply(t: Tree) = t match {
+      case CaseDef(Apply(AppliedTypeTree(fun, targs), patternVars), guard, expr) =>
+        Some((makeTermTree(fun), targs, patternVars, guard, expr))
+      case CaseDef(Apply(fn, patternVars), guard, expr) =>
+        Some((fn, Nil, patternVars, guard, expr))
+      case CaseDef(pat, guard, expr) =>
+        Some((pat, Nil, Nil, guard, expr))
+      case _ =>
+        None
+    }
+  }
+
+  class TranslateMatch(m: Match) {
+    val Match(expr, cases) = m
+
+    /** The list of valdefs to prepend to the rewritten match. */
+    var valDefs = Vector[Tree]()
+
+    /** Memoize the scrutinee expression. */
+    lazy val exprVal = currentUnit.freshTermName("Scrutinee")
+    lazy val expr1   = q"val $exprVal = $expr"
+
+    /** The transformed cases, side effecting valDefs as we go. */
+    val cases1 = cases map (x => (TranslateCaseDefs transform x).asInstanceOf[CaseDef])
+
+    /** The rewritten pattern match, if any type argument patterns were pulled out. */
+    val resultTree = valDefs match {
+      case Seq() => m
+      case _     => q"{ $expr1 ; ..$valDefs ; $exprVal match { case ..$cases1 } }"
+    }
+
+    object TranslateCaseDefs extends Transformer {
+      def apply(cd: CaseDef): CaseDef = transform(cd).asInstanceOf[CaseDef]
+
+      override def transform(t: Tree): Tree = t match {
+        case CaseDecompose(selector, targs, patternVars, guard, rhs) if targs.nonEmpty =>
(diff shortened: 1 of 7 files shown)


stew:
[here's a link to the jira ticket](https://issues.scala-lang.org/browse/SI-884)

src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  class TranslateMatch(m: Match) {
+    val Match(expr, cases) = m
+
+    /** The list of valdefs to prepend to the rewritten match. */
+    var valDefs = Vector[Tree]()
+
+    /** Memoize the scrutinee expression. */
+    lazy val exprVal = currentUnit.freshTermName("Scrutinee")
```
retronym, line 51: Needs a trailing `$` in the prefix:

```
scala> currentUnit.freshTermName("foo")
res0: $r.intp.global.TermName = foo1
```

src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
```diff
@@ -0,0 +1,113 @@
...
+    lazy val expr1   = q"val $exprVal = $expr"
```
retronym, line 52: We don't use quasiquotes in the compiler implementation elsewhere, partly because of the lack of ability to the component trees precisely, partly because of the somewhat bloated expansions, and partly keep one fewer dependency in the bootstrap. So we'd need to roll up our sleeves and do the manual `treeCopy.XXX`-s in their place.

src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
```diff
@@ -0,0 +1,113 @@
...
+  import global._
+
+  /** Entry point. */
+  object rewrite extends Transformer {
+    def apply(t: Tree): Tree = {
+      logResult("Transforming: " + t)(transform(t))
+    }
+    override def transform(t: Tree): Tree = t match {
+      case m: Match => new TranslateMatch(super.transform(m).asInstanceOf[Match]) resultTree
+      case _        => super.transform(t)
```
retronym, line 14: Too much recursion, here.

```scala
class Test {
  1 match {
    case 1 => 1 match {
      case 1 => 1 match { 
        case 1 => 1 match {
          case 1 => 1 match {
            case 1 => 1 match {
              case 1 => 1 match {
                case 1 =>
                case _ => "7"
              }
              case _ => "6"
            }
            case _ => "5"
          }
          case _ => "4"
        }
        case _ => "3"
      }
      case _ => "2"
    }
    case _ => "1"
  }
}
```

```patch
diff --git a/src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala b/src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
index d27ae1fe14..13837be767 100644
--- a/src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/MatchRewriter.scala
@@ -9,9 +9,15 @@ class MatchRewriter[G <: scala.tools.nsc.Global](val global: G) {
     def apply(t: Tree): Tree = {
       logResult("Transforming: " + t)(transform(t))
     }
-    override def transform(t: Tree): Tree = t match {
-      case m: Match => new TranslateMatch(super.transform(m).asInstanceOf[Match]) resultTree
-      case _        => super.transform(t)
+    override def transform(t: Tree): Tree = {
+      t match {
+        case Literal(Constant(s: String)) => println(s)
+        case _ =>
+      }
+      t match {
+        case m: Match => new TranslateMatch(super.transform(m).asInstanceOf[Match]) resultTree
+        case _        => super.transform(t)
+      }
     }
   }
```
Prints:
```
7
7
6
7
6
5
7
6
5
4
7
6
5
4
3
7
6
5
4
3
2
7
6
5
4
3
2
1

**Comment by retronym** [comment]
Overall, I'd prefer to see this translation in the pattern match translation, rather than as a rewrite of the parse tree

**Comment by paulp** [comment]
Yeah. I ported this from a macro, where I don't exactly have the same options. I got it to a non-macro working state but I'm going to defer to @milessabin as to how to advance from here.

**Comment by retronym** [comment]
More work would also be needed to support type apply extractor patterns in sub pattern position

```
scala> object Extractor { def unapply[T](a: Any) = Some(a) }
defined object Extractor

scala> 0 match { case Extractor[Int](x) => x }
res2: Any = 0

scala> (0, 0) match { case (Extractor[Int](x), _) => x }
<console>:12: error: not found: type Extractor
       (0, 0) match { case (Extractor[Int](x), _) => x }
                            ^
```

**Comment by retronym** [comment]
I'm unassigning the 2.12.2 milestone for now as that release is pretty close and this will take some more time to nail down.

**Comment by milessabin** [comment]
> I'm going to defer to @milessabin as to how to advance from here.

Happy to help, but I don't know what you're asking for here ...

**Comment by Blaisorblade** [comment]
Not to complain—I'm extremely happy this is even under the radar and don't want to derail it—but is there a chance for binding type variables in these patterns while looking at this, or to keep that in mind while looking at options?

Example use case:

```scala
trait Exp[T]
case class MapOption[T, U](f: Exp[T => U], arg: Exp[Option[T]]) extends Exp[Option[U]]
def f(term: Exp[T]) = term match {
  case Foo[t, u](f, arg) => ... // doesn't work
  case Foo(f, arg) => ... // works, no way to say `t`
  case term1: Foo[t, u] => // works but clumsy
}
```

paulp:
@milessabin I was referring to our conversation in the comments of [SI-8901](https://issues.scala-lang.org/browse/SI-8901) in which you said "If you open a PR against scala/scala then I'll commit to getting it across the finishing line," which successfully motivated me to take it this far. There are some straightforward improvements I will make based on @retronym's helpful analysis, but they won't add up to the finish line and that is all the time I have allocated for compiler hacking.

paulp:
@Blaisorblade I'll take a look.

milessabin:
@paulp understood.

paulp:
I think this isn't going to fly in its present form and I don't anticipate ever working on it again so I'm closing. Anyone interested is welcome to pick it back up.

PR #5776, opened 2017-03-14
cleanups and clarifications in versions.properties

lrytz:


versions.properties
```diff
@@ -1,27 +1,27 @@
...
+#  - scala-compiler: scala-xml, jline (% "optional")
+#  - scala-library-all: scala-xml, scala-parser-combinators, scala-swing
+# Other usages:
+#  - scala-asm: jar content included in scala-compiler
+#  - jline: shaded with JarJar and included in scala-compiler
+#  - partest: used for running the tests
 scala-xml.version.number=1.0.6
 scala-parser-combinators.version.number=1.0.5
 scala-swing.version.number=2.0.0
-scala-swing.version.osgi=2.0.0
```
lrytz, line 21: this is no longer used, it was only used in the ant build (8247b8f077)

PR #5778, opened 2017-03-14
SI-10225 Either docs compile

som-snytt:
...or they don't.

But without tut support, who knows.

Boy scout aligns stars and updates several
examples that didn't infer types in a way
that compiles.

### DIFF
--- a/src/library/scala/util/Either.scala
+++ b/src/library/scala/util/Either.scala
@@ -10,24 +10,24 @@ package scala
 package util
 
 /** Represents a value of one of two possible types (a disjoint union.)
- *  An instance of Either is either an instance of [[scala.util.Left]] or [[scala.util.Right]].
+ *  An instance of `Either` is an instance of either [[scala.util.Left]] or [[scala.util.Right]].
  *
- *  A common use of Either is as an alternative to [[scala.Option]] for dealing
- *  with possible missing values.  In this usage, [[scala.None]] is replaced
+ *  A common use of `Either` is as an alternative to [[scala.Option]] for dealing
+ *  with possibly missing values.  In this usage, [[scala.None]] is replaced
  *  with a [[scala.util.Left]] which can contain useful information.
  *  [[scala.util.Right]] takes the place of [[scala.Some]].  Convention dictates
- *  that Left is used for failure and Right is used for success.
+ *  that `Left` is used for failure and `Right` is used for success.
  *
- *  For example, you could use `Either[String, Int]` to detect whether a
- *  received input is a String or an Int.
+ *  For example, you could use `Either[String, Int]` to indicate whether a
+ *  received input is a `String` or an `Int`.
  *
  *  {{{
  *  import scala.io.StdIn._
  *  val in = readLine("Type Either a string or an Int: ")
- *  val result: Either[String,Int] = try {
- *      Right(in.toInt)
- *    } catch {
- *      case e: Exception => Left(in)
+ *  val result: Either[String,Int] =
+ *    try Right(in.toInt)
+ *    catch {
+ *      case e: NumberFormatException => Left(in)
  *    }
  *
  *  result match {
@@ -36,13 +36,13 @@ package util
  *  }
  *  }}}
  *
- *  Either is right-biased, which means that Right is assumed to be the default case to
- *  operate on. If it is Left, operations like map, flatMap, ... return the Left value unchanged:
+ *  `Either` is right-biased, which means that `Right` is assumed to be the default case to
+ *  operate on. If it is `Left`, operations like `map` and `flatMap` return the `Left` value unchanged:
  *
  *  {{{
  *  def doubled(i: Int) = i * 2
- *  Right(12).map(doubled) // Right(24)
- *  Left(23).map(doubled)  // Left(23)
+ *  Right(42).map(doubled) // Right(84)
+ *  Left(42).map(doubled)  // Left(42)
  *  }}}
  *
  *  Since `Either` defines the methods `map` and `flatMap`, it can also be used in for comprehensions:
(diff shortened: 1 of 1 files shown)


som-snytt:
It should be possible (lacking tut support) to paste the scaladoc into REPL, but [for this one](https://issues.scala-lang.org/browse/SI-10226).

You should be able to define `-Dscala.repl.continue=" *  "` or similar, to process the stars as a continuation prompt.

src/library/scala/util/Either.scala
```diff
@@ -245,53 +234,49 @@ sealed abstract class Either[+A, +B] extends Product with Serializable {
...
    *
-   * {{{
-   * Right(12).forall(_ > 10) // true
-   * Right(7).forall(_ > 10)  // false
-   * Left(12).forall(_ > 10)  // true
-   * }}}
+   *  {{{
+   *  Right(12).forall(_ > 10) // true
+   *  Right(7).forall(_ > 10)  // false
+   *  Left(12).forall((_: Int) > 10) // true
```
SethTisue: this is a tricky one — it's true because it's a `Left`, not because 12 is greater than 10. `Left(12).forall((_: Int) < 10)` is true, too. maybe this could be made more explicit?

src/library/scala/util/Either.scala
```diff
@@ -425,62 +405,58 @@ object Either {
...
-   *   for (r <- interactWithDB(someQuery).right) yield generateReport(r)
-   * if (report.isRight)
-   *   send(report)
-   * else
-   *   log("report not generated, reason was " + report.left.get)
-   * }}}
+   *  {{{
+   *  // using Option:
+   *  def interactWithDB(x: Query): Option[Result] =
+   *    try {
```
SethTisue: I'd suggest removing the unnecessary extra curly braces here (and below)

src/library/scala/util/Either.scala
```diff
@@ -425,62 +405,58 @@ object Either {
...
+   *  def interactWithDB(x: Query): Either[Exception, Result] =
+   *    try {
+   *      Right(getResultFromDatabase(x))
+   *    } catch {
+   *      case ex => Left(ex)
+   *    }
+   *
+   *   // this will only be executed if interactWithDB returns a Right
+   *   val report = for (r <- interactWithDB(someQuery).right) yield generateReport(r)
+   *   if (report.isRight)
```
SethTisue: we don't want to encourage this style of coding, I think, especially the `report.left.get` that follows. could this be replaced with a pattern match?

SethTisue (approved):

SethTisue:
I've hit "approve", but also suggested some further improvements. Feel free to adopt my suggestions or ignore them and hit "merge", as you see fit.

som-snytt:
@SethTisue improved. Could you re-approve? Do they call you the `Tisue Tracker` at work?

SethTisue:
They've called me worse.

SethTisue:
LGTM

After reading through the whole thing again, the only further issue that occurred to me has to do with the sentence "Since Either defines the methods map and flatMap, it can also be used in for comprehensions". `withFilter` is missing from the list, rightly, but the omission goes unmentioned. wdyt about adding something like "Note that Either does not define `withFilter`, so `if` guards can't be used in `Either`-based comprehensions. That's because no default or empty value can be supplied if the guard is false." 

Also throwing this out there on a "feel free to go ahead and merge anyway" basis.

som-snytt:
@SethTisue Frankly I never use Either, so this was a chance to suss it out, or a missed chance. I, too, didn't like the language about `map`, but I was running on fumes when I did the second pass. The missing `withFilter` seems like a big deal to me. How could there not be a ticket for that? It's not a doc bug, just a regular bug, or improvement if you will. Too bad we can't summon soc to close this out. Maybe if we deprecated Either, he'd come out of the woodwork.

Also, I was 50/50 on whether to use backticks on all mentions of class names. I consulted doc on Option, and added them back in everywhere.

SethTisue:
Hmmm, also, what about this part?

```
// It may be necessary to provide the type of the “missing” value, especially the type
// of the right value for `Left`. Otherwise, without any context that constrains the type,
// it might be inferred as `Nothing`:
```

What does it look like to "provide the type of the “missing” value"? It wasn't obvious to me without experimenting that it doesn't matter if you type-annotate `a` but you must type-annotate `c`.

SethTisue:
> The missing withFilter seems like a big deal to me. How could there not be a ticket for that? It's not a doc bug, just a regular bug, or improvement if you will

well, `filter` is defined on `Either` but it returns an `Option` (which is why it also provides `filterOrElse`); I don't think that possibility is open for `withFilter` because the types wouldn't line up (going off hunch/memory here)? It's hard to know how much to say about this in the doc without getting too deep into the constraints and design decisions.

som-snytt:
Line 56 shows necessary annotation. There are various explicit types in the doc which are merely documentary, and that confused me because it suggests they are required.

Line 106 is an example where I suppressed explicit unnecessary type.

I stopped caring some time ago, only because of personal issues that have affected my world view recently, but my other idea is that colorized example code could do something special with uses of the documented identifier, like bold or a special color. Maybe also a color for values which are typed as a type that is documented.

som-snytt:
OK then, I'll merge for now. But this is just stupid. Even if there's a good reason for it, it's still stupid. I'm not going to advertise it.

```
scala> val rr = for (r <- Right(42).right if r > 50) yield r+1
<console>:11: error: value withFilter is not a member of scala.util.Either.RightProjection[Nothing,Int]
       val rr = for (r <- Right(42).right if r > 50) yield r+1
                                    ^

scala> val rr = for (r <- Right(42) if r > 50) yield r+1
<console>:11: error: value withFilter is not a member of scala.util.Right[Nothing,Int]
       val rr = for (r <- Right(42) if r > 50) yield r+1
                               ^

```

SethTisue:
:shipit:

SethTisue:
thank you for taking this on.

SethTisue:
maybe in the next version we could add some `Left(pad)` and `Left(shark)` jokes

som-snytt:
I did almost write something like, `Left is left alone` by right-bias.

I see the conundrum with filter was in specifying the value to "swap" to when filtering out.

PR #5779, opened 2017-03-15
Update the spec for trailing commas (SIP-27)

dwijnand:
Fixes scala/scala-dev#299.

### DIFF
--- a/spec/01-lexical-syntax.md
+++ b/spec/01-lexical-syntax.md
@@ -553,6 +553,22 @@ but are required to be properly nested.  Therefore, a comment like
 `/* /* */` will be rejected as having an unterminated
 comment.
 
+## Multi-line and trailing commas
+
+Trailing commas are allowed when followed by a newline and a closing
+parentheses, brackets or braces (`)`, `}`, and `]`, respectively), such as:
+
+```scala
+foo(
+  23,
+  "bar",
+  true,
+)
+```
+
+This feature was added with
+[SIP-27](http://docs.scala-lang.org/sips/completed/trailing-commas.html).
+
 ## XML mode
 
 In order to allow literal inclusion of XML fragments, lexical analysis
--- a/spec/01-lexical-syntax.md
+++ b/spec/01-lexical-syntax.md
@@ -556,7 +556,7 @@ comment.
 ## Multi-line and trailing commas
 
 Trailing commas are allowed when followed by a newline and a closing
-parentheses, brackets or braces (`)`, `}`, and `]`, respectively), such as:
+parentheses, square brackets or curly braces (`)`, `]`, and `}`, respectively), such as:
 
 ```scala
 foo(
--- a/spec/01-lexical-syntax.md
+++ b/spec/01-lexical-syntax.md
@@ -555,8 +555,9 @@ comment.
 
 ## Trailing Commas in Multi-line Expressions
 
-If a comma is followed by a newline end and then a right parenthesis, square
-bracket or curly brace then such a trailing comma is ignored. For example:
+If a comma (`,`) is followed immediately, ignoring whitespace, by a newline and
+a closing parenthesis (`)`), bracket (`]`), or brace (`}`), then the comma is
+treated as a "trailing comma" and is ignored. For example:
 
 ```scala
 foo(
--- a/spec/01-lexical-syntax.md
+++ b/spec/01-lexical-syntax.md
@@ -566,9 +566,6 @@ foo(
 )
 ```
 
-This feature was added with
-[SIP-27](http://docs.scala-lang.org/sips/completed/trailing-commas.html).
-
 ## XML mode
 
 In order to allow literal inclusion of XML fragments, lexical analysis
--- a/spec/01-lexical-syntax.md
+++ b/spec/01-lexical-syntax.md
@@ -555,8 +555,9 @@ comment.
 
 ## Multi-line and trailing commas
 
-Trailing commas are allowed when followed by a newline and a closing
-parenthesis, square bracket or curly brace (`)`, `]`, and `}`, respectively), such as:
+If a comma is followed by a line end and then a right parenthesis, square
+bracket or curly brace (`)`, `]`, and `}`, respectively) then such a trailing
+comma is ignored. For example:
 
 ```scala
 foo(
--- a/spec/01-lexical-syntax.md
+++ b/spec/01-lexical-syntax.md
@@ -553,11 +553,10 @@ but are required to be properly nested.  Therefore, a comment like
 `/* /* */` will be rejected as having an unterminated
 comment.
 
-## Multi-line and trailing commas
+## Trailing Commas in Multi-line Expressions
 
-If a comma is followed by a line end and then a right parenthesis, square
-bracket or curly brace (`)`, `]`, and `}`, respectively) then such a trailing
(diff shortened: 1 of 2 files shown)


dwijnand:
Part of the "fix" for scala/scala-dev#299 I should say.

dwijnand:
Disk space issues.

/rebuild

spec/01-lexical-syntax.md
```diff
@@ -553,6 +553,22 @@ but are required to be properly nested.  Therefore, a comment like
 `/* /* */` will be rejected as having an unterminated
 comment.
 
+## Multi-line and trailing commas
+
+Trailing commas are allowed when followed by a newline and a closing
```
som-snytt: delete extra "a": "by a newline and closing parentheses" etc. or say "a closing parenthesis" (singular) etc.

spec/01-lexical-syntax.md
```diff
@@ -553,6 +553,22 @@ but are required to be properly nested.  Therefore, a comment like
...
+```scala
+foo(
+  23,
+  "bar",
+  true,
+)
+```
+
+This feature was added with
+[SIP-27](http://docs.scala-lang.org/sips/completed/trailing-commas.html).
```
som-snytt: Every language feature was added (at some point). I'm not sure an external reference is desirable. If the SIP explains something useful, it should be inlined.

dwijnand: I thought it might be useful, but I'm happy to drop it if it's better not to have it.

spec/01-lexical-syntax.md
dwijnand: You're right. I went with the singular approach. Thanks.

SethTisue:
rather than adding a Changelog entry here, I would suggest removing the Changelog section entirely, since it is unmaintained and the information in it is now only of antiquarian interest

dwijnand:
BTW I can fully appreciate this being seen as too minimal. I'm not well-versed on the spec, so I thought I'd just start with _something_ and go from there.

spec/01-lexical-syntax.md
adriaanm: I suggest bringing this closer to the implementation, which ignores a (single) comma if the following token is a right-parenthesish that itself comes after a line end.

adriaanm: By the way, when looking at the implementation again, I noticed that the only other use of `afterLineEnd` is predicated on `!applyBracePatch` (presumably for the IDE). Should we do that here too?

dwijnand: > I suggest bringing this closer to the implementation, which ignores a (single) comma if the following token is a right-parenthesish that itself comes after a line end.

I can certainly write what you wrote. But I don't understand what difference(s) you're highlighting. Is it the "ignores"? "newline" vs "line end"?

---

> By the way, when looking at the implementation again, I noticed that the only other use of `afterLineEnd` is predicated on `!applyBracePatch` (presumably for the IDE). Should we do that here too?

I had a look at `applyBracePatch` by I can't grok what it does. If you can provide some more context, or even better a test case where you expect the current logic to fall over I'd be happy to look into it.

adriaanm: My suggestion is to remove the ambiguity in the plural "commas" and what it means for a comma to be "allowed".

adriaanm: Regarding the implementation, I'm not up to speed on the IDE's use of the parser, but it looks related to brace patching:

>  return `true` if subsequent end of line handling should be suppressed.

It's overridden in `scala.tools.nsc.ast.parser.Scanners.UnitScanner#applyBracePatch` according to IntelliJ.

dwijnand: Tracked as https://github.com/scala/scala-dev/issues/347

dwijnand:
We've shipped the change officially, so we should probably update the spec sooner rather than later. I'm still happy to revise the wording here, given more feedback.

spec/01-lexical-syntax.md
```diff
@@ -553,6 +553,20 @@ but are required to be properly nested.  Therefore, a comment like
 `/* /* */` will be rejected as having an unterminated
 comment.
 
+## Multi-line and trailing commas
```
som-snytt: Trailing Commas in Multi-line Expressions

spec/01-lexical-syntax.md
```diff
@@ -553,6 +553,20 @@ but are required to be properly nested.  Therefore, a comment like
...
+
+If a comma is followed by a line end and then a right parenthesis, square
+bracket or curly brace (`)`, `]`, and `}`, respectively) then such a trailing
+comma is ignored. For example:
```
som-snytt: specese is "newline" and parentheses are defined on pg 1. Suggestion to follow after walk to school.

som-snytt:
I noticed trailing comma is OK in a block expression. That's potentially confusing because blocks have a result expression. Lexically, it is still simple.
```
scala> :pa

{
 42,
}

res7: Int = 42
```
I was thinking whether trailing comma has any interaction with `nl` tokenization, but I guess trailing comma is allowed only where newline is disabled.

spec/01-lexical-syntax.md
som-snytt: If a comma (`,`) is followed immediately by a newline and a closing parenthesis (`)`), bracket (`]`), or brace (`}`), then the comma is treated as a "trailing comma" and is ignored.

Other specese doesn't say square or curly. The section on newlines uses "treated" in a casual sense. All these closing limiters are in the "parentheses" class defined earlier, but it's confusing to say "a parenthesis-class delimiter", etc.

som-snytt: "immediately" seems useful to say, but this is modulo spaces. I wanted to say "token" somewhere to emphasize this.

spec/01-lexical-syntax.md
```diff
@@ -553,6 +553,20 @@ but are required to be properly nested.  Therefore, a comment like
...
+
+```scala
+foo(
+  23,
+  "bar",
+  true,
+)
+```
```
som-snytt, line 568: I see `foo` is already used in spec examples, although IMHO it's not just casual but slightly undignified.

Was `23` intended to be `42`? There must be a specese style guide somewhere.

som-snytt:
My earlier suggested revision is hidden because of collision with the update. Apparently, there's no way to preserve my backticks without further research.

> If a comma (,) is followed immediately by a newline and a closing parenthesis ()), bracket (]), or brace (}), then the comma is treated as a "trailing comma" and is ignored.

spec/01-lexical-syntax.md
dwijnand: I picked up using 23 from shapeless :)

dwijnand:
@som-snytt I like your reworded explanation so I've updated to that, having included a "ignoring whitespace" clause. WDYT?

I'm not sure what to action from your comment on the block example or your comment on the use of "foo".

Btw:

> you can use `backticks` in quotes `:D`

retronym:
@adriaanm / @som-snytt Do you have further review feedback, or are we ready to merge this?

som-snytt:
LGTM

AFAICR. Also IEITMH (it's early in the morning here).

adriaanm (approved):

PR #5780, opened 2017-03-15
Use a single repository in the bootstrap job

lrytz:
Recently, I changed the bootstap script to publish the "locker" build to scala-release-temp
and only the bootstrapped "quick" build to scala-integration. This commit reverts back to the
previous mechanism where "locker" is published to the same repo (scala-integration) and later
overwritten.

The reason is that we want to use scala-release-temp for publishing integration builds of commits
that are not (yet) merged to scala/scala. Such builds are useful for preliminary testing /
benchmarking of PRs / development revisions. This means that we want to add scala-release-temp
as a resolver in the community build and the benchmark job. If we have same-named releases in
both repos (one "locker", one "quick", so they are not the same), we would not know which one
is being picked up.

If we want to avoid the overwriting in the future, we could work on a solution that sets a different
version number for "locker", but we have to be careful when building the modules; maybe setting the
scalaBinaryVersion would be enough.

### DIFF
--- a/scripts/jobs/integrate/bootstrap
+++ b/scripts/jobs/integrate/bootstrap
@@ -3,7 +3,7 @@
 # Script Overview
 #  - determine scala version
 #  - determine module versions
-#  - build minimal core (aka locker) of Scala, use the determined version number, publish to scala-release-temp
+#  - build minimal core (aka locker) of Scala, use the determined version number, publish to scala-integration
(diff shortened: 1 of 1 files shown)


lrytz:
(even without merging this, the change is already in place, the intermediate script was adjusted: https://github.com/scala/scala-jenkins-infra/commit/2c72708229506c6906ca76e696804763f42dde79)

PR #5782, opened 2017-03-17
Improve serializability of lambdas that call methods in module

retronym:
Depending on how it is written, a reference to an member of an
enclosing module can be either be qualified by `This(moduleClass)`
or by `Ident(module)`. The former was causing a unnecsseary capture
of the module class.

This commit substitutes eligible `this` references when lifting
the closure body into a method.

Fixes scala/bug#10233

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/TreeGen.scala
+++ b/src/compiler/scala/tools/nsc/ast/TreeGen.scala
@@ -308,8 +308,17 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     val useMethodParams = new TreeSymSubstituter(fun.vparams.map(_.symbol), methParamSyms)
     // we're now owned by the method that holds the body, and not the function
     val moveToMethod = new ChangeOwnerTraverser(fun.symbol, methSym)
+    def substThisForModule(tree: Tree) = {
+      // Rewrite This(enclModuleClass) to Ident(enclModuleClass) to avoid unnecessary capture of the module
+      // class, which might hamper serializability.
+      //
+      // Analagous to this special case in ExplicitOuter: https://github.com/scala/scala/blob/d2d33ddf8c/src/compiler/scala/tools/nsc/transform/ExplicitOuter.scala#L410-L412
+      // that understands that such references shouldn't give rise to outer params.
+      val enclosingStaticModules = owner.enclClassChain.filter(x => !x.hasPackageFlag && x.isModuleClass && x.isStatic)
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -308,8 +308,17 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
...
     // we're now owned by the method that holds the body, and not the function
     val moveToMethod = new ChangeOwnerTraverser(fun.symbol, methSym)
+    def substThisForModule(tree: Tree) = {
+      // Rewrite This(enclModuleClass) to Ident(enclModuleClass) to avoid unnecessary capture of the module
+      // class, which might hamper serializability.
+      //
+      // Analagous to this special case in ExplicitOuter: https://github.com/scala/scala/blob/d2d33ddf8c/src/compiler/scala/tools/nsc/transform/ExplicitOuter.scala#L410-L412
+      // that understands that such references shouldn't give rise to outer params.
+      val enclosingStaticModules = owner.enclClassChain.filter(x => !x.hasPackageFlag && x.isModuleClass && x.isStatic)
+      enclosingStaticModules.foldLeft(tree)((tree, moduleClass) => tree.substituteThis(moduleClass, gen.mkAttributedIdent(moduleClass.sourceModule)) )
```
retronym, line 318: This could be fused into a single transform with some effort to generalize `ThisSubstituter` and `SubstThisMap`. That would reduce the overhead for lambdas owned by a chain of static modules.

adriaanm:
Sorry for the delay. I'm okay with the focussed fix, but am wondering whether this is something we could do more generally in lambdalift?

retronym:
That sounds like a good idea, I'll give it a shot.

retronym:
Looking at this again, I think the right place to fix this is actually in `Delambdafy`. I can't quite figure out the how to keep all the plates spinning at the same time, though. I'll come back to this next week...

andrewsmartin:
@retronym Any updates on this?

retronym:
My current thinking is to reconsider the patch as it stands for inclusion in 2.12.3, and revisit this afterwards to see if a cleaner solution is possible. @adriaanm WDYT?

adriaanm:
Ok with me if there's a new ticket for follow up.

retronym:
Tracking the revisit/refactor as https://github.com/scala/scala-dev/issues/401

lrytz:
Could add tests around non-local modules

retronym:
> Could add tests around non-local modules

Done

lrytz (approved):

PR #5786, opened 2017-03-20
Fix Greek Mythology references in compiler package

janekdb:
Improve documentation in some packages and correct some typos in these directories,

 - compiler
 - library

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -161,7 +161,7 @@ trait Implicits {
   }
 
   /* Map a polytype to one in which all type parameters and argument-dependent types are replaced by wildcards.
-   * Consider `implicit def b(implicit x: A): x.T = error("")`. We need to approximate debruijn index types
+   * Consider `implicit def b(implicit x: A): x.T = error("")`. We need to approximate de Bruijn index types
    * when checking whether `b` is a valid implicit, as we haven't even searched a value for the implicit arg `x`,
    * so we have to approximate (otherwise it is excluded a priori).
    */
@@ -358,8 +358,8 @@ trait Implicits {
     val undetParams = if (isView) Nil else context.outer.undetparams
     val wildPt = approximate(pt)
 
-    private val runDefintions = currentRun.runDefinitions
-    import runDefintions._
+    private val stableRunDefsForImport = currentRun.runDefinitions
+    import stableRunDefsForImport._
 
     def undet_s = if (undetParams.isEmpty) "" else undetParams.mkString(" inferring ", ", ", "")
     def tree_s = typeDebug ptTree tree
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -1348,7 +1348,7 @@ trait Namers extends MethodSynthesis {
 
       // Add a () parameter section if this overrides some method with () parameters
       val vparamSymssOrEmptyParamsFromOverride =
-        if (overridden != NoSymbol && vparamSymss.isEmpty && overridden.alternatives.exists(_.info.isInstanceOf[MethodType])) ListOfNil // NOTEL must check `.info.isInstanceOf[MethodType]`, not `.isMethod`!
+        if (overridden != NoSymbol && vparamSymss.isEmpty && overridden.alternatives.exists(_.info.isInstanceOf[MethodType])) ListOfNil // NOTE: must check `.info.isInstanceOf[MethodType]`, not `.isMethod`!
         else vparamSymss
 
       val methSig = deskolemizedPolySig(vparamSymssOrEmptyParamsFromOverride, resTp)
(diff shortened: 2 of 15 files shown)


SethTisue:
(CI failure is real)

janekdb:
Fix this
```
[error] /home/jenkins/workspace/scala-2.12.x-validate-publish-core/src/compiler/scala/tools/nsc/typechecker/Implicits.scala:363: overriding value runDefinitions in class Typer of type Implicits.this.global.definitions.RunDefinitions;
[error]  value runDefinitions has weaker access privileges; it should not be private
[error]     private val runDefinitions = currentRun.runDefinitions
[error]                 ^
[error] one error found
```

janekdb:
Compile error fixed. Should be plain sailing here on in.

image

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -360,8 +360,8 @@ trait Implicits {
     val undetParams = if (isView) Nil else context.outer.undetparams
     val wildPt = approximate(pt)
 
-    private val runDefintions = currentRun.runDefinitions
-    import runDefintions._
+    private val currentRunRunDefinitions = currentRun.runDefinitions
+    import currentRunRunDefinitions._
```
Ichoran: Ow.  That is not a very friendly name.  If you're going to use so many letters, maybe at least explain why this whole thing is necessary `val stableRunDefsForImport = currentRun.runDefinitions`

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -1348,7 +1348,7 @@ trait Namers extends MethodSynthesis {
 
       // Add a () parameter section if this overrides some method with () parameters
       val vparamSymssOrEmptyParamsFromOverride =
-        if (overridden != NoSymbol && vparamSymss.isEmpty && overridden.alternatives.exists(_.info.isInstanceOf[MethodType])) ListOfNil // NOTEL must check `.info.isInstanceOf[MethodType]`, not `.isMethod`!
+        if (overridden != NoSymbol && vparamSymss.isEmpty && overridden.alternatives.exists(_.info.isInstanceOf[MethodType])) ListOfNil // NOTE: Must check `.info.isInstanceOf[MethodType]`, not `.isMethod`!
```
Ichoran: `Must` should be `must` as it is a sentence fragment.  (If we're going to be really pedantic.)

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
SethTisue: daDoRunRunDefinitions http://1.bp.blogspot.com/-xLVS4NExYCU/U9xUfmsNVcI/AAAAAAAABlQ/u6zx6uPJLis/s1600/shaun+4.jpg

dwijnand: Ya running and ya running. But ya can't run away from yourself.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
janekdb: Now lowercase matching the majority case,

```
$ grep -rh -oE '(NOTE:.*)' src/|cut -c1-7|sort|uniq -c
      1 NOTE: `
      1 NOTE: *
      2 NOTE: a
      1 NOTE: A
      1 NOTE: b
     11 NOTE: c
      2 NOTE: C
      1 NOTE: d
      2 NOTE: e
      1 NOTE: E
      3 NOTE: f
      2 NOTE: g
      1 NOTE: h
      1 NOTE: i
      2 NOTE: I
      2 NOTE: m
      2 NOTE: o
      1 NOTE: q
      2 NOTE: r
      1 NOTE: s
      1 NOTE: S
      6 NOTE: t
      1 NOTE: T
      3 NOTE: U
      1 NOTE: V
      4 NOTE: w
      1 NOTE: W
```

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
janekdb: @Ichoran - thanks for thinking about this. Improvement applied!

SethTisue (approved):
on behalf of the Nicolaas Govert de Bruijn estate, I thank you.

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -117,13 +117,13 @@ trait TypeDiagnostics {
    */
   final def exampleTuplePattern(names: List[Name]): String = {
     val arity = names.length
-    val varPatterNames: Option[List[String]] = sequence(names map {
+    val varPatternNames: Option[List[String]] = sequence(names map {
```
l0rinc, line 120: minor: maybe remove `var` prefix if it's a `val` :)
```scala
val patternNames
```

l0rinc (approved):

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
janekdb: In context I think the `val` is about variables so it's okay to leave as-is,

```
    val varPatterNames: Option[List[String]] = sequence(names map {
      case name if nme.isVariableName(name) => Some(name.decode)
      case _                                => None
    })
```
WDYT?

l0rinc (approved):

janekdb:
Fixed conflicts.

l0rinc (approved):

SethTisue:
thanks! looking forward to the PR that straightens out the scalac situation w/r/t qabbalism, the medieval era, etc.

PR #5787, opened 2017-03-20
SI-10240 Better pos for infix type arg

som-snytt:
The range position for the type apply in `1 ->[Int] 2`
should include the last type arg.

It would be nice to include the right bracket.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -824,7 +824,10 @@ self =>
 
       def mkSelection(t: Tree) = {
         def sel = atPos(opPos union t.pos)(Select(stripParens(t), op.encode))
-        if (targs.isEmpty) sel else atPos(left.pos)(TypeApply(sel, targs))
+        if (targs.isEmpty) sel
+        else atPos(sel.pos union targs.last.pos withPoint sel.pos.point) {
+          TypeApply(sel, targs)
+        }
       }
       def mkNamed(args: List[Tree]) = if (isExpr) args map treeInfo.assignmentToMaybeNamedArg else args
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -824,7 +824,10 @@ self =>
 
       def mkSelection(t: Tree) = {
         def sel = atPos(opPos union t.pos)(Select(stripParens(t), op.encode))
-        if (targs.isEmpty) sel else atPos(left.pos)(TypeApply(sel, targs))
+        if (targs.isEmpty) sel
+        else atPos(sel.pos union targs.last.pos withPoint sel.pos.point) {
```
adriaanm: `sel` is a `def` (for `def`ensive tree duplication?) I suggest turning it into a `val`, along with its position, so that both are evaluated once

adriaanm (changes requested):

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
som-snytt: Good eyes! I guess those spectacles are a recent prescription.

som-snytt:
I wish there was a beer called `Mr Moors`.

retronym (approved):

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
som-snytt: sorry, next time!

som-snytt: The follow-up was the second commit, this diff was outdated.

PR #5788, opened 2017-03-20
remove/privatize deprecated stuff from collection.concurrent.TrieMap

SethTisue:
so we can have a deprecation-free parallel collections module

reference: https://github.com/scala/scala-parallel-collections/pull/16


SethTisue:
review by @Ichoran

lrytz (approved):

PR #5790, opened 2017-03-20
unbundle scala-library-all, scala-swing, scala-parser-combinators

SethTisue:


SethTisue:
there is some justification on the discontinuance of scala-library-all at https://github.com/scala/scala-dev/issues/323#issuecomment-286183284

as for scala-swing and scala-parser-combinators, they were modularized in Scala 2.11 but bundled to ease the transition. Scala 2.12 kept them, because in general we didn't remove things from 2.12. but now it's time. ultimately, the Scala Platform is intended to pick up the slack in some form.

readers of e.g. _Programming in Scala_ (Odersky, Venners & Spoon) will need to figure out how to obtain and add the dependencies. so it goes.

SethTisue:
note that scala-xml is not included in this because (as per https://issues.scala-lang.org/browse/SI-9560) work would be needed to cut the remaining (pretty thin) dependency of Scaladoc on scala-xml. it would be wonderful if someone volunteered to tackle that.

dwijnand (approved):
Looks complete (and therefore good) to me.

> readers of _Programming in Scala_ (Odersky, Venners & Spoon) will need to figure out how to obtain and add the dependencies. so it goes

I'm not sure when the UnPickler jumps into action, but it looks like it tries to give guidance:

https://github.com/scala/scala/blob/f77139532d5b6030eaa69a6ff9ed07e7ea4a4dfe/src/reflect/scala/reflect/internal/pickling/UnPickler.scala#L222-L224

SethTisue:
> UnPickler [...] looks like it tries to give guidance

good eye. added a commit that adds scala-parallel-collections to the list

build.sbt
```diff
@@ -778,7 +760,7 @@ lazy val scalaDist = Project("scala-dist", file(".") / "target" / "scala-dist-di
     ),
     publishArtifact in (Compile, packageSrc) := false
   )
-  .dependsOn(libraryAll, compiler, scalap)
+  .dependsOn(library, reflect, compiler, scalap)
```
lrytz, line 763: Does this remove xml, swing and parser combinators from the default repl (in the distribution archive)? I guess it's fine, but would be finer if we had a good way to add things to the default repl's classpath..

versions.properties
```diff
@@ -16,9 +16,6 @@ scala.binary.version=2.13.0-M1
 
 # external modules shipped with distribution, as specified by scala-library-all's pom
```
lrytz: update comment

build.sbt
dwijnand:
```scala
$ scala -cp \
> ~/.ivy2/cache/org.scala-lang.modules/scala-xml_2.12/bundles/scala-xml_2.12-1.0.6.jar:\
> ~/.ivy2/cache/org.scala-lang.modules/scala-swing_2.12/bundles/scala-swing_2.12-2.0.0.jar:\
> ~/.ivy2/cache/org.scala-lang.modules/scala-parser-combinators_2.12/bundles/scala-parser-combinators_2.12-1.0.5.jar
cat: /release: No such file or directory

scala> import scala.xml._
import scala.xml._

scala> import scala.swing._
import scala.swing._

scala> import scala.util.parsing.combinator._
import scala.util.parsing.combinator._
```

... with the caveat that you need to (a) have those jars locally and (b) know the full path (that it's org.scala-lang.modules, that the it's a "bundle" jar..)

SethTisue: > Does this remove swing and parser combinators from the default repl

yes

as for making it easier again to access them, I would think that would happen in the context of https://github.com/scala/scala-dev/issues/326

> Does this remove scala-xml from the default repl

no

```
~/scala.213/build/pack % bin/scala
Welcome to Scala 2.13.0-20170321-002852-96c02a3 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> <i>foo</i>
res0: scala.xml.Elem = <i>foo</i>
```

versions.properties
SethTisue: fixed

lrytz:
@SethTisue there's a merge conflict in `versions.properties`.

Is everyone fine with removing removing xml/swing/parser-combinators from the standard distribution?

SethTisue:
> there's a merge conflict

fixed

SethTisue:
after this is merged, changes will be needed in the scala-dist and scala-dist-smoketest repos

adriaanm:
LGTM too. Could someone make the changes in the distro repos and coordinate merging with this PR?

lrytz:
Or we go for YOLO and deal with the fallout, the 2.13.x build is broken anyway (osgi tests failing)

SethTisue:
https://github.com/scala/scala-dist/pull/177

SethTisue:
https://github.com/scala/scala-dist-smoketest/pull/14

SethTisue:
> Or we go for YOLO

TIL California has a whole Yolo County. wonder what housing prices are like there.

lrytz:
I think this should fix the boostrap failure introduced by https://github.com/scala/scala/pull/5810. Merging now to give it a try.

lrytz:
https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-release-main/164/console

lrytz:
The bootstrap job passed: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-bootstrap/318/

PR #5791, opened 2017-03-21
remove a bunch of old cruft

SethTisue:
the horror, the horror.


SethTisue:
25048bc is the doozy here. this stuff clutters up my `git grep` output constantly

dwijnand (approved):
:+1:

You're going to jump 7 positions by deletions with this one, curtesy of test/disabled/coder/dict.txt. :-)

lrytz:
agreed on 25048bc, i would even be in for `pending`.

SethTisue:
> i would even be in for pending.

I didn't include that in the first cut because it isn't quiiiiiite as moldy and ancient as the others. But yes, I would definitely be in favor of axing pending too, if others agree as well.

SethTisue:
one does sometimes see comments in the code like ` !!! Make pending/run/t5866b.scala work`, but even after wholesale removal, it doesn't take wizard-level git skills to find a particular file like that in the history

I went ahead and provisionally added an additional commit that removes `pending`.

SethTisue:
Adriaan likes it (source: personal communication)

SethTisue:
(there were no objections at the team meeting)

SethTisue:
> You're going to jump 7 positions by deletions with this one

(it doesn't actually work that way, it turns out, it's still by number of commits, not lines deleted. I don't understand it exactly)

SethTisue:
> You're going to jump 7 positions by deletions with this one

screen shot 2017-03-21 at 2 41 38 pm

🏅

PR #5792, opened 2017-03-21
SI-7936 access protected members in package-protected class

lrytz:
A super access to a protected member in a package-protected class is allowed through
an intermediate public class. The fix for SI-5162 actually introduced an error for this
case, because the scala compiler would generate an INVOKESPECIAL with the package-protected
class as receiver, which is illegal. However, we can use the public intermediate class in
the invocation signature.

This is very similar to SI-4283.


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4882,8 +4882,6 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           }
           NoSymbol
         }
-        if (phase.erasedTypes && qual.isInstanceOf[Super] && tree.symbol != NoSymbol)
-          qual setType tree.symbol.owner.tpe
```
lrytz: This code is almost 12 years old.. https://github.com/scala/scala/commit/2a5f62338c#diff-b0ec9c916463796fd7ff1d2ed73d46e2L810

test/files/jvm/t4283/Test.scala
```diff
@@ -1,4 +0,0 @@
-
-object Test extends App {
```
lrytz, line 2: this test was removed because a more exhaustive version of it exists in https://github.com/scala/scala/tree/2.12.x/test/files/run/t4283

lrytz:
this took longer than expected, i hope i covered all cases now.. 🤞

lrytz:
OK, while I think it's going in the right direction, the kind of errors that are showing while sorting things out ask for not rushing this in 2.12.2 :) currently `VerifyError: Bad <init> method call`

milessabin:
@lrytz is it possible you're hitting the same problem I ran in to [here](https://github.com/milessabin/si2712fix-plugin/blob/master/project/FixSuperCtorCall.scala)?

lrytz:
I've got some greens now. A nice side-effect of this fix is that it reduces the cases where one has to manually provide a Java interface as direct parent to allow emitting a super call. For example

```
// Java
interface T { default int f() { return 1; } }

// Scala
trait U extends T
class C extends U { def t = super.f }
```

now compiles using `invokespecial U.f`.

There are still situations where a parent needs to be provided, for example

```
// Java
interface T { default int f() { return 1; } }

// Scala
trait U extends T { override def f = super.f + 1 }
class C extends U
```

doesn't compile: the super accessor is implemented as `invokespecial T.f`, and `U.f` would not work, so `T` needs to be added as a direct parent.

retronym:
This LGTM on a first pass review. Needs a rebase and I also need to try it out in anger.

retronym (approved):

lrytz:
Rebased

lrytz:
Should run this through a community build.

PR #5793, opened 2017-03-21
Respell elidable checking function and make more similar to namesake

janekdb:
The layout and message edits bring out the similarity of the methods.


SethTisue:
perhaps @som-snytt would like to work on adding an `isRisible` method

som-snytt:
The better joke was "Elision fields."

lrytz (approved):

PR #5795, opened 2017-03-22
Remove unused 2.8.0 Scala version val

janekdb:


adriaanm (approved):

dwijnand (approved):
LGTM, too.

https://github.com/search?l=Scala&q=twoDotEight&type=Code&utf8=%E2%9C%93

22 Scala results, and they all look like copies of this file

PR #5796, opened 2017-03-22
Fix creation of negative Durations

Jasper-M:
See https://groups.google.com/forum/#!topic/scala-user/I2Yd3KmoTc4

Fixes scala/bug#9949

### DIFF
--- a/src/library/scala/concurrent/duration/Duration.scala
+++ b/src/library/scala/concurrent/duration/Duration.scala
@@ -125,7 +125,7 @@ object Duration {
     else if (nanos > Long.MaxValue || nanos < Long.MinValue)
       throw new IllegalArgumentException("trying to construct too large duration with " + nanos + "ns")
     else
-      fromNanos((nanos + 0.5).toLong)
+      fromNanos(nanos.round)
   }
 
   private[this] final val  µs_per_ns = 1000L
--- a/test/files/jvm/duration-tck.scala
+++ b/test/files/jvm/duration-tck.scala
@@ -202,4 +202,12 @@ object Test extends App {
   ((2 seconds fromNow).timeLeft: FiniteDuration) < 4.seconds mustBe true
   val finite3: FiniteDuration = 3.5 seconds span
 
+  // scala/bug#9949
+  Duration(-1.0, DAYS) mustBe Duration(-1, DAYS)
+  Duration("-10 s").toNanos mustBe -10000000000L
+  Duration(1.0, DAYS) mustBe Duration(1, DAYS)
+  Duration("10 s").toNanos mustBe 10000000000L
+  
+  // scala/bug#10320
+  Duration("6803536004516701ns").toNanos mustBe 6803536004516701L
 }


test/files/run/duration_negative.scala
```diff
@@ -0,0 +1,14 @@
...
+  // result in the same Duration
+  assert(Duration(-1, DAYS) == Duration(-1.0, DAYS))
+  
+  // A Duration of -10 seconds should be -10000000000 long
+  assert(Duration("-10 s").toNanos == -10000000000l)
+  
+  // Positive Durations should still work
+  assert(Duration(1, DAYS) == Duration(1.0, DAYS))
+  assert(Duration("10 s").toNanos == 10000000000L)
+}
```
dwijnand: Please could you convert this to a unit test instead?

Jasper-M: Sure. I wasn't sure where to test it and this was the only place I found other tests for Duration.

test/junit/scala/concurrent/duration/DurationTest.scala
```diff
@@ -0,0 +1,27 @@
+package scala.concurrent.duration
```
viktorklang: IMO this test should instead of being its own file, be added to [duration-tck](https://github.com/scala/scala/blob/2.12.x/test/files/jvm/duration-tck.scala)

dwijnand: Damn, we're sending mixed signals now :-/

@viktorklang I suggested creating a unit test, because this is pure standard library testing, so it doesn't need partest. See https://github.com/scala/scala-dev/issues/295 and https://github.com/scala/scala-partest/issues/75.

viktorklang: FWIW duration-tck is also pure standard library testing. I'd rather keep related tests together so that migration (if needed) can be done wholesale.

Jasper-M: I agree the tests fit better in the existing set of tests for Duration, where they can eventually be migrated together with the others. Shall we go with that option then?
I have to say finding the right place to test this functionality was definitely the hardest part of this contribution.

dwijnand: Yep. Go with Viktor's suggestion of putting them in duration-tck.scala. Sorry for misguiding you. And thank you for the contribution.

viktorklang: @Jasper-M If you feel adventurous, I guess you could tackle porting the duration-tck to JUnit :)

Jasper-M: Not that I don't love JUnit, but I think I'll keep that for a potential future pull request :)

viktorklang: That's OK, I just had to ask. ;)

src/library/scala/concurrent/duration/Duration.scala
```diff
@@ -124,6 +124,8 @@ object Duration {
       Undefined
     else if (nanos > Long.MaxValue || nanos < Long.MinValue)
       throw new IllegalArgumentException("trying to construct too large duration with " + nanos + "ns")
+    else if (nanos < 0)
+      fromNanos((nanos - 0.5).toLong)
```
l0rinc: Is this a `Double` to `Long` conversion problem at that scale?
Wouldn't:
```scala
fromNanos(nanos.toLong)
```
fix the issue?

(ps. I get `Error: Could not find or load main class Test`, not sure how to run this test...)

test/files/jvm/duration-tck.scala
```diff
@@ -202,4 +202,9 @@ object Test extends App {
   ((2 seconds fromNow).timeLeft: FiniteDuration) < 4.seconds mustBe true
   val finite3: FiniteDuration = 3.5 seconds span
 
+  // test precision of Duration.apply
```
l0rinc: what about `+0.0` and `-0.0`?

src/library/scala/concurrent/duration/Duration.scala
l0rinc: @adriaanm, if `Double` is unstable for big values (`~10^13`), shouldn't we disallow `nanos` as a `Double`, if its sole purpose seems to be to infinite `Duration` (I'm sure `Long.MaxValue` nanos is pretty close to infinite anyway, i.e. `300` years)?

l0rinc (changes requested):
this ±½ is very fishy, you might have discovered a bug

src/library/scala/concurrent/duration/Duration.scala
Jasper-M: `toLong` truncates the decimal part. The `+ 0.5` part is meant to round the decimal number to the nearest integer number. `Duration(0.5, NANOSECONDS) == Duration(1, NANOSECONDS)` and `Duration(0.4, NANOSECONDS) == Duration(0, NANOSECONDS)`. Not sure though if using Double over Long is necessary.

Jasper-M: I have to add that apparently `java.lang.Math.round(-0.5) == 0` and `java.lang.Math.round(-0.51) == -1`, while in this pull request `Duration(-0.5, NANOSECONDS) == Duration(-1, NANOSECONDS)`. So if this is meant to emulate standard java.lang rounding, just using `java.lang.Math.round` or `scala.math.round` might be better.

l0rinc: If we want to truncate, we shouldn't do operations on a double: it's probably the reason for the loss of precision.
I personally don't think `0.5ns` should be rounded to `1ns`, but we could use `math.rint` instead, if we want to round, i.e.
```scala
else
  fromNanos(math.rint(nanos).toLong)
```
which works for negatives the same way as for positives.

l0rinc: If `0.5ns` == `-0.5ns` == `0ns` we could simply use:
```scala
else
  fromNanos(nanos.toLong)
```

l0rinc: Opinions @dwijnand, @viktorklang?

Jasper-M: I don't really have an opinion on exactly how and where to round, but I agree that `Duration("-1 d")` should be `-1 day` and not `-86399999999999 nanoseconds`. Whether everything gets rounded down, or gets scala.math.rounded, or something else, I have no opinion on.

Blaisorblade:
Is this PR for https://github.com/scala/bug/issues/9949? Didn't find a link.

src/library/scala/concurrent/duration/Duration.scala
Blaisorblade: (double+0.5).toLong is broken anyway on a few `Double`s which `round` handles correctly, as @melrief reported on GitHub. Using `round` should fix the original issue as well, though I didn't test the integrated version.

See https://github.com/scala/scala/pull/5913 and https://github.com/scala/bug/issues/10320

Jasper-M:
Didn't even know there was an existing issue for that.

src/library/scala/concurrent/duration/Duration.scala
Blaisorblade: Also: IIUC, scala/bug#10320 only affects `Double` which uses all of the available 52 bits of mantissa. More bugs might lurk elsewhere. I'm no IEEE-754 expert, but let's maybe not try to reimplement `round`, at least not without significantly more coverage. Especially not when we're taking the effort to parse a String anyway.

Jasper-M: @Blaisorblade My first attempt was trying to stay in line with the existing code which I assumed had a reason not to `round`. But I think you're right, especially if there are bugs in `(double+0.5).toLong`, that `round` is the only reasonable option here.

retronym:
@Jasper-M If you agree with @Blaisorblade's suggestion (your comments above lead me to believe you do), could you please update this PR with his suggestion to use `.round`?

Jasper-M:
Yes, sorry, I was planning to but forgot.

Jasper-M:
@retronym Fixed to `round`, added test for scala/bug#10320 and added links to the bug tickets.

SethTisue:
thanks Jasper (and reviewers)!

PR #5798, opened 2017-03-22
Use the correct Context in AmbiguousImplicitError

milessabin:
`context0` is the `Context` passed as an argument to `AmbiguousImplicitError` whereas `context` is the `var context` inherited from `Typer` via the self type of `ImplicitSearch`.

I believe that the intention was that the argument, `context0`, be used as the target for `issueAmbiguousError` but that the error was missed because the only current call site passes in the typer context anyway.

This has tripped me up now, because I've been attempting to use `AmbiguousImplicitError` with a different context (with a different value for `ambiguousErrors`) and it has been completely ignored.

All (par)test pass.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -1293,7 +1293,7 @@ trait ContextErrors {
           case _ => Nil
         }
 
-        context.issueAmbiguousError(AmbiguousImplicitTypeError(tree,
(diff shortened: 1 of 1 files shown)


retronym:
Thanks for spotting this, @milessabin. We've conservatively pushed this to 2.12.3, as I'd like to spend some time validating your assumptions and trying to engineer a test case that demonstrates the fix.

milessabin:
Not a problem.

retronym:
I agree with your analysis that this is a bug, but was benign given the current usage of this method in the compiler.

I also looked around for any other instances of the problem, but didn't find any.

PR #5461, opened 2016-10-17
SI-9967 External linking in scaladoc [WIP]

jgreen-zynga:
https://issues.scala-lang.org/browse/SI-9967

```
On 2.12.0-RC2, scaladoc appears to be generating fragment identifiers
Ex: index.html#com.foo.Bar
Instead of the new directory structure based urls. Ex: http://bog.us/com/foo/Bar.html
Example link generated when referencing a Long primitive as a return value:
<a href="http://www.scala-lang.org/api/2.12.0-RC2/index.html#scala.Long" class="extype" target="_top">Long</a>
Expected result:
http://www.scala-lang.org/api/2.12.0-RC2/scala/Long.html
```

I've signed the scala CLA

I don't know who would be good to review this, and I can't figure out how to run test/SI-191, hoping the continuous integration will pick it up and inform me.

PR #5799, opened 2017-03-22
SI-9967 External linking in scaladoc (continued from PR #5461)

haveo:
https://issues.scala-lang.org/browse/SI-9967

Trying to get #5461 over the finish line.

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/Settings.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/Settings.scala
@@ -253,13 +253,13 @@ class Settings(error: String => Unit, val printMsg: String => Unit = println(_))
     }
   }
 
-  def appendIndex(url: String): String = url.stripSuffix("index.html").stripSuffix("/") + "/index.html"
+  def stripIndex(url: String): String = url.stripSuffix("index.html").stripSuffix("/") + "/"
 
   lazy val extUrlMapping: Map[String, String] = docExternalDoc.value flatMap { s =>
     val idx = s.indexOf("#")
     if (idx > 0) {
       val (first, last) = s.splitAt(idx)
-      Some(new File(first).getCanonicalPath -> appendIndex(last.substring(1)))
+      Some(new File(first).getCanonicalPath -> stripIndex(last.substring(1)))
     } else {
       error(s"Illegal -doc-external-doc option; expected a pair with '#' separator, found: '$s'")
       None
--- a/src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
@@ -19,6 +19,7 @@ import scala.xml.Elem
 import scala.xml.dtd.DocType
 import scala.collection._
 import java.io.Writer
+import java.net.URI
 
 /** An html page that is part of a Scaladoc site.
   * @author David Bernard
@@ -145,8 +146,14 @@ abstract class HtmlPage extends Page { thisPage =>
         <span class="extmbr" name={ mbr.qualifiedName }>{ inlineToHtml(text) }</span>
     case Tooltip(tooltip) =>
       <span class="extype" name={ tooltip }>{ inlineToHtml(text) }</span>
-    case LinkToExternal(name, url) =>
-      <a href={ url } class="extype" target="_top">{ inlineToHtml(text) }</a>
+    case LinkToExternalTpl(name, baseUrlString, dtpl: TemplateEntity) =>
+      val baseUrl = new URI(Page.makeUrl(baseUrlString, Page.templateToPath(dtpl)))
+      val url = if (name.isEmpty) baseUrl
+                else new URI(baseUrl.getScheme, baseUrl.getSchemeSpecificPart, name)
+      if (hasLinks)
+        <a href={ url.toString } class="extype" name={ dtpl.qualifiedName }>{ inlineToHtml(text) }</a>
+      else
+        <span class="extype" name={ dtpl.qualifiedName }>{ inlineToHtml(text) }</span>
     case _ =>
       inlineToHtml(text)
   }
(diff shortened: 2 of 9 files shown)


haveo:
I signed the CLA

src/scaladoc/scala/tools/nsc/doc/Settings.scala
```diff
@@ -253,7 +253,7 @@ class Settings(error: String => Unit, val printMsg: String => Unit = println(_))
     }
   }
 
-  def appendIndex(url: String): String = url.stripSuffix("index.html").stripSuffix("/") + "/index.html"
+  def appendIndex(url: String): String = url.stripSuffix("index.html").stripSuffix("/") + "/"
```
szeiger: What about `file:` URLs? You can't rely on a rewrite to `index.html`. Linking to the parent dir is a useful option but it should not be the only one and not the default.

src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
```diff
@@ -145,8 +145,12 @@ abstract class HtmlPage extends Page { thisPage =>
         <span class="extmbr" name={ mbr.qualifiedName }>{ inlineToHtml(text) }</span>
     case Tooltip(tooltip) =>
       <span class="extype" name={ tooltip }>{ inlineToHtml(text) }</span>
-    case LinkToExternal(name, url) =>
-      <a href={ url } class="extype" target="_top">{ inlineToHtml(text) }</a>
+    case LinkToExternalTpl(name, baseUrl, dtpl: TemplateEntity) =>
```
szeiger: Why does `LinkToExternalTpl` have a `Tpl` type parameter when there is only a single handled case for `TemplateEntity` here? Is the type parameter necessary? If so, can it at least have `TemplateEntity` as an upper bound?

src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
```diff
@@ -145,8 +145,12 @@ abstract class HtmlPage extends Page { thisPage =>
...
+      def url = Page.makeUrl(baseUrl, Page.templateToPath(dtpl)) + name
```
szeiger: I don't see any url-encoding. `name` is produced by `MemberLookupBase.externalSignature` without any encoding. `Page.templatePath` uses `NameTransformer.encode`, so the result is *probably* safe but this needs to be double-checked (or just run through the encoder anyway).

src/scaladoc/scala/tools/nsc/doc/html/Page.scala
```diff
@@ -13,6 +13,58 @@ import scala.reflect.NameTransformer
...
+    val (file, pack) =
+      if(tpl.isPackage) {
+        ("index.html", tpl)
+      } else {
+        downInner(doName(tpl), tpl)
+      }
+    file :: downPacks(pack)
+  }
+
+  def makeUrl(baseUrl : String, path : List[String]) : String = baseUrl.stripSuffix("/") + "/" + path.reverse.mkString("/")
```
szeiger, line 53: I recommend doing all URI-related computations strongly typed with `java.net.URI` instead of stringly typed to ensure that paths are concatenated correctly and to make it clear what has been encoded and what has not.

I did not look at the rest of the scaladoc code to see how pervasive this kind of URI handling is. If it is already done with string manipulation in many cases a rewrite would best be left for another PR.

src/scaladoc/scala/tools/nsc/doc/html/Page.scala
```diff
@@ -13,6 +13,58 @@ import scala.reflect.NameTransformer
...
+
+  /** A relative link from a page's path to some destination path.
+    * @param destPath The path that the link will point to. */
+  def relativeLinkTo(thisPagePath: List[String], destPath: List[String]): String = {
```
szeiger, line 57: Case in point: With `java.net.URI` this would simply be `thisPagePath.relativize(destPath)`

szeiger (changes requested):

src/scaladoc/scala/tools/nsc/doc/Settings.scala
haveo: This is probably more of a naming problem as `appendIndex` is only used in `extUrlMapping` which is only used to get the base URL to which the file name is appended. This means that we shouldn't be linking to dirs directly.

src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
haveo: This is indeed weird. As far as I understand this template is always a `NoDocTemplate` so I'll try to make that explicit. This way the pattern matching should be exhaustive.

src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
haveo: Internal links seem to be handled the same way (append `#` then the signature without any encoding). I'd also be wary of encoding it as a URL as the forbidden characters are not the same. What would you encode it with?

src/scaladoc/scala/tools/nsc/doc/html/Page.scala
haveo: I didn't look at the whole code of scaladoc but `templateToPath` has simply been moved from `Page` to its object companion so I don't have any reason to think it would be much different. A quick `grep` did not seem to find much of any use of `java.net` in scaladoc. I'd rather leave this cleanup to another PR.

src/scaladoc/scala/tools/nsc/doc/html/HtmlPage.scala
szeiger: Quoting too much won't hurt. <https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#URI-java.lang.String-java.lang.String-java.lang.String-java.lang.String-java.lang.String-> knows how to do it correctly.

haveo:
@szeiger are there still some changes you need me to make?

szeiger:
The changes look good but the commits should be squashed. (I'd generally prefer to keep separate commits from different authors but since the original commit doesn't even compile we shouldn't have it in the history.)

SethTisue:
@szeiger is this mergeable for 2.12.3?

szeiger (approved):

PR #5801, opened 2017-03-23
Consistify trailing punctuation in spec

janekdb:
Three types of change,

 - Drop trailing comma on some lists
 - Add missed commas on some lists
 - Drop `  .` on some examples
 - Correct =/== on linearization examples

I have assumed trailing `  .` in `scala` examples was not significant - I couldn't infer any meaning in it but happy to be corrected.

## TODO
 - [x] Address linearization comments
 - [x] Address comments

### DIFF
--- a/spec/03-types.md
+++ b/spec/03-types.md
@@ -507,7 +507,7 @@ Assume the class definitions
 
 ```scala
 class Ref[T]
-abstract class Outer { type T } .
+abstract class Outer { type T }
 ```
 
 Here are some examples of existential types:
@@ -530,7 +530,7 @@ Ref[_ <: java.lang.Number]
 The type `List[List[_]]` is equivalent to the existential type
 
 ```scala
-List[List[t] forSome { type t }] .
+List[List[t] forSome { type t }]
 ```
 
 ###### Example
--- a/spec/06-expressions.md
+++ b/spec/06-expressions.md
@@ -222,9 +222,14 @@ the linearization of class `D` is `{D, B, A, Root}`.
 Then we have:
 
 ```scala
-(new A).superA == "Root",
-                          (new C).superB = "Root", (new C).superC = "B",
-(new D).superA == "Root", (new D).superB = "A",    (new D).superD = "B",
+(new A).superA == "Root"
+
+(new C).superB == "Root"
+(new C).superC == "B"
+
+(new D).superA == "Root"
+(new D).superB == "A"
+(new D).superD == "B"
 ```
 
 Note that the `superB` function returns different results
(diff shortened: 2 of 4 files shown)


janekdb:
The rich diff option avoids the puzzling blood-bath of red on the first file,
image

spec/06-expressions.md
```diff
@@ -224,7 +224,7 @@ Then we have:
 ```scala
 (new A).superA == "Root",
```
l0rinc: Please join these lines also :)

spec/07-implicits.md
```diff
@@ -218,7 +218,7 @@ which implicit arguments are searched is
 
 ```scala
 List[List[Int]] => Ordered[List[List[Int]]],
-List[Int] => Ordered[List[Int]]
+List[Int] => Ordered[List[Int]],
```
l0rinc, line 221: the list seems to continue here with `Int =>`

spec/07-implicits.md
```diff
@@ -239,7 +239,7 @@ of types for which implicit arguments are searched is
 
 ```scala
 Throwable => Ordered[Throwable],
-Throwable => Ordered[Throwable],
+Throwable => Ordered[Throwable]
 ...
```
l0rinc: if we interpret `...` as more elements to follow, we should keep the `,`.
Otherwise you could delete that also

spec/03-types.md
```diff
@@ -507,7 +507,7 @@ Assume the class definitions
 
 ```scala
 class Ref[T]
-abstract class Outer { type T } .
```
l0rinc, line 510: :+1:, why was there a trailing `.` in the first place?

l0rinc (changes requested):

spec/07-implicits.md
janekdb: Yes, so adding the missing comma makes sense here don't you think?

spec/07-implicits.md
janekdb: Agreed. Restored.

janekdb:
@paplorinc Ready for review.

spec/06-expressions.md
```diff
@@ -222,9 +222,14 @@ the linearization of class `D` is `{D, B, A, Root}`.
 Then we have:
 
 ```scala
-(new A).superA == "Root",
-                          (new C).superB = "Root", (new C).superC = "B",
-(new D).superA == "Root", (new D).superB = "A",    (new D).superD = "B",
+(new A).superA == "Root"
```
l0rinc, line 225: :+1:

spec/07-implicits.md
l0rinc: My mistake, yes :)
:+1:

l0rinc (approved):

SethTisue:
thank you!

PR #5802, opened 2017-03-23
Implement ListBuffer.isEmpty / nonEmpty / head efficiently

jrudolph:
Uses the extra length information in comparison to list to provide efficient implementations.

Evaluating these three methods turns up with about 6-7% of akka-http message parsing.

From the profiles it looks the likely reason are polymorphic invokeinterface calls against List (i.e. just forwarding to `start.isEmpty` etc wouldn't help).

### DIFF
--- a/src/library/scala/collection/mutable/ListBuffer.scala
+++ b/src/library/scala/collection/mutable/ListBuffer.scala
@@ -119,6 +119,10 @@ final class ListBuffer[A]
   // Don't use the inherited size, which forwards to a List and is O(n).
   override def size = length
 
+  // Override with efficient implementations using the extra size information available to ListBuffer.
+  override def isEmpty: Boolean = len == 0
+  override def nonEmpty: Boolean = len > 0
+
   // Implementations of abstract methods in Buffer
 
   override def apply(n: Int): A =


src/library/scala/collection/mutable/ListBuffer.scala
```diff
@@ -119,6 +119,11 @@ final class ListBuffer[A]
   // Don't use the inherited size, which forwards to a List and is O(n).
   override def size = length
 
+  // Override with efficient implementations using the extra size information available to ListBuffer.
+  override def isEmpty: Boolean = len == 0
+  override def nonEmpty: Boolean = len > 0
+  override def head: A = if (len > 0) start.asInstanceOf[::[A]].head else Nil.head
```
lrytz: This one doesn't look very cute.. `start.head` would be a virtual call with two cases, so C2 should be able to inline it. But before C2 it's probably slower than the monomorphic calls.

lrytz (commented):
Can you elaborate a bit on the kind of benchmarks you performed?

src/library/scala/collection/mutable/ListBuffer.scala
jrudolph: Yes, you are probably right about `head`. I removed that one.

jrudolph:
I basically just saw these turning up in play / akka-http benchmarks.

It seems one reason that I saw `head` turning up at all was that it hit the `MaxInlineDepth` at the wrong place, so that the `head` call wouldn't benefit from the bimorphic invocation optimization in C2 you hinted at above. `TraversableForwarder` adds a few frames which makes it more likely that a call against `List` (which should be always bimorphic) ends up going through a polymorphic call-site when not all forwarder frames cannot be inlined. The code I added for `head` is very similar to what C2 would optimize to, but I agree that it is not clear if optimizing at such a level makes sense.

For the `isEmpty` / `nonEmpty` case, it seems obvious that a single integer comparison would be faster than calling through to `List` to figure it out.

lrytz:
The default `MaxInlineLevel` is possibly low for Scala. I think @retronym experimented with a  patch to the JDK so that small methods (35 bytes/`MaxInlineSize`) are not counted. Anyway, i think the current patch makes sense. cc @Ichoran

jvican:
@lrytz @retronym Is that JDK patch available? I'd like to have a look and experiment with it if it's open-source.

lrytz:
I searched a bit before but couldn't find it

dwijnand:
Was it this one? It's a JDK patch by Jason, but it's about forwarders not `MaxInlineSize`.

https://gist.github.com/retronym/d27090a68570485c3e329a52db0c7b25

jrudolph:
Interesting information. 

FWIW, in this case, methods from `TraversableForwarder` might not qualify as a forwarder in the sense of the patch as they use e.g. `underlying.nonEmpty` where `underlying` is a method call itself.

retronym:
My JDK patch was intended to make the extra indirection of the trait encoding "invisible" to the inliner heuristics. I've been meaning to get back to it, but I haven't been able to get my dev environment setup for OpenJDK builds recently :(

retronym:
I'm seeing a 6% improvement to the performance of our compiler benchmark suite with `-XX:MaxInlineLevel=18` (the default is 9). I haven't tried fine tuning of the value.

Related discussion on SO: [Why does the JVM has a max inline depth?](https://stackoverflow.com/questions/32503669/why-does-the-jvm-have-a-maximum-inline-depth)

It will be interesting to see how much of that improvement I could get under the default limit with my JDK patch applied.

Ichoran:
@lrytz - Yes, this patch looks sensible to me too.

lrytz (approved):

PR #5807, opened 2017-03-27
Named Arguments & Implicits

m4dc4p:
Fixes a bug where implicit definition is not resolved when using named arguments at call site. (https://issues.scala-lang.org/browse/SI-7809)

Updates README with info on using SBT_OPTS to increase JVM heap size if OOM occurs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,11 @@ bootstrap build locally.
 For history on how the current scheme was arrived at, see
 https://groups.google.com/d/topic/scala-internals/gp5JsM1E0Fo/discussion.
 
+Note: If compilation fails with "Out of Memory" errors, try increasing the heap
+size:
+
+  $ SBT_OPTS=-Xmx8192M
+
 ### Using the sbt Build
 
 Core commands:
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4599,10 +4599,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           })
           def errorInResult(tree: Tree) = treesInResult(tree) exists (err => typeErrors.exists(_.errPos == err.pos))
 
-            val newArgs = args.collect( {
-                case AssignOrNamedArg(_, r) => r
-                case t => t
-            })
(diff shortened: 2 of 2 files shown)


m4dc4p:
CLA signed.

m4dc4p:
@SethTisue

README.md
```diff
@@ -91,6 +91,11 @@ bootstrap build locally.
 For history on how the current scheme was arrived at, see
 https://groups.google.com/d/topic/scala-internals/gp5JsM1E0Fo/discussion.
 
+Note: If compilation fails with "Out of Memory" errors, try increasing the heap
+size:
+
+  $ SBT_OPTS=-Xmx8192M
```
lrytz, line 97: This value is really big, `1536m` is good enough. Also, we should mention here again that we recommend https://github.com/paulp/sbt-extras, which sets `-Xmx1536m` by default. So it could be something like:

Note: If compilation fails with "Out of Memory" errors, set `SBT_OPTS=-Xmx1536m`, or use [sbt-extras](https://github.com/paulp/sbt-extras) (recommended) which defines this setting by default.

Feel free to submit this as a separate PR.

lrytz (changes requested):
Hi & welcome!

This is going to be more complicated, unfortunately.. Simply removing the argument names leads to unexpected / wrong results.

```
class Test {
  def m(a: Int) = 0
}

object A extends App {
  implicit class Ext(val f: Test) {
    def m(notMyName: String) = 1
  }
  assert((new Test).m("works") == 1)
  assert((new Test).m(a = "compiles") == 1) // weird, there's no parameter named `a` in the selected method
}

object B extends App {
  implicit class Ext(val f: Test) {
    def m(a: Int, b: Int) = a - b
  }
  assert((new Test).m(b = 5, a = 10) == 5) // ups...
}
```

m4dc4p:
Excellent counter-example, thanks for looking.

m4dc4p:
@lrytz Can I rebase and force push this PR, or is it better to close and create a new branch?

SethTisue:
rebasing and force-pushing is totally fine, that's usually how we work

SethTisue:
closing for inactivity. we can reopen it if activity resumes

PR #5808, opened 2017-03-28
Prefer library string manipulation to local reimplementation

janekdb:


src/scalap/scala/tools/scalap/scalax/util/StringUtil.scala
```diff
@@ -10,7 +10,7 @@ import java.beans.Introspector
 
 object StringUtil {
 
-  def trimStart(s: String, prefix: String) = if (s != null && s.startsWith(prefix)) s.substring(prefix.length) else s
+  def trimStart(s: String, prefix: String) = if (s != null) s.stripPrefix(prefix) else s
```
l0rinc: Could you please test whether we need that null check?
If not, you could probably replace `trimStart` with `stripPrefix` everywhere!

janekdb:
@paplorinc - Your suggestion has been incorporated. In both cases it doesn't seem as if the string can ever be null so this removed some unnecessarily defensive coding.

l0rinc (approved):

PR #5809, opened 2017-03-28
Allow inlining into trait constructors

lrytz:
Fixes scala/scala-dev#350

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/InlinerHeuristics.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/InlinerHeuristics.scala
@@ -64,13 +64,6 @@ class InlinerHeuristics[BT <: BTypes](val bTypes: BT) {
     }).filterNot(_._2.isEmpty).toMap
   }
 
-  private def isTraitStaticSuperAccessorName(s: String) = s.endsWith("$")
-  private def traitStaticSuperAccessorName(s: String) = s + "$"
-
-  private def isTraitSuperAccessor(method: MethodNode, owner: ClassBType): Boolean = {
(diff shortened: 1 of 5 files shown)


retronym:
Thanks, this will eliminate a closure allocation in `trait AliasTypeRef { require(...) }`!

src/compiler/scala/tools/nsc/backend/jvm/opt/InlinerHeuristics.scala
```diff
@@ -64,7 +64,7 @@ class InlinerHeuristics[BT <: BTypes](val bTypes: BT) {
     }).filterNot(_._2.isEmpty).toMap
   }
 
-  private def isTraitStaticSuperAccessorName(s: String) = s.endsWith("$")
+  private def isTraitStaticSuperAccessorName(s: String) = s != "$init$" && s.endsWith("$")
```
retronym: Do we emit any other static methods (e.g. static-ified lambda lifted or lambda impl methods) that could possibly have the trailing `$` and also be misclassified?

retronym: ... it might be more robust to look for a static method named `foo$(self)` that only calls `self.foo`.

lrytz: done, thanks

SethTisue:
@lrytz should this go in 2.12.2?

lrytz:
It's not a regression and a benign bug, so 2.12.3 is just fine.

retronym:
@lrytz This needs a rebase (https://github.com/scala/scala/compare/2.12.x...retronym:review/5809?expand=1 should do the trick). Otherwise LGTM.

lrytz:
Thanks for the rebase! I looked at the diff of the diffs, LGTM.

PR #5810, opened 2017-03-28
Set scala.binary.version to 2.12 until 2.13.0-M1 is out

lrytz:
The current override in VersionUtil hard-codes the binary version
to `2.12`, which means that the `-Dscala.binary.version` override
in the bootstrap script does not have its intended effect. So we
were in fact using the 2.12 binary modules also when testing the
boostrapped 2.13 builds.

### DIFF
--- a/project/VersionUtil.scala
+++ b/project/VersionUtil.scala
@@ -163,10 +163,7 @@ object VersionUtil {
   /** Build a dependency to a Scala module with the given group and artifact ID */
   def scalaDep(group: String, artifact: String, versionProp: String = null, scope: String = null, compatibility: String = "binary") = {
     val vp = if(versionProp eq null) artifact else versionProp
-    // TODO for 2.13.0-M1 -- eventually it will no longer work to use 2.12 modules,
-    // but at the moment, there aren't any 2.13.0-M1 modules to use yet, so hardcode 2.12 for now.
-    // val m = group % (artifact + "_" + versionProps(s"scala.$compatibility.version")) % versionNumber(vp)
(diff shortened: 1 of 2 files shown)


lrytz:
This reverts part of https://github.com/scala/scala/pull/5581.

lrytz:
review by @SethTisue

SethTisue:
@lrytz looks like this is causing bootstrap to fail. at https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-bootstrap/314/consoleFull:

```
[error] 2 of 12 test tasks failed:
[error] - osgiTestEclipse/test
[error] - library/mima
```

I won't have time to look at it today or tomorrow.

lrytz:
Ah, I didn't see that, thanks for the hint. I'll take a look.

lrytz:
I think the build failures will be solved by https://github.com/scala/scala/pull/5790, which removes the `scala-swing` dependency. Since https://github.com/scala/scala/commit/1371a216d81836b79d9b60afd5481c088bf5c438 we no longer build swing in the bootstrap script, however the swing module is still needed by osgi it seems. In 2.12.x this was fine because the binary version didn't change. But 2.13.x is currently fully cross-versioned, so the build searches for `scala-swing_2.13.0-pre-27a48e3;2.0.0`, which it doesn't find as it wasn't built.

lrytz:
Yep, #5790 fixed it, https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-bootstrap/318/ passed.

PR #5811, opened 2017-03-28
Make ImplicitInfo hashCode consistent with equals.

milessabin:
`equals` on `ImplicitInfo` is defined in terms of `=:=` on types,
```scala
override def equals(other: Any) = other match {
  case that: ImplicitInfo =>
    this.name == that.name &&
    this.pre =:= that.pre &&
    this.sym == that.sym
  case _ => false
}
```
which can identify non-equal but equivalent types. On the other hand `hashCode` is defined in terms of `##` which is unlikely to yield equal values for types which are equivalent under `=:=` but not equal,

```scala
override def hashCode = name.## + pre.## + sym.##
```
I noticed this while attempting to remove duplicates from a `List[ImplicitInfo]` using `distinct` ... the result contained multiple `ImplicitInfo`s which compared `==` because `distinct` is implemented in terms of a `HashSet`.

It's not clear to me what the impact of this bug is, but it could have an effect on the performance of the `improvesCache`.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -255,7 +255,10 @@ trait Implicits {
           this.sym == that.sym
       case _ => false
     }
-    override def hashCode = name.## + pre.## + sym.##
+    override def hashCode = {
+      import scala.util.hashing.MurmurHash3._
+      finalizeHash(mix(mix(productSeed, name.##), sym.##), 2)
+    }
     override def toString = (
       if (tpeCache eq null) name + ": ?"
       else name + ": " + tpe
--- a/test/junit/scala/tools/nsc/typechecker/Implicits.scala
+++ b/test/junit/scala/tools/nsc/typechecker/Implicits.scala
@@ -1,0 +1,39 @@
+package scala.tools.nsc
+package typechecker
+
+import org.junit.Assert._
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+import scala.tools.testing.BytecodeTesting
+
+@RunWith(classOf[JUnit4])
+class ImplicitsTests extends BytecodeTesting {
+  import compiler.global._, definitions._, analyzer._
+
+  @Test
+  def implicitInfoHashCode(): Unit = {
+    val run = new global.Run
+
+    enteringPhase(run.typerPhase) {
+      val T0 = IntClass.tpeHK
+      val T1 = refinedType(List(T0), NoSymbol)
+
+      assert(T0 =:= T1)
+      assert(T0 != T1)
+      assert(T0.hashCode != T1.hashCode)
+
+      val I0 = new ImplicitInfo(TermName("dummy"), T0, NoSymbol)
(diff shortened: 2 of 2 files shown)


milessabin:
Actually `normalize` doesn't seem to do the trick here. Perhaps we should just drop the hash term for the type and use,
```
override def hashCode = name.## + sym.##
```

dwijnand:
I did some git log spelunking and this has been this way since its inception (Feb 2009). The commit message says:

> Fixed an ambiguity problem with implcits, and a spurious cyclic
reference error mentioned by Jorge. Tightened overriding checks (test
case is new collection libraries).

https://github.com/scala/scala/commit/266df9f05ecd11e67166e57cc58236646f2f50bb#diff-04639e477fcdf844cf3ae0e3e4bf5d7eR638

milessabin:
By the looks of that commit and the description of the issues addressed, it seems that a correct implementation of `equals` was the primary focus. I doesn't look like there were any hash maps or sets with `ImplicitInfo` keys or elements in that era so a faulty `hashCode` implementation would not have been immediately obvious ... let's face it, it's hardly obvious now.

milessabin:
I've removed `pre` from the hash computation and added a Junit test for `ImplicitInfo` `equals`/`hashCode` consistency.

retronym (changes requested):
Almost LGTM, thanks for this.

I agree with your suggestion that incoherent hashCode appears to have been a benign problem for `improveCache`, it would just have just resulted in some extra cache misses.

After your change, we'll have a longer linear search in the hash bucket if there are many implicit infos over the same symbol that are drawn from different prefixes, which could be a slow down certain implicit searches.

There is also a hashmap of `ImplictInfo`-s used by `allImplicits`, which is used by the presentation compiler to autocomplete extension methods. That ought to be more reliable after your change.

While we're in this area, I'd suggest that we combine the component hashCodes more robustly:

```scala
override def hashCode = {
  (name, sym).##  // note, pre is not included in the hash as we can't do this in a way compatible with =:=
}
```

retronym:
> After your change, we'll have a longer linear search in the hash bucket if there are many implicit infos over the same symbol that are drawn from different prefixes, which could be a slow down certain implicit searches.

Looking at the surrounding code, I see that the `improvesCache` is only used for static symbols (the owner chain is module and package classes, only) which won't have a variety of prefixes.

```
        if (info1.sym.isStatic && info2.sym.isStatic) {
          improvesCache get ((info1, info2)) match {
            case Some(b) => if (Statistics.canEnable) Statistics.incCounter(improvesCachedCount); b
            case None =>
              val result = isStrictlyMoreSpecific(info1.tpe, info2.tpe, info1.sym, info2.sym)
              improvesCache((info1, info2)) = result
              result
          }
        } else isStrictlyMoreSpecific(info1.tpe, info2.tpe, info1.sym, info2.sym)
```

milessabin:
@retronym do you want me to rework the `hashCode` implementation as you've done with a `Tuple2` or inline it in terms of `MurmurHash3`?

retronym:
The latter would be ideal.

milessabin:
Reworked `hashCode` in terms of `MurmurHash3` and squashed.

retronym (approved):

PR #5816, opened 2017-03-31
Allow user-defined `[un]apply` in case companion

adriaanm:
Don't emit a synthetic `apply` (or `unapply`) when it would
clash with an existing one. This allows e.g., a `private apply`,
along with a `case class` with a `private` constructor.

We have to retract the synthetic method in a pretty roundabout way,
as we need the other methods and the owner to be completed already.
Unless we have to complete the synthetic `apply` while completing
the user-defined one, this should not be a problem. If this does
happen, this implies there's a cycle in computing the user-defined
signature and the synthetic one, which is not allowed.

First submitted as #5730 for 2.11.9, see also scala/scala-dev#352 (as in, to be investigated if the change relates to this change, though unlikely since it already regressed in 2.12.1)

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -613,7 +613,15 @@ trait Namers extends MethodSynthesis {
       noDuplicates(selectors map (_.rename), AppearsTwice)
     }
 
-    def enterCopyMethod(copyDef: DefDef): Symbol = {
+    class CompleterWrapper(completer: TypeCompleter) extends TypeCompleter {
+      val tree = completer.tree
+
+      override def complete(sym: Symbol): Unit = {
+        completer.complete(sym)
+      }
+    }
+
+    def copyMethodCompleter(copyDef: DefDef): TypeCompleter = {
       val sym      = copyDef.symbol
       val lazyType = completerOf(copyDef)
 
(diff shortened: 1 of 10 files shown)


adriaanm:
Diagnosed why 2.12.1 was already failing, luckily the same fix as in 2.11.10 applies 

> The problem was already present before the introduction of
> `applyUnapplyMethodCompleter`, as 63f7b35 (in #5294) introduced
> a similar bug where the `PolyTypeCompleter`'s `typeParams` override
> was masked.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -632,14 +643,63 @@ trait Namers extends MethodSynthesis {
...
       }
 
-      sym setInfo {
-        mkTypeCompleter(copyDef) { sym =>
-          assignParamTypes()
-          lazyType complete sym
-        }
+      mkTypeCompleter(copyDef) { sym =>
+        assignParamTypes()
+        lazyType complete sym
```
lrytz: should we use `CompleterWrapper` here as well? maybe we have the same bug caused by a missing `typeParams` override lingering around.

lrytz (commented):
I have the feeling that you caught some old revision when porting this change to 2.12.x. For example, I made a comment in the 2.11 pr https://github.com/scala/scala/pull/5730#pullrequestreview-24329809 about a code comment mentioning `addIfMissing` which doesn't exist. You then fixed this. Now, that code comment is back in the 2.12 PR here.

Similarly for a code comment saying `rule out locked symbols from the owner.info.member call`, which I didn't understand, and you then clarified it. Here we have the old version again.

Also, I remember that there were some additional changes that were later removed in order to have a minimal change in 2.11.x, see your comment: https://github.com/scala/scala/pull/5730#issuecomment-282398536. Should we bring those back? I don't remember what they were.

adriaanm:
You're right. Sorry! I missed the commits that addressed your review feedback 😊

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: Good point. Looking into it

SethTisue:
@adriaanm what does the "on-hold" status here mean...?

adriaanm:
Even though the last commit is green, this is not ready (need to implement Lukas's review feedback).

src/compiler/scala/tools/nsc/typechecker/Namers.scala
adriaanm: Turns out I already had a commit for that in my wip branch (79a7015) :-) Pushed to this one, plus some more boy scouting.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -632,14 +640,63 @@ trait Namers extends MethodSynthesis {
...
+          def retainOnlyParamAnnots(param: Symbol) =
+            param setAnnotations (param.annotations filter AnnotationInfo.mkFilter(ParamTargetClass, defaultRetention = false))
+
+          sym.info.paramss.foreach(_.foreach(retainOnlyParamAnnots))
+
+          // If there's a same-named locked symbol, we're currently completing its signature.
+          // This means it (may) refer to us, and is thus either overloaded or recursive without a signature.
+          // rule out locked symbols from the owner.info.member call
+          val scopePartiallyCompleted =
+            companionContext.scope.lookupAll(sym.name).exists(existing => existing != sym && existing.hasFlag(LOCKED))
```
lrytz: do you remember why we did this differently in the 2.11.x branch? `new HasMember(ownerInfo, sym.name, BridgeFlags | SYNTHETIC, LOCKED).apply()` https://github.com/scala/scala/pull/5730/files#diff-44437929b3df2683c7ed998c20eb5b03R665

src/reflect/scala/reflect/internal/tpe/FindMembers.scala
```diff
@@ -285,4 +285,18 @@ trait FindMembers {
       initBaseClasses.head.newOverloaded(tpe, members)
     }
   }
+
+  private[scala] final class HasMember(tpe: Type, name: Name, excludedFlags: Long, requiredFlags: Long) extends FindMemberBase[Boolean](tpe, name, excludedFlags, requiredFlags) {
```
lrytz: This class is not used

lrytz: oh yes it is sorry

src/compiler/scala/tools/nsc/typechecker/Namers.scala
lrytz: Ah, I should look at the full diff, not individual commits. Sorry.

lrytz (approved):
nice additional cleanups, thanks!

lrytz:
The last refactoring https://github.com/scala/scala/pull/5816/commits/ade53a123c1edce12db442ee74b636d130e7e0f2 broke scalameta in the community build:

```
[scalameta] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/scalameta-94370b607c4b26a9c7e8869bcfea3e88bcf250d6/scalahost/src/main/scala-2.12.1/scala/meta/internal/scalahost/ScalahostAnalyzer.scala:22: not found: value mkTypeCompleter
[scalameta] [error]     override def selfTypeCompleter(tree: Tree) = mkTypeCompleter(tree) { sym =>
[scalameta] [error]                                                  ^
[scalameta] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.7-RC1/project-builds/scalameta-94370b607c4b26a9c7e8869bcfea3e88bcf250d6/scalahost/src/main/scala-2.12.1/scala/meta/internal/scalahost/ScalahostAnalyzer.scala:22: not found: value mkTypeCompleter
[scalameta] [error]     override def selfTypeCompleter(tree: Tree) = mkTypeCompleter(tree) { sym =>
[scalameta] [error]                                                  ^
[scalameta] [error] one error found
```

https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1414/consoleFull

cc @SethTisue @xeno-by

adriaanm:
I'm working on a follow-up PR for this. I over-refactored a bit. Will bring back usage of the completer factory methods. 

See #5844

PR #5817, opened 2017-03-31
Fix typo in JavaConverters doc

ashawley:
There is no such thing as scala.collection.mutable.concurrent.Map

    error: object concurrent is not a member of package scala.collection.mutable

Introduced in 2908236a.  Found by @cescude


dwijnand (approved):

SethTisue:
good catch, thanks!

PR #5818, opened 2017-03-31
comonadic comprehensions

shimib:
This is a proposal (POC) to provide a new keyword (cofor) for co-monadic comprehensions in Scala.
The work is based over Dominic Orchard and Alan Mycroft paper: https://www.cl.cam.ac.uk/~dao29/publ/codo-notation-orchard-ifl12.pdf

If you find the implementation flawed, it is me to blame and not them!

The 'cofor' expression returns a function: (T[A] => B).
The function's type MUST be stated explicitly in the call-site. e.g.:
val result: Zipper[Person] => Boolean = cofor...

the syntax:
cofor(inputpatten) {
pattern1 <- gen1
pattern2 <- gen2
...
} yield body

the 'cofor' and generator must use types that provide:
def map(A => B)
def extract: A
def coflatMap(T[A] => B)

Implementation is much more complex than 'for' desugaring as stated in the paper.

Current implementation supports full Scala patterns in the input and generator patterns.

currently missing:
Desugared tree positioning validation (should not affect surrounding code)
Quasiquotes/macros/reification -- this is the next phase in the implementation.

The flag to control the keyword: -Ycofor-extension
If the flag is not enabled, current Scala code should not be affected.

Compatibility: adding a new keyword breaks existing use of 'cofor' identifier and it should be back-quoted.

Some discussion and example can is in: https://contributors.scala-lang.org/t/providing-co-monadic-comprehensions/518

A SIP was also submitted: http://docs.scala-lang.org/sips/pending/comonadic-comprehensions.html

This PR is targeted for TLS. The same PR was opened for 2.12.x: scala#5725

### DIFF
--- a/bincompat-forward.whitelist.conf
+++ b/bincompat-forward.whitelist.conf
@@ -389,6 +389,10 @@ filter {
         problemName=MissingMethodProblem
     },
     {
+        matchName="scala.reflect.runtime.Settings.YcoforExtension"
+        problemName=MissingMethodProblem
+    },
+    {
         matchName="scala.reflect.runtime.Settings.YpartialUnification"
         problemName=MissingMethodProblem
     },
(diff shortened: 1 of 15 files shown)


SethTisue:
SIP committee didn't think this was going to go forward

PR #5819, opened 2017-04-01
Applicative desugaring in for comprehensions

ghik:
This PR is a proof of concept of new syntax bringing support of applicative desugaring into `for` comprehensions.

I know that such a change requires prior discussion and ultimately, a SIP. Here's a topic on Scala contributors forum where all the details of the proposal are explained: https://contributors.scala-lang.org/t/new-syntax-applicative-desugaring-in-for-comprehensions-with-pr/690


SethTisue:
a SIP on this was rejected by the committee as premature

PR #5820, opened 2017-04-03
Improve performance of the backend

retronym:
In aggregate, I measure a 3.5% improvement from this PR.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
@@ -461,10 +461,10 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
       locals.reset(isStaticMethod = methSymbol.isStaticMember)
       jumpDest = immutable.Map.empty[ /* LabelDef */ Symbol, asm.Label ]
       // populate labelDefsAtOrUnder
-      val ldf = new LabelDefsFinder
+      val ldf = new LabelDefsFinder(dd.rhs)
       ldf.traverse(dd.rhs)
-      labelDefsAtOrUnder = ldf.result.withDefaultValue(Nil)
-      labelDef = labelDefsAtOrUnder(dd.rhs).map(ld => (ld.symbol -> ld)).toMap
+      labelDefsAtOrUnder = ldf.result
+      labelDef = ldf.directResult.map(ld => (ld.symbol -> ld)).toMap
       // check previous invocation of genDefDef exited as many varsInScope as it entered.
       assert(varsInScope == null, "Unbalanced entering/exiting of GenBCode's genBlock().")
       // check previous invocation of genDefDef unregistered as many cleanups as it registered.
@@ -591,7 +591,7 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
        * but the same vars (given by the LabelDef's params) can be reused,
        * because no LabelDef ends up nested within itself after such duplication.
        */
-      for(ld <- labelDefsAtOrUnder(dd.rhs); ldp <- ld.params; if !locals.contains(ldp.symbol)) {
+      for(ld <- labelDefsAtOrUnder.getOrElse(dd.rhs, Nil); ldp <- ld.params; if !locals.contains(ldp.symbol)) {
         // the tail-calls xform results in symbols shared btw method-params and labelDef-params, thus the guard above.
         locals.makeLocal(ldp.symbol)
       }
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
@@ -462,7 +462,7 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
       jumpDest = immutable.Map.empty[ /* LabelDef */ Symbol, asm.Label ]
       // populate labelDefsAtOrUnder
       val ldf = new LabelDefsFinder(dd.rhs)
-      ldf.traverse(dd.rhs)
+      ldf(dd.rhs)
       labelDefsAtOrUnder = ldf.result
       labelDef = ldf.directResult.map(ld => (ld.symbol -> ld)).toMap
       // check previous invocation of genDefDef exited as many varsInScope as it entered.
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -298,19 +298,29 @@ abstract class BTypes {
    * referring to BTypes.
    */
   sealed trait BType {
-    final override def toString: String = this match {
-      case UNIT   => "V"
-      case BOOL   => "Z"
-      case CHAR   => "C"
-      case BYTE   => "B"
-      case SHORT  => "S"
-      case INT    => "I"
-      case FLOAT  => "F"
-      case LONG   => "J"
-      case DOUBLE => "D"
-      case ClassBType(internalName) => "L" + internalName + ";"
-      case ArrayBType(component)    => "[" + component
-      case MethodBType(args, res)   => "(" + args.mkString + ")" + res
+    final override def toString: String = {
+      val builder = new java.lang.StringBuilder(64)
+      buildString(builder)
+      builder.toString
+    }
+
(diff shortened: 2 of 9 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
```diff
@@ -461,10 +461,10 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
       locals.reset(isStaticMethod = methSymbol.isStaticMember)
       jumpDest = immutable.Map.empty[ /* LabelDef */ Symbol, asm.Label ]
       // populate labelDefsAtOrUnder
-      val ldf = new LabelDefsFinder
+      val ldf = new LabelDefsFinder(dd.rhs)
       ldf.traverse(dd.rhs)
```
lrytz: could be `.apply()` now

src/compiler/scala/tools/nsc/backend/jvm/BytecodeWriters.scala
```diff
@@ -118,10 +135,20 @@ trait BytecodeWriters {
     def writeClass(label: String, jclassName: String, jclassBytes: Array[Byte], outfile: AbstractFile) {
       assert(outfile != null,
              "Precisely this override requires its invoker to hand out a non-null AbstractFile.")
-      val outstream = new DataOutputStream(outfile.bufferedOutput)
+      if (outfile.file != null) {
+        try {
+          Files.write(outfile.file.toPath, jclassBytes)
+        } catch {
+          case _: java.nio.file.NoSuchFileException =>
```
lrytz, line 142: would it be too slow to check `java.nio.file.Files.exists(p.getParent)`?

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
```diff
@@ -108,8 +108,6 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
 
     assert(classSym != NoSymbol, "Cannot create ClassBType from NoSymbol")
     assert(classSym.isClass, s"Cannot create ClassBType from non-class symbol $classSym")
-    assertClassNotArrayNotPrimitive(classSym)
-    assert(!primitiveTypeToBType.contains(classSym) || isCompilingPrimitive, s"Cannot create ClassBType for primitive class symbol $classSym")
```
lrytz, line 112: I'm a bit sad to see those go.. They were helpful for finding bugs, and they are also useful documentation. Can we put them under a flag or elide them? Or even leave them as comments.

src/compiler/scala/tools/nsc/backend/jvm/BytecodeWriters.scala
retronym: I found some small but measurable improvement from this approach, as there are typically many classfiles per directory, which makes the optimism pays off. I could split this into two commits so that we could reproduce that measurement.

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
retronym: I'll resurrect under `-Xdev` (which is also enabled by `-Ydebug`)

retronym:
```
```

src/compiler/scala/tools/nsc/backend/jvm/BytecodeWriters.scala
mkeskells: I think that there is a  better solution in https://github.com/scala/scala/pull/5815, that will only check a given directory once, and will not call createDirectories unless it the first time that the directory is reached

mkeskells: @lrytz each stat call is expensive in windows particually

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -298,19 +298,29 @@ abstract class BTypes {
...
-      case BYTE   => "B"
-      case SHORT  => "S"
-      case INT    => "I"
-      case FLOAT  => "F"
-      case LONG   => "J"
-      case DOUBLE => "D"
-      case ClassBType(internalName) => "L" + internalName + ";"
-      case ArrayBType(component)    => "[" + component
-      case MethodBType(args, res)   => "(" + args.mkString + ")" + res
+    final override def toString: String = {
```
mkeskells, line 301: this is non optimal still

There is a better version I think in https://github.com/scala/scala/pull/5815/files#diff-61481becb11478d78c0b577ecad981aaR320

that version caches the result and shares the (java) StringBuilder when building a compound type, and avoid builder generation for primative types

lrytz: @mkeskells it would be good to extract small, self-contained pieces of #5815 that can be submitted as individual PRs. This will make reviewing much easier and also avoid duplicate work.

src/compiler/scala/tools/nsc/backend/jvm/BytecodeWriters.scala
retronym: I like the approach of caching the results of file stats, but opted not to do that in this PR because it takes a bit of care to get the lifecycle of that cache sorted out (we need to start fresh on a new compiler run, but the current design of `ByteCodeWriters` didn't have a hook for that part of the lifecycle.)

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
retronym: Caching all the intermediate strings is an interesting idea. Usually I'd be a bit cautious about that for a recursive structure, as it requires exponential space wrt the depth of the tree. BTypes, however, only have "deep" trees for higher dimensional arrays, so this might not be a practical concern here.

src/compiler/scala/tools/nsc/backend/jvm/BytecodeWriters.scala
mkeskells: @retronym it is hooked in this my PR as the OutputDirectories are created new for each run

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: @retronym should be easy to bake off as we have the profiler and this is a measurable part of the backend CPU, and we have stable allocation counters

Not sure if the allocation counters are in the version that you have

mkeskells: @lrytz I will be working on that but I also have a day job. I was trying to get more tests to pass

I have seperately opened a discussion with @adriaanm about this coordination

adriaanm:
I'm looking into options for that. Will get back to you on that thread.

src/compiler/scala/tools/nsc/backend/jvm/BytecodeWriters.scala
mkeskells: I have split out an alternative as https://github.com/scala/scala/pull/5834

will run performance test overnight and post up is successful

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: @lrytz split out several PRs for this https://github.com/scala/scala/pull/5831 https://github.com/scala/scala/pull/5834 https://github.com/scala/scala/pull/5833 https://github.com/scala/scala/pull/5832

haven't test them or got he performance figures - will run overnight and post up if successful

lrytz (commented):
This LGTM, but it overlaps with @mkeskells'  https://github.com/scala/scala/pull/5831 (BType descriptor generation) and https://github.com/scala/scala/pull/5834 (Classfile writing IO). These PRs are more comprehensive than the corresponding changes here. We can either remove the two commits here or rebase the other PRs.

retronym:
My impression is that both of those PRs are worth pursuing, but that the initial versions herein will be a good starting point (simpler code changes / deliver much of the performance improvement), so it would make sense to merge this and rebase the other changes. That way, if we have to revert the more involved changes due to a regression or undesirable performance consequence, we'd be back to square 2, instead of square 1.

lrytz:
Sounds good to me.

lrytz (approved):

PR #5821, opened 2017-04-03
Revert #5664 because the binary incompatible change leaks via erasure

adriaanm:
This reverts #5664 because the binary incompatible change leaks via erasure. We'll need to cut 2.11.10 and declare 2.11.9 a dud.

Originally discovered and reported in detail by @xuwei-k and @sjrd at https://github.com/scala-js/scala-js/issues/2847#issuecomment-290769846 (see also the [2.11.9 release issue]( https://github.com/scala/scala-dev/issues/330#issuecomment-290747384))


adriaanm:
@szeiger, @lrytz could you summarize your findings here for the release notes.

lrytz:
I'll do that, will write something.

lrytz:
PR [#5664](https://github.com/scala/scala/pull/5664) added a number of optimzied implementations for certain collection operations on a specific type. For example, `List.filter` received a new implementation overriding the one inherited from `TraversableLike`. In a simplified view, 2.11.8 had the following structure:

```scala
package test.collection.immutable
trait TraversableLike[+A, +Self] {
  def filter(p: A => Boolean): Self = ???
}
class List[+A] extends TraversableLike[A, List[A]]
```

Due to type erasure, method `filter` in classfile `TraversableLike` has return type `Object`.

We cannot add an overriding implementation of `filter` in class `List` in a binary compatible way, as [noted in a comment on PR #5653](https://github.com/scala/scala/pull/5653#issuecomment-275099015):

```scala
class List[+A] extends TraversableLike[A, List[A]] {
  override def filter(p: A => Boolean): List[A] = ???
}
```
The new method would have return type `List` in bytecode (to perform the override, the Scala compiler would generat a "bridge" method). An invocation in client code of this new, more specific method would not work with the 2.11.8 library on the classpath.

To prevent this issue, an intermediate class was added. This class was supposed to be completely internal and never appear in user code.

```scala
private[immutable] trait FilteredTraversableInternal[+A, +Repr] extends TraversableLike[A, Repr] {
  override def filter(p: A => Boolean): Repr = ???
}
class List[+A] extends TraversableLike[A, List[A]] with FilteredTraversableInternal[A, List[A]]
```

The new trait is private to package `immutable`, so user code cannot introduce references to it. Unfortunately, the erasure phase of the Scala compiler does introduce references to this new class into user code. This means that user code compiled with 2.11.9 cannot run with a 2.11.8 library, as the class is missing.

Assume we have the following user code:

```scala
class C {
  def m(o: Option[List[String]]): Unit = o.get.filter(_ => true)
}
```

Compiling this with the original collections code and de-compiling `C.m` with [cfr](http://www.benf.org/other/cfr/) gives

```scala
public void m(Option<List<String>> o) {
  ((TraversableLike)o.get()).filter(...);
}
```

Due to erasure, the method `Option.get` has return type `Object` in bytecode. The Scala compiler introduces a cast to `TraversableLike` in order to make the invocation of method `filter` valid - the `Object` class does not have such a method.

If we re-compile the same example with the `FilteredTraversableInternal` intermediate class, the cast inserted by the Scala compiler now uses the `FilteredTraversableInternal`, because that is the class in which `filter` is defined. This way, a reference to the new trait finds its way into user code.

lrytz:
we also need a short version of that.. :)

adriaanm:
Thanks @lrytz! @szeiger, let me know when you feel this is ready to go.

szeiger (approved):

szeiger:
FTR, I don't think there is any part of this that can be salvaged. Everything relies on injecting a private superclass and is therefore not binary compatible.

PR #5822, opened 2017-04-03
SI-9937 find nested java classes if InnerClass entry is missing

lrytz:
When a classfile has a reference to an inner class C$D but no
InnerClass entry for it, the classfile parser would use the
top-level symbol C$D. In a different classfile, if there's also
a reference to C$D, but the InnerClass entry exists, the
symbol D owned by C (C.D) would be used. Therefore the two
signatures would be incompatible, which can lead to a spurious
type error.

Also, when an inner symbol C.D is resolved, the top-level
symbol C$D is invalidated and removed from the scope. A
subsequent lookup of the top-level symbol C$D (from a classfile
with a missing InnerClass entry) would fail.

This patch identifies the case when a class name containing
a $ is being looked up in a package. It splits the name, resolves
the outer class, and then searches for a member class.

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
@@ -380,11 +380,41 @@ abstract class ClassfileParser {
   }
 
   private def lookupClass(name: Name) = try {
-    if (name containsChar '.')
-      rootMirror getClassByName name
-    else
+    def lookupTopLevel = {
+      if (name containsChar '.')
+        rootMirror getClassByName name
+      else
       // FIXME - we shouldn't be doing ad hoc lookups in the empty package, getClassByName should return the class
-      definitions.getMember(rootMirror.EmptyPackageClass, name.toTypeName)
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -380,11 +380,36 @@ abstract class ClassfileParser {
...
-    else
+    def lookupTopLevel = {
+      if (name containsChar '.')
+        rootMirror getClassByName name
+      else
       // FIXME - we shouldn't be doing ad hoc lookups in the empty package, getClassByName should return the class
-      definitions.getMember(rootMirror.EmptyPackageClass, name.toTypeName)
+        definitions.getMember(rootMirror.EmptyPackageClass, name.toTypeName)
+    }
+
```
lrytz: need to add a comment what this hack is about :)

lrytz:
`error: Class scala.Function1$mcZD not found - continuing with a stub`

adriaanm (approved):
Very well described solution! Just curious, is there a reason to look in the static scope before the instance scope?

lrytz:
No, there's no particular reason I did it in this order.

PR #5785, opened 2017-03-20
Various performance optimizations / experiments [ci: last-only]

retronym:
Puship WIP here to facilitate benchmarking.

PR #5823, opened 2017-04-04
Compiler performance improvements around Symbols and Types

retronym:
Seems to bring about 4% in aggregate.

Thematically cherry-picked from #5785

_UPDATE_ : benchmark results, with the latest iteration of this PR.

Before:
```
$ for sha in 2.12.3-bin-9acab45-SNAPSHOT 2.12.3-bin-9b28b01-SNAPSHOT ; do sbt "set scalaVersion in compilation := \"$sha\"" "hot -psource= -f3 -wi 10 -jvmArgs -Xmx1G"; sleep 10; done

[info] HotScalacBenchmark.compile                                         sample  240  1299.221 ± 6.034  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00                           sample       1270.874          ms/op
[info] HotScalacBenchmark.compile:compile·p0.50                           sample       1291.846          ms/op
[info] HotScalacBenchmark.compile:compile·p0.90                           sample       1319.109          ms/op
[info] HotScalacBenchmark.compile:compile·p0.95                           sample       1354.655          ms/op
[info] HotScalacBenchmark.compile:compile·p0.99                           sample       1433.592          ms/op
[info] HotScalacBenchmark.compile:compile·p0.999                          sample       1434.452          ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999                         sample       1434.452          ms/op
[info] HotScalacBenchmark.compile:compile·p1.00                           sample       1434.452          ms/op
```

After:

```
[info] HotScalacBenchmark.compile                                         sample  251  1251.348 ± 7.449  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00                           sample       1214.251          ms/op
[info] HotScalacBenchmark.compile:compile·p0.50                           sample       1243.611          ms/op
[info] HotScalacBenchmark.compile:compile·p0.90                           sample       1277.166          ms/op
[info] HotScalacBenchmark.compile:compile·p0.95                           sample       1335.466          ms/op
[info] HotScalacBenchmark.compile:compile·p0.99                           sample       1418.933          ms/op
[info] HotScalacBenchmark.compile:compile·p0.999                          sample       1451.229          ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999                         sample       1451.229          ms/op
[info] HotScalacBenchmark.compile:compile·p1.00                           sample       1451.229          ms/op
```


src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -696,21 +696,23 @@ trait Types
      *  }}}
      */
     def memberInfo(sym: Symbol): Type = {
-//      assert(sym ne NoSymbol, this)
+      require(sym ne NoSymbol, this)
```
l0rinc: this seems to be more than an optimization
@adriaanm: https://github.com/paplorinc/scala/commit/1f6f7f8aa94c622665a35343de8108ea66a787b7#diff-ecaa4198a04261cd0e00a708b695eafbL689

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -1106,13 +1109,15 @@ trait Types
...
-/*
-    override def typeOfThis: Type = typeSymbol.typeOfThis
-    override def bounds: TypeBounds = TypeBounds(this, this)
-    override def prefix: Type = NoType
-    override def typeArgs: List[Type] = List()
-    override def typeParams: List[Symbol] = List()
-*/
+    override def boundSyms: Set[Symbol] = emptySymbolSet
+
+    /*
```
l0rinc: minor: you may want to delete commented code instead of formatting it

src/reflect/scala/reflect/internal/util/Collections.scala
```diff
@@ -61,6 +61,19 @@ trait Collections {
     head
   }
 
+  final def sameElementsEquals(thiss: List[AnyRef], that: List[AnyRef]): Boolean = {
+    // Probably immutable, so check reference identity first (it's quick anyway)
+    (thiss eq that) || {
+      var these = thiss
+      var those = that
+      while (!these.isEmpty && !those.isEmpty && these.head.equals(those.head)) {
```
l0rinc, line 69: minor: could probably be `while (these.nonEmpty && those.nonEmpty`

src/reflect/scala/reflect/internal/util/Collections.scala
```diff
@@ -61,6 +61,19 @@ trait Collections {
     head
   }
 
+  final def sameElementsEquals(thiss: List[AnyRef], that: List[AnyRef]): Boolean = {
+    // Probably immutable, so check reference identity first (it's quick anyway)
+    (thiss eq that) || {
```
l0rinc, line 66: question: this inlining of `thiss.sameElements(that)` seems like a micro-optimization to my untrained eye :)

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -2135,14 +2135,17 @@ trait Types
...
       import scala.util.hashing.MurmurHash3._
-      val hasArgs = args ne Nil
       var h = productSeed
       h = mix(h, pre.hashCode)
       h = mix(h, sym.hashCode)
-      if (hasArgs)
-        finalizeHash(mix(h, args.hashCode()), 3)
-      else
-        finalizeHash(h, 2)
+      var i = 0
```
l0rinc: could be inited with `2` to avoid `2 + i` later

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -2135,14 +2135,17 @@ trait Types
...
+      var elem = args
```
l0rinc: minor: rename to `elems` instead?

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -829,7 +829,7 @@ self =>
       def mkNamed(args: List[Tree]) = if (isExpr) args map treeInfo.assignmentToMaybeNamedArg else args
       val arguments = right match {
         case Parens(args) => mkNamed(args)
-        case _            => List(right)
+        case _            => right :: Nil
```
l0rinc, line 832: the first is more readable, is the second that much faster?

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -1076,7 +1076,8 @@ abstract class ClassfileParser {
         cls setInfo completer
         mod setInfo completer
         mod.moduleClass setInfo loaders.moduleClassLoader
-        List(cls, mod.moduleClass) foreach (_.associatedFile = file)
+        cls.associatedFile = file
```
l0rinc, line 1092: if you want to retain the previous structure, you could probably
```scala
(cls.associatedFile, mod.moduleClass.associatedFile) = (file, file)
```

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -184,8 +184,8 @@ private[internal] trait GlbLubs {
   /** Eliminate from list of types all elements which are a supertype
     *  of some other element of the list. */
   private def elimSuper(ts: List[Type]): List[Type] = ts match {
```
l0rinc, line 186: I don't claim to understand it, but I have a question:
This seems to be an `O(n^2)` operation (i.e. for every element search all remaining). Would it be faster to cache the parents in a `Set` and check the set for inconsistencies for the previously encountered parents instead?
That could make it linear, I think.
(would probably be slower if the number of elements is small)

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -195,8 +195,8 @@ private[internal] trait GlbLubs {
     *  of some other element of the list. */
   private def elimSub(ts: List[Type], depth: Depth): List[Type] = {
     def elimSub0(ts: List[Type]): List[Type] = ts match {
-      case List() => List()
-      case List(t) => List(t)
+      case List() => ts
```
l0rinc, line 197: could probably be:
```scala
case List() | List(_) => ts
```
instead

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
```diff
@@ -101,11 +101,16 @@ private[internal] trait TypeMaps {
       case tr @ TypeRef(pre, sym, args) =>
         val pre1 = this(pre)
         val args1 = (
-          if (trackVariance && args.nonEmpty && !variance.isInvariant && sym.typeParams.nonEmpty)
-            mapOverArgs(args, sym.typeParams)
-          else
+          if (trackVariance && args.nonEmpty && !variance.isInvariant) {
+            val tparams = sym.typeParams
```
l0rinc, line 105: minor: to make it more obvious what happened to the `sym.typeParams.nonEmpty` part of the `if`, this could probably be simplified to
```scala
if (sym.typeParams.nonEmpty) mapOverArgs(args, sym.typeParams)
else args mapConserve this
```

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
```diff
@@ -722,9 +722,20 @@ private[internal] trait TypeMaps {
       else subst(tp, sym, from.tail, to.tail)
       )
 
+    private def fromContains(syms: Set[Symbol]): Boolean = {
+      if (!syms.isEmpty) {
```
l0rinc: minor: can probably be `syms.nonEmpty && {`

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2361,15 +2367,16 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
     /** Returns all symbols overridden by this symbol. */
     final def allOverriddenSymbols: List[Symbol] = {
-      def loop(xs: List[Symbol]): List[Symbol] = xs match {
-        case Nil     => Nil
+      @tailrec
+      def loop(xs: List[Symbol], result: List[Symbol]): List[Symbol] = xs match {
+        case Nil     => result
         case x :: xs =>
```
l0rinc, line 2381: minor: could also be `y :: ys` to avoid name collision with parent match

l0rinc (approved):

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -3604,7 +3604,10 @@ trait Types
       if (sym.isAliasType && sameLength(sym.info.typeParams, args) && !sym.lockOK)
         throw new RecoverableCyclicReference(sym)
 
-      TypeRef(pre, sym, args)
+      if ((args eq Nil) && (pre eq NoPrefix))
+        sym.tpeHK // opt lean on TypeSymbol#tyconCache, rather than interning a type ref.
```
lrytz, line 3621: Could this be pushed further down into `TypeRef.apply`?

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -1085,7 +1085,8 @@ trait Types
     override def baseTypeSeq: BaseTypeSeq = supertype.baseTypeSeq
     override def baseTypeSeqDepth: Depth = supertype.baseTypeSeqDepth
     override def baseClasses: List[Symbol] = supertype.baseClasses
-  }
+    override def boundSyms: Set[Symbol] = emptySymbolSet
```
lrytz: since `mapOver` recurses into components, it seems there are some unnecessary calls to `resultType.boundSyms`:
```
MethodType
    override def boundSyms = resultType.boundSyms ++ params

PolyType
    override def boundSyms = immutable.Set[Symbol](typeParams ++ resultType.boundSyms: _*)

```
and an unnecessary override in `NullaryMethodType`

```
    override def boundSyms = resultType.boundSyms
```

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -3788,7 +3788,7 @@ trait Types
   private var uniques: util.WeakHashSet[Type] = _
   private var uniqueRunId = NoRunId
 
-  protected def unique[T <: Type](tp: T): T = {
+  protected def unique[T <: Type](tp: T): T =  {
```
lrytz, line 3804: ` `

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2928,6 +2938,34 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     // are a case accessor (you can also be a field.)
     override def isCaseAccessorMethod = isCaseAccessor
 
+    def typeAsMemberOf(pre: Type): Type = {
```
lrytz:
```
Found 2 binary incompatibilities (22 were filtered out)
 * abstract synthetic method
   scala$reflect$runtime$SynchronizedSymbols$SynchronizedMethodSymbol$$$outer()scala.reflect.runtime.SynchronizedSymbols
   in interface
   scala.reflect.runtime.SynchronizedSymbols#SynchronizedMethodSymbol is
   present only in current version
 * abstract synthetic method
   scala$reflect$runtime$SynchronizedSymbols$SynchronizedMethodSymbol$$super$typeAsMemberOf(scala.reflect.internal.Types#Type)scala.reflect.internal.Types#Type
   in interface
   scala.reflect.runtime.SynchronizedSymbols#SynchronizedMethodSymbol is
   present only in current version
```

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -23,9 +23,6 @@ trait Namers extends MethodSynthesis {
   import global._
```
lrytz: i cannot find the commit referenced in the commit message (8dc4dbf0edde0f83b76c6a31d1f741689fa2e67b), i guess it's 1f6f7f8aa94c622665a35343de8108ea66a787b7

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2928,6 +2938,34 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
...
+      // in place until after the compiler has completed the typer phase.
+      //
+      // Out of caution, I've also disable caching if there are active type completers, which also
+      // mutate symbol infos during val and def return type inference based the overriden member.
+      if (isPastTyper || lockedCount > 0) return pre.computeMemberType(this)
+
+      if (mtpeRunId == currentRunId) {
+        if ((mtpePre eq pre) && (mtpeInfo eq info)) return {
+          mtpeResult
+        }
```
lrytz: could merge the two `if`s, also the return block is a bit funny

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -696,21 +696,23 @@ trait Types
...
-    def memberType(sym: Symbol): Type = sym.tpeHK match {
-      case OverloadedType(_, alts) => OverloadedType(this, alts)
+    def memberType(sym: Symbol): Type = sym match {
+      case meth: MethodSymbol =>
+        meth.typeAsMemberOf(this)
+      case _ =>
+        computeMemberType(sym)
+    }
+
+    def computeMemberType(sym: Symbol): Type = sym.tpeHK match { //@M don't prematurely instantiate higher-kinded types, they will be instantiated by transform, typedTypeApply, etc. when really necessary
```
lrytz: the comment can probably be dropped, since adriaan removed it in 1f6f7f8aa94c622665a35343de8108ea66a787b7

lrytz (commented):
Looking good!

src/reflect/scala/reflect/internal/util/Collections.scala
retronym: I've got a bit of an aversion to `nonEmpty` (a mixed in trait method) in hot code because in some cases it is harder for JIT to devirtualize and inline. Not sure whether that's the case here.

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
retronym: Yeah, we're lacking an optimization on `List.apply` to avoid boxing an array to pass as varargs. We've got a [ticket open](https://github.com/scala/scala-dev/issues/341) for that improvement, but in the meantime I'm happy enough with the readability of `::` to use it in the few places where this makes a material difference.

src/reflect/scala/reflect/internal/util/Collections.scala
retronym: The key difference is using `.equals` rather than `.==`. The latter is routed through `BoxesRuntime.equals` which needs to perform a few type tests on the operands before delegating to `equals` in order to paper over the differences between primitive and boxed numeric equality.

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
retronym: `Type`-s aren't amenable to use in a `HashSet` because `=:=` isn't compatible with `hashCode`.  This code here is actually using subtyping, which makes it harder again to find ways to beat `O(N^2)`.

src/reflect/scala/reflect/internal/Types.scala
retronym: Yep, I agree. Things seem to work with your proposed changes. I'm pushed a commit that actually avoids calling `boundSyms` altogether, which avoids the overhead of creating sets just to check if one of the `from` symbols exists in there, I'm pretty sure it will be faster to iterate over, e.g, `MethodSymbol#typeParams` directly and compare search for element to `from`.

We can actually short-circuit that search in many cases by pre-computing the range of symbol ids in `from` and by noting whether it has any term symbols.

src/reflect/scala/reflect/internal/Types.scala
retronym: IIRC, we can't (easily) do that because `tpeHK` calls that `TypeRef.apply`, so we'd loop.

src/reflect/scala/reflect/internal/Types.scala
lrytz: Cool! `boundSyms` already confused me not long ago (I don't remember the details).

src/reflect/scala/reflect/internal/Types.scala
lrytz: OK, I see.. We could clean up TypeRef construction one day.

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
```diff
@@ -703,18 +703,30 @@ private[internal] trait TypeMaps {
...
     protected def toType(fromtp: Type, tp: T): Type
 
     protected def renameBoundSyms(tp: Type): Type = tp match {
-      case MethodType(ps, restp) =>
-        createFromClonedSymbols(ps, restp)((ps1, tp1) => copyMethodType(tp, ps1, renameBoundSyms(tp1)))
-      case PolyType(bs, restp) =>
-        createFromClonedSymbols(bs, restp)((ps1, tp1) => PolyType(ps1, renameBoundSyms(tp1)))
-      case ExistentialType(bs, restp) =>
+      case MethodType(ps, restp) if fromHasTermSymbol && fromContains(ps) =>
+        createFromClonedSymbols(ps, restp)((ps1, tp1) => copyMethodType(tp, ps1, tp1))
```
lrytz: I don't understand why we don't need the recursive calls to `renameBoundSyms` anymore.

retronym: The result of this method is recursed into with `mapOver`, so we'll get back to here with the result type.

retronym:
My measurements suggest that the last two commits (removing `boundSyms`, and avoiding futile comparisons of eta-expanded type refs in `<:<` and `=:=`) offer a further 2% speedup. 🏎

lrytz:
Very nice!

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2931,6 +2943,31 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
...
 
+    def typeAsMemberOf(pre: Type): Type = {
+      // We used to cache member types more pervasively, but we can't get away with that
+      // any more because of t8011.scala, which demonstrates a problem with the extention methods
+      // phase. As it moves a method body to an extension methid in the companion, it substitutes
+      // the new type parameter symbols into the method body, which mutates the base type sequence of
+      // a local class symbol. We can no longer assume that`mtpePre eq pre` is a sufficient condition
+      // to use the cached result here.
+      //
+      // Rather than throwing away the baby with the bathwater, lets at least try to keep the caching
```
adriaanm: 👍

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -696,21 +698,23 @@ trait Types
...
+        meth.typeAsMemberOf(this)
+      case _ =>
+        computeMemberType(sym)
+    }
+
+    def computeMemberType(sym: Symbol): Type = sym.tpeHK match {
+      case OverloadedType(_, alts) =>
+        OverloadedType(this, alts)
       case tp =>
-        // Correct caching is nearly impossible because `sym.tpeHK.asSeenFrom(pre, sym.owner)`
```
adriaanm: Is this comment worth keeping around?

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4239,38 +4259,44 @@ trait Types
...
-         (!symLo.hasVolatileType || symHi.hasVolatileType || tpHi.isWildcard)) // sub-member must not introduce volatility
-      else if (symHi.isAbstractType)
-        ((tpHi.bounds containsType tpLo) &&
-         kindsConform(symHi :: Nil, tpLo :: Nil, preLo, symLo.owner))
-      else // we know `symHi.isAliasType` (see above)
-        tpLo =:= tpHi
+      val symHiInfo = symHi.info
+      if (symHiInfo == WildcardType) {
+        if (symHi.isTerm) (!symHi.isStable || symLo.isStable)     // sub-member must remain stable
+        else true
```
adriaanm: Are we guaranteed `symHi.isAbstractType` in this case? Can an alias type's `info` ever be a wildcard? (If so, the lower symbol's info must also be a wildcard)

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -360,10 +364,22 @@ trait TypeComparers {
...
+    }
 
     (    (tp1.typeSymbol eq NothingClass)       // @M Nothing is subtype of every well-kinded type
       || (tp2.typeSymbol eq AnyClass)           // @M Any is supertype of every well-kinded type (@PP: is it? What about continuations plugin?)
-      || isSub(tp1.normalize, tp2.normalize) && annotationsConform(tp1, tp2)  // @M! normalize reduces higher-kinded case to PolyType's
+      || (if (isTrivialTypeRef(tp1) && isTrivialTypeRef(tp2)) {
+            val r1 = tp1.typeSymbolDirect.isNonBottomSubClass(tp2.typeSymbolDirect) // OPT faster than comparing eta-expanded types
+            val r2 = isSub(tp1.normalize, tp2.normalize)
+            if (r1 != r2)
+              getClass
```
adriaanm: stray debug artifact?

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
adriaanm: Would be good to elaborate on this in a comment

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
```diff
@@ -698,18 +703,30 @@ private[internal] trait TypeMaps {
     // OPT this check was 2-3% of some profiles, demoted to -Xdev
     if (isDeveloper) assert(sameLength(from, to), "Unsound substitution from "+ from +" to "+ to)
 
+    private[this] var fromHasTermSymbol = false
+    private[this] var fromMin = Int.MaxValue
```
adriaanm: Kind of like a mini-bloom filter?

src/reflect/scala/reflect/internal/Types.scala
som-snytt: The first thing I'm going to type tomorrow is `// is this comment worth keeping?`  I hope I don't accidentally read it six months from now, because it would blow my brain i.e. mind.  The other one I should plant for my future self is `// what were you thinking?`

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
retronym: Funny, I was re-reading about bloom filters while making this change. I guess this is kindred (we can cheaply determine non-existence in a set without iterating the full set), but OTOH we there isn't any superimposition of information.

I did learn about an [Zatocoding](https://en.wikipedia.org/wiki/Superimposed_code) in my travels, sort of an analog bloom filter for librarians.

src/reflect/scala/reflect/internal/Types.scala
retronym: The common case that I'm trying to optimize is when `symHi` is the member of a prototype for `HasMember` (e.g `{ def extensionMethod: ?}`)`. I'll tweak this to restrict this new fast path to `symHiInfo == WildcardType && symHi.isTerm` and let types flow down the the logic below.

src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
retronym: > The result of this method is recursed into with mapOver, so we'll get back to here with the result type.

Actually, taking another look at this, I think there is a potential problem with existential types, as `newExistentialType` collapses nested existential types. If some construction of existentials snuck past that, and an `ExistentialType(qs1, ExistentialType(qs2, res))` appeared here, we'd fail to look at `qs2`.

```scala
  /** A creator for existential types which flattens nested existentials.
   */
  def newExistentialType(quantified: List[Symbol], underlying: Type): Type =
    if (quantified.isEmpty) underlying
    else underlying match {
      case ExistentialType(qs, restpe) => newExistentialType(quantified ::: qs, restpe)
      case _                           => ExistentialType(quantified, underlying)
    }
```

As per @adriaanm's offline suggestion, I'm going to split this change into a separate PR and address the review comments there.

retronym: Actually, the old code was already assuming existentials has been flattened.

src/reflect/scala/reflect/internal/Types.scala
retronym: I've pushed this change now.

src/reflect/scala/reflect/internal/Types.scala
retronym:
```
// Ceci n'est pas un commentaire
```

som-snytt: `// comment?`

retronym:
I believe I've addressed the review comments here.

l0rinc (approved):
Awesome job, please update the description with the latest stats

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -180,7 +180,11 @@ trait TypeComparers {
     !settings.isScala211 || methodHigherOrderTypeParamsSameVariance(low, high) || low.variance.isInvariant
 
   def isSameType2(tp1: Type, tp2: Type): Boolean = {
-    def retry(lhs: Type, rhs: Type) = ((lhs ne tp1) || (rhs ne tp2)) && isSameType(lhs, rhs)
+    def retry(lhs: Type, rhs: Type) = ((lhs ne tp1) || (rhs ne tp2)) && {
+      def sameSyms(tp1: Type, tp2: Type) = tp1.typeSymbolDirect eq tp2.typeSymbolDirect
+      val skipRetry = lhs.isInstanceOf[PolyType] && rhs.isInstanceOf[PolyType] && sameSyms(tp1, lhs) && sameSyms(tp2, rhs)
```
adriaanm: Could you convince me that we don't need to take into account potential differences in type params / type application due to normalization? I'm not sure how, but I'm also not sure it's impossible that `normalizePlus` could take e.g., `[T, U](T, U)` to `[T, U](U, T)` for one of `rhs`/`lhs`, so that now `[T,U]Swap[U,T] =:= [T,U]Tuple2[T,U]`

Also, this feels easier to follow:
```
      val skipRetry =
        (lhs eq tp1) && (rhs eq tp2) ||
          (lhs.isInstanceOf[PolyType] && rhs.isInstanceOf[PolyType] && sameSyms(tp1, lhs) && sameSyms(tp2, rhs))

      !skipRetry && isSameType(lhs, rhs)
```

adriaanm (commented):
The subtype change LGTM, but have a question on the sameType one.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
retronym: I've reworked that commit a to be more in line with the subtyping change (computing `skip` before actually calling `normalize`). Given that it is pretty subtle, let's take our time to review it in #5885, rather than in this omnibus.

retronym:
@paplorinc I've added the benchmark results to the description.

l0rinc:
@retronym the `cnt` seems to be different for the two runs, is that expected? I don't know how to compare them this way.

retronym:
The benchmark runs for a fixed duration, so after the speed up then count of executions is higher

The results suggest a 3.5% speed up. 1251/1299

l0rinc:
Thanks, good job!

retronym:
@adriaanm I've resolved the merge conflict here.

retronym:
ping @adriaanm for a (final?) review round here; ditto for #5875 #5882.

adriaanm (approved):

som-snytt:
I don't know what I'm going to do with my 3.5%. Probably catch up on Supergirl.

PR #5824, opened 2017-04-04
Avoid excessive IO in classfile parser

retronym:
The class file parser, used to read the java-defined classes from
the classpath, includes logic to search the output path for a .java
source file that corresponds to the value of the SourceFile attribute.

I haven't been able to figure out the rationale for that fix, but
it involves a non-neglible overhead, so this commits disables it
in the batch compiler.

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
@@ -872,13 +872,26 @@ abstract class ClassfileParser {
           parseExceptions(attrLen)
 
         case tpnme.SourceFileATTR =>
-          val srcfileLeaf = readName().toString.trim
-          val srcpath = sym.enclosingPackage match {
-            case NoSymbol => srcfileLeaf
-            case rootMirror.EmptyPackage => srcfileLeaf
-            case pkg => pkg.fullName(File.separatorChar)+File.separator+srcfileLeaf
-          }
-          srcfile0 = settings.outputDirs.srcFilesFor(in.file, srcpath).find(_.exists)
+          if (forInteractive) {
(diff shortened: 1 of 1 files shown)


mkeskells:
@retronym Nice fix
slightly related - have you ever run with a security manager installed - logging all file opens would be a systematic way to spot all of the IO

lrytz (approved):

retronym:
> slightly related - have you ever run with a security manager installed - logging all file opens would be a systematic way to spot all of the IO

@mkeskells I've used YourKit to identify the IO bottlenecks so far. This one was a bit of surprise!

dragos:
@retronym I don't think that needs to stay in the IDE code path either. I think this is the ticket (at least, it links to the right commit on scalac): https://scala-ide-portfolio.assembla.com/spaces/scala-ide/tickets/realtime_cardwall?ticket=2689

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -872,13 +872,26 @@ abstract class ClassfileParser {
...
+            // opt: disable this code in the batch compiler for performance reasons.
+            // it appears to be looking for the .java source file mentioned in this attribute
+            // in the output directories of scalac.
+            //
+            // References:
+            // https://issues.scala-lang.org/browse/SI-2689
+            // https://github.com/scala/scala/commit/7315339782f6e19ddd6199768352a91ef66eb27d
+            // https://github.com/scala-ide/scala-ide/commit/786ea5d4dc44065379a05eb3ac65d37f8948c05d
+            //
+            // TODO: can we disable this altogether? Does Scala-IDE actually intermingle source and classfiles in a way
```
fommil, line 885: we don't do anything funky in ensime that rings a bell here

PR #5825, opened 2017-04-04
per run immutable settings

mkeskells:
settings are mutable vars

within a run they dont change, so create a data structure with immutable data that is optimised for the read access

this build on work from @retronym 

based on the normal test config - this is a minor cpu gain. I expected to see some allocation gain, but I presume this is optimized away

based on the results below this is about a 2.8% gain based on the mean of after 10 90%

```

ALL
              00_baseline	60	 11620.94 [+2.46% -0.82%]	 11064.06 [+2.42% -0.85%]	  2793.84 [+1.11% -0.99%]
10_perRunImmutableSettings	60	 10728.01 [+2.59% -0.89%]	 10515.10 [+2.47% -0.90%]	  2798.69 [+1.11% -0.99%]

after 10 90%
              00_baseline	46	 10212.12 [+1.37% -0.93%]	  9996.60 [+1.25% -0.94%]	  2783.93 [+1.00% -1.00%]
10_perRunImmutableSettings	46	  9919.79 [+1.03% -0.96%]	  9780.57 [+1.03% -0.96%]	  2788.79 [+1.00% -1.00%]
```
here is a breakdown of the phases that seem to benefit most
```
after 10 90% typer, no GC
              00_baseline	42	  2620.33 [+1.40% -0.92%]	  2598.59 [+1.30% -0.92%]	   584.64 [+1.00% -1.00%]
10_perRunImmutableSettings	43	  2562.37 [+1.05% -0.95%]	  2556.32 [+1.05% -0.95%]	   583.26 [+1.00% -1.00%]

after 10 90% patmat, no GC
              00_baseline	44	   447.58 [+1.22% -0.87%]	   444.96 [+1.19% -0.88%]	   109.19 [+1.00% -1.00%]
10_perRunImmutableSettings	42	   428.68 [+1.18% -0.89%]	   421.88 [+1.19% -0.85%]	   110.05 [+1.00% -1.00%]


after 10 90% pickler, no GC
              00_baseline	46	   121.46 [+1.12% -0.91%]	   122.28 [+1.15% -0.77%]	    20.84 [+1.00% -1.00%]
10_perRunImmutableSettings	46	    93.08 [+1.43% -0.79%]	    92.73 [+1.52% -0.67%]	    20.82 [+1.00% -1.00%]

after 10 90% jvm, no GC
              00_baseline	37	  2725.29 [+1.24% -0.93%]	  2665.54 [+1.13% -0.94%]	   617.55 [+1.00% -1.00%]
10_perRunImmutableSettings	37	  2584.80 [+1.05% -0.97%]	  2559.54 [+1.05% -0.97%]	   617.08 [+1.00% -1.00%]

```

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -75,7 +75,6 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
   // alternate constructors ------------------------------------------
 
   override def settings = currentSettings
-  def currentRunSettings = curRun.runSettings
 
   /** Switch to turn on detailed type logs */
   var printTypings = settings.Ytyperdebug.value
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1075,7 +1075,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
 
   /** A Run is a single execution of the compiler on a set of units.
    */
-  class Run extends RunContextApi with RunReporting with RunParsing {
+  class Run extends RunContextApi with RunReporting with RunParsing with RunSettings {
     /** Have been running into too many init order issues with Run
      *  during erroneous conditions.  Moved all these vals up to the
      *  top of the file so at least they're not trivially null.
@@ -1540,6 +1540,12 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     }
   } // class Run
 
+
+  class PerRunSettings extends PerRunSettingsBase {
+    override val isScala211: Boolean = settings.isScala211
+    override val isScala212: Boolean = settings.isScala212
+  }
+  override protected def PerRunSettings: PerRunSettings = new PerRunSettings
   def printAllUnits() {
     print("[[syntax trees at end of %25s]]".format(phase))
     exitingPhase(phase)(currentRun.units foreach { unit =>
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -274,7 +274,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
       body
   }
 
-  override protected def isDeveloper = settings.developer || super.isDeveloper
+  override protected def isDeveloper = settings.developer.boolValue || super.isDeveloper
 
   /** This is for WARNINGS which should reach the ears of scala developers
    *  whenever they occur, but are not useful for normal users. They should
@@ -303,7 +303,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
   }
 
   @inline final override def debuglog(msg: => String) {
-    if (settings.debug)
+    if (settings.debug.boolValue)
       log(msg)
   }
 
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1545,9 +1545,10 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
 
 
   class PerRunSettingsGlobal extends PerRunSettings {
-    override val isScala211: Boolean = settings.isScala211
-    override val isScala212: Boolean = settings.isScala212
-    val isScala213: Boolean = settings.isScala213
+    override val isScala213: Boolean = settings.isScala213
+    override val isScala212: Boolean = isScala213 || settings.isScala212
+    override val isScala211: Boolean = isScala212 || settings.isScala211
+
     val debug = settings.debug.boolValue
 
     override val Xexperimental = settings.Xexperimental
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1551,6 +1551,8 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     val isScala213: Boolean = settings.isScala213
     val debug = settings.debug.boolValue
 
+    override val Xexperimental = settings.Xexperimental
+
     val current = settings
     //optimise settings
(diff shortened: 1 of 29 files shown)


src/compiler/scala/tools/nsc/settings/MutableSettings.scala
```diff
@@ -455,6 +455,7 @@ class MutableSettings(val errorFn: String => Unit)
     type T = Boolean
     protected var v: Boolean = false
     override def value: Boolean = v
+    def boolValue: Boolean = v
```
som-snytt, line 458: There is an implicit conversion from BooleanSetting to Boolean that obviates a member. It currently resides in `reflect.internal.settings.MutableSettings`. As a matter of taste, `boolValue` is not my favorite name. :)

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -1540,6 +1543,72 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     }
   } // class Run
 
+
+  class PerRunSettingsGlobal extends PerRunSettings {
+    override val isScala211: Boolean = settings.isScala211
+    override val isScala212: Boolean = settings.isScala212
+    val isScala213: Boolean = settings.isScala213
```
som-snytt: These tests are cheap now. https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala#L90

som-snytt:
I would be wary of indirections that make settings a maintenance burden. The reflect layer of settings certainly is burdensome. 

For naming, I'd suggest using `settings` for the thing that is routinely consulted, such as the immutable per-run settings. I would name the rarely-used mutable thing something else, such as `settings0` or `config`.

src/compiler/scala/tools/nsc/util/RunnableInPhase.scala
```diff
@@ -0,0 +1,34 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, Awaitable, Future}
+import scala.tools.nsc.{Global, Phase}
+object RunnableInPhase {
+  val idGen = new AtomicInteger
+}
+/**
+  * a wrapper to allow Runnables to be associated to a phase. Primarily useful for profiling
```
lrytz, line 12: Can you add a detailed example how this class can be used for profiling?

lrytz:
> I would be wary of indirections that make settings a maintenance burden. The reflect layer of settings certainly is burdensome.

I have to agree, the current organization of settings is already hard to follow, this adds another dimension.. Some things are really not very pretty, e.g., `override val YoptTrace_isSetByUser: Boolean = settings.YoptTrace.isSetByUser`.

How about something along these lines. It would keep things together in one place.

```scala
import collection.mutable

class Settings {
  val debugSetting = BooleanSetting("-Ydebug", "...", false)
  var debug: Boolean = debugSetting.cached(debugSetting.value)(debug = _)
}

trait Setting[T] {
  def name: String
  def desc: String
  def default: T

  private[this] var _value: T = _
  final def value = _value
  final def value_=(v: T): Unit = {
    _value = v
    updateCached.foreach(_.apply())
  }

  private val updateCached: mutable.Set[() => Unit] = mutable.Set.empty

  def cached[T](get: => T)(setVar: T => Unit) = {
    updateCached += { () => setVar(get) }
    get
  }
}

case class BooleanSetting(name: String, desc: String, default: Boolean) extends Setting[Boolean]
```

```scala
scala> val s = new Settings

scala> s.debug
res0: Boolean = false

scala> s.debugSetting.value = true

scala> s.debug
res1: Boolean = true
```

src/compiler/scala/tools/nsc/settings/MutableSettings.scala
mkeskells: This is a change from @retronym, with a comment "Avoid needless Boolean.unbox in hot code paths."

All of this PR is non functional changes, so the implicit function is not really a concern. I think this is a CPU vs memory trade off, an extra field on a handful of objects against N calls and M call sites

That said it is what I inherited and I have not looked into that the effect is in either the generated code or the runtime performance, but I am sure that @retronym can comment as to if this change is useful or not

What would be the preferred name if this access is kept?

src/compiler/scala/tools/nsc/util/RunnableInPhase.scala
mkeskells: see https://github.com/scala/scala/pull/5815 for use - this enables the profiler to be aware of the background thread
profiler is in https://github.com/scala/scala/pull/5760 (early version) and current working version in https://github.com/rorygraves/scalac_perf/tree/2.12.x_profile_baseline

when used with -Yprofile-enable:true

this is a pre-req for the profiler and not used directly in this PR

the profiler generates the figures that I quoted in the PR header

src/compiler/scala/tools/nsc/settings/MutableSettings.scala
som-snytt: OK, boxing, thanks. It would take years of research to determine a good name. Maybe `asBoolean`. Is it true that the inlined impicit could return `s.asBoolean` and you wouldn't see the name in that usage.

mkeskells:
@lrytz looks like that pattern would have the same effect @retronym any thoughts?

src/compiler/scala/tools/nsc/settings/MutableSettings.scala
mkeskells: @som-snytt I think that the alternative below from @lrytz may avoid the boxing as well, and be a sumpler change as this should not affect user code - it could just change BooleanSetting etc

Will have a look when I can get other PRS unstalled

mkeskells:
will come back to this later with a different approach

PR #5826, opened 2017-04-04
WIP: Deduplicate safeToInt - Approach undecided - comments please

janekdb:
Not ready for merge.

I started looking at this,

```
  // XXX way past time for this to be central
  def parseInt(x: String): Option[Int] =
    try   { Some(x.toInt) }
    catch { case _: NumberFormatException => None }
```

Along the way I found one near duplication of this could be removed without compile or test errors.

I'm looking for some guidance on whether the object commented out in this PR should be removed or not.

### DIFF
--- a/src/compiler/scala/tools/cmd/FromString.scala
+++ b/src/compiler/scala/tools/cmd/FromString.scala
@@ -22,44 +22,45 @@ abstract class FromString[+T](implicit m: OptManifest[T]) extends PartialFunctio
   def targetString: String = m.toString
 }
 
-object FromString {
+//object FromString {
   // We need this because we clash with the String => Path implicits.
-  private def toDir(s: String)  = new Directory(new java.io.File(s))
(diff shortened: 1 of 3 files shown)


lrytz:
@janekdb the code in `scala.tools.cmd` should probably live somewhere else? As far as i can see only `object CommandLineParser` is used by the compiler. It seems to be used in partest, https://github.com/scala/scala-partest/blob/1.1.x/src/main/scala/scala/tools/partest/nest/RunnerSpec.scala, not sure if anywhere else.

janekdb:
@lrytz - Thanks for comments. Have now discovered `sbt testAll`. Closing for now.

janekdb:
Having now also discovered the paulp command line tool builder functionality buried inside the compiler I am wondering if this code should be extracted serving two purposes,

1. Reduce compiler code base
2. Provide option to package this into a standalone library which would provide a way to advertize it's existence.

PR #5828, opened 2017-04-05
`CompleterWrapper` delegates `typeParams`.

adriaanm:
Fixes the problem reported with #5730 by @xuwei-k and @sjrd in scala/scala-dev#352.


lrytz (approved):

lrytz:
Maybe we can think of a different design for 2.12 that avoids this trap..

adriaanm:
Will work on improving my fix for 2.12.2 today!

PR #5831, opened 2017-04-05
optimise Btypes descriptor generation

mkeskells:
part of https://github.com/scala/scala/pull/5815 isolated to aid review

performance result look reasonable. The 2 results are for the 2 commits
```
after 10 90% jvm, no GC
             14_cacheDesc	38	  2336.37 [+1.03% -0.96%]	  2321.96 [+1.03% -0.96%]	   618.26 [+1.00% -1.00%]
         15_cacheDesc_ASM	37	  2325.06 [+1.03% -0.97%]	  2302.36 [+1.02% -0.97%]	   608.97 [+1.00% -1.00%]

```

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -478,20 +478,7 @@ abstract class BTypes {
      *  - for an OBJECT type, the 'L' and ';' are not part of the range of the created Type
      *  - for an ARRAY type, the full descriptor is part of the range
      */
-    def toASMType: asm.Type = this match {
-      case UNIT   => asm.Type.VOID_TYPE
-      case BOOL   => asm.Type.BOOLEAN_TYPE
-      case CHAR   => asm.Type.CHAR_TYPE
-      case BYTE   => asm.Type.BYTE_TYPE
-      case SHORT  => asm.Type.SHORT_TYPE
-      case INT    => asm.Type.INT_TYPE
-      case FLOAT  => asm.Type.FLOAT_TYPE
-      case LONG   => asm.Type.LONG_TYPE
-      case DOUBLE => asm.Type.DOUBLE_TYPE
-      case ClassBType(internalName) => asm.Type.getObjectType(internalName) // see (*) above
-      case a: ArrayBType            => asm.Type.getObjectType(a.descriptor)
-      case m: MethodBType           => asm.Type.getMethodType(m.descriptor)
-    }
+    def toASMType: asm.Type
 
     def asRefBType       : RefBType       = this.asInstanceOf[RefBType]
     def asArrayBType     : ArrayBType     = this.asInstanceOf[ArrayBType]
@@ -499,7 +486,9 @@ abstract class BTypes {
     def asPrimitiveBType : PrimitiveBType = this.asInstanceOf[PrimitiveBType]
   }
 
-  sealed abstract class PrimitiveBType(override val descriptor:String) extends BType {
+  sealed abstract class PrimitiveBType(
+      override val descriptor:String,
+      override val toASMType: asm.Type) extends BType {
     private[BTypes] override def buildDescriptor(sb: java.lang.StringBuilder): Unit = {
       sb.append(descriptor)
     }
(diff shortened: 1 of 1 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -298,29 +298,23 @@ abstract class BTypes {
...
+      *  - int[]: [I
+      *  - Object m(String s, double d): (Ljava/lang/String;D)Ljava/lang/Object;
+      */
+    def descriptor : String
+
+    /**
+      * generally used by descriptor to help build the descriptor. Allows several descriptors to be build from the same stringbuilder
+      * @param sb
+      */
+    private[BTypes] def buildDescriptor(sb: java.lang.StringBuilder): Unit
```
lrytz, line 324: I'd prefer to use a pattern match here instead of virtual dispatch. With an overload `override def descriptor(sb: java.lang.StringBuilder): String` in Class/Array/MethodBType this should be possible.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -848,6 +845,24 @@ abstract class BTypes {
    * infos using `get`, but it reports inliner warnings for missing infos that prevent inlining.
    */
   final case class ClassBType(internalName: InternalName) extends RefBType {
+
+    override def descriptor: String = {
+      if (descriptorCache eq null) {
+        val sb = new java.lang.StringBuilder(internalName.length + 2)
+        buildDescriptor(sb)
```
lrytz, line 848: there will probably be concurrent accesses to this method once we parallelize the backend, no?

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: Not sure what you are getting at
As I see it you either use a pattern match or virtual dispatch
Getting run of the 'this match { case ...} ' pattern in the code aids encapsulation, safety and performance IMHO

Why an overload? They are behaviourly different?

I am turn as to whether the buildDescriptor should return the descriptor  or generate i as a side effect. Either way there are 2 effects, and I could not come up with a a better name before my attention moved

lrytz: I think we generally prefer matching on `this` for sealed data types like `BType`

mkeskells: @lrytz I think that it is a very bad idea. I did see large amounts or this elsewhere, and it is on my todo list. Is this personal preference or lightbend policy?

My view - 

Given that you dont have warnings as fatal, it is fragile

From a perfomance perspective I would expect it to be non-optimal. Has this been benchmarked? I know that @retronym is a dab hand with JMH :-)
I would suggest that we need some figures to compare and determine what the effective costs and benefits are

Virtual dispatch, particularly from abstract classes rather than traits is very fast. Pattern matching is a sequential operation, slowed with unapply calls. Inner classes (e.g., from the global cake) make this even more onerous

I agree that is some cases it may be appropriate - code share, cold paths, but these paths are hot

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: Yes it would be concurrent, but I believe that it is threadsafe

I don't believe that the cost of lock is worth the potential gain in lost CPU

the answer is the same (==) so 2 threads race one will win. Small chance that we will have 2 thread with identical strings, but this is not harmful

There is no chance of a incomplete string being visible to another thread as String constructor has a memory fence

retronym:
I wasn't able to clearly measure an [improvement from this change](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=56daf3a%7C14132b4&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@) on our benchmark server.

That _might_ be because we compile with an oversized heap (initial/max=2G) to try to avoid GC noise in the benchmark run, which could tilt things in favour of the status quo.

We're still getting the hang of benchmarking, so this is just a datapoint, not a conclusive answer.

rorygraves:
I'm planning on re-profiling this one during the week - hopefully I can add some more clarity on it.

adriaanm:
What's the latest on the benchmarks? The window for 2.12.4 is closing soon (sign-off must happen by Wed Sep 27).

mkeskells:
I will not have time to met that cut off. Trying to work on the parallelization of the jvm first, but I doubt that will be ready in time either

adriaanm:
No problem, thanks for the update.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
retronym: > Pattern matching is a sequential operation, slowed with unapply calls

In general, your point is valid, and in some cases the performance does make a difference (e.g. in `mapOver` #5906 ), but it is important to note that most pattern matching is on case class constructors or plain old `instanceOfs` and doesn't involve unapply calls. So we should balance readability (which is sometimes enhanced with a pattern match) vs performance by confirming that there is an improvement in benchmarks from a refactoring.

adriaanm:
I'm closing this PR because it has been dormant for a while now. Happy to receive a resubmission!

PR #5832, opened 2017-04-05
optimise completeSilentlyAndCheckErroneous

mkeskells:
performance result - disappointing. I havent looked into the detail as it why - 
```
after 10 90% jvm, no GC
      11_completeSilently	39	  2540.75 [+1.03% -0.97%]	  2507.61 [+1.04% -0.95%]	   601.74 [+1.00% -1.00%]
```

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -233,7 +233,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
       sym.isErroneous
     }
 
-  @inline def withoutReporting[T](fn : => T) = {
+  @inline private def withoutReporting[T](fn : => T) = {
     val currentReporter = reporter
     reporter = NoReporter
     try fn finally reporter = currentReporter
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -12,6 +12,7 @@ import scala.tools.nsc.io.AbstractFile
 import GenBCode._
 import BackendReporting._
 import scala.reflect.internal.Flags
+import scala.tools.nsc.reporters.NoReporter
 
 /*
  *  Traits encapsulating functionality to convert Scala AST Trees into ASM ClassNodes.
@@ -228,16 +229,15 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
   def completeSilentlyAndCheckErroneous(sym: Symbol): Boolean =
     if (sym.hasCompleteInfo) false
     else {
-      val originalReporter = global.reporter
-      val storeReporter = new reporters.StoreReporter()
-      global.reporter = storeReporter
-      try {
-        sym.info
-      } finally {
-        global.reporter = originalReporter
-      }
+      withoutReporting(sym.info)
       sym.isErroneous
     }
+
+  @inline def withoutReporting[T](fn : => T) = {
+    val currentReporter = reporter
+    reporter = NoReporter
+    try fn finally reporter = currentReporter
+  }
 
 
   /*
--- a/src/compiler/scala/tools/nsc/reporters/NoReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/NoReporter.scala
@@ -1,0 +1,9 @@
+package scala.tools.nsc.reporters
+import scala.reflect.internal.util.Position
+
+/**
+  * A reporter that ignores reports
+  */
+object NoReporter extends Reporter{
+  override protected def info0(pos: Position, msg: String, severity: Severity, force: Boolean): Unit = ()
+}


lrytz:
How should I read the numbers? What's `Count`? The *MS times are larger, so does that mean that this patch caused a slight slowdown?

lrytz:
Also, the confidence intervals of the *MS numbers seem to be too large to show a measurable difference here.

mkeskells:
HI @lrytz 

the test is run 60 times- 

- we ignore the first 10 hence "after 10"
- we ignore the worst 10% ( outliers) hence "90%"  
- this is for the jvm phase only hence "jvm"
- we filter the result to ignore those with a GC that occurred within the run of the phase hence "no jvm"

the "count" column is the number of samples remaining after the above filtering

the "AllWallMS" is taken as the mean ( and ranges) for the start and end of the phase measured via System.nanotime

the "CPU_MS" in taken from the mxBean from the current thread. This is lower granularity, but useful for detecting IO pauses etc. We also gather the user mode time for the thread, but this is not displayed in the results above

"Allocated" is taken from the thread allocation counter, an indicates the number of bytes allocated in the heap expressed in MB

mkeskells:
as for the actual number reported - the Allocated show a slight reduction, about what I expect, but the CPU increase is unexpected and I will dig into this. It is on the border of confidence ( 1.6% mean increase, with a range in measurements of ~1%) 

I didn't expect to see much change in performance, but certainly not an increase

As the only change is  from 
```
     val originalReporter = global.reporter
      val storeReporter = new reporters.StoreReporter()		
      global.reporter = storeReporter		
      try {		
        sym.info		
      } finally {		
        global.reporter = originalReporter		
      }
```
to 
```
withoutReporting(sym.info)
...
 @inline def withoutReporting[T](fn : => T) = {
    val currentReporter = reporter
    reporter = NoReporter
    try fn finally reporter = currentReporter
  }
...
object NoReporter extends Reporter{
  override protected def info0(pos: Position, msg: String, severity: Severity, force: Boolean): Unit = ()
}
``` 
NoReporter does less work ( if called)
no constructor to call (so the expected minimal allocation reduction)
the orig code path refers to global.reporter, as opposed to the visible field
the only other difference is the use of  inline - I would have expected this to be optimized away

I would love to understand what is wrong with the rationale above

lrytz:
Thanks @mkeskells for these explanations. I'm wondering why did you start optimizing this particular spot, did it show up in profiling?

I think `withoutReporting` is not inlined because it's a non-final method in a non-final class. This is not so easy to see unfortunately because we have plenty of inliner warnings already, this just added one..

```
[info] Compiling 293 Scala sources to /home/jenkins/workspace/scala-2.12.x-validate-test/build/quick/classes/compiler...
...
[warn] there were 43 inliner warnings; re-run enabling -opt-warnings for details, or try -help
```

Check if you see the desired speedup after making the method final.

mkeskells:
@lrytz I dont recall the reason for the optimisation. Either it would have been showing up in the profiler or code inspection, I would presume the latter. This is just going through https://github.com/scala/scala/pull/5815 and pulling out all of the changes as separate reviewable items as requested

mkeskells:
@lrytz Not inlined - doh schoolboy error :-(  

too used to having warnings as fatal from the day job. Will retry, and add a longer run, and post the results when I have them

mkeskells:
Result after 200 cycles. _2 is with https://github.com/scala/scala/pull/5832/commits/e0d2830f838dcf8abe7bec1c9a0d528e9ebdd45c
```
after 10 90% jvm, no GC
              B0_baseline	130	  3334.91 [+1.06% -0.96%]	  3226.08 [+1.05% -0.95%]	   604.20 [+1.01% -1.00%]
     11_completeSilently_2	131	  3313.66 [+1.04% -0.96%]	  3206.35 [+1.03% -0.96%]	   602.04 [+1.01% -1.00%]
```

machine blue screened some time after this after so cpu figures are suspect. Will rerun

mkeskells:
reran. Not sure why the results have stepped (upgrades?). The comparison is the important part though
```
after 10 90% jvm, no GC
              B0_baseline	139	  3020.22 [+1.50% -0.95%]	  2962.79 [+1.11% -0.95%]	   614.98 [+1.00% -1.00%]
    11_completeSilently_2	136	  2932.35 [+1.05% -0.97%]	  2904.07 [+1.05% -0.96%]	   611.21 [+1.00% -1.00%]
```

lrytz:
Looks good to me, thanks! IUC, this saves around 3% of the `jvm` phase.

Would you like to get the first commit (`RunnableInPhase`) in as well? If so I'd like it to have a bit more documentation: what it's used for and how.

mkeskells:
@lrytz can we add the additional documentation with the contribution for the profiler? Hard to describe how to use it without the other part of the puzzle

there is an earlier version in https://github.com/scala/scala/pull/5760 of the generation, but that doesnt have the MT support, and has a few more bugs and less feature, but some things in that branch that relate to your CI, so I have created another PR https://github.com/scala/scala/pull/5848 so you can see the changes to the RunnableInPhase, but no documentation in there yet

to generate the result summary the tool that we use is https://github.com/rorygraves/perf_tester
this is a little rough. Its not really a standalone tool yet. Not sure where this would be useful.

I think that as a CI task, but not sure how that would work with the infra that you have - it would need to run on the dedicated hardware as a performance test. What could we do to link this to your existing performance test work - which I think only runs after merge

lrytz:
> @lrytz can we add the additional documentation with the contribution for the profiler? Hard to describe how to use it without the other part of the puzzle

Yes, that sounds good. Can we move this class out of this PR then?

mkeskells:
rebased and removed profiling commit

mkeskells:
@lrytz can this PR be merged?

lrytz (approved):
Yes, looks good!

PR #5833, opened 2017-04-05
minor BTypes optimisations

mkeskells:
performance result - disapointing - havent looked why
```
after 10 90% jvm, no GC
          12_minor_BTypes	34	  2513.46 [+1.03% -0.97%]	  2501.84 [+1.02% -0.97%]	   603.24 [+1.00% -1.00%]
```

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -123,17 +123,30 @@ abstract class BTypes {
    * has the method.
    */
   val indyLambdaImplMethods: mutable.AnyRefMap[InternalName, mutable.LinkedHashSet[asm.Handle]] = recordPerRunCache(mutable.AnyRefMap())
+
+  /**
+    * add methods
+    * @return the added methods. Note the order is undefined
+    */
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -123,17 +123,30 @@ abstract class BTypes {
    * has the method.
    */
   val indyLambdaImplMethods: mutable.AnyRefMap[InternalName, mutable.LinkedHashSet[asm.Handle]] = recordPerRunCache(mutable.AnyRefMap())
+
+  /**
+    * add methods
+    * @return the added methods. Note the order is undefined
+    */
   def addIndyLambdaImplMethod(hostClass: InternalName, handle: Seq[asm.Handle]): Seq[asm.Handle] = {
```
retronym, line 131: Thanks for pointing out that the order is unspecified. I was sort of worried that this is a (small) problem, and that bytecode generation of `lambdaDeserialize` will be non-deterministic, but it looks like that part is taken care of by the `LinkedHashSet` here, and the non-deterministic order in the return value is benign given the way it is used in the inliner.

retronym (approved):

PR #5834, opened 2017-04-05
reduce backend IO load

mkeskells:
cache created directory, 
faster IO and parsing of dir structure
dont use AbstractFile for files that you want to create

performance results - looks good
```
after 10 90% jvm, no GC
              13_reduceIO	38	  1713.87 [+1.04% -0.96%]	  1701.48 [+1.05% -0.96%]	   600.59 [+1.00% -1.00%]
```


mkeskells:
will have a look at the test failure tonight

lrytz:
The two failures are:

```
!!  293 - run/macro-range                           [non-zero exit code]

Exception in thread "main" java.lang.ClassFormatError: Extra bytes at the end of class file Test
```

```
!! 1205 - run/t5717.scala                           [non-zero exit code]

/home/jenkins/workspace/scala-2.12.x-validate-test/test/files/run/t5717-run.obj/a
java.nio.file.FileAlreadyExistsException: /home/jenkins/workspace/scala-2.12.x-validate-test/test/files/run/t5717-run.obj/a
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:88)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:384)
	at java.nio.file.Files.createDirectory(Files.java:674)
	at java.nio.file.Files.createAndCheckIsDirectory(Files.java:781)
	at java.nio.file.Files.createDirectories(Files.java:727)
	at scala.tools.nsc.backend.jvm.OutputDirectories$PathDirInfo$PathDir.dir$lzycompute(BytecodeWriters.scala:48)
	at scala.tools.nsc.backend.jvm.OutputDirectories$PathDirInfo$PathDir.dir(BytecodeWriters.scala:48)
	at scala.tools.nsc.backend.jvm.OutputDirectories$PathDirInfo$PathDir.file(BytecodeWriters.scala:49)
	at scala.tools.nsc.backend.jvm.BytecodeWriters.getFile(BytecodeWriters.scala:121)
	at scala.tools.nsc.backend.jvm.BytecodeWriters.getFile$(BytecodeWriters.scala:114)
	at scala.tools.nsc.backend.jvm.BCodeHelpers.getFile(BCodeHelpers.scala:23)
	at scala.tools.nsc.backend.jvm.GenBCode$BCodePhase.sendToDisk$1(GenBCode.scala:433)
```

mkeskells:
@lrytz thanks for the diagnostic - will have a look 

Is the expectation that the compiler will delete whatever files to create the directories needed, and similarly delete directories to create files, when the names conflict. The test case is not specific (t5717), though it should be easy to fix to whatever is the expectation (either of the above seems dangerous though)

Always wary about deleting trees :(

I can fix this when I get a clarification on the expectation above

For the other issue - I suspect that this is due to there being duplicate classes with the same name in the test - as I read macro-range both "Range" and "Test" are duplicated in the root package. Does this test compile all 3 files are they incrementally compiler in phases

Sorry for my ignorance of this but haven't really looked at these tests and how then run

lrytz:
Running tests is easy in sbt (https://github.com/scala/scala#using-the-sbt-build)

About `t5717`: I also got confused and it took me a while. The compiler in fact reports an error message. Before SI-5717 was fixed, the compiler was crashing with an exception. The test only ensures that the compiler doesn't crash. I clarified that here: https://github.com/scala/scala/pull/5835

About `macro-range`: Partest supports separate compilation based on filenames. It compiles sources with the same `_n` ending in groups, and in increasing order. Each stage uses a new compiler instance.

In the example, it first compiles `Common_1.scala`, then `Expansion_Impossible_2.scala`, and finally `Expansion_Possible_3.scala`. When compiling the second file, a warning is emitted (not visible in the `.check` file):
```
warning: some macros could not be expanded and code fell back to overridden methods;
recompiling with generated classfiles on the classpath might help.
one warning found
```
The third file contains the exact same definitions of `Range` and `Test`, but because it's compiled with the macro implementation on the classpath, macro expansion succeeds. Compiling the third file should just the corresponding, existing classfiles.

som-snytt:
Paulp's first fix for SI-5717 was destructive, but the use case of outputting to the current dir is not far-fetched. I typically have a scratch or tmp dir where I have misc test classes, and occasionally a text file that is unfortunately named and clashes with a package. Maybe there's a useful heuristic, such as specifying an output dir makes it trashable.

mkeskells:
@lrytz thanks for that detail - 

I will add a catch and some appropriate diagnostic. Probably Monday night when I am back

I have add a fix for the macro-range issue in https://github.com/scala/scala/pull/5834/commits/8dd934a8f3030d4ece65b2917865c2c654fd17ef

I think that this was caught by luck though. It was by chance that this test ( and this test only) wrote a slightly smaller file in the second pass ( not sure why) 
I think that there should be a test to ensure that the file is just the content of the bytecode written, and will add one for that.

What is the naming convention of these tests?

I was aware of how to run the tests from sbt (https://github.com/scala/scala#using-the-sbt-build), but sbt partest fails on windows, and the ++ syntax doesn't work to override the build dir (from cmd)

lrytz:
> What is the naming convention of these tests?

If there's a ticket on https://github.com/scala/bug/issues, it's tNNNN, otherwise be there's none..

lrytz:
> sbt partest fails on windows, and the ++ syntax doesn't work to override the build dir (from cmd)

Ah! We have a nightly build on windows, and it's passing (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-windows/). However, this runs in cygwin, not cmd. @szeiger might be able to help out. Our README just says "Windows may work if you use Cygwin"..

szeiger:
"Runs in Cygwin" is overstating it a bit. In Cygwin you get the sbt launcher bash script instead of the Windows cmd script but that's it. The JVM processes are not Cygwin-aware, so once you're in the JVM there shouldn't be any difference. The only exception I can think of is calling an external "diff" tool to show diffs in case of partest failures. If there's no POSIX-compatible `diff` on the path, this will fail. We should replace it with a JVM-based diffing library.

I'd go looking at the sbt launcher and its configuration as possible sources of errors. AFAIR configuration of JVM parameters is done differently in the bash and cmd launchers, so this could explain different results (when the parameters that you expect are not actually used).

som-snytt:
I was just looking at partest to make it pass -Xlint and also https://github.com/scala/scala-partest/issues/55 oh wait I guess https://github.com/scala/scala-partest/pull/81 about diff order, and I noticed that it will use git diff or else difflib as available, so possibly under --update-check it should always use difflib for reproducibility. Edit: or I guess it always just logs the log output; the diff is just for human consumption, either diff suffices to pass/fail the test. I'll look again later when I have time, but that seems obvious.

szeiger:
@som-snytt That's probably the right explanation. We already know from the problems leading to https://github.com/scala/scala/pull/5733 that @mkeskells' `git` output is not to be trusted.

mkeskells:
@som-snytt @szeiger I ran on a clean build from the start or last week after https://github.com/scala/scala/pull/5733 was merged

szeiger:
@mkeskells #5733 wouldn't fix this but it's the same symptom. Executing `git` on your machine produces output that it shouldn't produce. The correct fix here is to install git properly so that this doesn't happen. But in order to make our build more robust we shouldn't rely on the output of external commands for things that can be easily done in the sbt process.

mkeskells:
@szeiger I preumed that it fixed git usage. I suppose this was just to get some minimal set running for the mkPack?

mkeskells:
all test finally pass - can squash, but will leave so the reviewers can see the changes, and squash when I get the nod

test/files/run/t5717.scala
```diff
@@ -13,9 +13,19 @@ object Test extends StoreReporterDirectTest {
   // Let the compiler tell us output dir
   // val sc = newCompiler("-cp", classpath, "-d", testOutput.path)
   // val out = sc.settings.outputDirs.getSingleOutput.get
+  // TODO check the reported errors
```
lrytz: for this, just rebase this PR on current 2.12.x, https://github.com/scala/scala/pull/5835 is merged

lrytz (commented):
Thanks for the changes, they look good! Also thanks for the new tests.

I implemented a few cleanups here: https://github.com/rorygraves/scalac_perf/pull/2. We're trying to keep the formatting (more or less) consistent in this repo, so it'd be nice if you could use `t: T` (not `t:T` or `t :T`). Some other details:
  - No empty line before the first / after the last expression in a block (class, method)
  - Empty line between methods, classes
  - No space in `foo(arg`
  - Space in `if (..)`, `C {`, `foo = {`

You can squash the changes into one commit.

Finally, the comment https://github.com/scala/scala/pull/5832#issuecomment-293522266 about `RunnableInPhase` also applies here.

szeiger:
@mkeskells That fix was only for using git to get the commit SHA and date, not for any other uses (but `git diff` should be the only one left now).

mkeskells:
is there a intellij formatter for the code style anywhere

lrytz:
I don't think so, also we don't have our conventions written down, I just mentioned a few things that people seem to follow here. Things are not too strict either.

It would be nice to adopt scalfmt, but require a bit of work..

lrytz:
This patch looks good to me now, I think you can squash everything into one. I'd also remove the first commit (`RunnableInPhase`) and treat that in a separate PR.

mkeskells:
Squashed and removed the first commit (profiler enabler), and rebased as requested

The rebase changed so that t5717 now has validated output so I have changed the expectation to match the modified test
was 
```
error: error writing a/B: t5717-run.obj/a/B.class: t5717-run.obj/a is not a directory
```
now 
```
error: error writing a/B: t5717-run.obj/a
error: error writing a/b/B: t5717-run.obj/a/b: Not a directory
error: error writing a: t5717-run.obj/./a.class: Is a directory
```

mkeskells:
will revist this after @lrytz has completed his refactor

PR #5836, opened 2017-04-07
[WIP] Reestablish support for dependent types in TypeTags

Jasper-M:
[This commit](https://github.com/scala/scala/commit/524c90d09f2fb4687b312f2c7597393978d50b6a) outlawed free types from TypeTag, for good reasons. Alas this also means things like the following don't work:

```
val a = new { type T = Int }
typeOf[a.T]

trait Foo { type T }
val foo = new Foo { type T = Int }
typeOf[foo.T]
```
But this does work, which seems a bit inconsistent to me:
```
object a { type T = Int }
typeOf[a.T]
```

This pull request attempts to bring back support for the first type of path dependent types, while keeping the other illegal cases like local classes in the land of WeakTypeTag.

### DIFF
--- a/src/compiler/scala/reflect/reify/codegen/GenSymbols.scala
+++ b/src/compiler/scala/reflect/reify/codegen/GenSymbols.scala
@@ -140,7 +140,7 @@ trait GenSymbols {
   def reifyFreeType(binding: Tree): Tree =
     reifyIntoSymtab(binding.symbol) { sym =>
       if (reifyDebug) println("Free type: %s (%s)".format(sym, sym.accurateKindString))
-      state.reificationIsConcrete = false
+      if (!sym.hasAttachment[TypeIsConcrete.type]) state.reificationIsConcrete = false
       val name: TermName = nme.REIFY_FREE_PREFIX append sym.name
       Reification(name, binding, mirrorBuildCall(nme.newFreeType, reify(sym.name.toString), mirrorBuildCall(nme.FlagsRepr, reify(sym.flags)), reify(origin(sym))))
     }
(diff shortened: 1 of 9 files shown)


Jasper-M:
Is it normal that `typeOf[foo.T]` _does_ work? (currently with or without this pull request)

```scala
scala> trait Foo { type T }
defined trait Foo

scala> val foo = new Foo { type T }
foo: Foo = $anon$1@3b8f25c1

scala> typeOf[foo.T]
res0: reflect.runtime.universe.Type = foo.T

scala> val bar = new Foo { type T = Int }
bar: Foo{type T = Int} = $anon$1@bd6d2b2

scala> typeOf[bar.T]
<console>:16: error: No TypeTag available for bar.T
       typeOf[bar.T]
             ^
```

And by extension, should these things also work?
```scala
val a: AnyRef{ type T } = new { type T }
typeOf[a.T]

val b: AnyRef{ type F[x] } = new { type F[x] }
typeOf[b.F[Int]]

val c: AnyRef{ type T >: Some[Int] <: Option[Int] } = new { type T >: Some[Int] <: Option[Int] }
typeOf[c.T]

val d: AnyRef{ type A; type T = A } = new { type A; type T = A }
typeOf[d.T]

val e: AnyRef{ type T = a.T } = new { type T = a.T }
typeOf[e.T]
```

retronym:
@xeno-by Do you have some time to help @Jasper-M answer those questions?

adriaanm:
Happy to reschedule for 2.12.4 if we can make some progress on this. ping @xeno-by

Jasper-M:
I'm not sure if I'll be able to do much for this in time for 2.12.4 though, unless the current implementation is pretty close to an actual solution.
Nevertheless I would be interested to know which types really need to be outlawed from `TypeTag` and how well the set of types that the implementation in 2.12.3 outlaws matches with that.

adriaanm:
To be honest, I'd have to find some time to sit down and think about this, and not sure I'm going to be able to prioritize this in the next few months. Plenty of work to do for 2.13. I'm guessing Eugene is too busy with the compiler performance arms race ;-)

SethTisue:

PR #5838, opened 2017-04-10
SI-2458 Make spec example live test

som-snytt:
Synchronize the live test with the spec update,
which is trivial.

Also add a neg test showing that an imported name
remains ambiguous even if it resolves to the
definition in scope with which it is ambiguous.


lrytz (approved):

lrytz:
thanks!

PR #5839, opened 2017-04-10
Deal with undefined offset (#10255)

raboof:
In the scaladoc viewer javascript.

Unfortunately I must admit I haven't 'properly' tested this change myself as the sbt project didn't load for me, but it *seems* straightforward enough...

https://github.com/scala/bug/issues/10255

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.js
+++ b/src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.js
@@ -113,20 +113,26 @@ function handleKeyNavigation() {
         scroller.container = $container;
 
         scroller.scrollDown = function($elem) {
-            var yPos = $elem.offset().top; // offset relative to viewport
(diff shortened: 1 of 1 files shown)


src/scaladoc/scala/tools/nsc/doc/html/resource/lib/index.js
```diff
@@ -113,20 +113,26 @@ function handleKeyNavigation() {
         scroller.container = $container;
 
         scroller.scrollDown = function($elem) {
-            var yPos = $elem.offset().top; // offset relative to viewport
-            if ($container.height() < yPos || (yPos - $("#search").height()) < 0) {
-                $container.animate({
-                    scrollTop: $container.scrollTop() + yPos - $("#search").height() - 10
-                }, 200);
+            var offset = $elem.offset(); // offset relative to viewport
+            if (offset !== undefined) {
```
retronym, line 117: This idiom seems to work according to https://jsfiddle.net/s1ssgcex/

retronym (approved):

PR #5840, opened 2017-04-10
Stricter type parameters of collection.mutable.HashEntry

cypok:
`next` in `scala.collection.mutable.HashEntry` is of type `Object` in bytecode. This requires additional interface casts to `HashEntry` (e.g. in `HashTable.resize()`).

This change makes semantics of type parameters more clear and improves generated bytecode a little bit.

### DIFF
--- a/src/library/scala/collection/mutable/HashEntry.scala
+++ b/src/library/scala/collection/mutable/HashEntry.scala
@@ -12,7 +12,7 @@ package mutable
 /** Class used internally.
  *  @since 2.8
  */
-trait HashEntry [A, E] {
+trait HashEntry [A, E <: HashEntry[A, E]] {
   val key: A
   var next: E = _
 }


cypok:
Please review @Ichoran

cypok:
@axel22, please review.

axel22:
It is difficult to predict the code that the JIT compiler would emit here - my guess is that the first compilation of the `asInstanceOf` on `next` would result in type check plus a deopt (~2 extra instructions), which is probably hoisted out of loops - so, I would not expect a huge performance difference.
Still, I think this change is ok, and I don't think there is much (any) code out there that violates the assumption about the upper bound.

LGTM

szeiger (approved):
Since the trait is marked for internal use I'm not worried too much about external users. It is used by `scala-parallel-collections` but all uses there should still work with the upper bound.

PR #5841, opened 2017-04-11
SI-7960 and SI-10023 Improve REPL Output for val/var and Method Definitions

mattfowler:
Proposed fix for SI-7960 and SI-10023.  REPL now prints vals vars and defs as follows:

```
scala> val a = 1
val a: Int = 1

scala> var b = 2
var b: Int = 2

scala> def foo = 3
def foo: Int

scala> def foo() = 4
def foo(): Int

scala> def foo2(a: Int, b: Int) = a + b
def foo2(a: Int, b: Int): Int

scala> def foo3(a: Int)(b: Int) = (c: Int) => a + b + c
def foo3(a: Int)(b: Int): Int => Int
```

For `def foo = 3` it's not clear to me if this should actually output `val foo: Int` which is what is actually defined, but would be slightly confusing.

Not sure if it might also make sense to put "defined" in front of the outputs here to make it consistent with elsewhere.

If this looks good I can update the failing REPL based tests.  It looks like some error outputs are affected as well (see t7157.check for example).

### DIFF
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -2526,7 +2526,12 @@ trait Types
     //TODO this may be generalised so that the only constraint is dependencies are acyclic
     def approximate: MethodType = MethodType(params, resultApprox)
 
-    override def safeToString = paramString(this) + resultType
+    override def safeToString = {
+      resultType match {
+        case MethodType(_, _) => paramString(this) + resultType
+        case _                => paramString(this) + ": " + resultType
+      }
+    }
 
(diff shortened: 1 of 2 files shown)


adriaanm:
I have a commit for this in my repl rework for 2.13: https://github.com/scala/scala/commit/61d180b3c517fe90c9066a0e56260c9a558dc681

adriaanm:
I'm closing this PR because it has been dormant for a while now. This is fixed in 2.13.

som-snytt:
"This is fixed in 2.13" is the new "Already fixed in Ammonite."

PR #5842, opened 2017-04-11
SI-10193 - returns ListMap on mapValues method

Rohithyeravothula:


lrytz:
This is not binary compatible and cannot be done in 2.12.x. Probably this will be fixed anyway in 2.13 by the new collections design (being developed here https://github.com/scala/collection-strawman), cc @szeiger

retronym:
I'm going to close this. Feel free to re-open on the 2.13.x branch to continue discussion about the change.

PR #5844, opened 2017-04-11
Revert some of ade53a123. Use completer factory methods.

adriaanm:
Scalameta et al need to be able to customize the
type completer behavior, so we must use factory methods
to instantiate them, rather than instantiating the classes
directly.

/cc @xeno-by 

Should unblock scala/scala-dev#331


SethTisue:
I'll test this in the scalameta repo

UPDATE: testing it in the scalameta repo wasn't trivial since paradise is versioned by Scala minor version. rather than mess with it I'm using dbuild, with community.dbuild cut back to just scalameta's recursive dependencies.

UPDATE 2: aigh, dbuild failed, maybe some problem with `2.12.2-bin-2804c63-SNAPSHOT` version number format? conflicting xml versions (one with `_2.12` one with no Scala version suffix). so I tested by doing `publishLocal` from paradise first, and that worked.

lrytz (approved):

PR #5845, opened 2017-04-12
Actually retract clashing synthetic apply/unapply

adriaanm:
The completer set the IS_ERROR flag and I assumed the typer
dropped a synthetic tree with a symbol with that flag, because
the tree was not shown in -Xprint output.

A separate compilation run seems necessary to reproduce.
The original `ClashPoly` in pos/userdefined_apply.scala test also
reproduces the bug if `BaseCP` is compiled separately from `ClashPoly`.

Thankfully, the community build caught my mistake.

Fix scala/bug#10261

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -664,7 +664,17 @@ trait Namers extends MethodSynthesis {
 
           if (suppress) {
             sym setInfo ErrorType
+
+            // There are two ways in which we exclude the symbol from being added in typedStats::addSynthetics,
+            // because we don't know when the completer runs with respect to this loop in addSynthetics
+            //  for (sym <- scope)
+            //    for (tree <- context.unit.synthetics.get(sym) if shouldAdd(sym)) {
+            //      if (!sym.initialize.hasFlag(IS_ERROR))
+            //        newStats += typedStat(tree)
+            // If we're already in the loop, set the IS_ERROR flag and trigger the condition `sym.initialize.hasFlag(IS_ERROR)`
             sym setFlag IS_ERROR
+            // Or, if we are not yet in the addSynthetics loop, we can just retract our symbol from the synthetics for this unit.
+            companionContext.unit.synthetics -= sym
 
             // Don't unlink in an error situation to generate less confusing error messages.
             // Ideally, our error reporting would distinguish overloaded from recursive user-defined apply methods without signature,
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3168,7 +3168,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           for (sym <- scope)
             // OPT: shouldAdd is usually true. Call it here, rather than in the outer loop
             for (tree <- context.unit.synthetics.get(sym) if shouldAdd(sym)) {
-              newStats += typedStat(tree) // might add even more synthetics to the scope
+              // if the completer set the IS_ERROR flag, retract the stat (currently only used by applyUnapplyMethodCompleter)
+              if (!sym.initialize.hasFlag(IS_ERROR))
+                newStats += typedStat(tree) // might add even more synthetics to the scope
               context.unit.synthetics -= sym
             }
           // the type completer of a synthetic might add more synthetics. example: if the
--- a/test/files/run/t10261/Companion_1.scala
+++ b/test/files/run/t10261/Companion_1.scala
@@ -1,0 +1,4 @@
+trait Companion[T] {
+  def parse(value: String): Option[T]
+  def apply(value: String): T = parse(value).get
+}
--- a/test/files/run/t10261/Test_2.scala
+++ b/test/files/run/t10261/Test_2.scala
@@ -1,0 +1,14 @@
+import scala.util.Try
+
+object C extends Companion[C] {
+  def parse(v: String) = if (v.nonEmpty) Some(new C(v)) else None
+}
+
+case class C(value: String)
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    assert(Try{C("")}.isFailure, "Empty value should fail to parse") // check that parse is used to validate input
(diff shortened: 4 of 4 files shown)


adriaanm:
Once the 2.12 community build is green, I'll backport to 2.11.x and with blushing cheeks cut 2.11.11....

adriaanm:
Lesson learned: always develop a new feature on the latest version first, and make sure the community build there is ok (since it's grown twice as a big as on 2.11!!)

adriaanm:
I still don't understand why removing the symbol from its scope isn't enough to keep the backend from emitting it. Also, why does the separately compiled scenario trigger the bug but not the one where everything is compiled at once? I'd like to understand this before tagging 2.12.2.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3168,7 +3168,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           for (sym <- scope)
             // OPT: shouldAdd is usually true. Call it here, rather than in the outer loop
             for (tree <- context.unit.synthetics.get(sym) if shouldAdd(sym)) {
-              newStats += typedStat(tree) // might add even more synthetics to the scope
+              if (!sym.initialize.hasFlag(IS_ERROR)) // if the completer set the IS_ERROR flag, retract the stat
```
lrytz: could use `!sym.initialize.isError`

lrytz:
I see the failure also when compiling the two files at once, for both the test case in the commit as well as the one in https://github.com/scala/bug/issues/10261

```
➜  sandbox git:(2.12.x) cat Test.scala
trait Companion { def apply(value: String): Any = ??? }
object o extends Companion
case class o(value: String)
➜  sandbox git:(2.12.x) qsc -version
Scala compiler version 2.12.2-bin-18157b9 -- Copyright 2002-2016, LAMP/EPFL and Lightbend, Inc.
➜  sandbox git:(2.12.x) qsc Test.scala
warning: an unexpected type representation reached the compiler backend while compiling Test.scala: <error>. If possible, please file a bug on issues.scala-lang.org.
scala.MatchError: <error> (of class scala.reflect.internal.Types$ErrorType$)
```

lrytz:
But, the example from `pos/userdefined_apply` that you mention only fails in separate compilation for me

```
class BaseCP[T] {
  def apply(x: T): ClashPoly = if (???) ClashPoly(1) else ???
}
```

```
object ClashPoly extends BaseCP[Int]
case class ClashPoly private(x: Int)
```

```
➜  sandbox git:(2.12.x) qsc A.scala B.scala
➜  sandbox git:(2.12.x) qsc A.scala && qsc B.scala
warning: an unexpected type representation reached the compiler backend while compiling B.scala: <error>. If possible, please file a bug on issues.scala-lang.org.
scala.MatchError: <error> (of class scala.reflect.internal.Types$ErrorType$)
```

lrytz (approved):
The reason why supressed synthetic apply/unapply symbols were not added (i.e., the good case) is because they were removed from their scope (https://github.com/scala/scala/blob/12c240d69b6958d2c8f03a7728c097dd215011cd/src/compiler/scala/tools/nsc/typechecker/Namers.scala#L678)

In the typer we have

```
def addSynthetics .. {
...
          for (sym <- scope)
            for (tree <- context.unit.synthetics.get(sym) if shouldAdd(sym)) {
              newStats += typedStat(tree)
...
}
```

If the `sym` is not yet initialized, it is still in the scope, therefore the tree is added (even if the completer then marks it `IS_ERROR`). So it depends on completion order, which is influenced by many factors.

lrytz:
Ah, we should add

```
            companionContext.unit.synthetics -= sym
```

here: https://github.com/scala/scala/blob/12c240d69b6958d2c8f03a7728c097dd215011cd/src/compiler/scala/tools/nsc/typechecker/Namers.scala#L666-L667

adriaanm:
> If the sym is not yet initialized, it is still in the scope, therefore the tree is added (even if the completer then marks it IS_ERROR). So it depends on completion order, which is influenced by many factors.

Thank you. That explains it.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
adriaanm: I like being explicit about the flags :)

adriaanm:
> Ah, we should add
> 
>             companionContext.unit.synthetics -= sym

@lrytz done, changes amended

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -664,7 +664,17 @@ trait Namers extends MethodSynthesis {
...
+
+            // There are two ways in which we exclude the symbol from being added in typedStats::addSynthetics,
+            // because we don't know when the completer runs with respect to this loop in addSynthetics
+            //  for (sym <- scope)
+            //    for (tree <- context.unit.synthetics.get(sym) if shouldAdd(sym)) {
+            //      if (!sym.initialize.hasFlag(IS_ERROR))
+            //        newStats += typedStat(tree)
+            // If we're already in the loop, set the IS_ERROR flag and trigger the condition `sym.initialize.hasFlag(IS_ERROR)`
             sym setFlag IS_ERROR
+            // Or, if we are not yet in the addSynthetics loop, we can just retract our symbol from the synthetics for this unit.
```
lrytz, line 676: this is not strictly necessary as we also unlink the symbol from its scope below, so the loop in `addSynthetics` wouldn't find it. but it keeps the `synthetics` map cleaner.

adriaanm: Right! Good point. Maybe I should take another vacation or something.

adriaanm:
Follow up for 
https://github.com/scala/scala/pull/5816

src/compiler/scala/tools/nsc/typechecker/Namers.scala
som-snytt: Now I definitely need a vacation. https://github.com/scala/scala/pull/8597

adriaanm: ^^ Nice cadeautje under my 🎄.

PR #5846, opened 2017-04-12
Actually retract clashing synthetic apply/unapply

adriaanm:
Also make this whole retraction of apply/unapply in case of a
clashing user-defined member conditional on `-Xsource:2.12`.

It turns out, as explained by lrytz, that the retraction mechanism
was fragile because it relied on the order in which completers are run.
We now cover both the case that:

  - the completer was run, the `IS_ERROR` flag was set, and the
  symbol was unlinked from its scope before `addSynthetics`
  in `typedStat` iterates over the scope (since the symbol is
  already unlinked, the tree is not added, irrespective of its flags).
  For this case, we also remove the symbol from the synthetics in
  its unit (for cleanliness).

  - the completer is triggered during the iteration in `addSynthetics`,
  which needs the check for the `IS_ERROR` flag during the iteration.

Before, the completer just unlinked the symbol and set the IS_ERROR flag,
and I assumed the typer dropped a synthetic tree with a symbol with
that flag, because the tree was not shown in -Xprint output.
In reality, the completer just always happened to run before the
addSynthetics loop and unlinked the symbol from its scope in the
test cases I came up with (including the 2.11 community build).

Thankfully, the 2.12 community build caught my mistake, and lrytz provided
a good analysis and review.

Fix scala/bug#10261


sjrd (approved):
> Also make this whole retraction of apply/unapply in case of a
clashing user-defined member conditional on `-Xsource:2.12`.

👍

adriaanm:
(Forgot the flags files of course :-))

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -691,6 +691,16 @@ trait Namers extends MethodSynthesis {
...
+            // because we don't know when the completer runs with respect to this loop in addSynthetics
+            //  for (sym <- scope)
+            //    for (tree <- context.unit.synthetics.get(sym) if shouldAdd(sym)) {
+            //      if (!sym.initialize.hasFlag(IS_ERROR))
+            //        newStats += typedStat(tree)
+            //  (1) If we're already in the loop, set the IS_ERROR flag and trigger the condition
+            //      `sym.initialize.hasFlag(IS_ERROR)` in typedStats::addSynthetics,
+            //  (2) Or, if we are not yet in the addSynthetics loop (and we're not going to emit an error anyway),
+            //      we unlink the symbol from its scope.
             sym setFlag IS_ERROR
```
lrytz, line 704: - [X] add `companionContext.unit.synthetics -= sym`?

lrytz (commented):
- [X] also add `[backport]` to the commit msg

densh (approved):
I think it's much safer now under a flag. 👍

lrytz (approved):

PR #5847, opened 2017-04-12
SI-10260 better error messages for raw Java types

Jasper-M:
Attempt at fixing scala/bug#10260. It's not very pretty, but since compilation will fail anyway I assume a lot is permitted. You can now basically copy&paste the type that the error message suggests.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -704,7 +704,7 @@ abstract class RefChecks extends Transform {
                           // Compilation has already failed so we shouldn't have to worry overmuch
                           // about forcing types.
                           if (underlying.isJavaDefined && pa.typeArgs.isEmpty && abstractSym.typeParams.nonEmpty)
-                            ". To implement a raw type, use %s[_]".format(pa)
+                            s". To implement this raw type, use ${rawToExistential(pa)}"
                           else if (pa.prefix =:= pc.prefix)
                             ": their type parameters differ"
                           else
--- a/test/files/neg/t8244.check
+++ b/test/files/neg/t8244.check
@@ -1,4 +1,4 @@
-Test_2.scala:9: error: value exxx is not a member of ?0
+Test_2.scala:9: error: value exxx is not a member of T
     raw.t.exxx // java.lang.ClassCastException: java.lang.String cannot be cast to X
           ^
 one error found
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -3876,8 +3876,8 @@ trait Types
   }
 
   def typeParamsToExistentials(clazz: Symbol, tparams: List[Symbol]): List[Symbol] = {
-    val eparams = mapWithIndex(tparams)((tparam, i) =>
-      clazz.newExistential(newTypeName("?"+i), clazz.pos) setInfo tparam.info.bounds)
+    val eparams = tparams map (tparam =>
+      clazz.newExistential(tparam.name.toTypeName, clazz.pos) setInfo tparam.info.bounds)
 
     eparams map (_ substInfo (tparams, eparams))
   }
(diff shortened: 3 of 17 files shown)


src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -703,8 +720,13 @@ abstract class RefChecks extends Transform {
                           // TODO: what is the optimal way to test for a raw type at this point?
                           // Compilation has already failed so we shouldn't have to worry overmuch
                           // about forcing types.
-                          if (underlying.isJavaDefined && pa.typeArgs.isEmpty && abstractSym.typeParams.nonEmpty)
-                            ". To implement a raw type, use %s[_]".format(pa)
+                          if (underlying.isJavaDefined && pa.typeArgs.isEmpty && abstractSym.typeParams.nonEmpty) {
+                            prepareRawTypeForMessage(abstractSym).fold(
+                              ". To implement a raw type, use %s[_]".format(pa)
```
Jasper-M: I think this case should actually never happen. Am I right? Should that be an error instead of a default then? If yes, which?

Jasper-M:
Review by @adriaanm ?

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
adriaanm: Yep, I think you can implement this message as `s". To implement this raw type, use ${rawToExistential(pa)}"` with a slight variation on a helper method it uses (patch below).

```diff
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -3877,7 +3877,7 @@ trait Types
 
   def typeParamsToExistentials(clazz: Symbol, tparams: List[Symbol]): List[Symbol] = {
     val eparams = mapWithIndex(tparams)((tparam, i) =>
-      clazz.newExistential(newTypeName("?"+i), clazz.pos) setInfo tparam.info.bounds)
+      clazz.newExistential(tparam.name.toTypeName, clazz.pos) setInfo tparam.info.bounds)
 
     eparams map (_ substInfo (tparams, eparams))
   }
```

Would you be interested in pursuing this change and see if we can make this work for the whole compiler, or alternatively make a second `rawToExistentialPreservingNames`?

adriaanm (changes requested):
Thanks for looking into this! Sorry about the delay. 

I agree the current message is a bit confusing. I think we can implement this more directly, as suggested below. Also, instead of "better error message", could you sharpen the title? ("Display raw type as existential" or something like that)

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
Jasper-M: Yes I will look into this in a week or so, when I have access to a real computer. By 'make this work for the whole compiler' do you mean resolving all subsequent test failures?

adriaanm: Yes, it's worth a shot, I think, but it could also be too hairy. If you could experiment with this a bit and report back, we can decide whether to duplicate the rawToExistential transform or not.

Jasper-M:
@adriaanm As far as I can see only the following tests fail:
```
neg/t8244
neg/t8244e 
run/t5418b.scala
run/existentials-in-compiler.scala
```
To my untrained eyes they all look like `--update-check` would do.

adriaanm:
Cool! Could you submit the updated check files so we can have a look (more easily -- sorry, a bit swamped)? @retronym, what do you think?

Jasper-M:
No prob.

test/files/neg/t8244.check
```diff
@@ -1,4 +1,4 @@
-Test_2.scala:9: error: value exxx is not a member of ?0
+Test_2.scala:9: error: value exxx is not a member of T
```
adriaanm, line 1: would be interesting to investigate this one -- should indicate it's an existential skolem

adriaanm (approved):
Approving because I think it's a step in the right direction. I think it could still be refined (will the user get confused because the existentially quantified names are now the same as the universally quantified, i.e., the class's own type params?). The whole cooking Java raw types thing is a big hack, though,... We really shouldn't use modifyInfo to destroy the original info in cookJavaRawInfo, but I don't see any other way either (except for considering this conversion every time).

adriaanm:
Thanks for your patience.

adriaanm:
PS: de groeten daar in Heverlee ;-)

Jasper-M:
Thanks ;) maar ik ben een week geleden vertrokken in Heverlee :D

PR #5848, opened 2017-04-12
add per-phase profiling to scalac

mkeskells:


src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -338,6 +338,16 @@ trait ScalaSettings extends AbsScalaSettings
 
   def YstatisticsEnabled = Ystatistics.value.nonEmpty
 
+  val YprofileEnabled = BooleanSetting("-Yprofile-enabled", "Enable profiling.")
```
ShaneDelmore, line 342: This whole pr is super useful. Thank you!

src/compiler/scala/tools/nsc/profile/ExternalToolHook.java
```diff
@@ -0,0 +1,17 @@
+package scala.tools.nsc.profile;
+
+/**
```
jvican, line 3: Can we simplify and make these docs clearer? There are typos, no capital letter and the actual meat of the docs is hidden under unnecessary words :smile:.

rorygraves:
Better description (from updated git commit):

    Add support for phase profiling generating inforamtion per phase:
      - Wall Clock Time (ms) - wall clock time
      - Idle Time (ms) - time blocked
      - cpu time (ms) - reported cpu time
      - user time (ms) - reported user time
      - allocated (mb) - bytes allocated during phase
      - retainedHeap (mb) - retained heap after phase (enable GC for more accurate counts)
      - gcTime (ms) - time spent in GC.

    Adds four flags:

    -Yprofile-enabled
      Add per phase profiling information

    -Yprofile-destination:<file>
      Where to write profiling information (defaults to console if not set)

    -Yprofile-run-gc
      Run GC between phases to help gemerate more more accurate memory usage in profiling

    -Yprofile-external-tool:<phase>
      Enable calling of ExternalToolHook.before/after around the target phase

retronym (approved):
I have some ideas about ways to extend this, but I think its a great starting point and we can include in 2.12.3.

Future work: 
  - Rather than rendering the profile results to the logger after each compiler run, aggregate them over many runs of the compiler and present summary statistics at the end.
  - Try to reuse some of the MXBean based profiler code from JMH rather than DIY (unfortunately this isn't straightforward with the current API of JMH)

rorygraves:
Great. 

I agree those are both good ideas.

PR #5849, opened 2017-04-13
Use partest 1.1.1

lrytz:
During bootstrapping 2.13.x, partest is re-built. Partest has Xlint
fatal warnings enabled.

scala/scala#5402 enabled new unused warnings, which fails the 1.1.0
build.

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -19,6 +19,6 @@ scala.binary.version=2.12
 #  - jline: shaded with JarJar and included in scala-compiler
 #  - partest: used for running the tests
 scala-xml.version.number=1.0.6
-partest.version.number=1.1.0
+partest.version.number=1.1.1
 scala-asm.version=5.1.0-scala-2
 jline.version=2.14.3


SethTisue:
Why didn't we hit this in 2.12?

lrytz:
i didn't want to interfere with the 2.12.2 release process

lrytz:
Should I change it to 2.12.x? For me it doesn't matter, I'll have to merge 2.12.x -> 2.13.x anyway.

SethTisue:
Holding off on the partest upgrade on the 2.12.x branch until after 2.12.2 makes sense.

But, what I meant to ask was: why didn't the same issue cause 2.12.x to fail bootstrap as well?

lrytz:
Because we're not rebuilding partest during the 2.12 boostrap, we're just using the binary compatible published version. 2.13 version numbers are all binary incompatible until 2.13.0 is out, so every bootstrap will rebuild the modules.

For PR validation, we keep using the last release, hoping that it won't break due to a binary incompatible change. So the 2.13 branch uses the `_2.12` partest for now. Once M1 is out, we'll switch to `_2.13.0-M1` until M2, and so forth.

PR #5852, opened 2017-04-13
SI-8935 valueOfTerm for repl-class-based

som-snytt:
Descend reflectively through members in the path.

Also, let `SessionTest` use check files. Embedding the session transcript is not convenient for updating.

There are a few use cases for the embedded session.

One wrongly updated check file: test/files/run/t8918-unary-ids.check 

A nice property of the session test is that the input is the output; you don't have to guess if it's correct, or correlate line comments like `// compiles` or `// warns` in the output.

Fixes scala/bug#8935


adriaanm (approved):
Thanks!

PR #5857, opened 2017-04-19
Fix lambda deserialization in classes with 252+ lambdas

retronym:
Create a lambda deserializer per group of target methods,
and call these sequentially trapping the particular pattern
of exception that is thrown when a target method is absent
from the map.

Fixes scala/bug#10232

```
  // access flags 0x100A
  private static synthetic $deserializeLambda$(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object;
    TRYCATCHBLOCK L0 L1 L1 java/lang/IllegalArgumentException
   L0
    ALOAD 0
    INVOKEDYNAMIC lambdaDeserialize(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object; [
      // handle kind 0x6 : INVOKESTATIC
      scala/runtime/LambdaDeserialize.bootstrap(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/invoke/MethodHandle;)Ljava/lang/invoke/CallSite;
      // arguments:
      // handle kind 0x6 : INVOKESTATIC
      Test$.$anonfun$main$1$adapted(Lscala/Function1;)Ljava/lang/Object;,
      // handle kind 0x6 : INVOKESTATIC
      Test$.$anonfun$lambdas$1(Ljava/lang/Object;)Ljava/lang/String;,

      ...

      Test$.$anonfun$lambdas$249(Ljava/lang/Object;)Ljava/lang/String;,
      // handle kind 0x6 : INVOKESTATIC
      Test$.$anonfun$lambdas$250(Ljava/lang/Object;)Ljava/lang/String;
    ]
    ARETURN
   L1
    ALOAD 0
    INVOKEDYNAMIC lambdaDeserialize(Ljava/lang/invoke/SerializedLambda;)Ljava/lang/Object; [
      // handle kind 0x6 : INVOKESTATIC
      scala/runtime/LambdaDeserialize.bootstrap(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/invoke/MethodHandle;)Ljava/lang/invoke/CallSite;
      // arguments:
      // handle kind 0x6 : INVOKESTATIC
      Test$.$anonfun$lambdas$251(Ljava/lang/Object;)Ljava/lang/String;,
      // handle kind 0x6 : INVOKESTATIC
      Test$.$anonfun$lambdas$252(Ljava/lang/Object;)Ljava/lang/String;,

      ...

      // handle kind 0x6 : INVOKESTATIC
      Test$.$anonfun$lambdas$256(Ljava/lang/Object;)Ljava/lang/String;
    ]
    ARETURN
    MAXSTACK = 2
    MAXLOCALS = 1

```


retronym:
Unfortunately, this change uses exceptions as control flow for lambda#252 and above, but that seems to be the best way around the compatibility constraint that the newly generated code for `$lambdaDeserialize$` should work when run with older standard libraries.

I've prototyped a change to avoid the overhead of filling in the stack trace of the `IllegalArgumentException` used for control flow (https://github.com/scala/scala/compare/2.12.x...retronym:ticket/10232-stackless), but I'm reluctant to merge that it also relies on hacks to avoid requiring API changes to the runtime support.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -87,15 +100,40 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
...
+      mv.visitLabel(labels(0))
+      emitLambdaDeserializeIndy(groups(i))
+      mv.visitLabel(labels(1))
       mv.visitInsn(ARETURN)
-      mv.visitEnd()
+      mv.visitLabel(labels(2))
+      mv.visitVarInsn(ASTORE, 1)
+      mv.visitVarInsn(ALOAD, 1)
+      mv.visitMethodInsn(INVOKEVIRTUAL, "java/lang/IllegalArgumentException", "getCause", "()Ljava/lang/Throwable;", false)
+      mv.visitJumpInsn(IFNULL, nextLabel(i))
```
lrytz: should there be an `ALOAD 1; ATHROW` here?

lrytz:
Looks good otherwise. I guess for 2.13 we can have a solution that doesn't use exceptional control flow? You could mark the commit `[nomerge]`.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
retronym: Actually, I've just noticed that the second `throw IllegalArgumentException` in `LambdaDeserializer` is detritus that should have been removed in https://github.com/scala/scala/commit/131402fd5fe8c064ef5cfffbe568507cbdf37990.

I'll remove that `throw` and then simplify this PR (no need to check `getCause == null` here, just swallow any `IAE` unconditionally in all but the terminal `indy` call.

retronym:
Pushed the discussed simplification.

>  I guess for 2.13 we can have a solution that doesn't use exceptional control flow? You could mark the commit `[nomerge]`.

I'd rather merge this into 2.13.x and then adapt the result, I've opened a ticket to track that: https://github.com/scala/scala-dev/issues/373

lrytz:
Sure, that's fine too.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -87,15 +99,34 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
...
+    val groups: Array[Array[Handle]] = implMethodsArray.grouped(targetMethodGroupLimit).toArray
+    val numGroups = groups.length
+
+    import scala.tools.asm.Label
+    val initialLabels = Array.fill(numGroups - 1)(new Label())
+    val terminalLabel = new Label
+    def nextLabel(i: Int) = if (i == numGroups - 2) terminalLabel else initialLabels(i + 1)
+
+    for ((label, i) <- initialLabels.iterator.zipWithIndex) {
+      mv.visitTryCatchBlock(label, nextLabel(i), nextLabel(i), "java/lang/IllegalArgumentException")
```
lrytz: Can you add `jlIllegalArgumentExceptionRef` to `CoreBTypes`, like [`jlClassCastExceptionRef`](https://github.com/scala/scala/blob/v2.12.2/src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala#L105), then use `.internalName`? This ensures that the `ClassBType` is part of the `classBTypeFromInternalName` map, which is excpected for computing stack map frames during class writing.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -66,15 +66,27 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
...
    * private static Object $deserializeLambda$(SerializedLambda l) {
-   *   return indy[scala.runtime.LambdaDeserialize.bootstrap](l)
+   *   <FOR N in 0..NUM_GROUPS-2>
+   *   try {
+   *    return indy[scala.runtime.LambdaDeserialize.bootstrap, targetMethodGroup$N](l)
+   *   } catch {
+   *     case i: IllegalArgumentException => // swallow
+   *   }
+   *   </FOR>
+   *   return indy[scala.runtime.LambdaDeserialize.bootstrap, targetMethodGroup${NUM_GROUPS-1}](l)
```
lrytz: the generated structure is more like

```
try return indy[scala.runtime.LambdaDeserialize.bootstrap, targetMethodGroup$0](l)
catch {
  case i: IllegalArgumentException =>
    try return indy[scala.runtime.LambdaDeserialize.bootstrap, targetMethodGroup$1](l)
    catch {
      case i: IllegalArgumentException =>
        ...
          return indy[scala.runtime.LambdaDeserialize.bootstrap, targetMethodGroup${NUM_GROUPS-1}](l)
    }
}
```

retronym: Is there a difference? In any case, your version is clearer, so I'll use that.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
retronym: Done.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
lrytz: Just a detail to make the comment align better with the implementation.

src/compiler/scala/tools/nsc/backend/jvm/CoreBTypes.scala
```diff
@@ -103,6 +103,7 @@ class CoreBTypes[BTFS <: BTypesFromSymbols[_ <: Global]](val bTypes: BTFS) {
   lazy val jlCloneableRef            : ClassBType = classBTypeFromSymbol(JavaCloneableClass)        // java/lang/Cloneable
   lazy val jiSerializableRef         : ClassBType = classBTypeFromSymbol(JavaSerializableClass)     // java/io/Serializable
   lazy val jlClassCastExceptionRef   : ClassBType = classBTypeFromSymbol(ClassCastExceptionClass)   // java/lang/ClassCastException
+  lazy val jlIllegalArgExceptionRef  : ClassBType = classBTypeFromSymbol(IllegalArgExceptionClass)  // java/lang/ClassCastException
```
lrytz: Comment is out of synch..

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -87,15 +99,34 @@ class BackendUtils[BT <: BTypes](val btypes: BT) {
...
+    val groups: Array[Array[Handle]] = implMethodsArray.grouped(targetMethodGroupLimit).toArray
+    val numGroups = groups.length
+
+    import scala.tools.asm.Label
+    val initialLabels = Array.fill(numGroups - 1)(new Label())
+    val terminalLabel = new Label
+    def nextLabel(i: Int) = if (i == numGroups - 2) terminalLabel else initialLabels(i + 1)
+
+    for ((label, i) <- initialLabels.iterator.zipWithIndex) {
+      mv.visitTryCatchBlock(label, nextLabel(i), nextLabel(i), coreBTypes.jlIllegalArgExceptionRef.internalName)
```
lrytz: `coreBTypes._` is already imported

lrytz (approved):

PR #5858, opened 2017-04-19
Bump version numbers after 2.13.0-M1

lrytz:


lrytz:
Note that the scalacheck dependency doesn't use the `scala.binary.version` property in `versions.properties`, it's [hard-coded to the `_2.12`](https://github.com/scala/scala/blob/v2.13.0-M1/build.sbt#L45). So we don't need to wait for scalacheck.

lrytz:
This PR basically comes too late, because there was a nightly run on the tagged revision, which would overwrite the downloadable distribution packages (https://downloads.lightbend.com/scala/2.13.0-M1/scala-2.13.0-M1.tgz). We had this problem during the 2.12.0 release, the file hashes kept changing.

Fortunately, this time the bootstrap build failed (https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-bootstrap/336/console, https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-bootstrap/336/artifact/logs/builds/*view*/)

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: com.typesafe.sbt#sbt-pgp;0.8.1: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

But IMO we should really disable that automation, we don't seem to use it anyway.

PR #5859, opened 2017-04-19
Update ticket references and bugtracker URL [ci: last-only]

lrytz:


lrytz:
Anyone motivated to review this (ideally before there are merge conflicts..)?

SethTisue:
a few stray references remain:

```
src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala:                 |if this is reproducible, please report bug at https://issues.scala-lang.org/
src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala:                 |if this is reproducible, please report bug at http://issues.scala-lang.org/
src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala:            "If possible, please file a bug on issues.scala-lang.org.")
```

LGTM otherwise

lrytz:
Thanks, Seth. Pushed a fixup for these three, not sure why I missed those.

Note that not all of the individual commits would pass CI validation, I think some compiler messages are changed in different commits than `.check` files.

adriaanm:
When the time comes, shall we squash on merge?

adriaanm (approved):
👍

lrytz:
we can also squash, as you prefer. i tried to split things a bit up by topic, but maybe it's not all that useful in the history.

adriaanm:
Yes, that helped for reviewing, but I don't see us reverting these individual commits.

PR #5863, opened 2017-04-25
Fix MiMa feature request URL

dwijnand:
I accidentally mislead @szeiger in https://github.com/scala/scala/pull/5190#r66948268.

Please retarget to whatever branch this type of change should target.


szeiger (approved):
AFAIK we're not planning to do any more 2.11 releases so 2.12.x should be OK.

dwijnand:
Btw this is relevant because @2m's been fixing this issue, so in a near future it's possible the scala build could use sbt-mima as a plugin like every other user of MiMa.

https://github.com/typesafehub/migration-manager/pull/173

PR #5864, opened 2017-04-26
Optimize symbol substitution

retronym:
Rather than asking each type for its set of boundSyms, and then
intersecting that with `from`, just dive into `renameBoundSyms`
and compare the directly bound poly/method/existential params with
`from`.

Optimize this comparison by pre-computing the symbol id range of
`from` (typically small), to cheaply rule out some types without
needing to traverse `from`.

Deprecate the now-unused `Type#boundSyms`


retronym:
Split out of #5823

adriaanm (approved):
Thanks for adding the comments. Nice to see dealing with symbol-binding types become more generic (as in dotty). Would be cool if we could write:

```
protected def renameBoundSyms(tp: Type): Type = tp match {
  case tp: SymbolBindingType  => copyWithClonedSymbols(tp)
  case _ => tp
}
```

PR #5865, opened 2017-04-26
Revert "Don't cache `MethodSymbol`'s `memberType`."

retronym:
This reverts commit 1f6f7f8aa94c622665a35343de8108ea66a787b7.

... but in a limited way: we only cache member types in the
typer compiler phase.


retronym:
Split out of #5823

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2931,6 +2941,37 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     // are a case accessor (you can also be a field.)
     override def isCaseAccessorMethod = isCaseAccessor
 
+    def typeAsMemberOf(pre: Type): Type = {
```
jvican, line 2944: Great comment. Thanks.

adriaanm (approved):

PR #5867, opened 2017-04-27
Include the parts of a compound/refinement type in implicit scope.

hrhino:
According to the spec, the parts of a type `T1 with ... with Tn { ... }` are the union of the parts of `T1`, ..., `Tn`. Without this, given an implicit view from `(A) => B` defined in `A`'s companion:

    val t: B = new A {}

and

    val t: B = { object a extends A {}; a }

had different behavior (only the latter compiled), which seemed surprising.

I somewhat feel like I'm about to run headlong into some dark pit which Chesterton so kindly erected a fence around for people like me, but I can't figure out what this check was meant to accomplish (it was added with relatively little comment in commit b7bdf048b1 back in 2011).

Fixes scala/bug#10288; review by @adriaanm


milessabin:
This looks like a good candidate for the next Typelevel Scala release.

retronym:
Thanks for digging your way into this pit!

To be able to include this patch in the a 2.12 minor release, we'd need to predicate the new behaviour on `-Xsource:2.13`. 

```scala
if (settings.isScala213) newBehaviour else oldBehaviour
```

I think https://github.com/scala/scala/commit/aaf919859f5dfa295aac8846bc109ceb87984a7e#diff-7c03397456d3b987549fd039d6b639c6R516 was the first to exclude refinement/anon classes from contributing to companion implicits. @odersky Can you remember the motivation?

hrhino:
Good to know, @retronym! Flag check added and amended.

And yes, I'm interested in knowing what the original reason for this check was, as well.

@milessabin, would you like that I made a PR also against TLS?

edit: whoops, that wasn't the button I meant to press

milessabin:
@hrhino no need I can merge this PR directly.

retronym:
I'd still like to chase down the opinion from @odersky about this one before we merge. Please bear with my. We're cutting 2.12.3 very shortly, so I'm going to change the milestone here to 2.12.4.

adriaanm:
> I think aaf9198#diff-7c03397456d3b987549fd039d6b639c6R516 was the first to exclude refinement/anon classes from contributing to companion implicits

Without accompanying motivation/spec changes, we'll have to assume it was an oversight.

adriaanm:
@retronym could you run a benchmark on this PR to ensure the extra flag check doesn't affect performance. Once that's done, I'm happy to merge.

adriaanm (approved):
Let's run the benchmarks on the merge commit. If there's an issue, we'll back out the change on Monday.

PR #5870, opened 2017-04-28
Move tuple[23]ToZippedOps to Tuple[23] companion objects. Fixes scala/bug#10293

kamilkloch:


kamilkloch:
Signed the CLA.

adriaanm:
Welcome! Thank you for your contribution, but we cannot make a change like this in the 2.12.x series because we promise [binary compatibility](http://docs.scala-lang.org/overviews/core/binary-compatibility-of-scala-releases.html).

We can discuss this in the context of 2.13.0-M2. That said, what's the rationale? (The answer to this should also be included in your PR description and commit message.)

kamilkloch:
Hello @adriaanm, thanks for the response. I tried to explain the rationale in scala/bug#10293.

1. Only ATuple2Ops and ATuple3Ops are defined in Predef.scala. If we are to be consistent, why not all the way to Tuple22? 

2. `tuple2ToZippedOps` defined in `Predef.scala` has higher precedence than any implicit conversion defined in the companion object of a user-defined class. See scala/bug#10293 for an illustration. Implicit conversion `tuple2ToZippedOps` defined Predef.scala is IMHO way too strong. It deals with tuples, and as such belongs to the Tuple's companion object.

3. Tuple2Zipped.Ops adds 2 methods (invert & zipped), both of which require the generic types to be traversable. Yet this constraint is missing in the definition of tuple2ToZippedOps:
```
implicit def tuple2ToZippedOps[T1, T2](x: (T1, T2))   
```

Fixes scala/bug#10293.

adriaanm:
It's a reasonable concern, and we should find a way to improve this in 2.13, where we're reworking the collections. Our binary compatibility guarantees here are set in stone, though -- the best we can do in 2.12 is some form of deprecation.

szeiger (changes requested):
https://github.com/scala/bug/issues/10293 looks like a problem worth solving but there are several problems with the current changes:

- As Adriaan pointed out, the motivation (e.g. ticket numbers) should be in the commit message, not just a github comment.

- Notice how `Tuple2.scala` and `Tuple3.scala` start with a comment saying "GENERATED CODE: DO NOT EDIT". Any changes to these files need to be incorporated into `genprod.scala` (and then regenerated because we check in both the source and the generated files).

- An explicit companion object for a case class does not automatically extend the correct Function type (like an implicitly generated one would), so this needs to be added.

- Would moving the methods into `LowPriorityImplicits` work instead? This would be a much simpler change.

- The current handling of the implicit evidence values was probably motivated by performance. `Tuple2Zipped` and `Tuple3Zipped` don't take implicit parameters because they could not be value classes anymore if they did. Instead the individual methods require these parameters.  Adding "dummy" implicits to the implicit conversions probably has significant performance implications. There should be a strong motivation for introducing such a performance regression (or benchmarks that show that performance is unaffected).

kamilkloch:
* Understood, I shall include the ticket number in commit messsages. 

* I was aware of the "GENERATED CODE: DO NOT EDIT", just did not know how to re-generate `TupleX` files and wanted to present a proof-of-concept. 

*  As for the explicit companion object of a case class - thank you, I was not aware if that. 

* `LowPriorityImplicits` trick won't work, unfortunately. I tried it. If I am not mistaken, the trick works for the  resolution of the implicits in the companion objects, as opposed to the explicit import of the implicit conversions via  `import Predef._`

* Performance-wise, `zipped` captures implicit conversions, so in the end there should be little difference. 

* `Tuple2Zipped` adds two methods - `invert` and `zipped` - which have little in common. 
  - `zipped` captures implicit conversions to Traversable types (for the sole purpose that it works for Strings and Arrays), 
  - `invert` (btw, why not `transpose`?) does nto capture implicits and works only for Scala traversables.

* Please take a look at https://github.com/scala/scala/pull/5870/commits/fa822108e6bd270c952d4a5d57ca8b3bfca1cd1d
with `TupleXZipped.Ops` split into `TupleXZipped.InvertOps` and `TupleXZipped.ZippedOps`.

szeiger:
After thinking some more about this and digging in the commit history I'm not sure what's the best way forward (or if this should be changed at all). The origin of the current implementation is https://github.com/scala/scala/commit/cf18d879d3f43d5a0c15c5c0af72f88c25605a08. According to the commit comment the motivation for making `zipped` and `invert` extension methods in the first place was to move the implementations out of `genprod.scala` and into "proper" source files.

1. Should https://github.com/scala/bug/issues/10293 be considered a bug at all? If `zipped` and `invert` were instance methods (as originally intended and implemented), you wouldn't be able to replace them with your own extension methods and this would be an easy "won't fix".

2. If we accept the current implicits-based state, where should the implicits for the extension methods be defined? On one hand it is preferable to scope implicits as narrowly as possible for performance reasons. On the other hand this hurts discoverability. At least `package scala` and `scala.Predef` are two well-known locations where all conversions for the standard library are added. Is there any precedent for adding always available extension methods outside these standard locations? A cursory search didn't turn up anything.

3. Should we make it possible to override standard extension methods of core types like `Tuple2` in such a casual way (by providing a different extension method with higher priority)? Maybe a better way forward is to suggest selective imports from Predef so you have to be explicit about removing the standard conversions.

kamilkloch:
Stefan, thanks for the response. I will try to response to your 3 points. 

1. The proposition is to merely change the scope of the implicit conversions (from the global `Predef.scala` into the companion objects), and not to revert to instance methods. 

2. Hmm, discoverability is handled by IDE, I do not really buy this argument ;) Companion object seems to be a perfect and sensible place for implicit conversions/extensions. Global scope is bad. See SO questions how to disable `any2stringadd`. Also , there is a reason why TLS 4 added configurable default imports: https://github.com/typelevel/scala/blob/typelevel-readme/notes/typelevel-4.md#configurable-default-imports-pull5350-andyscott

3. Sure, selective imports from `Predef.scala` would work (see TLS 4), but isn't it a real pain to use? For instance, as in scala/bug#10293, the code does not work as expected, so you have to search through the default imports, locate the source of the problem, and then do not forget it to disable it in _every_ file where you want to use your extension method. All because a wildcard import of implicits trumps companion object implicits.

szeiger:
> The proposition is to merely change the scope of the implicit conversions

Sure, but the reason for making them extension methods rather than proper instance methods in the first place was to get them out of genprod. Putting them into the companion objects means putting them back into genprod. The only reason for still making them extension methods in this case is to allow others to replace them with higher-priority extension methods. 

> Hmm, discoverability is handled by IDE, I do not really buy this argument ;) Companion object seems to be a perfect and sensible place for implicit conversions/extensions. Global scope is bad. See SO questions how to disable `any2stringadd`

If relying on IDEs was a sufficiently good solution people would be able to find `any2stringadd`. I doubt it would be easier if `any2stringadd` was defined in the companion object of `Any`. But it wouldn't be possible to disable it anymore. Having it in Predef makes this easy.

> so you have to search through the default imports, locate the source of the problem, and then do not forget it to disable it in *every* file where you want to use your extension method

The alternative of silently replacing a method that is available by default in every Scala source file seems worse to me. From my experience with designing "clever" APIs I would advise to use a different name for your method. If you really want to call it `zipped` a better way forward is to push for https://github.com/scala/scala/pull/5350 so you get the configurable default imports like in TLS.

szeiger:
Any other opinions on this PR? @SethTisue? My preference would be to close it because the proposed changes IMHO do not come up ahead when weighing the pros and cons against the status quo, and to recommend https://github.com/scala/scala/pull/5350 as the solution for https://github.com/scala/bug/issues/10293.

kamilkloch:
I see my arguments against global 'zipped' implicits, strangely available only for 2/22 tuples (Tuple2 and Tuple3) did not come through. I still believe that cherry-picking default imports, as in #5350, is inferior to letting the implicit precedence do the work automatically.

lrytz:
After looking at this for a while, it seems to me that the concerns @szeiger mentions outweigh the benefits here. If you want the same functionality for larger tuples, you can do it. If you want a different functionality, it's a good idea to give your extension method a different name.

I totally agree we should deprecate `invert` and introduce `transpose`.

szeiger:
I hadn't paid much attention to those comments, but I agree. `transpose` would be better. Assigning to M3 for now.

kamilkloch:
OK, I am closing this PR and will submit a new one with regard to deprecating `invert` with `transpose`.

PR #5872, opened 2017-04-28
Enable -howtorun:repl -i -e

som-snytt:
The runner concatenates -i files and -e expression
to run the code as a script. The addition of -I
broke detection of -i there, which is now fixed.

In addition, allow explicit `-howtorun:repl` which
is now useful for `-e`, so that REPL will handle
`-i` and `-I` and then evalute the expression as
usual (and as the `-help` text implies).

Fixes: scala/bug#10284

Where `dbl.sc` defines overloaded `f` (or redefines it if not pasted):
```
$ skala -howtorun:repl -nc -I dbl.sc -e "println(f(42))"
 found   : Int(42)
 required: String
       println(f(42))
                 ^

$ skala -howtorun:repl -nc -i dbl.sc -e "println(f(42))"
int 42

$ skala -howtorun:repl -nc -i dbl.sc -e "f(42)"
res0: String = int 42

$ skala -howtorun:script -nc -i dbl.sc -e "f(42)"


```
I restored REPL output in REPL mode by default, as a use case.


src/repl/scala/tools/nsc/interpreter/InteractiveReader.scala
```diff
@@ -46,6 +46,13 @@ object InteractiveReader {
     }
 
   def apply(): InteractiveReader = SimpleReader()
+
+  // a non-interactive SimpleReader that returns the given text
+  def apply(text: String): InteractiveReader = SimpleReader(
```
adriaanm: Could we move this to `SimpleReader`, as method `fromString` or `constant` or somesuch?

som-snytt: Maybe. `apply` gives you a `SimpleReader` for some reason. `apply(text)` gives you the same thing, but it only reads from that input. As a user, I don't care that it's a `SimpleReader`. WDYT? If I were closer to the city, I'd buy you a beer to discuss.

adriaanm: The owing of beers is all mine! I'm probably overthinking it. By the way, I've just started work on reorganizing the repl codebase to simplify maintenance and stealing of features from ammonite. My goal for the month is some more separation of concerns. Will open a ticket to discuss once I get a sense of the concrete direction I'm headed.

som-snytt: Yes, I expected a clean break from spark support at some point. Also, Ammonite is asking for community funding. What a great idea. I mean dollars, not just beer support.

som-snytt:
I acceded to Adriaan's wishes for future `Adriaanite` REPL.

It took me forever to write a stupid test, similar to an early test I contributed when fixing `scala -i` for https://github.com/scala/bug/issues/7805.

`ReplTest` was dropping two lines of output for the welcome banner, except there is no welcome banner for `-e`. Part of the reason it took me a while to see it, is that both settings and the output capture for testing in `ILoop` are complicated. Turning up debug doesn't display unmangled test transcript.

I need that beer now.

src/repl/scala/tools/nsc/interpreter/SimpleReader.scala
```diff
@@ -43,6 +43,13 @@ object SimpleReader {
 
   def apply(in: BufferedReader = defaultIn, out: JPrintWriter = defaultOut, interactive: Boolean = true): SimpleReader =
     new SimpleReader(in, out, interactive)
+
+  // a non-interactive SimpleReader that returns the given text
+  def apply(text: String): SimpleReader = apply(
```
adriaanm, line 48: Cheers!

adriaanm (approved):

src/partest-extras/scala/tools/partest/ReplTest.scala
```diff
@@ -30,21 +30,20 @@ abstract class ReplTest extends DirectTest {
...
+          case welcome(s) => s
+          case s          => s
+        }
+      } else {
+        lines.drop(header.lines.size)
       }
-    } else {
-      lines drop header.lines.size
-    }) map normalize
+    clean.map(normalize)
```
som-snytt, line 46: Previously seen at https://github.com/scala/scala/pull/5220/commits/6d8c565c8cba6405b0b9c6466776d1922b2488ba

PR #5874, opened 2017-04-28
README: note about potential SBT plugin dependency issues

ceedubs:
I recently got tripped up for a while by my global Ensime SBT plugin causing
cyclical dependency issues when running `sbt` in the scala repo. An
internet search has shown me that I'm not the first person to encounter
this issue. I'm hoping to save others some trouble. Let me know if
there's a better way to address this or if there are any wording changes
that you'd prefer.


adriaanm:
Thanks!

PR #5876, opened 2017-04-29
Conditionally lint macros and expansions

som-snytt:
Adds a flag `-Ywarn-macros`.

Don't lint macros unless they request it.

If warning, then traverse the original tree of
macro expansions to witness usages.

Fixes scala/bug#10296

### DIFF
--- a/src/compiler/scala/tools/nsc/settings/Warnings.scala
+++ b/src/compiler/scala/tools/nsc/settings/Warnings.scala
@@ -16,7 +16,19 @@ trait Warnings {
   val fatalWarnings = BooleanSetting("-Xfatal-warnings", "Fail the compilation if there are any warnings.")
 
   // Non-lint warnings.
-
+  val warnMacros           = ChoiceSetting(
+    name    = "-Ywarn-macros",
+    helpArg = "mode",
+    descr   = "Enable lint warnings on macro expansions.",
+    choices = List("none", "before", "after", "both"),
+    default = "before",
+    choicesHelp = List(
+      "Do not inspect expansions or their original trees when generating unused symbol warnings.",
+      "Only inspect unexpanded user-written code for unused symbols.",
+      "Only inspect expanded trees when generating unused symbol warnings.",
+      "Inspect both user-written code and expanded trees when generating unused symbol warnings."
+    )
+  )
   val warnDeadCode         = BooleanSetting("-Ywarn-dead-code", "Warn when dead code is identified.")
   val warnValueDiscard     = BooleanSetting("-Ywarn-value-discard", "Warn when non-Unit expression results are unused.")
   val warnNumericWiden     = BooleanSetting("-Ywarn-numeric-widen", "Warn when numerics are widened.")
--- a/src/compiler/scala/tools/nsc/settings/Warnings.scala
+++ b/src/compiler/scala/tools/nsc/settings/Warnings.scala
@@ -15,7 +15,7 @@ trait Warnings {
   // Warning semantics.
   val fatalWarnings = BooleanSetting("-Xfatal-warnings", "Fail the compilation if there are any warnings.")
 
-  // Non-lint warnings.
+  // Non-lint warnings. -- TODO turn into MultiChoiceEnumeration
   val warnMacros           = ChoiceSetting(
     name    = "-Ywarn-macros",
     helpArg = "mode",
--- a/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
@@ -473,7 +473,7 @@ trait TypeDiagnostics {
         "readResolve", "readObject", "writeObject", "writeReplace"
       ).map(TermName(_))
 
-      class UnusedPrivates(traverseCheck: Tree => Tree, isOriginal: Boolean) extends Traverser {
+      class UnusedPrivates extends Traverser {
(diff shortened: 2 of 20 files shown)


src/compiler/scala/tools/nsc/settings/Warnings.scala
```diff
@@ -17,6 +17,7 @@ trait Warnings {
 
   // Non-lint warnings.
 
+  val warnMacros           = BooleanSetting("-Ywarn-macros", "Enable warnings on macro expansions.")
```
xeno-by: Would it be possible to include both unexpanded and expanded code in lint checks?

som-snytt: I didn't give a lot of thought to this. For this ticket, I think linting looks at both. The other use case is the ticket whether to emit warnings on unexpanded code. So maybe the space of options is none, both, unexpanded and expanded.

dragos:
Is there a downside in checking both original and expanded trees? Both usages should be counted always since the definition can't be removed otherwise. And if that's the case, there isn't a need for an additional flag either.

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -492,14 +492,15 @@ trait TypeDiagnostics {
...
+          && (settings.warnMacros || !sym.isMacro)
         )
         def qualifiesType(sym: Symbol) = !sym.isDefinedInPackage
         def qualifies(sym: Symbol) = (
              (sym ne null)
           && (sym.isTerm && qualifiesTerm(sym) || sym.isType && qualifiesType(sym))
         )
 
-        override def traverse(t: Tree): Unit = {
+        override def traverse(t: Tree): Unit = if (settings.warnMacros || !hasMacroExpansionAttachment(t)) {
```
som-snytt: Does this mean: lint if `t` has nothing to do with a macro, or if they want linted macros?

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -544,6 +545,8 @@ trait TypeDiagnostics {
               case _                 =>
             }
           }
+          val expandee = macroExpandee(t)
+          if (expandee ne t) traverse(expandee)
           super.traverse(t)
```
som-snytt: Does this mean: if this is a macro and t is not the original expandee, then lint it; always lint the expanded. The flag is checked in traverse before actually descending into either tree. So it's either both trees or neither. Is that it?

lrytz:
@som-snytt what's the status here? There are some questions, also by yourself. Iulian suggests that the linter should check both trees (which it does as far when the new flag is enabled), and that we could just always do that, and not introduce the `-Ywarn-macros` flag. That sounds good to me, too. WDYT?

som-snytt:
I'll update this when I have a chance. My answer about multi choice, to lint both (by default), either or none of expandee and expanded, seems right to me. I'll try to motivate with a use case, but macro magic means a given macro might either consume or produce something that is spuriously warnable. Inevitably, someone will complain they can't use a macro lib because of linting; a flag is an escape hatch at least.

fommil:
I never realised how much I wanted this until I started turning linters / wartremover to the max :-)

som-snytt:
@adriaanm That was so easy, I thought I'd push my luck. After I take care of a small fire at the office, I'll re-review. I started looking yesterday with my new macro vision. But my home computer just died, etc.

som-snytt:
Pushed what I commented in July, default is to lint original trees, optionally lint again with expanded trees, or both, or neither.

The previous fix traversed both at once, which is less discriminating.

adriaanm:
Turned my review into a commit. Looking GTY? EDIT: couldn't resist a small `push -f`

som-snytt:
One minor comment, otherwise thanks for another good idea.

I also didn't convert strings to the `MultiChoiceEnumeration` thing. Because lazy and pressed for time.

Also, for the mixed case, the message doesn't distinguish whether the unusage is before or after expansion.

adriaanm:
The reviewer reviewed, as they say in French (something about "the watered sprinkler"). Fixed the `traverse`, but turning the rest into TODOs.

som-snytt:
@adriaanm thanks, it's like The Flash showed up to help me finish vacuuming.

adriaanm:
Zoom zoom zoom! Or were you thinking of the black adder one?

stewSquared:
Can someone clarify what `both` means?
Suppose we have `-Ywarn-unused` and `-Ywarn-macros:both`
Does it error when a val is unused either before OR after macro expansion?
Or does it error only when a val is unused before AND after?

ie., does `both` generally result in more or fewer warnings?

Edit: My local testing seems to imply the interpretation of "either before or after" is correct.
If so, how then do we emit warnings only when unused before AND after? `none`?

som-snytt:
@stewSquared It runs the check both before and then after expansion, so you're liable to see more warnings.  Both before and after trees must look clean. The missing mode is "either", where a symbol could be used before but not after and conversely for other symbols. That would result in fewer warnings. I.e., some usages could be introduced via expansion, others eliminated.

stewSquared:
I see. Thanks! I suppose `either` could be the most conservative option, but it look like it would be a bit trickier to implement.

This leaves `after` as the conservative option for most people. (`none` is then the least conservative, since spurious warnings can result from failure to find usages in either the expanded or unexpanded trees.)

PR #5879, opened 2017-05-01
-Xlint:unused -Ywarn-unused is intuitive

som-snytt:
The `-Xlint:unused` setting is just a boolean, but it
drives an expanding option on `-Ywarn-unused`.
Normally, expanding options can't be unset, so add
API for that.

This makes `-Xlint:_,-unused` work as advertised.
Other combinations such as `-Xlint -Ywarn-unused:-imports`
are also supported. That's because explicit unset trumps
an expanding option.


som-snytt:
I requested Lukas as a reviewer because he has been through the flames of optimizer options. Or optimiser options.

Possibly, an option should keep its setting history, and then compute a value on demand. Then it could say, "You set and unset me, you seem to be confused. Let me offer some help."

lrytz (approved):
Thanks!

PR #5882, opened 2017-05-03
Optimize CommandLineParser to avoid exponential cost

retronym:
Echoing the performance bug in the bash scripts (scala/bug#9279),
the command line parser exhibited exponential performance.

When using the command line arguments used to compile the standard
library, which contains ~ 700 source files, I noticed about 1% of
compilation time was being spend in evaluating the regex `(\S+)(.*)`.
Inspection of the parser reveals that this was repeatedly being
called on the rest of the command line as the parser consumed
each argument.

This commit switches to using a regex that only matches the
current argument, and then makes sure to split the input
string using structural sharing of the backing char[] (by using
`String.substring`, rather than high level Scala collections
operations from `StringLike#span`).

After this change, I no longer observed this area of code in the
top profiler results.

### DIFF
--- a/src/compiler/scala/tools/cmd/CommandLineParser.scala
+++ b/src/compiler/scala/tools/cmd/CommandLineParser.scala
@@ -39,7 +39,7 @@ object CommandLineParser {
     private val regex = """(\S+)""".r
     def unapply(s: String): Option[(String, String)] = {
       regex.findPrefixOf(s) match {
-        case Some(prefix) => Some(prefix, s.substring(prefix.length, s.length))
+        case Some(prefix) => Some(prefix, s.substring(prefix.length))
         case None => None
       }
     }
@@ -61,7 +61,7 @@ object CommandLineParser {
     else argument(trimmed) match {
       case Right((arg, next)) =>
         val leadingWhitespaceLen = next.prefixLength(Character.isWhitespace)
-        val rest = next.substring(leadingWhitespaceLen, next.length)
+        val rest = next.substring(leadingWhitespaceLen)
         if (leadingWhitespaceLen == 0 && rest.nonEmpty)
           Left("Arguments should be separated by whitespace.") // TODO: can this happen?
         else
--- a/src/compiler/scala/tools/cmd/CommandLineParser.scala
+++ b/src/compiler/scala/tools/cmd/CommandLineParser.scala
@@ -35,7 +35,15 @@ object CommandLineParser {
   }
   private object DoubleQuoted extends QuotedExtractor('"')
   private object SingleQuoted extends QuotedExtractor('\'')
-  private val Word = """(\S+)(.*)""".r
+  object Word {
+    private val regex = """(\S+)""".r
+    def unapply(s: String): Option[(String, String)] = {
+      regex.findPrefixOf(s) match {
+        case Some(prefix) => Some(prefix, s.substring(prefix.length, s.length))
+        case None => None
+      }
+    }
+  }
 
   // parse `in` for an argument, return it and the remainder of the input (or an error message)
   // (argument may be in single/double quotes, taking escaping into account, quotes are stripped)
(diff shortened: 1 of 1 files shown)


retronym:
```
[error] Test scala.tools.nsc.backend.jvm.DirectCompileTest.testSeparateCompilation failed: scala.tools.cmd.CommandLineParser$ParseException: Illegal argument:  -d /tmp/1493802034123-0 -cp /tmp/1493802034123-0, took 0.003 sec
[error]     at scala.tools.cmd.CommandLineParser$.$anonfun$tokenize$1(CommandLineParser.scala:93)
[error]     at scala.tools.cmd.CommandLineParser$.tokenize(CommandLineParser.scala:97)
[error]     at scala.tools.testing.BytecodeTesting$.newCompilerWithoutVirtualOutdir(BytecodeTesting.scala:138)
[error]     at scala.tools.testing.BytecodeTesting$.compileClassesSeparately(BytecodeTesting.scala:173)
[error]     at scala.tools.nsc.backend.jvm.DirectCompileTest.testSeparateCompilation(DirectCompileTest.scala:79)
[error]     ...
```

retronym:
I'll re-open when I address the test failure.

src/compiler/scala/tools/cmd/CommandLineParser.scala
```diff
@@ -52,10 +60,13 @@ object CommandLineParser {
     if (trimmed.isEmpty) Right((accum.reverse, ""))
     else argument(trimmed) match {
       case Right((arg, next)) =>
-        (next span Character.isWhitespace) match {
-          case("", rest) if rest.nonEmpty => Left("Arguments should be separated by whitespace.") // TODO: can this happen?
-          case(ws, rest)                  => commandLine(rest, arg :: accum)
-        }
+        val prefixLen = next.prefixLength(Character.isWhitespace)
```
adriaanm: For extra tidiness, could avoid the`leadingWhitespace` string with:

```
val leadingWhitespaceLen = next.prefixLength(Character.isWhitespace)
val rest = next.substring(leadingWhitespaceLen, next.length)

if (leadingWhitespaceLen == 0 && rest.nonEmpty)
```

src/compiler/scala/tools/cmd/CommandLineParser.scala
```diff
@@ -35,7 +35,15 @@ object CommandLineParser {
   }
   private object DoubleQuoted extends QuotedExtractor('"')
   private object SingleQuoted extends QuotedExtractor('\'')
-  private val Word = """(\S+)(.*)""".r
+  object Word {
+    private val regex = """(\S+)""".r
+    def unapply(s: String): Option[(String, String)] = {
+      regex.findPrefixOf(s) match {
+        case Some(prefix) => Some(prefix, s.substring(prefix.length, s.length()))
```
adriaanm: Inconsistent use of `length()` and `length`

adriaanm (changes requested):
Looks good, but I do have a few nits to pick.

src/compiler/scala/tools/cmd/CommandLineParser.scala
som-snytt: laugh + confused emoji might be ambiguous. @ jenkins, this comment should not require human intervention, sorry, I mean intervention by a team developer. Jenkins, old chap, you are the humanist i.e. most human, not "human bigot". Can robots be bigoted or only roboted?

adriaanm: I've been puzzling over this comment. Could you help a human out? Am I being too nitpicky?

Not sure which smiley I'll get for this one, but I just noticed there's an overload of `substring` that takes just a `beginIndex`.

som-snytt: I just meant a linter should save you time on this sort of style comment.

I hope you're catching some of the Mr Rogers marathon. He's so good.

retronym:
I've denitted the patch.

adriaanm (approved):

src/compiler/scala/tools/cmd/CommandLineParser.scala
```diff
@@ -52,10 +60,12 @@ object CommandLineParser {
...
     else argument(trimmed) match {
       case Right((arg, next)) =>
-        (next span Character.isWhitespace) match {
-          case("", rest) if rest.nonEmpty => Left("Arguments should be separated by whitespace.") // TODO: can this happen?
-          case(ws, rest)                  => commandLine(rest, arg :: accum)
-        }
+        val leadingWhitespaceLen = next.prefixLength(Character.isWhitespace)
+        val rest = next.substring(leadingWhitespaceLen)
+        if (leadingWhitespaceLen == 0 && rest.nonEmpty)
+          Left("Arguments should be separated by whitespace.") // TODO: can this happen?
```
som-snytt, line 66: Not quite an answer, but as per the comment on the quote extractor:
```
scala> scala.tools.cmd.CommandLineParser.tokenize(""""abc"def""")
scala.tools.cmd.CommandLineParser$ParseException: Arguments should be separated by whitespace.
```

som-snytt: Fixed in the latest denitting.

PR #5883, opened 2017-05-03
Precompute Setting.isScala211 etc

retronym:
This avoids the small, but non-neglible cost of evaluating
the regex that implements `source.value >= version212`, etc.

This is a less ambitious alternative to #5825 which
addresses the worst culprit of inefficient settings lookups.

### DIFF
--- a/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
@@ -83,11 +83,19 @@ trait ScalaSettings extends AbsScalaSettings
    * The previous "-source" option is intended to be used mainly
    * though this helper.
    */
-  def isScala211: Boolean = source.value >= ScalaVersion("2.11.0")
-  private[this] val version212 = ScalaVersion("2.12.0")
-  def isScala212: Boolean = source.value >= version212
-  private[this] val version213 = ScalaVersion("2.13.0")
-  def isScala213: Boolean = source.value >= version213
+  private[this] var _isScala211: Boolean = _
+  private[this] var _isScala212: Boolean = _
+  private[this] var _isScala213: Boolean = _
+  private[this] val defaultScalaVersion = ScalaVersion("2.12.0")
+  def isScala211: Boolean = _isScala211
+  def isScala212: Boolean = _isScala212
+  def isScala213: Boolean = _isScala213
+  private def initIsScala(value: ScalaVersion): Unit = {
+    _isScala211 = value >= ScalaVersion("2.11.0")
+    _isScala212 = value >= ScalaVersion("2.12.0")
+    _isScala213 = value >= ScalaVersion("2.13.0")
+  }
+  initIsScala(defaultScalaVersion)
 
   /**
    * -X "Advanced" settings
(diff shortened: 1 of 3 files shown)


rorygraves:
Isn't isScala211 and isScala212 defunct now (always true in a 2.12 build)?

retronym:
You can still use `-Xsource:2.10` to disable language features (that have been designed with that setting in mind) that might be hampering migration.

rorygraves:
Oh, I see. I had not realised you could do that.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -131,7 +132,11 @@ trait ScalaSettings extends AbsScalaSettings
   val sourceReader       = StringSetting       ("-Xsource-reader", "classname", "Specify a custom method for reading source files.", "")
   val reporter           = StringSetting       ("-Xreporter", "classname", "Specify a custom reporter for compiler messages.", "scala.tools.nsc.reporters.ConsoleReporter")
   val strictInference    = BooleanSetting      ("-Xstrict-inference", "Don't infer known-unsound types")
-  val source             = ScalaVersionSetting ("-Xsource", "version", "Treat compiler input as Scala source for the specified version, see scala/bug#8126.", initial = ScalaVersion("2.12"))
+  val source             = ScalaVersionSetting ("-Xsource", "version", "Treat compiler input as Scala source for the specified version, see scala/bug#8126.", initial = ScalaVersion("2.12")).withPostSetHook( setting => {
+    _isScala211 = setting.value >= ScalaVersion("2.11.0")
+    _isScala212 = setting.value >= ScalaVersion("2.12.0")
+    _isScala213 = setting.value >= ScalaVersion("2.13.0")
+  })
```
lrytz: Hmm, given the build failures, I wonder if these variables are being set when the `-Xsource` flag is not present.

retronym:
D'oh! Pushed a fix, but it is starting to get a bit ugly.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -83,11 +83,19 @@ trait ScalaSettings extends AbsScalaSettings
...
    */
-  def isScala211: Boolean = source.value >= ScalaVersion("2.11.0")
-  private[this] val version212 = ScalaVersion("2.12.0")
-  def isScala212: Boolean = source.value >= version212
-  private[this] val version213 = ScalaVersion("2.13.0")
-  def isScala213: Boolean = source.value >= version213
+  private[this] var _isScala211: Boolean = _
+  private[this] var _isScala212: Boolean = _
+  private[this] var _isScala213: Boolean = _
+  private[this] var defaultScalaVersion = ScalaVersion("2.12.0")
```
dwijnand: doesn't look like it needs to be a var

som-snytt:
Sorry to be obtuse, but I thought it just uses `compare` to compare the two values. What regex?

retronym:
Oh, looks like the problem is just that:

```
def isScala211: Boolean = source.value >= ScalaVersion("2.11.0")
```

doesn't hoist out `ScalaVersion.apply("2.11.0")` (which regex-parses the string) into a val.

lrytz:
Hah, i was just about to add my review, then github told me "You can only approve open PRs" :)

lrytz:
I also overlooked @som-snytt's comment, sorry!

som-snytt:
One of you did complain that I was too lazy to valify the `isScala211`. That was probably Adriaan. I remember I was about to check if that was used much, but I must have forgotten, or it got stranded on a branch somewhere.

som-snytt:
No, it was @retronym https://github.com/scala/scala/pull/5539#discussion_r93143644 

Mea culpa. Sorry I made a joke. "I didn't want to push it." Get it? "git push"? Ha, ha. It was the holiday season, I guess.

I made a PR https://github.com/scala/scala/pull/5890 from my old cygwin box, which was a useful exercise. Obviously, because hulu.

PR #5884, opened 2017-05-03
Optimize erasure typer by avoiding tree hashcode

retronym:
Profilers showed that during the namer/typer handoff via the
transformed hashmap, calls to Tree.hashCode (delegating to the native
System.identityHashCode) were a hotspot.

This commit avoids routes around this in the erasure phase type
checker.


adriaanm (approved):
Would be nice if we could get rid of that `transformed` map one day.

PR #5889, opened 2017-05-05
Optimize Iterator.combinations

retronym:
Avoid a creating of temporary array just to calculate the sum
of a slice.

REPL session to show the equivalence of the new approach:
```
scala> Array(1,2,3,4).slice(1, 2).sum
res0: Int = 2

scala> Array(1,2,3,4).view(1, 2).sum
res1: Int = 2

scala> Array(1,2,3,4).slice(1, 2)
res2: Array[Int] = Array(2)

scala> Array(1,2,3,4).view(1, 2)
res3: scala.collection.mutable.IndexedSeqView[Int,Array[Int]] = SeqViewS(...)
```

### DIFF
--- a/src/library/scala/collection/SeqLike.scala
+++ b/src/library/scala/collection/SeqLike.scala
@@ -232,7 +232,13 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
       if (idx < 0)
         _hasNext = false
       else {
-        var sum = nums.slice(idx + 1, nums.length).sum + 1
+        // OPT: hand rolled version of `sum = nums.view(idx + 1, nums.length).sum + 1`
+        var sum = 1
+        var i = idx + 1
+        while (i < nums.length) {
+          sum += nums(i)
+          i += 1
+        }
         nums(idx) -= 1
         for (k <- (idx+1) until nums.length) {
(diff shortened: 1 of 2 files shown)


src/library/scala/collection/SeqLike.scala
```diff
@@ -232,7 +232,7 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
       if (idx < 0)
         _hasNext = false
       else {
-        var sum = nums.slice(idx + 1, nums.length).sum + 1
+        var sum = nums.view(idx + 1, nums.length).sum + 1
```
Ichoran: This is still considerably slower than a direct computation.  If it matters at all, it should be switched to an accumulator.

retronym: I've pushed a new version with a plain-old loop. I've also added a unit test in a preceding commit to show the behaviour hasn't changed.

src/library/scala/collection/SeqLike.scala
```diff
@@ -9,7 +9,9 @@
 package scala
 package collection
 
-import immutable.{ List, Range }
+import java.util.stream.Collectors
```
xuwei-k: unused import?

szeiger (approved):

PR #5890, opened 2017-05-05
Compute also isScala211 only once

som-snytt:
Retronym also did this follow-up, which also
cleans up former cachings.

### DIFF
--- a/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
@@ -83,7 +83,8 @@ trait ScalaSettings extends AbsScalaSettings
    * The previous "-source" option is intended to be used mainly
    * though this helper.
    */
-  def isScala211: Boolean = source.value >= ScalaVersion("2.11.0")
+  private[this] val version211 = ScalaVersion("2.11.0")
+  def isScala211: Boolean = source.value >= version211
   private[this] val version212 = ScalaVersion("2.12.0")
   def isScala212: Boolean = source.value >= version212
   private[this] val version213 = ScalaVersion("2.13.0")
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -83,7 +83,8 @@ trait ScalaSettings extends AbsScalaSettings
    * The previous "-source" option is intended to be used mainly
    * though this helper.
    */
-  def isScala211: Boolean = source.value >= ScalaVersion("2.11.0")
+  private[this] val version211 = ScalaVersion("2.11.0")
+  def isScala211: Boolean = source.value >= version211
```
janekdb, line 87: Off topic: `isGteqScala211` is a more accurate name!

som-snytt: On topic! Maybe `wasScala211`.

dwijnand: I've defined these before in build files as `isScala211Plus` (i.e `2.11+`).

janekdb: `atLeastScala211` is accurate and mildly more abstract than `gteqScala211`.

retronym (approved):
Thanks for following this up, LGTM.

som-snytt:
How do they score this in the bug derby?

retronym:
A solid 7.0, accounting for the degree of difficulty in dodging the question of `{is,was,isAtLeast,remindsOneOf}Scala211`

som-snytt:
I meant in sports like basketball and hockey where they have "assists" or maybe baseball where a relief pitcher can get a "save". We need a scorecard. Or a "box score". FOSS can get complicated, and that's even modulo the licenses.

PR #5891, opened 2017-05-06
Fix queue ++ (bug #10298)

edmundnoble:
Sorry bout that. This looks to have fixed the issue; thanks for the help @Ichoran.

### DIFF
--- a/src/library/scala/collection/immutable/Queue.scala
+++ b/src/library/scala/collection/immutable/Queue.scala
@@ -108,8 +108,14 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
 
   override def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Queue[A], B, That]): That = {
     if (bf eq Queue.ReusableCBF) {
-      val thatQueue = that.asInstanceOf[Queue[B]]
-      new Queue[B](thatQueue.in ++ (thatQueue.out reverse_::: this.in), this.out).asInstanceOf[That]
+      val newIn =
+        if (that.isInstanceOf[Queue[_]]) {
+          val thatQueue: Queue[B] = that.asInstanceOf[Queue[B]]
+          thatQueue.in ++ (thatQueue.out reverse_::: this.in)
+        } else {
+          (new ListBuffer[B] ++= that.seq).prependToList(this.in)
+        }
+      new Queue[B](newIn, this.out).asInstanceOf[That]
     } else {
       super.++(that)(bf)
     }
--- a/test/files/run/t10298.scala
+++ b/test/files/run/t10298.scala
@@ -1,0 +1,17 @@
+import collection.immutable._
+
+object Test {
+
+  def main(args: Array[String]): Unit = {
+    assert((Queue(1) ++ Vector(2)) == Queue(1, 2))
+
+    assert(((Queue(1).++(Vector(2))(collection.breakOut)): Vector[Int]) == Vector(1, 2))
+
+    assert(((Queue(1) :+ 2) ++ Vector(3)) == Queue(1, 2, 3))
+
+    assert(((1 +: Queue(2)) ++ Vector(3)) == Queue(1, 2, 3))
+
+    assert(((1 +: Queue(2)) ++ (3 +: Queue(4))) == Queue(1, 2, 3, 4))
+  }
+
+}


src/library/scala/collection/immutable/Queue.scala
```diff
@@ -108,8 +108,12 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
 
   override def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Queue[A], B, That]): That = {
     if (bf eq Queue.ReusableCBF) {
-      val thatQueue = that.asInstanceOf[Queue[B]]
-      new Queue[B](thatQueue.in ++ (thatQueue.out reverse_::: this.in), this.out).asInstanceOf[That]
+      if (that.isInstanceOf[Queue[_]]) {
```
Ichoran: It would be clearer if you did something like `val newIn = if (that.isInstanceOf[Queue[_]]) ...` so you didn't have to repeat the `new Queue[B](_, this.out).asInstanceOf[That]` part.

edmundnoble: Done.

Ichoran (approved):
LGTM.

src/library/scala/collection/immutable/Queue.scala
```diff
@@ -108,8 +108,14 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
 
   override def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Queue[A], B, That]): That = {
     if (bf eq Queue.ReusableCBF) {
-      val thatQueue = that.asInstanceOf[Queue[B]]
-      new Queue[B](thatQueue.in ++ (thatQueue.out reverse_::: this.in), this.out).asInstanceOf[That]
+      val newIn =
+        if (that.isInstanceOf[Queue[_]]) {
+          val thatQueue: Queue[B] = that.asInstanceOf[Queue[B]]
```
som-snytt, line 113: Isn't this the same as an unchecked match `case q: Queue[B] =>`?

Ichoran: Yeah but I already asked him to rewrite it once.  I guess it might be more idiomatic if it were a match.  (Personally I don't really like unchecked matches; if one is going to cheat I prefer to `asInstanceOf` it.)

edmundnoble: This is why I avoided writing as a match, yes. If it had more cases I'd have reconsidered.

ghost:
Thank you for the fix @edmundnoble!

test/files/run/t10298.scala
```diff
@@ -0,0 +1,17 @@
...
+  def main(args: Array[String]): Unit = {
+    assert((Queue(1) ++ Vector(2)) == Queue(1, 2))
+
+    assert(((Queue(1).++(Vector(2))(collection.breakOut)): Vector[Int]) == Vector(1, 2))
+
+    assert(((Queue(1) :+ 2) ++ Vector(3)) == Queue(1, 2, 3))
+
+    assert(((1 +: Queue(2)) ++ Vector(3)) == Queue(1, 2, 3))
+
+    assert(((1 +: Queue(2)) ++ (3 +: Queue(4))) == Queue(1, 2, 3, 4))
```
mingthemerciless, line 14: Shouldn't there be test cases for appending an empty vector, and for appending a vector that has more than 1 element?

edmundnoble: There's no test case unless it's calling a method on `Queue`, because that's what I changed.

src/library/scala/collection/immutable/Queue.scala
```diff
@@ -108,8 +108,14 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
...
+          thatQueue.in ++ (thatQueue.out reverse_::: this.in)
+        } else {
+          (new ListBuffer[B] ++= that.seq).prependToList(this.in)
```
mingthemerciless, line 116: Sorry, but this looks wrong. Do we really want the behaviour that `Queue(1,2) ++ Vector(3,4) == Queue(1,2,4,3)`, even though `Queue(1,2) ++ Queue(3,4) == Queue(1,2,3,4)` ? 

I'm pretty sure we should prepend the reverse of `that.seq` to `this.in` - just like we do when we `enqueue` an `Iterable`.

test/files/run/t10298.scala
mingthemerciless: 'Appending' in my comment above refers to the `++` method on `Queue`. 

I suggest at least the following additional test cases:
`Queue(1) ++ Vector(2,3)`
`(Queue(1) :+ 2) ++ Vector(3,4)`
`(1 +: Queue(2)) ++ Vector(3,4)`

`Queue(1) ++ Vector.empty`
`(Queue(1) :+ 2) ++ Vector.empty`
`(1 +: Queue(2)) ++ Vector.empty`

src/library/scala/collection/immutable/Queue.scala
edmundnoble: There's no need to apologize. I'm taking a look now.

mingthemerciless: It's not an apology. It's an expression of my distress at the level of testing and scrutiny that accompany changes to such important libraries. Thanks for investigating.

edmundnoble: Yet another app-breaking bug 😞 There is no such comprehensive, constructive testing for the Scala compiler or stdlib. All there is is regression tests; the entire Scala community is punished for every contributor's mistakes, until the next version, for months at a time. @Ichoran is working on it at the moment for collections; for now, I have the fix and can avoid actually reversing `that.seq`. Thanks a lot for the test-cases and report.

Edit: I don't know when this fix can get out to people; @SethTisue may have more.

test/files/run/t10298.scala
edmundnoble: All of these have been added. Thank you :)

src/library/scala/collection/immutable/Queue.scala
jvican: > There is no such comprehensive, constructive testing for the Scala compiler or stdlib. All there is is regression tests; the entire Scala community is punished for every contributor's mistakes, until the next version, for months at a time.
  
This seems to big of an statement.

mingthemerciless: Scary. Thanks for quick turnaround on the fix. 

How can I go about getting a picture of the test coverage on the standard library, contribute suggested test cases?

Ichoran: @mingthemerciless - You can try running collections-laws for a start, which ought to catch these sorts of things, and says how good its coverage is.  If it doesn't catch stuff like this, we should add a law.  We also should include it in the automatic test suite before release, instead of having me intermittently running it manually.  But I'm making it more friendly for other contributors (by making more of it Scala code instead of doing weird text replacement off of files with unique syntax), so maybe not just yet.

One of the problems with the test cases for the rest of the compiler and library is that there is no good way to get a picture of the test coverage.  The policy is generally to write sufficient tests to cover the interestingly different cases when fixing something (so at least things always get better), but that isn't clearly stated or carefully-enough checked for.  Getting an extra pair of eyes on PRs before they're merged is really helpful in this regard, because it's all too easy to glance at something quickly and think, "that looks okay".  (To help with that, just look for open PRs.)

SethTisue: the bigger questions here about test coverage, Scala, release timing, would really be better discussed on https://contributors.scala-lang.org

SethTisue: > We also should include (scala-collections-laws) in the automatic test suite

this is now https://github.com/scala/scala-dev/issues/411

PR #5892, opened 2017-05-07
[WIP] Implement annotation parsing in JavaParsers.

hrhino:
This causes the java parser to get somewhat more complex, but allows for macros and compiler plugins to correctly inspect java-defined classes for annotations.

Moved a few utility methods into `{Parsers,Scanners}Common` as well, so I can reuse them in the java ones.

This involves a change to the t4788 tests, as `SAnnotation` now correctly has `RetentionPolicy.SOURCE`, and is therefore ignored by the check in `BCodeHelpers#BCAnnotGen.shouldEmitAnnotation`; also, ASM Textifier emits an `// invisible` marker now that `CAnnotation` has `RetentionPolicy.CLASS`. I'm not sure what the chances that anyone is relying upon this behavior are, but since it does the Right Thing for separate compilation runs, they probably should not be too disappointed.

This still drops the annotations in `typ()` and `compilationUnit()`, because I don't know where the right place to put them is. `PackageDef` has `def mods = NoMods`, which I could change to `var mods = NoMods` but I'm not sure where I'd have to change to make them stick to which `Symbol`. `typ()` makes a `RefTree` and I have no idea how to reasonably annotate that.

Either way it should be better than what we have right now.

Review by @densh 

Fixes scala/bug#8928.

### DIFF
--- a/src/compiler/scala/tools/nsc/javac/JavaParsers.scala
+++ b/src/compiler/scala/tools/nsc/javac/JavaParsers.scala
@@ -234,11 +234,21 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
 
     // -------------------- specific parsing routines ------------------
 
-    def qualId(): RefTree = {
-      var t: RefTree = atPos(in.currentPos) { Ident(ident()) }
-      while (in.token == DOT) {
+    def qualId(orClassLiteral: Boolean = false): Tree = {
+      var t: Tree = atPos(in.currentPos) { Ident(ident()) }
+      var done = false
+      while (!done && in.token == DOT) {
         in.nextToken()
-        t = atPos(in.currentPos) { Select(t, ident()) }
+        t = atPos(in.currentPos) {
+          if (orClassLiteral && in.token == CLASS) {
+            in.nextToken()
+            done = true
+            val tpeArg = convertToTypeId(t)
+            TypeApply(q"_root_.scala.Predef.classOf", tpeArg :: Nil)
+          } else {
+            Select(t, ident())
+          }
+        }
       }
       t
     }
@@ -267,7 +277,7 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
       }
 
     def typ(): Tree = {
-      annotations()
+      annotations() // TODO: put these somewhere?
       optArrayBrackets {
         if (in.token == FINAL) in.nextToken()
         if (in.token == IDENTIFIER) {
@@ -326,20 +336,59 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
     }
 
     def annotations(): List[Tree] = {
-      //var annots = new ListBuffer[Tree]
+      var annots = new ListBuffer[Tree]
       while (in.token == AT) {
         in.nextToken()
-        annotation()
+        annots += annotation()
       }
-      List() // don't pass on annotations for now
+      annots.toList
     }
 
-    /** Annotation ::= TypeName [`(` AnnotationArgument {`,` AnnotationArgument} `)`]
+    /** Annotation ::= TypeName [`(` [AnnotationArgument {`,` AnnotationArgument}] `)`]
      */
-    def annotation() {
-      qualId()
-      if (in.token == LPAREN) { skipAhead(); accept(RPAREN) }
-      else if (in.token == LBRACE) { skipAhead(); accept(RBRACE) }
+    def annotation(): Tree = {
+      def annArg(): Tree = {
+        def annVal(): Tree = {
+          tryLiteral() match {
+            case Some(lit) => atPos(in.currentPos)(Literal(lit))
+            case _ if in.token == AT =>
+              in.nextToken()
+              annotation()
+            case _ if in.token == LBRACE =>
+              atPos(in.pos) {
+                val elts = inBracesOrNil(commaSeparated(annVal()))
+                Apply(ArrayModule_overloadedApply, elts: _*)
+              }
+            case _ if in.token == IDENTIFIER =>
+              qualId(orClassLiteral = true)
+          }
+        }
+
+        if (in.token == IDENTIFIER) {
+          qualId(orClassLiteral = true) match {
+            case name: Ident if in.token == EQUALS =>
+              in.nextToken()
+              /* name = value */
+              gen.mkNamedArg(name, annVal())
+            case rhs =>
+              /* implicit `value` arg with constant value */
+              gen.mkNamedArg(nme.value, rhs)
+          }
+        } else {
+          /* implicit `value` arg */
+          gen.mkNamedArg(nme.value, annVal())
+        }
+      }
+
+      atPos(in.pos) {
+        val id = convertToTypeId(qualId())
+        val args =
+          if (in.token == LPAREN) inParensOrNil {
+            if (in.token == RPAREN) Nil
+            else commaSeparated(atPos(in.pos)(annArg()))
+          } else Nil
+        New(id, args :: Nil)
+      }
     }
 
     def modifiers(inInterface: Boolean): Modifiers = {
@@ -353,7 +402,7 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
         in.token match {
           case AT if (in.lookaheadToken != INTERFACE) =>
             in.nextToken()
-            annotation()
+            annots :+= annotation()
           case PUBLIC =>
             isPackageAccess = false
             in.nextToken()
@@ -408,10 +457,10 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
 
     def typeParam(): TypeDef =
       atPos(in.currentPos) {
-        annotations()
+        val anns = annotations()
         val name = identForType()
         val hi = if (in.token == EXTENDS) { in.nextToken() ; bound() } else EmptyTree
-        TypeDef(Modifiers(Flags.JAVA | Flags.DEFERRED | Flags.PARAM), name, Nil, TypeBoundsTree(EmptyTree, hi))
+        TypeDef(Modifiers(Flags.JAVA | Flags.DEFERRED | Flags.PARAM, typeNames.EMPTY, anns), name, Nil, TypeBoundsTree(EmptyTree, hi))
       }
 
     def bound(): Tree =
@@ -435,7 +484,7 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
 
     def formalParam(): ValDef = {
       if (in.token == FINAL) in.nextToken()
-      annotations()
+      val anns = annotations()
       var t = typ()
       if (in.token == DOTDOTDOT) {
         in.nextToken()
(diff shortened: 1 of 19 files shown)


retronym:
This patch looks in very good shape already. I'd like to get it into 2.12.3, even though time is short.

I've got a few small suggestions in https://github.com/retronym/scala/tree/review/5892

  - Avoid using quasiquotes (so far, the compiler doesn't depend on them)
  - Initialize symbols that are newly referenced by the parser (the parser needs to be threadsafe, and forcing the lazy `.info` of `Symbol`-s while forming references to them isn't)
  - In general, prefer `isNonBottomSubclass` to `isSubclass`. Not sure if this was exploitable in this instance.

retronym:
It would be good to promote  `test/files/pos/t8928` to a `run` test in order to add another level of cross checking: do we get the consistent answers about the annotations via Scala runtime reflection? That is, is [`JavaMirrors::copyAnnotations`](https://github.com/scala/scala/blob/9acab45aeeadef2f63da69faf81465cc15599789/src/reflect/scala/reflect/runtime/JavaMirrors.scala#L676-L691) correct and being called in all the right places?

hrhino:
Thanks! I'll add those changes in; all good things to keep in mind.

I'm working on making `t8928` a `run` test, and I found something surprising in `JavaMirrors`:
```
      // todo. find out the exact order of assocs as they are written in the class file
      // currently I'm simply sorting the methods to guarantee stability of the output
      override lazy val assocs: List[(Name, ClassfileAnnotArg)] = (
        jann.annotationType.getDeclaredMethods.sortBy(_.getName).toList map (m =>
          TermName(m.getName) -> toAnnotArg(m.getReturnType -> m.invoke(jann))
        )
      )
```

It turns out that making it a `run` test as currently written won't work, because the test right now asserts ordering of the annotation parameters. Do you think it would be better to remove the ordering in the test and add a comment pointing to an issue in scala/bug (and then fix it in a separate PR), or to fix the ordering in runtime right now?

EDIT: that's not even close to trivial...

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -267,7 +277,7 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
       }
 
     def typ(): Tree = {
-      annotations()
+      annotations() // TODO: put these somewhere?
```
lrytz, line 280: the classfile parser also doesn't handle `RuntimeVisibleTypeAnnotations`, so you're in good company :)

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -326,20 +336,55 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
       while (in.token == AT) {
         in.nextToken()
-        annotation()
+        annots += annotation()
       }
-      List() // don't pass on annotations for now
+      annots.toList
     }
 
     /** Annotation ::= TypeName [`(` AnnotationArgument {`,` AnnotationArgument} `)`]
```
lrytz: The grammer is not correct here: http://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7

It should be ``Annotation ::= TypeName [`(` [AnnotationArgument {`,` AnnotationArgument}] `)`]``

Incidentally, the following crashes the compiler (after your patch):

```
➜  sandbox git:(be39803acc) ✗ cat A.java
public @interface A { }
class T {
  @A() void t2() { }
}
➜  sandbox git:(be39803acc) ✗ qsc A.java
error: scala.MatchError: None (of class scala.None$)
	at scala.tools.nsc.javac.JavaParsers$JavaParser.annVal$1(JavaParsers.scala:352)
	at scala.tools.nsc.javac.JavaParsers$JavaParser.annArg$1(JavaParsers.scala:379)
	at scala.tools.nsc.javac.JavaParsers$JavaParser.$anonfun$annotation$4(JavaParsers.scala:385)
	at scala.tools.nsc.ast.parser.ParsersCommon$ParserCommon.tokenSeparated(Parsers.scala:96)
	at scala.tools.nsc.ast.parser.ParsersCommon$ParserCommon.commaSeparated(Parsers.scala:107)
	at scala.tools.nsc.javac.JavaParsers$JavaParser.$anonfun$annotation$3(JavaParsers.scala:385)
```

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -326,20 +336,55 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
...
      */
-    def annotation() {
-      qualId()
-      if (in.token == LPAREN) { skipAhead(); accept(RPAREN) }
-      else if (in.token == LBRACE) { skipAhead(); accept(RBRACE) }
+    def annotation(): Tree = {
+      def annArg(): Tree = {
+        def annVal(): Tree = {
```
lrytz, line 351: This method is too restrictive. The [Java syntax](http://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7) allows `ConditionalExpression` here, which can be many things. For example, the following compiles with javac, but gives a syntax error with this PR:

```
➜  sandbox git:(be39803acc) ✗ cat A.java
public @interface A {
  int value();
}
class T {
  public static final int x = 1;
  @A(1 + x) void t2() { }
}
➜  sandbox git:(be39803acc) ✗ qsc A.java
A.java:6: error: `)' expected but `+' found.
  @A(1 + x) void t2() { }
       ^
A.java:7: error: illegal start of type
}
^
two errors found
```

If we make the field `x` non-final, it's no longer a compile-time constant, and the Java compiler shows an error "element value must be a constant expression". However, this is not a syntax error.

Supporting all valid Java annotations would require a huge amount of work: we'd have to start parsing Java expressions and constant fold them. In the Scala compiler this is currently type-driven (ConstantType), but in the case of java we already have a type for compile-time constants (`x` is `int` above). See also an older comment here: https://github.com/scala/bug/issues/5333#issuecomment-292401305

However, this should not stop us from supporting basic Java annotations. But we have to make sure that the Scala compiler doesn't reject Java code that the Java compiler accepts.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
```diff
@@ -894,10 +939,10 @@ trait JavaParsers extends ast.parser.ParsersCommon with JavaScanners {
       var pos = in.currentPos
       val pkg: RefTree =
         if (in.token == AT || in.token == PACKAGE) {
-          annotations()
+          annotations() // TODO: put these somewhere?
```
lrytz, line 946: interesting :) i wouldn't know where to put them anyway.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
hrhino: Ah, that's what I get for retyping it from memory rather than just copypasta.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
hrhino: Indeed. The fact that this gets called if `in.token == PACKAGE` seems surprising, as well.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
hrhino: Hmm. So, I can easily skip the value until the next comma (modulo parens). But... in that case, what would you expect the value to be? In the bit where we check to see if we can give a `static final` field a `ConstantType`, we give up and return the declared type if it's not a literal (which makes sense). 

However, I'm not sure what `Tree` we could give them. `EmptyTree`?
 `TypeTree().setType(whateverType)`? Someone will be surprised.

Should there be a warning? It seems possibly cruel to people who use `-Xfatal-warnings`, but since it should only affect Java source which the user can change, it's not the end of the world. If there were a way to make a `Tree` which warns the first time you look at it... but that's silly.

I could file a bug about that and label it "notatallquickfix", and we could call this "known restriction".

What do you think?

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
hrhino: Re-pushed with that fixed.

retronym:
I've re-assigned to 2.12.4, I think this is a bit too risky to add in so shortly before the 2.12.3 release. Thanks for your patience.

src/compiler/scala/tools/nsc/javac/JavaParsers.scala
retronym: Generally, we avoid any warnings for Java sources that run through the compiler.

Perhaps we should just plough ahead and add support for parsing `ConditionalExpression`. That's a necessary step to the full solution, but should be useful in and of itself, even before we can assign the right `ConstantType`.

lrytz: I also think an eager warning is too strict, there is probably a lot of Java code in mixed projects that (unnecessarily / unknowingly) goes through the Scala compiler.

It's always so easy to find a good reason for yet another compiler flag.. Maybe we can avoid that somehow, I don't have an idea right now though. I was thinking of a meta-annotation (or language import) that people could put on their Java code if they want strict Java annotation parsing. But because we're early in the compiler, there's no symbolic information available yet, so we cannot reliably identify a meta-annotation or import.

How to represent annotations with arguments that we cannot constant fold, I'm not sure. There are two pieces to the puzzle: one is arguments we cannot parse (because we don't parse Java expressions), and the other is expressions that we cannot constant fold. Maybe we can exclude the second because we only parse expressions that we can also constant fold. Just for reference, the following works (with the current form of this PR):
```
public @interface A {
  int value();
  public static class T {
    public static final int x = 1;
    @A(x) void t2() { }
  }
}
```

We could drop annotations entirely, or use `ErroneousAnnotation` (which already exists) without issuing an error. I think my preference is to drop them and have a strict mode to get warnings.

hrhino:
I think I'll take a look at @retronym's suggestion to "plough ahead". It causes possible sadness already unrelated to this PR; see scala/bug#10410.

adriaanm:
I'm really happy you're looking into this, but I would be more comfortable if this moved to 2.13. What do you think?

hrhino:
Please. I've been poking at it when I have time, but the full solution with constant parsing and such is a little harder than I was expecting when I picked up this ticket :)

If you'd like to close it for now I'm okay with that as well; I'll reopen it when it's ready.

adriaanm:
Thanks, I'm happy to change the milestone and keep it open.

adriaanm:
To keep the queue focused, I'll close this for now. Let me know if there's anything I can do to help.

hrhino:
Thanks! I'd like to take the suggestion to parse constants from Java, but that's not trivial and I have a lot on my plate these days.

retronym:
@hrhino Would you be interested in reviving this, at least to get the basics working? 

```
$ tail *.java *.scala
==> JavaClass.java <==
public class JavaClass {
	@Deprecated public static void foo() {}
}

==> ScalaClient.scala <==
class ScalaClient {
	JavaClass.foo()
}


$ scalac -d target JavaClass.java ScalaClient.scala -Xfatal-warnings -deprecation && javac -cp target -d target JavaClass.java # No deprecation warning with CompilerOrder.Mixed

$ javac -d target JavaClass.java && scalac -cp target -d target ScalaClient.scala -Xfatal-warnings -deprecation # deprecation warning is issued with CompilerOrder.JavaThenScala
ScalaClient.scala:2: warning: method foo in class JavaClass is deprecated: see corresponding Javadoc for more information.
	JavaClass.foo()
                  ^
one warning found
one error found

```

PR #5894, opened 2017-05-09
Broken link fixed in README.md

0zAND1z:
ENSIME link for non-IDEA Editors in #IDE Setup is now fixed. Reaches directly to http://ensime.org/editors/


adriaanm:
Thanks!

PR #5895, opened 2017-05-09
Optimizer: avoid inlining of accidental SAM types

lrytz:
Before, any Scala defined class that happened to have a SAM was considered
a function type for the inliner heuristics. This can lead to unexpected
inlining, large methods and degraded performance. Jason recently saw such a
case when adding an abstract method `mapOver` to `Type`:

```
def typed1(tree: Tree, mode: Mode, pt: Type): Tree = {
  def typedInAnyMode(tree: Tree): Tree = ...

  ... typedInAnyMode(tree) ...
}
```

Lambdalift adds a `pt: Type` parameter to `typedInAnyMode`, so the callsite
forwards the `pt` parameter. This tirggered inlining once `Type` was a SAM,
and method `typed1` became very large.

Now we only consider `scala.FunctionN` and types annotated `@FunctionalInterface`
as SAM types.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
@@ -550,7 +550,10 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
     val isEffectivelyFinal = classSym.isEffectivelyFinal
 
     val sam = {
-      if (classSym.isEffectivelyFinal) None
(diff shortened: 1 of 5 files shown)


lrytz:
I quickly ran `hot -p source=scalap` on my laptop (I didn't have a separate machine at hand, so just a quick test) before and after this change, the results look the same.

retronym:
The divergence in what the front and back-end consider to be a SAM type is a bit unfortunate. But I think that the heuristic you've used here is better than that status quo to avoid surprising over-inlining.

retronym (approved):

PR #5896, opened 2017-05-10
Don't attempt to publish if there's no private key.

milessabin:
Having the publish step only performed if the private key is available prevents [builds of forks from erroring](https://travis-ci.org/typelevel/scala/branches) unnecessarily. See gitter discussion [here](https://gitter.im/scala/contributors?at=5911803a4746d29d249a0a8d).

If this is merged I'll back/forward port to 2.11.x and 2.13.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,7 @@ env:
 
 # the key is restricted using forced commands so that it can only upload to the directory we need here
 after_success:
-  - if [ "${PRIV_KEY_SECRET}" != "" ] ; then
-      openssl aes-256-cbc -pass "pass:$PRIV_KEY_SECRET" -in spec/id_dsa_travis.enc -out spec/id_dsa_travis -d -a
-      chmod 600 spec/id_dsa_travis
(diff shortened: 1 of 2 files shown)


dwijnand:
Travis CI is configured to not build PRs in scala/scala. But this one should be the exception.

One way is by triggering it via API: https://docs.travis-ci.com/user/triggering-builds/.

Two issues I found right off the bat are:

* you need semicolons because the whole thing is presented as one bash command
* the last command needs to lose its single quotes

There might be more.

milessabin:
If this was my build I'd pull the whole thing out as an external script ... opinions?

dwijnand:
Personally I would go that way too. Much better than bash-in-YAML.

SethTisue:
>  pull the whole thing out as an external script

+1

milessabin:
Split out to an external script.

SethTisue:
thanks Miles!

SethTisue:
run at e.g. https://travis-ci.org/scala/scala/builds/247669362 looks OK

PR #5897, opened 2017-05-10
Identify partial functions in PartialFunction.apply

dwijnand:


dwijnand:
This is to avoid this trap, shared by @SystemFw:

```scala
val a: PartialFunction[Int,String] = {
  case 1 => "Normal"
}

val b = PartialFunction[Int,String] {
  case 1 => "isDefined is going to return always true!"
}
```

src/library/scala/PartialFunction.scala
```diff
@@ -245,7 +245,10 @@ object PartialFunction {
   /** Converts ordinary function to partial one
    *  @since   2.10
    */
-  def apply[A, B](f: A => B): PartialFunction[A, B] = { case x => f(x) }
+  def apply[A, B](f: A => B): PartialFunction[A, B] = f match {
+    case pf: PartialFunction[A, B] => pf
```
Jasper-M, line 249: I don't think this case will trigger, because the compiler only instantiates an instance of `PartialFunction[A,B]` when the expected type is `PartialFunction[A,B]`. So this will only help for 
```
val pf: PartialFunction[Int,String] = { case 1 => "1" }
val f = PartialFunction(pf)
```

dwijnand: You're right, this doesn't fix as much as I'd hoped. :(

```scala
scala> def funcToPf[A, B](f: A => B): PartialFunction[A, B] = f match {
     |   case pf: PartialFunction[A, B] => pf
     |   case _                         => { case x => f(x) }
     | }
funcToPf: [A, B](f: A => B)PartialFunction[A,B]

scala> funcToPf[Int, String]({ case 1 => "one" }) isDefinedAt 2
res2: Boolean = true

scala> funcToPf[Int, String]({ case 1 => "one" }: PartialFunction[Int, String]) isDefinedAt 2
res3: Boolean = false
```

dwijnand:
As correctly pointed out by @Jasper-M this doesn't solve that trap.

Perhaps `{ case ..` notation should always return partial functions, but that would require a SIP.

Jasper-M:
Perhaps the best thing to do other than a SIP is:
```
def apply[A, B](f: PartialFunction[A, B]): PartialFunction[A, B] = f
def fromFunction[A, B](f: A => B): PartialFunction[A, B] = { case x => f(x) }
```

SystemFw:
I think @Jasper-M solution is the most appropriate. `apply` is too neutral a name for this sort of conversion, and I think it's only motivated by having a very convenient notation for passing a `Function` where a `PartialFunction` is expected (which I think ultimately points out that the sub typing relation between `Function` and `PartialFunction` is backwards)

dwijnand:
But that would be a backwards incompatible change. :-/

SethTisue:
this could be revived for 2.13.x. deprecate `apply`, add `fromFunction`

szeiger:
See https://github.com/scala/scala/pull/5698 for a change to PF type inference in 2.13, including a discussion of the different meanings of pattern-matching anonymous functions depending on expected type.

PR #5898, opened 2017-05-10
Don't stripmargin source text

som-snytt:
Introduced by c3aca109e95e2259d9909f8457a1422c5c995940

Fixes scala/bug#9016

Other bugs fall under the umbrella issue but are sure to be fixed by now in `Adriaanite`.

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/IMain.scala
+++ b/src/repl/scala/tools/nsc/interpreter/IMain.scala
@@ -865,7 +865,7 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends 
         |${preambleHeader format lineRep.readName}
         |${envLines mkString ("  ", ";\n  ", ";\n")}
         |$importsPreamble
-        |${indentCode(toCompute)}""".stripMargin
+        |%s""".stripMargin.format(indentCode(toCompute))
       def preambleLength = preamble.length - toCompute.length - 1
 
       val generate = (m: MemberHandler) => m extraCodeToEvaluate Request.this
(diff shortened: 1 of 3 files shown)


dwijnand (approved):
Amazing.

src/repl/scala/tools/nsc/interpreter/IMain.scala
```diff
@@ -865,7 +865,7 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends
         |${preambleHeader format lineRep.readName}
         |${envLines mkString ("  ", ";\n  ", ";\n")}
         |$importsPreamble
-        |${indentCode(toCompute)}""".stripMargin
+        |%s""".stripMargin.format(indentCode(toCompute))
```
som-snytt, line 868: Maybe `stuff + indentCode` is more straightforward here.

adriaanm: Good catch! I was just in the same area in my repl-modularize branch. I would get rid of `stripMargin` entirely and assemble with `+`. The next line also looks fishy:
```
      def preambleLength = preamble.length - toCompute.length - 1
```

Where does this `-1` come from? The assumption that `indentCode` appends a `\n`?

som-snytt: That goes back to retronym completion. I'd have to investigate. I see it changed in my change you reverted? Presumably it was off by one at some point. https://github.com/scala/scala/commit/f279894a3efe84b85e9becfd0ce96aaa0c21bfbd

It should really import int + or - => offWhichWay, then `len offWhichWay 1`.

adriaanm: I'll roll that investigation into my PR (coming once the currently in-flight repl PRs are merged).

PR #5901, opened 2017-05-12
Pretty print symbolic names in the REPL

Jasper-M:
Before, when you defined or imported things in the REPL, the encoded name is shown to the user. For some reason only vals are currently pretty printed:

```
scala> def + = ???
$plus: Nothing

scala> class +
defined class $plus

scala> type + = Nothing
defined type alias $plus

scala> object * { def + = ??? }
defined object $times

scala> import *.{+ => -}
import $times.{$plus=>$minus}

scala> lazy val + = ???
+: Nothing = <lazy>

scala> val + = null
+: Null = null
```
New behavior:

```
scala> def + = ???
+: Nothing

scala> class +
defined class +

scala> type + = Nothing
defined type alias +

scala> object * { def + = ??? }
defined object *

scala> import *.{+ => -}
import *.{+ => -}

scala> lazy val + = ???
+: Nothing = <lazy>

scala> val + = null
+: Null = null
```
For the imports I had to duplicate some code from `Printers`. An alternative would be to make `TreePrinter` print decoded names for imports, but I guess there's a reason why that isn't being done currently.

### DIFF
--- a/src/reflect/scala/reflect/internal/Printers.scala
+++ b/src/reflect/scala/reflect/internal/Printers.scala
@@ -49,11 +49,16 @@ trait Printers extends api.Printers { self: SymbolTable =>
   /** Turns a path into a String, introducing backquotes
    *  as necessary.
    */
-  def backquotedPath(t: Tree): String = {
+  def backquotedPath(t: Tree): String = backquotedPathInternal(t, symName)
+  def decodedBackquotedPath(t: Tree): String = backquotedPathInternal(t, decodedSymName)
+  
+  private def backquotedPathInternal(t: Tree, stringify: (Tree, Name) => String): String = {
     t match {
-      case Select(qual, name) if name.isTermName  => s"${backquotedPath(qual)}.${symName(t, name)}"
-      case Select(qual, name) if name.isTypeName  => s"${backquotedPath(qual)}#${symName(t, name)}"
-      case Ident(name)                            => symName(t, name)
+      case Select(qual, name) if name.isTermName  => 
+        s"${backquotedPathInternal(qual, stringify)}.${stringify(t, name)}"
+      case Select(qual, name) if name.isTypeName  => 
+        s"${backquotedPathInternal(qual, stringify)}#${stringify(t, name)}"
+      case Ident(name)                            => stringify(t, name)
       case _                                      => t.toString
     }
   }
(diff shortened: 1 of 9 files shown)


Jasper-M:
Review @retronym @som-snytt ?

som-snytt:
@adriaan 's `Adriaanite` REPL will surely have a refactored facility for pluggable printing.

There's also in-flight https://github.com/scala/scala/pull/5841

The duplication is too bad. I can't tell at a glance what is intended to be decoded in Printers. With all those flags, I'd expect a boolean that says if I want decoded names. The boolean param with default is classically hard to follow.

src/reflect/scala/reflect/internal/Printers.scala
```diff
@@ -49,11 +49,16 @@ trait Printers extends api.Printers { self: SymbolTable =>
   /** Turns a path into a String, introducing backquotes
    *  as necessary.
    */
-  def backquotedPath(t: Tree): String = {
+  def backquotedPath(t: Tree): String = backquotedPathInternal(t, symName)
+  def decodedBackquotedPath(t: Tree): String = backquotedPathInternal(t, decodedSymName)
+  
+  private def backquotedPathInternal(t: Tree, toString: (Tree, Name) => String): String = {
```
som-snytt: Maybe a name other than `toString`? `printer`, `nameFn`.

src/reflect/scala/reflect/internal/Printers.scala
```diff
@@ -49,11 +49,16 @@ trait Printers extends api.Printers { self: SymbolTable =>
...
     t match {
-      case Select(qual, name) if name.isTermName  => s"${backquotedPath(qual)}.${symName(t, name)}"
-      case Select(qual, name) if name.isTypeName  => s"${backquotedPath(qual)}#${symName(t, name)}"
-      case Ident(name)                            => symName(t, name)
+      case Select(qual, name) if name.isTermName  => 
+        s"${backquotedPathInternal(qual, toString)}.${toString(t, name)}"
+      case Select(qual, name) if name.isTypeName  => 
+        s"${backquotedPathInternal(qual, toString)}#${toString(t, name)}"
```
som-snytt: I can't tell if this includes `pkg.C` as `pkg#C`.

Jasper-M: I see what you mean. That should be the same logic as before though...

Jasper-M: Tested, and it seems to work correctly.

retronym:
The REPL has been substantially refactored in the 2.13.x branch, with more changes planned. I'd encourage you to resubmit this change to that branch so we can avoid merge conflicts and have a bit more freedom to perform the necessary refactorings in the tree printing code to avoid the need for copy/paste. Thanks!

PR #5903, opened 2017-05-13
REPL: decouple the UI (shell) from the compiler [ci: last-only]

adriaanm:
Split in front- and back-end. Drop shaded jline (was needed for spark shell; we should be able to offer a better solution with the coming frontend improvements).

Review follow-ups:
  - [x] compare to 05cc3e2 for error message indenting
  - [x] tab completion on first line should also result in tab completion once repl initializes
  - [x] check we're ok with different sbt versions (standardization work on API for sbt / spark / .. will be a separate effort -- forthcoming!)

Known (temporary) regressions in this PR:
  - lost syntax highlighting because it was done in the compiler, while it should be in the shell (UI)
  - similarly, the repl no longer exposes an object that can be used to manipulate the shell, because we're trying to reduce this coupling 
  - error reporting uses a heuristic to determine whether the offending line is on the console (to avoid repeating it) -- can we improve this?


som-snytt:
Now I know why it's called a shell game.

src/repl-frontend/scala/tools/nsc/interpreter/shell/InteractiveReader.scala
```diff
@@ -44,18 +44,14 @@ object InteractiveReader {
     }
 
   def apply(): InteractiveReader = SimpleReader()
-
-  // a non-interactive InteractiveReader that returns the given text
```
som-snytt: that comment gave me a chuckle; I almost pointed it out in the PR.

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -50,7 +49,34 @@ class  ReplReporter(intp: IMain) extends ConsoleReporter(intp.settings, Console.
...
+       val lineContent = posIn.lineContent
+
+       val preambleLines = intp.currentRequest.preambleEndPos.line
+
+       super.printMessage(s"On line ${posIn.line - preambleLines}:")
+       super.printMessage(indentation + lineContent)
+       super.printMessage(indentation + posIn.lineCaret)
+       super.printMessage(indented(msg))
+     }
+   }
```
som-snytt: Great idea to lead with caret (and not with a stick). I'd forgotten this was part of that reversion. I see I needed to pick out pastes, which are not indented? For reference 05cc3e2271d2c2226ded33bef5a03f0c70c6f66d

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -63,7 +89,7 @@ class  ReplReporter(intp: IMain) extends ConsoleReporter(intp.settings, Console.
     else Console.println("[init] " + msg)
   }
 
-  override def displayPrompt() {
+  override def displayPrompt(): Unit = {
     if (intp.totalSilence) ()
     else super.displayPrompt()
   }
```
som-snytt: From the reverted, `override def displayPrompt() = if (!intp.totalSilence) super.displayPrompt()`  There must be a word for this style of supplying explicit unit values. `Unit saturation`? There is a ticket where the lub crashes...

src/repl/scala/tools/nsc/interpreter/IMain.scala
```diff
@@ -817,6 +817,15 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends
     def value      = Some(handlers.last) filter (h => h.definesValue) map (h => definedSymbols(h.definesTerm.get)) getOrElse NoSymbol
     val lineRep = new ReadEvalPrint()
 
+    private val preambleEndMember = "$ipos"+{lineRep.lineId}
+    // don't look until we've (tried) attributing the trees in unit!
+    lazy val preambleEndPos: Position =
+      unit.body.find {
+        case td: TypeDef => td.name.toString == preambleEndMember
+        case _ => false
+      }.map(_.pos).getOrElse(NoPosition)
```
som-snytt: Clever! Another idea is an innocuous import. How about a macro called `$up` or `^` that emits the innocuous marker tree? We just watched our disk of "Up" again, a very effective movie. "Squirrel!"

src/repl-frontend/scala/tools/nsc/interpreter/shell/InteractiveReader.scala
```diff
@@ -34,6 +32,8 @@ trait InteractiveReader {
     // hack necessary for OSX jvm suspension because read calls are not restarted after SIGTSTP
     if (scala.util.Properties.isMac) restartSysCalls(readOneLine(prompt), reset())
     else readOneLine(prompt)
+
+  def initCompletion(completion: Completion): Unit = {}
```
som-snytt: I would encourage unit saturation here.  To review, `postInit` only ever meant `initCompletion`, but I wonder if `sbt` would like a hook somewhere, there's also `loopPostInit` or similar. Cf https://github.com/scala/scala/pull/5655

src/repl-frontend/scala/tools/nsc/interpreter/shell/InteractiveReader.scala
```diff
@@ -117,31 +115,7 @@ class SplashLoop(reader: InteractiveReader, prompt: String) extends Runnable {
...
-
-  override protected[interpreter] def readOneLine(prompt: String): String = ???   // unused
-  override protected[interpreter] def readOneKey(prompt: String): Int     = ???   // unused
-
-  override def readLine(prompt: String): String = reader.readLine(prompt)
-}
-object SplashReader {
-  def apply(reader: InteractiveReader)(postIniter: InteractiveReader => Unit) =
-    new SplashReader(reader, postIniter)
-}
```
som-snytt: One annoying (to me) behavior is that trying tab completion on first line is thrown away. I'd like it to wait for a PC and then autocomplete. Otherwise I have to hit tab for a couple of seconds (feels like ten). I've gotten into the habit of entering a dummy value just to wait for warm compiler.

adriaanm:
(Now that I have the tests passing, I'll work on addressing the first round of review feedback. Thanks!)

src/repl-frontend/scala/tools/nsc/interpreter/shell/InteractiveReader.scala
adriaanm: I don't think sbt used `postInit`, if that's what you're concerned about? I in-sourced its console interface to be sure.

Are you saying the unit saturation levels are sufficient, or are you encouraging more of it? If so, how? :-)

src/repl-frontend/scala/tools/nsc/interpreter/shell/InteractiveReader.scala
adriaanm: Ditto. I'll add that as a follow-up.

src/repl-frontend/scala/tools/nsc/interpreter/shell/InteractiveReader.scala
som-snytt: I think I was thinking that I prefer `def f(): Unit = ()` style instead of braces. Are braces the new tabs?  Obviously it doesn't matter.

src/repl-frontend/scala/tools/nsc/interpreter/shell/ILoop.scala
```diff
@@ -432,15 +452,12 @@ class ILoop(defaultIn: BufferedReader = null, protected val out: JPrintWriter =
...
-      savingReplayStack {
-        file applyReader { reader =>
-          in = if (verbose) new SimpleReader(reader, out, interactive = true) with EchoReader
-               else SimpleReader(reader, out, interactive = false)
-          echo(s"Loading $file...")
-          loop()
-        }
+  def interpretAllFrom(file: File, verbose: Boolean = false): Unit = {
+    savingReplayStack {
+      file applyReader { reader =>
```
adriaanm: - [x] regression -- this did more than changing verbosity, also need to pass in new reader

src/repl/scala/tools/nsc/interpreter/Power.scala
```diff
@@ -105,17 +106,12 @@ class Power[ReplValsImpl <: ReplVals : ru.TypeTag: ClassTag](val intp: IMain, re
...
+//        repldbg("Not a package class! " + packageClass)
         Set()
       }
     }
   }
 
-  private def customBanner = replProps.powerBanner.option flatMap {
-    case f if f.getName == "classic" => Some(classic)
-    case f => io.File(f).safeSlurp()
-  }
```
som-snytt, line 118: No more paulp classic banner? Or do you plan to bring it back? Maybe as a `motd` feature?

src/repl/scala/tools/nsc/interpreter/ReplProps.scala
```diff
@@ -3,26 +3,27 @@
...
-import java.util.{ Formattable, FormattableFlags, Formatter }
 
-class ReplProps {
+trait ReplProps {
   private def bool(name: String) = BooleanProp.keyExists(name)
   private def int(name: String)  = Prop[Int](name)
 
   // This property is used in TypeDebugging. Let's recycle it.
-  val colorOk = Properties.coloredOutputEnabled
+  private val colorOk = Properties.coloredOutputEnabled
```
som-snytt: It would be so nice if `private` had `private[this]` semantics. I don't suppose you could sneak that in under cover of the construction dust.

src/repl/scala/tools/nsc/interpreter/ReplProps.scala
```diff
@@ -46,23 +47,11 @@ class ReplProps {
...
-  }
-  val continuePrompt = encolor(continueText)
-
-  // Next time.
-  //def welcome = enversion(Prop[String]("scala.repl.welcome") or shellWelcomeString)
-  def welcome = enversion {
-    val p = Prop[String]("scala.repl.welcome")
-    if (p.isSet) p.get else shellWelcomeString
-  }
+  val welcomeString  = Prop[String]("scala.repl.welcome").option getOrElse shellWelcomeString
```
som-snytt: The `Next time` comment promises that `Prop or other` is available now.

som-snytt: Is this the next time?

src/repl/scala/tools/nsc/interpreter/ReplProps.scala
```diff
@@ -73,10 +62,25 @@ class ReplProps {
...
   val format = Prop[String]("scala.repl.format")
 
-  val replAutorunCode = Prop[JFile]("scala.repl.autoruncode")
-  val powerInitCode   = Prop[JFile]("scala.repl.power.initcode")
-  val powerBanner     = Prop[JFile]("scala.repl.power.banner")
+  val replAutorunCode = Prop[File]("scala.repl.autoruncode")
+  val powerInitCode   = Prop[File]("scala.repl.power.initcode")
+  val powerBanner     = Prop[File]("scala.repl.power.banner")
 
-  val vids = bool("scala.repl.vids")
```
som-snytt: FSR, I want to add, "video killed the radio star."

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -3,23 +3,67 @@
...
+
+  var unwrapStrings: Boolean
+  def unmangle(s: String): String
+
+  // compiler is initialized
+  def initializeComplete: Boolean
+
+  def lastInputSingleLine: Boolean
+  def preambleLineDelta: Int
+}
```
som-snytt: Repl gets some config chops, not to be confused with settings or isettings etc! Or is it ReplProps? I could never remember.

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -3,23 +3,67 @@
...
+
+// TODO: move this out of imain and into iloop
+/** Like ReplGlobal, a layer for ensuring extra functionality. */
```
som-snytt: My least favorite comment. `As opposed to totally gratuitous layers which ensure nothing but screw with your stack trace, ha, ha, ha!`

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -3,23 +3,67 @@
...
- */
-class  ReplReporter(intp: IMain) extends ConsoleReporter(intp.settings, Console.in, new ReplStrippingWriter(intp)) {
   def printUntruncatedMessage(msg: String) = withoutTruncating(printMessage(msg))
 
   /** Whether very long lines can be truncated.  This exists so important
    *  debugging information (like printing the classpath) is not rendered
    *  invisible due to the max message length.
    */
-  private var _truncationOK: Boolean = !intp.settings.verbose
+  private var _truncationOK: Boolean = !settings.verbose
```
som-snytt: The leading underscore means it could have been initialized with an underscore. Maybe I should add a StackOverflow Q&A for that?

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
```diff
@@ -105,9 +146,22 @@ class  ReplReporter(intp: IMain) extends ConsoleReporter(intp.settings, Console.
...
+
+  import config._
+  private def strip(str: String): String =
+    if (unwrapStrings) unmangle(str) else str
+
+  private def truncate(str: String): String =
+    if (isTruncating && (maxPrintString != 0 && str.length > maxPrintString))
+      (str take maxPrintString - 3) + "..."
+    else str
+}
```
som-snytt: A StrippingWriter is when I fantasize about J. K. Rowling. Who is also a PrintWriter, a writer in print, pant pant.

I'd thumbsup `DemanglingWriter` with similar verbs like `demangle`, maybe by the third step in modularizing REPL. I mean WREPL. Where the W is for wrapped writer, of course.

src/repl/scala/tools/nsc/interpreter/ReplVals.scala
```diff
@@ -17,12 +17,11 @@ import scala.reflect.runtime.{universe => ru}
 abstract class ReplVals { }
 
 class StdReplVals(final val intp: IMain) extends ReplVals {
-  // TODO bring back repl, reader, completion and history?
+  // TODO bring back repl, reader, isettings, completion and history?
```
som-snytt: please just don't call it `isettings`, if it must be recalled. I realize these names were conceived in the aughts, but one ought not `i-` everything.

src/repl/scala/tools/nsc/interpreter/ReplVals.scala
```diff
@@ -17,12 +17,11 @@ import scala.reflect.runtime.{universe => ru}
...
   // TODO should intp be lazy?
   final lazy val power                    = intp.power
   final lazy val vals                     = this
   final lazy val global: intp.global.type = intp.global
-  final lazy val isettings                = intp.isettings
```
som-snytt: 👍

src/repl/scala/tools/nsc/interpreter/package.scala
```diff
@@ -33,186 +23,15 @@ import scala.util.Try
...
-  type JClass         = java.lang.Class[_]
-  type JList[T]       = java.util.List[T]
-  type JCollection[T] = java.util.Collection[T]
-  type JPrintWriter   = java.io.PrintWriter
-  type InputStream    = java.io.InputStream
-  type OutputStream   = java.io.OutputStream
-
-  val IR = Results
-
-  implicit def postfixOps = scala.language.postfixOps // make all postfix ops in this package compile without warning
```
som-snytt, line 44: Because what can go wrong with excellent coders working on an excellent code base?

src/repl/scala/tools/nsc/interpreter/package.scala
```diff
@@ -33,186 +23,15 @@ import scala.util.Try
...
-
-  def staticTypeTag[T: ClassTag]: ru.TypeTag[T] = ru.TypeTag[T](
-    ru.runtimeMirror(ourClassloader),
-    new TypeCreator {
-      def apply[U <: ApiUniverse with Singleton](m: Mirror[U]): U # Type =
-        m.staticClass(classTag[T].runtimeClass.getName).toTypeConstructor.asInstanceOf[U # Type]
-  })
-
-  /** This class serves to trick the compiler into treating a var
-   *  (intp, in ILoop) as a stable identifier.
```
som-snytt, line 65: I know it tricked me.

src/repl/scala/tools/nsc/interpreter/Power.scala
adriaanm: That'll be an enterprise feature. Just kidding. It moved a bit closer to the co-source of the banner printer.

src/repl-frontend/scala/tools/nsc/interpreter/shell/ILoop.scala
```diff
@@ -1137,7 +1137,12 @@ object ILoop {
           }
         }
 
-        val repl = new ILoop(ShellConfig(settings), input, output)
+        val config = ShellConfig(settings)
+        val repl = new ILoop(config, input, output) {
+          // remove versioning info (for reproducible test results),
+          // or suppress altogether if we're not welcoming
+          override def welcome = if (welcoming) config.welcomeString else ""
+        }
```
som-snytt: That's a welcome change.

som-snytt:
I just came across an old example of unit saturation with emoticon:
```
def start()     = { futureThread ;() }
```
It seemed like a great design choice at the time.  A winking funny face.

test/files/run/repl-paste-parse.check
```diff
@@ -1,5 +1,5 @@
 
-scala>
+scala>
```
som-snytt: This is, of course, REPL's revenge for the local suffix space thing.

src/repl-frontend/scala/tools/nsc/interpreter/shell/ILoop.scala
adriaanm: :-D

test/files/run/repl-paste-parse.check
adriaanm: It's outspacing me for now. Some interesting design choices around the splash loop, where to get input from, and that loopy load command.

test/files/run/t10284.scala
```diff
@@ -5,9 +5,7 @@
 import scala.tools.partest.{ReplTest, Welcoming}
 import scala.tools.nsc.{GenericRunnerSettings, Settings}
 
-// Welcoming just fakes ReplTest into not stripping leading lines of output
-// since REPL doesn't emit a header for -e
-object Test extends ReplTest with Welcoming {
+object Test extends ReplTest {
```
som-snytt, line 8: Not very welcoming.

som-snytt:
Since paulp [modularized the repl](https://github.com/scala/scala/commit/48cc8b47fcadaa187026ca0422178c9094e4b412) four years ago, maybe change the title to "A second step in modularizing the repl."

adriaanm:
I squashed my stream of consciousness commits. Will follow up on your comments tomorrow.

som-snytt:
I forgot to say :) a couple of times. I know working on REPL is relatively thankless. Or REPLatively.

One thing to decouple is the particulars of wrapping. I had one branch where wrapping happens as a phase that can be swapped in via plugin. I still like that model for accommodating various runtime behaviors. For the longer-time. I'd be happy to do that again on top of `Adriaanite`.

adriaanm:
The inferred :-) conferred plenty of :-) here :-)

Regarding doing the wrapping in a separate phase, that's exactly what Jason suggested when we discussed next steps for the rework yesterday! I'm going to try to wrap this PR up before I head to Copenhagen (all of next week). After that, I'm planning to work on upgrading to jline3, implementing parsing-as-you-type and multi-line edits, and bringing back syntax highlighting (the only regression due to the split between repl backend / shell in this PR).

I'll create some tickets on scala/scala-dev to divide up the work today. In addition to reworking the wrapping for run-time semantics, we should also bring back #5220

src/repl/scala/tools/nsc/interpreter/package.scala
adriaanm: sadly, for now, the rabbit just moved to a different hat (will domesticate it further, but probably not in this PR)

test/files/run/t10284.scala
adriaanm: No need to fake it anymore.

src/repl/scala/tools/nsc/interpreter/ReplProps.scala
adriaanm: I hesitate to add more stuff to the Prop abstraction.

src/repl/scala/tools/nsc/interpreter/ReplReporter.scala
adriaanm: Yep! I'd like to at least make a clear distinction between configuring the shell (the UI, `ShellConfig`) and the interpreter (classpath and other run-time characteristics, `interpreterSettings`) and the backing compiler (`settings`)

src/repl/scala/tools/nsc/interpreter/ReplVals.scala
adriaanm: roger

src/repl/scala/tools/nsc/interpreter/package.scala
adriaanm: The whole package object must go!

src/repl-frontend/scala/tools/nsc/interpreter/shell/Reporter.scala
```diff
@@ -176,23 +185,41 @@ class ReplReporterImpl(val config: ShellConfig, settings: Settings = new Setting
...
+    if (!totalSilence) printlnAndFlush(msg)
+    else if (isTrace) printlnAndFlush("[silent] " + msg)
 
   override def displayPrompt(): Unit =
-    if (!totalSilence) super.displayPrompt()
+    if (!totalSilence) {
+      out.println()
+      out.print("a)bort, s)tack, r)esume: ")
+      out.flush()
+      Console.in.read match { // TODO: use repl reader?
```
som-snytt, line 200: I guess the new test will tell you.  https://github.com/scala/scala/blob/v2.12.2/test/junit/scala/tools/nsc/reporters/ConsoleReporterTest.scala#L144

som-snytt:
+1 for a direct approach with a sit-down. Please ask Adriaan stand a round, as he can probably expense it. He's so stand-up, he'd probably stand it out of pocket. Good luck to all.

adriaanm:
I can stand around with the best of 'em.

som-snytt:
I wish everyone could enjoy that bon mot. It really lifted my spirits. To illustrate, I'm listening on repeat to my child's favorite Rachel Platten https://www.youtube.com/watch?v=bwB9EMpW8eY

I would like that every contribution to FOSS expresses, "Love, you're not alone."

eed3si9n:
Not sure what the implication would be, but in Zinc 1 we are implementing more fine-grained Console API that sbt-pamflet implemented to emulate console in markdown. We can discuss more in the direct stand-down - https://github.com/sbt/zinc/blob/1.0/internal/compiler-bridge/src/main/scala/xsbt/InteractiveConsoleInterface.scala

adriaanm:
@retronym, I'd like to see if we can get this reviewed and merged next week. I've gone over it with @lrytz in person -- happy to schedule a call if that helps digesting the change.

som-snytt:
You don't actually digest change, it just goes right through you. Like when a toddler swallows a dime. The problem is when they inhale it instead. So the lesson is, don't inhale.

build.sbt
```diff
@@ -640,7 +640,7 @@ lazy val test = project
     fork in IntegrationTest := true,
     javaOptions in IntegrationTest += "-Xmx2G",
     testFrameworks += new TestFramework("scala.tools.partest.sbt.Framework"),
-    testOptions in IntegrationTest += Tests.Argument("-Dpartest.java_opts=-Xmx1024M -Xms64M"),
+    testOptions in IntegrationTest += Tests.Argument("-Dpartest.java_opts=-Xmx1024M -Xms64M"), // -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005
```
lrytz: nit - remove that comment

retronym:
Not sure yet if the problem lies in my `scala-pr` script or in the POM files published after this change, but:

```
⚡ (set -x; scala-pr 5903)
+ scala-pr 5903
+ PR=5903
+ shift
++ scala-pr-version 5903
++ PR=5903
++ scala-ref-version pull/5903/head
++ REF=pull/5903/head
++ shift
+++ curl -H 'Authorization: token XXXXXXXXX' --silent https://api.github.com/repos/scala/scala/git/refs/pull/5903/head
+++ jq -r '.object.sha[0:7]'
++ SHORT_SHA=87f41ea
+++ curl --silent -H 'Authorization: token XXXXXXX' -H 'Accept: application/vnd.github.VERSION.raw' 'https://api.github.com/repos/scala/scala/contents/build.sbt?ref=pull/5903/head'
+++ grep 'baseVersion in Global'
+++ gsed -r 's/.*"(.*)".*/\1/'
++ BASE_VERSION=2.13.0
++ CROSS=-bin
++ echo 2.13.0
++ grep -q -E '\d+\.\d+.0'
++ CROSS=-pre
++ VERSION=2.13.0-pre-87f41ea-SNAPSHOT
++ echo 2.13.0-pre-87f41ea-SNAPSHOT
+ scala-launch 2.13.0-pre-87f41ea-SNAPSHOT
+ V=2.13.0-pre-87f41ea-SNAPSHOT
+ shift
+ _scala_launch scala.tools.nsc.MainGenericRunner 2.13.0-pre-87f41ea-SNAPSHOT -usejavacp
+ RUNNER=scala.tools.nsc.MainGenericRunner
+ shift
+ VERSION=2.13.0-pre-87f41ea-SNAPSHOT
+ shift
++ coursier fetch -q -p -r https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots -r https://scala-ci.typesafe.com/artifactory/scala-integration org.scala-lang:scala-compiler:2.13.0-pre-87f41ea-SNAPSHOT
+ java -classpath /Users/jz/.coursier/cache/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.13.0-M1/1.0.6/scala-xml_2.13.0-M1-1.0.6.jar:/Users/jz/.coursier/cache/v1/https/scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-compiler/2.13.0-pre-87f41ea-SNAPSHOT/scala-compiler-2.13.0-pre-87f41ea-20170525.235750-1.jar:/Users/jz/.coursier/cache/v1/https/scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-reflect/2.13.0-pre-87f41ea-SNAPSHOT/scala-reflect-2.13.0-pre-87f41ea-20170525.235750-1.jar:/Users/jz/.coursier/cache/v1/https/scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-library/2.13.0-pre-87f41ea-SNAPSHOT/scala-library-2.13.0-pre-87f41ea-20170525.235750-1.jar scala.tools.nsc.MainGenericRunner -usejavacp
Welcome to Scala 2.13.0-20170525-235632-87f41ea (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).
Exception in thread "main" java.lang.NoClassDefFoundError: jline/console/history/History
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at scala.tools.nsc.interpreter.jline.InteractiveReader.<init>(JLineReader.scala:31)
	at scala.tools.nsc.interpreter.shell.ILoop.defaultIn$lzycompute(ILoop.scala:62)
	at scala.tools.nsc.interpreter.shell.ILoop.defaultIn(ILoop.scala:59)
	at scala.tools.nsc.interpreter.shell.ILoop.$anonfun$run$1(ILoop.scala:947)
	at scala.tools.nsc.interpreter.shell.ILoop.run(ILoop.scala:944)
	at scala.tools.nsc.MainGenericRunner.runTarget$1(MainGenericRunner.scala:80)
	at scala.tools.nsc.MainGenericRunner.run$1(MainGenericRunner.scala:83)
	at scala.tools.nsc.MainGenericRunner.process(MainGenericRunner.scala:94)
	at scala.tools.nsc.MainGenericRunner$.main(MainGenericRunner.scala:99)
	at scala.tools.nsc.MainGenericRunner.main(MainGenericRunner.scala)
Caused by: java.lang.ClassNotFoundException: jline.console.history.History
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 46 more
```

retronym:
Looks like `jline_embedded` is absent from `scala-compiler.jar`.

```
⚡ jar tf /Users/jz/.coursier/cache/v1/https/scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-compiler/2.13.0-pre-87f41ea-SNAPSHOT/scala-compiler-2.13.0-pre-87f41ea-20170525.235750-1.jar | grep console/history/History

⚡ jar tf /Users/jz/.coursier/cache/v1/https/scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-compiler/2.13.0-pre-1521e00-SNAPSHOT/scala-compiler-2.13.0-pre-1521e00-20170530.172855-1.jar | grep console/history/History
scala/tools/jline_embedded/console/history/History$Entry.class
scala/tools/jline_embedded/console/history/History.class
```

retronym:
I see this was intended in the commit "Split REPL into front / back; drop embedded jline", but maybe we need to introduce the hard dependency on the actual JLine in its place?

```scala
lazy val compiler = configureAsSubproject(project)
   ...
    // These are only needed for the POM:
    libraryDependencies ++= Seq(scalaXmlDep, jlineDep % "optional"),
```

adriaanm:
> libraryDependencies ++= Seq(scalaXmlDep, jlineDep % "optional"),

Could the issue be with coursier? "optional" means "non-transitive" in maven-speak, as far as I know. It's not optional in the sense that it needn't be on the classpath (that's "provided")

adriaanm:
Looks like `keepOptional` defaults to `false`:  https://github.com/coursier/coursier/blob/7b1bbae2085bbcc075c803ce3aaf15c6b9bd80f4/cli/src/main/scala-2.11/coursier/cli/Options.scala#L10

src/repl-frontend/scala/tools/nsc/interpreter/shell/Reporter.scala
```diff
@@ -0,0 +1,225 @@
...
+            case -1 => (msg, "")
+            case n => (msg.substring(0, n), msg.substring((n + 1) min msg.length))
+          }
+
+        // add newline to get away from prompt when we're reporting on a script/paste
+        if (locationPrefix != consoleLinePrefix) printMessage("")
+
+        printMessage(indentation + posIn.lineContent)
+        printMessage(indentation + posIn.lineCaret)
+        printMessage(s"$locationPrefix${posIn.line - preambleLineDelta}: $msgFirstLine")
```
retronym: Off by one error here?

```
scala> {
     |   "".rev
     | }
         "".rev
            ^
On line 3: error: value rev is not a member of String
```

retronym:
Here's a pastie-corner case that gives a negative position error. `:paste`, `{`, `^D`

```
scala> :paste

{

The pasted code is incomplete!


       {
       ^
<pastie>:-8: error: Missing closing brace `}' assumed here
```

retronym:
Good call, [adding `--keep-optional`](https://gist.github.com/retronym/6b5cea4d94d51fbcf5d7a0587404492e) does the trick.

src/repl/scala/tools/nsc/interpreter/Interface.scala
```diff
@@ -0,0 +1,283 @@
...
+  protected def cursor: Int
+  protected def buf: String
+
+  def cleanup(): Unit
+
+  def print: String
+
+  def typeAt(start: Int, end: Int): String
+
+  def candidates(tabCount: Int): (Int, List[String])
```
retronym, line 294: This is somewhat REPL specific, and it seems odd to already be String-ifiying the types at this level of API. I'm happy to proceed with the refactoring and refine later, though.

retronym:
I've got another round of cleanups for the Repl API and for power mode in https://github.com/retronym/scala/tree/review/5903

Nothing that needs to hold up this PR, though.

dwijnand:
> Could the issue be with coursier? "optional" means "non-transitive" in maven-speak, as far as I know. It's not optional in the sense that it needn't be on the classpath (that's "provided")

My understanding is that "provided" means provided by something other than dependency resolution (typically the runtime context, like an app server). My interpretation of "optional" _is_ "needn't be on the classpath".

lrytz:
> `<pastie>:-8: error: Missing closing brace '}' assumed here`

I cannot reproduce this, I get line number 2 (same as 2.12.2)

src/repl/scala/tools/nsc/interpreter/Interface.scala
```diff
@@ -0,0 +1,283 @@
...
+  def updateSettings(arguments: List[String]): Boolean
+
+  def initializeComplete: Boolean
+
+  // initializes the compiler, returning false if something went wrong
+  def initializeCompiler(): Boolean
+
+  def classPathString: String
+
+  def quietRun[T](code: String): Result
```
lrytz: is there a compatibility reason for the type param?

src/repl/scala/tools/nsc/interpreter/Interface.scala
```diff
@@ -0,0 +1,283 @@
...
+    *
+    */
+  def setContextClassLoader(): Unit
+}
+
+/**
+  * Interface to the repl for use by the frontend (shell, the UI).
+  *
+  * The interface should not depend on symbols and types,
+  * just simple trees and positions, so that we could run the shell
```
lrytz: Where are the trees? I couldn't spot any. (Maybe they are hidden in the forest.)

test/files/run/repl-trim-stack-trace.check
```diff
@@ -1,28 +1,28 @@
-Welcome to Scala
+Welcome to Scala %1$#s (%3$s, Java %2$s).
```
lrytz: could we just remove the `Welcoming` trait from repl-trim-stack-trace and t7805-repl-i, or are we testing something useful here?

src/repl-frontend/scala/tools/nsc/interpreter/jline/FileBackedHistory.scala
```diff
@@ -5,18 +5,50 @@
...
 trait FileBackedHistory extends JLineHistory with PersistentHistory {
   def maxSize: Int
+  import java.nio.file.StandardOpenOption.{CREATE, APPEND, TRUNCATE_EXISTING}
+
+  val charSet: Charset = implicitly[Codec].charSet
+
+  private lazy val historyPath = {
+    val fs = FileSystems.getDefault
+
+    // This would really have been sufficient for our property getting infrastructure
```
lrytz, line 30: let's put it somewhere in the standard library then :-)

src/repl-frontend/scala/tools/nsc/interpreter/shell/ILoop.scala
```diff
@@ -864,125 +887,121 @@ class ILoop(defaultIn: BufferedReader = null, protected val out: JPrintWriter =
...
+          interpreterInitialized.countDown()
+
+          if (intp.reporter.hasErrors) {
+            echo("Interpreter encountered errors during initialization!")
+            throw new InterruptedException
+          }
+
+          // scala/bug#7418 Now that the interpreter is initialized, and `interpretPreamble` has populated the symbol table,
+          // enable TAB completion (we do this before blocking on input from the splash loop,
+          // so that it can offer tab completion as soon as we're ready).
```
lrytz: I don't understand what exactly changed, in 2.12.2 tab-completion works on the first line when waiting for a bit (until the compiler is initialized).

lrytz (commented):
This looks all good, just a few minor comments in-line.

One commit message says: `TODO: bring back color output` - what's the plan there?

A small detail: in order to exit the repl on the first line, I have to press `^D` twice now, that's a regression.

```
$ scala-pr 5903
Welcome to Scala 2.13.0-20170525-235632-87f41ea (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).

scala>
scala> :quit
```

adriaanm:
> TODO: bring back color output - what's the plan there?

https://github.com/adriaanm/scala/commit/c8d475e (tokenizer borrowed from dotty)
(Saving this for the next PR)

src/repl/scala/tools/nsc/interpreter/Interface.scala
adriaanm: Haha, they are potential trees. Not yet materialized.

src/repl/scala/tools/nsc/interpreter/Interface.scala
adriaanm: nope, dropped

src/repl-frontend/scala/tools/nsc/interpreter/jline/FileBackedHistory.scala
adriaanm: added a TODO to keep this PR nice and svelte 😏

src/repl/scala/tools/nsc/interpreter/Interface.scala
adriaanm: The goal for the first version was to be as minimal and decoupled from the compiler as possible. We could provide a richer interface using Trees, but I don't think the UI should have access to the whole symbol table. Ultimately, I'd like to have multiple shells that can use any 2.13.x compiler.

src/repl-frontend/scala/tools/nsc/interpreter/shell/Reporter.scala
adriaanm: The error message you're seeing does look wrong, but I think it's correct to subtract the line number of the last line of the preamble from the line number of the error to get the line number as the user sees it.

adriaanm:
> in order to exit the repl on the first line, I have to press ^D twice now, that's a regression.

Fix coming

adriaanm:
>> <pastie>:-8: error: Missing closing brace '}' assumed here

> I cannot reproduce this, I get line number 2 (same as 2.12.2)

Also can't repro.

src/repl-frontend/scala/tools/nsc/interpreter/jline/FileBackedHistory.scala
lrytz: sure, i didn't mean to do it here :)

adriaanm:
Thanks for the feedback! I think this should be ready to go now.

eed3si9n:
In my Spec2 [CompilerMatcher](https://github.com/eed3si9n/scalaxb/blob/e346c7baf1de9dacff0961390672487c2754b290/integration/src/test/scala/CompilerMatcher.scala#L105-L107), I do something like:

```scala
      val holder0 = allCatch opt {
        main.lastReq.lineRep.call("$result")
      }
```

to grab the last result and match against some expected value. Do I now use `mostRecentVar`?

retronym:
I'm still seeing the off-by-one error.

```
⚡ scala-pr 5903
Welcome to Scala 2.13.0-20170613-182959-5a6acca (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> {
     |   "".rev
     | }
         "".rev
            ^
On line 3: error: value rev is not a member of String
```

The generated code is:

```


object $read {
object $iw {
object $iw {

private type $ipos3 = _root_.scala.Nothing;
val res0 =
{
"".rev
}

}
}

}
```

The extra line taken by the `val res0 =\n` seems not to be accounted for.

retronym:
I think the principled way out this quagmire is to at parse the input as given (hey presto, we'll have the right positions!), and then add the synthetic wrappers (wrapping templates, imports, and result capturing valdefs) with a tree transformation, rather than the textual mess in `requestFromLine` and `Wrapper`.

retronym:
Here's another related bug in error positions:

```

scala> "".rev
          ^
       error: value rev is not a member of String

scala> println("blerg"); "".rev
           ^
       error: value rev is not a member of String
```

adriaanm:
I'm at the limit of what I can manage to do with the current encoding. I'm very much in favor of taking a more direct approach, to remove any wrapping during type checking, and wrapping trees for execution later in the pipeline. 

For now, all I can think of is to turn off the optimization where I try to point at the code that's already on the console.

adriaanm:
> Do I now use mostRecentVar?

@eed3si9n, would it make sense to add an `eval` method to the `Repl` interface that returns the value of the expression?

retronym:
> For now, all I can think of is to turn off the optimization where I try to point at the code that's already on the console.

👍, I'm happy to merge after that.

eed3si9n:
> would it make sense to add an eval method to the Repl interface that returns the value of the expression?

@adriaanm Yes. I think retrieving the last value is a common use case for REPL embedding.

src/repl/scala/tools/nsc/interpreter/Interface.scala
```diff
@@ -55,9 +55,13 @@ trait ReplCore {
...
-  * The interface should not depend on symbols and types,
-  * just simple trees and positions, so that we could run the shell
-  * in a separate process from the compiler, for example.
+  * The interface should not depend on symbols and types (the compiler's internal state).
+  * At most, expose untyped trees and positions in addition to standard Java types.
+  * This decoupling would allow running the shell in a separate thread, or even
+  * in a separate process from the compiler. It should also be possible to write
+  * a new REPL frontend using this interface, and be compatible across minor compiler releases.
+  *
+  * (The first iteration of this interface is restricted to Strings, but we could loosen that.)
```
lrytz: Nit: it does use Positions currently (`def lastWarnings: List[(Position, String)]`, `trait ReplRequest { def preambleEndPos: Position }`

lrytz (approved):
:shipit:

adriaanm:
I like the suggestion of parsing user-input as-is, @retronym! WIP: https://github.com/adriaanm/scala/compare/repl-modularize...repl-modularize-positions

adriaanm:
What a slog -- TODO:
  - [x] rangepos fun: run/t6381.scala
  - [x] xml syntax (jvm/xml05.scala, jvm/interpreter.scala)

adriaanm:
@retronym, I plan to squash 460a985..7e5b06b, but will leave them separate to simplify reviewing.

adriaanm:
i could lift the syntax highlighting one (5346987) out of that range if you prefer

test/files/jvm/interpreter.check
```diff
@@ -308,8 +308,9 @@ scala> /*
   */
 */
 
+scala> 
 
-You typed two blank lines.  Starting a new command.
```
adriaanm, line 315: there was no incomplete input, so this message should not have triggered

adriaanm:
Rebased, and incorporated magic imports.

src/repl/scala/tools/nsc/interpreter/Imports.scala
```diff
@@ -1,8 +1,7 @@
 /* NSC -- new Scala compiler
- * Copyright 2005-2013 LAMP/EPFL
- * @author  Paul Phillips
+ * Copyright 2005-2016 LAMP/EPFL
```
dwijnand: 2017?

dwijnand:
Real minor stuff:

* there's a random added yet empty src/repl-frontend/scala/tools/nsc/InterpreterLoop.scala file
* src/repl/scala/tools/nsc/interpreter/PresentationCompilerCompleter.scala is now an empty file, without being deleted, bizarrely

test/files/run/t7747-repl.check
```diff
@@ -236,45 +236,39 @@ package $line10 {
...
-          super.<init>;
-          ()
-        };
-        val res3 = List(BippyBups, PuppyPups, Bingo)
-      };
-      val $iw = new $iw.<init>
-    };
-    val $iw = new $iw.<init>
+    import _root_.scala.tools.nsc.interpreter.$u007B$u007B;
+    import _root_.scala.tools.nsc.interpreter.$u007B$u007B;
```
retronym: Not sure what leads to this, but we could collapse sequences of these into a single import.

test/files/run/t7747-repl.check
```diff
@@ -236,45 +236,39 @@ package $line10 {
       super.<init>;
       ()
     };
-    sealed class $iw extends _root_.java.io.Serializable {
```
retronym: Begone, ye wretched wrapper!

adriaanm:
Thanks, will clean up those files.

Also, one (hopefully last test fail):
  - test/files/neg/warn-unused-imports

retronym:
The ANSIfication of results seems a bit heavy handed now.

image

adriaanm:
Alright, this should be ready for merge, if it looks good. I'm going to defer the work on simplifying wrapping to another PR, where we'll also standardize on the class-based wrapping, and just generally try to dial the wrapping noise back to a minimum. That, and upgrading to JLine 3 will have to wait for M3.

retronym (approved):

felixmulder:
@retronym, @adriaanm - hey guyz, Dotty has a `Show` type class in the library. It's used to render rhs in a friendly way, e.g:

```scala
scala> "\twaaaaaaaaaaaaaaaaaaaaaat"
res0: String =     waaaaaaaaaaaaaaaaaaaaaat
```
in scalac, but in dotty:

```scala
scala> "\twaaaaaaaaaaaaaaaaaaaaaat"
val res0: String = "\twaaaaaaaaaaaaaaaaaaaaaat"
```
As soon as case classes are hlists, there could be show instances for them as well by default. It requires a bit more fiddling, but it looks nice and is copy-pastable. The original idea (of course) came from ghci and the ocaml repl.

adriaanm:
> Dotty has a `Show` type class in the library. It's used to render rhs in a friendly way

Cool, thanks for the pointer! Will be nice to have this as part of the std lib. (Maybe even in 2.13? /cc @szeiger)

adriaanm:
🎈

som-snytt:
Just back from vacation and look fwd to trying it out. Is there an initiative to provide a uniform API for Scala 2 and 3/dotty? I saw a comment about a dotr rewrite, but I didn't see if unification was a goal. Choices like jline or not would be nice to decouple from folks writing REPL support. Can I use `Amirite` I mean `Ammonite` with Scala 2 or 3-alpha in a pluggable way?

adriaanm:
Welcome back! I'd very much like for us to converge on a compiler-backed API that can be provided uniformly across versions, as well as consumed by multiple frontends.

adriaanm:
Consider this the proto-strawman proposal.

src/repl-frontend/scala/tools/nsc/Interpreter.scala
```diff
@@ -0,0 +1,122 @@
+package scala.tools.nsc
+
+import scala.tools.nsc.interpreter.{IMain, Repl, ReplCore}
+import scala.tools.nsc.interpreter.shell.{ILoop, ReplReporterImpl, ShellConfig}
+import scala.tools.nsc.reporters.Reporter
+
+// Pretty gross contortion to satisfy the de facto interface expected by sbt.
```
adriaanm, line 7: /cc @smarter: this is the grossness we use to integrate with sbt ;-)

PR #5908, opened 2017-05-16
Route errors and warnings to the ContextReporter.

dragos:
The context reporter is where error reporting policy is defined, so type diagnostics should use it as well.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
@@ -638,23 +638,23 @@ trait TypeDiagnostics {
               else if (sym.isModule) s"object ${sym.name.decoded}"
               else "term"
             )
-            reporter.warning(pos, s"$why $what in ${sym.owner} is never used")
+            context.warning(pos, s"$why $what in ${sym.owner} is never used")
           }
           for (v <- p.unsetVars) {
-            reporter.warning(v.pos, s"local var ${v.name} in ${v.owner} is never set: consider using immutable val")
+            context.warning(v.pos, s"local var ${v.name} in ${v.owner} is never set: consider using immutable val")
           }
           for (t <- p.unusedTypes) {
             val sym = t.symbol
             val wrn = if (sym.isPrivate) settings.warnUnusedPrivates else settings.warnUnusedLocals
             if (wrn) {
               val why = if (sym.isPrivate) "private" else "local"
-              reporter.warning(t.pos, s"$why ${sym.fullLocationString} is never used")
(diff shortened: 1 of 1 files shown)


src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -698,11 +698,8 @@ trait TypeDiagnostics {
         } else f
       }
       def apply(tree: Tree): Tree = {
-        // Error suppression (in context.warning) would squash some of these warnings.
-        // It is presumed if you are using a -Y option you would really like to hear
-        // the warnings you've requested; thus, use reporter.warning.
         if (settings.warnDeadCode && context.unit.exists && treeOK(tree) && exprOK)
-          reporter.warning(tree.pos, "dead code following this construct")
+          context.warning(tree.pos, "dead code following this construct")
```
dragos, line 702: This change reverts a conscious choice made by @adriaanm to circumvent the context reporter. In an offline discussion we agreed to revisit, so here it is.

IMO it's better to go via the Context for two reasons:
- if the user wanted to silence a warning I think that should take precedence over a general/coarse-grained command line option
- consistency. it is/will be simpler to enforce reporting policies if all code goes through the same gate

adriaanm (approved):

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
retronym: One practical concern is that this will probably silence the dead code warning if the preceding statement is, e.g. a call to a deprecated method.

som-snytt:
I thought context expressed something different; the unused check runs more as a post-processing step.

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
dragos: ~I am not sure I fully understand this, but I *think* it shouldn't be a problem, based on the fact that this check is indeed run after typer, as a post-processing step, and the context reporter is the one from the root context.~

I couldn't reproduce this scenario. The following code shows both dead-code warnings:

```Scala
class Foo {
  @deprecated
  def show(x: Int): Nothing = ???

  show(???)
  println("d")
}

$ qscalac -d /tmp -Ywarn-dead-code private.scala
private.scala:5: warning: dead code following this construct
  show(???)
       ^
private.scala:5: warning: dead code following this construct
  show(???)
      ^
warning: there was one deprecation warning
warning: there was one deprecation warning (since 2.11.0)
warning: there were two deprecation warnings in total; re-run with -deprecation for details
5 warnings found
```

PR #5911, opened 2017-05-17
Avoid generic array instantation in hot paths

retronym:
Before:

```
[info] 
[info] HotScalacBenchmark.compile                                         sample  500  1049.371 ± 3.062  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00                           sample       1000.342          ms/op
[info] HotScalacBenchmark.compile:compile·p0.50                           sample       1047.527          ms/op
[info] HotScalacBenchmark.compile:compile·p0.90                           sample       1075.839          ms/op
[info] HotScalacBenchmark.compile:compile·p0.95                           sample       1082.130          ms/op
[info] HotScalacBenchmark.compile:compile·p0.99                           sample       1113.462          ms/op
[info] HotScalacBenchmark.compile:compile·p0.999                          sample       1176.502          ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999                         sample       1176.502          ms/op
[info] HotScalacBenchmark.compile:compile·p1.00                           sample       1176.502          ms/op
```

After:

```
[info] HotScalacBenchmark.compile                                         sample  500  1045.162 ± 2.888  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00                           sample        984.613          ms/op
[info] HotScalacBenchmark.compile:compile·p0.50                           sample       1046.479          ms/op
[info] HotScalacBenchmark.compile:compile·p0.90                           sample       1068.499          ms/op
[info] HotScalacBenchmark.compile:compile·p0.95                           sample       1075.839          ms/op
[info] HotScalacBenchmark.compile:compile·p0.99                           sample       1090.498          ms/op
[info] HotScalacBenchmark.compile:compile·p0.999                          sample       1117.782          ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999                         sample       1117.782          ms/op
[info] HotScalacBenchmark.compile:compile·p1.00                           sample       1117.782          ms/op
```

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -1349,9 +1349,20 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
           lambdaTarget.name.toString,
           methodBTypeFromSymbol(lambdaTarget).descriptor,
           /* itf = */ isInterface)
-      val receiver = if (isStaticMethod) Nil else lambdaTarget.owner :: Nil
-      val (capturedParams, lambdaParams) = lambdaTarget.paramss.head.splitAt(lambdaTarget.paramss.head.length - arity)
-      val invokedType = asm.Type.getMethodDescriptor(asmType(functionalInterface), (receiver ::: capturedParams).map(sym => typeToBType(sym.info).toASMType): _*)
+      val numCapturedParams = lambdaTarget.paramss.head.length - arity
+      val (capturedParams, lambdaParams) = lambdaTarget.paramss.head.splitAt(numCapturedParams)
+
+      val invokedTypeArgs: Array[asm.Type] = if (isStaticMethod) {
+        val result = new Array[asm.Type](numCapturedParams)
+        capturedParams.iterator.map(sym => typeToBType(sym.info).toASMType).copyToArray(result)
+        result
+      } else {
+        val result = new Array[asm.Type](numCapturedParams + 1)
+        result(0) = typeToBType(lambdaTarget.owner.info).toASMType
+        capturedParams.iterator.map(sym => typeToBType(sym.info).toASMType).copyToArray(result, 1)
+        result
+      }
+      val invokedType = asm.Type.getMethodDescriptor(asmType(functionalInterface), invokedTypeArgs: _*)
       val constrainedType = new MethodBType(lambdaParams.map(p => typeToBType(p.tpe)), typeToBType(lambdaTarget.tpe.resultType)).toASMType
       val samMethodType = methodBTypeFromSymbol(sam).toASMType
       val markers = if (addScalaSerializableMarker) classBTypeFromSymbol(definitions.SerializableClass).toASMType :: Nil else Nil
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -501,7 +501,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
       currentRun.symData get sym match {
         case Some(pickle) if !nme.isModuleName(newTermName(jclassName)) =>
           val scalaAnnot = {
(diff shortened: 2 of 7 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1349,9 +1349,20 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
+      val (capturedParams, lambdaParams) = lambdaTarget.paramss.head.splitAt(numCapturedParams)
+
+      val invokedTypeArgs: Array[asm.Type] = if (isStaticMethod) {
+        val result = new Array[asm.Type](numCapturedParams)
+        capturedParams.iterator.map(sym => typeToBType(sym.info).toASMType).copyToArray(result)
+        result
+      } else {
+        val result = new Array[asm.Type](numCapturedParams)
+        result(0) = typeToBType(lambdaTarget.owner.info).toASMType
+        capturedParams.iterator.map(sym => typeToBType(sym.info).toASMType).copyToArray(result, 1)
```
rorygraves: There is a bug here.  This loses the last parameter. The array should be ```numCapturedParams+1```  - copyToArray will simply throw away the arguments that do not fix.

rorygraves: Is ```numCapturedParams``` ever 0?  If so this could be optimised further.

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1349,9 +1349,20 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
           lambdaTarget.name.toString,
           methodBTypeFromSymbol(lambdaTarget).descriptor,
           /* itf = */ isInterface)
-      val receiver = if (isStaticMethod) Nil else lambdaTarget.owner :: Nil
-      val (capturedParams, lambdaParams) = lambdaTarget.paramss.head.splitAt(lambdaTarget.paramss.head.length - arity)
-      val invokedType = asm.Type.getMethodDescriptor(asmType(functionalInterface), (receiver ::: capturedParams).map(sym => typeToBType(sym.info).toASMType): _*)
+      val numCapturedParams = lambdaTarget.paramss.head.length - arity
+      val (capturedParams, lambdaParams) = lambdaTarget.paramss.head.splitAt(numCapturedParams)
```
mkeskells, line 1353: there is a special case where arity == 0 which avoid the splitAt

List.splitAt could be optimized where split == 0 and in for Nil - seperate PR probably

src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
```diff
@@ -501,7 +501,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
       currentRun.symData get sym match {
         case Some(pickle) if !nme.isModuleName(newTermName(jclassName)) =>
           val scalaAnnot = {
-            val sigBytes = ScalaSigBytes(pickle.bytes.take(pickle.writeIndex))
+            val sigBytes = ScalaSigBytes(java.util.Arrays.copyOf(pickle.bytes, pickle.writeIndex))
```
mkeskells, line 504: why this change - isnt this what .take does as we fixed the Array methods

retronym: Good point. We had to revert that change in #5821 because it wasn't binary compatible. We should re-apply the original patch to the 2.13.x branch.

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
retronym: Good catch. Turns out the second branch `(!isStatic`) is never taken, as we chose to make all lambda implementation methods static, which meant we could leave them public without having to mangle the names. But I've pushed a fix in case this becomes live code in the future.

retronym:
I"m going to close this PR for now as the benchmark isn't conclusive. I re-ran it and got 1054ms vs 1056ms.

PR #5914, opened 2017-05-19
Remove :kind

dwijnand:
Fixes scala/bug#8529


som-snytt:
This was rather unkind.

PR #5915, opened 2017-05-20
Add "Xlint:implicit-no-annotation"

edmundnoble:
Wondering if this needs a more self-explanatory name. Warns on implicits without type annotations.

Edit: Also please let me know if I am doing this in the wrong phase; I went as early as possible just because this info is available right away.

### DIFF
--- a/project/ScalaOptionParser.scala
+++ b/project/ScalaOptionParser.scala
@@ -108,7 +108,7 @@ object ScalaOptionParser {
     "-g" -> List("line", "none", "notailcails", "source", "vars"),
     "-target" -> List("jvm-1.5", "jvm-1.6", "jvm-1.7", "jvm-1.8"))
   private def multiChoiceSettingNames = Map[String, List[String]](
-    "-Xlint" -> List("adapted-args", "nullary-unit", "inaccessible", "nullary-override", "infer-any", "missing-interpolator", "doc-detached", "private-shadow", "type-parameter-shadow", "poly-implicit-overload", "option-implicit", "delayedinit-select", "by-name-right-associative", "package-object-classes", "unsound-match", "stars-align"),
+    "-Xlint" -> List("adapted-args", "nullary-unit", "inaccessible", "nullary-override", "infer-any", "missing-interpolator", "doc-detached", "private-shadow", "type-parameter-shadow", "poly-implicit-overload", "option-implicit", "delayedinit-select", "by-name-right-associative", "package-object-classes", "unsound-match", "stars-align", "no-implicit-annotation"),
     "-language" -> List("help", "_", "dynamics", "postfixOps", "reflectiveCalls", "implicitConversions", "higherKinds", "existentials", "experimental.macros"),
     "-opt" -> List("l:none", "l:default", "l:method", "l:project", "l:classpath", "unreachable-code", "simplify-jumps", "empty-line-numbers", "empty-labels", "compact-locals", "nullness-tracking", "closure-elimination", "inline-project", "inline-global"),
     "-Ystatistics" -> List("parser", "typer", "patmat", "erasure", "cleanup", "jvm")
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -718,6 +718,9 @@ trait Namers extends MethodSynthesis {
       if (isScala) {
         if (nme.isSetterName(tree.name)) ValOrVarWithSetterSuffixError(tree)
         if (tree.mods.isPrivateLocal && tree.mods.isCaseAccessor) PrivateThisCaseClassParameterError(tree)
+        if (tree.mods.isImplicit && !owner.isTerm && settings.warnImplicitNoAnnotation && tree.tpt.isEmpty) {
+          reporter.warning(tree.pos, s"implicit value ${tree.name} has no type annotation")
+        }
       }
 
       if (isScala && deriveAccessors(tree)) enterGetterSetter(tree)
@@ -748,6 +751,10 @@ trait Namers extends MethodSynthesis {
       case DefDef(mods, name, _, _, _, _) =>
         val bridgeFlag = if (mods hasAnnotationNamed tpnme.bridgeAnnot) BRIDGE | ARTIFACT else 0
         val sym = enterInScope(assignMemberSymbol(tree)) setFlag bridgeFlag
+
+        if (tree.symbol.isImplicit && settings.warnImplicitNoAnnotation && tree.tpt.isEmpty) {
+          reporter.warning(tree.pos, s"implicit definition ${tree.name} has no type annotation")
+        }
 
         val completer =
           if (sym hasFlag SYNTHETIC) {
--- a/src/compiler/scala/tools/nsc/settings/Warnings.scala
+++ b/src/compiler/scala/tools/nsc/settings/Warnings.scala
@@ -101,6 +101,7 @@ trait Warnings {
     val PackageObjectClasses   = LintWarning("package-object-classes",    "Class or object defined in package object.")
     val UnsoundMatch           = LintWarning("unsound-match",             "Pattern match may not be typesafe.")
     val StarsAlign             = LintWarning("stars-align",               "Pattern sequence wildcard must align with sequence component.")
+    val ImplicitNoAnnotation   = LintWarning("implicit-no-annotation",    "Implicit definitions and fields must have type annotations.")
     val Constant               = LintWarning("constant",                  "Evaluation of a constant arithmetic expression results in an error.")
     val Unused                 = LintWarning("unused",                    "Enable -Ywarn-unused:imports,privates,locals,implicits.")
 
@@ -124,6 +125,7 @@ trait Warnings {
   def warnPackageObjectClasses   = lint contains PackageObjectClasses
   def warnUnsoundMatch           = lint contains UnsoundMatch
   def warnStarsAlign             = lint contains StarsAlign
+  def warnImplicitNoAnnotation   = lint contains ImplicitNoAnnotation
   def warnConstant               = lint contains Constant
   def lintUnused                 = lint contains Unused
 
--- a/test/files/neg/implicit-no-annotation.check
+++ b/test/files/neg/implicit-no-annotation.check
@@ -1,0 +1,12 @@
+implicit-no-annotation.scala:2: warning: implicit definition int has no type annotation
+  implicit def int = 42
+               ^
+implicit-no-annotation.scala:3: warning: implicit value string has no type annotation
+  implicit val string = "hello"
+               ^
+implicit-no-annotation.scala:12: warning: implicit definition localString has no type annotation
+    implicit def localString = "hello"
+                 ^
+error: No warnings can be incurred under -Xfatal-warnings.
+three warnings found
+one error found
--- a/test/files/neg/implicit-no-annotation.flags
+++ b/test/files/neg/implicit-no-annotation.flags
@@ -1,0 +1,1 @@
+-Xlint:implicit-no-annotation -Xfatal-warnings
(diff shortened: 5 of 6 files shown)


sjrd:
An implicit `val` in a *local* scope (e.g., within a `def`) should be allowed not to have a type annotation. Only field `val`s require a type annotation.

However, all `def`s require a type annotation.

edmundnoble:
Why exactly? I am working off of the comment in `isValid` in Implicits.scala, which seems to state that implicits without type annotations will not be considered for implicit resolution unless they occur before the search location in the source.

sjrd:
Because https://github.com/lampepfl/dotty/issues/1784 and https://github.com/lampepfl/dotty/pull/1785.

Local implicit vals are only visible in the scope that follows them, so it is safe to let their type be inferred. Moreover, it is extremely annoying to write explicit types for local implicit vals. For example the Scala.js codebase is full of
```scala
def someTransformation(tree: Tree): Tree = {
  implicit val pos = tree.pos
  ...
}
```

edmundnoble:
Makes sense to me. I'll try to make the change.

edmundnoble:
Change made and tested.

sjrd:
Thanks. I guess someone from the core team has to take over the technical review at this point.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -710,6 +710,9 @@ trait Namers extends MethodSynthesis {
       if (isScala) {
         if (nme.isSetterName(tree.name)) ValOrVarWithSetterSuffixError(tree)
         if (tree.mods.isPrivateLocal && tree.mods.isCaseAccessor) PrivateThisCaseClassParameterError(tree)
+        if (tree.mods.isImplicit && !owner.isTerm && settings.warnImplicitNoAnnotation && tree.tpt.isEmpty) {
+          reporter.warning(tree.pos, s"implicit value ${tree.name} has no type annotation")
```
retronym, line 722: Should be `.name.decoded` to avoid displaying encoded names like '$colon$colon`.

retronym: The term "type annotation" is widely understood but the spec talks about "result types" in this context (and about "type ascriptions" after expressions).

A precedent for wording of this sort of message is:

```
scala> def foo = { return 1; 1 }
<console>:11: error: method foo has return statement; needs result type
       def foo = { return 1; 1 }
                   ^
```

`s/needs/should have a` and I think it fits here.

retronym:
Here's the corresponding check in dotty: https://github.com/lampepfl/dotty/blob/d524e28ebc4a5dc445a11df2548d7c9ea586b745/compiler/src/dotty/tools/dotc/typer/Namer.scala#L1062

project/ScalaOptionParser.scala
```diff
@@ -108,7 +108,7 @@ object ScalaOptionParser {
     "-g" -> List("line", "none", "notailcails", "source", "vars"),
     "-target" -> List("jvm-1.5", "jvm-1.6", "jvm-1.7", "jvm-1.8"))
   private def multiChoiceSettingNames = Map[String, List[String]](
-    "-Xlint" -> List("adapted-args", "nullary-unit", "inaccessible", "nullary-override", "infer-any", "missing-interpolator", "doc-detached", "private-shadow", "type-parameter-shadow", "poly-implicit-overload", "option-implicit", "delayedinit-select", "by-name-right-associative", "package-object-classes", "unsound-match", "stars-align"),
+    "-Xlint" -> List("adapted-args", "nullary-unit", "inaccessible", "nullary-override", "infer-any", "missing-interpolator", "doc-detached", "private-shadow", "type-parameter-shadow", "poly-implicit-overload", "option-implicit", "delayedinit-select", "by-name-right-associative", "package-object-classes", "unsound-match", "stars-align", "no-implicit-annotation"),
```
retronym, line 111: Perhaps: `implicit-result-type`

src/compiler/scala/tools/nsc/typechecker/Namers.scala
retronym: (I guess this is to avoid overloading the term "annotation" with `@annotations`.)

retronym (commented):
It is also worth looking at the history of the corresponding rule in the `wartremover` lint tool: https://github.com/wartremover/wartremover/commits/master/core/src/main/scala/wartremover/warts/ExplicitImplicitTypes.scala

It evolved to allow `{ impicit x => ... }` lambda params without an result type on the `ValDef` representing the lamdba param `x`; added an some sort of exception for implicit classes.

project/ScalaOptionParser.scala
retronym: Without the rename, there is a typo here:

`s/no-implicit-annotation/implicit-no-annotation`

(The fact that you need to repeat yourself here is my fault, sorry...)

retronym:
When adding a new warning, it's a can be helpful to run it across a few projects and review the output. That often turns up false positives. You can also post a gist of the output to help us review.

Here's how that looks for "re-starring" (using your modified compiler as the new reference compiler):

```
% sbt setupPublishCore publishLocal | tail -n 20

% git diff -U1
diff --git a/build.sbt b/build.sbt
index 8557e1280d..1e70b7d9ca 100644
--- a/build.sbt
+++ b/build.sbt
@@ -145,2 +145,3 @@ lazy val commonSettings = clearSourceAndResourceDirectories ++ publishSettings +
   fork in run := true,
+  scalacOptions in Compile ++= Seq("-Xlint:implicit-no-annotation"),
   scalacOptions in Compile in doc ++= Seq(

% sbt -Dstarr.version=2.12.3-bin-f092e93-SNAPSHOT test:compile

```
https://gist.github.com/53444a774c7675d7f115bd58b65ab498

project/ScalaOptionParser.scala
edmundnoble: I'm curious; so my motivation was that it seems to be that all of these lint options are named for what they warn for. Nullary-unit, inaccessible, adapted args, these are all the "bad" that the lint option actually warns for itself. "implicit-result-type" seems inconsistent, because it would be named for the condition needed to *not* warn.

Jasper-M:
Does this also warn for implicit objects? Because they have the same problem.
Dotty doesn't seem to have a problem with implicit objects though, so maybe they can be fixed.

adriaanm:
@edmundnoble, what's your plan regarding getting this across the finish line in time for 2.12.4? This would be a nice lint option and it looks like there isn't much more polish needed. To be included in 2.12.4, PRs must be mergeable (approved by reviewer) by Wed Sep 27.

project/ScalaOptionParser.scala
edmundnoble: I thought about it and I agree the rename is the most consistent option.

edmundnoble:
@retronym I've rebased, and added tests for the special cases you've mentioned: they didn't require any more changes to the warning check. I also changed "type annotation" to "result type" in the lint option and the warning itself.

I'm considering maybe relaxing the `implicit def` condition to allow no result type if the RHS is just a `new` call, but I think this is worth merging into 2.12.4 as it is now.
EDIT: I'm considering this exception based mostly on implicit conversions into `AnyVal` classes which usually take this form as well as the cases I encountered after re-starring.

project/ScalaOptionParser.scala
Ichoran: This still looks unfixed?

adriaanm:

adriaanm:
We ran out of time here for 2.12.4, but let's get this into a nightly as soon as all review comments are implemented.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
som-snytt: Not that further motivation is needed, but I expected it to warn about lacking an implicit not found annotation.

project/ScalaOptionParser.scala
som-snytt: I can attest that naming these options is hard. I suspect no one will ever request `-Xlint:poly-implicit-overload`. You warn for `implicit-lacks-type`, or `-Xlint:infer-X`, since the warning is triggered by inference. `-Xlint:infer-implicit-type`.

martin-g:
Ping! Is it a good time to merge this PR for 2.12.5 ?

SethTisue:
should I do a community build run so we can grep the log for the new warning and see what turns up?

lrytz:
@edmundnoble as far as I can see there are still a few minor changes to be done:
  - rename the option from `no-implicit-annotation` to `implicit-result-type` across the PR (including test file names)
  - use `tree.name.decoded` instead of `tree.name` in the warning messages, thest for such names
  - update the warning messages from `has no type annotation` to `should have a result type`

@SethTisue the logic looks correct, so you can run it through the community build already now.

SethTisue:
@edmundnoble note that due to recent Jenkins changes, you'll need to rebase, too.

SethTisue:
I'm sorry this PR sat so long — we are really trying to do better on this in 2018

gah, and actually I can't run (not easily, anyway) the community build until this is rebased, the validate-publish-core job was deleted. @edmundnoble would you mind rebasing onto current 2.12.x...?

lrytz:
@edmundnoble / @SethTisue I rebased and pushed -f to `edmundnoble/implicit-no-annotation` (there were no conflicts)

SethTisue:
community build run queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2635/ ~~(404 til Jenkins finishes shaving his toes)~~

lrytz:
That didn't go very far:
```
ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
```

SethTisue:
doh. trying again: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2643/consoleFull

SethTisue:
gah, that timed out. queued run 2645. builds from 2643 should still be cached, so when it comes time to grep the logs, we'll need to grep both of them together.

SethTisue:
(note that these runs don't explicitly enable the new flag everywhere, so we're only seeing the warnings for projects that have `-Xlint` enabled. this is probably sufficient to evaluate the quality.)

lrytz:
The build timed out; there's 125 lines with "has no type annotation" in the log, checking if those are all intended would give a good estimate.

SethTisue:
in run 2645 which is a continuation of 2643, there are a bunch of new compiler errors like:

```
[kind-projector] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.11/project-builds/kind-projector-cad88983fe673ab2ad72b4a5f1d7ca5de5e38d0e/src/main/scala/KindProjector.scala:93:14: not enough arguments for method require: (requirement: Boolean, message: => Any)Unit.
[kind-projector] [error] Unspecified value parameter message.
[kind-projector] [error]       require(i >= 0)
[kind-projector] [error]              ^
```

SethTisue:
this needs another rebase, now, in order to get better community build results

som-snytt:
I can't tell without a map if the territory entered by this PR is Kafkaesque, Communist state, or Monty Python. There's a special community badge for "PR merged one year after Adriaan says it must be merged by Friday to make [the next point release]." It would be funnier, though, if it were delayed by naming the option. That would be Monty Python.

adriaanm:
Closing due to inactivity. Happy to accept a resubmission, though that should go to 2.13.x now.

som-snytt:
Maybe the community build should be peer-to-peer, with "volunteer" computers all over the world verifying every scala project for every commit.

adriaanm:
We promise we won't mine bitcoin in macros.

som-snytt:
This noble effort landed not as a lint but under `-Xsource:3`.

PR #2340, opened 2013-03-31
Added a :kind command to the REPL

folone:
### Description

Because Scala supports working with higher kinded types, we might want to be able to inspects kinds as well as types. This pull request adds a simple `:kind` command, implemented by @eed3si9n in his ["Introduction to Scalaz: day 4"](http://eed3si9n.com/learning-scalaz-day3) post, to the Scala REPL. Here is an example output:

``` scala
scala> :kind scala.Option
Option's kind is * -> *

scala> :k scalaz.Unapply
Unapply's kind is ((* -> *) -> *) -> * -> *

scala> import scalaz._
import scalaz._

scala> :k Monad // Finds locally imported types.
Monad's kind is (* -> *) -> *
This is a type constructor that takes type constructor(s): a higher-kinded type.

scala> :k Nonexisting
<console>:14: error: not found: value Nonexisting
              Nonexisting
              ^

scala> class Foo
defined class Foo

scala> new Foo { def empty = true }
res0: Foo{def empty: Boolean} = $anon$1@786aceba

scala> :k res0
Foo{def empty: Boolean}'s kind is *
```
### Future improvements

As pointed out by @retronym [here](https://twitter.com/retronym/status/318343952942125057), variance and type bounds are also part of the kind and we might want to include them in command's output.
### Acknowledgements

The function for `:kind` feature is originally implemented by @eed3si9n 
I got a lot of help from @xeno-by and @retronym while working on this feature, thank you for that.

Happy Easter everyone!

PR #5916, opened 2017-05-20
REPL: Improve :kind by only accepting type expressions, Fixes SI-8529: 

eed3si9n:
Fixes scala/bug#8529

The :kind implementation added in scala/scala#2340 was buggy and misleading in part because we could not find a robust way to turn a type expression string into a `Type`. See also [Getting type information inside scala repl via IMain](http://stackoverflow.com/questions/15678616/getting-type-information-inside-scala-repl-via-imain/15694003#15694003). The problem was that we didn't know how to pick up types defined or imported in the REPL, so #2340 faked it by treating the given type expression as a term.

That's the major source of problems reported in scala/bug#8529. This PR improves the String-to-type-symbol conversion by using `exprTyper.typeOfTypeString` repeatedly with `[Nothing]`. It's hacky, but a whole class of "expression" can now be called out as not being a type.

### rejection of terms

Before:

```scala
scala> :kind 5
scala.Int's kind is A
```

After:

```scala
scala> :kind 5
<console>: error: type 5 was not found
```

### fixes companion object mixup

Before:

```scala
scala> :kind -v Predef.Pair
scala.Predef.Pair's kind is A
*
This is a proper type.
```

After:

```scala
scala> :kind -v Predef.Pair
Pair's kind is F[+A1,+A2]
* -(+)-> * -(+)-> *
This is a type constructor: a 1st-order-kinded type.
```

### fixes type application not being treated as a proper type

Before:

```scala
scala> :kind -v Predef.Pair[Int, Int]
scala.Tuple2's kind is F[+A1,+A2]
* -(+)-> * -(+)-> *
This is a type constructor: a 1st-order-kinded type.
```

After:

```scala
scala> :kind -v Predef.Pair[Int, Int]
Pair[Int,Int]'s kind is A
*
This is a proper type.
```

### fixes partially applied type alias

Before:

```scala
scala> type IntTuple[+A] = (Int, A)
defined type alias IntTuple

scala> :kind IntTuple
<console>:12: error: not found: value IntTuple
       IntTuple
       ^
```

After:

```scala
scala> type IntTuple[+A] = (Int, A)
defined type alias IntTuple

scala> :kind IntTuple
IntTuple's kind is F[+A]
```

### fixes type lambdas

Before:

```scala
scala> :kind ({type l[A] = Either[Int, A]})#l
<console>:1: error: ';' expected but '#' found.
({type l[A] = Either[Int, A]})#l
                              ^

scala> trait Nat[-F1[_], +F2[_]] {}
defined trait Nat

scala> type ~>[-F1[_], +F2[_]] = Nat[F1, F2]
defined type alias $tilde$greater

scala> :kind ({type l[F[-_]] = F ~> List})#l
<console>:1: error: ';' expected but '#' found.
({type l[F[-_]] = F ~> List})#l
                             ^
```

After:

```scala
scala> :kind ({type l[A] = Either[Int, A]})#l
scala.util.Either[Int,?]'s kind is F[+A]

scala> trait Nat[-F1[_], +F2[_]] {}
defined trait Nat

scala> type ~>[-F1[_], +F2[_]] = Nat[F1, F2]
defined type alias $tilde$greater

scala> :kind ({type l[F[-_]] = F ~> List})#l
Nat[?,[+A]List[A]]'s kind is X[-F[A]]
```

### picking things up from REPL

```scala
scala> trait Functor[F[_]] {}
...
defined trait Functor

scala> :kind -v Functor
Functor's kind is X[F[A]]
(* -> *) -> *
This is a type constructor that takes type constructor(s): a higher-kinded type.

scala> object Bar { class Bop }
defined object Bar

scala> import Bar.Bop
import Bar.Bop

scala> :kind Bop
Bar.Bop's kind is A
```

### helpful help message

I added a bit of documentation in `:help` as well.

```
scala> :help kind

:kind [-v] <type>
Displays the kind of a given type.

    -v      Displays verbose info.

Kind is a system to describe types and type constructors, which are themselves also types.
`Int`, `Option[Int]` and every other types that you can make a value out of are called
a "proper type" and denoted as `A` using Scala notion, or as * symbol.

    scala> :kind Option[Int]
    Option[Int]'s kind is A

In the above `Option` is an example of a first-order type constructor, which is denoted as
`F[A]` using Scala notation, or * -> * using the star notation. Because of variance,
it's actually `F[+A]`.

    scala> :k -v Option
    Option's kind is F[+A]
    * -(+)-> *
    This is a type constructor: a 1st-order-kinded type.

When you have more complicated type parameters, :kind command can be used to find out
what you need to pass in.

    scala> trait ~>[-F1[_], +F2[_]] {}
    scala> :kind ~>
    ~>'s kind is X[-F1[A1],+F2[A2]]

This shows that `~>` accepts something of `F[A]` kind such as `List` and `Vector`.
```


Jasper-M:
Is it possible to suppress output like this?
```
<console>:1: error: identifier expected but integer literal found.
def $foo: 5 = ???
          ^
```
I think the REPL should try to expose as little of its internals as possible to the user.

adriaanm:
> Is it possible to suppress output like this?

Yes, see beQuietDuring/beSilentDuring/printResults.

eed3si9n:
> Is it possible to suppress output like this?

Most internal interpreter calls are in fact suppressed. I'm displaying the error message in question intentionally when a type was not found like this (note `false` that's passed in):

```diff
+        case NoType =>
+          exprTyper.typeOfExpression(s"def $$foo: $typeString = ???", false)
+          echo(":kind requires a type.")
```

This is analogous to `:type` command letting know the user that the term expression is not valid:

```scala
scala> :type "1" / 0
<console>:12: error: value / is not a member of String
       "1" / 0
           ^

scala> :type Foo
<console>:12: error: not found: value Foo
       Foo
       ^
```

I'd be happy to change to something like

```scala
scala> :kind 5
<console>: error: type 5 was not found
```

if you guys think that's better.

eed3si9n:
btw I've added d465df421e1caebcca89db577933531c49ee8c6e last week so the error message is simpler:

```
scala> :kind 5
<console>: error: type 5 was not found
```

Jasper-M:
FWIW I think it's better this way.

src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -295,14 +295,47 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
...
+       |    This is a type constructor: a 1st-order-kinded type.
+       |
+       |When you have more complicated type parameters, :kind command can be used to find out
+       |what you need to pass in.
+       |
+       |    scala> trait ~>[-F1[_], +F2[_]] {}
+       |    scala> :kind ~>
+       |    ~>'s kind is X[-F1[A1],+F2[A2]]
+       |
+       |This shows that `~>` accepts something of `F[A]` kind such as `List` and `Vector`.""".stripMargin
```
SethTisue: suggested revised text:

```text
Displays the kind of a given type.

    -v      Displays verbose info.

"Kind" is a word used to classify types and type constructors
according to their level of abstractness.

Concrete, fully specified types such as `Int` and `Option[Int]`
are called "proper types" and denoted as `A` using Scala
notation, or with the `*` symbol.

    scala> :kind Option[Int]
    Option[Int]'s kind is A

In the above, `Option` is an example of a first-order type
constructor, which is denoted as `F[A]` using Scala notation, or
* -> * using the star notation. `:kind` also includes variance
information in its output, so if we ask for the kind of `Option`,
we actually see `F[+A]`:

    scala> :k -v Option
    Option's kind is F[+A]
    * -(+)-> *
    This is a type constructor: a 1st-order-kinded type.

When you have more complicated types, `:kind` can be used to find
out what you need to pass in.

    scala> trait ~>[-F1[_], +F2[_]] {}
    scala> :kind ~>
    ~>'s kind is X[-F1[A1],+F2[A2]]

This shows that `~>` accepts something of `F[A]` kind, such as
`List` or `Vector`.
```

eed3si9n: Thanks for the revision. I'll adopt it.

retronym: I've pushed that change to your branch.

retronym (approved):

src/repl/scala/tools/nsc/interpreter/ILoop.scala
eed3si9n: Awesome. Thanks for the merge!

PR #5917, opened 2017-05-22
fix @switch annotation positions

xuwei-k:
I believe `@switch def` is no effect.


retronym (approved):

SethTisue:
thanks Kenji — nice catch

PR #5918, opened 2017-05-23
Fixes scala/bug#10328:PipedSource should join Source thread to wait for the exit value

chengpohi:
Try to fix https://github.com/scala/bug/issues/10328, this issue is caused by when try to wait `first.exitValue` and `second.exitValue`, we are hoping the `source: PipeSource` `thread` has finished running(finished transferring). 

Fix: when get the `exitValue`, we should block and wait the `source` `thread` finish running by `source.join()`.

### DIFF
--- a/src/library/scala/sys/process/ProcessImpl.scala
+++ b/src/library/scala/sys/process/ProcessImpl.scala
@@ -147,6 +147,7 @@ private[process] trait ProcessImpl {
           throw err
         }
       runInterruptible {
+        source.join()
         val exit1 = first.exitValue()
         val exit2 = second.exitValue()
         // Since file redirection (e.g. #>) is implemented as a piped process,
--- a/test/junit/scala/sys/process/PipedProcessTest.scala
+++ b/test/junit/scala/sys/process/PipedProcessTest.scala
@@ -9,6 +9,7 @@ import java.lang.reflect.InvocationTargetException
 import scala.concurrent.{Await, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.control.Exception.ignoring
+import org.junit.Assert.assertEquals
 
 // Each test normally ends in a moment, but for failure cases, waits two seconds.
 // scala/bug#7350, scala/bug#8768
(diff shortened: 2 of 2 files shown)


src/library/scala/sys/process/ProcessImpl.scala
```diff
@@ -147,6 +147,7 @@ private[process] trait ProcessImpl {
           throw err
         }
       runInterruptible {
+        source.join()
         val exit1 = first.exitValue()
```
som-snytt, line 151: I wonder why this isn't sufficient. I might have broken it previously in `CompoundProcess` at line 102, where I left the line comment `thread.join()`. (I'll check it out later.)

lrytz:
ping @som-snytt for the review, maybe we can still get this in 2.12.3

som-snytt:
Sorry for delay, I spent a day reviewing the package during last holiday, today I'll take another look. I think asking for the exitvalue of a process should mean it's "done, done", including i/o. I see a process can say it has no exit value. Apparently it works for regular processes but not for special redirects like `< url` or `url | consumer` which masquerade as process and are a ThreadProcess.

I may not be the sharpest tool in the shed, but the plethora of threads and conveniences make the process package hard to follow. With 2.13 and java 8, some simplifications are possible. Not sure how integration with java api while supporting url process will look like.

som-snytt:
LGTM

My two questions are why doesn't `sink.join()` get the same treatment and which change broke the behavior? I guess those can be deferred.

som-snytt (approved):

src/library/scala/sys/process/ProcessImpl.scala
som-snytt: This was my comment in May. FSR github review made it a comment from now.

The answer to my question is `ThreadProcess` just sets up the stream for redirection. `PipedProcess` consumes the stream. `ThreadProcess` itself doesn't know if or when the i/o succeeded, alas.

som-snytt:
I had forgotten the history of the code, trying to manage resources.

2012, on 2.10.x:
https://github.com/scala/bug/issues/5439
https://github.com/scala/bug/issues/6488

2014, on 2.12.x:
https://github.com/scala/bug/issues/7350
https://github.com/scala/scala/commit/67b3b9da7a517f702c331b42987f1debe49184d6
https://github.com/scala/bug/issues/10007

The refactor from summer 2014 is on 2.12.x.

lrytz:
@chengpohi maybe would be great to hear what you think on the two questions above https://github.com/scala/scala/pull/5918#issuecomment-315459718

src/library/scala/sys/process/ProcessImpl.scala
chengpohi: Hi @som-snytt and @lrytz  , Yes, Since the `Spawn` start a new `ThreadProcess` to run the `PipedProcess`, but `I/O` is also async running. so **uncomment** `thread.join()` will not fix this. we need to wait the `I/O` by `source.join()`  in the new  `ThreadProcess`.

som-snytt:
The answer to my second question (how did my last change break it) is that it already had this behavior after the earlier refactor, but was broken worse by the subsequent hang which my change addressed.

My answer to the first question, should it wait on `#>` was not to hold up the PR.

retronym:
Resubmitted with a minor reformatting of the commit message in #5998.

@chengpohi For future pull requests, I'd also suggest to create a branch in your repo with the fix, rather than commiting to your `2.12.x` branch. Following this guideline makes it easier to manage your changes.

chengpohi:
Sure, Thanks @retronym.

som-snytt:
Thanks @retronym I looked at the `process` package so long, I forgot to look at basic commit hygiene.

PR #5921, opened 2017-05-26
Remove new modifier on case classes in compiler

janekdb:
Based on IntelliJ IDEA 2017.1.2 code analysis.


som-snytt:
That's OK, but the compiler just rewrites them as new. I don't know what IntelliJ is trying to express.

janekdb:
Superfluous syntax I would suppose.

SethTisue:
thanks!

PR #5923, opened 2017-05-31
Improve the scaladoc -d param description

dkomlen:
Overriding the MutableSettings settings will keep the last setting added.
This enables scaladoc to use same -d param with different description
from scalac.

fixes scala/bug#9609

### DIFF
--- a/src/compiler/scala/tools/nsc/settings/MutableSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/MutableSettings.scala
@@ -213,9 +213,15 @@ class MutableSettings(val errorFn: String => Unit)
     }
 
   // a wrapper for all Setting creators to keep our list up to date
-  private def add[T <: Setting](s: T): T = {
-    allSettings += s
-    s
+  private[nsc] def add[T <: Setting](setting: T): T = {
+
+    allSettings.find(_.name == setting.name) match {
+      case Some(s) => allSettings -= s
+      case None =>
+    }
+
+    allSettings += setting
+    setting
   }
 
   def BooleanSetting(name: String, descr: String) = add(new BooleanSetting(name, descr))
@@ -233,7 +239,7 @@ class MutableSettings(val errorFn: String => Unit)
   def MultiStringSetting(name: String, arg: String, descr: String, helpText: Option[String]) = add(new MultiStringSetting(name, arg, descr, helpText))
   def MultiChoiceSetting[E <: MultiChoiceEnumeration](name: String, helpArg: String, descr: String, domain: E, default: Option[List[String]] = None) =
     add(new MultiChoiceSetting[E](name, helpArg, descr, domain, default))
-  def OutputSetting(outputDirs: OutputDirs, default: String) = add(new OutputSetting(outputDirs, default))
+  def OutputSetting(outputDirs: OutputDirs, default: String, descr: String) = add(new OutputSetting(outputDirs, default, descr))
   def PhasesSetting(name: String, descr: String, default: String = "") = add(new PhasesSetting(name, descr, default))
   def StringSetting(name: String, arg: String, descr: String, default: String, helpText: Option[String]) = add(new StringSetting(name, arg, descr, default, helpText))
   def ScalaVersionSetting(name: String, arg: String, descr: String, initial: ScalaVersion, default: Option[ScalaVersion] = None) =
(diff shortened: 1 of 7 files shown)


SethTisue:
from the Scala open source spree, Copenhagen 2017

jvican (commented):
LGTM, good job!

SethTisue:
@som-snytt you're always up in the settings code, want to take a quick look?

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -55,7 +55,7 @@ trait ScalaSettings extends AbsScalaSettings
   // argfiles is only for the help message
   /*val argfiles = */ BooleanSetting    ("@<file>", "A text file containing compiler arguments (options and source files)")
   val classpath     = PathSetting       ("-classpath", "Specify where to find user class files.", defaultClasspath) withAbbreviation "-cp"
-  val d             = OutputSetting     (outputDirs, ".")
+  val d             = OutputSetting     (outputDirs, ".", "Destination for generated classfiles.")
```
lrytz, line 58: Instead of overriding the setting, we could pull out the description string into a method here and override that in `doc/Settings`. That would avoid the creation of two `OutputSetting` instances.

dkomlen: True, my reasoning here was to make a generic approach where any overridden setting in the hierarchy would set the new setting value. For example if there is a need to change setting "-x" for scaladoc or any other settings inheriting from scalac settings, all that is needed is:

```scala
override val x = StringSetting(...)
```
With method approach we would have to change the base settings by adding a new method every time.

If the performance/memory is a concern for creating the extra OutputSetting I can make it a method instead. Let me know what you think.

src/compiler/scala/tools/nsc/settings/MutableSettings.scala
```diff
@@ -213,9 +213,15 @@ class MutableSettings(val errorFn: String => Unit)
...
-    s
+  private[nsc] def add[T <: Setting](setting: T): T = {
+
+    allSettings.find(_.name == setting.name) match {
+      case Some(s) => allSettings -= s
+      case None =>
+    }
+
+    allSettings += setting
+    setting
```
som-snytt, line 224: Setting doesn't have a useful hashCode or equals. Maybe ensure unique names this way:
```
diff --git a/src/compiler/scala/tools/nsc/settings/MutableSettings.scala b/src/compiler/scala/tools/nsc/settings/MutableSettings.scala
index a176552..b03b2a1 100644
--- a/src/compiler/scala/tools/nsc/settings/MutableSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/MutableSettings.scala
@@ -214,6 +214,7 @@ class MutableSettings(val errorFn: String => Unit)
 
   // a wrapper for all Setting creators to keep our list up to date
   private def add[T <: Setting](s: T): T = {
+    allSettings -= s
     allSettings += s
     s
   }
@@ -384,6 +385,13 @@ class MutableSettings(val errorFn: String => Unit)
     private var _deprecationMessage: Option[String] = None
     override def deprecationMessage = _deprecationMessage
     def withDeprecationMessage(msg: String): this.type = { _deprecationMessage = Some(msg) ; this }
+
+    override def hashCode = name.hashCode
+
+    override def equals(other: Any) = other match {
+      case s: Setting => name == s.name
+      case _ => false
+    }
   }

```
There are a lot of settings, so linear search suffers.

dkomlen: Added your suggestion in second commit.

dkomlen:
Updated git author e-mail in both commits with forced push on my branch.

som-snytt:
`test/files/run/settings-parse.scala` uses `MutableSettings.toString` which depends on iteration order of the hash set of settings. Probably `AbsSettings.toString` should sort first.

adriaanm:
Ping! 👋  For this to ship as part of 2.12.4, it will have to be mergeable by Sep 27. In this case, that means an approved review.

dkomlen:
Hello. 👋 I'd more than happy to get this shipped. I've forwarded your ping to scala/contributors for someone to take a look.

adriaanm:
It looks like we'll need to push this to 2.12.5. Before we merge, let's also make sure this doesn't break tooling (e.g. Scala IDE).

PR #5927, opened 2017-06-06
Fix completion of x.bar.<TAB>, where bar takes dependent implicits

retronym:
Extended the logic that typechecks the located tree if it is
an unapplied method to also work for polymorphic methods.

Fixes scala/bug#10353


mpollmeier:
@retronym can this PR be merged? I don't want to rush, but would love to see this released in 2.12.3.

adriaanm (approved):

PR #5929, opened 2017-06-07
Improve presentation compiler with Shapeless macros

retronym:
`-Ymacro-expand` discard was introduced to leave prefixes and
arguments of implicit applications in the typechecked tree in the
presentation compiler to facilite completion, type-at-cursor, etc
within those trees.

It seems that this mode interferes with implicit searches that involve
Shapeless's `mkLazy` macro.

This commit simply turns off the macro expansion discarding if
the current macro application is part of an application of implicit
arguments. There is no trees corresponding to source code in that
position, so we the concerns about IDE functionality are moot.

I couldn't disentangle the bug report from circe and shapeless,
so I've manually tested that the compiler from this commit
makes the following test project compile.

https://github.com/retronym/t9716

Fixes scala/bug#9716

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -237,6 +237,9 @@ trait Contexts { self: Analyzer =>
 
     /** Types for which implicit arguments are currently searched */
     var openImplicits: List[OpenImplicit] = List()
+    final def isSearchingForImplicitParam: Boolean = {
+      openImplicits.nonEmpty && openImplicits.exists(x => !x.isView)
+    }
 
     /* For a named application block (`Tree`) the corresponding `NamedApplyInfo`. */
     var namedApplyBlockInfo: Option[(Tree, NamedApplyInfo)] = None
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -262,7 +262,21 @@ trait Implicits {
 
   /** A class which is used to track pending implicits to prevent infinite implicit searches.
    */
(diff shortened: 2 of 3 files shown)


retronym:
@mpollmeier @milessabin Perhaps you could give the combination of this and #5927 a spin in Ensime ([instructions](https://github.com/scala/bug/issues/10353#issuecomment-305995374)) to see if it makes life better? Be sure to disable the any manual setting of `-Ymacro-expand:normal` in the presentation compiler options that might have been added as a workaround for this bug.

/cc @fommil

fommil:
we use `discard` https://github.com/ensime/ensime-sbt/blob/2.0/src/main/scala/EnsimePlugin.scala#L219 but only because somebody told me to do it. What should we be using?

BTW, somewhat related me and @dragos started writing https://github.com/ensime/ensime-plugin-implicits but ran out of time. This will have huge benefits for presentation compiler users in implicit-heavy codebases.

retronym:
Prior to this change, both discard and normal mode would lead to problems
in different circumstances. Users of Scala IDE found that by explicitly
selecting normal, which stops the pres compiler defaulting to discard,
shapeless worked better. But they would sacrifice code completion in the
arguments of non-implicit macros

After this change, discard should work for both use cases. I want to make
sure anyone testing this patch tests it under that configuration.

mpollmeier:
@retronym I tested with ensime and now the autocompletion works fine in both cases :)
I didn't have to change anything other than the scala jars in `.ensime`. As @fommil said, ensime uses `-Ymacro-expand:discard` by default.

fommil:
awesome, looking forward to this! @mpollmeier if you'd like to help out with presentation compiler testing / performance there are a few things I have in mind, none of them particularly onerous :smile:

adriaanm (approved):

PR #5930, opened 2017-06-07
Remove deprecated view bounds in library

janekdb:
Replace view bounds with implicit parameter.

`evidence$1` is the name used by the parser when converting the view bound
to a parameter. Picking this name instead of say `ev` results in an unchanged
method signature.

### DIFF
--- a/src/library/scala/math/Ordering.scala
+++ b/src/library/scala/math/Ordering.scala
@@ -146,7 +146,7 @@ trait LowPriorityOrderingImplicits {
    *  turn up if nothing else works.  Since `Ordered[A]` extends
    *  `Comparable[A]` anyway, we can throw in some Java interop too.
    */
-  implicit def ordered[A <% Comparable[A]]: Ordering[A] = new Ordering[A] {
+  implicit def ordered[A](implicit @deprecatedName('evidence$1, "2.12.3") ev: A => Comparable[A]): Ordering[A] = new Ordering[A] {
     def compare(x: A, y: A): Int = x compareTo y
   }
   implicit def comparatorToOrdering[A](implicit cmp: Comparator[A]): Ordering[A] = new Ordering[A] {
--- a/src/library/scala/math/PartiallyOrdered.scala
+++ b/src/library/scala/math/PartiallyOrdered.scala
@@ -25,24 +25,24 @@ trait PartiallyOrdered[+A] {
    *  - `x == 0`   iff   `'''this''' == that`
    *  - `x > 0`    iff   `'''this''' &gt; that`
    */
-  def tryCompareTo [B >: A <% PartiallyOrdered[B]](that: B): Option[Int]
+  def tryCompareTo [B >: A](that: B)(implicit @deprecatedName('evidence$1, "2.12.3") ev: B => PartiallyOrdered[B]): Option[Int]
 
-  def <  [B >: A <% PartiallyOrdered[B]](that: B): Boolean =
+  def <  [B >: A](that: B)(implicit @deprecatedName('evidence$1, "2.12.3") ev: B => PartiallyOrdered[B]): Boolean =
     (this tryCompareTo that) match {
       case Some(x) if x < 0 => true
       case _ => false
     }
-  def >  [B >: A <% PartiallyOrdered[B]](that: B): Boolean =
+  def >  [B >: A](that: B)(implicit @deprecatedName('evidence$1, "2.12.3") ev: B => PartiallyOrdered[B]): Boolean =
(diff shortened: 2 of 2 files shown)


src/library/scala/math/Ordering.scala
```diff
@@ -146,7 +146,7 @@ trait LowPriorityOrderingImplicits {
    *  turn up if nothing else works.  Since `Ordered[A]` extends
    *  `Comparable[A]` anyway, we can throw in some Java interop too.
    */
-  implicit def ordered[A <% Comparable[A]]: Ordering[A] = new Ordering[A] {
+  implicit def ordered[A](implicit evidence$1: A => Comparable[A]): Ordering[A] = new Ordering[A] {
```
lrytz: maybe add a `// todo` for 2.13 do change the parameter name?

janekdb: I've used `deprecatedName` as a way to achieve this. WDYT?

janekdb: This is the code I'm using to determine the compiler output,

```
trait Test {
  implicit def ordered1[A <% Comparable[A]]: Unit = ()

  implicit def ordered2[A](implicit evidence$1: A => Comparable[A]): Unit = ()

  implicit def ordered3[A](implicit @deprecatedName('evidence$1, "2.12.3") ev: A => Comparable[A]): Unit = ()
}
```

```
10:31 $ ~/lib/scala-2.12.2/bin/scalac -Xprint:parser TestEvName.scala
[[syntax trees at end of                    parser]] // TestEvName.scala
package <empty> {
  abstract trait Test extends scala.AnyRef {
    def $init$() = {
      ()
    };
    implicit def ordered1[A](implicit evidence$1: _root_.scala.Function1[A, Comparable[A]]): Unit = ();
    implicit def ordered2[A](implicit evidence$1: _root_.scala.Function1[A, Comparable[A]]): Unit = ();
    implicit def ordered3[A](implicit @new deprecatedName(scala.Symbol("evidence$1"), "2.12.3") ev: _root_.scala.Function1[A, Comparable[A]]): Unit = ()
  }
```

janekdb:
@lrytz : TODO:

I've used `deprecatedName` as a way to achieve this. WDYT?

This is the code I'm using to determine the compiler output,

```
trait Test {
  implicit def ordered1[A <% Comparable[A]]: Unit = ()

  implicit def ordered2[A](implicit evidence$1: A => Comparable[A]): Unit = ()

  implicit def ordered3[A](implicit @deprecatedName('evidence$1, "2.12.3") ev: A => Comparable[A]): Unit = ()
}
```

```
10:31 $ ~/lib/scala-2.12.2/bin/scalac -Xprint:parser TestEvName.scala
[[syntax trees at end of                    parser]] // TestEvName.scala
package <empty> {
  abstract trait Test extends scala.AnyRef {
    def $init$() = {
      ()
    };
    implicit def ordered1[A](implicit evidence$1: _root_.scala.Function1[A, Comparable[A]]): Unit = ();
    implicit def ordered2[A](implicit evidence$1: _root_.scala.Function1[A, Comparable[A]]): Unit = ();
    implicit def ordered3[A](implicit @new deprecatedName(scala.Symbol("evidence$1"), "2.12.3") ev: _root_.scala.Function1[A, Comparable[A]]): Unit = ()
  }
```

src/library/scala/math/Ordering.scala
janekdb: See main conversation.

lrytz:
Good suggestion! I just realized that changing a parameter name (and marking it `@deprecatedName`) affects the classfile as we emit parameter names since https://github.com/scala/scala/commit/c78d771e6f.

```
$ cat Test.scala
class C {
  def m(@deprecatedName('hai, "2.12.3") bai: Int): Int = 0
}
$ scalac Test.scala
$ scala -e 'println(classOf[C].getDeclaredMethods.apply(0).getParameters.apply(0).getName)'
bai
```

If we care about this `@deprecatedName` basically becomes useless..

janekdb:
Okay, who can say if we care about this? If we do care then we replace `deprecatedName` with a `TODO` in this commit but what happens with `deprecatedName` in the future? When can it be used?

Alternatives to `TODO` comment

 1. Move to `2.13` branch using `ev` immediately (dodges issue but issue remains)
 2. Assume `deprecatedName` is not problematic (or will be fixed) and continue as-is

My preference is 2.

retronym:
I'd be in favour of moving this PR to 2.13.x. The risk is low, but I don't think there is enough advantage to having this cleanup in the minor release.

PR #5932, opened 2017-06-09
Remove unnecessary semicolons

janekdb:


janekdb:
Crushes *.scala source byte size down to 99.99977436%

SethTisue:
thanks!

PR #5933, opened 2017-06-10
Changing length function to more functional programming way

SakalyaDeshpande:
Current implementation uses vars and a loop.So reimplementing it in a FP way.

### DIFF
--- a/src/library/scala/collection/LinearSeqOptimized.scala
+++ b/src/library/scala/collection/LinearSeqOptimized.scala
@@ -45,13 +45,11 @@ trait LinearSeqOptimized[+A, +Repr <: LinearSeqOptimized[A, Repr]] extends Linea
    *  Note: the execution of `length` may take time proportional to the length of the sequence.
    */
   def length: Int = {
-    var these = self
-    var len = 0
-    while (!these.isEmpty) {
-      len += 1
-      these = these.tail
+    def loop(len:Int): Int = {
+      if (this.isEmpty) len
+      else loop(len + 1, this.tail)
     }
-    len
+    loop(0,this)
   }
 
   /** Selects an element by its index in the $coll.
--- a/src/library/scala/collection/LinearSeqOptimized.scala
+++ b/src/library/scala/collection/LinearSeqOptimized.scala
@@ -46,10 +46,10 @@ trait LinearSeqOptimized[+A, +Repr <: LinearSeqOptimized[A, Repr]] extends Linea
    */
   def length: Int = {
     def loop(len:Int): Int = {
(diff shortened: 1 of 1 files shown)


janekdb:
While it is nice to avoid `var` there might be a cost in terms of performance and in this case stack depth. So at the least this would need,

* Profiling
* Annotation with @tailrec

If this change has any prospect of being accepted it must be faster than the current approach so you should seek to establish that first.

More generally this would also need

* Squashing down to one commit
* Fixing of compile error
* Format the change lines, for example add a trailing space after the comma in the `loop` call.

Use `sbt` to compile and run tests locally

    $ sbt clean test

One question for you: If we set aside our convictions regarding fp versus non-fp how would you describe the key benefit of this change in the context of this narrow example?

You would also need to have a compelling argument for changing just this one instance of a `while` loop that increments a `var`. What makes this instance special? At a guess that are around 400+ similar uses of `while`. Will you be proposing a programme of activity to convert all into a tail-recursive format? 

````
$ grep while -A5 $(find src -name '*.scala')|grep '+='|cat -n|tail -5
   473	src/interactive/scala/tools/nsc/interactive/Global.scala-    toBeRemoved ++= toBeRemovedAfterRun
   474	src/interactive/scala/tools/nsc/interactive/tests/core/PresentationCompilerRequestsWorkingMode.scala-      buf += source.position(pos - 1) // we need the position before the first character of this marker
   475	src/interactive/scala/tools/nsc/interactive/tests/Tester.scala-      result += 1
   476	src/interactive/scala/tools/nsc/interactive/tests/Tester.scala-        buf += pos
   477	src/interactive/scala/tools/nsc/interactive/Picklers.scala:    while (start < bs.length && start < cs.length && bs(start) == cs(start)) start += 1
```

rorygraves:
@sakky11  Ignoring the compile error for a minute and following on @janekdb 's comment.   
To ensure this is performant you need to make sure the code is tail recursive so it is optimised (basically into the mutable code you see).  So add ```@tailrec``` otherwise this could could lead to a stack overflow on larger collections.

For performance critical code you would need to add a benchmark:
See https://github.com/scala/scala/tree/2.12.x/test/benchmarks and for a specific example look at https://github.com/scala/scala/blob/2.12.x/test/benchmarks/src/main/scala/scala/collection/immutable/ListBenchmark.scala

You can see an example of PR with benchmarking results here: 
https://github.com/scala/scala/pull/5755

lrytz:
This is just a matter of taste, using a local var in such a case is not considered bad style. I'm closing this for now, since there's no clear gain - anyone who disagrees, please comment!

PR #5934, opened 2017-06-10
Remove unnecessary imports and import braces

janekdb:


rorygraves:
👍

lrytz:
Thanks!

SethTisue:
(was this discovered with an automated tool? if so, it might be nice, in the future, if such commits indicated what tool was used)

dwijnand:
It would be even nicer if any such tool were setup to mechanically enforce these changes, rather than require constant supervision and/or manual maintenance: fix not just the issue, but the fact that the issue can arise. (But avoid https://xkcd.com/1319/)

PR #5937, opened 2017-06-11
Expose sha and commit date on VersionUtil

estsauver:
One of the things mentioned in the hacking guide is setting your local.sbt to enable sha based versioning to make it possible to easily use the scala compiler from another project. Those docs were out of date, as the script that exposed the sha had been removed and pulled into VersionUtil.

This change exposes a new property (GitProperties) to allow for referring to the sha in the basenameSuffix. The git properties had to be refactored out of the versionProperties because versionProperties depends on basenameSuffix.

See linked PR in docs here: https://github.com/scala/scala-lang/pull/668


estsauver:
@SethTisue @szeiger I think you two look like the best folks to review this change.

lrytz (approved):

szeiger (approved):

SethTisue:
thank you!

PR #5938, opened 2017-06-11
Fix generation of derived value classes wrapping Unit, Null, and Nothing

hrhino:
All sorts o' specialness going on here. `Unit` morphs into a `BoxedUnit` when it's in a field, but void when it's the return type of a method, which is expected. This means, though, that the unboxing method of a Unit-wrapping value class has the signature `()V`, not `()Lscala/runtime/BoxedUnit`, so attempting to use its value in the equals method spits out some wonderful invalid bytecode, instead. Similar sadness occurs for Nothing and Null as well.

The "solution" is to not even bother to check for equality, as we've only got at most one legitimate value of each of these types. Because the code is shared with `case class`es, this also changes the bytecode we generate for them. Obviously this is an "unrelated change" as far as the bugs this is meant to fix go, but it's innocuous enough as far as I can tell.

I also slipped a constructor call into the generated `ClassCastException` that gets thrown when we are asked to emit a cast for a primitive type in `BCodeBodyBuilder`, so we generate valid bytecode in that case. I've got no idea how `NEW; DUP; ATHROW` is possibly ever legitimate, but hey.

Fixes scala/bug#9240
Fixes scala/bug#10361

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -528,8 +528,10 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
             else if (l.isPrimitive) {
               bc drop l
               if (cast) {
+                devWarning(s"Tried to emit impossible cast from primitive type $l to $r (at ${app.pos})")
                 mnode.visitTypeInsn(asm.Opcodes.NEW, jlClassCastExceptionRef.internalName)
                 bc dup ObjectRef
+                mnode.visitMethodInsn(asm.Opcodes.INVOKESPECIAL, jlClassCastExceptionRef.internalName, INSTANCE_CONSTRUCTOR_NAME, "()V", true)
                 emit(asm.Opcodes.ATHROW)
               } else {
                 bc boolconst false
--- a/src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala
@@ -159,25 +159,41 @@ trait SyntheticMethods extends ast.TreeDSL {
     def thatCast(eqmeth: Symbol): Tree =
       gen.mkCast(Ident(eqmeth.firstParam), clazz.tpe)
 
-    /* The equality method core for case classes and inline classes.
+    /* The equality method core for case classes and derived value classes.
+     * Generally:
      * 1+ args:
      *   (that.isInstanceOf[this.C]) && {
      *       val x$1 = that.asInstanceOf[this.C]
      *       (this.arg_1 == x$1.arg_1) && (this.arg_2 == x$1.arg_2) && ... && (x$1 canEqual this)
      *      }
-     * Drop canBuildFrom part if class is final and canBuildFrom is synthesized
+     * Drop:
+     * - canEqual part if class is final and canEqual is synthesized.
+     * - test for arg_i if arg_i has type Nothing, Null, or Unit
+     * - asInstanceOf if no equality checks need made (see scala/bug#9240, scala/bug#10361)
      */
     def equalsCore(eqmeth: Symbol, accessors: List[Symbol]) = {
+      def usefulEquality(acc: Symbol): Boolean = {
+        val rt = acc.info.resultType
+        rt != NothingTpe && rt != NullTpe && rt != UnitTpe
+      }
+
       val otherName = context.unit.freshTermName(clazz.name + "$")
       val otherSym  = eqmeth.newValue(otherName, eqmeth.pos, SYNTHETIC) setInfo clazz.tpe
-      val pairwise  = accessors map (acc => fn(Select(mkThis, acc), acc.tpe member nme.EQ, Select(Ident(otherSym), acc)))
+      val pairwise  = accessors collect {
+        case acc if usefulEquality(acc) =>
+          fn(Select(mkThis, acc), acc.tpe member nme.EQ, Select(Ident(otherSym), acc))
+      }
       val canEq     = gen.mkMethodCall(otherSym, nme.canEqual_, Nil, List(mkThis))
       val tests     = if (clazz.isDerivedValueClass || clazz.isFinal && syntheticCanEqual) pairwise else pairwise :+ canEq
 
-      thatTest(eqmeth) AND Block(
-        ValDef(otherSym, thatCast(eqmeth)),
-        AND(tests: _*)
-      )
+      if (tests.isEmpty) {
+        thatTest(eqmeth)
+      } else {
+        thatTest(eqmeth) AND Block(
+          ValDef(otherSym, thatCast(eqmeth)),
+          AND(tests: _*)
+        )
+      }
     }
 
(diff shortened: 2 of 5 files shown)


src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala
```diff
@@ -159,25 +159,39 @@ trait SyntheticMethods extends ast.TreeDSL {
...
      *      }
-     * Drop canBuildFrom part if class is final and canBuildFrom is synthesized
+     * Drop:
+     * - canEqual part if class is final and canEqual is synthesized.
+     * - test for arg_i if arg_i has type Nothing, Null, or Unit
+     * - asInstanceOf if no equality checks need made (see scala/bug#9240, scala/bug#10361)
      */
     def equalsCore(eqmeth: Symbol, accessors: List[Symbol]) = {
+      def usefulEquality(acc: Symbol): Boolean =
+        acc.info.resultType != NothingTpe && acc.info.resultType != NullTpe && acc.info.resultType != UnitTpe
```
hrhino: I would think that these three types would have an alias somewhere, but I couldn't find it. Feel free to make me swap it out for something else.

test/files/run/wacky-value-classes.scala
```diff
@@ -0,0 +1,19 @@
+// scala/bug#9240
+final class AnyValNothing(val self: Nothing) extends AnyVal
+final class AnyValNull   (val self: Null   ) extends AnyVal
+// scala/bug#10361
```
dwijnand: Your references are inverted. The below is scala/bug#9240, and the ones above are scala/bug#10361.

test/files/run/wacky-value-classes.scala
```diff
@@ -0,0 +1,19 @@
...
+final class AnyValUnit   (val self: Unit   ) extends AnyVal
+
+object Test extends App {
+  def avn = new AnyValNull(null)
+  assert(avn == avn)
+  /*this throws NPE right now b/c null.hashCode() does...*/
+  //assert(avn.hashCode() == 0)
```
dwijnand: Switch this (and avu below) to `.##`. If this here returns 0 rather than throw an NPE then you also fixed https://github.com/scala/bug/issues/7396 with this patch!

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -526,12 +526,19 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
...
               genConversion(l, r, cast)
             else if (l.isPrimitive) {
-              bc drop l
               if (cast) {
-                mnode.visitTypeInsn(asm.Opcodes.NEW, jlClassCastExceptionRef.internalName)
-                bc dup ObjectRef
-                emit(asm.Opcodes.ATHROW)
+                if (l == UNIT && r == srBoxedUnitRef) {
+                  /* special case: we can "cast" V to Lscala/runtime/BoxedUnit; by summoning the latter */
+                  mnode.instructions.add(backendUtils.getBoxedUnit)
```
lrytz: I think this should be handled in erasure, at the place where the cast is inserted, see where [`cast`](https://github.com/scala/scala/blob/9ad9297e35455d3e6a36f28250c8afbf9fff0071/src/compiler/scala/tools/nsc/transform/TypeAdaptingTransformer.scala#L102) is invoked.

src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala
lrytz: The three comparisons look OK to me, but you could extract `acc.info.resultType` into a local.

lrytz (changes requested):

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
sjrd: :+1: It would also fix the problem for the three backends at the same time :)

test/files/run/wacky-value-classes.scala
hrhino: Woops. Good catch, thanks!

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
hrhino: How rude of me, I almost forgot that non-JVM backends exist... Changed!

test/files/run/wacky-value-classes.scala
hrhino: Sadly, that problem is deeper than this, as far as I can tell. No hat tricks here :disappointed:

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -530,6 +530,7 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
               if (cast) {
                 mnode.visitTypeInsn(asm.Opcodes.NEW, jlClassCastExceptionRef.internalName)
                 bc dup ObjectRef
+                mnode.visitMethodInsn(asm.Opcodes.INVOKESPECIAL, jlClassCastExceptionRef.internalName, INSTANCE_CONSTRUCTOR_NAME, "()V", true)
```
lrytz: This certainly looks right, but I wonder how we can end up here. We'd need an `x.asInstanceOf[T]` where `x` has primitive type. Any idea? When I write `1.asInstanceOf[String]`, the value is boxed.

lrytz (approved):
Very nice work, thanks a lot!

lrytz:
- [x] TODO: run [jardiff](https://github.com/scala/jardiff) -- no diff in library+reflect+compiler

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
sjrd: FWIW, Scala.js [has corresponding code](https://github.com/scala-js/scala-js/blob/99e6017f55427d83fb1cd35c55def9ed8e875e11/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L2344-L2352) in its codegen, but I'm not sure it is ever exercised ^^

hrhino: I can't figure out how to get this branch to get hit... we should have already inserted the boxing by now. This somewhat implies to me that this *could*, possibly should, be an assertion or similar; before this patch, this would always generate bytecode that doesn't pass verification.

retronym: I agree with your analysis.

When in doubt about old code like this, we tend to add a `devWarning` (which issues a warning under `-Ydebug` or `-Xdev`) rather than a hard `assert` to fail gracefully

hrhino: Devs are duly warned. Good idea!

PR #5940, opened 2017-06-13
Remove unused format arg in interpreter echo

janekdb:


src/repl/scala/tools/nsc/interpreter/ILoop.scala
```diff
@@ -614,8 +614,8 @@ class ILoop(in0: Option[BufferedReader], protected val out: JPrintWriter)
     if (f.exists) {
       addedClasspath = ClassPath.join(addedClasspath, f.path)
       intp.addUrlsToClassPath(f.toURI.toURL)
-      echo("Added '%s' to classpath.".format(f.path, intp.global.classPath.asClassPathString))
-      repldbg("Added '%s'.  Your new classpath is:\n\"%s\"".format(f.path, intp.global.classPath.asClassPathString))
+      echo("Added '%s' to classpath.".format(f.path))
+      repldbg("Added '%s'. Your new classpath is:\n\"%s\"".format(f.path, intp.global.classPath.asClassPathString))
```
som-snytt, line 618: @adriaanm is refactoring/rewriting. Also, it would be natural to change this to interpolated string, which is immune to the error. There should be a style rule about format strings containing only `%s`.

janekdb: Let's close this if @adriaanm will include it.

adriaanm: Let's merge this here. Since this method is deprecated, I've removed it in my 2.13 refactoring (it will be replaced by a more general mechanism of adding dependencies similar to ammonite). We're currently not planning to backport the 2.13 repl to 2.12, so, in general, I would ask that changes to the repl target 2.13 by default (once #5903 is merged ;-))

janekdb: @adriaanm - could you mark this a passing review in that case?

adriaanm (approved):
Thanks!

PR #5944, opened 2017-06-14
Fix code gen for Outer.super[Q].foo

lrytz:
Consider

    class B extends T { class C { B.super[T].f }}

After flatten, that call is ` B$C.this.$outer().super[T].f()`.

In 2.11, mixin translates this to `A$class.f(B$C.this.$outer())`.
In 2.12, the tree is passed unchanged to the backend.

In `genApply` we assumed that in `Apply(Select(Super(qual, ... )))`,
`qual` is a `This` tree, so we just emitted `ALOAD_0`, which caused
the `$outer()` call to get lost. Now we invoke `genLoad(qual)`.

Fixes scala/bug#10290.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -317,9 +317,13 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
           }
           else {
             mnode.visitVarInsn(asm.Opcodes.ALOAD, 0)
-            generatedType =
-              if (tree.symbol == ArrayClass) ObjectRef
-              else classBTypeFromSymbol(claszSymbol)
+            // When compiling Array.scala, the constructor invokes `Array.this.super.<init>`. The expectedType
+            // is `[Object` (computed by typeToBType, the type of This(Array) is `Array[T]`). If we would set
+            // the generatedType to `Array` below, the call to adapt at the end would fail. The situation is
+            // similar for primitives (`I` vs `Int`).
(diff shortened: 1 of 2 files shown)


lrytz:
- [x] TODO: jardiff that -- empty

sjrd:
Aaand yup, Scala.js has exactly the same bug, for exactly the same reason: https://github.com/scala-js/scala-js/issues/3013 :smile:

lrytz:
@sjrd @densh you guys are so quick, I just wanted to notify you :)

dwijnand:
"code gen" is a trigger word :D

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -317,9 +317,12 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
           }
           else {
             mnode.visitVarInsn(asm.Opcodes.ALOAD, 0)
-            generatedType =
-              if (tree.symbol == ArrayClass) ObjectRef
```
lrytz, line 321: This special case is old, it goes back to https://github.com/scala/scala/commit/0ad5e129f3. Note that on a current 2.12.x, that code path is never triggered when compiling library+compiler+reflect. With this PR we get here again, but `Object` doesn't work when the expected type is `[Object`.

lrytz: (It seems we need more special cases for primitive types, locally I just got `Cannot emit primitive conversion from Lscala/Boolean; to Z - Boolean.scala`).

lrytz:
jardiff is empty (library+reflect+compiler)

retronym (approved):

PR #5946, opened 2017-06-14
Fix check for varargs calling convention in overrides of Java methods

hrhino:
The comment explains it all, but to repeat: Scala-defined varargs methods take a `Seq` parameter; Java-defined varargs methods take an `Array`. `transformArgs` in `uncurry` needs to wrap the passed args in the right data structure. Checking for `fun.isJava` works quite nicely, except for when we've made a superaccessor for some Java-defined method, the superaccessor takes an `Array`, but does not have `isJava` set, and therefore we wrap the args in a `Seq`, netting us a tasty `ClassCastException` at runtime.

The solution: check with `isJavaVarArgsMethod` instead, which can capture this distinction correctly.

Alternate solution: change `superaccessors` to generate a scala-ish varargs method signature. I didn't pick this solution because, as of right now, `superaccessors` knows not of varargs, and it might be better left that way.

Fixes scala/bug#10368

Review by @retronym 

Quoth the jardiff on scala/scala: [No changes.](https://gist.github.com/hrhino/eb88a0ee510e81e1726448c0675bc653)

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -580,7 +580,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     val runsRightAfter = None
   } with Delambdafy
 
-  // phaseName = "bcode"
+  // phaseName = "jvm"
   object genBCode extends {
     val global: Global.this.type = Global.this
     val runsAfter = List("cleanup")
--- a/src/compiler/scala/tools/nsc/transform/UnCurry.scala
+++ b/src/compiler/scala/tools/nsc/transform/UnCurry.scala
@@ -285,10 +285,16 @@ abstract class UnCurry extends InfoTransform
           }
         }
 
+        /* Java-style varargs = expects `Array` rather than `Seq`
+         * Note that `fun.isJavaDefined` is not good enough because
+         * if we override a varargs method defined in Java, `superaccessors`
+         * will make us a superaccessor which also takes `Array` rather than `Seq`.
+         * See scala/bug#10368 */
+        val javaStyleVarArgs = isJavaVarArgsMethod(fun)
         var suffix: Tree =
           if (treeInfo isWildcardStarArgList args) {
             val Typed(tree, _) = args.last
-            if (isJava)
+            if (javaStyleVarArgs)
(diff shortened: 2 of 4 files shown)


src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -580,7 +580,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     val runsRightAfter = None
   } with Delambdafy
 
-  // phaseName = "bcode"
+  // phaseName = "jvm"
```
hrhino, line 583: This of course being The Real Fix.

hrhino:
/rebuild

This... looks spurious? I hope...

```
[error] Error during tests:
[error] 	Forked test harness failed: java.lang.NullPointerException
[error] 	at sbt.TestEvent$$anonfun$overallResult$1.apply(TestReportListener.scala:72)
[error] 	at sbt.TestEvent$$anonfun$overallResult$1.apply(TestReportListener.scala:71)
[error] 	at scala.collection.IndexedSeqOptimized$class.foldl(IndexedSeqOptimized.scala:51)
[error] 	at scala.collection.IndexedSeqOptimized$class.foldLeft(IndexedSeqOptimized.scala:60)
[error] 	at scala.collection.mutable.WrappedArray.foldLeft(WrappedArray.scala:34)
[error] 	at scala.collection.TraversableOnce$class./:(TraversableOnce.scala:138)
[error] 	at scala.collection.AbstractTraversable./:(Traversable.scala:105)
[error] 	at sbt.TestEvent$.overallResult(TestReportListener.scala:71)
[error] 	at sbt.TestEvent$$anon$1.<init>(TestReportListener.scala:66)
[error] 	at sbt.TestEvent$.apply(TestReportListener.scala:65)
[error] 	at sbt.React.react(ForkTests.scala:137)
[error] 	at sbt.ForkTests$$anonfun$mainTestTask$1$Acceptor$2$.run(ForkTests.scala:76)
[error] 	at java.lang.Thread.run(Thread.java:745)
```

SethTisue:
yeah that's https://github.com/scala/scala-dev/issues/277 (at least offhand it looks the same)

lrytz (approved):

lrytz:
Nice, thanks!

PR #5947, opened 2017-06-15
Support serializable on MatchError

estsauver:
Since MatchError ultimately extends throwable, it needs to implement the Serializable interface. To do this, we need to eagerly evaluate the error message.

Closes scala/bug#10369

### DIFF
--- a/src/library/scala/MatchError.scala
+++ b/src/library/scala/MatchError.scala
@@ -19,9 +19,9 @@ package scala
  *  @version 1.1, 05/03/2004
  *  @since   2.0
  */
-final class MatchError(obj: Any) extends RuntimeException {
+final class MatchError(@transient obj: Any) extends RuntimeException {
   /** There's no reason we need to call toString eagerly,
-   *  so defer it until getMessage is called.
+   *  so defer it until getMessage is called or object is serialized
    */
   private lazy val objString = {
     def ofClass = "of class " + obj.getClass.getName
@@ -31,6 +31,12 @@ final class MatchError(obj: Any) extends RuntimeException {
     } catch {
       case _: Throwable => "an instance " + ofClass
     }
+  }
+
+  @throws[java.io.ObjectStreamException]
+  private def writeReplace(): Object = {
+    objString
+    this
   }
 
   override def getMessage() = objString
--- a/test/junit/scala/MatchErrorSerializationTest.scala
+++ b/test/junit/scala/MatchErrorSerializationTest.scala
@@ -1,0 +1,23 @@
+package scala
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+/**
+  * Created by estsauver on 6/15/17.
+  */
+@RunWith(classOf[JUnit4])
+class MatchErrorSerializationTest {
+
+  @Test
+  def canSerializeMatchError = {
+    import java.io._
+    val matchError = new MatchError(new Object)
+    val barrayOut = new ByteArrayOutputStream()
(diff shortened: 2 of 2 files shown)


src/library/scala/MatchError.scala
```diff
@@ -10,6 +10,8 @@
 
 package scala
 
+import java.beans.Transient
```
retronym: Accidentally imported?

src/library/scala/MatchError.scala
```diff
@@ -19,11 +21,11 @@ package scala
  *  @version 1.1, 05/03/2004
  *  @since   2.0
  */
-final class MatchError(obj: Any) extends RuntimeException {
+final class MatchError(@transient obj: Any) extends RuntimeException {
   /** There's no reason we need to call toString eagerly,
    *  so defer it until getMessage is called.
    */
-  private lazy val objString = {
+  private val objString = {
```
retronym: The comment would need to be updated here to match the new behaviour.

However, I'm a little concerned about making this change, calling `toString` was intentionally deferred in 46d5e73c11bccd0e892429e4c3f2ac1a02cea2a9.

It might be possible to defer `toString` until the point of serialization, with something like this (not tested):

```scala
final class MatchError(@transient obj: Any) extends RuntimeException {
  /** There's no reason we need to call toString eagerly,
   *  so defer it until getMessage is called. We do trigger the message creation on serialization
   *
   */
  private var objString: String = null

  @throws[ObjectStreamException]
  private def writeReplace(): Object = {
    objString = computeObjString
    this
  }

  private def computeObjString: String = {
    def ofClass = "of class " + obj.getClass.getName
    if (obj == null) "null"
    else try {
      obj.toString() + " (" + ofClass + ")"
    } catch {
      case _: Throwable => "an instance " + ofClass
    }
  }

  override def getMessage() = {
    if (objString == null) computeObjString
    objString
  }
}

```

src/library/scala/MatchError.scala
estsauver: Yah, intellij auto import was a little over zealous.

estsauver:
Pushed your suggested changes. I made a couple changes so that the method to compute the string would only ever be called at most once.

src/library/scala/MatchError.scala
hrhino: Is there any reason why a `lazy val` wouldn't still work? I did this (on 2.12.1) and the test still passed, and the bytecode appears unobjectionable. It makes this a very small change.

```scala
final class MatchError(@transient obj: Any) extends RuntimeException with Serializable {
  private lazy val message = {
    def ofClass = "of class " + obj.getClass.getName
    if (obj == null) "null"
    else try {
      obj.toString + " (" + ofClass + ")"
    } catch {
      case _: Throwable => "an instance " + ofClass
    }
  }
  @throws[java.io.ObjectStreamException]
  private def writeReplace(): Object = {
    message // force eval
    this
  }
  override def getMessage = message
}
```

estsauver: @hrhino ... you're absolutely right. I just started implementing it one way and then didn't finish thinking through how the other changes could make it easier.

There's no need to reimplement lazy vals :)

estsauver:
I added @hrhino 's suggestion and the change ended up being very pleasantly small. :)

estsauver:
@retronym should be ready for review again.

retronym (approved):
LGTM. The only downside is that we won't be able to deserialize `MatchError`-s from prior versions of the standard library, but we don't have strong guarantees about that, especially not about `Throwable`-s, so I'm willing to live with this.

SethTisue:
thank you!

PR #5949, opened 2017-06-18
Teach the optimizer about type tests on primitive box types.

hrhino:
Currently, we generate a box, `.isInstanceOf` call, and then unbox, even for something as simple as:

    (1: Any) match { case i: Int => i; case _ => 0 }

The optimizer currently doesn't know enough to handle that case, and emits naive code which `new`s up a box for `i`, checks if it is truly a `java.lang.Integer`, then takes the value back out of the box to return.

The simple fix for this involves replacing the `INSTANCEOF` node in question with either a `ICONST_1` or an `ICONST_0`, depending on whether or not the box is legit. Doing that brings up another optimization opportunity: currently `ICONST_1; ISEQ` and similar instruction patterns go untouched. The other part of this commit adds another case to the jump simplifier in `LocalOpt` to notice this situation, and either remove the pair or substitute a `GOTO`, depending on whether the branch would always or never be chosen.

This latter optimization is unlikely to be important, as the JVM has reasonably good branch prediction logic, but it does give us very nice bytecode for the above match:

    ICONST_1
    IRETURN

What more (less) could you ask for?

Review by @lrytz 

Jardiff that: https://gist.github.com/hrhino/8ea27a11444d492b63ccfe1d0154d623


src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -255,6 +255,12 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
             case c: EscapingConsumer =>
               assert(keepBox, s"found escaping consumer, but box is eliminated: $c")
 
+            case BoxedPrimitiveTypeCheck(consumer, success) =>
+              toReplace(consumer) =
+                new InsnNode(POP) ::
```
lrytz: I think we need to distinguish two cases here, if `keepBox` is false, there's nothing on the stack to pop:

```scala
              val loadRes = new InsnNode(if (success) ICONST_1 else ICONST_0)
              if (keepBox) toReplace(consumer) = List(getPop(1), loadRes)
              else toReplace(consumer) = List(loadRes)
```

But we could also leave away the entire `case BoxedPrimitiveTypeCheck` and rely on `postExtractionAdaptationOps` in the case below, no? It would lead to an temporary `xLOAD n; POP` that's then eliminated by push-pop.

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -681,6 +687,9 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
           else if (isPredefAutoUnbox(mi) && typeOK(mi)) BoxKind.checkReceiverPredefLoad(mi, prodCons).map(ModuleGetter(_, mi))
           else None
 
+        case ti: TypeInsnNode if insn.getOpcode == INSTANCEOF =>
+          Some(BoxedPrimitiveTypeCheck(ti, ti.desc == kind.boxClass))
```
lrytz: how about `(1: Integer).isInstanceOf[Object]`? I think we should only catch the success case here (or maybe we can enumerate all types that can possibly succeed?)

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -726,6 +735,9 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
         else if (fi.getOpcode == PUTFIELD) Some(StaticSetterOrInstanceWrite(fi))
         else None
 
+      case ti: TypeInsnNode if ti.getOpcode == INSTANCEOF =>
+        Some(BoxedPrimitiveTypeCheck(ti, ti.desc == kind.refClass))
```
lrytz: same as above

src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala
```diff
@@ -924,6 +924,47 @@ object LocalOptImpls {
...
+      * the GOTO and `l` will be removed by DCE (if it's not jumped into from somewhere else).
+      */
+    def simplifyConstantConditions(instruction: AbstractInsnNode): Boolean = {
+      def replace(jump: JumpInsnNode, success: Boolean): Boolean = {
+        if (success) method.instructions.insert(jump, new JumpInsnNode(GOTO, jump.label))
+        replaceJumpByPop(jump)
+        true
+      }
+
+      (instruction, previousExecutableInstruction(instruction, jumpTargets)) match {
```
lrytz: it seems we don't need to call `previousExecutableInstruction` if `instruction` is not a `ConditionalJump`.

lrytz (commented):
Thanks a lot, these lead to very nice bytecode cleanups!

It would be good to split up the two changes in separate commits.

I'd like to have more tests for the cast elimination, see https://github.com/scala/scala/blob/2.12.x/test/junit/scala/tools/nsc/backend/jvm/opt/MethodLevelOptsTest.scala#L245 for existing. It might be a good idea to move the existing ones into a separate `BoxUnboxTest` file first.

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
hrhino: Yeah, I think there are very few possibilities here (`kind.boxClass`, `Number`, `Comparable`, `Object`, and `Serializable`). But you're right:

```
scala> def f = (1: Integer).isInstanceOf[Object]
f: Boolean

scala> :javap f
/* ... */
  public boolean f();
    descriptor: ()Z
    flags: ACC_PUBLIC
    Code:
      stack=2, locals=1, args_size=1
         0: getstatic     #24                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
         3: iconst_1
         4: invokevirtual #28                 // Method scala/Predef$.int2Integer:(I)Ljava/lang/Integer;
         7: instanceof    #4                  // class java/lang/Object
        10: ireturn
/* ... */
```

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -681,6 +684,10 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
           else if (isPredefAutoUnbox(mi) && typeOK(mi)) BoxKind.checkReceiverPredefLoad(mi, prodCons).map(ModuleGetter(_, mi))
           else None
 
+        case ti: TypeInsnNode if insn.getOpcode == INSTANCEOF =>
+          val success = ti.desc == kind.boxClass || primBoxSupertypes.contains(ti.desc)
```
lrytz: Not entirely correct: `Character` and `Boolean` don't implement` Number`

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -650,6 +650,9 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
       else Type.getReturnType(mi.desc).getInternalName
     }
 
+    private val primBoxSupertypes =
+      Set(coreBTypes.jiSerializableRef, coreBTypes.jlComparableRef, coreBTypes.jlNumberRef, coreBTypes.ObjectRef).map(_.internalName)
```
lrytz: You could get the list of supertypes programmatically through the ClassBType's `info.get.parents / interfaces` and recursing a bit. Should not be too expensive to compute.

test/junit/scala/tools/nsc/backend/jvm/opt/BoxUnboxTest.scala
```diff
@@ -0,0 +1,325 @@
...
+import scala.tools.nsc.backend.jvm.AsmUtils._
+import scala.tools.partest.ASMConverters._
+import scala.tools.testing.BytecodeTesting
+import scala.tools.testing.BytecodeTesting._
+
+/**
+  * Tests for boxing/unboxing optimizations.
+  */
+@RunWith(classOf[JUnit4])
+class BoxUnboxTest extends BytecodeTesting {
```
lrytz, line 19: If it's not too annoying, could you prepend a simple commit that creates this new test file without other changes? This way, the actual changes are easier to spot.

hrhino: Done and done.

src/compiler/scala/tools/nsc/backend/jvm/analysis/ProdConsAnalyzerImpl.scala
```diff
@@ -435,7 +435,7 @@ trait ProdConsAnalyzerImpl {
  * The ASM built-in SourceValue analysis yields an empty producers set for such values. This leads
  * to ambiguities. Example (in Java one can re-assign parameter):
  *
- *   void foo(int a) {
+ *   int foo(int a) {
```
hrhino, line 438: Just scouting out around here, in case anyone decides to compile the java in the scaladoc and gets surprised.

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -650,6 +650,15 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
       else Type.getReturnType(mi.desc).getInternalName
     }
 
+    private val primBoxSupertypes: Map[InternalName, Set[InternalName]] = {
```
lrytz, line 653: the map contains `(java/lang/Void,Set(java/lang/Object, java/lang/Void))`, but it's probably benign

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -681,6 +690,10 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
           else if (isPredefAutoUnbox(mi) && typeOK(mi)) BoxKind.checkReceiverPredefLoad(mi, prodCons).map(ModuleGetter(_, mi))
           else None
 
+        case ti: TypeInsnNode if insn.getOpcode == INSTANCEOF =>
+          val success = ti.desc == kind.boxClass || primBoxSupertypes(kind.boxClass).contains(ti.desc)
```
lrytz: The first test is redundant, as `primBoxSupertypes` contains the `boxClass`.

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
```diff
@@ -713,7 +713,7 @@ class BoxUnbox[BT <: BTypes](val btypes: BT) {
     private def refClass(mi: MethodInsnNode): InternalName = mi.owner
     private def loadZeroValue(refZeroCall: MethodInsnNode): List[AbstractInsnNode] = List(loadZeroForTypeSort(runtimeRefClassBoxedType(refZeroCall.owner).getSort))
 
-    private val refSupertypes =
+    val refSupertypes =
```
lrytz: is this change accidental?

lrytz (commented):
Thanks a lot, almost there I think!

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
hrhino: So it does, now.

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
hrhino: Yeah, I was messing with these in the repl and forgot to switch back.

src/compiler/scala/tools/nsc/backend/jvm/opt/BoxUnbox.scala
hrhino: I can't think of how it would become an issue when compiling scala code.

lrytz (approved):
LGTM now! We'll merge it once 2.12.3 is tagged.

PR #5950, opened 2017-06-18
Implement `tails` method efficiently on List and Stream

hakonhw:
I do not feel this is a complete solution, we are not sure Traversable[A] can be cast to Repr. Which is the case with StringOps.

### DIFF
--- a/src/library/scala/collection/LinearSeqOptimized.scala
+++ b/src/library/scala/collection/LinearSeqOptimized.scala
@@ -315,4 +315,7 @@ trait LinearSeqOptimized[+A, +Repr <: LinearSeqOptimized[A, Repr]] extends Linea
     }
     last
   }
+
+  override /*TraversableLike*/
+  def tails: Iterator[Repr] = Iterator.iterate(repr)(_.tail).takeWhile(_.nonEmpty) ++ Iterator(newBuilder.result)
 }
--- a/test/junit/scala/collection/LinearSeqOptimizedTest.scala
+++ b/test/junit/scala/collection/LinearSeqOptimizedTest.scala
@@ -16,4 +16,20 @@ class LinearSeqOptimizedTest {
     assertEquals(2, "abcde".toList.indexWhere(_ == 'c', -1))
     assertEquals(2, "abcde".toList.indexWhere(_ == 'c', -2))
   }
+
+  @Test def test_efficientTails_list_SI9892: Unit = {
+    val tails = List(1,2,3,4).tails.toList
+
+    assert(tails(0).tail eq tails(1))
+    assert(tails(0).tail.tail eq tails(2))
+    assert(tails(1).tail eq tails(2))
+    assert(tails(3).tail eq tails(4))
+    assert(tails(4) eq List())
+  }
+
+  @Test def test_efficientTails_stream_SI9892: Unit = {
+    val stream = Stream.from(1)
+    val tails = stream.tails.toStream
+    assert(tails.head eq stream)
+  }
 }


hakonhw:
context: ~~https://issues.scala-lang.org/browse/SI-9892~~ https://github.com/scala/bug/issues/9892

SethTisue:
I think a good next step would be to add a failing test that shows why this solution isn't right

hakonhw:
I created the FailedTraversable class and a test to show that "tail" work, but not "tails". I feel the problem here is that there is nothing in place to make sure Repr can be cast to Traversable[A]. I am sure there are good reasons for it, but this means newBuilder is the only way to safely convert between the two types. 

Some suggestions of how to fix it:

1. (I think this one is impossibly out of scope) Constrain Repr with something like this: [+A, +Repr <: Traversable[A @uV]] 

2. Find every corner case as with StringOps and implement tails differently there. Write a ton of tests.

3. Pass implicit conversion from Repr to Traversable[A @uV] in the tails function (i imagine this would break a ton of stuff)


I am beginning to think the cost of fixing this is a bit too high, especially if there is a collection library rewrite planned. What do you think?

szeiger:
I'd put an optimized implementation into `List` since that should be the only collection type that can benefit from it (perhaps `Stream`, too). It's not safe to cast `Repr` to `Traversable`, therefore this should not be the default implementation in `TraversableLike`.

I think the situation in the collection strawman is generally the same. The default implementation there should be lazy but you have the same relationship between `Repl` and `Iterable` and the same optimization can be applied to `List`.

hakonhw:
It sounds smart to add a optimization only where needed. 

How do we know which collections needs to be optimized? I can see that algorithms on List will use tails because of the O(n) complexity when accessing element n, but is this "linkedness" (sorry) defined in any trait? (I saw an earlier pull request that LinearSeq was common to streams and List, is this it?)

I want to be DRY so in case there is no common trait, how can I avoid creating yet another new trait for the collections to inherit?

szeiger:
`LinearSeqOptimized` looks like a good place to add this. It should cover all cases and it also enforces `Repr <: LinearSeqOptimized`.

hakonhw:
Great! Should I do it in a new pull request?

szeiger:
You can add another commit here or `git push -f` a modified state. In the end it needs to be squashed anyway. No need to split up the history of this change by creating another PR.

hakonhw:
Ok, This time the fix is much simpler. I still feel that the original tails is not optimal, but it seems it takes too much effort to fix it, at least for me.

adriaanm:
ping @szeiger

adriaanm:
@Ichoran, would you have time to sign off on this today (otherwise I'll have to push this to 2.12.5)

SethTisue:
in the test code that was added, shouldn't `eq` be used every time, instead of `assertEquals`?

SethTisue:
also nbd but the PR name says bug 9892, but the commit message says "Fixes scala/bug/#10304"... better to reference the original ticket than the duplicate

adriaanm:
I'm sorry we didn't get to reviewing this in time. Let's make sure this gets polished & merged ASAP in 2.12.x and 2.13.x.

hakonhw:
Not quite ASAP but I made the changes, hopefully I got the the assertions correct this time.

SethTisue (approved):
LGTM. final review/merge by @szeiger?

src/library/scala/collection/LinearSeqOptimized.scala
```diff
@@ -315,4 +315,7 @@ trait LinearSeqOptimized[+A, +Repr <: LinearSeqOptimized[A, Repr]] extends Linea
     }
     last
   }
+
+  override /*TraversableLike*/
+  def tails: Iterator[Repr] = Iterator.iterate(repr)(_.tail).takeWhile(_.nonEmpty) ++ Iterator(newBuilder.result)
```
retronym, line 320: The existing implementation if `inits`/`tails` use:

https://github.com/scala/scala/blob/403f910996ebd5cbf327e8cc978d03b906b83ea7/src/library/scala/collection/TraversableLike.scala#L804-L807

I wonder if `Iterator(Nil) map (x => (newBuilder ++= x).result)` in that code could simplified to the version above  (which, although simpler, still took me a little while to recognize as `take(0)` 😄 )

hakonhw: If i remember correctly, it looks like this in TraversableLike because the only "path" from Traversable[A @uV] to Repr is through newBuilder. Traversable is not constrained to Repr. So `Iterator(Nil) map (x => (newBuilder ++= x).result)` is basically a cast to Iterator[Repr] from Iterator[Traversable[A @uV]]

SethTisue:
ping @szeiger

Ichoran:
The algorithm looks good (not the one I'd have chosen, but it looks correct to me).  But how can this be binary compatible?  I guess the expectation is that nobody is going to inherit from `LinearSeqOptimized` save the collections library, so if we get this addition at all, we also get `List` and `Stream` that depends on it?

hakonhw:
Maybe I misunderstand some aspects of binary compatibility, but LinearSeqOptimized already inherits from TraversableLike, so this addition of tails won’t change its interface. Also LinearSeqOptimized has constrained type Repr to LinearSeqOptimized, which solves the problem in TraversableLike where Repr was only «connected» to TraversableLike through newBuilder. However the behavior will change to not returning copies, but that was the issue itself. List and Stream already inherits from LinearSeqOptimized so this change will also affect them the same way, I wrote tests for both List and Stream to verify that they solve the issue :)

szeiger (approved):
LGTM. Adding or removing overrides is binary compatible in general. We often run into problems with collections where the return type is `Repr` because that gets refined and therefore erases to a different type when you override. This is not a problem here because `Iterator[Repr]` always erases to `Iterator`.

SethTisue:
thanks @hodga, glad we finally got this across the finish line, and in time for 2.12.5, too! 👏

PR #5951, opened 2017-06-19
Avoid diverging implicit scope search

szeiger:
Updated version of #5742. Filtering out empty infos & targeting 2.13.

Storing empty infos into the infoMap (thus avoiding recomputation)
prevents the endless recursion.

This is an alternative to https://github.com/scala/scala/pull/5734 that
avoids the expensive annotation removal and the regression in
`pos/t1203a.scala`.

Fixes scala/bug#10081


szeiger:
@retronym Any objections to merging this?

retronym (approved):

milessabin:
I think this might be responsible for the regression relative to 2.12.2 reported in https://github.com/scala/bug/issues/10425.

PR #5952, opened 2017-06-20
Invalidate symbols for artifact classfiles, refactor classfile parser

lrytz:
No longer run the classfile parser on Scala generated classfiles that
don't have a Scala signature (module classes, inner classes, etc).

Various cleanups in the classfile parser, minimize the work performed
on Scala classfiles. Before, the attributes section was parsed twice:
once to find the ScalaSig attribute, the second time to find the
ScalaSignature in the RuntimeVisibleAnnotations. Now everything happens
in the first iteration.

Also fixes a bug in the backend: classes ending in `$` did not get a
ScalaSignature by mistake. They were filtered out by the name-based
test that is supposed to identify module classes.

Also cleans up an unnecessary abstraction in the representation of
ScalaSignature annotations.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -501,7 +501,7 @@ abstract class BCodeHelpers extends BCodeIdiomatic with BytecodeWriters {
      */
     def getAnnotPickle(jclassName: String, sym: Symbol): Option[AnnotationInfo] = {
       currentRun.symData get sym match {
-        case Some(pickle) if !nme.isModuleName(newTermName(jclassName)) =>
+        case Some(pickle) if !sym.isModuleClass => // pickles for module classes are in the companion / mirror class
           val scalaAnnot = {
             val sigBytes = ScalaSigBytes(pickle.bytes.take(pickle.writeIndex))
             AnnotationInfo(sigBytes.sigAnnot, Nil, (nme.bytes, sigBytes) :: Nil)
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
@@ -12,6 +12,7 @@ import scala.tools.nsc.io.AbstractFile
 import GenBCode._
 import BackendReporting._
 import scala.reflect.internal.Flags
+import scala.tools.nsc.backend.jvm.BCodeHelpers.ScalaSigBytes
 import scala.tools.nsc.reporters.NoReporter
 
 /*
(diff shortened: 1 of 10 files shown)


retronym (approved):
Such an improvement! LGTM.

We should consider backporting the "no scala sig for classes ending in '$'" fix, if it is a regression.

lrytz:
Good point, I'll submit a backport for this fix.

Jardiff (lib+reflect+comp) looks good:

```
➜  sandbox git:(c8b0e1da51) jardiff a b
WARN: unable to invoke scalap on: a/scala/concurrent/duration/package.class: Unexpected failure
WARN: unable to invoke scalap on: b/scala/concurrent/duration/package.class: Unexpected failure
diff --git a/scala/runtime/Nothing$.class.scalap b/scala/runtime/Nothing$.class.scalap
new file mode 100644
index 0000000..367d133
--- /dev/null
+++ b/scala/runtime/Nothing$.class.scalap
@@ -0,0 +1,4 @@
+package scala.runtime
+sealed abstract class Nothing$ extends scala.Throwable {
+  def this() = { /* compiled code */ }
+}
diff --git a/scala/runtime/Null$.class.scalap b/scala/runtime/Null$.class.scalap
new file mode 100644
index 0000000..2a0e3a4
--- /dev/null
+++ b/scala/runtime/Null$.class.scalap
@@ -0,0 +1,3 @@
+package scala.runtime
+sealed abstract class Null$ extends scala.AnyRef {
+}
```

When compiling the compiler, a number of scala-artifact-classfiles are being completed. They fall in two categories. When completed, they are now unlinked and marked non-existing:

`scala.annotation.meta.package$`: parsed during typer when computing
  `lazy val metaAnnotations: Set[Symbol] = getPackage(TermName("scala.annotation.meta")).info.members filter (_ isSubClass StaticAnnotationClass) toSet`

`scala.runtime.AbstractFunction2$mcZDD$sp` and many more specialized classfiles. They are completed as part of the fix of https://github.com/scala/bug/issues/5545, https://github.com/scala/scala/commit/c5f68c11d3d8b9f12fe8608d7cea5d33c260c793. We could revert that fix: we don't parse the classfile anymore, so the crash with incompatible signatures doesn't happen. But the symbols would still get completed, namely during the backend (as described in the bug report).

I'll leave the fix in place. In general, when a new class symbol is created synthetically with a name that matches a classfile that already exists on the classpath, it's probably a good idea to unlink the existing symbol (by completing it) and its companionModule/companionModuleClass.

retronym:
I've seen some code in the wild that is broken by this change.

`x.isInstanceOf[Enumeration$Val]` / `classOf[Enumeration$Val]` will no longer work. `Class.forName("scala.Enumeration$Val")`  is an alternative.

lrytz:
Ah, as in `x.getClass isAssignableFrom Class.forName("scala.Enumeration$Val")`. I guess that's ok.

PR #5956, opened 2017-06-22
Optimize classpath implementation to speed up SBT

retronym:
SBT is a heavy user of `Classpath.findClass`, and its performance regressed, especially for large classpaths, under the new "flat" classpath implementation. The new implementation shipped in 2.11 experimentally (`-YclasspathImpl:flat`), and was promoted to the default an only implementation in 2.12.

In 2.12, the performance was improved some in 0f9a704a71 / #5112. This commit goes further by optimizing the lookup of a given class in a JAR classpath entry, and, most importantly, by avoiding a linear scan of the classpath on each lookup by indexing which classpath entries contribute classes to each package.

I tested a [2.11 backport](https://github.com/retronym/scala/tree/backport/fasterFindClass) of this patch using https://github.com/guardian/frontend as a test case.

Results: https://gist.github.com/3c89c21c54d27e566dec57a16a4bf208

tl;dr; No significant difference in performance between the classpath implementations after this change. The baseline performance was 55s, which regressed to 79s under `-YclasspathImpl:flat` and then progressed back to 55s using the same flat classpath implementation as in this PR.

I've also tried to [patch SBT](https://github.com/sbt/sbt/compare/0.13...retronym:faster-flatclasspath?expand=1) to use `Symbol.associatedFile` rather than scanning the classpath, but I haven't got that change into shape yet.

Fixes scala/bug#10289

### DIFF
--- a/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
+++ b/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
@@ -21,32 +21,29 @@ import scala.tools.nsc.util.ClassRepresentation
  */
 case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
   override def findClassFile(className: String): Option[AbstractFile] = {
-    @tailrec
-    def find(aggregates: Seq[ClassPath]): Option[AbstractFile] =
-      if (aggregates.nonEmpty) {
-        val classFile = aggregates.head.findClassFile(className)
-        if (classFile.isDefined) classFile
-        else find(aggregates.tail)
-      } else None
-
-    find(aggregates)
+    val (pkg, simpleClassName) = PackageNameUtils.separatePkgAndClassNames(className)
+    aggregatesForPackage(pkg).iterator.map(_.findClassFile(className)).collectFirst {
+      case Some(x) => x
+    }
+  }
+  private[this] val packageIndex: collection.mutable.Map[String, Seq[ClassPath]] = collection.mutable.Map()
+  private def aggregatesForPackage(pkg: String): Seq[ClassPath] = packageIndex.synchronized {
+    packageIndex.getOrElseUpdate(pkg, aggregates.filter(_.hasPackage(pkg)))
   }
 
+  // This method is performance sensitive as it is used by SBT's ExtractDependencies phase.
   override def findClass(className: String): Option[ClassRepresentation] = {
-    @tailrec
-    def findEntry(aggregates: Seq[ClassPath], isSource: Boolean): Option[ClassRepresentation] =
-      if (aggregates.nonEmpty) {
-        val entry = aggregates.head.findClass(className) match {
-          case s @ Some(_: SourceFileEntry) if isSource => s
-          case s @ Some(_: ClassFileEntry) if !isSource => s
-          case _ => None
-        }
-        if (entry.isDefined) entry
-        else findEntry(aggregates.tail, isSource)
-      } else None
+    val (pkg, simpleClassName) = PackageNameUtils.separatePkgAndClassNames(className)
 
-    val classEntry = findEntry(aggregates, isSource = false)
-    val sourceEntry = findEntry(aggregates, isSource = true)
+    def findEntry(isSource: Boolean): Option[ClassRepresentation] = {
+      aggregatesForPackage(pkg).iterator.map(_.findClass(className)).collectFirst {
+        case Some(s: SourceFileEntry) if isSource => s
+        case Some(s: ClassFileEntry) if !isSource => s
+      }
+    }
+
+    val classEntry = findEntry(isSource = false)
+    val sourceEntry = findEntry(isSource = true)
 
     (classEntry, sourceEntry) match {
       case (Some(c: ClassFileEntry), Some(s: SourceFileEntry)) => Some(ClassAndSourceFilesEntry(c.file, s.file))
@@ -72,6 +69,7 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
   override private[nsc] def sources(inPackage: String): Seq[SourceFileEntry] =
     getDistinctEntries(_.sources(inPackage))
 
+  override private[nsc] def hasPackage(pkg: String) = aggregates.exists(_.hasPackage(pkg))
   override private[nsc] def list(inPackage: String): ClassPathEntries = {
     val (packages, classesAndSources) = aggregates.map { cp =>
       try {
--- a/src/compiler/scala/tools/nsc/util/ClassPath.scala
+++ b/src/compiler/scala/tools/nsc/util/ClassPath.scala
@@ -23,6 +23,7 @@ trait ClassPath {
   def asURLs: Seq[URL]
 
   /** Empty string represents root package */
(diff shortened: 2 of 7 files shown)


dragos:
LGTM! 

Perhaps not directly related, but still in this area of the code: we need to provide a classpath implementation in our codebase and it would be very helpful to widen the visibility of some classes/methods. They don't seem to be consistently `private[nsc]` anyway. It's mostly the `*FileEntryImpl` classes, that even though are implementing a trait, are used in many places as return type or [type argument](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala#L199).

If that has chances to be merged I'm happy to open a PR.

src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
```diff
@@ -21,32 +21,29 @@ import scala.tools.nsc.util.ClassRepresentation
...
+
+    def findEntry(isSource: Boolean): Option[ClassRepresentation] = {
+      aggregatesForPackage(pkg).iterator.map(_.findClass(className)).collectFirst {
+        case Some(s: SourceFileEntry) if isSource => s
+        case Some(s: ClassFileEntry) if !isSource => s
+      }
+    }
+
+    val classEntry = findEntry(isSource = false)
+    val sourceEntry = findEntry(isSource = true)
```
lrytz, line 46: You could search for both entries in a single traversal, but it's probably not going to make a difference, as there are few aggregates for a given package.

lrytz (approved):
👍

lrytz:
> widen the visibility of some classes/methods

I think that should be fine.

retronym:
@dragos I'm open to opening up access in those cases.

I think some of the cases you pointed out can be cleaned up by referring to the public types, as in https://github.com/retronym/scala/tree/topic/refactorClassPath

smarter:
> I've also tried to patch SBT to use Symbol.associatedFile rather than scanning the classpath, but I haven't got that change into shape yet.

Interestingly, the dotty  sbt phases do not use `findClass` and were already using `associatedFile`, though now I'm wondering if that means we're tracking something incorrectly: https://github.com/lampepfl/dotty/blob/1d0579622da87655c6d32c4e02e7907e75a4f7b5/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala#L89-L90

retronym:
We should give this a spin through the [community build](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1616) before merging.

> Interestingly, the dotty sbt phases do not use findClass and were already using associatedFile, though now I'm wondering if that means we're tracking something incorrectly: 

@smarter My patch seemed to have the correct behaviour, but didn't seem to improve performance in my test as I had expected.

smarter:
> @smarter My patch seemed to have the correct behaviour, but didn't seem to improve performance in my test as I had expected.

So `findClass` is not the culprit?

MikeSkells:
Hi @retronym as discussed directly, I have some work in this area. Happy to share

I have a classpath cache that can use a file watcher to detect changes

This is showing a 20% reduction in memory allocation and around 10% in CPU

I think that my chnages should work well with yours - I have not been changing the AggregateClassPath, but focussed on the other structures.
I will email you directly to discuss

raw figures
```
 
                  RunName                        AllWallMS                             CPU_MS                         Allocated
       00_linker_baseline  510        8,592.24 [-2.68% +2.96%]            8,467.01 [-2.38% +2.97%]            2,588.77 [-0.04% +0.04%]
        00_linker_cp_raw2 529        7,761.38 [-6.41% +9.11%]            7,558.16 [-6.56% +8.12%]            2,124.32 [-0.06% +0.06%]
```

src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
mkeskells: done the in the BasicClassPath of https://github.com/scala/scala/pull/5957

retronym:
> So findClass is not the culprit?

The best theory is that I've made some silly error in my attempt to modify SBT and/or my attempt to measure the difference. My gut feel is still that switching to `associatedFile` should also restore previous performance.

retronym:
Rebased and made a minor reversion of a change of signature to quiet MiMa, and tried again to schedule a community build at https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1621/

retronym:
Testing this locally in the Akka SBT build has flushed out a problem.

```
warn]                            ^
[error] /Users/jz/code/akka/akka-remote/src/main/java/akka/remote/WireFormats.java:3088: Class com.google.protobuf.InvalidProtocolBufferException not found - continuing with a stub.
[error]     boolean hasSerializerId();
[error]      ^
[error] Unable to locate class corresponding to inner class entry for Descriptor in owner com.google.protobuf.Descriptors
[error] Unable to locate class corresponding to inner class entry for Descriptor in owner com.google.protobuf.Descriptors
[error] /Users/jz/code/akka/akka-remote/src/main/java/akka/remote/WireFormats.java:6: Class com.google.protobuf.ByteString not found - continuing with a stub.
[error] public final class WireFormats {
[error]                    ^
[error] Unable to locate class corresponding to inner class entry for Descriptor in owner com.google.protobuf.Descriptors
[error] /Users/jz/code/akka/akka-remote/src/main/java/akka/remote/WireFormats.java:2385: Class com.google.protobuf.InvalidProtocolBufferException not found - continuing with a stub.
[error]       public akka.remote.WireFormats.AcknowledgementInfo getDefaultInstanceForType() {
[error]              ^
```

retronym:
> Testing this locally in the Akka SBT build has flushed out a problem.

Sigh, it slipped through my fingers and I can't reproduce now. Will see how the community build fares and try to retrace my steps locally in the Akka build. Marked as WIP so we don't merge just yet.

src/compiler/scala/tools/nsc/util/ClassPath.scala
```diff
@@ -22,18 +22,43 @@ trait ClassPath {
...
+   * Returns the class file and / or source file for a given external name, e.g., "java.lang.String".
+   * If there is both a class file and source file, the compiler can decide whether to read the
+   * class file or compile the source file.
+   *
+   * Internally this seems to be used only by `ScriptRunner`, but only to call `.isDefined`. That
+   * could probably be implemented differently.
+   *
+   * Externally, it is used by sbt's compiler interface:
+   * https://github.com/sbt/sbt/blob/v0.13.15/compile/interface/src/main/scala/xsbt/CompilerInterface.scala#L249
+   * Jason has some improvements for that in the works (https://github.com/scala/bug/issues/10289#issuecomment-310022699)
```
lrytz, line 60: could remove parts of this comment

retronym:
> Testing this locally in the Akka SBT build has flushed out a problem.

That turned out to be unrelated to this PR, and has been fixed separately in #5976.

jvican:
> @smarter My patch seemed to have the correct behaviour, but didn't seem to improve performance in my test as I had expected.
  
Zinc 1.0 already checks for `associatedFile` and @romanowski (the author of the commit that introduced its use) experienced speedups in the codebase he worked on.

PR #5960, opened 2017-06-26
User-defined case companion apply: unlink default getters

lrytz:
When a synthetic case companion `apply` method is unlinked because
there's a matching user-defined method (see PR 5730), we also have to
unlink the default getters to avoid clashes.

Fixes scala/bug#10389

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -678,7 +678,14 @@ trait Namers extends MethodSynthesis {
             // which could upset other code paths)
             if (!scopePartiallyCompleted)
               companionContext.scope.unlink(sym)
+
+            for (a <- sym.attachments.get[CaseApplyDefaultGetters]; defaultGetter <- a.defaultGetters) {
+              companionContext.unit.synthetics -= defaultGetter
+              companionContext.scope.unlink(defaultGetter)
+            }
           }
+
+          sym.removeAttachment[CaseApplyDefaultGetters] // no longer needed once the completer is done
         }
       }
 
@@ -1544,6 +1551,14 @@ trait Namers extends MethodSynthesis {
             if (!isConstr)
               methOwner.resetFlag(INTERFACE) // there's a concrete member now
             val default = parentNamer.enterSyntheticSym(defaultTree)
+            if (meth.name == nme.apply && meth.hasAllFlags(CASE | SYNTHETIC)) {
+              val att = meth.attachments.get[CaseApplyDefaultGetters].getOrElse({
+                val a = new CaseApplyDefaultGetters()
+                meth.updateAttachment(a)
+                a
+              })
+              att.defaultGetters += default
+            }
             if (default.owner.isTerm)
               saveDefaultGetter(meth, default)
           }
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -678,7 +678,14 @@ trait Namers extends MethodSynthesis {
...
             if (!scopePartiallyCompleted)
               companionContext.scope.unlink(sym)
+
+            for (a <- sym.attachments.get[CaseApplyDefaultGetters]; defaultGetter <- a.defaultGetters) {
+              companionContext.unit.synthetics -= defaultGetter
+              companionContext.scope.unlink(defaultGetter)
+            }
           }
+
+          sym.removeAttachment[CaseApplyDefaultGetters]
```
adriaanm: It's not clear to me why we remove the attachment? I guess it's no longer needed, but we could start using it for other purposes later, so it would be good to add a comment to make it clear it doesn't *have* to be removed for correctness.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -1544,6 +1551,14 @@ trait Namers extends MethodSynthesis {
             if (!isConstr)
               methOwner.resetFlag(INTERFACE) // there's a concrete member now
             val default = parentNamer.enterSyntheticSym(defaultTree)
+            if (meth.name == nme.apply && meth.hasFlag(CASE)) {
```
adriaanm: Technically, right now we could restrict it to methods that also have the `SYNTHETIC` flag, right?

adriaanm: Oh, right, the `CASE` flag is not set for user-defined method either.

src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala
```diff
@@ -29,6 +29,12 @@ trait NamesDefaults { self: Analyzer =>
   // as an attachment in the companion module symbol
   class ConstructorDefaultsAttachment(val classWithDefault: ClassDef, var companionModuleClassNamer: Namer)
 
+  // Attached to the synthetic companion `apply` method symbol generated for case classes, holds
+  // the set contains all default getters for that method. If the synthetic `apply` is unlinked in
+  // its completer because there's a user-defined matching method (PR #5730), we have to unlink the
+  // default getters as well.
+  class CaseApplyDefaultGetters(val defaultGetters: mutable.Set[Symbol] = mutable.Set.empty)
```
adriaanm: Add a quick note on the (current) lifetime of the attachment? Similar to my comment at `sym.removeAttachment`

adriaanm (changes requested):
Looks good! Small request for clarification.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
lrytz: Added `SYNTHETIC`

src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala
lrytz: Done!

src/compiler/scala/tools/nsc/typechecker/Namers.scala
lrytz: 👍

adriaanm (approved):

PR #5936, opened 2017-06-10
Drive bitmap computation from field, not accessor

adriaanm:
Since bitmaps are only needed if there is a field, and the field carries
the transient annotation, don't go around via the accessor (we'd miss the
annotation).

Before we had a fields phase, this was tricky to do, but now it's
pretty straightforward to operate on field symbols, as we synthesize
the relevant ones, along with their bitmaps, during the fields phase
instead of waiting until mixin. As an extra bonus, the "artifact" fields
such as outer pointers etc don't exist yet, so we don't have to
exclude them.
 
With thanks to @sarahgerweck for the analysis!

TODO:
  - [x] jardiff

Fix scala/bug#10244

PR #5966, opened 2017-06-28
Fix regression under -Xcheckinit: param accessors don't need init checking

retronym:
In #5936, some checks were removed that resulted in checkinit machinery
being synthesized for param accessor fields (ie, constructor vals).

This commit dials this back to the old behaviour, avoiding unnecessary
fields and code, and an unwanted VerifyError.

I've added a basic test case to show that checkinit errors are still
raised when we want them to be, and that the verify error is gone.

Fixes scala/community-builds#566

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
+++ b/src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
@@ -140,7 +140,7 @@ trait AccessorSynthesis extends Transform with ast.TreeDSL {
 
         if (field.isLazy)
           if (field hasAnnotation TransientAttr) BITMAP_TRANSIENT else BITMAP_NORMAL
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/transform/AccessorSynthesis.scala
```diff
@@ -140,7 +140,7 @@ trait AccessorSynthesis extends Transform with ast.TreeDSL {
 
         if (field.isLazy)
           if (field hasAnnotation TransientAttr) BITMAP_TRANSIENT else BITMAP_NORMAL
-        else if (doCheckInit && !(field hasFlag DEFAULTINIT | PRESUPER))
+        else if (doCheckInit && !(field hasFlag DEFAULTINIT | PRESUPER | PARAMACCESSOR))
```
adriaanm, line 143: 👍

adriaanm (approved):
Thanks, @retronym!

PR #5968, opened 2017-06-30
Fix 5355: wrong error message for cyclic structural types

Jasper-M:
The pattern match in `TypeDiagnostics` also matched methods in structural types that do have a return type.
Strictly match on vals and defs with an empty `TypeTree` as return type. Show the default error message otherwise.

Fixes scala/bug#5355

I could add an extra case to add a custom error message for structural types, but I think the default error message is sufficient. The information about which reference exactly is cyclic seems to have been thrown away anyway, except for the position.


lrytz (approved):

lrytz:
Thanks, @Jasper-M

PR #5969, opened 2017-06-30
Proper laziness for by-name args of right-associative operators

### LINKED ISSUE #1980: by-name argument incorrectly evaluated on :-ending operator
```scala
scala> def foo() = { println("foo") ; 5 }
foo: ()Int

scala> class C { def m1(f: => Int) = () ; def m2_:(f: => Int) = () }
defined class C

scala> val c = new C
c: C = C@96d484

scala> c m1 foo()

scala> foo() m2_: c
foo
```
But it is not evaluated if invoked directly:
```scala
scala> c.m2_:(foo())

scala>
```

szeiger:
This fixes scala/bug#1980 by changing the
desugaring of right-associative operator syntax in the spec such that
by-name operands now get the same desugaring as left-associative
operators (except for the reversed operands, of course). Only by-value
operands are pulled out into intermediate vals to preserve their
left-to-right evaluation order.

The (revised) implementation is as follows:

- `Parsers` still performs the `val` desugaring for all calls, except
  that the generated synthetic names use the new `RIGHT_ASSOC_OP_PREFIX`
  to identify them later.

- Everything else happens in `Typers`: After typechecking a ValDef
  resulting from desugaring of a right-associative operator its Symbol
  and RHS are stored in a Map (without knowing at that point if they are
  by-name or by-value).

- After typechecking a method application with an Ident for one of these
  Symbols, check if the parameter is by-name in which case it is
  replaced with the RHS and the Symbol added to a Set.

- After typechecking a Block, check for a leading ValDef with the
  Symbol that was inlined and remove the ValDef.

Fixes scala/bug#1980

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -45,6 +45,11 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
   final val shortenImports = false
 
+  // All typechecked RHS of ValDefs for right-associative operator desugaring
+  private val rightAssocValDefs = new mutable.AnyRefMap[Symbol, Tree]
+  // Symbols of ValDefs for right-associative operator desugaring which are passed by name and have been inlined
+  private val inlinedRightAssocValDefs = new mutable.HashSet[Symbol]
+
   // allows override of the behavior of the resetTyper method w.r.t comments
   def resetDocComments() = clearDocComments()
 
@@ -53,6 +58,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
     resetContexts()
     resetImplicits()
     resetDocComments()
+    rightAssocValDefs.clear()
+    inlinedRightAssocValDefs.clear()
   }
 
   sealed abstract class SilentResult[+T] {
@@ -2067,7 +2074,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           } else tpt1.tpe
           transformedOrTyped(vdef.rhs, EXPRmode | BYVALmode, tpt2)
         }
-      treeCopy.ValDef(vdef, typedMods, sym.name, tpt1, checkDead(rhs1)) setType NoType
+      val vdef1 = treeCopy.ValDef(vdef, typedMods, sym.name, tpt1, checkDead(rhs1)) setType NoType
+      if (sym.isSynthetic && sym.name.startsWith(nme.RIGHT_ASSOC_OP_PREFIX))
+        rightAssocValDefs += ((sym, vdef1.rhs))
+      vdef1
     }
 
     /** Enter all aliases of local parameter accessors.
@@ -2478,7 +2488,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           if (result0.nonEmpty) checkPure(result0, supple = true)
         }
 
-        treeCopy.Block(block, statsTyped, expr1)
+        // Remove ValDef for right-associative by-value operator desugaring which has been inlined into expr1
+        val statsTyped2 = statsTyped match {
+          case (vd: ValDef) :: Nil if inlinedRightAssocValDefs remove vd.symbol => Nil
+          case _ => statsTyped
+        }
+
+        treeCopy.Block(block, statsTyped2, expr1)
           .setType(if (treeInfo.isExprSafeToInline(block)) expr1.tpe else expr1.tpe.deconst)
       } finally {
         // enable escaping privates checking from the outside and recycle
@@ -3611,6 +3627,15 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                   case _ => tp
                 }
 
+                // Inline RHS of ValDef for right-associative by-value operator desugaring
+                val args2 = (args1, mt.params) match {
+                  case ((ident: Ident) :: Nil, param :: Nil) if param.isByNameParam && rightAssocValDefs.contains(ident.symbol) =>
+                    inlinedRightAssocValDefs += ident.symbol
+                    val rhs = rightAssocValDefs.remove(ident.symbol).get
+                    rhs.changeOwner(ident.symbol -> context.owner) :: Nil
+                  case _ => args1
+                }
+
                 /*
                  * This is translating uses of List() into Nil.  This is less
                  *  than ideal from a consistency standpoint, but it shouldn't be
@@ -3622,7 +3647,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                 if (args.isEmpty && canTranslateEmptyListToNil && fun.symbol.isInitialized && ListModule.hasCompleteInfo && (fun.symbol == List_apply))
                   atPos(tree.pos)(gen.mkNil setType restpe)
                 else
-                  constfold(treeCopy.Apply(tree, fun, args1) setType ifPatternSkipFormals(restpe))
+                  constfold(treeCopy.Apply(tree, fun, args2) setType ifPatternSkipFormals(restpe))
               }
               checkDead.updateExpr(fun) {
                 handleMonomorphicCall
--- a/test/files/run/t1980.scala
+++ b/test/files/run/t1980.scala
@@ -1,0 +1,75 @@
+class LazyList[+A](expr: => LazyList.Evaluated[A]) {
+  def #:: [B >: A](elem: => B): LazyList[B] = new LazyList(Some((elem, this)))
+  def ##:: [B >: A](elem: B): LazyList[B] = new LazyList(Some((elem, this)))
+  def force: Unit = expr.foreach(_._2.force)
+}
+
+object LazyList {
+  type Evaluated[+A] = Option[(A, LazyList[A])]
+  object Empty extends LazyList[Nothing](None)
+  def empty[A]: LazyList[A] = Empty
+}
+
+object Test extends App {
+  def foo(i: Int) = { println("foo "+i); i }
+  def bar(i: Int) = { println("bar "+i); i }
+  println("1. defining")
+  val xs1 = foo(1) ##:: foo(2) ##:: foo(3) ##:: LazyList.empty
+  val xs2 = bar(1) #:: bar(2) #:: bar(3) #:: LazyList.empty
+  println("1. forcing")
+  xs1.force
+  xs2.force
+
+  {
+    println("2. defining")
+    class C { def f_:(x: => Int)(implicit y: Int): () => Int = (() => x) }
+    val c = new C
+    implicit val i = 1
+    def k = { println("hi"); 1 }
+    val x1 = c.f_:(k)
+    val x2 = k f_: c
+    println("2. forcing")
+    println(x1())
+    println(x2())
+  }
+
+  {
+    println("3. defining")
+    class C { def f_:[T](x: => T): () => T = (() => x) }
+    val c = new C
+    def k = { println("hi"); 1 }
+    val x1 = k f_:[Any] c
+    println("3. forcing")
+    println(x1())
+  }
+
+  // Ensure the owner chain is changed correctly when inlining
+  {
+    println("4. defining")
+    class C { def f_:(x: => Int): () => Int = (() => x) }
+    val c = new C
+    val x1 = c.f_:({ val xxx = 1; xxx })
+    val x2 = { val yyy = 1; yyy } f_: c
+    val x3 = { val yyy = 1; yyy } f_: c
+    println("4. forcing")
+    println(x1())
+    println(x2())
+    println(x3())
+  }
+
+  // Overloaded operator with by-name and by-value variants
+  {
+    println("5. defining")
+    class C {
+      val saved = new collection.mutable.ArrayBuffer[() => String]
+      def force: Unit = saved.foreach(_.apply())
+      def :: (x: Int): C = this
+      def :: (x: => String): C = { saved += (() => x); this }
+    }
+    def genI(i: Int): Int = { println("Int "+i); i }
+    def genS(s: String): String = { println("String "+s); s }
+    val c = genI(1) :: genS("2") :: genI(3) :: genS("4") :: (new C)
+    println("5. forcing")
+    c.force
+  }
+}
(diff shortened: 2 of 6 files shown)


lrytz:
Additional test case that could be added (works correctly):

```
scala> class C { def f_:(x: => Int)(implicit y: Int) = 0 }
scala> val c = new C
scala> implicit val i = 1
scala> def k = { println("hi"); 1 }

scala> c.f_:(k)
res0: Int = 0

scala> k f_: c
res1: Int = 0
```

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1703,7 +1707,7 @@ abstract class RefChecks extends Transform {
               assert(sym != NoSymbol, "transformCaseApply: name = " + name.debugString + " tree = " + tree + " / " + tree.getClass) //debug
               enterReference(tree.pos, sym)
             }
-            tree
+            eliminatedRightAssocValDefs.getOrElse(tree.symbol, tree)
```
lrytz: In general we cannot just move trees around like this, because the owner chain of symbols defined in the tree might mess up. This is maybe what's causing the following crash:
```
scala> class C { def f_:(x: => Int) = 0 }
defined class C

scala> val c = new C
c: C = C@4dd94931

scala> { val x = 1; x } f_: c
java.util.NoSuchElementException: key not found: value x
	at scala.collection.MapLike.default(MapLike.scala:230)
	at scala.collection.MapLike.default$(MapLike.scala:229)
	at scala.collection.AbstractMap.default(Map.scala:59)
	at scala.collection.mutable.HashMap.apply(HashMap.scala:61)
	at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder$locals$.load(BCodeSkelBuilder.scala:391)
```

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1714,6 +1718,9 @@ abstract class RefChecks extends Transform {
                            // probably not, until we allow parameterised extractors
             tree
 
+          case Block((vd: ValDef) :: Nil, expr) if vd.symbol.isLazy && vd.name.toString.startsWith(nme.RIGHT_ASSOC_OP_PREFIX) =>
+            eliminatedRightAssocValDefs += ((vd.symbol, vd.rhs))
```
lrytz: Missing call to `transform` on `vd.rhs`?

lrytz (changes requested):
I like the idea, it's nice to see this being fixed! Could we do the transformation in typers? Why is there the intermediate step of marking the symbol lazy?

lrytz:
Yet another test case (that works correctly):

```
scala> class C { def f_:[T](x: => T) = 0 }
scala> val c = new C
scala> def k = { println("hi"); 1 }

scala> k f_:[Any] c
res9: Int = 0
```

szeiger:
There are two steps because we first have to typecheck the application before we know that the definition can be removed. AFAICT typers only does a single full transformation of the tree. Could be done in `typedBlock` on the way up?

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
szeiger: Yes, I think there should be one

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
szeiger: Hm, I assumed it would be safe here because I'm only moving it within the same block. But maybe that's no longer true in all cases after uncurry?

lrytz:
> Could be done in typedBlock on the way up

That's what I was thinking, it might work that way. Also, instead of "abusing" the `LAZY` flag, we could use a symbol attachment.

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
szeiger: Wait, I'm not moving it within the same block. I'm *eliminating* the block.

szeiger: OK, I see. It's owned by the `ValDef` that gets eliminated, so I have to change it anyway, even within the same block.

szeiger:
Looks like the `typedBlock` approach works. I also added your additional tests and fixed the owner chain problem.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2474,7 +2484,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
         }
 
-        treeCopy.Block(block, statsTyped, expr1)
+        // Remove ValDef for right-associative by-value operator desugaring which has been inlined into expr1
+        val statsTyped2 = statsTyped match {
+          case (vd: ValDef) :: Nil if inlinedRightAssocValDefs contains vd.symbol => Nil
+          case _ => statsTyped
+        }
+
+        treeCopy.Block(block, statsTyped2, expr1)
```
lrytz: This leaves a single-expression block in place, but I guess that's fine. We'd have to handle it outside `typedBlock`, as this method returns a `Block`.

```
scala> def foo = { 1 f_: c; 2 }
[[syntax trees at end of                     typer]] // <console>

        def foo: Int = {
          {
            $line5.$read.$iw.$iw.c.f_:(1)
          };
          2
        }
```

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3591,6 +3607,15 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                   case _ => tp
                 }
 
+                // Inline RHS of ValDef for right-associative by-value operator desugaring
+                val args2 = (args1, mt.params) match {
+                  case ((ident: Ident) :: Nil, param :: Nil) if param.isByNameParam && rightAssocValDefs.contains(ident.symbol) =>
+                    inlinedRightAssocValDefs += ident.symbol
+                    val rhs = rightAssocValDefs(ident.symbol)
```
lrytz: I'd prefer a symbol attachment over the two collections. Something like `class RightAssocValDefAttachment(var rhsInlined)`, added in `typedValDef`, the var could be set to `true` here. Then you can use `getAndRemoveAttachment` in `typedBlock`.

We could even just add an empty marker attachment `RightAssocValDefInlined` here, and skip the test in `typedValDef` - what value does it add?

test/files/run/t1980.scala
```diff
@@ -0,0 +1,43 @@
...
+  xs1.force
+  xs2.force
+
+  {
+    class C { def f_:(x: => Int)(implicit y: Int) = 0 }
+    val c = new C
+    implicit val i = 1
+    def k = { println("hi"); 1 }
+    c.f_:(k)
+    k f_: c
```
lrytz: maybe also test lazy evaluation here (and in all tests below), just to make sure.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
szeiger: > I'd prefer a symbol attachment over the two collections.

Are you sure? I checked for other uses of attachments and they all seem to be for communication between phases. I didn't find any precedent for data of a single phase being stored in attachments.

> skip the test in typedValDef - what value does it add?

It stores the RHS which is needed in doTypedApply. Is there a better way to get it?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
szeiger: `typedBlock` could be changed to return a `Tree` but I didn't want to add unnecessary complications. The empty blocks don't seem to cause any problem.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
lrytz: > Are you sure?

To me it feels more like keeping state local, but it's fine either way in the end.



> It stores the RHS which is needed in doTypedApply

Right, of course, I missed that.

odersky:
This is very promising, but I think the desugaring needs to go to defs instead of lazy vals. If

    xs: { def &: (x: => T): U }
  
then, logically

    e &: xs

should be the same as 

    xs.&:(e)

But it is only if it is expanded to

    def x$ = xs; e &: (x$)

Or, the spec could simply demand that by-name arguments are not lifted out. That would be even clearer.

szeiger:
> Or, the spec could simply demand that by-name arguments are not lifted out. That would be even clearer.

But that's exactly what it does in this PR.

lrytz:
Alternative pattern, suggested in peer-reviewing with Jason
  - add an attachment to the outer block (containing the rassoc$ val-def) during parsing
  - catch that in `typedBlock`, type-check the function part of the invocation, if it's by-name, inline the ValDef rhs before even typing it

That way we could live without the hash map / set altogether.

szeiger:
> catch that in `typedBlock`, type-check the function part of the invocation, if it's by-name, inline the ValDef rhs before even typing it

You can't check if it's by-name before typing the RHS because the method could have mixed by-name and by-value overloads. You need to type the RHS before doing overload resolution to find the right method.

lrytz:
Nice one! Could you add a test for that case?

retronym:
We should (separately) pursue an analogous change for the desugaring of default arguments:

```scala
scala> def foo(a: => Any, b: => Any) = a
foo: (a: => Any, b: => Any)Any

scala> foo(b = toString, a = toString) //print

{
  val x$1: () => String @scala.reflect.internal.annotations.uncheckedBounds = (() => $iw.this.toString());
  val x$2: () => String @scala.reflect.internal.annotations.uncheckedBounds = (() => $iw.this.toString());
  $line9.$read.$iw.$iw.foo(x$2.apply(), x$1.apply())
} // : Any
```

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -44,6 +44,11 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
   final val shortenImports = false
 
+  // All typechecked RHS of ValDefs for right-associative operator desugaring
+  val rightAssocValDefs = new mutable.AnyRefMap[Symbol, Tree]
+  // Symbols of ValDefs for right-associative operator desugaring which are passed by name and have been inlined
+  val inlinedRightAssocValDefs = new mutable.HashSet[Symbol]
```
retronym: `private`?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2063,7 +2070,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           } else tpt1.tpe
           transformedOrTyped(vdef.rhs, EXPRmode | BYVALmode, tpt2)
         }
-      treeCopy.ValDef(vdef, typedMods, sym.name, tpt1, checkDead(rhs1)) setType NoType
+      val vdef1 = treeCopy.ValDef(vdef, typedMods, sym.name, tpt1, checkDead(rhs1)) setType NoType
+      if (sym.isSynthetic && sym.name.toString.startsWith(nme.RIGHT_ASSOC_OP_PREFIX))
```
retronym: Should need the to `toString` here.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2474,7 +2484,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           if (result0.nonEmpty) checkPure(result0, supple = true)
         }
 
-        treeCopy.Block(block, statsTyped, expr1)
+        // Remove ValDef for right-associative by-value operator desugaring which has been inlined into expr1
+        val statsTyped2 = statsTyped match {
+          case (vd: ValDef) :: Nil if inlinedRightAssocValDefs contains vd.symbol => Nil
```
retronym: Could eagerly clear the entry from the Map here `if (inlinedRightAssocValDefs.remove(vd.symbol).isDefined) =>`

src/compiler/scala/tools/nsc/typechecker/Typers.scala
retronym: Again, I'd like to use `Map#remove` here to clean up as we go.

retronym:
Another approach here would be to modify the parser to emit a regular application, but mark it with a `RightAssociative` tree attachment. This would be typechecked as is, and afterwards we could lift out the val for strict arguments.

Pros:
 - Better type inference (the formal parameter type of the method (assuming non-overloaded would be used expected type when typechecking the argument.
 - Implementation is closer to the treatment of named/default argument desugaring

Cons:
 - Change in tree shape might have downstream effects on tools that look at the pre-typer tree (maybe quasiquotes, IDEs, ???). While these effects might actually simplify things, they are hard to predict
 - "better" type inference can break existing code in corner cases.

(I don't want to filibuster this PR with the alternative proposal, it can come as a follow up if we think it is worth pursuing.)

We also should consider how easy/hard alternatives are to spec. Currently it quite prescriptive:

> If op is right associative, the same operation is interpreted as `{ val x=e1; e2.op(x ) }`, where x is a fresh name.

Perhaps we could abstract this to specify the evaluation without specifying the desugaring. The spec for named/default applications is also currently prescriptive but doesn't discuss the treatment of by-name params.

szeiger:
> Another approach here would be to modify the parser to emit a regular application, but mark it with a `RightAssociative` tree attachment. This would be typechecked as is, and afterwards we could lift out the val for strict arguments

Yes, I was also thinking about this option since I read your previous proposal yesterday. It would be nice to get the same level of type inference for right-associative operators that you get for other operators and method calls.

lrytz (commented):
LGTM! I think you can squash all in one (and eliminate the no-op changes in RefChecks.scala)

odersky:
I am still very positive on this, but believe it definitely needs a SIP.

jvican:
Can you prepare a quick SIP @szeiger and we discuss it in the next meeting? I'm scheduling one for this month.

lrytz:
He did already https://github.com/scala/scala.github.com/pull/805

szeiger:
Moving to M3, pending SIP

adriaanm:
@szeiger, could you update the PR description to reflect the latest implementation strategy?

szeiger:
Rebased and updated commit comment. Now that SIP-34 was accepted this should be ready to merge for M3.

adriaanm (approved):

PR #5970, opened 2017-07-01
Spec relaxed import behavior

som-snytt:
Implementation was massaged to fix 2133, 3160, 3836
and related tickets around import ergonomics, but
the spec lagged.

This specifies importing multiple type aliases and
that importable requires accessible.

Fixes scala/bug#2133

### DIFF
--- a/spec/02-identifiers-names-and-scopes.md
+++ b/spec/02-identifiers-names-and-scopes.md
@@ -92,7 +92,10 @@ without introducing an ambiguity.
 ```scala
 object X { type T = annotation.tailrec }
 object Y { type T = annotation.tailrec }
-object Z { import X._, Y._, annotation.{tailrec => T} ; @T def f: Int = { f ; 42 } }
+object Z {
+  import X._, Y._, annotation.{tailrec => T}  // OK, all T mean tailrec
+  @T def f: Int = { f ; 42 }                  // error, f is not tail recursive
+}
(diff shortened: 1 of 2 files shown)


spec/02-identifiers-names-and-scopes.md
```diff
@@ -83,6 +83,18 @@ package util {
...
+As a convenience, multiple bindings of a type identifier to the same
+underlying type is permitted. This is possible when import clauses introduce
+a binding of a member type alias with the same binding precedence, typically
+through wildcard imports. This allows redundant type aliases to be imported
+without introducing an ambiguity.
+
+```scala
+object X { type T = annotation.tailrec }
+object Y { type T = annotation.tailrec }
+object Z { import X._, Y._, annotation.{tailrec => T} ; @T def f: Int = { f ; 42 } }
```
lrytz: `f` is not tail recursive 😄 maybe it should be to prevent confusion?

som-snytt: I thought it was more clever to show that the annotation does something? Because if silent, did it actually demonstrate anything? I considered a comment `// error, not tail recursive`, which in retrospect is clarifying.

lrytz: Thanks :)

lrytz (approved):

PR #2494, opened 2013-05-04
makes sense of implicit macros!

xeno-by:
Supersedes https://github.com/scala/scala/pull/2464

review @paulp @retronym

PR #5971, opened 2017-07-01
Prevent `inferImplicitValue` from reporting false divergent implicits

ruippeixotog:
Fixes https://github.com/scala/bug/issues/10398.

`universe.analyzer.inferImplicit` requires the tree in which the implicit is going to be inserted, which is used to detect divergent implicits. This PR makes `Typers#inferImplicitValue` provide a tree from `openImplicits` if such a tree exists. This seems more accurate than always providing `EmptyTree`, which for the purposes of divergent implicit detection doesn't provide the means to distinguish between searches made by macros at different call sites.

This is my first contribution to the Scala compiler and I do not have enough insight about its architecture to know if this is the right approach or if this will affect negatively something else. Please let me know if I missed something or if there is a better option!

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -482,24 +482,27 @@ trait Implicits {
       // otherwise, the macro writer could check `c.openMacros` and `c.openImplicits` and do `c.abort` when expansions are deemed to be divergent
       // upon receiving `c.abort` the typechecker will decide that the corresponding implicit search has failed
       // which will fail the entire stack of implicit searches, producing a nice error message provided by the programmer
-      (context.openImplicits find { case OpenImplicit(info, tp, tree1) => !info.sym.isMacro && tree1.symbol == tree.symbol && dominates(pt, tp)}) match {
-         case Some(pending) =>
-           //println("Pending implicit "+pending+" dominates "+pt+"/"+undetParams) //@MDEBUG
-           DivergentSearchFailure
-         case None =>
-           try {
-             context.openImplicits = OpenImplicit(info, pt, tree, isView) :: context.openImplicits
-             // println("  "*context.openImplicits.length+"typed implicit "+info+" for "+pt) //@MDEBUG
-             val result = typedImplicit0(info, ptChecked, isLocalToCallsite)
-             if (result.isDivergent) {
-               //println("DivergentImplicit for pt:"+ pt +", open implicits:"+context.openImplicits) //@MDEBUG
-               if (context.openImplicits.tail.isEmpty && !pt.isErroneous)
-                 DivergingImplicitExpansionError(tree, pt, info.sym)(context)
-             }
-             result
-           } finally {
-             context.openImplicits = context.openImplicits.tail
+      val existsDominatedImplicit = tree != EmptyTree && context.openImplicits.exists {
+        case OpenImplicit(nfo, tp, tree1) => !nfo.sym.isMacro && tree1.symbol == tree.symbol && dominates(pt, tp)
+      }
+
+        if(existsDominatedImplicit) {
+          //println("Pending implicit "+pending+" dominates "+pt+"/"+undetParams) //@MDEBUG
+          DivergentSearchFailure
+        } else {
+         try {
+           context.openImplicits = OpenImplicit(info, pt, tree, isView) :: context.openImplicits
+           // println("  "*context.openImplicits.length+"typed implicit "+info+" for "+pt) //@MDEBUG
+           val result = typedImplicit0(info, ptChecked, isLocalToCallsite)
+           if (result.isDivergent) {
+             //println("DivergentImplicit for pt:"+ pt +", open implicits:"+context.openImplicits) //@MDEBUG
+             if (context.openImplicits.tail.isEmpty && !pt.isErroneous)
+               DivergingImplicitExpansionError(tree, pt, info.sym)(context)
            }
+           result
+         } finally {
+           context.openImplicits = context.openImplicits.tail
+         }
        }
     }
 
--- a/test/files/run/macro-implicit-decorator.check
+++ b/test/files/run/macro-implicit-decorator.check
@@ -1,0 +1,3 @@
+Successful()
+Failed(List(MyTC[Boolean]))
+Failed(List(MyTC[Boolean]))
--- a/test/files/run/macro-implicit-decorator.flags
+++ b/test/files/run/macro-implicit-decorator.flags
@@ -1,0 +1,1 @@
+-language:experimental.macros
--- a/test/files/run/macro-implicit-decorator/Macros_1.scala
+++ b/test/files/run/macro-implicit-decorator/Macros_1.scala
@@ -1,0 +1,33 @@
+import scala.reflect.macros.whitebox
+
+trait Derivation[A]
+
+object Derivation {
+  case class Successful[A]() extends Derivation[A]
+  case class Failed[A](failures: List[String]) extends Derivation[A]
+
+  var failures = List.empty[String]
+
+  def materializeDerivationImpl[A](c: whitebox.Context)(implicit tt: c.WeakTypeTag[A]): c.Tree = {
+    import c.universe._
+
+    c.inferImplicitValue(weakTypeOf[A]) match {
+      case EmptyTree if c.openImplicits.length == 1 =>
+        q"Derivation.Failed[${weakTypeOf[A]}](Nil)"
+
+      case EmptyTree =>
+        failures ::= weakTypeOf[A].toString
+        q"Derivation.Failed[${weakTypeOf[A]}](Nil)"
+
+      case _ if c.openImplicits.length == 1 && failures.nonEmpty =>
+        val tree = q"Derivation.Failed[${weakTypeOf[A]}](List(..$failures))"
+        failures = Nil
+        tree
+
+      case _ =>
+        q"Derivation.Successful[${weakTypeOf[A]}]()"
+    }
+  }
+
+  implicit def materializeDerivation[A]: Derivation[A] = macro materializeDerivationImpl[A]
+}
--- a/test/files/run/macro-implicit-decorator/Test_2.scala
+++ b/test/files/run/macro-implicit-decorator/Test_2.scala
@@ -1,0 +1,17 @@
+// https://github.com/scala/bug/issues/10398
+
+class CustomClass
+
+trait MyTC[A]
+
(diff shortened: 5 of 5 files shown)


src/compiler/scala/reflect/macros/contexts/Typers.scala
```diff
@@ -41,7 +41,14 @@ trait Typers {
 
   def inferImplicitValue(pt: Type, silent: Boolean = true, withMacrosDisabled: Boolean = false, pos: Position = enclosingPosition): Tree = {
     macroLogVerbose(s"inferring implicit value of type $pt, macros = ${!withMacrosDisabled}")
-    universe.analyzer.inferImplicit(universe.EmptyTree, pt, false, callsiteTyper.context, silent, withMacrosDisabled, pos, (pos, msg) => throw TypecheckException(pos, msg))
+    // If there is an open implicit in the context, use its tree as the one for which the implicit needs to be inserted.
+    // Otherwise, use the empty tree. This prevents false divergent implicit errors like the one in
+    // https://github.com/scala/bug/issues/10398.
+    val tpe = callsiteTyper.context.openImplicits match {
```
hrhino: This is a `Tree`, no? Calling it `tpe` is somewhat misleading.

ruippeixotog: You're right, I was distracted. I have just changed its name to `tree`.

ruippeixotog:
The build is failing because of this partest: [macro-sip19-revised](https://github.com/scala/scala/tree/2.12.x/test/files/run/macro-sip19-revised). However, after taking a look, I don't really understand how the macro in question works. I did some tests and both in `master` and after this fix the macro recurses infinitely (well, until it reaches some Scala compiler limit of 100 open calls). I wonder if that was intended by @xeno-by and what was the reasoning behind it...

src/compiler/scala/reflect/macros/contexts/Typers.scala
```diff
@@ -41,7 +41,14 @@ trait Typers {
...
     macroLogVerbose(s"inferring implicit value of type $pt, macros = ${!withMacrosDisabled}")
-    universe.analyzer.inferImplicit(universe.EmptyTree, pt, false, callsiteTyper.context, silent, withMacrosDisabled, pos, (pos, msg) => throw TypecheckException(pos, msg))
+    // If there is an open implicit in the context, use its tree as the one for which the implicit needs to be inserted.
+    // Otherwise, use the empty tree. This prevents false divergent implicit errors like the one in
+    // https://github.com/scala/bug/issues/10398.
+    val tree = callsiteTyper.context.openImplicits match {
+      case oi :: _ => oi.tree
+      case _ => universe.EmptyTree
+    }
+    universe.analyzer.inferImplicit(tree, pt, false, callsiteTyper.context, silent, withMacrosDisabled, pos, (pos, msg) => throw TypecheckException(pos, msg))
```
retronym: In a standard (non-macro) implicit search, the nested searches that try to fill in implicit paramaters are made in `Typer.applyImplicitArgs` with:

```
inferImplicitFor(paramTp, tree = fun, context, reportAmbiguous = context.reportErrors)
```

`fun` is the tree with an `MethodType(implicit .. )`, ie the application ends up looking like `Apply(fun, implicitArgs)`.

As the search recurses, these trees are stacked in a `List[OpenImplicit]`, which, as you've found, is scanned with:

```
(context.openImplicits find { case OpenImplicit(info, tp, tree1) => !info.sym.isMacro && tree1.symbol == tree.symbol && dominates(pt, tp)}) match {
```

to detect if the current implicitly called method appears higher up in the stack. If so, and if the current expected type isn't reducing in complexity, we declare divergence.

`implicit` macros are given a free pass in this check.

I'm wondering whether using `openImplicits.head` is really what we want here. For one thing, this would mean different behaviour depending on whether the macro was called explicitly or implicitly.

Perhaps what we should to is special case `EmptyTree` in the divergence check, to avoid accidentally satisisfying `tree.symbol` == tree1.symbol` by virtue of `EmptyTree` and `NoSymbol`.

retronym: Another possibility is to use `Ident(NoSymbol.newTermSymbol())` as the tree here. That would have the advantage of only requiring changes to the macro system, which lessens the risk of unexpected regressions.

retronym: @adriaanm WDYT?

retronym:
BTW, great job in debugging this, I tried and failed :)

src/compiler/scala/reflect/macros/contexts/Typers.scala
adriaanm: Special casing on EmptyTree sounds like a good idea to me. Why is tree.symbol NoSymbol though? Can we safely we exclude those trees from the divergence check? (I don't know)

ruippeixotog: @retronym thanks for the explanation, it was insightful and easy to follow :)

If I understood correctly, you are suggesting to disable the divergence check for every case where `tree == EmptyTree`, which amounts to every `inferImplicitValue` call from a macro, right? What would happen if the implicit truly diverges? My guess is that the compiler would give up after some open implicit limit, which would indeed result in no false positives and false negatives at the cost of some extra time and memory expanding unnecessary implicits.

With the solution in this PR I believe that there wouldn't be different behavior depending on whether the macro was called explicitly or implicitly. If a macro gets called explicitly and triggers an implicit search like the example `Derivation` macro, only the first expansion gets called with an `EmptyTree`. Subsequent macro calls must already have a non-empty `openImplicits` list, which would still provide the information to disambiguate between calls.

If we want to keep the divergence check, the tree at the head of `openImplicits` seems to be the most accurate one, as it indicates the "latest" known location that caused the macro to run. For the `None` case in the check, would `callsiteTyper.context.tree` be an acceptable alternative? From what I could gather debugging, it is always the tree that first triggered the macro (in the partest it would be, for example, `println(implicitly[Derivation[MyTC[CustomClass]]])`).

I'm basing my reasoning on the premise that for the divergence implicit to work best, the macro context should provide the most accurate tree that triggered the current implicit search, which allows the divergence checker to detect method call "loops" with non-decreasingly complex type parameters.

Once more, please let me know if I'm missing some use case! Both special-casing `EmptyTree` and using fresh names seem solutions that would work for my case, so if you think it's the best option, I'm happy to change this PR :)

retronym:
If suggest the emptytree special case, but I'd ask you to try to craft an
example that exploits the missing divergence check.

ruippeixotog:
Here's an minimal example showing the effects of the missing divergence check: https://gist.github.com/ruippeixotog/fb78d1be2c4a67fb989732a000e7c698.

I can't think of any example where not having divergence checks gets a "good" implicit to fail or a bad one to work, it just causes the expansion of more implicits/macros. It may translate into functional differences if the macro has side-effects, uses mutable state or depends in some strange way on the context, but it would be a very convoluted case and macros should be prepared against that anyway.

retronym:
There is a risk that this change effectively revokes the free pass granted to implicit macros in https://github.com/scala/scala/pull/2494/files#diff-7c03397456d3b987549fd039d6b639c6R446

Looking back at that change, I can't really find a succinct justification as to why implicit macros need the flexibility of managing their own divergence checking. There was a demonstration of a problem in the definition of divergence checking with alias types ([discussion](https://groups.google.com/forum/#!topic/scala-internals/VFRis3ewrAo). But I can't see why the answer here was "disable divergence checks for macros" rather than "try to fix the inconsistency".

But if we take it as a given that we can't make divergence more likely for implicit macros, I think we'll need to special case `EmptyTree`.

retronym:
`run/macro-sip19-revised` passes under https://github.com/retronym/scala/tree/review/5971

(The test might itself still be flawed, based on your analysis.)

ruippeixotog:
SI-7167, looking at the description and at the fix, seems to be targeted at macro-only implicit chains such as the `Complex` example there, and not at `inferImplicitValue` calls. While I can see why macro-only chains should get optional control over divergence checking, I expect `inferImplicitValue` to work just as if I asked for an `implicit val` at the macro call site.

I see two ways of special casing `EmptyTree`: we can either:
1. add a new condition to `context.openImplicits find { case OpenImplicit(info, tp, tree1) => ... }`;
2. disable that line if there are _any_ empty tree in `context.openImplicits`.

What option were you thinking of?

Currently, `inferImplicitValue` inside a macro triggers an implicit search where the root type being searched is associated with an `EmptyTree`. If that root type triggers other implicit searches, they are associated with the "right" tree as long as there are no more macros involved. In that case I think that divergence is still checked unless we choose option 2. But I also think that, as macros are not involved in the "inner" implicit search, there would not be any need to fully disable divergence checking, as the cases that motivated SI-7167 would not be applicable.

The most complicated (and possibly more contrived) cases are the ones in this PR, where macros are intertwined with regular implicit searches like this:

```
implicitly[Macro[A]]   // not checked for divergence since it's a macro  
implicitly[A]          // triggered from Macro[A] - called with EmptyTree
implicitly[Macro[B]]   // not checked for divergence since it's a macro
implicitly[B]          // triggered from Macro[B] - called with EmptyTree
(...)
```

Both options above would disable divergence checking in this case and the macro would have all the work of doing that. The current solution of this PR simply would not allow the `A`, `B`… searches to result in divergence, while still having it disabled on `Macro[A]`, `Macro[B]`… - the target of SI-7167.

That said, if you still prefer special-casing `EmptyTree` tell me what option above you prefer and I'll change this PR :)

ruippeixotog:
I'm sorry, I commented without seeing your response above :) That would be option 1 then.

As a real-world example of my problem, I'm trying to build a macro that improves implicit not found messages like [this](https://github.com/pureconfig/pureconfig/pull/279/files#diff-5cf8784c0894deb4e1f75accb0ac1bfbR51). In this case the number of opened macros (which depends on the divergence checker) matters to me, but I can surely do the divergence checking inside my macro.

ruippeixotog:
@retronym I have just updated the PR with your solution. Please let me know if you need me to do anything else.

(one question: I'm trying to `publishLocal` this fix to test in my project, but SBT is throwing a `java.util.NoSuchElementException: None.get` at [build.sbt:405](https://github.com/scala/scala/blob/2.12.x/build.sbt#L405). Do you have any idea what I'm doing wrong?)

hrhino:
scala/scala-dev#399, perhaps?

ruippeixotog:
That's it, I was also using coursier. I disabled it and the error disappeared. Thanks!

retronym (approved):

PR #5972, opened 2017-07-03
Fixes #10399: Adding an element will enter dead loop while constructing a mutable BitSet with an empty array.

joymufeng:
Constructing a mutable BitSet with an empty array will return an empty instance. 
Fixes scala/bug#10399

### DIFF
--- a/src/library/scala/collection/mutable/BitSet.scala
+++ b/src/library/scala/collection/mutable/BitSet.scala
@@ -77,7 +77,13 @@ class BitSet(protected final var elems: Array[Long]) extends AbstractSet[Int]
     }
   }
 
-  protected def fromBitMaskNoCopy(words: Array[Long]): BitSet = new BitSet(words)
+  protected def fromBitMaskNoCopy(words: Array[Long]): BitSet = {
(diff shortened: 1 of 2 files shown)


src/library/scala/collection/mutable/BitSet.scala
```diff
@@ -77,7 +77,13 @@ class BitSet(protected final var elems: Array[Long]) extends AbstractSet[Int]
     }
   }
 
-  protected def fromBitMaskNoCopy(words: Array[Long]): BitSet = new BitSet(words)
+  protected def fromBitMaskNoCopy(words: Array[Long]): BitSet = {
+    if (words.isEmpty) {
```
lrytz: `words.length == 0` would avoid the call to an extension method on ArrayOps.

joymufeng: Thanks, I will correct it.

joymufeng: @lrytz I have corrected it, could you please have a check!

lrytz (approved):
Thanks!

PR #5973, opened 2017-07-03
Structural invocation: find method by name if by signature fails

lrytz:
In a structural invocation, if the receiver object doesn't have
a method with the expected paramter types, check if there's a
single method matching the name. If so, invoke this method.

This fixes parts of https://github.com/scala/bug/issues/10334,
namely the parts that regressed with the new lambda encoding.
But the underlying issue still remains.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/CleanUp.scala
+++ b/src/compiler/scala/tools/nsc/transform/CleanUp.scala
@@ -108,7 +108,7 @@ abstract class CleanUp extends Statics with Transform with ast.TreeDSL {
               if (method ne null)
                 return method
               else {
-                method = ScalaRunTime.ensureAccessible(forReceiver.getMethod("xyz", methodCache.parameterTypes()))
+                method = ScalaRunTime.lookupStructuralMethod(forReceiver, "xyz", methodCache.parameterTypes())
                 methodCache.add(forReceiver, method)
                 return method
               }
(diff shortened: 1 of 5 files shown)


sjrd:
Ouch.

I'd like to stress that we won't ever be able to implement the same behavior in Scala.js. If at least there wasn't the test `!x.isBridge` in `lookupStructuralMethod`, we could make the linker to support the same fallback. But with that test, it cannot be replicated: the Scala.js linker has no idea what is or isn't a "bridge".

Reflective calls are broken, but at least at the moment they are uniformly broken across JVM and JS. We put a lot of effort into replicating the JVM behavior, even bug-compatibility. It would be very disappointing to see that compatibility go away :worried:

lrytz:
The isBridge test is not needed for the test case, and I didn't come up with an example where it's actually needed. I added it because it seems to make sense, but we can get rid of it.

I don't have any idea for a proper fix of the issue..

src/library/scala/runtime/ScalaRunTime.scala
```diff
@@ -139,6 +139,19 @@ object ScalaRunTime {
   // More background at ticket #2318.
   def ensureAccessible(m: JMethod): JMethod = scala.reflect.ensureAccessible(m)
 
+  def lookupStructuralMethod(c: Class[_], m: String, params: Array[Class[_]]): JMethod = {
```
lrytz, line 142: This doesn't pass MiMa. It's more than an internal issue: if we compile new 2.12 code to use this method, it won't run when an older 2.12 library is on the classpath.

src/compiler/scala/tools/nsc/transform/CleanUp.scala
```diff
@@ -108,7 +108,7 @@ abstract class CleanUp extends Statics with Transform with ast.TreeDSL {
               if (method ne null)
                 return method
               else {
-                method = ScalaRunTime.ensureAccessible(forReceiver.getMethod("xyz", methodCache.parameterTypes()))
+                method = ScalaRunTime.lookupStructuralMethod(forReceiver, "xyz", methodCache.parameterTypes())
```
retronym, line 111: How about emitting:

```scala
method = try {
    ScalaRunTime.ensureAccessible(forReceiver.getMethod("xyz", methodCache.parameterTypes())) 
  } catch {
    case e: NoSuchMethodException => try { lookupStructuralMethod(...) } catch { case _: Throwable => throw e }
}
```

To profit from the bug fix, you'd need to the latest standard library available at runtime, but structural calls that used to work will continue to work with the old library.

lrytz: We could also emit the entire body of `lookupStructuralMethod` in CleanUp in 2.12.x, and introduce the ScalaRunTime method only in 2.13.x

lrytz:
Alternative in https://github.com/scala/scala/pull/5977

retronym:
Closing this one for now. This approach might still complement the alternative fix (although it sounds there will still be cases that are impossible to unambiguously resolve), so let's pursue this further on 2.13.x

PR #5975, opened 2017-07-05
Two fixes for overload resolution for PartialFunctions

szeiger:
The extension of shape-based overload resolutions for pattern-matching
anonymous functions in https://github.com/scala/scala/pull/5698 did not
work correctly in all cases:

- Type unification for shapes needs to treat `PartialFunction` types
  the same as `FunctionN` and SAM types. They were considered for
  unification but the parameter type was not extracted correctly.

- When the matched argument types are constructed after a successful
  shape match, a pattern-matching anonymous function has to be typed as
  `PartialFunction` instead of `FunctionN`.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3368,50 +3368,80 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                 catch { case _: IllegalArgumentException => args.map(_ => Nil) } // fail safe in case formalTypes fails to align to argslen
               else args.map(_ => Nil) // will type under argPt == WildcardType
 
-            val (args1, argTpes) = context.savingUndeterminedTypeParams() {
+            def typedSingleArg(arg: Tree, argPtAlts: List[(Type, Symbol)], amode: Mode, matchAsPartial: Boolean): (Tree, Type, Boolean) = {
+              def typedArg0(tree: Tree) = {
+                // if we have an overloaded HOF such as `(f: Int => Int)Int <and> (f: Char => Char)Char`,
+                // and we're typing a function like `x => x` for the argument, try to collapse
+                // the overloaded type into a single function type from which `typedFunction`
+                // can derive the argument type for `x` in the function literal above
+                val (argPt, maybeRetype) =
+                  if (argPtAlts.isEmpty) (WildcardType, false)
+                  else if (treeInfo.isFunctionMissingParamType(tree)) (functionProto(argPtAlts), false)
+                  else if (treeInfo.isPartialFunctionMissingParamType(tree)) {
+                    if(matchAsPartial) {
+                      val (pfAlts, nonPfAlts) = argPtAlts.partition(ts => isPartialFunctionType(ts._1))
+                      if (pfAlts.nonEmpty && nonPfAlts.nonEmpty) (partialFunctionProto(argPtAlts), true)
+                      else if (pfAlts.nonEmpty) (partialFunctionProto(argPtAlts), false)
+                      else (functionProto(argPtAlts), false)
+                    } else (functionProto(argPtAlts), false)
+                  } else (WildcardType, false)
+                val tree2 = if (maybeRetype) duplicateAndKeepPositions(tree) else tree
+                val argTyped = typedArg(tree2, amode, BYVALmode, argPt)
+                (argTyped, argTyped.tpe.deconst, maybeRetype)
+              }
+
+              arg match {
+                // scala/bug#8197/scala/bug#4592 call for checking whether this named argument could be interpreted as an assign
+                // infer.checkNames must not use UnitType: it may not be a valid assignment, or the setter may return another type from Unit
+                // TODO: just make it an error to refer to a non-existent named arg, as it's far more likely to be
+                //       a typo than an assignment passed as an argument
+                case AssignOrNamedArg(lhs@Ident(name), rhs) =>
+                  // named args: only type the righthand sides ("unknown identifier" errors otherwise)
+                  // the assign is untyped; that's ok because we call doTypedApply
+                  typedArg0(rhs) match {
+                    case (rhsTyped, tp, partial) => (treeCopy.AssignOrNamedArg(arg, lhs, rhsTyped), NamedType(name, tp), partial)
+                  }
+                case treeInfo.WildcardStarArg(_) =>
+                  typedArg0(arg) match {
+                    case (argTyped, tp, partial) => (argTyped, RepeatedType(tp), partial)
+                  }
+                case _ =>
+                  typedArg0(arg)
+              }
+            }
+
+            val (args1, argTpes, maybeRetypeArgs) = context.savingUndeterminedTypeParams() {
               val amode = forArgMode(fun, mode)
-
-              map2(args, altArgPts) { (arg, argPtAlts) =>
-                def typedArg0(tree: Tree) = {
-                  // if we have an overloaded HOF such as `(f: Int => Int)Int <and> (f: Char => Char)Char`,
-                  // and we're typing a function like `x => x` for the argument, try to collapse
-                  // the overloaded type into a single function type from which `typedFunction`
-                  // can derive the argument type for `x` in the function literal above
-                  val argPt =
-                    if (argPtAlts.isEmpty) WildcardType
-                    else if (treeInfo.isFunctionMissingParamType(tree)) functionProto(argPtAlts)
-                    else if (treeInfo.isPartialFunctionMissingParamType(tree)) partialFunctionProto(argPtAlts)
-                    else WildcardType
-
-                  val argTyped = typedArg(tree, amode, BYVALmode, argPt)
-                  (argTyped, argTyped.tpe.deconst)
-                }
-
-                arg match {
-                  // scala/bug#8197/scala/bug#4592 call for checking whether this named argument could be interpreted as an assign
-                  // infer.checkNames must not use UnitType: it may not be a valid assignment, or the setter may return another type from Unit
-                  // TODO: just make it an error to refer to a non-existent named arg, as it's far more likely to be
-                  //       a typo than an assignment passed as an argument
-                  case AssignOrNamedArg(lhs@Ident(name), rhs) =>
-                    // named args: only type the righthand sides ("unknown identifier" errors otherwise)
-                    // the assign is untyped; that's ok because we call doTypedApply
-                    typedArg0(rhs) match {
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3374,7 +3378,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                   // the overloaded type into a single function type from which `typedFunction`
                   // can derive the argument type for `x` in the function literal above
                   val argPt =
-                    if (argPtAlts.nonEmpty && treeInfo.isFunctionMissingParamType(tree)) functionProto(argPtAlts)
+                    if (argPtAlts.isEmpty) WildcardType
+                    else if (treeInfo.isFunctionMissingParamType(tree)) functionProto(argPtAlts)
+                    else if (treeInfo.isPartialFunctionMissingParamType(tree)) partialFunctionProto(argPtAlts)
```
lrytz: A difference is that now the compiler will generate a PartialFunction even if the selected alternative has a Function parameter type
```
object T {
  def m(f: Object => Int, s: String): Int = 1
  def m(f: String => Int, s: Object): Int = 1
}

T.m({ case x => 2 }, null)
```
generates
```
T.m(({
          @SerialVersionUID(value = 0) final <synthetic> class $anonfun extends scala.runtime.AbstractPartialFunction[Object,Int] with Serializable {
            def <init>(): <$anon: Object => Int> = {
              $anonfun.super.<init>();
              ()
            };
            final override def applyOrElse[A1 <: Object, B1 >: Int](x1: A1, default: A1 => B1): B1 = ((x1.asInstanceOf[Object]: Object): Object @unchecked) match {
              case (x @ _) => 2
              case (defaultCase$ @ _) => default.apply(x1)
            };
            final def isDefinedAt(x1: Object): Boolean = ((x1.asInstanceOf[Object]: Object): Object @unchecked) match {
              case (x @ _) => true
              case (defaultCase$ @ _) => false
            }
          };
          new $anonfun()
        }: PartialFunction[Object,Int]), null);
```

In 2.13.0-M1

```
T.m(((x0$1: Object) => x0$1 match {
          case (x @ _) => 2
        }), null);
```

test/files/run/InferOverloadedPartialFunction.scala
```diff
@@ -45,3 +45,12 @@ object Test extends App {
...
 }
+
+trait SortedMap {
+  def collect[B](pf: PartialFunction[(String, Int), B]): Int = 0
+  def collect[K2 : Ordering, V2](pf: PartialFunction[(String, Int), (K2, V2)]): Int = 0
+  def collectF[B](pf: Function1[(String, Int), B]): Int = 0
+  def collectF[K2 : Ordering, V2](pf: Function1[(String, Int), (K2, V2)]): Int = 0
+  def foo(xs: SortedMap) = xs.collect { case (k, v) => 1 }
+  def fooF(xs: SortedMap) = xs.collectF { case (k, v) => 1 }
+}
```
lrytz: Maybe add a (few) test(s) where the overloads mix Function and PartialFunction parameter types (and also assert the expected method is selected).

lrytz (commented):
Looks good!

src/compiler/scala/tools/nsc/typechecker/Typers.scala
szeiger: My first impulse was to write "this works as intended and specified in https://github.com/scala/scala/pull/5698" but the spec change doesn't actually mandate it and it would be more efficient to create a `Function1` in cases where that's sufficient.

szeiger: Looks like a non-trivial problem. Perhaps extract the argument types from `fun` after `inferMethodAlternative` (is there a simple way to do that?) and then rebuild `args1` with `Function1` types in place of `PartialFunction` where it is sufficient?

szeiger:
I've added the changes necessary to type Match blocks as Function1 again after overload resolution unless the inferred alternative actually requires a PartialFunction.

lrytz:
Hmm, that last commit adds quite a bit of complexity. How about using a PF expected type if some of the overloads takes a PF? Then we could still end up with a PF passed as F, but only when overloads mix F and PF parameter types. What do you think?

szeiger:
Most of the complexity comes from retyping in the case of mixed overloads. This is the whole `if(maybeRetypeArgs contains true)` block. The rest looks bigger than it is because abstracting out `typedSingleArg` changed the indentation.

> How about using a PF expected type if some of the overloads takes a PF? Then we could still end up with a PF passed as F, but only when overloads mix F and PF parameter types

This is already done (in the lines following https://github.com/scala/scala/pull/5975/files#diff-4eab1aad4533a31c10565971e90f73eaR3382) to avoid copying the tree in the non-mixed cases. If we remove the `maybeRetypeArgs` handling we end up with `PartialFunction` if any overload required it.

szeiger:
How do we move this forward? Should I cut it back to a simpler impl that doesn't optimize the mixed overload case? I'd like to get this into M2 because we're running into these bugs in collections.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3359,53 +3363,85 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+              val args2 = if(maybeRetypeArgs contains true) {
+                // There were pattern-matching anonymous functions in the argument list which got typed as PartialFunction
+                // but if the inferred alternative does not actually need a PartialFunction we retype it as Function1.
+                def paramTypes(tpe: Type): List[Type] = tpe match {
+                  case ExistentialType(_, qtpe) => paramTypes(qtpe)
+                  case MethodType(syms, _) => syms.map(_.tpe)
+                  case TypeRef(_, _, tpes) => tpes.init
+                  case _ => Nil
+                }
+                val selectedParams = paramTypes(normalize(fun.tpe))
```
lrytz: It's surprising me a bit that `fun.tpe` can be `ExistentialType` or `TypeRef`, what are examples? How about using `fun.symbol.tpe` instead, would that always be a `MethodType`?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3359,53 +3363,85 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                if(selectedParams.nonEmpty) {
+                  context.savingUndeterminedTypeParams() {
+                    val amode = forArgMode(fun, mode)
+                    val keepTypes = map3(selectedParams, maybeRetypeArgs, args1) { (param, maybeRetype, arg1) =>
+                      if (maybeRetype && !isPartialFunctionType(param)) None else Some(arg1)
+                    }
+                    map3(args, altArgPts, keepTypes) {
+                      case (_, _, Some(keep)) => keep
+                      case (arg, argPtAlts, None) => typedSingleArg(arg, argPtAlts, amode, matchAsPartial = false)._1
```
lrytz, line 3438: @retronym wdyt about throwing away a type checked tree and typing it again? Are you aware of other places where we do this? Any potential problems that come to mind with side-effects of type checking, for examples related to definitions within that tree? I cannot think of any from the top of my head.

The alternative would be to generate a PartialFunction instance in case an overloaded method has alternatives with `Function1` and `PF` parameter types, and the `Function1` alternative is chosen.

lrytz, line 3438: You could probably use the selected function's parameter type here as expected type to type check the argument, so you wouldn't need to go through `typedSingleArg` / `functionProto` again. Just a thought, maybe it's better to keep the two type checking invocations as similar as possible.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
szeiger: I expected `MethodType` but `TypeRef` was the most common case. `MethodType` and `ExistentialType` came up in some partest cases but I didn't check for more details. I'll give `fun.symbol.tpe` a try.

szeiger: `ClassArgsTypeRef` shows up in the last two test cases (passing a case literal to a method with mixed overloads) in `run/InferOverloadedPartialFunction.scala`, both for `fun.tpe` and `fun.symbol.tpe`:

```scala
  assert(ys.collect { case (k, v) => 1 } == 0)
  assert(ys.collect { case (k, v) => (1, 1) } == 2)
```

I wasn't able to reproduce the `ExistentialType` anymore with `partest pos neg run`. Maybe it would show up in a full test run.

szeiger:
This is the simpler version that doesn't retype arguments: https://github.com/scala/scala/compare/2.13.x...szeiger:fix/pf-overloads2

src/compiler/scala/tools/nsc/typechecker/Typers.scala
lrytz: Ah, the source of the non-method types is the call to `normalize` - why is it there? Otherwise you should get a `PolyType` or `MethodType`

szeiger: Ah, right, `normalize` converts the method type to a function type. Just cargo-cult programming (copied from some place that did something similar). I've pushed a change that doesn't normalize anymore, now I get the expected `MethodType` or `PolyType`.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3422,12 +3422,11 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                 // There were pattern-matching anonymous functions in the argument list which got typed as PartialFunction
                 // but if the inferred alternative does not actually need a PartialFunction we retype it as Function1.
                 def paramTypes(tpe: Type): List[Type] = tpe match {
-                  case ExistentialType(_, qtpe) => paramTypes(qtpe)
+                  case PolyType(_, restpe) => paramTypes(restpe)
                   case MethodType(syms, _) => syms.map(_.tpe)
-                  case TypeRef(_, _, tpes) => tpes.init
-                  case _ => Nil
+                  case ErrorType => Nil
```
lrytz: I'd go for the more cautious `case _`

lrytz:
Let's merge your "simpler version that doesn't retype arguments" (https://github.com/scala/scala/compare/2.13.x...szeiger:fix/pf-overloads2) for M2. We can re-visit the other version for M3, but I think it's not worth the additional complexity.

SethTisue:
@szeiger should this remain in the PR queue...?

szeiger:
The mixed overloads case is probably not important enough to warrant further optimization.

PR #5976, opened 2017-07-06
Add missing phase travel when searching a companion in the backend

lrytz:
When looking for the companion of a nested class A$C in the backend, a
pahse travel is necessary to avoid finding the package-owned symbol for
A$C.class (created by SymbolLoaders) instead of the symbol C owned by A.

BTypesFromSymbols contains many such phase travels already, this fixes
a missing case.

Fixes scala/scala-dev#402

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
@@ -582,7 +582,11 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
     def keepMember(sym: Symbol) = sym.isMethod && !scalaPrimitives.isPrimitive(sym)
     val classMethods = classSym.info.decls.iterator.filter(keepMember)
     val methods = if (!classSym.isJavaDefined) classMethods else {
-      val staticMethods = classSym.companionModule.info.decls.iterator.filter(m => !m.isConstructor && keepMember(m))
+      // Phase travel important for nested classes (scala-dev#402). When a java class symbol A$B
+      // is compiled from source, this ensures that `companionModule` doesn't return the `A$B`
+      // symbol created for the `A$B.class` file on the classpath, which might be different.
+      val companion = exitingPickler(classSym.companionModule)
+      val staticMethods = companion.info.decls.iterator.filter(m => !m.isConstructor && keepMember(m))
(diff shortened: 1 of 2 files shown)


sjrd:
Huh. Is it also what we should be doing in Scala.js instead of [our `companionModuleClass` hack](https://github.com/scala-js/scala-js/blob/b83411d91c00ef5d18599d0a3948b1f56cc2dcdd/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L186-L192)?

lrytz:
That sounds plausible, give it a try!

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
```diff
@@ -582,7 +582,11 @@ class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
     def keepMember(sym: Symbol) = sym.isMethod && !scalaPrimitives.isPrimitive(sym)
     val classMethods = classSym.info.decls.iterator.filter(keepMember)
     val methods = if (!classSym.isJavaDefined) classMethods else {
-      val staticMethods = classSym.companionModule.info.decls.iterator.filter(m => !m.isConstructor && keepMember(m))
+      // Phase travel important for nested classes (scala-dev#402). When a java class symbol A$B
+      // is compiled from source, this ensures that `companionModule` doesn't return the `A$B`
+      // symbol created for the `A$B.class` file on the classpath, which might be different.
+      val companion = exitingPickler(classSym.companionModule)
+      val staticMethods = companion.info.decls.iterator.filter(m => !m.isConstructor && keepMember(m))
```
retronym, line 589: I took a quick look to see why the flatten info transformer doesn't move the module symbol into the package scope. This [`!sym.isStaticModule`](https://github.com/scala/scala/blob/01307fab3dd11c4ca99f7d4605b4a34e76605a27/src/compiler/scala/tools/nsc/transform/Flatten.scala#L75) check seems to be related.

retronym (approved):
LGTM, this fix follows the prevailing pattern in the backend.

sjrd:
> That sounds plausible, give it a try!

Thanks, I will. https://github.com/scala-js/scala-js/issues/3042

retronym:
Logging the idea to fix this in `flatten` as https://github.com/scala/scala-dev/issues/403

PR #5977, opened 2017-07-06
Add a bridge to LMF-generated lambdas with the instantiated signature

lrytz:
Currently, a lambda object for `(s: String) => String` only gets the
`apply(Object)Object` method (LMF generates the necessary casts). When
using such a lambda through a structural type `{def apply(s: String): String}`,
the reflective lookup for the apply mehthod fails.

This patch asks LMF to generate a bridge method with the instantiated signature.

Fixes the regressed parts of scala/bug#10334

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -1361,10 +1361,14 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
   private def visitInvokeDynamicInsnLMF(jmethod: MethodNode, samName: String, invokedType: String, samMethodType: asm.Type,
                                         implMethodHandle: asm.Handle, instantiatedMethodType: asm.Type,
                                         serializable: Boolean, markerInterfaces: Seq[asm.Type]) = {
-    import java.lang.invoke.LambdaMetafactory.{FLAG_MARKERS, FLAG_SERIALIZABLE}
+    import java.lang.invoke.LambdaMetafactory.{FLAG_MARKERS, FLAG_SERIALIZABLE, FLAG_BRIDGES}
+    // scala/bug#10334: make sure that a lambda object for `T => U` has a method `apply(T)U`, not only the `(Object)Object`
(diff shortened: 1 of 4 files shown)


retronym:
Testing idea:

  - what happens with clashes with inherited methods from user defined SAM types?
  - value classes, void return

lrytz:
It seems the VM treats it as an override if there's a clashing method inherited, which makes sense, as there's no override flag in the VM.

Using the following SAM interface (written in Java, because a Scala trait with a non-abstract method gets an `$init$` method, which [causes lambdas to be expanded to anonymous classes](https://github.com/scala/scala/blob/5a25c2a6e33df72f9f1691309d82bcd16cdfd33a/src/compiler/scala/tools/nsc/transform/Erasure.scala#L1317))

```java
public interface A<T> {
  public String apply(T s);
  public default String apply(String s) { return "hi"; }
}
```

```scala
object Test {
    val t: A[String] = s => s
    println((t: A[_]).apply("there")) // uses indyLambda
  }
}
```

The type ascription is needed to prevent "ambiguous reference to overloaded definition".

In 2.12.2, this prints `hi`, after this patch it prints `there`. However, 2.12.2 with `-Ydelambdafy:inline` also prints `there`. So maybe that's an improvement.

lrytz:
Void return is handled by `anonfun$adapted`:

```scala
trait T[T] { def apply(x: Int): T }
object Test {
    val t: T[Unit] = x => println(x)
    println(t.apply(1))
  }
}
```

gives `indyLambda $anonfun$main$1$adapted(I)Ljava/lang/Object;`, which invokes `$anonfun$main$1(I)V`.

lrytz:
Also value classes are handled by `anonfun$adapted`

```scala
trait T[T] { def apply(x: Int): T }
class C(val x: Int) extends AnyVal { override def toString = s"$x" }
object Test {
    val t: T[C] = x => new C(x)
    println(t.apply(1))
  }
}
```

gives `indyLambda $anonfun$main$1$adapted(I)Ljava/lang/Object;`, which invokes `$anonfun$main$1(I)I` and boxes its result in a `new C`.

lrytz:
Added the examples above as tests

retronym (approved):

PR #5979, opened 2017-07-07
Disable tag driven publication of distribution

retronym:
The bootstrap script is called within PR validation
and nightly builds, in addition to being used in the
release process.

It used to try to to automatically determine which
of these contexts was active based on the whether
HEAD was suitable tagged.

However, if same commit was rebuilt later as a nightly,
new binaries would be created and overwrite the official
ones.

This commit removes this logic. `publishToSonatype=yes`
will need to be explicitly provided in then environment
of the release build.

References scala/scala-dev#387

### DIFF
--- a/scripts/jobs/integrate/bootstrap
+++ b/scripts/jobs/integrate/bootstrap
@@ -14,9 +14,7 @@
 
 # Specifying the Scala version:
 #  - To build a release (this enables publishing to sonatype):
-#    - Either specify SCALA_VER_BASE. You may also specify SCALA_VER_SUFFIX, the Scala version is SCALA_VER=$SCALA_VER_BASE$SCALA_VER_SUFFIX.
-#    - Or have the current HEAD tagged as v$base$suffix
-#    - To prevent staging on sonatype (for testing), set publishToSonatype to anything but "yes"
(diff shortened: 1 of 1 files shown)


retronym:
Hmm, maybe I'm mis-reading the request in that ticket. We're trying to address:

> Before the next nightly, advance HEAD on scala/scala past the tag. Otherwise, the nightly build will re-stage the release and overwrite the distribution archives on S3. TODO: should we disable this?

lrytz:
Yeah, I think the plan is to remove this part: https://github.com/scala/scala/blob/2.12.x/scripts/jobs/integrate/bootstrap#L300-L311

lrytz:
Also https://github.com/scala/scala/blob/2.12.x/scripts/jobs/integrate/bootstrap#L18-L19

lrytz:
> publishToSonatype=yes
> will need to be explicitly provided in then environment
> of the release build.

Actually, just setting the `SCALA_VER_BASE` is enough

scripts/jobs/integrate/bootstrap
```diff
@@ -15,8 +15,6 @@
 # Specifying the Scala version:
 #  - To build a release (this enables publishing to sonatype):
 #    - Either specify SCALA_VER_BASE. You may also specify SCALA_VER_SUFFIX, the Scala version is SCALA_VER=$SCALA_VER_BASE$SCALA_VER_SUFFIX.
```
lrytz: "Either" is stale now

lrytz (approved):

PR #5980, opened 2017-07-08
Fix configuration of SBT incremental compiler

retronym:
We want to use the (old) rules for incremental compilation
after editing files with macro definitions, as the new
rules lead to a lot of recompiles.

We'd previously configured this, but the change had been
somehow broken by a more recent change to the incremental
compiler options.

This commit moves both of the configs to the same part of
our build, which seems to make it stick:

```
% sbt consoleProject
> (incOptions in LocalRootProject).eval.recompileOnMacroDef
false
```

Fixes scala/scala-dev#337

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -337,8 +337,6 @@ lazy val library = configureAsSubproject(project)
         "-doc-root-content", (sourceDirectory in Compile).value + "/rootdoc.txt"
       )
     },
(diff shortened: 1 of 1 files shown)


szeiger (approved):
It looks like all projects do indeed inherit this setting from the root project (which I find rather unusual and surprising).

PR #5983, opened 2017-07-08
Fix regression with eta expansion of implicit method

retronym:
In #5327, a change was made to typedEta to accept an
original (ie, pre-typechecked) tree to be used in a
fallback path.

However, the caller provided an original tree larger
than the actual tree being typechecked.

This commit just passes the part of the orig tree that
corresponds to the tree we're eta expanding, rather than
the entire `Typed(methodValue, functionPt)` tree.

That avoids an infinite loop in typechecking the erroneous
code in the test case.

Fixes scala/bug#10279

### DIFF
--- a/test/files/neg/t10279.scala
+++ b/test/files/neg/t10279.scala
@@ -1,0 +1,10 @@
+object Test {
+
+  def foo(i: Int)(implicit s: String): String = ???
+
+  def test(implicit s: String) {
+    // foo(1) _
+  }
+
+  foo(1) _
+}
--- a/test/files/neg/t10279.scala
+++ b/test/files/neg/t10279.scala
@@ -6,5 +6,8 @@ object Test {
     // foo(1) _
   }
 
-  foo(1) _
+  val bar = foo(1) _
+
+  def fooSimple(implicit x: Int): Int = x
+  val barSimple = fooSimple _
 }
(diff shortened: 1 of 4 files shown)


retronym:
Review in good time by @adriaanm (if/when the new dependently-typed member allows it.)

test/files/neg/t10279.scala
```diff
@@ -0,0 +1,10 @@
+object Test {
+
+  def foo(i: Int)(implicit s: String): String = ???
+
+  def test(implicit s: String) {
+    // foo(1) _
+  }
+
+  foo(1) _
```
lrytz: let's add the slightly simpler test case from https://github.com/scala/bug/issues/10299
```
scala> def foo(implicit x: Int): Int = x
scala> val bar = foo _
```

adriaanm (approved):
LGTM. Added Lukas's test case

lrytz:
```
!!   65 - run/byname.scala                          [output differs]
% diff /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/run/byname-run.log /home/jenkins/workspace/scala-2.12.x-validate-test@2/test/files/run/byname.check
@@ -1,4 +1,3 @@
-warning: there were two deprecation warnings (since 2.12.0); re-run with -deprecation for details
 test no braces completed properly
 test no braces r completed properly
```

adriaanm:
Push -f'ed to Jason's branch again :)

PR #5984, opened 2017-07-08
Correct scoping in typing application to block

retronym:
Restrict a special case in typechecking applications to
the desugaring of right assoc operators, as in general
it's not sound to rewrite `{stats; expr}(a)` to `{stats; expr(a)}`,
because it changes visibility of names. Not sure why we even
need this special case, but that's a thread to pull on another day.

In the test case, the argument was binding to the inner
definition of `x`, which shouldn't have been in scope.

TODO: can we drop this when we eta-expand more aggressively?

Fixes scala/bug#7881


lrytz:
Doesn't bootstrap, you get a few of those:
```
[error] /home/jenkins/workspace/scala-2.12.x-validate-test@2/src/library/scala/Function.scala:24: missing argument list for method /: in trait TraversableOnce
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `/: _` or `/:(_)(_)` instead of `/:`.
[error]   def chain[a](fs: Seq[a => a]): a => a = { x => (x /: fs) ((x, f) => f(x)) }
[error]                                                     ^
```

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4740,8 +4740,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       }
 
       def typedApply(tree: Apply) = tree match {
-        case Apply(Block(stats, expr), args) =>
-          typed1(atPos(tree.pos)(Block(stats, Apply(expr, args) setPos tree.pos.makeTransparent)), mode, pt)
+        case Apply(Block((cd: ClassDef) :: Nil, expr), args) if cd.name containsName tpnme.ANON_CLASS_NAME =>
```
lrytz: would `cd.symbol.isSynthetic` work?

adriaanm:
Bumping to 2.12.5, feel free to undo.

retronym:
Okay, looks like this special case also covers the right associative operator desugaring, and needs to accept:

```
{
  <synthetic> <artifact> val rassoc$1 = x;
  fs.$div$colon(rassoc$1)
}(((x, f) => f(x)))
```

retronym:
I've generalized the condition to deal with synthetic symbols/modifiers and retargetted to 2.13.

adriaanm:
I push -f'ed a reworked version that seems to special case enough

PR #4079, opened 2014-10-28
SI-8941 Idempotent presentation compilation of implicit classes

retronym:
When we name an implicit class, `enterImplicitWrapper` is called,
which enters the symbol for the factory method into the
owning scope. The tree defining this factory method is stowed into
`unit.synthetics`, from whence it will be retrieved and incorporated
into the enclosing tree during typechecking (`addDerivedTrees`).
The entry in `unit.synthetics` is removed at that point.

However, in the presentation compiler, we can typecheck a unit
more than once in a single run. For example, if, as happens
in the enclosed test, a call to ask for a type at a given
position interrupts type checking of the entire unit, we
can get into a situation whereby the first type checking
invocation has consumed the entry from `unit.synthetics`,
and the second will crash when it can't find an entry.

Similar problems have been solved in the past in
`enterExistingSym` in the presentation compiler. This method
is called when the namer encounters a tree that already has
a symbol attached. See 0b78a0196 / 148736c3df.

This commit takes a two pronged approach.

First, `enterExistingSym` is extended to handle implicit classes.
Any previous factory method in the owning scope is removed, and
`enterImplicitWrapper` is called to place a new tree for the factory
into `unit.synthetics` and to enter its symbol into the owning scope.

Second, the assertions that could be tripped in `addDerivedTrees`
and in `ImplicitClassWrapper#derivedSym` have been converted to
positioned errors.

The first change is sufficient to fix this bug, but the second
is also enough to make the enclosed test pass, and has been retained
as an extra layer of defence.

Review by @lrytz / @dragos

PR #5987, opened 2017-07-08
Don't lose track of imports in presentation compiler

retronym:
Fixes regression in that the reporter bisected to #4079.

I haven't reverted that change, but rather noticed that
the path through namers that enters trees that already
have symbols assigned was not extending the context chain
for imports.

Fixes scala/bug#9283

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -281,7 +281,11 @@ trait Namers extends MethodSynthesis {
       }
       tree.symbol match {
         case NoSymbol => try dispatch() catch typeErrorHandler(tree, this.context)
-        case sym      => enterExistingSym(sym, tree)
+        case sym      =>
+          tree match {
+            case tree@Import(_, _) => enterExistingSym(sym, tree).make(tree)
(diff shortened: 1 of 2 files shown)


lrytz (approved):

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -281,7 +281,11 @@ trait Namers extends MethodSynthesis {
       }
       tree.symbol match {
         case NoSymbol => try dispatch() catch typeErrorHandler(tree, this.context)
-        case sym      => enterExistingSym(sym, tree)
+        case sym      =>
+          tree match {
+            case tree@Import(_, _) => enterExistingSym(sym, tree).make(tree)
```
xeno-by, line 286: Why wasn't this leading to problems in the batch compiler?

retronym: I think `tree.symbol` is null, at least for imports, at this point in the batch compiler, which goes to the code just above in `dispatch()`.

```
case tree @ Import(_, _)                           => enterImport(tree); returnContext = context.make(tree)
```

I'll test that theory out by adding some assertions before I merge, won't have time for a few days though.

PR #5988, opened 2017-07-09
scala/bug#5961 Add Ordering composition methods

NthPortal:
Add `Ordering.orElse(Ordering[T]):Ordering[T]` and
`Ordering.orElseBy[S:Ordering](f:T=>S):Ordering[T]`

Fixes [scala/bug#5961](https://github.com/scala/bug/issues/5961)


lrytz:
```
[error] Test scala.SerializationStabilityTest.testAll failed: java.io.InvalidClassException: scala.math.Ordering$Int$; local class incompatible: stream classdesc serialVersionUID = -941572818022838553, local class serialVersionUID = -5532304942524776010, took 0.108 sec
```
https://github.com/scala/scala/blob/2.13.x/test/junit/scala/SerializationStabilityTest.scala needs to be updated

NthPortal:
@lrytz By running the sbt task listed at the top of the file?

lrytz:
Yes, that should do it!

src/library/scala/math/Ordering.scala
```diff
@@ -122,6 +122,49 @@ trait Ordering[T] extends Comparator[T] with PartialOrdering[T] with Serializabl
...
+
+  /** Given f, a function from T into S, creates an Ordering[T] whose compare
+    * function returns the result of this Ordering's compare function,
+    * if it is non-zero, or else a result equivalent to:
+    *
+    * {{{
+    * Ordering[S].compare(f(x), f(y))
+    * }}}
+    *
+    * This function is analogous to passing the result of `Ordering.by(f)`
```
SethTisue: isn't it equivalent, not merely analogous?

SethTisue:
why would adding `def`s affect serialization?

src/library/scala/math/Ordering.scala
NthPortal: Yes, I suppose it is (though slightly cheaper in terms of allocation).

I perhaps should not have borrowed the wording from `by` as much.

NthPortal:
Is there something I need to do to fix the `combined` status check?

lrytz:
Yes, you can squash all your changes into a single commit. We [keep our git history clean](https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#clean-commits-clean-history), no intermediate commits.

NthPortal:
Okay. I wasn't sure because force-pushing is generally considered bad practice.

(Also, The description for `combined` seems to indicate that the problem is previous build failures, not that isn't more than one commit in general; I don't know if someone can tweak the message.)

lrytz:
Thanks! We force push all the time, like when we screw up a release or so. Just kidding :-)

szeiger:
@SethTisue is this PR good to go for M2?

SethTisue:
> is this PR good to go for M2

I don't know... nobody has answered my question "why would adding defs affect serialization". I'd like to understand that. on the one hand, maybe the PR doesn't need to be held up over it, but on the other hand, there's no rush to get this in M2, either.

lrytz:
`Ordering` doesn't have an `@serialVersionUID`, so the uid is computed, and in a very conservative fashion: adding a public method changes the uid (https://stackoverflow.com/questions/998925/will-adding-a-method-change-the-java-calculated-serialversionuid-on-my-class). I remember researching this topic a while ago, but cannot find any comment or note or somesuch that I made.

Anyway, it would be good to add the annotation to prevent this issue in future.

lrytz:
Ah, here's my comment, in the last paragraph of the summary: https://github.com/scala/scala/pull/5429

lrytz:
And https://github.com/scala/scala-dev/issues/237

SethTisue:
@lrytz thanks

for verifying that this actually fills a hole, https://stackoverflow.com/questions/14695833/scala-idiom-for-ordering-by-multiple-criteria/14696410 was helpful

(personally, I've always used the tuple method to make composite orderings. but, this still seems like a nice addition)

SethTisue (approved):

PR #5990, opened 2017-07-10
[REVERTED] Add infer Product with Serializable linter flag

jvz:
This adds a new lint warning, -Xlint:infer-product-with-serializable
which warns when an inferred value or method type is Product with
Serializable. This is oftentimes a programmer error from mixing up what
would otherwise be incompatible types (which could be caught by the
-Xlint:infer-any flag otherwise) that have the common PwS supertype. As
this inferred type tends to be useless, this lint flag helps avoid
errors related to this inference.

Requesting review by @adriaanm in particular as this seems related to his expertise.

### DIFF
--- a/src/compiler/scala/tools/nsc/settings/Warnings.scala
+++ b/src/compiler/scala/tools/nsc/settings/Warnings.scala
@@ -73,7 +73,8 @@ trait Warnings {
     val NullaryUnit            = LintWarning("nullary-unit",              "Warn when nullary methods return Unit.",                                    true)
     val Inaccessible           = LintWarning("inaccessible",              "Warn about inaccessible types in method signatures.",                       true)
     val NullaryOverride        = LintWarning("nullary-override",          "Warn when non-nullary `def f()' overrides nullary `def f'.",                true)
-    val InferAny               = LintWarning("infer-any",                 "Warn when a type argument is inferred to be `Any`.",                        true)
+    val InferAny               = LintWarning("infer-any",                 "Warn when a type argument, variable definition or method definition is inferred to be `Any`.", true)
+    val InferPwS               = LintWarning("infer-pws",                 "Warn when a type argument, variable definition, or method definition is inferred to be `Product with Serializable`.")
     val MissingInterpolator    = LintWarning("missing-interpolator",      "A string literal appears to be missing an interpolator id.")
     val DocDetached            = LintWarning("doc-detached",              "A Scaladoc comment appears to be detached from its element.")
     val PrivateShadow          = LintWarning("private-shadow",            "A private field (or class parameter) shadows a superclass field.")
@@ -97,6 +98,7 @@ trait Warnings {
   def warnInaccessible           = lint contains Inaccessible
   def warnNullaryOverride        = lint contains NullaryOverride
   def warnInferAny               = lint contains InferAny
+  def warnInferPwS               = lint contains InferPwS
   def warnMissingInterpolator    = lint contains MissingInterpolator
   def warnDocDetached            = lint contains DocDetached
   def warnPrivateShadow          = lint contains PrivateShadow
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -551,21 +551,37 @@ trait Infer extends Checkable {
         }
       }
       val targs = solvedTypes(tvars, tparams, tparams map varianceInTypes(formals), upper = false, lubDepth(formals) max lubDepth(argtpes))
+      def warnIfInferred(warn: Type => Boolean) = {
+        if (context.reportErrors && !fn.isEmpty) {
+          targs.withFilter(warn).foreach { targ =>
+            reporter.warning(fn.pos, s"a type was inferred to be `$targ`; this may indicate a programming error.")
+          }
+        }
+      }
+      def canWarnAbout(explicitlyTyped: List[Type] => Boolean): Boolean = {
+        val loBounds = tparams map (_.info.bounds.lo)
+        val hasExplicitType = pt :: restpe :: formals ::: argtpes ::: loBounds exists (tp => explicitlyTyped(tp.dealiasWidenChain))
+        !hasExplicitType
+      }
       // Can warn about inferring Any/AnyVal as long as they don't appear
       // explicitly anywhere amongst the formal, argument, result, or expected type.
       // ...or lower bound of a type param, since they're asking for it.
-      def canWarnAboutAny = {
-        val loBounds = tparams map (_.info.bounds.lo)
-        def containsAny(t: Type) = (t contains AnyClass) || (t contains AnyValClass)
-        val hasAny = pt :: restpe :: formals ::: argtpes ::: loBounds exists (_.dealiasWidenChain exists containsAny)
-        !hasAny
+      def canWarnAboutAny = canWarnAbout(_ exists (t => (t contains AnyClass) || (t contains AnyValClass)))
+      if (settings.warnInferAny && canWarnAboutAny) {
+        warnIfInferred {
+          _.typeSymbol match {
+            case AnyClass | AnyValClass => true
+            case _ => false
+          }
+        }
       }
-      if (settings.warnInferAny && context.reportErrors && !fn.isEmpty && canWarnAboutAny) {
-        targs.foreach(_.typeSymbol match {
-          case sym @ (AnyClass | AnyValClass) =>
-            reporter.warning(fn.pos, s"a type was inferred to be `${sym.name}`; this may indicate a programming error.")
-          case _ =>
-        })
+      // Ditto for Product with Serializable
+      def canWarnAboutPwS = canWarnAbout(tps => (tps exists (_ contains ProductRootClass)) && (tps exists (_ contains SerializableClass)))
+      if (settings.warnInferPwS && canWarnAboutPwS) {
+        warnIfInferred {
+          case RefinedType(ProductRootTpe :: SerializableTpe :: _, scope) if scope.isEmpty => true
+          case _ => false
+        }
       }
       adjustTypeArgs(tparams, tvars, targs, restpe)
     }
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -1070,7 +1070,24 @@ trait Namers extends MethodSynthesis {
 
       val defnTpe = widenIfNecessary(tree.symbol, rhsTpe, pt)
       tree.tpt defineType defnTpe setPos tree.pos.focus
-      tree.tpt.tpe
+      val tpe = tree.tpt.tpe
+      // if enabled, validate that the now inferred val or def type isn't PwS
+      if (settings.warnInferPwS && context.reportErrors) {
+        tpe match {
+          case RefinedType(ProductRootTpe :: SerializableTpe :: _, scope) if scope.isEmpty =>
+            reporter.warning(tree.pos, s"a type was inferred to be `$tpe`; this may indicate a programming error")
+          case _ =>
+        }
+      }
+      // if enabled, validate the now inferred type isn't Any or AnyVal
+      if (settings.warnInferAny && context.reportErrors) {
+        tpe match {
+          case AnyTpe | AnyValTpe =>
+            reporter.warning(tree.pos, s"a type was inferred to be `$tpe`; this may indicate a programming error")
+          case _ =>
+        }
+      }
+      tpe
     }
 
     // owner is the class with the self type
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -712,6 +712,7 @@ trait Definitions extends api.StandardDefinitions {
     def tupleComponents(tp: Type)      = tp.dealiasWiden.typeArgs
 
     lazy val ProductRootClass: ClassSymbol = requiredClass[scala.Product]
+    lazy val ProductRootTpe: Type          = ProductRootClass.tpe
       def Product_productArity          = getMemberMethod(ProductRootClass, nme.productArity)
       def Product_productElement        = getMemberMethod(ProductRootClass, nme.productElement)
       def Product_iterator              = getMemberMethod(ProductRootClass, nme.productIterator)
--- a/src/reflect/scala/reflect/runtime/JavaUniverseForce.scala
+++ b/src/reflect/scala/reflect/runtime/JavaUniverseForce.scala
@@ -339,6 +339,7 @@ trait JavaUniverseForce { self: runtime.JavaUniverse  =>
     definitions.AbstractFunctionClass
     definitions.MacroContextType
     definitions.ProductRootClass
+    definitions.ProductRootTpe
     definitions.Any_$eq$eq
     definitions.Any_$bang$eq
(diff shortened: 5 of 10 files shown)


src/compiler/scala/tools/nsc/settings/Warnings.scala
```diff
@@ -74,6 +74,7 @@ trait Warnings {
     val Inaccessible           = LintWarning("inaccessible",              "Warn about inaccessible types in method signatures.",                       true)
     val NullaryOverride        = LintWarning("nullary-override",          "Warn when non-nullary `def f()' overrides nullary `def f'.",                true)
     val InferAny               = LintWarning("infer-any",                 "Warn when a type argument is inferred to be `Any`.",                        true)
+    val InferProductWithSerializable = LintWarning("infer-product-with-serializable", "Warn when a type argument is inferred to be `Product with Serializable`.")
```
hrhino: This does not do what it says it does.

```
> scala -Xlint:infer-any -Xlint:infer-product-with-serializable
[info] Running scala.tools.nsc.MainGenericRunner -usejavacp -Xlint:infer-any -Xlint:infer-product-with-serializable
Welcome to Scala 2.12.3-20170710-215443-d31a518 (OpenJDK 64-Bit Server VM, Java 1.8.0_131).

scala> def f[A](as: A*) = 12
f: [A](as: A*)Int

scala> f(1, "2")
<console>:13: warning: a type was inferred to be `Any`; this may indicate a programming error.
       f(1, "2")
       ^
res0: Int = 12

scala> f((1,2), List(3, 4))
res1: Int = 12

scala> 
```

Instead it does (as the test cases say):

```
scala> val f = if (true) (1,2) else List(3, 4)
<console>:11: warning: a type was inferred to be `Product with Serializable`; this may indicate a programming error
       val f = if (true) (1,2) else List(3, 4)
           ^
f: Product with Serializable = (1,2)

scala> 
```

which is :+1: :+1:, but personally I think it should do both of the above.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -1070,7 +1070,16 @@ trait Namers extends MethodSynthesis {
 
       val defnTpe = widenIfNecessary(tree.symbol, rhsTpe, pt)
       tree.tpt defineType defnTpe setPos tree.pos.focus
-      tree.tpt.tpe
+      val tpe = tree.tpt.tpe
+      // if enabled, validate that the now inferred val or def type isn't PwS
+      if (settings.warnInferProductWithSerializable && context.reportErrors) {
+        tpe match {
+          case RefinedType(ProductRootTpe :: SerializableTpe :: _, _) =>
```
hrhino: Possibly check that `scope.isEmpty` here? I'm unsure how it would really make a difference, though.

src/reflect/scala/reflect/internal/Definitions.scala
```diff
@@ -712,6 +712,7 @@ trait Definitions extends api.StandardDefinitions {
     def tupleComponents(tp: Type)      = tp.dealiasWiden.typeArgs
 
     lazy val ProductRootClass: ClassSymbol = requiredClass[scala.Product]
+    lazy val ProductRootTpe: Type          = typeOf[scala.Product]
```
hrhino: `ProductRootClass.tpe` seems to be more in line with what is done elsewhere in this file.

hrhino (commented):
Glad you got this going!

hrhino:
Also, having changed `Definitions.scala`, you now are morally obligated by the tests to take a trip to `JavaUniverseForce.scala` and spread the good word of `ProductRootTpe` ;)

src/compiler/scala/tools/nsc/settings/Warnings.scala
jvz: Ah yes, the problems of copying semi-related examples. I need to update the docs as well. Adding inference checks for parameters makes sense.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jvz: I'm kind of unclear of the purpose there.

src/reflect/scala/reflect/internal/Definitions.scala
jvz: Makes sense to me!

src/compiler/scala/tools/nsc/typechecker/Namers.scala
lrytz: `RefinedType` is defined as `case class RefinedType(override val parents: List[Type], override val decls: Scope)`, you could check that `decls` is empty.

jvz:
Alright, I've updated the docs, added parameter checking, did a little refactoring for infer-any to support PwS in the same way, added the type to JavaUniverseForce, added more tests for parameter inference, and checked for an empty scope. I think that about covers all the current comments.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -1070,7 +1070,16 @@ trait Namers extends MethodSynthesis {
...
+            reporter.warning(tree.pos, s"a type was inferred to be `$tpe`; this may indicate a programming error")
+          case _ =>
+        }
+      }
+      tpe
```
lrytz, line 1090: Do we still need the check here?

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -551,21 +551,37 @@ trait Infer extends Checkable {
...
+        }
       }
-      if (settings.warnInferAny && context.reportErrors && !fn.isEmpty && canWarnAboutAny) {
-        targs.foreach(_.typeSymbol match {
-          case sym @ (AnyClass | AnyValClass) =>
-            reporter.warning(fn.pos, s"a type was inferred to be `${sym.name}`; this may indicate a programming error.")
-          case _ =>
-        })
+      // Ditto for Product with Serializable
+      def canWarnAboutPwS = canWarnAbout(tps => (tps exists (_ contains ProductRootClass)) && (tps exists (_ contains SerializableClass)))
```
lrytz, line 579: could you write a few more tests that exploit this path?

lrytz (commented):
very nice work!

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jvz: The Namer check is for vals and defs, while the Infer check is for generic parameters. I can't find a suitable place in the compiler that would work for both situations without any duplication unfortunately, though that's mainly due to my unfamiliarity with the codebase.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
jvz: I can add some more that are similar to the infer-any checks, though I'm not as clear about how this parameter situation arises compared to vals and defs.

szeiger:
Is it clear that the type will always be `Product with Serializable` rather than `Serializable with Product`?

jvz:
The typically inferred type is fully `Product with Serializable with Object`, though I'm not sure what logic dictates the ordering. If it makes sense, I could add a check for `RefinedType(SerializableTpe :: ProductRootTpe :: _, _)` as well.

Also, I've pushed an update to the partest. I've added a couple more tests there.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
jvz: Almost forgot to fix this check. Updated as well.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -551,21 +551,37 @@ trait Infer extends Checkable {
...
-      def canWarnAboutAny = {
-        val loBounds = tparams map (_.info.bounds.lo)
-        def containsAny(t: Type) = (t contains AnyClass) || (t contains AnyValClass)
-        val hasAny = pt :: restpe :: formals ::: argtpes ::: loBounds exists (_.dealiasWidenChain exists containsAny)
-        !hasAny
+      def canWarnAboutAny = canWarnAbout(_ exists (t => (t contains AnyClass) || (t contains AnyValClass)))
+      if (settings.warnInferAny && canWarnAboutAny) {
+        warnIfInferred {
+          _.typeSymbol match {
+            case (AnyClass | AnyValClass) => true
```
som-snytt: The parens are extra.

som-snytt:
Dangling issue for improvement in this area https://github.com/scala/bug/issues/9211

I remember there was once a partest that occasionally printed `Serializable with Product`.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
jvz: Removed.

jvz:
I've been unable to find a situation to cause the reverse ordering of types like that, but I can change the RefinedType check to look for the opposite ordering. I'm not confident that changing them to check for any location besides the first two types would be appropriate, but I could also go that route if there aren't any issues.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
lrytz: Ah, I see. Would it make sense to add the `infer-any` check here as well?

jvz: That would be handy feature improvement. I'll go ahead and add it in.

lrytz:
Maybe you can also figure out why the following doesn't warn:

```
scala> def f(x: Boolean) = List(if(x) (1,2) else Nil)
f: (x: Boolean)List[Product with Serializable]
```

The same is true for `infer-any`

szeiger:
AFAIR we had a reversed order at some point during the migration from ant to sbt. I don't remember the exact reason, probably a change in the classpath.

jvz:
I've pushed an update that adds val and def checks for Any and AnyVal (along with neg and neg-neg test updates) along with updates to the command line docs.

src/compiler/scala/tools/nsc/settings/Warnings.scala
```diff
@@ -97,6 +98,7 @@ trait Warnings {
   def warnInaccessible           = lint contains Inaccessible
   def warnNullaryOverride        = lint contains NullaryOverride
   def warnInferAny               = lint contains InferAny
+  def warnInferProductWithSerializable = lint contains InferProductWithSerializable
```
som-snytt: I wouldn't be averse to using `warnInferPwS` everywhere. Then you preserve alignment, which is king.

som-snytt:
I was surprised the following doesn't currently warn, and I guess still won't.
```
scala> def f[A](a: A) = 42
f: [A](a: A)Int

scala> f { if (true) 17 else "" } //print
   $line3.$read.$iw.$iw.f[Any](if (true)
  17
else
  "") // : Int

scala> f { if (true) 17 else "" } //print
res0: Int = 42

```

src/compiler/scala/tools/nsc/settings/Warnings.scala
jvz: Updated to use the PwS acronym for brevity.

jvz:
Is there anything else you all would like me to change in this PR? Or can it be merged for 2.12.4?

adriaanm:
Reviewers, could one of you add the "approved" stamp?

som-snytt:
LGTM

in case that's still the stamp of approval.

adriaanm:
The newfangled github stamp is the preferred indicator these days, but it's good to see scabot can still apply reviewed labels.

som-snytt:
Maybe there's a difference between "leaving review comments" (one-offs) vs "starting a review" where the full review has the approve button?

IATM like the politicians said in ads, "I approved this ~message~ modification."

jvz:
Thanks for the reviews!

SethTisue:
under `-Xlint` this didn't used to warn:

```
trait T {
  def thunk(): Any
  def value = thunk()
}
```

but since this PR was merged (at least, I'm assuming this is the PR responsible, I didn't bisect), it gives:

```
[warn] /Users/tisue/tmp/any-warning/S.scala:3:7: a type was inferred to be `Any`; this may indicate a programming error
[warn]   def value = thunk()
[warn]       ^
```

this was caught by the Scala 2.12 community build; the new warning causes several projects with `-Xfatal-warnings` enabled to fail compilation

SethTisue:
ah, I guess this is "adds val and def checks for Any and AnyVal"

SethTisue:
@jvz can you look through the new warnings in https://scala-ci.typesafe.com/view/All/job/scala-2.12.x-integrate-community-build/1992/consoleFull and make sure that they are actually intended...? we can always temporarily disable `-Xfatal-warnings` for the failing projects in the context of the community build, figuring the project owners will address the warnings after 2.12.4 come out

SethTisue:
(there is a larger discussion to be had... not here on this PR... about how much, and what kind, of linting it's appropriate to keep adding to scalac. it would be really nice if linting were its own phase (or perhaps a compiler plugin) and were Scalafix based, rather than just inserted into compiler code everywhere)

som-snytt:
It's like having to write `case _: Throwable`, which isn't even lint.

I can attest to the great head ache of wading through community build hiccups. It's like walking on gravel in thin-soled shoes.

Was the contributor sufficiently warned? Besides the CLA, there ought to be a release form you have to fill out.

som-snytt:
The `better-files` warning is a style challenge. It needs tpolecat's `void`.

```
  def apply[A](disposeMethod: A => Any): Disposable[A] = new Disposable[A] {
    override def dispose(resource: A) = {
      val _ = disposeMethod(resource)
    }
  }

```

som-snytt:
Bookmarking:
```
[fs2] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.9/project-builds/fs2-9cd5cb7138c541c306011296c1d74d4e95332a41/core/shared/src/main/scala/fs2/AsyncPull.scala:96: a type was inferred to be `Product with Serializable with fs2.internal.FreeC[F,A]`; this may indicate a programming error.
[fs2] [error]       def get = FreeC.Eval(r.get).flatMap(_.fold(FreeC.Fail(_), FreeC.Pure(_)))
[fs2] [error]                                             ^
```
I thought something like that came in up the discussion?

jvz:
>a type was inferred to be `Product with Serializable with fs2.internal.FreeC[F,A]`

This doesn't sound correct. My check is supposed to only be finding things with just the first two types, and it wouldn't really make sense in that example to declare a type. The discussion did bring up the question about what if a type was inferred to be `Serializable with Product` instead of the other way around, though I never found that to be the case.

@SethTisue as for the thunk one, that looks like the appropriate warning being detected, though it does seem to be a bit overzealous in that you already do have an explicit type for `Any` above it. As it is, though, I do prefer to be explicit about types in public functions unless it's extremely trivial (like string constants).

Also, as to adding linter checks to the compiler, yes, I'd rather see it in a separate tool in the future that would not only be easier to modify (scalac's source code is hard to follow), though when I put this PR together, I wasn't really aware of scalacheck, wartremover, et al. Perhaps this would be a nice focus in Dotty?

jvz:
Oh, and that check for the inferred Any, I believe I added that after the PwS check. See the review comments by @lrytz

adriaanm:
I'm sorry we had to revert this (see scala/bug#10535), but the release deadline is too close to refine this in time for 2.12.4. I would love to get a refinement into 2.12.5!

PR #5991, opened 2017-07-14
Translate appropriate JVM accessor flags into annotations.

hrhino:
a.k.a., once more into the ClassfileParser fray.

I'm having enough fun on another branch with buffing up `JavaParsers` for dealing with pesky Java expression syntax, so I thought, why not go have some fun elsewhere. Thus:

In `JavaMirrors`, add these synthetic annotations along with the real ones (and the `@throws` annotation which we already synthesize in the same place). In `ClassfileParser`, inspect the access flags and apply the appropriate annotations then. 

Happily, we were already doing The Right Thing[tm] for these classes if we loaded their symbols via `JavaParsers`, so for today that file escapes unscathed.

I gave the ol' college try to doing the same for `synchronized`, but sadly, I dropped out. There's no sane way to access it from the reflection API unless you're very familiar with the internals of the compiler and love you some casts.

Fixes scala/bug#10042.

### DIFF
--- a/src/reflect/scala/reflect/internal/JavaAccFlags.scala
+++ b/src/reflect/scala/reflect/internal/JavaAccFlags.scala
@@ -1,5 +1,5 @@
 /* NSC -- new Scala compiler
- * Copyright 2005-2013 LAMP/EPFL
+ * Copyright 2005-2017 LAMP/EPFL
  * @author  Paul Phillips
  */
 package scala
@@ -12,13 +12,13 @@ import JavaAccFlags._
 import ClassfileConstants._
 
 /** A value class which encodes the access_flags (JVMS 4.1)
- *  for a field, method, or class. The low 16 bits are the same
- *  as those returned by java.lang.reflect.Member#getModifiers
- *  and found in the bytecode.
- *
- *  The high bits encode whether the access flags are directly
- *  associated with a class, constructor, field, or method.
- */
+  * for a field, method, or class. The low 16 bits are the same
+  * as those returned by java.lang.reflect.Member#getModifiers
+  * and found in the bytecode.
+  *
+  * The high bits encode whether the access flags are directly
+  * associated with a class, constructor, field, or method.
+  */
 final class JavaAccFlags private (val coded: Int) extends AnyVal {
   private def has(mask: Int) = (flags & mask) != 0
   private def flagCarrierId  = coded >>> 16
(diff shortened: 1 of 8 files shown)


lrytz:
Looks very good! In https://github.com/scala/bug/issues/10042#issuecomment-292442813 Jason mentions that JavaMirrors should add an `@throws` annotation but doesn't - is this still relevant?

hrhino:
I think someone's beaten me to it: https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/runtime/JavaMirrors.scala#L690

And indeed:

```
Welcome to Scala 2.12.1 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> import reflect.runtime.universe._

scala> typeOf[java.util.concurrent.Callable[_]].decl(TermName("call"))
res0: reflect.runtime.universe.Symbol = method call

scala> .annotations
res1: List[reflect.runtime.universe.Annotation] = List(throws[java.lang.Exception](classOf[java.lang.Exception]))

scala>
```

lrytz:
Thanks for checking! The code you link to is a lot older (https://github.com/scala/scala/commit/f1701f704a2485fcc2eb6d5d8b5d0228beddd9b3) than the comment in the ticket. So it was probably a misunderstanding from Jason's part.

For this PR, I'd like to keep it open until we cut 2.12.3 (next week), hope that's OK with you.

hrhino:
No worries! I don't actually need this fix; this is just what I do for "fun"..

lrytz:
CC @jvican, you mention [this bug being a blocker for you](https://github.com/scala/bug/issues/10042#issuecomment-292442808) - I guess you worked around it in the meantime.

jvican:
@lrytz Thanks, this is pretty useful! I worked around this at the time using the reflection API :smile:.

test/files/run/t10042/Checks_0.scala
```diff
@@ -0,0 +1,47 @@
+package test
+
+import reflect.api.Universe
+
+class Checks[U <: Universe with Singleton](universe: U) {
```
retronym, line 5: Great way to share the test!

src/reflect/scala/reflect/internal/JavaAccFlags.scala
```diff
@@ -56,6 +56,19 @@ final class JavaAccFlags private (val coded: Int) extends AnyVal {
...
+    */
+  def toScalaAnnotations(syms: SymbolTable): List[syms.AnnotationInfo] = {
+    import syms._
+    def annInfo(asym: ClassSymbol) = AnnotationInfo(asym.tpe, Nil, Nil)
+    var anns: List[AnnotationInfo] = Nil
+    if (isNative)    anns ::= annInfo(definitions.NativeAttr)
+    if (isTransient) anns ::= annInfo(definitions.TransientAttr)
+    if (isVolatile)  anns ::= annInfo(definitions.VolatileAttr)
+    anns
+  }
```
retronym, line 71: Thanks for sharing this code between the compiler and reflective universes.

retronym (approved):

PR #5992, opened 2017-07-14
Fix for scala/bug#5946.

milessabin:
This also turns the crasher reported in https://github.com/scala/bug/issues/4225 into an error.

Whilst this does the right thing, it's not clear to me that the failing case shouldn't have been caught further up the call chain.

I also attempted to modify and use `mkAttributedQualifierIfPossible` in `mkAttributedRef`, however this appears to reject cases which `mkAttributedQualifier` is happy to handle, and it's not clear to me if this is because the conditions in `mkAttributedQualifierIfPossible` are incorrect or if it has some more specific use than its name suggests.

I'd be more than happy to forward port this to 2.13.x.

### DIFF
--- a/src/reflect/scala/reflect/internal/TreeGen.scala
+++ b/src/reflect/scala/reflect/internal/TreeGen.scala
@@ -143,11 +143,15 @@ abstract class TreeGen {
 
   /** Builds a reference to given symbol with given stable prefix. */
   def mkAttributedRef(pre: Type, sym: Symbol): RefTree = {
-    val qual = mkAttributedQualifier(pre)
-    qual match {
-      case EmptyTree                                  => mkAttributedIdent(sym)
-      case This(clazz) if qual.symbol.isEffectiveRoot => mkAttributedIdent(sym)
-      case _                                          => mkAttributedSelect(qual, sym)
+    pre match {
+      case TypeRef(NoPrefix, _: TypeSkolem, Nil) => mkAttributedIdent(sym)
+      case _ =>
+        val qual = mkAttributedQualifier(pre)
+        qual match {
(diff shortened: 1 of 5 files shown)


src/reflect/scala/reflect/internal/TreeGen.scala
```diff
@@ -143,11 +143,15 @@ abstract class TreeGen {
 
   /** Builds a reference to given symbol with given stable prefix. */
   def mkAttributedRef(pre: Type, sym: Symbol): RefTree = {
-    val qual = mkAttributedQualifier(pre)
-    qual match {
-      case EmptyTree                                  => mkAttributedIdent(sym)
-      case This(clazz) if qual.symbol.isEffectiveRoot => mkAttributedIdent(sym)
-      case _                                          => mkAttributedSelect(qual, sym)
+    pre match {
+      case TypeRef(NoPrefix, _: TypeSkolem, Nil) => mkAttributedIdent(sym)
```
adriaanm, line 147: I apologize for not leaving a comment here earlier. I was thinking about a nicer way of doing this, but rather than wait until I have a concrete suggestion, I'll already say this special case makes me uneasy. Why does a skolem show up? Should the type have been deskolemized before we pass it to `mkAttributedRef`?

adriaanm:
Could you please refine/expand your commit messages? The title should be short and prescriptive/imperative (don't mention tickets here), then a body with the description (why/how/known limitations). End with references to tickets.

adriaanm:
I'd like to get to the root of this before we merge it, pushing to 2.12.5

milessabin:
Obsoleted by https://github.com/scala/scala/pull/5999.

PR #5993, opened 2017-07-15
No noisy options for REPL eval compilation

som-snytt:
As for REPL startup code, disable noisy options
when compiling the uninteresting `eval` object
used for printing results. (Unless debug is on.)

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/IMain.scala
+++ b/src/repl/scala/tools/nsc/interpreter/IMain.scala
@@ -105,6 +105,29 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends 
     try body
     finally if (!saved) settings.nowarn.value = false
   }
+  def withSuppressedSettings[A](body: => A): A = {
+    val ss = this.settings
+    import ss._
+    val wasWarning = !nowarn
+    val noisy = List(Xprint, Ytyperdebug, browse)
+    val current = (Xprint.value, Ytyperdebug.value, browse.value)
+    val noisesome = wasWarning || noisy.exists(!_.isDefault)
+    if (isReplDebug || !noisesome) body
+    else {
+      Xprint.value = List.empty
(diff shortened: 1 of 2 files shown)


src/repl/scala/tools/nsc/interpreter/IMain.scala
```diff
@@ -105,6 +105,29 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends
     try body
     finally if (!saved) settings.nowarn.value = false
   }
+  def withSuppressedSettings[A](body: => A): A = {
```
lrytz: Can we avoid the duplication with https://github.com/scala/scala/blob/v2.12.2/src/repl/scala/tools/nsc/interpreter/ILoop.scala#L958 ?

som-snytt: At first I was just lazy; the other diff is that we want warnings (?) in one case but not the other. Probably I was just lazy.

som-snytt:
Refactored at some cost of fragility. Just a convenience thing, it's OK if prefer to ignore repl tweaks at this point.

adriaanm:
Happy to merge this here. From now on, 2.13  is a better target for repl improvements.

som-snytt:
@adriaanm Thanks, I'll just open regression tickets against 2.13. Kidding! Sorry to waste your PR queue time with little stuff that bugs only me and a handful of 2.12 users.  I try to be scala-team-friendly. Also I try not to @adriaanm every time I'm on twitter or gitter and think, Hey I bet @adriaanm would think this is funny.

adriaanm:
The only time wasted here is me trying to come up with witty replies. Luckily there's no drafts folder for my failed github comments to collect and haunt me.

adriaanm:
This won't merge forward to 2.13.x -- will need a separate port if we need these changes there.

PR #5995, opened 2017-07-16
Add Option.{when,unless}

japgolly:
I use these all the time and find them very useful and a significant improvement to readability at the call-site. They're really tiny and so I thought this would be a good first contribution to stdlib. :)

### DIFF
--- a/src/library/scala/Option.scala
+++ b/src/library/scala/Option.scala
@@ -28,6 +28,19 @@ object Option {
    *  the collections hierarchy.
    */
   def empty[A] : Option[A] = None
+
+  /** When a given condition is true, evaluates the `a` argument and returns
+   *  Some(a). When the condition is false, `a` is not evaluated and None is
+   *  returned.
+   */
+  def when[A](cond: Boolean)(a: => A): Option[A] =
+    if (cond) Some(a) else None
+
+  /** Unless a given condition is true, this will evaluate the `a` argument and
+   *  return Some(a). Otherwise, `a` is not evaluated and None is returned.
+   */
+  @inline def unless[A](cond: Boolean)(a: => A): Option[A] =
+    when(!cond)(a)
 }
 
 /** Represents optional values. Instances of `Option`


lrytz:
Without feeling too strongly about it, I wouldn't add two methods just to save those 7 characters. The `if/else` also reads more naturally to me.

```
Option.when(cond)(x)
if (cond) Some(x) else None
```

Though for more complex expressions the code becomes more readable, so it might be good to have around:

```
if (!cond) None else {
  ...
  ...
  Some(....)
}

Option.when(cond) {
   ...
   ...
   ...
}
```

som-snytt:
Should it be `if (cond) Option(a)` to behave like `Option(a).filter(_ => cond)`?

SethTisue:
> to behave like `Option(a).filter(_ => cond)`

you mean w/r/t null?

I don't think you mean exactly like that, since we don't always want to evaluate `a`

Ichoran:
@lrytz - The problem with `if/else` is that it doesn't chain nicely.  You have stuff like

```scala
if (p)
  Some(x).filter(q).map(f).flatMap(y =>
    if (r(y)) Some(g(y))
    else None
  )
else None
```

instead of

```scala
Option.when(p)(x).filter(q).map(f).
  flatMap(y => Option.when(r(y))(g(y)))
```

It's not the 7 characters that matter, it's that you don't have to come back to the `else` case.

That said, I'm not 100% convinced that this is the right way to do it.  Because you can also envision

```scala
p.toOption(x).filter(q).map(f).
  flatMap(y => r(y) toOption g(y))
```

which flows even more clearly IMO.

And there's also a strong argument for letting people choose their own library to handle this sort of thing.

szeiger:
Moving to M3 since there's still some discussion needed (and I forgot to change the target milestone last Friday...)

SethTisue:
I don't think the `p.toOption(x)...` thing reads well or is idiomatic in Scala (adding methods on `Boolean`), I think the `Option.when/unless` approach is much nicer

som-snytt:
This would be a great moment to bring back the Scalawags podcast to discuss.

Ichoran:
@SethTisue - Well, I'd probably call `toOption` `selects` or `rejects`, so you get

```scala
(p selects x).filter(q).map(f).
  flatMap(y => r(y) selects g(y))
```
I think that reads well but I agree it's not as idiomatic as `Option.when`.  `Option.when` is a great deal better than `if/else` in some scenarios.  (Worse in others.)

som-snytt:
Scalawags + Ichoran. I know the Scalawags were always slick fast talkers, but a slow episode with Ichoran would be really cool.

japgolly:
Yes, this isn't about saving 7 characters, it's about saving cognitive effort and improving clarify in certain situations. Now it's not suitable to all situations in which it's technically applicable, but I continue to come across many in which it provides a significant benefit to readability. It's closer to natural language, I've never asked my wife *"If they have orange juice, get me some, else get me none"*, I ask *"If they have orange juice, get me some."* :smile:

Here are some examples from a quick scan of my code:
```scala
submitButton("Login", Option.when(p.formEnabled is Enabled)(attemptLogin))

s => Option.when(s.header.userId != userId)(AccessDenied)

Option.when(waitTimes.nonEmpty)((waitTimes.head, Retries(waitTimes.tail)))

ViewsMenu.Component(Option.unless(mode == Mode.EmptyProject)(p.filterDead)),

  private val dispatcher: DispatchLogic.Request => F[DispatchLogic.Response] =
    logic.cacheUsualPaths(
      ( logic.main
      | Option.when(Props.testMode)(logic.loginApi)
      | Option.when(Props.devMode)(logic.quickDev).flatten
      ).withFallback(logic.fallback)
    )
```

Also, to be preemptive for a minute, I know this can be implemented in userland (as I've already done for myself); to me, this PR is about providing all Scala users with the ability, out-of-the-box, to be clearer and more descriptive in their own code when these cases comes up. They seem to come up for me often enough which is why I'm making the assumption that it will be similar for others too.

retronym:
Surveying the landscape for prior art.

```scala
object Either {
  def cond[A, B](test: Boolean, right: => B, left: => A): Either[A, B] =
    if (test) Right(right) else Left(left)
}

object PartialFunction {
  def cond[T](x: T)(pf: PartialFunction[T, Boolean]): Boolean = pf.applyOrElse(x, constFalse)
  def condOpt[T,U](x: T)(pf: PartialFunction[T, U]): Option[U] = pf.lift(x)
```

Nothing quite analogous in `Try` / `Future`, but we should consider whether/how we can consistently extend this concept to those types.

nafg:
@Ichoran The thing is, there's no "advance warning" that a method on Boolean would return Option. Maybe it should select or reject a Future, or something I (as a newbie) don't even know about. I think Option.{when,unless} is much more discoverable and self-explanatory.

ritschwumm:
i've been using something like this for years now, and i'm still happy with it.
it's short, readable (especially without dots and parens), and doesn't evaluate the value that goes into the Option/Either until necessary.
guard and prevent probably deserve better names, though.


```scala
implicit final class BooleanExt(peer:Boolean) {
    def guard[T](trueValue: =>T):Option[T] =
            if (peer)   Some(trueValue)
            else        None
    
    def prevent[T](falseValue: =>T):Option[T] =
            if (!peer)  Some(falseValue)
            else        None
    
    def either[U,V](falseLeft: =>U, trueRight: =>V):Either[U,V] =
            if (peer)   Right(trueRight)
            else        Left(falseLeft)
}
```

shawjef3:
I very much want this method. This definitely isn't about 7 characters. It's about saving 7 characters each time I want to do this.

@ritschwumm  I would rather have those methods in the Option and Either objects, since those are the natural places for functions that create Options and Eithers, respectively.

shawjef3:
> Should it be if (cond) Option(a) to behave like Option(a).filter(_ => cond)?

This would actually work if you used LazyOption from Scalaz or another package. Perhaps that's why scalaz.Option doesn't have the method in this pull request.

adriaanm:
I like this addition. I also agree we should try to align the naming across Either/Future/Option, but maybe it's those old types that should adjust to the new guard. `when` / `unless` read well.

I'm +1 on merging this PR and get this incremental progress into M3. Can I suggest one more crazy idea, though? 😁  How about parametrizing `if` on the "zero-plus-m***d" you'd like to use? I've often been annoyed by how unwieldy it gets to do lifted conditionals.

  - `if[Option] (cond) x` == `if (cond) Some(x) else None`
  - `if[Either] (cond) x else y` == `if (cond) Right(x) else Left(y)`
  - `if[List] (cond) x ` == `if (cond) x :: Nil else Nil`
  - `if[NotNull] (cond) x ` == `if (cond) x else null` 😓 

Tbd: chaining, type inference,...

adriaanm (approved):
Thanks for your patience. Congrats on your first addition to the stdlib!

adriaanm:
I would welcome a follow-up to implement retronym's suggestion of aligning names in the other abstractions, or at least update the docs to cross reference them.

japgolly:
Thank you for the merge! Happy to contribute!!

PR #5996, opened 2017-07-17
Make `completingStack` global and use `ArrayBuffer`

jvican:
Replace the use of `List` by the use of an already allocated array
buffer. I'm curious if this will make a difference (most probably not),
but since it's in a such a delicate part of the compiler it's better to
rest assured I guess...


retronym (approved):

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -701,14 +703,14 @@ abstract class UnPickler {
       *
       * Useful for reporting on stub errors and cyclic errors.
       */
-    private var completingStack = List.empty[Symbol]
+    private var completingStack = new mutable.ArrayBuffer[Symbol](128)
```
retronym: When compiling the compiler, no instance of this stack appears to grow beyond length 1. In a REPL session I found a case where the stack had two elements. So 128, the initial size below of the `ArrayBuffer`, seems too large.

I was somewhat surprised to see that we maintain a separate stack for each `Unpickler.Scan`, rather than having a global stack. Could you comment on why that is?

retronym (changes requested):
Retracting approval while we discuss the scope of `completingStack` and the initial size.

adriaanm:

jvican:
Will update before that.

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
jvican: > I was somewhat surprised to see that we maintain a separate stack for each Unpickler.Scan, rather than having a global stack. Could you comment on why that is?
  
This is an oversight... Last commit makes it global.

jvican:
This has been updated as per your feedback @retronym.

adriaanm:

src/reflect/scala/reflect/internal/pickling/UnPickler.scala
```diff
@@ -45,6 +45,12 @@ abstract class UnPickler {
     }
   }
 
+  /** Keep track of the symbols pending to be initialized.
+    *
+    * Useful for reporting on stub errors and cyclic errors.
+    */
+  private var completingStack = new mutable.ArrayBuffer[Symbol](24)
```
retronym: Should be a `val`, otherwise LGTM

jvican: Done. Sorry for the oversight.

adriaanm:
conditional LGTM met :-)

PR #5999, opened 2017-07-19
Stabilize receiver of extension method application so that implicits accessible via the prefix can be candidate implicit arguments

### LINKED ISSUE #4225: Implicit Lookup fails for implicit view on nested type.
```scala
trait Foo {
 val name : String
 trait Bar {
   def doSomething : String
   override def toString = "Foo(" + name + ").Bar -> " + doSomething
 }
 object Bar {
   implicit def fromFunction(func : Function0[String]) = new Bar {
     def doSomething = func()
   }
 }
 def andThen(b : Bar) = b
}

object Foo {
  def foo(n : String) = new Foo {
    override val name = n
  }
}
```

And then run in the REPL
```scala
Welcome to Scala version 2.8.1.final (OpenJDK 64-Bit Server VM, Java 1.6.0_20).

scala> Foo.foo("foo") andThen (() => "Bar")
java.lang.AssertionError: assertion failed
	at scala.Predef$$.assert(Predef.scala:77)
	at scala.tools.nsc.ast.TreeGen.mkAttributedQualifier(TreeGen.scala:82)
	at scala.tools.nsc.ast.TreeGen.mkAttributedQualifier(TreeGen.scala:45)
	at scala.tools.nsc.ast.TreeGen.mkAttributedRef(TreeGen.scala:107)
	at scala.tools.nsc.ast.TreeGen.mkAttributedStableRef(TreeGen.scala:149)
	at scala.tools.nsc.ast.TreeGen.mkAttributedQualifier(TreeGen.scala:60)
	at scala.tools.nsc.ast.TreeGen.mkAttributedQualifier(TreeGen.scala:45)
	at scala.tools.nsc.typechecker.Implicits$$ImplicitSearch.typedImplicit0(Implicits.scala:441)
	at scala.tools.nsc.typechecker.Implicits$$ImplicitSearch.typedImplicit(Implicits.scala:359)

```

=== What is the expected behavior? ===

I expect the implicit lookup to succeed in finding an implicit view for foo.Bar.

=== What do you see instead? ===

A stack trace

=== Additional information ===
(for instance, a link to a relevant mailing list discussion)

=== What versions of the following are you using? ===
  - Scala:  2.8.1
  - Java:   OpenJDK  1.6.0_20
  - Operating system:  Ubuntu

milessabin:
This fixes scala/bug#4225 and scala/bug#5946 by introducing a synthetic val for the unstable LHS of an implicit method application where the method type depends on the LHS.

As applied to this example, taken from scala/bug#4225,
```scala
class Foo {
  class Bar
  object Bar {
    implicit def mkBar: Bar = new Bar
  }
}

object f extends Foo

implicit class Ops[F <: Foo](val f0: F) {
  def op(i: Int)(implicit b: f0.Bar): f0.Bar = b
}

f.op1(23)
```
this transformation produces,
```
{
  val fresh = Ops[f.type](f)
  fresh.op1(23)
}
```
Because `fresh` is stable we retain the singleton type `F` and the implicit argument of type `f0.Bar` can be resolved against the implicit scope of `f.type`.

A real world example of this is the macro API issue reported in scala/bug#5946 ... following this PR the usage below compiles as expected,
```scala
class Ops(val g: scala.reflect.api.JavaUniverse) {
  def op[T: g.TypeTag] = ()
}
implicit def Ops(g: scala.reflect.api.JavaUniverse): Ops = new Ops(g)
scala.reflect.runtime.universe.op[Int]
```

In principle this transformation could also be applied for explicit applications, however there would no gain: to be able the write the explicit argument the user would already have had to explicitly introduce the necessary val.


retronym:
Minor admin note: could you please retitle the commit/PR with a description of the change, rather than the ticket reference? We've been leaving the ticket reference as a `Fixes scala/...#NNNN` in the PR description and commit message.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,45 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
               }
             )
-            else
-              typer1.typed(typer1.applyImplicitArgs(tree), mode, pt)
+            else {
+              if(tree.symbol != null && !tree.symbol.isConstructor) {
+                def stabilizeQual(tree: Tree): Option[(Tree, ValDef)] = tree match {
+                  case Select(qual, name) if qual.tpe.isStable => None
+                  case Select(qual, name) =>
+                    val vsym = context.owner.newValue(unit.freshTermName(), context.owner.pos, SYNTHETIC)
```
retronym: I think the `qual.pos.focus` will be more precise than `context.owner.pos` for the synthetic tree.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,45 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                    vsym.setInfo(qual.tpe)
+                    val vdef = ValDef(vsym, qual)
+                    val newQual = Ident(vsym).setType(qual.tpe)
+                    val newSelect = makeAccessible(treeCopy.Select(tree, newQual, name), tree.symbol, singleType(NoPrefix, vsym), newQual)._1
+                    if(newSelect.tpe.contains(vsym))
+                      Some((newSelect, vdef))
```
retronym: I don't see any `atPos` or `typedPos` that positions the synthetic trees, but I guess we can discuss the required positioning cleanups after we discuss the actual functional change.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,45 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                      val PolyType(tparams, restpe) = newLhs.tpe
+                      checkBounds(tree, NoPrefix, NoSymbol, tparams, targs, "")
+                      val typeApplyTpe = restpe.instantiateTypeParams(tparams, targs)
+                      val newTypeApply = treeCopy.TypeApply(tree, newLhs, args) setType typeApplyTpe
+                      (newTypeApply, vdef)
+                    }
+                  case _ => None
+                }
+
+                stabilizeQual(tree).map { case (newLhs, vdef) =>
```
retronym: I wonder if we can use `mt.isDependentMethodType` to drive whether the old or new logic is used.  I still don't have a clear picture of the problem or solution, so just brainstorming a little...

test/files/pos/t4225.scala
```diff
@@ -0,0 +1,22 @@
+object Test {
+  class Foo {
+    class Bar
+    object Bar {
+      implicit def mkBar: Bar = new Bar
+    }
+  }
+
+  implicit class Ops[F <: Foo](val f0: F) {
```
retronym: A test variant with a by-name implicit `f0` here might show up a problem with evaluation order by hoisting out the val. We deal with the same sort of issue in the named/default args transform, we would need to share some of that implementation in this case.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,45 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                      val targs = mapList(args)(treeTpe)
+                      val PolyType(tparams, restpe) = newLhs.tpe
+                      checkBounds(tree, NoPrefix, NoSymbol, tparams, targs, "")
+                      val typeApplyTpe = restpe.instantiateTypeParams(tparams, targs)
+                      val newTypeApply = treeCopy.TypeApply(tree, newLhs, args) setType typeApplyTpe
+                      (newTypeApply, vdef)
+                    }
+                  case _ => None
+                }
+
```
lrytz: Could you add some comment "rewrites `expr` to `fxpr` because ..., see scala/bug#X"? It will save time for people driving by here in the future.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Thanks for the pointer ... I'll investigate.

milessabin: That won't help here. The method type contains the prefix, but it's not a dependent method in the sense which `isDependentMethodType` is testing for.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Thanks.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Ditto ...

test/files/pos/t4225.scala
milessabin: I've added some test variants which cover by name arguments (for the ops class and for the ops methods) and (by name) right associative ops and they don't cause any problems. Do you have a particular scenario in mind which you think might be problematic?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Will do ...

milessabin:
@retronym PR title and description tweaked as requested.

soronpo:
@milessabin Forgive me if this is a stupid question, but is it possible to implement the same mechanism of synthetic val to resolve https://github.com/milessabin/shapeless/issues/738 with `ValueOf` in TLS ?

adriaanm:
The commit messages are still empty. Please start with a prescriptive/imperative title (to keep it short, don't mention tickets here), then a body with the description. End with references to tickets.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,50 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                 }
               }
             )
-            else
-              typer1.typed(typer1.applyImplicitArgs(tree), mode, pt)
+            else {
```
lrytz: This is the `else` expr of `if (original != EmptyTree && pt != WildcardType)`. The `if` expr is not obvious to me - do you understand what it does and why we need the two branches? And why / if your fix is not needed in the `if` case?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,50 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+              if(tree.symbol != null && !tree.symbol.isConstructor) {
```
lrytz: I guess you observed the `null` case, maybe under erroneous trees? A comment would help.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,50 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                def stabilizeQual(tree: Tree): Option[(Tree, ValDef)] = tree match {
```
lrytz: Could the transformation be done using this?

```
  case treeInfo.Applied(fun @ Select(qual, name), targs, argss) => ...
```

I also wonder if assigning the types manually to the Select / Apply / TypeApply nodes is necessary. Could we just build the trees and run the typer? Probably the new synthetic value would have to be entered in the typer's scope. Also, the select's symbol might need to be set manually to prevent going into overloading resolution. Similar case for named/default args: https://github.com/scala/scala/blob/v2.13.0-M2/src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala#L173-L189. See also https://github.com/scala/scala/pull/558.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,50 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                      val PolyType(tparams, restpe) = newLhs.tpe
+                      checkBounds(tree, NoPrefix, NoSymbol, tparams, targs, "")
+                      val typeApplyTpe = restpe.instantiateTypeParams(tparams, targs)
+                      val newTypeApply = treeCopy.TypeApply(tree, newLhs, args) setType typeApplyTpe
+                      (newTypeApply, vdef)
+                    }
+                  case _ => None
+                }
+
+                // Rewrites "qual.name ..." to "{ val lhs = qual ; lhs.name ... }" in cases where
```
lrytz: Add a reference to "scala/bug#5946"

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -880,8 +880,50 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                def stabilizeQual(tree: Tree): Option[(Tree, ValDef)] = tree match {
+                  case Select(qual, name) if qual.tpe.isStable => None
+                  case Select(qual, name) =>
+                    val vsym = context.owner.newValue(unit.freshTermName(), qual.pos.focus, SYNTHETIC | ARTIFACT)
+                    vsym.setInfo(qual.tpe)
+                    val vdef = ValDef(vsym, qual)
+                    val newQual = Ident(vsym).setType(qual.tpe)
+                    val newSelect0 = atPos(tree.pos)(treeCopy.Select(tree, newQual, name))
+                    val newSelect = makeAccessible(newSelect0, tree.symbol, singleType(NoPrefix, vsym), newQual)._1
+                    if(newSelect.tpe.contains(vsym))
```
lrytz: So this is to identify whether the transformation is necessary, right? Do you see a way to check this beforehand?

lrytz (commented):
The example here https://github.com/scala/bug/issues/4225#issuecomment-292389886 still doesn't compile

```
class Foo {
  class Bar
  object Bar {
    implicit def fromString(a: String) = new Bar
  }
  def andThen(b : Bar) = b
}

object Test {
  (new Foo) andThen ("Bar")
}
```

gives

```
Test.scala:10: error: type mismatch;
 found   : String("Bar")
 required: _1.Bar where val _1: Foo
  (new Foo) andThen ("Bar")
                     ^
one error found
```

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: If we have the types available to assign why would we run the typer? Wouldn't that be more expensive?

lrytz: It seems to me that inlining pieces of `typedSelect` and `typedApply` makes the code unnecessarily long / harder to maintain. I don't think we have a performance issue here, we'd just run through a few methods of the typer.

milessabin:
@lrytz you're right that this didn't in fact fix SI-4225, although it did fix some related issues.

I've now completely reworked the fix and it now solves both the issues this PR originally addressed and SI-4225 proper. The earlier commits are obsolete ... I should probably just squash this.

I think the implementation is now a lot simpler ... the problematic cases are detected directly in `typedSelect` (in effect beforehand, as you suggested) and then we walk up the context tree to find the appropriate place to insert the stabilizing val. The duplication of `typedSelect` and `typedApply` that you were unhappy about have also now gone.

milessabin:
The broadened scope of this fix means that one existing `neg` test has moved to `pos`: test/files/pos/t3995.scala ... the test is in fact recording an instance of 4255. Also a couple of `neg` check files have had to be updated because error messages now relate to a synthetic stabilizing val rather than an existential.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4959,6 +4959,37 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                  case _ => false
+                }
+                isInsertionNode(ctx.tree) && !isInsertionNode(ctx.outer.tree)
+              }
+
+              if(insertionContext != NoContext) {
+                val vsym = insertionContext.owner.newValue(unit.freshTermName(nme.STABILIZER_PREFIX), qual.pos.focus, SYNTHETIC | ARTIFACT | STABLE)
+                vsym.setInfo(qual.tpe)
+                insertionContext.scope enter vsym
+                val vdef = ValDef(vsym, qual)
```
lrytz: I think we need a `changeOwner` here, for symbols defined within the `qual`, similar to named / default arguments: https://github.com/scala/scala/blob/v2.13.0-M2/src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala#L173-L189

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4959,6 +4959,37 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                vdef.pos.makeTransparent
```
lrytz: It seems to me that this just computes a new `Position` value and discards it.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4959,6 +4959,37 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                def isInsertionNode(tree: Tree): Boolean = tree match {
+                  case _: Apply | _: TypeApply | _: Select => true
+                  case _ => false
+                }
+                isInsertionNode(ctx.tree) && !isInsertionNode(ctx.outer.tree)
+              }
+
+              if(insertionContext != NoContext) {
+                val vsym = insertionContext.owner.newValue(unit.freshTermName(nme.STABILIZER_PREFIX), qual.pos.focus, SYNTHETIC | ARTIFACT | STABLE)
+                vsym.setInfo(qual.tpe)
```
lrytz: I think we might have the same issue as https://github.com/scala/bug/issues/7694 here, fixed in named / default args with the `uncheckedBounds` annotation (https://github.com/scala/scala/blob/v2.13.0-M2/src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala#L173-L189)

```
scala> :pa -raw

trait A
trait B

class Foo[A2, B2 <: A2] {
  class Bar
  object Bar {
    implicit def fromString(a: String) = new Bar
  }
  def andThen(b : Bar) = b
}

object Test {
  def lub = if (true) (null: Foo[A, A]) else (null: Foo[B, B])
  (lub) andThen ("Bar")
}

[[syntax trees at end of                     typer]] // <pastie>

    def lub: Foo[_ >: A with B <: Object, _ >: A with B <: Object] = if (true)
      (null: Foo[A,A])
    else
      (null: Foo[B,B]);
    {
      <synthetic> <stable> <artifact> val stabilizer$1: Foo[_1,_2] = Test.this.lub;
      stabilizer$1.andThen(stabilizer$1.Bar.fromString("Bar"))
    }


         (lub) andThen ("Bar")
          ^
<pastie>:14: error: type arguments [_1,_2] do not conform to class Foo's type parameter bounds [A2,B2 <: A2]
```

lrytz (commented):
Nice, this is definitely much nicer than the previous solution!

adriaanm:
sjrd makes some good points at https://github.com/scala/bug/issues/10619, which we should keep in mind while reasoning about this change. Also, does it interact ok with the various rewrites we do to method invocations? (by-name, prefix, defaults)

I'm a little uneasy with this change, as we're basically adding a pretty ad-hoc form of wildcard capture (by giving the unstable prefix a name). Maybe that's ok, but we should think about this carefully in those terms. The Java type system and its rules around wildcard capture are pretty hard to follow.

sjrd:
@adriaanm I don't think the points I make in https://github.com/scala/bug/issues/10619 port over to this PR. There is one fundamental difference between this PR and the suggestions in the other issue: in this PR, we are only applying *term rewritings* that *preserve evaluation order*. In the other issues, the comments suggest to take a *path* from a *type* (where no evaluation order is even *defined*) and move it to a term position, thereby inventing some evaluation context for the term.

I do not see any problem with this PR as far as semantics are concerned.

test/files/pos/t4225.scala
```diff
@@ -0,0 +1,55 @@
...
+      def op2(i: Int)(j: Boolean)(implicit b: f0.Bar): f0.Bar = b
+      def op3[T](implicit b: f0.Bar): f0.Bar = b
+      def op4(i: => Int)(implicit b: f0.Bar): f0.Bar = b
+    }
+
+    f.op0
+    f.op1(23)
+    f.op2(23)(true)
+    f.op3[Int]
+    f.op4(23)
```
retronym: Could we introduce a `run` test to show how by-name params interact with this lifting? I'd imagine that we'd need to lift a def or a val with a `Function0` to preserve eval semantics.

retronym: If the qualifier is passed as a by-name arg to the implicit conversion, it doesn't seem suitably stable for a path (or equivalently, to be liftable to a val)

milessabin: Agreed. That would have to be excluded.

adriaanm:
I agree the intent preserves eval order, but, if I understood correctly, retronym identified one counter-example / needed exception. I'd like to see tests for all variations listed in my previous message (by-name, prefix, as well as defaults, just in case). For robustness, how about weird unstable things like a block that contains a definition? `{ class Foo ; qual }.m`?

The other aspect of the change is that we are offering a limited form of wildcard capture with this rewrite. What if you have a long chain of method invocations that now results in a bunch of extra local vals that you don't want but can't disable? How about the recursion involved with rewriting   chained calls?

lrytz:
There are some open comments: https://github.com/scala/scala/pull/5999#pullrequestreview-76943485 and here https://github.com/scala/scala/pull/5999#issuecomment-348281025

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Thanks for the pointer.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Gotcha.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
milessabin: Hmm ... you're right.

milessabin:
Also rebased.

milessabin:
@adriaanm 
> The other aspect of the change is that we are offering a limited form of wildcard capture with this rewrite. What if you have a long chain of method invocations that now results in a bunch of extra local vals that you don't want but can't disable? How about the recursion involved with rewriting chained calls?

A stabilizing val is _only_ inserted if necessary to make the term type check, so I don't think this is an issue. Prior to this PR the long chain you've described wouldn't compile without you manually inserting vals. Post this PR the long chain compiles, and there's nothing to prevent you from splitting it up as you would have been forced to previously.

test/files/pos/t4225.scala
milessabin: It's not actually the argument to the implicit conversion which must be strict, it's the argument of the extension method class. That ought to be prohibited as it stands, but it looks as though it's falling foul of https://github.com/scala/bug/issues/10619. The following is a small reproduction for 2.12.4 and 2.13.x (without this PR) that's relevant here,

```scala
scala> class Foo { class Bar }
defined class Foo

scala> class Unsound[F <: Foo](f: => F) { def quux(a: f.Bar): f.Bar = a }
defined class Unsound
```
The two occurrences of `f.Bar` in `quux` should not be treated as having identical prefixes.

That's orthogonal to this PR though, and I can't add a neg test for the scenario you're worried about yet because it would be frustrated by https://github.com/scala/bug/issues/10619.

milessabin:
Added some additional tests intended to address @retronym and @adriaanm's concerns.

lrytz (changes requested):
LGTM. Could you please:
  - squash the changes into a single commit
  - add a right-associative operator to the test for evaluation order, like:

```
scala> class Foo {
     |   class Bar
     |   object Bar {
     |     implicit def fromString(a: String) = new Bar
     |   }
     |   def andThen_:(b : Bar) = b
     | }

scala> { println("bar"); "Bar" } andThen_: { println("foo"); new Foo }
bar
foo
res0: Foo#Bar = Foo$Bar@769bd849

scala> { println("foo"); new Foo }.andThen_:({ println("bar"); "Bar" })
foo
bar
res1: Foo#Bar = Foo$Bar@2e71240b
```

lrytz:
Oh, and also add a version with `andThen_:` by-name; it works thanks to https://github.com/scala/scala/pull/5969.

milessabin:
@lrytz done. Note that your last request revealed a bug in byname arguments of right associative operators: https://github.com/scala/bug/issues/10693.

milessabin:
Gah ... forgot the check file ...

lrytz:
Could you update the commit message to add a short title line? https://github.com/scala/scala/blob/2.13.x/CONTRIBUTING.md#clean-commits-clean-history

milessabin:
Done. Sorry about that ... I must have lost the title when I combined the messages.

lrytz (approved):
Thank you!

PR #6000, opened 2017-07-20
Failing scalacheck tests fail build again

retronym:
Avoid forking Scalacheck tests when run by SBT to avoid
a bug in Scalacheck itself that doensn't report failures.
We've been exposed to this since the refactoring that stopped
using partest to manage scalacheck test execution.

Address the failures we'd accumulated during this time:

  - A performance driven change to `Symbol#allOverridingSymbols`
    reversed the order of the result, which didn't matter in the
    compiler but is material in Scaladoc and the IDE. I've reworked
    this to restore the previous order.
  - After our JIRA to GitHub issues migration, test files in the
    project went through a big name to replace the `SI_` prefix with `t`.
    We also needed to update references to those filenames in the tests.
    For consistency, I've also updated the class names within the files
    to correspond to the file names.

Fixes scala/scala-dev#282

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -579,7 +579,7 @@ lazy val scalacheck = project.in(file("test") / "scalacheck")
   .settings(disableDocs)
   .settings(disablePublishing)
   .settings(
-    fork in Test := true,
+    fork in Test := false,
     javaOptions in Test += "-Xss1M",
     libraryDependencies ++= Seq(scalacheckDep),
     unmanagedSourceDirectories in Compile := Nil,
--- a/build.sbt
+++ b/build.sbt
@@ -581,6 +581,9 @@ lazy val scalacheck = project.in(file("test") / "scalacheck")
   .settings(
     fork in Test := false,
     javaOptions in Test += "-Xss1M",
+    testOptions += Tests.Cleanup { loader =>
+      ModuleUtilities.getObject("scala.TestCleanup", loader).asInstanceOf[Runnable].run()
+    },
     libraryDependencies ++= Seq(scalacheckDep),
     unmanagedSourceDirectories in Compile := Nil,
     unmanagedSourceDirectories in Test := List(baseDirectory.value)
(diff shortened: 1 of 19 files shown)


retronym:
In a separate [branch](https://github.com/retronym/scala/tree/revert/alloverriding), I've tried some alternative/complementary changes:
  - Use JUnit as the test framework, rather than Scalacheck, by using a ScalaCheck => JUnit bridge. This has a nice collateral benefit of allowing one to run the entire suite in the IDE with a reach test UI, rather than running then individually as simple `main` methods.
  - Port the Scaladoc tests to direct JUnit tests, as they don't use any data generators from Scalacheck, they just make plain old assertions.

I'll defer that work to a later date to restrict this PR to what's required to unblock 2.12.3

retronym:
Actually, I now remember another advantate of the JUnit bridging approach that we should consider here: the parallel collections tests create thousands of threads and never clean them up, because there is no support for such a cleanup step in Scalacheck. Removing forking means those threads will live on in the SBT JVM. I'll see if I can find a solution for that problem under Scalacheck.

lrytz (approved):

lrytz:
> Removing forking means those threads will live on in the SBT JVM

IMO that's not necessarily blocking for 2.12.3

retronym:
Added a commit to clean up the threads before returning to SBT.

lrytz (approved):
Looks good!

SethTisue:
special thanks to @janekdb for noticing the problem and to @xuwei-k for correctly guessing the cause

SethTisue:
```diff
- // this test _will_ fail again some time in the future.
```

LOL

lrytz:
Why is the second commit marked `[nomerge]`?

retronym:
Because the parallel collections moved to a module in 2.13

We should check whether any of the problems here also exist in that module.

lrytz:
Right thanks!

PR #6004, opened 2017-07-24
Don't assume final member class in another comp. unit will stay final

retronym:
The optimization in #5099 that avoided needless capture of the enclosing
class, and hence improved serializability of local classes and functions,
went too far. It also caused constructor calls of member classes to
use `null` rather than the actual outer reference as the `$outer`
constructor argument. The enclosed test case exhibits this problem
by witnessing an null `Outer.this`.

This commit limits the strategy to use `null`-s to the outer refererences
of anonymous and local classes, which rules out cross compilation unit
effects (in absence of `-opt`.)

Fixes scala/bug#10423

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Constructors.scala
+++ b/src/compiler/scala/tools/nsc/transform/Constructors.scala
@@ -764,7 +764,7 @@ abstract class Constructors extends Statics with Transform with TypingTransforme
           primaryConstrBody.expr)
       })
 
-      if (omittableAccessor.exists(_.isOuterField) && !constructorStats.exists(_.exists { case i: Ident if i.symbol.isOuterParam => true; case _ => false}))
+      if ((exitingPickler(clazz.isAnonymousClass) || clazz.originalOwner.isTerm) && omittableAccessor.exists(_.isOuterField) && !constructorStats.exists(_.exists { case i: Ident if i.symbol.isOuterParam => true; case _ => false}))
         primaryConstructor.symbol.updateAttachment(OuterArgCanBeElided)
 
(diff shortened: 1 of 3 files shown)


retronym:
@lrytz I think the correctness problem here is serious enough to warrant inclusion of the fix in 2.12.3

lrytz (approved):
I agree to merge this for 2.12.3.

src/compiler/scala/tools/nsc/transform/Constructors.scala
```diff
@@ -764,7 +764,7 @@ abstract class Constructors extends Statics with Transform with TypingTransforme
           primaryConstrBody.expr)
       })
 
-      if (omittableAccessor.exists(_.isOuterField) && !constructorStats.exists(_.exists { case i: Ident if i.symbol.isOuterParam => true; case _ => false}))
+      if ((exitingPickler(clazz.isAnonymousClass) || clazz.originalOwner.isTerm) && omittableAccessor.exists(_.isOuterField) && !constructorStats.exists(_.exists { case i: Ident if i.symbol.isOuterParam => true; case _ => false}))
```
adriaanm, line 767: @odersky notes in the meeting that you could also do it for private classes.

retronym: Agreed.

I initially tried to put the condition into `Delambdafy` such that it only applies the "pass null rather than outer" trick if the current owner of the call site of the constructor and the class being constructed share the same enclosing top level class.

Note that just sharing a compilation unit isn't enough as top level definitions in a file can be split into separate files.

PR #6007, opened 2017-07-25
SI-9745 typedFunction keeps x => f(x)

som-snytt:
To type `x => f(x)`, `typedFunction` uses the expected
type and `f` to type `x`. Previously, it discarded the
result of typechecking `f`, which was a problem in the
case of a `g(e)` where `e` is "erroneous" but rewriteable
to an expr that checks, e.g., `g(i += 1)`.

This commit keeps the rewritten `g(i = i + 1)` instead
of trying to recover the erroneous tree.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -2901,7 +2901,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
     /** Type check a function literal.
      *
-     * Based on the expected type pt, potentially synthesize an instance of
+     *  Based on the expected type pt, potentially synthesize an instance of
      *   - PartialFunction,
      *   - a type with a Single Abstract Method (under -Xexperimental for now).
      */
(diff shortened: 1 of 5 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2960,35 +2960,34 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+          if (paramsMissingType.isEmpty || pt == ErrorType) (fun, null)
+          else fun.body match {
             // we can compare arguments and parameters by name because there cannot be a binder between
             // the function's valdefs and the Apply's arguments
             case Apply(meth, args) if (vparams corresponds args) { case (p, Ident(name)) => p.name == name case _ => false } =>
               // We're looking for a method (as indicated by FUNmode in the silent typed below),
               // so let's make sure our expected type is a MethodType
               val methArgs = NoSymbol.newSyntheticValueParams(argpts map { case NoType => WildcardType case tp => tp })
-
-              val result = silent(_.typed(meth, mode.forFunMode, MethodType(methArgs, respt)))
```
adriaanm, line 2972: Let's explore keeping the `silent` and a variation of throwing in a `.duplicate`  -> `silent(_.typed(meth.duplicate, mode.forFunMode, MethodType(methArgs, respt)))`. That fixed the issue for me as well. I'm hesitant to drop the `silent`, but grant that duplication is costly.

som-snytt: IIRC there was a previous PR that duplicates? For some reason, it seemed simpler just to accept the unsilent judgment; but I don't spend all day with the code, and also I don't remember what I did last summer. Maybe this is an opportunity to try some jmh-driven development.

adriaanm:
I'm going to defer until M4, because I'm dubious on dropping `silent`, but my `.duplicate` suggestion may need more performance investigation than we can afford in time for M3.

som-snytt:
@adriaanm Sorry I haven't had a chance to get back to this. It doesn't take much to blow my lunchtime budget for FOSS. In 2018, though, I hope to get better at contributing review time, which is probably more useful than marginal repairs. (But please don't quote me; I'll probably delete this resolution next December.)

adriaanm:
I'm pushing this to the RC, but I'm open to a refined fix either before (M5) or after (2.13.x)

SethTisue:
needs rebase

adriaanm:
Adopting as #7316

som-snytt:
@adriaanm I have a few other jobs around the house I've been meaning to tackle, if you have time.

adriaanm:
What’s your travel cost policy?

PR #6008, opened 2017-07-26
Fix race condition in quasiquotes in runtime reflection

retronym:
Quasiquotes expand into calls to `SyntacticTuple`, etc, which call
`Symbol#exists` as part of a bounds check against the maximal arity of
`TupleN`, etc.

However, `exists` was not threadsafe. This patch adds a test that
demonstrates this race condition, and adds the neccessary override to
synchronize this method.

It also includes some followup improvements/simplifications to the
territory.

Fixes scala/scala-dev#409


lrytz:
```
prev = /home/jenkins/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.12.0.jar
curr = /home/jenkins/workspace/scala-2.12.x-validate-test@2/build/pack/lib/scala-reflect.jar
Found 1 binary incompatibilities (27 were filtered out)
 * abstract synthetic method
   scala$reflect$runtime$SynchronizedSymbols$SynchronizedSymbol$$super$exists()Boolean
   in interface scala.reflect.runtime.SynchronizedSymbols#SynchronizedSymbol is
   present only in current version
```

The `scala.reflect.runtime` package is internal, so we can just add the filter?

lrytz (approved):

retronym:
@lrytz Mima filters added.

lrytz (approved):

PR #6009, opened 2017-07-26
Fix title tag hover in scaladoc

Philippus:
as can be seen on f.e. this page:
http://www.scala-lang.org/files/archive/api/current/scala/beans/BeanInfoSkip.html
when hovering over the links on the right the title tags contain line breaks and spaces in unwanted places.
Additionally two link elements were spread out over multiple lines


Philippus:
@janekdb is this PR something for you to review?

src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
```diff
@@ -90,15 +90,13 @@ trait EntityPage extends HtmlPage {
                           mbr match {
                             case dtpl: DocTemplateEntity =>
                               dtpl.companion.fold(<span class="separator"></span>) { c: DocTemplateEntity =>
-                                <a class="object" href={relativeLinkTo(c)} title={c.comment.fold("")(com => Page.inlineToStr(com.short))}></a>
+                                <a class="object" href={relativeLinkTo(c)} title={c.comment.fold("")(com => Page.inlineToStrForTitleTag(com.short))}></a>
```
adriaanm: Could we lift `_.fold("")(com => Page.inlineToStrForTitleTag(com.short))` to a method? It's repeated at least three times.

Philippus: Yes, I was contemplating this as well. Will make it part of this PR.

adriaanm (approved):
Great, thanks!

PR #6010, opened 2017-07-27
Fixes over aggressive pruning of implicit scope.

### LINKED ISSUE #10081: StackOverflowError in typer with AnnotatedType
The following code crashes typer:

```scala
trait A[_]
trait B[X] extends A[B[X @unchecked]] { this.x }
```

A is not essential (any type constructor will do), neither is x (any name lookup in "this" suffices) nor unchecked (any annotation does the trick).

The problem appears to be in Implicits.companionImplicitMap (at least that's where it manifests itself), with an endless recursion between getClassParts and getParts, called with types that appear to be the same but not identical. In the reproduction above, the bts(i) as seen in getClassParts are always:

```scala
bts(1) = A[B[... @... @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked @unchecked]]
```

milessabin:
Fixes https://github.com/scala/bug/issues/4947, https://github.com/scala/bug/issues/5340 and https://github.com/scala/bug/issues/10425.

When the implicit scope is constructed an initial pruning phase attempts to discard implicit members which it is believed would later be rejected as ambiguous. Members of a companion are discarded if they can be reached via more than one prefix, on the assumption that being visible via two or more paths means that the multiple values must conflict with one another. However, this does not take into account the possibility that the implicit members might depend on the prefix in a way which renders them non-ambiguous. This scenario is illustrated in https://github.com/scala/bug/issues/4947 and https://github.com/scala/bug/issues/5340. This PR address that by only pruning implicits which are independent of the prefix.

Whilst this PR tries to stay as close as possible to the current behaviour, I believe that it would be desirable to drop the early pruning of even the prefix-independent implicits. If we do that then we get improved error messages reporting ambiguity rather the ambiguous members being silently ignored which leads to otherwise unexplained failures of implicit resolution. For instance in the case of `test/files/neg/t5340.scala` which currently reports,
```
t5340.scala:15: error: type mismatch;
 found   : Quux[MyApp.r.E,MyApp.s.E]
 required: Int
  (new Quux[r.E, s.E]): Int // fails due to pre-stripping implicits which
   ^
one error found
``` 
we would get the following report,
```
t5340.scala:15: error: type mismatch;
 found   : Quux[MyApp.r.E,MyApp.s.E]
 required: Int
Note that implicit conversions are not applicable because they are ambiguous:
 both method conv in object E of type [T, U](b: Quux[T,U])Int
 and method conv in object E of type [T, U](b: Quux[T,U])Int
 are possible conversion functions from Quux[MyApp.r.E,MyApp.s.E] to Int
  (new Quux[r.E, s.E]): Int // fails due to pre-stripping implicits which
   ^
one error found
```
Which is clearly more informative (even if also a little baffling). I think that ambiguity via multiple paths is sufficiently rare that the performance benefits of this early pruning are likely to minimal and not enough to outweigh improved error messages.

Because this change can result in a symbol _legitimately_ being revisited, via a different prefix, we need to change the way that the a pruned path is represented. Instead of using empty lists, as introduced in #5951 (which fixes https://github.com/scala/bug/issues/10081) we use explicit sentinels to mark that a symbol has been visited at a given prefix yielding no additional implicits. Testing this yielded https://github.com/scala/bug/issues/10425 which appears to be a regression relative to 2.12.x. That's also fixed in this PR.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -219,6 +219,13 @@ trait Implicits {
       tpeCache
     }
 
+    def dependsOnPrefix: Boolean = pre match {
+      case SingleType(pre0, _) => tpe.exists(_ =:= pre0)
+      case _ => false
+    }
+
+    def isSearchedPrefix: Boolean = name == null && sym == NoSymbol
+
     def isCyclicOrErroneous: Boolean =
       if(sym.hasFlag(LOCKED)) true
       else {
@@ -292,6 +299,8 @@ trait Implicits {
     override def equals(other: Any) = other match { case that: AnyRef => that eq this  case _ => false }
     override def hashCode = 1
   }
+
+  def SearchedPrefixImplicitInfo(pre: Type) = new ImplicitInfo(null, pre, NoSymbol)
 
   /** A constructor for types ?{ def/type name: tp }, used in infer view to member
    *  searches.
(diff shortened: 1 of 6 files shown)


src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -215,6 +215,13 @@ trait Implicits {
       tpeCache
     }
 
+    def dependsOnPrefix: Boolean = pre match {
+      case SingleType(pre0, _) => tpe.exists(_ =:= pre0)
+      case _ => false
+    }
```
retronym, line 225: We'd need to think carefully about the performance here. Usually we need to cache the results of this sort of question.

milessabin: This would be called on a very infrequent execution path so I don't think it would have a measurable performance impact ... would a benchmark run improve your confidence level here?

retronym:
LGTM. Could you update the commit comment with the (excellent!) PR message before we merge? (Here's a suitable commit: https://github.com/scala/scala/commit/b3d4d01e38b27ad676c9e7874446c7715e30ff47)

milessabin:
Done. Also rebased.

PR #6012, opened 2017-07-28
Backend Refactoring

lrytz:
Isolate backend components to run parallel from the compiler frontend

Refactor the compiler backend to isolate components that are planned to
run in parallel (local optimizations, classfile writing) from the
compiler frontend (i.e., no access to the Global instance).

The long inheritance stack (`BCodeIdiomatic` - `BCodeHelpers` - ... -
`BCodeSyncAndTry`) is now isolated and simply a component of the new
`CodeGen` class. `CodeGen` itself is one component of `GenBCode`, the
backend class.

`GenBCode` has some other components
  -` BTypes`, used in `CodeGen` and `PostProcessor`. Many of the backend
    components that used to be in `BTypes` (`inliner`, `inlinerHeuristics`,
    `localOpt`) now moved to the `PostProcessor`.
  - `PostProcessor` implements global optimization (inlining), local
    optimizations and classfile writing. It has itself quite a number
    of components (`inliner`, `callGraph`, `classfileWriter`, ...)
  - `PostProcessorFrontendAccess` is an interface of functionality
    required in the `PostProcessor` that is implemented by accessing
    the frontend (`Global`). It synchronizes on the single `frontendLock`
    object.

The `PostProcessor` doesn't have a `Global`. It can access certain operations that depend on the frontend through `PostProcessorFrontendAccess`, which synchronizes on the `frontendLock`. Also the types in `CoreBTypes` are computed lazily by looking at symbols, so access is synchronized on the same lock.

The `BTypes` component is passed into `CodeGen` and the `PostProcessor`. Note that `GodeGen` has a `global`, so it can (and does) access various parts of the `PostProcessor` through `global.genBCode.postProcessor`, for example it updates `genBCode.postProcessor.callGraph.inlineAnnotatedCallsites` during code gen.

Quite a few cleanups all around.


lrytz:
This is ready for review, I'm not planning to add more right now. There's more that could be done, but that doesn't interfere with @mkeskells's plans started in https://github.com/scala/scala/pull/5815.

Review by @retronym. Start looking at [`GenBCode`](https://github.com/lrytz/scala/blob/backendRefactor/src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala#L13) and follow components from there. The structure should be much more easy to understand now.

The `PostProcessor` doesn't have a global. It can access certain operations that depend on the frontend through `PostProcessorFrontendAccess`, which synchronizes on the `frontendLock`. Also the types in `CoreBTypes` are computed lazily by looking at symbols, so access is synchronized on the same lock.

The `BTypes` component is passed into `CodeGen` and the `PostProcessor`. Note that `GodeGen` has a `global`, so it can (and does) access various parts of the `PostProcessor` through `global.genBCode.postProcessor`, for example it updates `genBCode.postProcessor.callGraph.inlineAnnotatedCallsites` during code gen.

I started a benchmark for the last commit, should be coming in here: https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=6ac6da8b61%7C58cfe9e76c&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1500736130730&to=1502458163200

Some notes for the TODOs on making local optimizations / classfile writing work in parallel:
  - maybe use Java's concurrent data structures (instead of scala.collection.concurrent.TrieMap)
  - `ByteCodeRepository`: probably needs synchronization to make sure a class is not parsed twice
  - `BTypesFromClassfile.classBTypeFromParsedClassfile`/`classBTypeFromClassNode`: probalby need to make sure that the same CalssBType is not created twice concurrently

Data structures to make concurrent
  - `BackendUtils.maxLocalsMaxStackComputed`/`indyLambdaImplMethods`
  - `ByteCodeRepository.javaDefinedClasses`
  - `LocalOpt.unreachableCodeEliminated`
  - `CallGraph.inlineAnnotatedCallsites`/`noInlineAnnotatedCallsites`

lrytz:
Performance is flat, build is green :)

lrytz:
The jardiff is non-empty (https://gist.github.com/lrytz/2079b2f9f933da06a3a55beb48c7b5fa): classes with a ` $deserializeLambda$` method now get an InnerClass entry for `MethodHandles$Lookup`. This is an accidental fix, I'll add a test for it and will figure out what fixed it.

By the way, @mkeskells, if you don't want to wait until this PR is merged you can also start working on top of my branch (https://github.com/lrytz/scala/tree/backendRefactor).

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -28,17 +28,17 @@ import scala.util.control.{NoStackTrace, NonFatal}
...
-  import postProcessor.{bTypes, bTypesFromClassfile, callGraph, frontendAccess}
+  import postProcessor.{bTypes, bTypesFromClassfile, callGraph}
   import bTypes._
   import callGraph.ClosureInstantiation
   import coreBTypes._
   import frontendAccess.compilerSettings
 
   // unused objects created by these constructors are eliminated by pushPop
-  private[this] lazy val sideEffectFreeConstructors: LazyVar[Set[(String, String)]] = perRunLazy {
+  private[this] lazy val sideEffectFreeConstructors: LazyVar[Set[(String, String)]] = perRunLazy(this) {
```
retronym: Is it a good idea for this to be both a `lazy val` and a `LazyVar`? The implementation of `PerRunInit` is not currently threasafe (no sync around mutation/access of `inits`), so I guess the invariant is that we expect that `sideEffectFreeConstructors` will have been called prior to `initialize` and any multi-threading.

src/compiler/scala/tools/nsc/backend/jvm/PerRunInit.scala
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Utility for backend components that have state that needs to be re-initialized at every compiler
+ * run, for example state that depends on compiler settings of frontend types (Symbols, Types).
+ *
+ * The trait provides an `initialize` method that runs all initializers added through `perRunLazy`.
+ */
+trait PerRunInit {
+  private val inits = ListBuffer.empty[() => Unit]
+
+  def perRunInit(init: => Unit): Unit = inits += (() => init)
```
retronym, line 14: Is it an error to call this after `initialize` has been called? Can we add an assertion?

src/compiler/scala/tools/nsc/backend/jvm/PerRunInit.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  def initialize(): Unit = inits.foreach(_.apply())
```
retronym, line 16: Can we discard the closures in `init` after forcing them, perhaps to allow some references to be GC-ed?

retronym:
(More) data structures to make concurrent:
  - `PostProcessor.generatedClasses`
  - `originalClosureInit.inlinedClones += newClosureInit` (not sure if naturally protected or not by the structure of `closureInstantiations`
  - `InlineSuccess.downstreamLog` (again, might not need it if we can reason that that parallel threads naturally work on different instances)

src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala
```diff
@@ -3,470 +3,71 @@
...
+      val genStart = Statistics.startTimer(BackendStats.bcodeGenStat)
+      super.run() // invokes `apply` for each compilation unit
+      Statistics.stopTimer(BackendStats.bcodeGenStat, genStart)
 
-    case class Item3(arrivalPos: Int,
-                     mirror:     SubItem3,
-                     plain:      SubItem3,
-                     bean:       SubItem3,
-                     outFolder:  scala.tools.nsc.io.AbstractFile) {
+      postProcessor.postProcessAndSendToDisk()
```
retronym, line 45: The architecture of generating all the `ClassNodes` in memory is well suited to compiling with global optimization enabled, but for non-optimized compile runs imposes an unnecessary memory burden. In that mode, I'd like to see code gen and classfile writing happen within the loop over compilation units, and make sure that we release the `ClassNode` instances in between.

I've sketched out what I'd like in a commit in https://github.com/scala/scala/compare/2.12.x...retronym:review/6012?expand=1

retronym:
I'd prefer to have explicit types for the component wiring, as in https://github.com/scala/scala/compare/2.12.x...retronym:review/6012?expand=1

retronym:
Could you please update the PR description with the important parts of your comment above and the motivation for the change?

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
retronym: Additionally, the blend of locks based on lazy vals and `frontEndLock` needs to be considered carefully for potential deadlocks.

lrytz: There's probably too many implicit invariants, but here they are:
  - `initialize` is called in each compiler run at the beginning when the GenBCode phase starts running. There's no multithreading at this point.
  - I kept the values here as lazy vals so that only those actually accessed are ever added to the `PerRunInit.this.inits` buffer (and therefore re-initializes in the next run).
  - I *think* there can't be a deadlock here. Initializing a lazy val here will create the `LazyVar` and add it to the `this.inits` buffer, and this happens synchronized on `this`.
  - Completing a `LazyVar` synchronizes on the `frontendLock`.

src/compiler/scala/tools/nsc/backend/jvm/PerRunInit.scala
lrytz: My thinking was that it's OK to call this after `initialize`, like the lazy vals in `CoreBTypes`.

src/compiler/scala/tools/nsc/backend/jvm/PerRunInit.scala
lrytz: We need to keep them around to re-initialize the `LazyVar`s in the beginning of the next run.

lrytz:
> (More) data structures to make concurrent:
> 
> PostProcessor.generatedClasses

It's read-only after code gen, but maybe that's not safe for a ListBuffer? Probably better be safe.

> originalClosureInit.inlinedClones += newClosureInit (not sure if naturally protected or not by the structure of closureInstantiations
> InlineSuccess.downstreamLog (again, might not need it if we can reason that that parallel threads naturally work on different instances)

I think in a first iteration the inliner will only run single-threaded, so these two should be OK. Inlining in parallel requires synchronization.

src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala
lrytz: That's a good suggestion. We can also run local optimizations in this mode. Can we do that in a follow-up PR?

lrytz:
Will take in your explicit type annotations and update the PR descriptions, probably also improve the commit messages.

retronym:
> It's read-only after code gen, but maybe that's not safe for a ListBuffer? Probably better be safe.

Depending how we implement interspersed code gen and postprocessing we might need it. Probably best to encapsulate access to it rather than exposing the buffer and synchronize defensively.

src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala
retronym: Yep, so long as we don't forget before 2.12.4.

lrytz:
> classes with a $deserializeLambda$ method now get an InnerClass entry for MethodHandles$Lookup.

Added a test: https://github.com/scala/scala/pull/6012/commits/5f83d78b40bf86a975ad17a7957b895b14237544

In 2.12.3, we do it in the wrong order (https://github.com/scala/scala/blob/v2.12.3/src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala#L268-L271)

retronym (approved):

PR #6013, opened 2017-07-29
Add original tree attachment

jvican:
Adds an original tree attachment that allows external tools (compiler
plugins like scalameta & zinc) to keep track of the previous, unadapted
tree. The reason why this is required is explained in SD-340: https://github.com/scala/scala-dev/issues/340.

This change enables the incremental compiler to detect changes in `final
val`s: https://github.com/sbt/zinc/issues/227.

It also enables a fix for scala/bug#10426 by allowing the scaladoc
compiler to let the compiler adapt constants without losing the tree that
will be shown in the Scaladoc UI.

To maintainers: I was thinking of the best way to test this, but
couldn't come up with an elegant one. Do you suggest a way I could write
a test for this? Is there a precedent in testing information carried in
the trees?

I think @lrytz is the right person to review this, since he suggested
this fix in the first place.

Fixes scala/bug#7173

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Erasure.scala
+++ b/src/compiler/scala/tools/nsc/transform/Erasure.scala
@@ -1226,13 +1226,17 @@ abstract class Erasure extends InfoTransform
         case Match(selector, cases) =>
           Match(Typed(selector, TypeTree(selector.tpe)), cases)
 
-        case Literal(ct) if ct.tag == ClazzTag
-                         && ct.typeValue.typeSymbol != definitions.UnitClass =>
-          val erased = ct.typeValue.dealiasWiden match {
-            case tr @ TypeRef(_, clazz, _) if clazz.isDerivedValueClass => scalaErasure.eraseNormalClassRef(tr)
-            case tpe => specialScalaErasure(tpe)
-          }
-          treeCopy.Literal(tree, Constant(erased))
+        case Literal(ct) =>
+          // We remove the original tree attachments in pre-easure to free up memory
+          val cleanLiteral = tree.removeAttachment[OriginalTreeAttachment]
+
+          if (ct.tag == ClazzTag && ct.typeValue.typeSymbol != definitions.UnitClass) {
+            val erased = ct.typeValue.dealiasWiden match {
+              case tr @ TypeRef(_, clazz, _) if clazz.isDerivedValueClass => scalaErasure.eraseNormalClassRef(tr)
+              case tpe => specialScalaErasure(tpe)
+            }
+            treeCopy.Literal(cleanLiteral, Constant(erased))
+          } else cleanLiteral
 
         case ClassDef(_,_,_,_) =>
           debuglog("defs of " + tree.symbol + " = " + tree.symbol.info.decls)
--- a/src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
@@ -165,4 +165,21 @@ trait StdAttachments {
   def markDynamicRewrite(tree: Tree): Tree = tree.updateAttachment(DynamicRewriteAttachment)
   def unmarkDynamicRewrite(tree: Tree): Tree = tree.removeAttachment[DynamicRewriteAttachment.type]
   def isDynamicRewrite(tree: Tree): Boolean = tree.attachments.get[DynamicRewriteAttachment.type].isDefined
+
+  /**
+   * Marks a tree that has been adapted by typer and sets the original tree that was in place before.
+   * 
+   * Keeping track of the original trees were is an important feature for some compiler plugins (like
+   * Scalameta) and the incremental compiler (Zinc). In both cases, adapting trees loses information
+   * in some sense and do not allow external tools to capture some information stored in user-defined
+   * trees that are optimized away by early phases (mostly, typer).
+   * 
+   * See how the absence of this attachment blocks Zinc: https://github.com/sbt/zinc/issues/227.
+   * Related: https://github.com/scala/scala-dev/issues/340.
+   * 
+   * This attachment is, at the moment, only used to keep track of constant-folded constants. It
+   * has a generic wording in the hope that in the future can be reused in the same context to keep
+   * track of other adapted trees.
+   */
+  case class OriginalTreeAttachment(original: Tree)
 }
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -995,8 +995,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         val sym = tree.symbol
         if (sym != null && sym.isDeprecated)
           context.deprecationWarning(tree.pos, sym)
-
-        treeCopy.Literal(tree, value)
+        // Keep the original tree in an annotation to avoid losing tree information for plugins
+        treeCopy.Literal(tree, value).updateAttachment(OriginalTreeAttachment(original))
       }
 
       // Ignore type errors raised in later phases that are due to mismatching types with existential skolems


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -27,7 +27,7 @@ import scala.reflect.macros.whitebox
  *  @author  Martin Odersky
  *  @version 1.0
  */
-trait Typers extends Adaptations with Tags with TypersTracking with PatternTypers {
+trait Typers extends Adaptations with Tags with TypersTracking with PatternTypers with StdAttachments {
```
lrytz: I that necessary? The self type (`Analyzer`) includes `StdAttachments`.

src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
```diff
@@ -165,4 +165,24 @@ trait StdAttachments {
...
+   * Related: https://github.com/scala/scala-dev/issues/340.
+   * 
+   * This attachment is, at the moment, only used to keep track of constant-folded constants. It
+   * has a generic wording in the hope that in the future can be reused in the same context to keep
+   * track of other adapted trees.
+   */
+  case class OriginalAdaptedTreeAttachment(original: Tree)
+  def markOriginal(tree: Tree, original: Tree) = tree.updateAttachment(OriginalAdaptedTreeAttachment(original))
+  def unmarkOriginal(tree: Tree): Tree = tree.removeAttachment[OriginalAdaptedTreeAttachment]
+  def isAdaptedTree(tree: Tree): Boolean = tree.attachments.get[OriginalAdaptedTreeAttachment].isDefined
```
lrytz: I personally prefer not to have those methods, they just make you go through an indirection when reading the code.

lrytz (commented):
I'd be interested to see how many such attachments are created, say when compiling the std lib. Could you do the experiment?

Also, should we eliminate the attachment at some phase? Maybe in erasure?

lrytz:
Related to https://github.com/scala/scala/pull/4880, which WIPs the idea of keeping all trees (with `ConstantType`) until erasure.

jvican:
> Related to #4880, which WIPs the idea of keeping all trees (with ConstantType) until erasure.
  
I'm aware of this effort, hope it gets to the finish line at some point. Until that gets merged, I think this is the perfect solution for Zinc -- I would like to support incremental changes in constants for 2.12.4 codebases.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
jvican: Yes, it was necessary -- sbt was failing to compile otherwise.

src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
jvican: I've done them to follow the convention. Honestly, I like it. I was hoping that there was a place in the compiler where the utils and definitions of attachments in the compiler was defined.
  
If we remove the attachment after erasure, we'll use both `unmarkOriginal` and `isAdaptedTree`, right? In any case, I'm happy to remove them if you think they should go away.

jvican:
> I'd be interested to see how many such attachments are created, say when compiling the std lib. Could you do the experiment?
  
How should I do this? Can you point me to some docs or previous examples?
  
> Also, should we eliminate the attachment at some phase? Maybe in erasure?
  
Yes, this is indeed a good idea.

lrytz:
> How should I do this? Can you point me to some docs or previous examples?

For example just a `HashSet` in some top-level object, add to it in the constructor of `OriginalAdaptedTreeAttachment`, and print its content at the [end of the run](https://github.com/scala/scala/blob/v2.12.3/src/compiler/scala/tools/nsc/Global.scala#L1466). Then `build/quick/bin/scalac -d /tmp $(find src/library -name '*.scala')`

>>  Also, should we eliminate the attachment at some phase? Maybe in erasure?
> Yes, this is indeed a good idea.

I'm not sure, I was just asking the question.

jvican:
>  For example just a HashSet in some top-level object, add to it in the constructor of OriginalAdaptedTreeAttachment, and print its content at the end of the run. Then build/quick/bin/scalac -d /tmp $(find src/library -name '*.scala')
  
Thanks, will try that.

> I'm not sure, I was just asking the question.
  
I think it's a good idea, neither Zinc, scaladoc or Scalameta traverse trees after typer, so it's a good tradeoff.

src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
adriaanm: I also prefer the more direct style (i.e., drop the additional methods). The method names are all pretty ambiguous. Also, why `Original*Adapted*` -- `OriginalTreeAttachment` captures the same info (the presence of the attachment implies the tree is no longer the original).

adriaanm:

src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
jvican: I'll use the more direct style, rename methods/names and update this PR.

adriaanm: Ok, as a reminder:  please do not include the ticket in the commit title/body. It should be in the PR description. (Title should be short, referencing the issue in the commit message causes noise on every push.)

lrytz: Our [CONTRIBUTING.md](https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#bug-fix) says

> At the end of the commit message, include "Fixes scala/bug#NNNN"

Should that be changed? I think it's good to have the reference in the commit message.

adriaanm: I'm fine with keeping it, I'll get over my OCD that hates seeing the stale commit references pile up until the git GC kicks in at github :-)

jvican:
@lrytz Our little experiment gives us the following data about adapted trees:
  
| Phases | Number |
| --------- | ---------- |
| typer | 12718 |
| patmat | 1300 |
| fields | 479 |
| specialize | 469 |
| total | 15063 |
  
The total includes a few adapted trees that were done by other phases but were not important enough to add to the table.

jvican:
I've added a second commit that cleans the original tree attachments in pre erasure so that we free up pointers to stale trees sooner in the pipeline. I think this hits the right trade-off, so thanks for the suggestion Lukas.
  
The PR is now updated and ready for another review.

adriaanm:
Seems like a real (though strange) error.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1226,13 +1226,19 @@ abstract class Erasure extends InfoTransform
...
           Match(Typed(selector, TypeTree(selector.tpe)), cases)
 
-        case Literal(ct) if ct.tag == ClazzTag
-                         && ct.typeValue.typeSymbol != definitions.UnitClass =>
-          val erased = ct.typeValue.dealiasWiden match {
-            case tr @ TypeRef(_, clazz, _) if clazz.isDerivedValueClass => scalaErasure.eraseNormalClassRef(tr)
-            case tpe => specialScalaErasure(tpe)
+        case Literal(ct) =>
+          val cleanLiteral = {
+            if (tree.attachments.get[OriginalTreeAttachment].isEmpty) tree
```
adriaanm: Use `contains` instead?

jvican: Done.

adriaanm: Or better yet, just use `remove` directly. If this pattern of checking whether the attachment is there before removing is faster, we should implement it once in `remove` (one method call that requires a class tag is also better than two)

jvican: Done in last commit!

lrytz (approved):

PR #6017, opened 2017-07-30
Fix description of the `next` method

Philippus:
closes https://github.com/scala/bug/issues/9423
the `next` method does not discard the element from the iterator


SethTisue:
maybe "and advances the iterator"? maybe "discard" isn't the right word, but I'm not sure it's improvement to say nothing about the effect of `next()` on the iterator's state

Philippus:
I've added that.

SethTisue (approved):

PR #6019, opened 2017-07-31
Fix Queue reversed append

edmundnoble:
https://github.com/scala/scala/pull/5891 introduced yet another bug which is that collections appended to `Queue`s are appended reversed. This is the fix; thanks very much @mingthemerciless for pointing it out.

### DIFF
--- a/src/library/scala/collection/immutable/Queue.scala
+++ b/src/library/scala/collection/immutable/Queue.scala
@@ -115,7 +115,9 @@ sealed class Queue[+A] protected(protected val in: List[A], protected val out: L
           val thatQueue: Queue[B] = that.asInstanceOf[Queue[B]]
           thatQueue.in ++ (thatQueue.out reverse_::: this.in)
         } else {
-          (new ListBuffer[B] ++= that.seq).prependToList(this.in)
+          val lb = new ListBuffer[B]
+          that.seq.foreach(_ +=: lb)
+          lb.prependToList(this.in)
         }
       new Queue[B](newIn, this.out).asInstanceOf[That]
     } else {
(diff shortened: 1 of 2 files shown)


Ichoran:
Can you please add tests that construct queues that have

  1. All elements in `in`
  2. All elements in `out`
  3. A mix of both

and make sure that the elements appear in the correct order when doing concatenating?

SethTisue:
perhaps this is a job for ScalaCheck?

Ichoran:
Writing the generator that reflects interestingly different cases in the internal structure is the hard part.  Queue is a bit weird that way.

edmundnoble:
Added the tests.

Ichoran:
@edmundnoble - It is hard for me to tell that you've covered all the cases.  Could you have a generator for queues of length 0; length 1 with the element either pre or post; and length 2 with all possibilities (both pre, one each, both post)?  Then all of these should be on both the LHS and RHS.  Plus you should have vectors of length relevant lengths on the RHS of every type of queue (all six).  Then you can run through the whole table, making sure (because the generators are parameterized) that they always generate (0, 1, 2, ...) so it's easy to check that they're correct.  It seems to me that there should be about 6*9 or thereabouts tests in total.

adriaanm:

edmundnoble:
@Ichoran The length of the vectors shouldn't matter, from what I can tell. I have 7 tests now, which fail on the previous iteration of queue concat and succeed now. EDIT: I did the "one each" case twice, once adding the first element first and once adding the last element first. I wasn't really clear what you meant by a generator, nor how it would be parameterized, but I've separated the assertions out from the data to make it clearer.

Ichoran (approved):
LGTM.  Simple fix, and the tests seem complete.

PR #6021, opened 2017-08-03
Decouple warning unused implicit parameters

som-snytt:
Take separate flags for warning about implicit
and explicit parameters. `-Ywarn-unused:params`
is an expanding option that enables both. That
allows `-Ywarn-unused:params,-implicits` to turn
off warning for implicits, in lieu of the
equivalent `-Ywarn-unused:explicits`, which is
a ridiculous name for the flag.

Fixes https://github.com/scala/bug/issues/10447


adriaanm (approved):
Nice to make this more explicit. I was reminded of a tangential discussion about eta-expanding by default, even when the expected type does not demand it. When refactoring, the missing params lint could be used to detect this. Also, could we say that params are not missing when eta-expansion is requested (explicitly) by the expected type of the expression (even if it's implicit in the code).

PR #6022, opened 2017-08-03
Make SeqLike.distinct access HashSet once per entry

jxtps:
The old distinct first does a contains (`!seen(x)`), then does the insert, with each of these operations doing the full elemToEntry, index, and search dance. Since HashSet.add returns a true if the element was added (false otherwise) we can get away with doing that dance only once.

### DIFF
--- a/src/library/scala/collection/SeqLike.scala
+++ b/src/library/scala/collection/SeqLike.scala
@@ -507,9 +507,8 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
     val b = newBuilder
     val seen = mutable.HashSet[A]()
     for (x <- this) {
-      if (!seen(x)) {
+      if (seen.add(x)) {
         b += x
-        seen += x
       }
     }
     b.result()


src/library/scala/collection/SeqLike.scala
```diff
@@ -507,9 +507,8 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
     val b = newBuilder
     val seen = mutable.HashSet[A]()
     for (x <- this) {
-      if (!seen(x)) {
+      if (seen.add(x)) {
         b += x
-        seen += x
```
retronym, line 512: In the case that `Set#apply` would have returned `true`, the new formulation actually does a bit more work by updating the set with the new element. So I'd expect the new version to be slightly slower when there are many copies of given elements.

It does appear to do less work in the case that `xs.distinct.size / xs.size` is closer to 1.

Would you care to test out these variations in a benchmark? See #5537 for a good example of a benchmark-driven performance patch.

retronym (changes requested):
Benchmarks needed to evaluate tradeoff

adriaanm:

retronym:
I'm going to close this due to inactivity. If you have the time/inclination to followup with the benchmarks, please @-mention me and I can reopen this PR.

PR #6024, opened 2017-08-04
Propagate the DEFAULTINIT flag to the underlying field

lrytz:
Fields initialized with `_` are excluded from initialization
checks under `-Xcheckinit`.

Since 5f86b1d94d, the compiler checks that flag on the field symbol
instead of the getter. Unfortunately the flag was not actually copied
to the field symbol, causing the init check to be added.

Fixes scala/bug#10439 and it also fixes scala/bug#10437.


dwijnand:
(You need to say "Fixes" multiple times: https://help.github.com/articles/closing-issues-using-keywords/#closing-multiple-issues)

retronym (approved):

PR #6025, opened 2017-08-04
[REVERTED] Type Alternative patterns as the lub of their branches.

hrhino:
This [came up on gitter](https://gitter.im/scala/scala?at=5970ec67c101bc4e3aa4d0e3) a few weeks back, and the spec is rather ambiguous on whether this happens or not, tersely claiming that "All alternative patterns are type checked with the expected type of the pattern". As such, this is a bit of a spaghetti PR (in the "throw it at the wall to see if it sticks" sense) motivated by checking Dotty and seeing that The Future Of Scala™ duly types alternatives in this way.

Something tells me you're going to ask for a `-Xsource:2.13` check on this one, but I'm being optimistic. I'll throw up a spec update if you think I should. 

Review by @adriaanm.


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -1174,7 +1174,14 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       } else tree.tpe match {
         case atp @ AnnotatedType(_, _) if canAdaptAnnotations(tree, this, mode, pt) => // (-1)
           adaptAnnotations(tree, this, mode, pt)
-        case ct @ ConstantType(value) if mode.inNone(TYPEmode | FUNmode) && (ct <:< pt) && canAdaptConstantTypeToLiteral => // (0)
+        case ct @ ConstantType(value) if mode.inNone(TYPEmode | FUNmode | PATTERNmode) && (ct <:< pt) && canAdaptConstantTypeToLiteral => // (0)
+          /* don't adapt in PATTERNmode only because of the duplicated alternatives
```
jvican: Message is a little bit confusing here since we don't adapt in three modes, not only `PATTERNmode`. Good work, btw!

hrhino: Thanks! I'll find out why we don't adapt in `TYPEmode` and `FUNmode` and expand the comment, if you'd like. The `PATTERNmode` case seemed like it would be unintuitive without more context, so that's why I put it in.

jvican: You can leave the other modes uncommented :smile:. Just make it clear that the comment only applies to pattern mode!

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -5214,9 +5222,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       }
 
       def typedAlternative(alt: Alternative) = {
-        context withinPatAlternative (
-          treeCopy.Alternative(tree, alt.trees mapConserve (alt => typed(alt, mode, pt))) setType pt
-        )
+        context withinPatAlternative {
+          val trees1 = alt.trees mapConserve (alt => typed(alt, mode, pt))
+          treeCopy.Alternative(tree, trees1) setType lub(trees1 map (_.tpe))
+        }
```
retronym: Corresponding code in dotty:

https://github.com/lampepfl/dotty/blob/972f6d5eff30539a5a8e4016ab2df3164adc76c3/compiler/src/dotty/tools/dotc/typer/Typer.scala#L1229-L1232

https://github.com/lampepfl/dotty/blob/972f6d5eff30539a5a8e4016ab2df3164adc76c3/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala#L501-L502

```scala
  def typedAlternative(tree: untpd.Alternative, pt: Type)(implicit ctx: Context): Alternative = track("typedAlternative") {
    val trees1 = tree.trees mapconserve (typed(_, pt))
    assignType(cpy.Alternative(tree)(trees1), trees1)
  }

  def assignType(tree: untpd.Alternative, trees: List[Tree])(implicit ctx: Context) =
    tree.withType(ctx.typeComparer.lub(trees.tpes))
```

retronym:
Here's another example of the sharper type:

```
⚡ scala
Welcome to Scala 2.12.2 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> def foo(a: Some[Any]) = (a : @unchecked) match { case Some(x @ ("a" | "b")) => x }
foo: (a: Some[Any])Any

⚡ scala-pr 6025
Welcome to Scala 2.12.4-20170804-175643-4e2502b (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> def foo(a: Some[Any]) = (a : @unchecked) match { case Some(x @ ("a" | "b")) => x }
foo: (a: Some[Any])String
```

This sort of change definitely needs to be predicated on `-Xsource:2.13`, as it can flow out into inferred types of user APIs etc.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -1174,7 +1174,15 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       } else tree.tpe match {
         case atp @ AnnotatedType(_, _) if canAdaptAnnotations(tree, this, mode, pt) => // (-1)
           adaptAnnotations(tree, this, mode, pt)
-        case ct @ ConstantType(value) if mode.inNone(TYPEmode | FUNmode) && (ct <:< pt) && canAdaptConstantTypeToLiteral => // (0)
+        case ct @ ConstantType(value) if mode.inNone(TYPEmode | FUNmode | PATTERNmode) && (ct <:< pt) && canAdaptConstantTypeToLiteral => // (0)
+          /* We don't adapt in PATTERNmode because patmat checks for duplicated
```
retronym, line 1178: Is this part of the change covered by the test? I could only envisage it mattering in code like `case Foo(1 | 1)`, which we don't want to constant fold based on the pattern's sharper type `lub(1.type, 1.type)` down to `case Foo(1)` before patmat has issued the warning.

hrhino: So, that's not covered by the test in this PR, but without it `run/t7290` fails.

The actual test is 
```scala
  case 0 | 0 => 0
```
which without this `inNone` fails to warn.

I'm trying to think whether this needs a `-Xsource:2.13` flag; as far as I can tell it shouldn't change anything visible to the user without the other change in this PR. Can you think of anything?

odersky:
We are still in dire need of a full spec for pattern matching, so I am afraid this is something of a grey area. For the moment I think the best we can do is follow the code of the dotty compiler as a guideline, because it's quite a bit clearer than scalac in this area. So 👍  from me.

retronym:
I believe we should use a weak LUB rather than a normal LUB, reasoning by analogy with typechecking of `if` and `match` expressions.

```
⚡ scala-pr 6025 -Xsource:2.13
Welcome to Scala 2.12.4-20170807-130532-b6af7e3 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

scala> def foo(a: Some[Any]) = (a : @unchecked) match { case Some(x @ (1 | 1)) => x }
foo: (a: Some[Any])Int

scala> def foo(a: Some[Any]) = (a : @unchecked) match { case Some(x @ (1 | 1L)) => x }
foo: (a: Some[Any])AnyVal

scala> List(1, 1L)
res1: List[Long] = List(1, 1)

scala> if (true) 1 else 1L
res0: Long = 1
```

retronym:
This subtlety suggests we should also boy-scout this area of the spec.

```
The type of the _pattern alternative_ is the [weak least upper bound](03-types.html#weak-conformance)of the types of `$p_1$ | $\ldots$ | $p_n$`
```

This spec addition should be submitted directly to the 2.13.x branch in a separate PR.

retronym:
Dotty gets the type right. I recall it treats weak subtyping differently from Scala, so I guess its call to `lub` automatically handles the quirks of numeric subtyping.

```
⚡ dotr-launch 0.2.0-RC1
Welcome to Scala.next (pre-alpha, git-hash: 54d7089)  (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).
scala> List(1, 1L)
val res0: List[Long] = List(1, 1)
scala> def foo(a: Some[Any]) = (a : @unchecked) match { case Some(x @ (1 | 1L)) => x }
def foo(a: Some[Any]): Long
```

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -230,14 +230,17 @@ private[internal] trait GlbLubs {
...
+    else if (shouldUseNumericLub(tps))
       numericLub(tps)
     else if (tps exists typeHasAnnotations)
       annotationsLub(lub(tps map (_.withoutAnnotations)), tps)
     else
       lub(tps)
   )
 
+  private def shouldUseNumericLub(tps: List[Type]) =
+    tps forall (tp => isNumericValueType(tp.widen))
```
hrhino, line 242: I quite nearly just put the `.widen` in `isNumericValueType`, which doesn't appear to break anything. It's only used (in the compiler) here and [in `TypeConstraints`](https://github.com/scala/scala/blob/f6e46eb2b2c84d2fc8ac26f8d400004e99aa43be/src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala#L136), and since I haven't yet found a way to get there with a numeric `ConstantType` I can't convince myself it doesn't cause breakage. Also then I'd arguably have to rename the method and I'm running out of my names budget for the week.

retronym: Would `deconst` be sufficient? For constant types, `widen` does the same thing. Not sure of any case where where `isNumericValueType(tp.widen)` and `isNumericValueType(tp.deconst)` would differ, but the latter is both more intention revealing and cautious.

Both questions (widen vs deconst, internally or externally) are worth thinking about this in the context of the [proposed support](https://github.com/scala/scala/pull/5310) for allowing constant types to be directly expressed in the language and for them to survive through type inference where appropriate. In particular, that PR changes `isNumericValueType` to [internally widen](https://github.com/scala/scala/pull/5310/files#diff-2b1bf15b53df94495f7155a69f822adeL1417).

/cc @milessabin

hrhino: `.deconst` is perfectly decent here, I think; I tried pretty hard to find an observable difference, but since [single types are evidently expected to conform to `AnyRef`](https://www.scala-lang.org/files/archive/spec/2.12/03-types.html#singleton-types) I can't trip up the lubber with something like 

```scala
  val i1 = 1
  val l1 = 1L

  type I1 = i1.type
  type L1 = l1.type

  def foo3(a: Option[Any]) = (a : @unchecked) match { case Some(x @ (I1 | L1)) => x }
  foo3(None) : Long
```

... which is probably for the best.

As I mentioned, I mostly but still don't quite understand what the other use of `weakLub` (being `isWeakSubType` in `TypeComparers`) uses it for, but its arguments [are already `dealias`ed](https://github.com/scala/scala/blob/1b7d57f2f90b2c851a5c60250c077721d01c48f8/src/reflect/scala/reflect/internal/tpe/TypeComparers.scala#L571) by the time they get there, so that shouldn't be a worry.

The fact that all of the (par)tests pass with `isNumericValueType` doing the `widen`ing, even in the singleton literals PR, gives me great hope that it's not the wrong thing to do. I'm more than happy to move it there to give 2.12.4 a sneak peek of the greatness which that PR will bring.

Does that sound reasonable to you?

hrhino: So I went and rebased this on top of the `-Yliteral-types` PR...

```
> scala -Xexperimental
[info] Running scala.tools.nsc.MainGenericRunner -usejavacp -Xexperimental
Welcome to Scala 2.12.4-20170731-141836-2899d86 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> type I1 = 1
defined type alias I1

scala> type L1 = 1L
defined type alias L1

scala> def foo1(a: Option[Any]) = (a : @unchecked) match { case Some(x @ (_ : I1 | _ : L1)) => x }
foo1: (a: Option[Any])AnyVal

scala>
```

which is of course remedied by a well placed `.dealiasWiden` in `isNumericValueType`, but...

Neither `.widen` nor `.deconst` will make that work, and I suppose it's a question whether we expect it to at all. `.dealiasWiden` works, and doesn't seem to blow anything up, but it seems slightly dubious and probably approaching reason to rename `isNumericValueType`.

retronym: I've noted the problem with `isNumericSubtype` over in the [other PR](https://github.com/scala/scala/pull/5310#issuecomment-322352052), your test case illuminates it really well.

I would expect the weak LUB of those aliases to be the same as that of the RHS's.

Given `type A1 = T1; type A2 = T2`, `weakLub(T1, T2) =:= weakLub(A1, A2)` should hold.

Setting literal constant types aside for now, I agree that we should put the `deconst` in `isNumericValueType`.

I'm on the verge of suggesting that we retarget all of this PR to 2.13.x, though. Is the bug fix important enough to warrant the risks on the stable branch?

hrhino: A wise man once told me "there ain't no such thing as a small change". I'd like to see how things play out over in the singleton type literals PR.

re: rebasing and retargeting, I'm not depending on this for anything other than excising a single, safe `.asInstanceOf` call. Only say the word and it shall be done.

retronym: > Only say the word and it shall be done.

Consider it uttered on grounds of the non-short tail of interactions with weak subtyping and the expanded horizons that `42.type` promises.

You can keep the `if (isScala213)` guard as it currently stands as `scalac213 -Xsource:2.12` is a supported thing, insofar as we remember and are able to maintain the historical behaviour.

milessabin: I'm happy to make the literal types PR consistent with this, but I'm unclear what people expect the semantics to be.

hrhino: I, personally, would expect, in pseudo-scala:

* `weakLub(1.type, 1.type) === 1.type`
* `weakLub(1.type, 2.type) === Int`
* `weakLub(1.type, Int) === Int`
* `weakLub(1.type, 1L.type) === Long`
* `weakLub(1.type, Long) === Long`
* `weakLub(1.type, ().type) === AnyVal`
* `weakLub(1.type, "".type) === Any`
* `weakLub(1.type, String) === Any`.

and various similar results. Moreover, I'd expect those to be consistent over type-level eta-expansion; given `type I1 = 1`, `weakLub(I1, Long) === Long` &c.

It seems to me that `.dealias.deconst` in `isNumericValueType` would produce that behaviour. I don't see anything in [the updated section of the spec](https://github.com/scala/scala/pull/5310/files#diff-c067e953d98d0b8fb993aeae8bdb0bd4) which would indicate otherwise.

hrhino:
Alright, rebased onto 2.13.x head, and a spec update slipped in there somehow. Let me know if the wording is alright.

Also, some kind fellow with more and greater powers than I should probably update the milestone. If I can do it, I can't figure out how.

retronym (approved):
Thanks for the spec update. This is ready to go IMO.

SethTisue:
fails bootstrap, see scala/scala-dev#455

SethTisue:
reverted by #6216

hrhino:
@SethTisue/@adriaanm since the release is going out and this isn't technically in it, should the release-notes label get taken off?

retronym:
Yep, done.

adriaanm:
The label can stay as long as the milestone is accurate. We filter on both
in the notes

Blaisorblade:
@hrhino Did you ever retry this, since it was only reverted due to a separate bug?

hrhino:
@Blaisorblade no, and thanks for the ping. I've been really super busy the last few weeks, and my OSS time has really gone out the window. I'll bookmark this and pick it back up.

hrhino:
@Blaisorblade I'm back to looking at this, and unsettlingly the exact same diff (modulo minor conflicts) passes tests and bootstraps just fine. I'll re-make this PR once I can explain what fixed the problem in scala/scala-dev#455.

PR #6027, opened 2017-08-07
look at all parameter lists for error

martijnhoekstra:
Fixes scala/bug#9138

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -674,11 +674,14 @@ abstract class RefChecks extends Transform {
               // If there is a concrete method whose name matches the unimplemented
               // abstract method, and a cursory examination of the difference reveals
               // something obvious to us, let's make it more obvious to them.
-              val abstractParams   = underlying.tpe.paramTypes
-              val matchingName     = clazz.tpe.nonPrivateMembersAdmitting(VBRIDGE)
-              val matchingArity    = matchingName filter { m =>
+              val abstractParams     = underlying.tpe.paramTypes
+              val abstractParamLists = underlying.paramLists
+              val matchingName       = clazz.tpe.nonPrivateMembersAdmitting(VBRIDGE)
+              val matchingArity      = matchingName filter { m =>
                 !m.isDeferred &&
                 (m.name == underlying.name) &&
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -674,19 +674,22 @@ abstract class RefChecks extends Transform {
               // If there is a concrete method whose name matches the unimplemented
               // abstract method, and a cursory examination of the difference reveals
               // something obvious to us, let's make it more obvious to them.
-              val abstractParams   = underlying.tpe.paramTypes
-              val matchingName     = clazz.tpe.nonPrivateMembersAdmitting(VBRIDGE)
-              val matchingArity    = matchingName filter { m =>
+              val abstractParams     = underlying.tpe.paramTypes
```
lrytz: `abstractParams` is now unused, can be removed

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -674,19 +674,22 @@ abstract class RefChecks extends Transform {
...
+              val abstractParamLists = underlying.paramLists
+              val matchingName       = clazz.tpe.nonPrivateMembersAdmitting(VBRIDGE)
+              val matchingArity      = matchingName filter { m =>
                 !m.isDeferred &&
                 (m.name == underlying.name) &&
+                (m.paramLists.length == abstractParamLists.length) &&
+                (m.paramLists.map(_.length).sum == abstractParamLists.map(_.length).sum) &&
                 (m.tpe.paramTypes.size == underlying.tpe.paramTypes.size) &&
```
lrytz: this check is redundant now, no?

lrytz (commented):
looks good otherwise!

martijnhoekstra:
@lrytz thank you for the review - at the moment I'm completely swamped with work, and despite the change being rather trivial, I will most likely not have time to address it until this weekend at the earliest.

adriaanm:
No problem, I pushed a quick commit to your branch.

som-snytt:
It turns out that the fix doesn't offer any advice on the original case.

The "smaller" test case actually tests something different. I wonder if there is a way to express that two tests test the same conditions. I guess it is an area of research. Does EPFL have any research positions open?

Nuance at https://github.com/scala/scala/pull/7506

PR #6030, opened 2017-08-08
Position negative numeric literals starting at the minus sign.

hrhino:
Previously the start position was the first digit; this caused scaladoc to not pick up on the minus sign.

Fixes scala/bug#10391; review by @VladUreche / @janekdb


som-snytt:
That's a good one. It reminds me of [this puzzler](http://scalapuzzlers.com/#pzzlr-041).

retronym (approved):

PR #6032, opened 2017-08-10
Set encoding to UTF-8 for partest to avoid failures on windows

retronym:
Partest uses the default encoding to load checkfiles, etc
[in](https://github.com/scala/scala-partest/blob/bd2d2c6c/src/main/scala/scala/tools/partest/package.scala#L83):

```
def fileContents: String    = try sf.slurp() catch { case _: java.io.FileNotFoundException => "" }
```

When it forks processes to run the compiled `run` tests, it explicitly
[sets](https://github.com/scala/scala-partest/blob/1c263205/src/main/scala/scala/tools/partest/nest/Runner.scala#L174)
`-Dfile.encoding=UTF-8`.

The old `./test/partest` script (removed in 0bf343e0) used to also set `-Dfile.encoding=UTF-8` to the top-level
partest process. We forgot to port this into the part of our SBT build that forks partest.

Fixes scala/scala-dev#412

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -677,7 +677,8 @@ lazy val test = project
     // test sources are compiled in partest run, not here
     sources in IntegrationTest := Seq.empty,
     fork in IntegrationTest := true,
-    javaOptions in IntegrationTest += "-Xmx2G",
+    javaOptions in IntegrationTest ++= "-Xmx2G" :: "-Dfile.encoding=UTF-8" :: Nil,
+    testOptions in IntegrationTest += Tests.Argument("-Dfile.encoding=UTF-8"),
     testFrameworks += new TestFramework("scala.tools.partest.sbt.Framework"),
     testOptions in IntegrationTest += Tests.Argument("-Dpartest.java_opts=-Xmx1024M -Xms64M"),
     testOptions in IntegrationTest += Tests.Argument("-Dpartest.scalac_opts=" + (scalacOptions in Compile).value.mkString(" ")),


retronym:
Testing on windows in https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-windows/702 (link 404 until Jenkins summons a windows box from the netherworld under the light of the next new moon)

SethTisue:
attn @hrhino

hrhino:
Dang, faster than me! I had thought about this, but it appeared to me that [we explicitly ask for UTF8](https://github.com/scala/scala-partest/blob/bd2d2c6c/src/main/scala/scala/tools/partest/package.scala#L33) when doing that.

Testing on a spare windows 10 box seemed to indicate that partest was reading the checkfile just fine; it appeared that the `...-run.log` file was written out non-UTF8ly. I don't grok the sbt build so fully as all that, though, so if this works, then awesome!

Sorry for the trouble :/

retronym:
Force pushed the patch I'd thought I'd written originally, trying again in https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-windows/704

dwijnand:
If partest is already opinionated that utf-8 is the encoding to use when forking, why not make it opinionated to use utf-8 when slurping? I.e fix partest instead of changing the build.

retronym:
I  haven't conclusively found where the error is, but I agree with your suggestion to fix it in partest, too. But given that we used to set it in the `partest` script, I think its reasonable to change our fork options in the brave new SBT world.

dwijnand:
I wonder if [scala.reflect.io.File.apply](https://github.com/scala/scala-partest/blob/v1.1.1/src/main/scala/scala/tools/partest/package.scala#L55) is falling back [system codec](https://github.com/scala/scala/blob/v2.10.6/src/library/scala/io/Codec.scala#L81) because the implicit Codec Harrison linked to [lacks its type annotation](https://github.com/scala/scala-partest/blob/v1.1.1/src/main/scala/scala/tools/partest/package.scala#L33).

hrhino:
It's not the implicit there that's the problem. All over everywhere in `scala.sys.process`, we call `new InputStreamReader`, which picks the default encoding. 

The forked java process is invoked with [the right encoding](https://github.com/scala/scala-partest/blob/1.1.x/src/main/scala/scala/tools/partest/nest/Runner.scala#L171), but [when we connect to it](https://github.com/scala/scala/blob/2.12.x/src/library/scala/sys/process/BasicIO.scala#L164) we read from the pipe with the default encoding, which winds up being that `Cp1252` mess, and we make garbage out of the forked process's stdout.

I'd argue that it wouldn't be ridiculous to pass around an `implicit codec: Codec` all throughout `scala.sys.process`, but that can't happen until 2.13.x.

retronym:
I've created https://github.com/scala/bug/issues/10461 to track the API problem in `sys.process`.

PR #6034, opened 2017-08-10
Collect all statistics and optimize checks

jvican:
References: scala/scala-dev#149.

This PR does the two following things, all in the domain of statistics:

1. It allows the collection of all the statistics (before, we were only
   reporting on a very narrow set of the available statistics). This is
   achieved by turning `canEnable` into a `def` instead of `val`. Related
   history: ebb719f#diff-438d9e5ac0dd1965aabd4776e1ee70eaR246/
2. It adds a mechanism to remove the overhead of checking for statistics
   in the common case (where statistics are disabled). It does so by
   reusing a mechanism proposed by Rémi Forax and brought to my
   attention by @retronym (more here: scala/scala-dev#149).

There is one pending issues with this PR, which hasn't been solved yet:

* How do we keep this implementation runtime-agnostic, allowing
  compilers like Scalajs and Scala.native to default to a correct, inefficient
  implementation?

I don't know how to do this (or how these compilers are replacing JVM-specific
implementations), so I would appreciate insights from @sjrd and @densh.

In the future, I could look into ways of expanding this implementation to all
our flag checking infrastructure, which is what the referenced ticket on the
top is about. However, I'd prefer to do this in a separate PR if possible. For
now, I wanted to create a very small change to see the prior results.

This PR is work-in-progress and I've just opened it to check the
performance penalty of my changes. Afterwards, I would like to detect
what's the performance penalty of having statistics enabled by default,
I'm sure we can draw interesting results from there.

### DIFF
--- a/src/reflect/scala/reflect/internal/util/BooleanContainer.java
+++ b/src/reflect/scala/reflect/internal/util/BooleanContainer.java
@@ -1,0 +1,30 @@
+package scala.reflect.internal.util;
+
+/**
+ * Represents a container with a boolean value that tells the compiler whether
+ * an option is enabled or not. This class is used for configuration purposes
+ * (see scala.reflect.internal.util.Statistics).
+ */
+class BooleanContainer {
+  private final boolean value;
+
+  public BooleanContainer(boolean value) {
+    this.value = value;
+  }
+  
+  public boolean isEnabledNow() {
+    return value;
+  }
+
+  protected final static class TrueContainer extends BooleanContainer {
+    TrueContainer() {
+        super(true);
+    }
+  }
+
+  protected final static class FalseContainer extends BooleanContainer {
+    FalseContainer() {
+        super(false);
+    }
+  }
+}
--- a/src/reflect/scala/reflect/internal/util/Statistics.scala
+++ b/src/reflect/scala/reflect/internal/util/Statistics.scala
@@ -256,7 +256,7 @@ quant)
    *
    *  to remove all Statistics code from build
    */
-  final val canEnable = _enabled
+  final def canEnable = _enabled
 
   /** replace with
    *
--- a/src/reflect/scala/reflect/internal/util/Statistics.scala
+++ b/src/reflect/scala/reflect/internal/util/Statistics.scala
@@ -3,51 +3,53 @@ package reflect.internal.util
 
 import scala.collection.mutable
 
+import java.lang.invoke.{SwitchPoint, MethodHandle, MethodHandles, MethodType}
+
 object Statistics {
 
   type TimerSnapshot = (Long, Long)
 
   /** If enabled, increment counter by one */
   @inline final def incCounter(c: Counter) {
-    if (_enabled && c != null) c.value += 1
+    if (canEnable && c != null) c.value += 1
   }
 
   /** If enabled, increment counter by given delta */
   @inline final def incCounter(c: Counter, delta: Int) {
-    if (_enabled && c != null) c.value += delta
+    if (canEnable && c != null) c.value += delta
   }
 
   /** If enabled, increment counter in map `ctrs` at index `key` by one */
   @inline final def incCounter[K](ctrs: QuantMap[K, Counter], key: K) =
-    if (_enabled && ctrs != null) ctrs(key).value += 1
+    if (canEnable && ctrs != null) ctrs(key).value += 1
 
   /** If enabled, start subcounter. While active it will track all increments of
    *  its base counter.
    */
   @inline final def startCounter(sc: SubCounter): (Int, Int) =
-    if (_enabled && sc != null) sc.start() else null
+    if (canEnable && sc != null) sc.start() else null
 
   /** If enabled, stop subcounter from tracking its base counter. */
   @inline final def stopCounter(sc: SubCounter, start: (Int, Int)) {
-    if (_enabled && sc != null) sc.stop(start)
+    if (canEnable && sc != null) sc.stop(start)
   }
 
   /** If enabled, start timer */
   @inline final def startTimer(tm: Timer): TimerSnapshot =
-    if (_enabled && tm != null) tm.start() else null
+    if (canEnable && tm != null) tm.start() else null
 
   /** If enabled, stop timer */
   @inline final def stopTimer(tm: Timer, start: TimerSnapshot) {
-    if (_enabled && tm != null) tm.stop(start)
+    if (canEnable && tm != null) tm.stop(start)
   }
 
   /** If enabled, push and start a new timer in timer stack */
   @inline final def pushTimer(timers: TimerStack, timer: => StackableTimer): TimerSnapshot =
-    if (_enabled && timers != null) timers.push(timer) else null
+    if (canEnable && timers != null) timers.push(timer) else null
 
   /** If enabled, stop and pop timer from timer stack */
   @inline final def popTimer(timers: TimerStack, prev: TimerSnapshot) {
-    if (_enabled && timers != null) timers.pop(prev)
+    if (canEnable && timers != null) timers.pop(prev)
   }
 
   /** Create a new counter that shows as `prefix` and is active in given phases */
@@ -247,29 +249,33 @@ quant)
     }
   }
 
-  private var _enabled = false
   private val qs = new mutable.HashMap[String, Quantity]
 
-  /** replace with
-   *
-   *    final val canEnable = false
-   *
-   *  to remove all Statistics code from build
-   */
-  final def canEnable = _enabled
+  /** Represents whether normal statistics can or cannot be enabled. */
+  @inline final def canEnable: Boolean = StatisticsStatics.areColdStatsEnabled()
 
-  /** replace with
-   *
-   *   final def hotEnabled = _enabled
-   *
-   * and rebuild, to also count tiny but super-hot methods
-   * such as phase, flags, owner, name.
-   */
-  final val hotEnabled = false
-
-  def enabled = _enabled
+  @inline def enabled = canEnable
   def enabled_=(cond: Boolean) = {
-    if (cond && !_enabled) {
+    if (cond && !canEnable) {
+      StatisticsStatics.enableColdStats()
+    } else if (!cond && canEnable) {
+      StatisticsStatics.disableColdStats()
+    }
+  }
+
+  /** Represents whether hot statistics can or cannot be enabled. */
+  @inline def hotEnabled: Boolean = canEnable && StatisticsStatics.areHotStatsEnabled()
+  def hotEnabled_=(cond: Boolean) = {
+    if (cond && !hotEnabled) {
+      StatisticsStatics.enableHotStats()
+    } else if (!cond && hotEnabled) {
+      StatisticsStatics.disableHotStats()
+    }
+  }
+
+  import scala.reflect.internal.Reporter
+  /** Reports the overhead of measuring statistics via the nanoseconds variation. */
+  def reportStatisticsOverhead(reporter: Reporter): Unit = {
       val start = System.nanoTime()
       var total = 0L
       for (i <- 1 to 10000) {
@@ -277,9 +283,7 @@ quant)
         total += System.nanoTime() - time
       }
       val total2 = System.nanoTime() - start
-      println("Enabling statistics, measuring overhead = "+
-              total/10000.0+"ns to "+total2/10000.0+"ns per timer")
-      _enabled = true
-    }
+      val variation = s"${total/10000.0}ns to ${total2/10000.0}ns"
+      reporter.echo(NoPosition, s"Enabling statistics, measuring overhead = $variation per timer")
   }
 }
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1213,6 +1213,11 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
       checkPhaseSettings(including = true, inclusions.toSeq: _*)
       checkPhaseSettings(including = false, exclusions map (_.value): _*)
 
+      // Report the overhead of statistics measurements per every run
+      import scala.reflect.internal.util.Statistics
+      if (Statistics.canEnable)
+        Statistics.reportStatisticsOverhead(reporter)
+
       phase = first   //parserPhase
       first
     }
--- a/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
+++ b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
@@ -370,6 +370,7 @@ trait ScalaSettings extends AbsScalaSettings
 
   val YoptLogInline = StringSetting("-Yopt-log-inline", "package/Class.method", "Print a summary of inliner activity; `_` to print all, prefix match to select.", "")
 
+  import scala.reflect.internal.util.Statistics
   object YstatisticsPhases extends MultiChoiceEnumeration { val parser, typer, patmat, erasure, cleanup, jvm = Value }
   val Ystatistics = {
     val description = "Print compiler statistics for specific phases"
@@ -379,10 +380,12 @@ trait ScalaSettings extends AbsScalaSettings
       descr   = description,
       domain  = YstatisticsPhases,
       default = Some(List("_"))
-    ) withPostSetHook { _ => scala.reflect.internal.util.Statistics.enabled = true }
+    ).withPostSetHook(_ => Statistics.enabled = true)
   }
 
   def YstatisticsEnabled = Ystatistics.value.nonEmpty
+  val YhotStatistics = BooleanSetting("-Yhot-statistics", "Print hot compiler statistics for all relevant phases")
+    .withPostSetHook(_ => Statistics.hotEnabled = true)
 
   val YprofileEnabled = BooleanSetting("-Yprofile-enabled", "Enable profiling.")
   val YprofileDestination = StringSetting("-Yprofile-destination", "file", "where to send profiling output - specify a file, default is to the console.", "").
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -5562,10 +5562,10 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         else
           typedInternal(tree, mode, pt)
       )
-      val startByType = if (Statistics.canEnable) Statistics.pushTimer(byTypeStack, byTypeNanos(tree.getClass)) else null
-      if (Statistics.canEnable) Statistics.incCounter(visitsByType, tree.getClass)
+      val startByType = if (Statistics.hotEnabled) Statistics.pushTimer(byTypeStack, byTypeNanos(tree.getClass)) else null
+      if (Statistics.hotEnabled) Statistics.incCounter(visitsByType, tree.getClass)
       try body
-      finally if (Statistics.canEnable) Statistics.popTimer(byTypeStack, startByType)
+      finally if (Statistics.hotEnabled) Statistics.popTimer(byTypeStack, startByType)
     }
 
     private def typedInternal(tree: Tree, mode: Mode, pt: Type): Tree = {
--- a/src/compiler/scala/tools/nsc/util/StatisticsInfo.scala
+++ b/src/compiler/scala/tools/nsc/util/StatisticsInfo.scala
@@ -19,12 +19,16 @@ abstract class StatisticsInfo {
 
   def print(phase: Phase) = if (settings.Ystatistics contains phase.name) {
     inform("*** Cumulative statistics at phase " + phase)
-    retainedCount.value = 0
-    for (c <- retainedByType.keys)
-      retainedByType(c).value = 0
-    for (u <- currentRun.units; t <- u.body) {
-      retainedCount.value += 1
-      retainedByType(t.getClass).value += 1
+
+    if (settings.YhotStatistics.value) {
+      // High overhead, only enable retained stats under hot stats
+      retainedCount.value = 0
+      for (c <- retainedByType.keys)
+        retainedByType(c).value = 0
+      for (u <- currentRun.units; t <- u.body) {
+        retainedCount.value += 1
+        retainedByType(t.getClass).value += 1
+      }
     }
 
     val quants =
--- a/src/reflect/scala/reflect/internal/SymbolTable.scala
+++ b/src/reflect/scala/reflect/internal/SymbolTable.scala
@@ -178,7 +178,7 @@ abstract class SymbolTable extends macros.Universe
 
   final def atPhaseStack: List[Phase] = List.tabulate(phStackIndex)(i => phStack(i))
   final def phase: Phase = {
-    if (Statistics.hotEnabled)
+    if (Statistics.canEnable)
       Statistics.incCounter(SymbolTableStats.phaseCounter)
     ph
   }
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -766,7 +766,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     }
 
     final def flags: Long = {
-      if (Statistics.hotEnabled) Statistics.incCounter(flagsCount)
+      if (Statistics.canEnable) Statistics.incCounter(flagsCount)
       val fs = _rawflags & phase.flagMask
       (fs | ((fs & LateFlags) >>> LateShift)) & ~((fs & AntiFlags) >>> AntiShift)
     }
@@ -1196,7 +1196,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
      * `assertOwner` aborts compilation immediately if called on NoSymbol.
      */
     def owner: Symbol = {
-      if (Statistics.hotEnabled) Statistics.incCounter(ownerCount)
+      if (Statistics.canEnable) Statistics.incCounter(ownerCount)
       rawowner
     }
     final def safeOwner: Symbol   = if (this eq NoSymbol) NoSymbol else owner
@@ -2767,7 +2767,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     private[this] var _rawname: TermName = initName
     def rawname = _rawname
     def name = {
-      if (Statistics.hotEnabled) Statistics.incCounter(nameCount)
+      if (Statistics.canEnable) Statistics.incCounter(nameCount)
       _rawname
     }
     override def name_=(name: Name) {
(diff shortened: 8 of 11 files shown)


src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -3,51 +3,58 @@ package reflect.internal.util
 
 import scala.collection.mutable
 
-object Statistics {
+import java.lang.invoke.{SwitchPoint, MutableCallSite, MethodHandle, MethodHandles, MethodType}
+
+object StatisticsHelper {
+  /** Represents the method handle for a parameterless method returning a primitive boolean. */
+  final val BooleanMethodType: MethodType = MethodType.methodType(classOf[Boolean]).unwrap
```
jvican: The `unwrap` here could probably be removed.

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -267,19 +308,20 @@ quant)
...
-    if (cond && !_enabled) {
+    if (cond && !canEnable) {
       val start = System.nanoTime()
       var total = 0L
       for (i <- 1 to 10000) {
         val time = System.nanoTime()
         total += System.nanoTime() - time
       }
       val total2 = System.nanoTime() - start
+      // Shouldn't we replace this println with a `reporter.info`?
```
jvican: I guess this is a TODO? I'd like to get some confirmation. Problem is that we have no access to reporters from here.

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -267,19 +308,20 @@ quant)
...
       println("Enabling statistics, measuring overhead = "+
               total/10000.0+"ns to "+total2/10000.0+"ns per timer")
-      _enabled = true
+      this.setValue(true)
```
jvican: Note that here we only can enable, but we cannot disable. Is this what we want, or would we like to enable/disable them for hot compiler? Personally, I think we could support both. It's true it will have an impact and will force the deoptimization of bunch a code, but this is the case when we enable statistics anyway.

lrytz:
Started the benchmarks, builds 929 and 930 in https://scala-ci.typesafe.com/benchq/tasks

sjrd:
Scala.js does not compile scalac nor scala-reflect, so I don't think there's any insights to be had from me here ;)

jvican:
> Scala.js does not compile scalac nor scala-reflect, so I don't think there's any insights to be had from me here ;)
  
Yeah, that was my intuition, but you commented on the ticket ()https://github.com/scala/scala-dev/issues/149 saying otherwise:
  
> Just pointing out that this will make it harder to compile scalac with non-JVM backends (duh, I can't just write "Scala.js" in those cases anymore ^^). It would be good if the way this is implemented is well isolated somewhere where it is easy to change the implementation for different platforms. (as opposed to spread this implementation detail all over the place)
  
Which made me think that you're interested in compiling the compiler with Scalajs at some point in the future. :smile:

sjrd:
True, I think it would be good to preemptively isolate such things in easy-to-swap-out files. But I don't have much interest in doing the work of porting scalac to Scala.js myself.

lrytz:
Benchmark data points dripping slowly in here: https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030%7Cbb3e997faa%7Cdbb772051f&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502073067834&to=1502363630576

lrytz:
It would be interesting to compare with a version where `canInline` is a compile-time constant (`final val canEnable = false`)

jvican:
> It would be interesting to compare with a version where canInline is a compile-time constant (final val canEnable = false)
  
If this is the case, we will not be able to switch on/off statistics, right?

lrytz:
> If this is the case, we will not be able to switch on/off statistics, right?

Yes indeed, but it would be interesting to see if it's faster than the current situation where we read the `canEnable` field.

jvican:
@lrytz Gotcha, it's a good idea to know the lower and upper bound. I've added two commits to check the performance for both `false` and `true`.

lrytz:
OK, should I run the benchmarks on these 3 new commits?

jvican:
Yes, please! @lrytz

lrytz:
Updated URL with the three new commits (no data yet): https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030%7Cbb3e997faa%7Cdbb772051f%7C6d006a5461%7C23007979e6%7Cb47d0f8b8b&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502004230118&to=1502368805803

jvican:
https://github.com/scala/scala/pull/6034/commits/dbb772051fdc4b05af6f36c8f4501e6d96cd705b is slower in the current results. I think this is due to `Global` and other classes setting the value of `enabled` when, by default, all uses of `enabled_=` should be executed only if `enabled` is false -- otherwise we cause the JVM to deoptimize. I'm giong to add a commit to test this hypothesis.

jvican:
I retract what I said before, there is no place where we use the setter unnecessarily. Also, I am removing the last commit since it fails because of scala/bug#10460.
  
Now that I have all the benchmark results, I see that making `_enabled` a compile-time constant (23007979e6) is more or less on par with 6d006a5461, sometimes one wins over the other one and viceversa (for example, in `scalap` 6d006a5461 wins, but in `better-files` does). I assume this variation is natural to the benchmark infrastructure and that they yield more or less the same performance. Given these results, I infer that the use of switchpoints has been a success -- we're able to get the same performance as a compile-time constant.
  
What I cannot grasp is how having a `private var _enabled` is in anyway better than the two previous alternatives, and yet the benchmarks report that this is the case. I assume this is noise, although it doesn't fall strictly into the error margin. I would appreciate if someone can have a look. The only explanations I see plausible are:
  
1. We've used different benchmarking machines for the first commit and the last one.
2. The machine to run the benchmarks for all the commits is the same, but it's sometimes unstable.
  
Everything is green. So at this point it would be good to get some feedback from a reviewer.

jvican:
Note that if the results look good, I'll need to remove https://github.com/scala/scala/pull/6034/commits/23007979e61ab441f10e57d24d25b9c955d143d8.

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -250,13 +257,47 @@ quant)
...
-   *    final val canEnable = false
-   *
-   *  to remove all Statistics code from build
+  private var switchPoint: SwitchPoint = new SwitchPoint()
+  private final val DefaultHandle: MethodHandle = {
+    val handle = MethodHandles.lookup().findVirtual(this.getClass, "defaultValue", StatisticsHelper.BooleanMethodType)
+    val bounded = handle.bindTo(this)
+    setTarget(bounded)
+    bounded
+  }
```
retronym: I'd suggest to put all the method handle contants into a Java source file so they can be try `static final` constants. 

https://wiki.openjdk.java.net/display/HotSpot/Deconstructing+MethodHandles

> If such a reference to a MethodHandle is held in a static final field then the runtime should be able to constant fold invocations on that reference and what it holds when inlining occurs.

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -267,19 +308,20 @@ quant)
    */
   final val hotEnabled = false
```
retronym: To get a better look at the performance overhead (or absense thereof), perhaps you could extend the switchpoint technique to `hotEnabled`.

jvican: I've decided to reuse `canEnable` for `hotEnabled` as well, so that when we do measure statistics we get the whole information. Let me know if you're ok with this.

src/reflect/scala/reflect/internal/util/Statistics.scala
jvican: This is cool. I'll try this.

jvican: @retronym https://github.com/scala/scala/pull/6034/commits/63953e13887c94b3b2c70349e3ccf146c3469e51.

jvican:
There seems to be some problem here with the CLA signature checker not detecting that I've signed the CLA in commit 01a4287 and blocking the rest of the commits from being compiled/tested. Any hints on how I could solve it?

jvican:
There's a noticeable hit in performance (~10%) when `_hotEnabled` is forwarding to `canEnable`. This can be seen [in the benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030|bb3e997faa|6d006a5461|63953e1388|5254231683&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502073067834&to=1502882030576). I'm surprised by this -- I presume this change exhausts the JVM inlining budget?
  
The use of statics in https://github.com/scala/scala/pull/6034/commits/63953e13887c94b3b2c70349e3ccf146c3469e51 seems to have no effect in performance compared to the previous approach.
  
What's your view on this @retronym? I haven't made `_hotEnabled` use its own switchpoint because I believe that `MutableCallSite`s can only have one target -- making such a change would require the switchpoint strategy to be more abstract, being potentially slower (I'm actually not sure if this would be noticeable or not).

jvican:
I think the way to move forward here is to actually remove `_hotEnabled` and use `canEnable` in all its call-sites. I don't think we should distinguish between the two. It makes sense to me to capture stats about those hot methods when I do want to see all the statistics.

jvican:
Okay, I've made some arrangements. You can check all the most important benchmarks here: https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030|bb3e997faa|6d006a5461|63953e1388|5254231683|b5a3aee4db&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502073067834&to=1502882030576.
  
At the end, the statics trick did have an effect on performance, https://github.com/scala/scala/pull/6034/commits/b5a3aee4db62529d085e6b0bde62dac040bc1f83 addresses an oversight I had.
  
I think this is already ready for another review.

jvican:
Okay, wrapping up, after all the experimentation that I've done with the different commits, my conclusions are:
  
1. Last two commits reduce the overhead of using this approach and enabling statistics compiler-wide. This approach uses `AlmostFinalValue` as proposed by Rémi Forax. This approach, even though boxed, is better than the one I started with (a special-cased implementation for Statistics) because it allows callers to forward to static fields (concretely `BooleanContainer.getDefault()`, instead of a non-static field). The previous approach caused a ~10% slowdown.
2. The statistics infrastructure now is a little bit more polished and enabling `-Ystatistics` works (not like before, that only showed a very constrained subset of all the available statistics).
3. The `println` in the setter for enabling statistics has been replaced by the use of a reporter, and the information on the variation of the statistics measurements happens in every run instead of per global.
  
All in all, this PR is ready for review. If you're okay with the final implementation, I'll squash accordingly. I'm planning to follow up with another proposal that will add more statistics.

jvican:
By the way, forgot to mention. Final benchmarks for:

* [Cold compilation](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030|bb3e997faa|6d006a5461|63953e1388|5254231683|b5a3aee4db|4119e8db7|6376343ae7&var-source=All&var-bench=ColdScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502349158966&to=1502913227133). No remarkable difference between any of my implementations.
* [Hot compilation](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030|bb3e997faa|6d006a5461|63953e1388|5254231683|b5a3aee4db|4119e8db7|6376343ae7&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502349158966&to=1502913227133). Second-to-last commit is the winner.

jvican:
@lrytz @retronym Can you have another look at this?

retronym:
What is the overhead of having statistics enabled? Ideally it should not skew the overall runtime too much (perhaps 10% overhead is okay). If the overhead is too high, we'll need to re-separate the toggle switches for hot and regular stats.

retronym:
Oh, I see that in https://github.com/scala/scala/pull/6034/commits/347d589158816848d7584af9366f6aa4696b2f24 you've done that already.

jvican:
> What is the overhead of having statistics enabled? Ideally it should not skew the overall runtime too much (perhaps 10% overhead is okay). If the overhead is too high, we'll need to re-separate the toggle switches for hot and regular stats.
> Oh, I see that in 347d589 you've done that already.
  
Right now, all of them are enabled and the overhead (when not enabled) is the one reported in my last comments. However, I have not computed the overhead when they are enabled because of https://github.com/scala/bug/issues/10460.

src/reflect/scala/reflect/internal/util/BooleanContainer.java
```diff
@@ -0,0 +1,61 @@
...
+  private static final AlmostFinalValue<BooleanContainer> DEFAULT_VALUE = new AlmostFinalValue<BooleanContainer>() {
+    @Override
+    protected BooleanContainer initialValue() {
+        return new FalseBooleanContainer();
+    }
+  };
+
+  private static final MethodHandle DEFAULT_VALUE_GETTER = DEFAULT_VALUE.createGetter();
+  
+  public static BooleanContainer getDefault() {
```
retronym: Should this actually be named `isStatisticsEnabled`? AFAICT it can't be used for more than one purpose, so it seems strange to leave it here.

If my assumptions are true, I'd suggest to move `getDefault` and the static fields it depends on to a class named `StatisticsStatics` or somesuch.

retronym:
We can use `-pextraArgs=-Ystatistics` to have the benchmark harness enable statistics via the setting.

```
for V in 2.12.4-bin-347d589-SNAPSHOT; 2.12.4-bin-042895b-SNAPSHOT 2.12.3; do sbt ++$V "hot -psource=scalap -f 1" "hot -psource=scalap -f 1 -pextraArgs=-Ystatistics"; done
YextraArgs=-Ystatistics"; done
```

Rough numbers, one fork on my local machine.

| Version | -Ystatistics:false |  -Ystatistics:true | |
|---------|------------|-------|--|
| 2.12.3   |   781      |   876          |  |
| 2.12.4-bin-a8c43dc | 801 | 1035 | parent of this PR |
| 2.12.4-bin-347d589 | 811 | 1035 | "hot" stats commented out |
| 2.12.4-bin-042895b |  816   |  1067 | |

 - The overhead of enabling statistics in 2.12.3 was low, largely because the setting work for many counters guarded by an `if` of a `false` compile time constant
 - The overhead of a working statistics implementation is 20-25%.
    - It doesn't seem to make much difference whether the counters that had been thought of as `hot` (and previously controlled with a separate constant) are enabled or not.

It would be prudent to figure out which statistics are really the hot ones, and maybe remove them.

retronym:
YourKit suggests that maintaining the `QuantMap` of the different tree subclasses to instantiation count (in `TreesStats.nodeByType`) and typer timings per tree subclass (`TyperStat.byTypeStack`) are the dominant overhead.

image

I have found the YourKit seems to overstate the cost of `Object.hashCode` (maybe its profiling somehow interferes with the intrisified nature of that?).

Anyway, removing stats: https://github.com/retronym/scala/commit/6e993e63fba1b52357fc79f14e1b161ccaff384c drops the cost of `-Ystatistics` somewhat, to 945ms (15% overhead) (from 1067ms, or 30% overhead, above).

Options:
  - Do nothing, defer a decision and implementation to a later PR
  - Move those "actually" hot statistics behind a secondary almost-final value and compiler option
  - Make them more efficient by sampling, rather than counting every invocation. Naively, bump a counter and sample when `counter % N == 0`. Less naively, bump a cheap pseudorandom value and sample when `counter % N == 0` (to avoid bias from patterns in data).
  - Delete them entirely

retronym:
`StatisticsInfo.print` is also fairly expensive (3% of total compile time) as it traverses all trees.

src/reflect/scala/reflect/internal/util/BooleanContainer.java
jvican: Hmm, the naming is generic is because I thought that this `BooleanContainer` could be used in other scenarios (that require an almost final boolean), and so didn't want to hardcode it. There's in theory nothing tied to statistics on this class on purpose.
  
I'm happy to follow your advice if you think that this genericity will not be worth it in the future... but please confirm :smile:.

retronym: Yes, please do. IMO this doesn't make any more sense than exposing `var default: Boolean` here :)

jvican:
Thanks for the in-depth analysis, Jason.
  
> Make them more efficient by sampling, rather than counting every invocation. Naively, bump a counter and sample when counter % N == 0. Less naively, bump a cheap pseudorandom value and sample when counter % N == 0 (to avoid bias from patterns in data).
  
I've always liked the idea of pseudorandom sampling. I think it could give good results on these super hot counters. However, before trying it out I think it's better to try easier alternatives like using a new almost final value for those hot stats you've identified + `StatisticsInfo.print`. If working, that should put the overhead of running normal statistics around 12%. This is implemented in https://github.com/scala/scala/pull/6034/commits/1866d1b0d1dfe67be0c6873441353f2871678840.
  
IMO, this effort could be split into a different PR. If https://github.com/scala/scala/pull/6034/commits/1866d1b0d1dfe67be0c6873441353f2871678840 doesn't reduce the overhead, I suggest merging what we have now (I'll clean up the git history first) and defer a decision and implementation to another PR.

jvican:
@retronym These are my results benchmarking locally with TurboBoost disabled. I could only benchmark 2.12.3 and the last commit in this PR, since the other artifacts weren't accessible, even though I have the following in a global sbt file:
```scala
resolvers ++= (
  if (scalaVersion.value.contains("-bin"))
     List("scala-integration" at "https://scala-ci.typesafe.com/artifactory/scala-integration/")
  else Nil
)
```
  
## Results for 2.12.4-bin-042895b-SNAPSHOT
```
[info] Setting version to 2.12.4-bin-042895b-SNAPSHOT
[info] Benchmark                                   (corpusVersion)  (extraArgs)  (source)    Mode  Cnt     Score    Error  Units
[info] HotScalacBenchmark.compile                           latest                 scalap  sample  220  1476.834 ± 34.233  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00             latest                 scalap  sample       1262.486           ms/op
[info] HotScalacBenchmark.compile:compile·p0.50             latest                 scalap  sample       1437.598           ms/op
[info] HotScalacBenchmark.compile:compile·p0.90             latest                 scalap  sample       1687.368           ms/op
[info] HotScalacBenchmark.compile:compile·p0.95             latest                 scalap  sample       1797.888           ms/op
[info] HotScalacBenchmark.compile:compile·p0.99             latest                 scalap  sample       2012.008           ms/op
[info] HotScalacBenchmark.compile:compile·p0.999            latest                 scalap  sample       2099.249           ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999           latest                 scalap  sample       2099.249           ms/op
[info] HotScalacBenchmark.compile:compile·p1.00             latest                 scalap  sample       2099.249           ms/op
[info] Benchmark                                   (corpusVersion)   (extraArgs)  (source)    Mode  Cnt     Score    Error  Units
[info] HotScalacBenchmark.compile                           latest  -Ystatistics    scalap  sample   68  1579.926 ± 54.677  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00             latest  -Ystatistics    scalap  sample       1411.383           ms/op
[info] HotScalacBenchmark.compile:compile·p0.50             latest  -Ystatistics    scalap  sample       1561.330           ms/op
[info] HotScalacBenchmark.compile:compile·p0.90             latest  -Ystatistics    scalap  sample       1707.291           ms/op
[info] HotScalacBenchmark.compile:compile·p0.95             latest  -Ystatistics    scalap  sample       1861.747           ms/op
[info] HotScalacBenchmark.compile:compile·p0.99             latest  -Ystatistics    scalap  sample       2168.455           ms/op
[info] HotScalacBenchmark.compile:compile·p0.999            latest  -Ystatistics    scalap  sample       2168.455           ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999           latest  -Ystatistics    scalap  sample       2168.455           ms/op
[info] HotScalacBenchmark.compile:compile·p1.00             latest  -Ystatistics    scalap  sample       2168.455           ms/op
```
  
## Results for 2.12.3
```
[info] Setting version to 2.12.3
[info] Benchmark                                   (corpusVersion)  (extraArgs)  (source)    Mode  Cnt     Score    Error  Units
[info] HotScalacBenchmark.compile                           latest                 scalap  sample  218  1489.257 ± 42.212  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00             latest                 scalap  sample       1247.805           ms/op
[info] HotScalacBenchmark.compile:compile·p0.50             latest                 scalap  sample       1442.841           ms/op
[info] HotScalacBenchmark.compile:compile·p0.90             latest                 scalap  sample       1716.729           ms/op
[info] HotScalacBenchmark.compile:compile·p0.95             latest                 scalap  sample       1867.409           ms/op
[info] HotScalacBenchmark.compile:compile·p0.99             latest                 scalap  sample       2194.040           ms/op
[info] HotScalacBenchmark.compile:compile·p0.999            latest                 scalap  sample       2214.593           ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999           latest                 scalap  sample       2214.593           ms/op
[info] HotScalacBenchmark.compile:compile·p1.00             latest                 scalap  sample       2214.593           ms/op
[info] Benchmark                                   (corpusVersion)   (extraArgs)  (source)    Mode  Cnt     Score    Error  Units
[info] HotScalacBenchmark.compile                           latest  -Ystatistics    scalap  sample   69  1545.206 ± 57.351  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00             latest  -Ystatistics    scalap  sample       1367.343           ms/op
[info] HotScalacBenchmark.compile:compile·p0.50             latest  -Ystatistics    scalap  sample       1503.658           ms/op
[info] HotScalacBenchmark.compile:compile·p0.90             latest  -Ystatistics    scalap  sample       1690.305           ms/op
[info] HotScalacBenchmark.compile:compile·p0.95             latest  -Ystatistics    scalap  sample       1844.445           ms/op
[info] HotScalacBenchmark.compile:compile·p0.99             latest  -Ystatistics    scalap  sample       2088.763           ms/op
[info] HotScalacBenchmark.compile:compile·p0.999            latest  -Ystatistics    scalap  sample       2088.763           ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999           latest  -Ystatistics    scalap  sample       2088.763           ms/op
[info] HotScalacBenchmark.compile:compile·p1.00             latest  -Ystatistics    scalap  sample       2088.763           ms/op
```
  
## Conclusion
I'm surprised by these results, up until a point in which I would like someone to rerun the benchmarks and confirm them. Following these results, the overhead between the two is negligible. As a reminder, the last commit moves the statistics infrastructure in the initializer of `Trees`, `Typers` and `StatisticsInfo.print` into a `hotEnabled` almost final value which is turned on by `-Yhot-statistics` compiler flag (if `-Ystatistics` is not present, hot statistics won't be shown). 
  
Note that last commits are failing because the builds for https://github.com/scala/scala/pull/6034/commits/9a2a0cdccf6ff71ad93d31195b1563fbb6811600 were aborted. Can someone restart the CI for them?

retronym:
I have this to access PR validation builds. 

```
resolvers ++= (
  if (scalaVersion.value.endsWith("-SNAPSHOT") || scalaVersion.value.contains("-nightly"))
    List(
      "pr-scala snapshots" at "https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/",
      Resolver.mavenLocal
    )
  else
    Nil
)
```

https://scala-ci.typesafe.com/artifactory/scala-integration/ contains fully bootstrapped builds triggered by the benchmark runs.

Even without these, it is possible to run `sbt setupPublishCore publishLocal` to publish to your `~/.ivy2`.

retronym:
/rebuild

https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/5457/consoleText
```
[info] + Range inclusiveByOne.take: OK, passed 100 tests.
Build timed out (after 150 minutes). Marking the build as aborted.
Build was aborted
Archiving artifacts
Notifying endpoint 'HTTP:https://scala-ci.typesafe.com/benchq/webhooks/jenkins'
```

jvican:
I've just benchmarked everything again to make sure the results are consistent.
  
1. [Logs for Scala 2.12.3](https://gist.github.com/jvican/a25baa6d3fa68556d0e2d11b7cd4559d)
1. [Logs for Scala 2.12.4-bin-042895b-SNAPSHOT](https://gist.github.com/jvican/8cfcb07076de0c67f2a79e301071cb41)
1. [Logs for Scala 2.12.4-bin-347d589-SNAPSHOT](https://gist.github.com/jvican/871c4539ec7a6d74d62fee53c658ae56)
1. [Logs for Scala 2.12.4-bin-1866d1b-SNAPSHOT](https://gist.github.com/jvican/97dac86f39432942a50758201ad3f489)
  
Rough numbers without error margins (which are more or less consistent):
  
| Version | -Ystatistics:false |  -Ystatistics:true | |
|---------|------------|-------|--|
| 2.12.3   |   1289      |   1386          |  |
| 2.12.4-bin-a8c43dc | ??? | ??? | parent of this PR |
| 2.12.4-bin-347d589 | 1346 | 1633 | "hot" stats commented out |
| 2.12.4-bin-042895b |  1357   |  1625 | all stats enabled & together |
| 2.12.4-bin-1866d1b |  1326   |  1370 | "hot" stats split |
  
I don't have benchmarks for `2.12.4-bin-a8c43dc` because `2.12.4-bin-1866d1b` provides almost the same benchmark results than 2.12.3, confirming that the overhead between the two is minimal, both where statistics are disabled and they are enabled.
  
I guess this can be reviewed again.

jvican:
I have squashed the commits of this PR. The previous git history can be found in the [`stats-non-squashed`](https://github.com/scalacenter/scala/tree/stats-non-squashed).

retronym:
Scheduled another [benchmark](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=bfc660a%7Ca641b833f4%7C7796bb7bf2%7C2165b701dc%7C851047970b%7C311a576030&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40&from=1500677652321&to=1505861652321) to be doubleplus sure.

retronym (approved):
Great work, thanks for your persistence in finding the optimal solution.

PR #6035, opened 2017-08-11
[REVERTED] Deprecate Tuple[23]Zipped.invert, rename to Tuple[23]Zipped.transpose.

kamilkloch:
Method `invert` defined in `Tuple[23]Zipped` object has an unfortunate name - `transpose` fits the result much better:
```scala
@ (Seq(1, 2, 3), Seq('a', 'b')).invert 
res4: Seq[(Int, Char)] = List((1, 'a'), (2, 'b'))
```
This PR deprecates `invert` and replaces it with `transpose`.
The topic already been discussed in scala/scala#5870

Fixes scala/scala-dev#413

### DIFF
--- a/src/library/scala/runtime/Tuple2Zipped.scala
+++ b/src/library/scala/runtime/Tuple2Zipped.scala
@@ -117,7 +117,14 @@ final class Tuple2Zipped[El1, Repr1, El2, Repr2](private val colls: (Traversable
 
 object Tuple2Zipped {
   final class Ops[T1, T2](private val x: (T1, T2)) extends AnyVal {
+    @deprecated("Use `transpose` instead", "2.13.0")
     def invert[El1, CC1[X] <: TraversableOnce[X], El2, CC2[X] <: TraversableOnce[X], That]
+      (implicit w1: T1 <:< CC1[El1],
+                w2: T2 <:< CC2[El2],
+                bf: scala.collection.generic.CanBuildFrom[CC1[_], (El1, El2), That]
+      ): That = transpose
+
+    def transpose[El1, CC1[X] <: TraversableOnce[X], El2, CC2[X] <: TraversableOnce[X], That]
       (implicit w1: T1 <:< CC1[El1],
                 w2: T2 <:< CC2[El2],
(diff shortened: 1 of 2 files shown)


lrytz (approved):

milessabin:
This breaks shapeless, so I'd like us to rethink here ...

Firstly, `transpose` isn't a much better name for the operation defined here than `invert` ... something akin to `sequence` would be more typical, however, that will most likely conflict with Cats and Scalaz.

milessabin:
FWIW, I agree that `invert` isn't a great name either, but the cure is worse than the disease ...

kamilkloch:
@milessabin I do believe `transpose` is the proper name, and it also exists in scala collections. Please take a look:
```scala
@ Seq(Seq(1, 2, 3), Seq(4, 5, 6)).transpose 
res0: Seq[Seq[Int]] = List(List(1, 4), List(2, 5), List(3, 6))
```

milessabin:
`transpose` is the right name for that operation on nested sequences. By analogy, `transpose` is the right name for the operation on nested tuples that the shapeless operation provides.

In cases where we're swapping the nesting of type constructors (ie. `F[G[T]] => G[F[T]]`) there is a very general operation, typically called `sequence` which does the job ... see the write up of the `Traverse` type class from Cats [here](http://eed3si9n.com/herding-cats/Traverse.html).

lrytz:
So you say `transpose` in the collections goes from `C[X[T]]` to `C[C[T]]` for any `X`, where the operation here goes from `Tuple[C[T]]` to `C[Tuple[T]]`. So `transpose` should go from `Tuple[C[T]]` to `Tuple[Tuple[T]]` (where possible for `C`).

I'm also a bit hesitant to introduce the name `sequence`, as we don't use it in the collections so far. The only place is in `Future`, as a factory. But maybe it's the right thing to do.. we could see how well it would go through a community build.

adriaanm:
Breaking shapeless and thus the community build is enough of a reason to (temporarily) revert this PR so that we can cut M3 soon. Can we just straight deprecate this and provide an extension somewhere else? Collection-contrib or so? /cc @szeiger

PR #6037, opened 2017-08-13
Dotty style downwards comparisons for implicit search and overloading resolution

milessabin:
This is a backport of https://github.com/lampepfl/dotty/commit/89540268e6c49fb92b9ca61249e46bb59981bf5a to scalac. As in the Dotty implementation the specificity comparison is as-if all contravariant positions in type constructors were covariant. The feature is enabled by specifying `-Xsource:3.0`.

Initially this broke `test/files/run/t2030.scala` in exactly the same way as in Dotty as reported by @smarter [here](https://github.com/lampepfl/dotty/pull/1246#issuecomment-220710595). Since then the landing of the new collections has fixed this.

This fixes https://github.com/scala/bug/issues/2509. It also aligns with the contravariant behaviour hoped for in https://github.com/scala/bug/issues/7768, although it differs on the covariant behaviour, which is unchanged. I think that's a reasonable outcome and that https://github.com/scala/bug/issues/7768 can be declared as fixed.


smarter:
Nice!

> The new specificity behaviour is behind flag -Ydotty-specificity.

I think `-Xsource:3.0` was proposed for this sort of things: https://github.com/scala/scala-dev/issues/374

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -221,6 +221,7 @@ trait ScalaSettings extends AbsScalaSettings
   val YdisableFlatCpCaching  = BooleanSetting    ("-YdisableFlatCpCaching", "Do not cache flat classpath representation of classpath elements from jars across compiler instances.")
   val YpartialUnification = BooleanSetting ("-Ypartial-unification", "Enable partial unification in type constructor inference")
   val Yvirtpatmat     = BooleanSetting    ("-Yvirtpatmat", "Enable pattern matcher virtualization")
+  val YdottySpecificity = BooleanSetting ("-Ydotty-specificity", "Enable Dotty compatible rules for overload selection by specifity")
```
som-snytt: > selection by specifity")

That's how I usually pronounce it.

milessabin: Curs mi speling ...

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -823,13 +823,46 @@ trait Infer extends Checkable {
...
+                def apply(tp: Type): Type = tp match {
+                  case TypeRef(pre, sym, args) if variance > 0 && sym.typeParams.exists(_.isContravariant) =>
+                    val flippedSym =
+                      subst.get(sym) match {
+                        case None =>
+                          val flippedClassSym = sym.owner.newClassSymbol(sym.name.toTypeName, sym.pos, sym.flags)
+                          val flippedParams = sym.rawInfo.typeParams.map { param =>
+                            val flippedFlags = if(param.isContravariant) (param.flags&(~CONTRAVARIANT))|COVARIANT else param.flags
+                            flippedClassSym.newTypeParameter(param.name.toTypeName, param.pos, flippedFlags) setInfo param.info
+                          }
```
retronym: `param.info` might refer to the original type parameters, so more substitution is needed. Cloning the class upfront takes care of all the rewiring.

```
scala> class C[-A <: C[A]]
defined class C

scala> val CClass = symbolOf[C[_]]
CClass: $r.intp.global.TypeSymbol = class C

scala> val CClass_clone = CClass.cloneSymbol(CClass.owner)
CClass_clone: CClass.TypeOfClonedSymbol = class C

scala> CClass_clone.rawInfo.typeParams.foreach(sym => if (sym.isContravariant) sym.resetFlag(Flag.CONTRAVARIANT).setFlag(Flag.COVARIANT))

scala> CClass_clone.defString
res3: String = class C[+A <: C[A]] extends AnyRef

scala> {settings.uniqid.value = true; println(CClass.defString); println(CClass_clone.defString); settings.uniqid.value = false}
class C#132392[-A#132393 <: C#132392[A#132393]] extends AnyRef#1932
class C#145639[+A#145640 <: C#132392[A#145640]] extends AnyRef#1932
```

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -823,13 +823,46 @@ trait Infer extends Checkable {
...
+              // Backport of fix for https://github.com/scala/bug/issues/2509
+              // from Dotty https://github.com/lampepfl/dotty/commit/89540268e6c49fb92b9ca61249e46bb59981bf5a
+              val flip = new TypeMap(trackVariance = true) {
+                var subst: Map[Symbol, Symbol] = Map.empty
+                def apply(tp: Type): Type = tp match {
+                  case TypeRef(pre, sym, args) if variance > 0 && sym.typeParams.exists(_.isContravariant) =>
+                    val flippedSym =
+                      subst.get(sym) match {
+                        case None =>
+                          val flippedClassSym = sym.owner.newClassSymbol(sym.name.toTypeName, sym.pos, sym.flags)
```
retronym: Surely the new class symbols created by `flip` will always be treated as unrelated by the subsequent `<:<`? Maybe that gives the expected result for the test cases.

Dotty's implementation appears to make a type alias instead.

```
class C[-A]
...
type C'[+A] = C[A] // synthetic (ill-varying) type alias
```

retronym (dismissed):
I'd also like to see how this would be described in the spec, assuming it were to become default behaviour. Sometimes these concepts are easier to describe in comments in the implenentation than they are in the more limited vocabulary of the spec.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -823,13 +823,46 @@ trait Infer extends Checkable {
...
     private def isAsSpecificValueType(tpe1: Type, tpe2: Type, undef1: List[Symbol], undef2: List[Symbol]): Boolean = tpe1 match {
       case PolyType(tparams1, rtpe1) =>
         isAsSpecificValueType(rtpe1, tpe2, undef1 ::: tparams1, undef2)
       case _                         =>
         tpe2 match {
           case PolyType(tparams2, rtpe2) => isAsSpecificValueType(tpe1, rtpe2, undef1, undef2 ::: tparams2)
-          case _                         => existentialAbstraction(undef1, tpe1) <:< existentialAbstraction(undef2, tpe2)
+          case _                         =>
+
+            if (settings.YdottySpecificity) {
```
retronym: This approach looks somewhat expensive, but we could cache the flipped version of each contravariant class symbol by spending a few fields in `ClassSymbol`.

test/junit/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -0,0 +1,97 @@
...
+import scala.tools.testing.BytecodeTesting
+
+class A
+class B extends A
+trait X[+T]
+trait Y[T]
+trait Z[-T]
+
+@RunWith(classOf[JUnit4])
+class InferencerTests extends BytecodeTesting {
```
retronym: 👍 for unit tests!

src/compiler/scala/tools/nsc/typechecker/Infer.scala
milessabin: Thanks. That simplifies things quite a bit.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
milessabin: Yes, this also helps.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
milessabin: I don't think that's quite what's going on in Dotty's implementation. In the `flip` in Dotty the `TypeMap` traversal appears to traverse through to the type members which correspond to the type params of the `TypeRef` in the scala implementation. These type members then have their variance individually flipped, via a type alias, as you say. But I think the mechanism in scalac has to be different ... we have to work at the `TypeRef`/type params level.

On your first point, yes you're right that new symbols will be unrelated by `<:<` except in the trivial case where the outer type constructor is the same in both cases (here the `subst` map ensures that the same flipped symbol is used). I'm fairly confident that my implementation here matches the one in Dotty, but it's possible that the Dotty implementation gets this wrong as well.

I think however, that in cases where the outer type constructor differs, we will always want to say that neither is specific as the other. The following is the main problem case (included in the latest commit, Dotty agrees with the implementation here),
```scala
trait Z[-T]
trait Quux[-T] extends Z[T]
class A
class B extends A
```
Given the above, what is the relative specificity of `Quux[B]` and `Z[A]`? Both Dotty and this implementation say that neither is as specific as the other even though `Z[B]` is as specific as `Z[A]` and `Quux[B]` is a subtype of `Z[B]`. @smarter what are your thoughts on this?

smarter: Intuitively I would expect `Quux[B]` to be more specific than `Z[A]` for the reasons you gave. The fact that it doesn't in dotty seems like an oversight to me /cc @odersky .

milessabin: @retronym is there any way of preserving the subtype relationships of the cloned outer `ClassSymbol`s?

smarter: I've opened https://github.com/lampepfl/dotty/issues/2974, you might want to weight in on whether you find the top-level-only-flipping semantics sensible or not.

smarter:
Could you try making `Ordering` contravariant on top of this PR ? (cf http://github.com/scala/bug/issues/7179, this might require using @unsafeVariance to deal with Java superclasses)

milessabin:
@smarter aside from the inevitable issues with `max` and `min` it looks fine.

milessabin:
Quick and dirty PoC of a contravariant `(Partial)Ordering` and `Equiv` here: https://github.com/milessabin/scala/commit/9ca6be83dd25564560eef72be21da958c52e8665.

I made `max` and `min` extension methods of `Ordering[T]` to avoid the variance issue and didn't bother to handle the `Float` and `Double` specializations.

It compiles and partests successfully.

smarter:
> I made max and min extension methods of Ordering[T] to avoid the variance issue

Would this work?
```scala
def max[U <: T](x: U, y: U): U = if (gteq(x, y)) x else y
```

milessabin:
> Would this work?

Yes, but it'd be a bit more work. The nested `Ops` class interferes.

milessabin:
Latest commit fixes the problem that @retronym pointed out by normalizing the LHS to the type constructor of the RHS via `baseType`. My reading of the discussion motivating this change is that people's expectations are that specificity runs in the same direction as extension rather than subtyping, so the use of base types at the outer level seems appropriate.

adriaanm:
I'd prefer to have this target 2.13, as it's about exploring dotty compatibility.

milessabin:
I'm fine with this going into 2.13.x. Would you like me to change `-Ydotty-specificity` to `-Xsource:2.13`?

The Dotty bug corresponding to the issue that @retronym raised has been fixed with same test case as included in this PR so we can be reasonably confident that the semantics match: https://github.com/lampepfl/dotty/issues/2974.

adriaanm:
Thanks, yes, let's standardize on `-Xsource`. Regarding the target of 2.13 v 3.0, we can tweak once it's clear what the impact on user code is.

ctongfei:
This is great :-) Cats can make all these typeclasses contravariant now!

milessabin:
Rebased against 2.13.x ...

@adriaanm where does this leave `-Xsource:2.13` ... should I just make this unconditional? Or do we need a `-Xsource:dotty`? Or something else?

smarter:
I would vote for making this unconditional so that it can influence the design of the collection-strawman

milessabin:
I would be very happy to make this unconditional ... @adriaanm?

adriaanm:
More work is needed before we can get to deciding whether this can go in and which flag (if any) it should be gated by. In short, 1) a correct implementation and 2) a detailed explanation of its pros & cons, including how it affects existing code bases and future ones (the new collections).

First, the code has evolved in dotty since the commit you've ported: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Applications.scala#L1138. Also, your port of the original commit is not obviously correct to me, and it didn't carry over the nice comment that accompanies the dotty implementation. Can you explain your implementation strategy and why it's different from dotty's? 

The higher-level concern is that this illustrates this is evolving in dotty. Are we ready to call the latest approach ready for production? We also need to evaluate how it affects existing code bases. If this helps the new collections design -- great! That's a good motivator and test bed, but it's not a sufficient condition.

smarter:
> The higher-level concern is that this illustrates this is evolving in dotty.

The only change to the behavior of dotty here was a fix for https://github.com/lampepfl/dotty/issues/2974 and Miles updated his PR to match the agreed-upon behavior. No other changes in this area are planned, though of course we need to see how well this works in practice. I think getting this PR in scalac and seeing what happens on the community build would be very helpful for that.

adriaanm:
Ok, thanks for the background. I'm all for experimentation here, and alignment with dotty, but this is a tricky part of the language and I would like to see a deep dive in how this is implemented and how it will impact existing code. 

I still don't understand how the code I linked to and the code in this PR relate. What does `FunctionOf` do in dotty, and why doesn't it appear here -- is it a cheaper way to flip variance? It would be nice to avoid the symbol cloning (cloning without substitution makes me suspicious, without looking too deeply). Should we do the comparison before or after existential abstraction?

smarter:
> I still don't understand how the code I linked to and the code in this PR relate. What does FunctionOf do in dotty, and why doesn't it appear here -- is it a cheaper way to flip variance?

Yeah basically you replace all type arguments `Foo` that appear in contravariant positions by a function type `Foo => ()` to flip their variance, it's kind of a hack :).

milessabin:
The impact will be similar, though not quite as significant, as partial unification ... people have been working around the old specificity behaviour wrt to contravariant type classes in a variety of ad hoc and more or less unsatisfactory ways for years. This change in Dotty and Scala gives people the behaviour they expected all along and makes the workarounds unnecessary.

Existing code will break, as was the case with partial unification, but I think that's a price worth paying, and I'm fairly confident that most people working with contravariant type classes would agree.

adriaanm:
This one is going to need more discussion and refinement. Moving to M4

milessabin:
Rebased.

milessabin:
Rebased and made unconditional.

milessabin:
I'm not able to reproduce the ScalaCheck failure reported above locally ... fingers crossed it's spurious.

SethTisue:
also fixes https://github.com/scala/bug/issues/7768 sez Miles

milessabin:
I've added a test confirming that this aligns with the hoped for behaviour in the contravariant case in https://github.com/scala/bug/issues/7768. The covariant case is unchanged, which I think is correct.

milessabin:
Squashed and rebased.

`testAll` runs locally for me ... hopefully the Jenkins failure will go away this time.

milessabin:
I'm still seeing the ScalaCheck failure on Jenkins, but not locally ... does anyone have any ideas? @SethTisue?

milessabin:
That's a ScalaCheck test _compile_ failure, BTW. Locally it compiles and runs just fine. The Jenkins log ends with,
```
[test] [error]   - scalacheck/test:compileIncremental failed: Compilation failed
```
Could this be some sort of incremental compilation failure? Stale binary artefacts in the workspace?

SethTisue:
> Stale binary artefacts in the workspace?

seems unlikely, https://scala-ci.typesafe.com/job/scala-2.13.x-validate-main/328/consoleFull shows `git clean -fdx` running

SethTisue:
"might be due to the files being compiled in a different order", suggests @smarter 

since we're otherwise stumped, some things that _could_ conceivably be relevant _somehow_:

* case-sensitive filesystem or not? I think Miles is on a Mac (case-insensitive), Jenkins is Linux (case sensitive)
* the ScalaCheck tests are all in the empty package
* old ScalaCheck build? I've addressed this in https://github.com/scala/scala/pull/6388

retronym:
FTR, I tried this locally to see if I could reproduce, but things compiled successfully:

```
/code/scala on pr/6037*
⚡ sbt ';clean;setupPublishCore;publishLocal;generateBuildCharacterPropertiesFile' && sbt -Dstarr.version=$(cat buildcharacter.properties |  grep '^maven.version.number=' | cut -d '=' -f 2) test:compile
```

Actually, giving this one more turn reproduces the error. (Note that the `scalacheck/test:clean` task actually does work without #6389 in SBT 0.13.17 because it uses a standard output directory.)

```
> scalacheck/test:clean

> scalacheck/test:compile
[info] Updating {file:/Users/jz/code/scala/}scalacheck...
[info] Resolving org.scala-sbt#test-interface;1.0 ...
[info] Done updating.
[info] Compiling 44 Scala sources to /Users/jz/code/scala/target/scalacheck/test-classes...
[error] /Users/jz/code/scala/test/scalacheck/treeset.scala:41: ambiguous implicit values:
[error]  both method newCanBuildFrom in object SortedSet of type [A](implicit ord: Ordering[A])scala.collection.generic.CanBuildFrom[scala.collection.immutable.SortedSet.Coll,A,scala.collection.immutable.SortedSet[A]]
[error]  and method newCanBuildFrom in class SortedSetFactory of type [A](implicit ord: Ordering[A])scala.collection.generic.CanBuildFrom[scala.collection.immutable.TreeSet.Coll,A,scala.collection.immutable.TreeSet[A]]
[error]  match expected type scala.collection.generic.CanBuildFrom[scala.collection.immutable.TreeSet[Int],(Int, Int),That]
[error]     subject.zip(subject.tail).forall { case (x, y) => x < y }
[error]                ^
[error] /Users/jz/code/scala/test/scalacheck/treeset.scala:41: value < is not a member of Any
[error]     subject.zip(subject.tail).forall { case (x, y) => x < y }
[error]                                                         ^
[error] two errors found
[error] (scalacheck/test:compileIncremental) Compilation failed
[error] Total time: 18 s, completed 07/03/2018 10:14:36 AM
```

This can be reproduced outside SBT.  Using my `argsFile` [task](https://gist.github.com/retronym/e0beca2da1150a57c584f83d7da0d82a), to give this [`test.args`](https://gist.github.com/a45b22fe7f7d52bf7a856d23d16a7276)

```
> scalacheck/test:argsFile
[info] Wrote compiler comand line to: /Users/jz/code/scala/target/scalacheck/test.args
> :quit

$ ./build/quick/bin/scalac -d /tmp @/Users/jz/code/scala/target/scalacheck/test.args
/Users/jz/code/scala/test/scalacheck/treeset.scala:41: error: ambiguous implicit values:
 both method newCanBuildFrom in object SortedSet of type [A](implicit ord: Ordering[A])scala.collection.generic.CanBuildFrom[scala.collection.immutable.SortedSet.Coll,A,scala.collection.immutable.SortedSet[A]]
 and method newCanBuildFrom in class SortedSetFactory of type [A](implicit ord: Ordering[A])scala.collection.generic.CanBuildFrom[scala.collection.immutable.TreeSet.Coll,A,scala.collection.immutable.TreeSet[A]]
 match expected type scala.collection.generic.CanBuildFrom[scala.collection.immutable.TreeSet[Int],(Int, Int),That]
    subject.zip(subject.tail).forall { case (x, y) => x < y }
               ^
/Users/jz/code/scala/test/scalacheck/treeset.scala:41: error: value < is not a member of Any
    subject.zip(subject.tail).forall { case (x, y) => x < y }
                                                        ^
two errors found
```

retronym:
Minimizes to:

```
object TreeSetTest {
  def test(subject: scala.collection.immutable.TreeSet[Int]) = subject.zip(subject)
}
```

```
⚡ qscalac  sandbox/test.scala
sandbox/test.scala:2: error: ambiguous implicit values:
 both method newCanBuildFrom in object SortedSet of type [A](implicit ord: Ordering[A])scala.collection.generic.CanBuildFrom[scala.collection.immutable.SortedSet.Coll,A,scala.collection.immutable.SortedSet[A]]
 and method newCanBuildFrom in class SortedSetFactory of type [A](implicit ord: Ordering[A])scala.collection.generic.CanBuildFrom[scala.collection.immutable.TreeSet.Coll,A,scala.collection.immutable.TreeSet[A]]
 match expected type scala.collection.generic.CanBuildFrom[scala.collection.immutable.TreeSet[Int],(Int, Int),That]
  def test(subject: scala.collection.immutable.TreeSet[Int]) = subject.zip(subject)
                                                                          ^
one error found
```

I haven't minimized whatever it is that makes this sometimes pass and sometimes fail.

milessabin:
@retronym do you get that result with the minimisation only on this branch or on 2.13.x as well? I'll try myself in a mo' if you haven't already.

milessabin:
@SethTisue no, I'm running locally on linux.

milessabin:
Confirmed reproduction locally with @retronym's minimization on this branch, but not on 2.13.x.

milessabin:
I turns out that this is _exactly_ the same implicit specificity issue that resulted in `test/files/pos/t2030.scala` being demoted to a neg test ... see @smarter's discussion here: https://github.com/lampepfl/dotty/pull/1246#issuecomment-220710595. So now the puzzle is: why did this ever succeed for me and @retronym locally?

I've fixed the ScalaCheck error and reinstated t2030.scala as a pos test by adding the `newCanBuildFrom` override in `TreeSet` that @smarter suggested in the comment linked to above ... presumably this will go away when the new collections land.

milessabin:
Jenkins appears to be wedged ... is there anything I could/should do to unwedge/kill/restart the job?

adriaanm:
Rebase on https://github.com/scala/scala/pull/6377?

adriaanm:
I guess that's easier said than done. In the mean time, a rebuild may be enough (I'll hit the button)

milessabin:
Squashed and rebased.

milessabin:
I've updated the spec as requested by @retronym in his review. @smarter @odersky I've adapted the text from the Dotty PR ... can you confirm that this matches your intent? https://github.com/scala/scala/pull/6037/files#diff-5d74b019862ed901d7800b4ace79fb5b.

spec/06-expressions.md
```diff
@@ -1476,11 +1476,21 @@ question: given
...
 - Given two members of types $T$ and $U$ which are neither parameterized nor polymorphic method types,
   the member of type $T$ is as specific as the member of type $U$ if
-  the existential dual of $T$ conforms to the existential dual of $U$.
+  the existential dual of $T$ is _as specific as_ the existential dual of $U$.
   Here, the existential dual of a polymorphic type
   `[$a_1$ >: $L_1$ <: $U_1 , \ldots , a_n$ >: $L_n$ <: $U_n$]$T$` is
   `$T$ forSome { type $a_1$ >: $L_1$ <: $U_1$ $, \ldots ,$ type $a_n$ >: $L_n$ <: $U_n$}`.
   The existential dual of every other type is the type itself.
+  A type $T$ is _as specific as_ a type $U$ if,
+  - the outermost type constructor of $T$ is unparametrized, or parametrized but not contravaraint,
```
smarter: `parametrized `-> `parameterized` for consistency with the rest of the spec. Also typo: contravaraint -> contravariant.

smarter:
@milessabin Looks reasonable to me.

milessabin:
@adriaanm can we move forward with this now?

adriaanm:
We're quizzing Martin on the current status of the spec & implementation of this feature in Dotty. Once we clarify that, I'll update you here. In any case, this PR will need some work to address my review: it also needs to port the comment from dotty, and I would rather see an implementation that's closer to dotty (using an internal FlipVariance marker trait, instead of cloning)

milessabin:
Sounds good, though I don't understand what you mean by "port the comment from Dotty"?

adriaanm:
The comment is here: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Applications.scala#L1137-L1163. We ended up discovering some inconsistencies, which are being addressed here: https://github.com/lampepfl/dotty/pull/4329

dwijnand:
Conflicts with changes in `src/library/scala/collection/immutable/TreeSet.scala`, so this needs another rebase..

xeno-by:
Since we now have a process to discuss changes that narrow the gap between Scala 2 and Scala 3, I suggest we follow this process with this pull request as well. Let's have this change on the agenda of one of the next SIP meetings?

smarter:
Oops commented on the wrong PR, sorry

milessabin:
Rebased, fixed tests broken by 2.13.x deprecations.

More importantly, changed to match the current Dotty semantics which are now to flip variances throughout the type, not just at the top level.

@smarter, I'd welcome your feedback on this.

@adriaanm do you want this behind an `-Xsource` flag? In which case, what version?

milessabin:
Hey ho ... this should be OK after squashing ...

milessabin:
Squashed.

adriaanm:
My current position is that this should go under `-Xsource:3.0`. It can then be discussed in a SIP meeting on whether we want this on by default in 2.14. I still think this is a bit of a hack, and the interaction with type inference needs to be fleshed out. See e.g. https://github.com/lampepfl/dotty/issues/4465. 

It's too late for this to influence the collections, so I don't think it's urgent from a 2.13 release planning perspective.

milessabin:
If you're happy to merge it with the `-Xsource:3.0` logic then I should be able to find time for it next week.

adriaanm:
👌

milessabin:
@adriaanm done. Merge when you're happy :-)

I've taken out the spec change for now because having the language conditionalized on the language version was a bit awkward. We should deal with that as part of the SIP for the feature.

adriaanm (approved):

SethTisue:
`test/files/run/t7768.scala` is failing on Windows. at e.g. https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-windows/658/consoleFull we see:

```
 Inv[A] Inv[B] Inv[C] Inv[D] Inv[E]
-Con[A] Con[A] Con[C] Con[C] Con[E]
-Cov[E] Cov[E] Cov[E] Cov[E] Cov[E]
+java.lang.NoClassDefFoundError: Con
...
+Caused by: java.lang.ClassNotFoundException: Con
...
```

@milessabin any idea why that might be...?

adriaanm:
🤣 Con is a reserved file name on windows 😱

SethTisue:
oh LOL, it must be because `con` is a reserved name on Windows filesystems?

SethTisue:
jinx

SethTisue:
this leaves me puzzled how Shapeless is able to call something `Aux`, which is also a dirty word in Redmond.

but, a glance at the source shows that Shapeless's `Aux` is a type alias, and therefore doesn't produce a .class file. This seems in keeping with our suspicions that @milessabin doesn't even exist at runtime.

PR with fix: #7024

som-snytt:
Naming is the hardest thing after all. In this case, `miless` is `s miles`, and like the Cheshire Cat, those smiles seem to evaporate. It's very mysterious and meta.

milessabin:
Oh. My. Goodness.

djspiewak:
**Narrator:** There comes a moment in everyone's life when they must choose what kind of a person they're going to be. A moment where steel is tempered in the unyielding fires of decision. A moment where all the churning demons of ten thousand unanswered slights rage as one and demand retribution.

In that moment, our hero knew: he wasn't a Windows user.

PR #6038, opened 2017-08-14
Mention caught throwables for trait Future.failed

magro:
Caught throwables / wrapping of fatal exceptions was documented for
`trait Future.onFailure`, but not for `trait Future.failed` which became
more prominent with the deprecation of `Future.onFailure` in 2.12.


adriaanm:
Makes sense to me, but I'm not sure about the finer points of exceptions in Futures. /cc @viktorklang could perhaps provide a quick yay/nay?

viktorklang:
Sounds good to me! (Apologies for late reply)

PR #6039, opened 2017-08-16
introduce scala.collection.Searching.searchBy

dehun:
Introduce searchBy. 
In case when you have a collection ordered not by regular ordering of elements and don't want to generate fake element for search.

works like this - 
```
val coll = (0 to 5).toVector
assertEquals(Found(1), coll.searchBy("1", _.toString))
```

### DIFF
--- a/src/library/scala/collection/Searching.scala
+++ b/src/library/scala/collection/Searching.scala
@@ -54,8 +54,8 @@ object Searching {
       */
     final def search[B >: A](elem: B)(implicit ord: Ordering[B]): SearchResult =
       coll match {
-        case _: IndexedSeqLike[A, Repr] => binarySearch(elem, 0, coll.length)(ord)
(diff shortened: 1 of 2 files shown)


dehun:
Signed CLA

adriaanm:
I'm sorry you haven't received feedback on this. We need to and will do better here.

That said, I am concerned about adding another indirection in the sort implementation. @szeiger, @retronym could you provide a pointer on how to benchmark this?

szeiger:
I recommend opening a new version of this against https://github.com/scala/collection-strawman which will be used in M4. The search implementations there are the same but the methods are not extension methods anymore so we don't need to manually dispatch. Micro-benchmarks should be added to the benchmark subproject in the strawman repo. I would assume the redirection through `identity` makes the existing cases slower but maybe the JIT is smart enough to optimize it away.

PR #6040, opened 2017-08-17
Don't skip the ValTypeCompleter from AccessorTypeCompleter

lrytz:
Don't skip the ValTypeCompleter in the AccessorTypeCompleter
When the AccessorTypeCompleter gets the type signature of the
ValDef, ensure that the ValTypeCompleter runs (which sets the
annotations on the field symbol) instead of just calling typeSig.

Fixes scala/bug#10471


adriaanm (approved):
LVGTM

PR #6043, opened 2017-08-20
Add Scaladoc tables

janekdb:
Parse a subset of GitHub flavoured markdown to define simple tables which can include inline wiki markdown. The specification for the table markdown is based on [GitHub Flavored Markdown Spec, Version 0.28-gfm (2017-08-01)](https://github.github.com/gfm/#tables-extension-).

The markdown used to structure the table follows [GitHub Flavoured Markdown Table Extension](https://github.github.com/gfm/#tables-extension-) with the restriction that pipe symbols at the start and end of row are mandatory and the clarification that delimiter row cells must have a minimum of three hyphens.

Note: Most of the following is irrelevant now the markdown is GHFM based..

```
  |     Define table cells
  ---   Default text alignment
  :---  Left text alignment
  ---:  Right text alignment
  :---: Center text alignment
```

Example.

A simple 2x2 table with a caption and header row,

```
  {|
  |+ Caption: Example Table
  ! header 1        ! header 2        !
  | row 1, column 1 | row 1, column 2 |
  | row 2, column 1 | row 2, column 2 |
  |}
```

Header and cell markdown can be repeated at the start of the heaer or row to allow | and ! to be included in content,

```
  {|
  !!  Disjunction!    !!  Conjunction!     !!
  ||| Logical OR = || ||| Logical AND = && |||
  |}
```

Fuller example including longer cell marks,

```
  {|
  |+ Caption: Example Table
  ! Header cell 1 ! Header cell 2 !
  | Row 1, column 1 | Row 1, column 2 |
  || Content with pipe symbol:  |A| := det(A) || final cell over
  two lines||
  |}
```

Caption, header and cell content can be broken over multiple lines and
include paragraphs and inline markdown,

```
    |Multiline
    cell
    content with __inline__ markdown

    Paragraph two.
```

Block level elements other than paragraphs are not yet parsed. Specifically these types are not yet supported,

 - Code blocks
 - Horizontal rules
 - Titles
 - Lists

A warning is issued when the rows and optional header do not all have the same number of cells.
This reserves an option to do something "wiki-ish" when cell counts differ.

Inspired by MediaWiki markdown with new row conciseness following GitHub flavoured markdown.

GitHub Project with subtasks: https://github.com/janekdb/scala/projects/1

## Live Demo

http://janekdb.github.io/scala/PR-6043/scala/test/scaladoc/tables/code/index.html

This page contains the unit test tables with the table markdown duplicated into a code block,

image

I'm hoping this makes it easier to review the PR and also furnishes a good idea of what you get for what you give.

## PR Status

WIP

## Review Comments

 - [x] More compact way of introducing new rows

## TODO: Move to GHFM

 - [x] Make GHFM compliant
 - [x] Add GHFM column alignment options
 - [x] Update commit message
 - [x] Document requirement for delimiting pipes on rows per @lrytz comment
 - [x] Add negative test for missing pipes

## TODO: Support multiline content cell with GHFM extension

 - [ ] Understand dotty 3rd party markdown parser
 - [ ] Explore conversion of pre-dotty markdown to allow dotty markdown parser to be used
 - [ ] Retain multiline support subject to review of dotty approach

## TODO

 - [x] Include table start in check for new paragraph
 - [x] Fix lines numbers in `table-warnings.check`
 - [x] Add progress check in `parseCells0`
 - [x] Parse `Caption` as `Seq[Block]` matching `Cell`.
 - [x] Fix missing paragraphs in cell content starting with `CellsParagraphsB` and friends 
 - [x] Remove new row markdown `|-` because final cell termination is sufficient
 - [x] Consolidate or remove identically structured test traits
 - [x] CSS style table to blend into current Scaladoc themes not including row alternation styles
 - [x] Consider review comment regarding paragraph element remembering CSS is the correct way to address this
 - [x] Decide which other block elements other than Paragraph to parse: Code, Lists, Title, HRule.
 - [x] Review commit message
 - [x] Update grammar on table() method comment
 - [x] Remove `isNewlineAnyTableMarks` or the duplicate
 - [x] Run static analysis
 - [x] Squash commits

## WONT

Item | Reason | Comment
--- | --- | ---
Be more lenient toward leading whitespace below table marks|Deliver current tables sooner|Possible future improvement|
Support code blocks in cells in first merge|Deliver current tables sooner|Parsing code blocks will be the next improvement. I'd rather separate this off into new work because parsing code block while avoid code duplication could be quite a significant transformation requiring some recursion to be added at the top level of the WikiParser.|
Drop requirement for table start and end markers|Deliver current tables sooner|Without the start and end markers detecting transitions into and out of table parsing mode would be more complex and invasive. If ever we want some table level options around table formatting etc there would be no where to locate this information. Main reason is the implementation complexity.

## Acceptance Testing

 - [x] Check HTML rendering of test traits
 - [x] No regressions on existing non-table markdown in Scala library

## Review

 - [x] Create version of tables.scala which includes the table markdown as a code block and share it
 
## Target Outcome Over the Course of a few PRs :)

 - [ ] Tables as seen in ScalaTest docs can be created without requiring non-markdown work

## Follow Ons

- Support code blocks.
- Support other block elements: Horizontal Rules, Titles, List
- Make table css responsive
- Make styling in CaptionsB, CaptionsMultilineB effective instead of code style being bold and indistinguisable for other text
- Consider applying a minimum width for table to avoid caption wrapping as seen in CaptionsC
- Make styling of "and" in CellsC style the word differently to surrounding text.
- Left/right/centre cell alignment

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
@@ -1,5 +1,5 @@
 /* NSC -- new Scala compiler
- * Copyright 2007-2017 LAMP/EPFL
+ * Copyright 2007-2018 LAMP/EPFL
  * @author  Manohar Jonnalagedda
  */
 
@@ -8,10 +8,12 @@ package doc
 package base
 
 import base.comment._
+import scala.annotation.tailrec
 import scala.collection._
 import scala.util.matching.Regex
 import scala.reflect.internal.util.Position
 import scala.language.postfixOps
+
 
 /** The comment parser transforms raw comment strings into `Comment` objects.
   * Call `parse` to run the parser. Note that the parser is stateless and
@@ -433,6 +435,9 @@ trait CommentFactoryBase { this: MemberLookupBase =>
   protected final class WikiParser(val buffer: String, pos: Position, site: Symbol) extends CharReader(buffer) { wiki =>
     var summaryParsed = false
 
+    // TODO: Convert to Char
+    private val TableCellStart = "|"
+
     def document(): Body = {
       val blocks = new mutable.ListBuffer[Block]
       while (char != endOfText)
@@ -442,7 +447,7 @@ trait CommentFactoryBase { this: MemberLookupBase =>
 
     /* BLOCKS */
 
-    /** {{{ block ::= code | title | hrule | listBlock | para }}} */
+    /** {{{ block ::= code | title | hrule | listBlock | table | para  }}} */
     def block(): Block = {
       if (checkSkipInitWhitespace("{{{"))
         code()
@@ -452,6 +457,8 @@ trait CommentFactoryBase { this: MemberLookupBase =>
         hrule()
       else if (checkList)
         listBlock
+      else if (check(TableCellStart))
+        table()
       else {
         para()
       }
@@ -490,7 +497,7 @@ trait CommentFactoryBase { this: MemberLookupBase =>
           jumpWhitespace()
           jump(style)
           val p = Paragraph(inline(isInlineEnd = false))
-          blockEnded("end of list line ")
+          blockEnded("end of list line")
           Some(p)
         }
 
@@ -542,6 +549,284 @@ trait CommentFactoryBase { this: MemberLookupBase =>
       repeatJump('-')
       blockEnded("horizontal rule")
       HorizontalRule()
+    }
+
+    /** {{{
+      * table         ::= headerRow '\n' delimiterRow '\n' dataRows '\n'
+      * content       ::= inline-content
+      * row           ::= '|' { content '|' }+
+      * headerRow     ::= row
+      * dataRows      ::= row*
+      * align         ::= ':' '-'+ | '-'+ | '-'+ ':' | ':' '-'+ ':'
+      * delimiterRow :: = '|' { align '|' }+
+      * }}}
+      */
+    def table(): Block = {
+
+      /* Helpers */
+
+      def peek(tag: String): Unit = {
+        val peek: String = buffer.substring(offset)
+        val limit = 60
+        val limitedPeek = peek.substring(0, limit min peek.length)
+        println(s"peek: $tag: '$limitedPeek'")
+      }
+
+      def nextIsCellStart = check(TableCellStart)
+
+      /* Accumulated state */
+
+      var header: Option[Row] = None
+
+      val rows = mutable.ListBuffer.empty[Row]
+
+      val cells = mutable.ListBuffer.empty[Cell]
+
+      def finalizeCells(): Unit = {
+        if (cells.nonEmpty) {
+          rows += Row(cells.toList)
+        }
+        cells.clear()
+      }
+
+      def finalizeHeaderCells(): Unit = {
+        if (cells.nonEmpty) {
+          if (header.isDefined) {
+            reportError(pos, "more than one table header")
+          } else {
+            header = Some(Row(cells.toList))
+          }
+        }
+        cells.clear()
+      }
+
+      def checkAny(terminators: List[String]) = terminators.exists(check)
+
+      def isEndOfText = char == endOfText
+
+      def isNewline = char == endOfLine
+
+      def skipNewline() = jump(endOfLine)
+
+      def contentNonEmpty(content: Inline) = content != Text("")
+
+      /**
+        * @param nextIsStartMark True if the next char is a cell mark prefix and not any non-cell mark.
+        * @param cellStartMark   The char the cell start mark is based on
+        * @param finalizeRow     Function to invoke when the row has been fully parsed
+        */
+      def parseCells(nextIsStartMark: => Boolean, cellStartMark: Char, finalizeRow: () => Unit): Unit = {
+        /* The first sequence of cellStartMark characters defines the markdown for new cells. */
+        def parseStartMark() = {
+          if (!jump(cellStartMark)) {
+            peek("Expected startMark")
+            sys.error("Precondition violated: Expected startMark.")
+          }
+          cellStartMark.toString
+        }
+
+        /* startMark is the only mark not requiring a newline first */
+        def makeInlineTerminators(startMark: String) = startMark :: Nil
+
+        val startPos = offset
+
+        val startMark = parseStartMark()
+
+        val inlineTerminators = makeInlineTerminators(startMark)
+
+        val content = Paragraph(inline(isInlineEnd = checkAny(inlineTerminators)))
+
+        parseCells0(content :: Nil, startMark, cellStartMark, inlineTerminators, nextIsStartMark, finalizeRow, startPos, offset)
+      }
+
+      // Continue parsing a table row.
+      //
+      // After reading inline content the follow conditions will be encountered,
+      //
+      //    Case : Next Chars
+      //    ..................
+      //      1  : end-of-text
+      //      2  : '|' '\n'
+      //      3  : '|'
+      //      4  : '\n'
+      //
+      // Case 1.
+      // State : End of text
+      // Action: Store the current contents, close the row, report warning, stop parsing.
+      //
+      // Case 2.
+      // State : The cell separator followed by a newline
+      // Action: Store the current contents, skip the cell separator and newline, close the row, stop parsing.
+      //
+      // Case 3.
+      // State : The cell separator not followed by a newline
+      // Action: Store the current contents, skip the cell separator, continue parsing the row.
+      //
+      // Case 4.
+      // State : A newline followed by anything
+      // Action: Store the current contents, report warning, skip the newline, close the row, stop parsing.
+      //
+      @tailrec def parseCells0(
+                                contents: List[Block],
+                                startMark: String,
+                                cellStartMark: Char,
+                                inlineTerminators: List[String],
+                                nextIsStartMark: => Boolean,
+                                finalizeRow: () => Unit,
+                                progressPreParse: Int,
+                                progressPostParse: Int
+                              ): Unit = {
+
+        def isStartMarkNewline = check(startMark + endOfLine)
+
+        def skipStartMarkNewline() = jump(startMark + endOfLine)
+
+        def isStartMark = check(startMark)
+
+        def skipStartMark() = jump(startMark)
+
+        def isNewlineCellStart = check(endOfLine.toString + cellStartMark)
+
+        def storeContents() = cells += Cell(contents.reverse)
+
+        val startPos = offset
+
+        // The ordering of the checks ensures the state checks are correct.
+        if (progressPreParse == progressPostParse) {
+          peek("no-progress-table-row-parsing")
+          sys.error("No progress while parsing table row")
+        } else if (isEndOfText) {
+          // peek("1: end-of-text")
+          // Case 1
+          storeContents()
+          finalizeRow()
+          reportError(pos, "unclosed table row")
+        } else if (isStartMarkNewline) {
+          // peek("2/1: start-mark-new-line")
+          // Case 2
+          storeContents()
+          finalizeRow()
+          skipStartMarkNewline()
+          // peek("2/2: start-mark-new-line")
+        } else if (isStartMark) {
+          // peek("3: start-mark")
+          // Case 3
+          storeContents()
+          skipStartMark()
+          val content = inline(isInlineEnd = checkAny(inlineTerminators))
+          // TrailingCellsEmpty produces empty content
+          val accContents = if (contentNonEmpty(content)) Paragraph(content) :: Nil else Nil
+          parseCells0(accContents, startMark, cellStartMark, inlineTerminators, nextIsStartMark, finalizeRow, startPos, offset)
+        } else if (isNewline) {
+          // peek("4: newline")
+          // Case 4
+          /* Fix and continue as there is no option to not return a table at present. */
+          reportError(pos, "missing trailing cell marker")
+          storeContents()
+          finalizeRow()
+          skipNewline()
+        } else {
+          // Case π√ⅈ
+          // When the impossible happens leave some clues.
+          reportError(pos, "unexpected table row markdown")
+          peek("parseCell0")
+          storeContents()
+          finalizeRow()
+        }
+      }
+
+      /* Parsing */
+
+      jumpWhitespace()
+
+      parseCells(nextIsCellStart, TableCellStart(0), finalizeHeaderCells)
+
+      while (nextIsCellStart) {
+        val initialOffset = offset
+
+        parseCells(nextIsCellStart, TableCellStart(0), finalizeCells)
+
+        /* Progress should always be made */
+        if (offset == initialOffset) {
+          peek("no-progress-table-parsing")
+          sys.error("No progress while parsing table")
+        }
+      }
+
+      /* Finalize */
+
+      /* Structural consistency checks */
+
+      /* Structural coercion */
+
+      // https://github.github.com/gfm/#tables-extension-
+      // TODO: The header row must match the delimiter row in the number of cells. If not, a table will not be recognized:
+      // TODO: Break at following block level element: The table is broken at the first empty line, or beginning of another block-level structure:
+      // TODO: Do not return a table when: The header row must match the delimiter row in the number of cells. If not, a table will not be recognized
+
+      if (cells.nonEmpty) {
+        reportError(pos, s"Parsed and unused content: $cells")
+      }
+      assert(header.isDefined, "table header was not parsed")
+      val enforcedCellCount = header.get.cells.size
+
+      def applyColumnCountConstraint(row: Row, defaultCell: Cell, rowType: String): Row = {
+        if (row.cells.size == enforcedCellCount)
+          row
+        else if (row.cells.size > enforcedCellCount) {
+          val excess = row.cells.size - enforcedCellCount
+          reportError(pos, s"Dropping $excess excess table $rowType cells from row.")
+          Row(row.cells.take(enforcedCellCount))
+        } else {
+          val missing = enforcedCellCount - row.cells.size
+          Row(row.cells ++ List.fill(missing)(defaultCell))
+        }
+      }
+
+      // TODO: Abandon table parsing when the delimiter is missing instead of fixing and continuing.
+      val delimiterRow :: dataRows = if (rows.nonEmpty)
+        rows.toList
+      else {
+        reportError(pos, "Fixing missing delimiter row")
+        Row(Cell(Paragraph(Text("-")) :: Nil) :: Nil) :: Nil
+      }
+
+      if (delimiterRow.cells.isEmpty) sys.error("TODO: Handle table with empty delimiter row")
+
+      val constrainedDelimiterRow = applyColumnCountConstraint(delimiterRow, delimiterRow.cells(0), "delimiter")
+
+      val constrainedDataRows = dataRows.toList.map(applyColumnCountConstraint(_, Cell(Nil), "data"))
+
+      /* Convert the row following the header row to column options */
+
+      val leftAlignmentPattern = "^:?-++$".r
+      val centerAlignmentPattern = "^:-++:$".r
+      val rightAlignmentPattern = "^-++:$".r
+
+      import ColumnOption._
+      /* Encourage user to fix by defaulting to least ignorable fix. */
+      val defaultColumnOption = ColumnOptionRight
+      val columnOptions = constrainedDelimiterRow.cells.map {
+        alignmentSpecifier =>
+          alignmentSpecifier.blocks match {
+            // TODO: Parse the second row without parsing inline markdown
+            // TODO: Save pos when delimiter row is parsed and use here in reported errors
+            case Paragraph(Text(as)) :: Nil =>
+              as.trim match {
+                case leftAlignmentPattern(_*) => ColumnOptionLeft
+                case centerAlignmentPattern(_*) => ColumnOptionCenter
+                case rightAlignmentPattern(_*) => ColumnOptionRight
+                case x =>
+                  reportError(pos, s"Fixing invalid column alignment: $x")
+                  defaultColumnOption
+              }
+            case x =>
+              reportError(pos, s"Fixing invalid column alignment: $x")
+              defaultColumnOption
+          }
+      }
+      blockEnded("table")
+      Table(header.get, columnOptions, constrainedDataRows)
     }
 
     /** {{{ para ::= inline '\n' }}} */
@@ -781,6 +1066,7 @@ trait CommentFactoryBase { this: MemberLookupBase =>
           checkSkipInitWhitespace('=') ||
           checkSkipInitWhitespace("{{{") ||
           checkList ||
+          check(TableCellStart) ||
           checkSkipInitWhitespace('\u003D')
         }
         offset = poff
--- a/src/scaladoc/scala/tools/nsc/doc/base/comment/Body.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/base/comment/Body.scala
@@ -1,5 +1,5 @@
 /* NSC -- new Scala compiler
- * Copyright 2007-2013 LAMP/EPFL
+ * Copyright 2007-2018 LAMP/EPFL
  * @author  Manohar Jonnalagedda
  */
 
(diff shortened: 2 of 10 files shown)


janekdb:
This is an example of table markdown which includes all types of supported marks,

```
/** A variety of decorators that enable converting between
 *  Scala and Java collections using extension methods, `asScala` and `asJava`.
 *
 *  The extension methods return adapters for the corresponding API.
 *
 * {|
 * |+ Conversions are supported via `asScala` and `asJava`
 * ! Scala                           ! Direction ! Java !
 * | scala.collection.Iterable       | <=>       | java.lang.Iterable |
 * | scala.collection.Iterator       | <=>       | java.util.Iterator |
 * | scala.collection.mutable.Buffer | <=>       | java.util.List |
 * | scala.collection.mutable.Set    | <=>       | java.util.Set |
 * | scala.collection.mutable.Map    | <=>       | java.util.Map |
 * | scala.collection.concurrent.Map | <=>       | java.util.concurrent.ConcurrentMap |
 * |}
 *
```
image

The unit test contains additional examples which include examples of multiline caption, header and cell content using inline wiki styling. Tables cannot contain other block elements.

At this point I've dealt with all the defects I could anticipate with the exception of one described below in *Known Issues*. I'l like to get this out to a wider audience now so the next set of defects can be found.

## Known Issues

Although it could be charitably regarded as adherent to the "spec" the markdown parsing does not tolerate leading whitespace before marks. To avoid wasting user time a subsequent PR will allow this to parse.

```
  // Should parse to table with a caption, header and a row
  /**
    * {|
    *  |+Leading
    *   whitespace before marks
    *    !!Not Yet Skipped!!TODO
    *    |-
    *    |A
    *       |B
    * |}
    */
  trait LeadingWhitespaceNotYetSkipped
```
Supporting this may require an extension to the `check*` method which so far I have managed to avoid. It would also require a few more unit tests and I feel since this limitation is well understood I'd rather get this in use and get some unanticipated cases reported.

Backlogged as a fast follower: https://github.com/scala/bug/issues/10475

## CSS Styling

The styling is a call to action for anyone with a passion for css. I will submit a PR if there is no interest from other quarters.

Follow on actions include

 - Converting some library Scaladoc to use tables where it makes an improvement
 - Updating the Scaladoc guide
 - Writing a blog post

adriaanm:
Cool! I'd prefer using GitHub-flavored markdown, as we do everywhere else. What do you think?

From https://guides.github.com/features/mastering-markdown/
> You can create tables by assembling a list of words and dividing them with hyphens - (for the first row), and then separating each column with a pipe |:

First Header | Second Header
------------ | -------------
Content from cell 1 | Content from cell 2
Content in the first column | Content in the second column

janekdb:
I surveyed a few markdown systems here: https://github.com/janekdb/scala/wiki/Table-Markdown-Survey

I believe the MediaWiki style markdown is easier to work with having used both GH and MW. It requires less attention to syntax than GH flavoured markdown and has the benefit of supporting vertical layout of cells in addition to horizontal layout which will work well with larger runs of inline markdown. MW can do everything GH can do AFAIK with less fiddling so I think it should be preferred.

Putting the choice into context here are the requirements I evaluated the markdown options against: https://github.com/janekdb/scala/wiki/Scaladoc-Tables-Requirements

adriaanm:
My opinion is that there's a strong case for standardization here, with feature completeness being a bit less urgent of a concern. Using GitHub markdown also makes it easier to preview

lrytz:
In general I agree to bias towards GFM, but here maybe the limitations are too severe?
  - GFM requires a header row
  - Cell content has to be on a single line
  - I like the option for a caption

Would it be OK adapt GFM to our needs?

On the other hand, the current proposal doesn't have functionality for left/center/right alignment.

I played a bit with the current state. One thing that's annoying for small tables is that each row requires a `|-` and a `|`. Example:

```
 * {|
 * |- | One | `1`
 * |- | Four | `4`
 * |}
```
It actually works without the first `|-`. I think it would be good if the `|` after `|-` was optional.

A detail in presentation: if two tables follow each other, the are sticked together. Maybe every table should be wrapped in a `<p>`?

The caption and header colors cold be a little less flashy :-)

I didn't get code blocks (`{{{ ... }}}`) to work within table cells - is that expected?

janekdb:
@lrytz —

 * Cell content is inline markdown only at this point — no block level elements.

 * The styling is deferred to a later commit.

 * Left/right/centre alignment and other styling options I was intending to address later after a first iteration that results in usable tables.

I have some views to share regarding GHFM or not, just need a bit of time to write them down.

Thanks for the feedback. I am surprised the example was parsed but getting some feedback was the purpose of the PR!

lrytz:
OK, thanks for the clarifications!

I definitely think that allowing to use multiple lines for a cell's content (or even a row's) is very important. Otherwise the usefulness of this new feature would be severely limited.

I guess a good source of real-world examples is scalatest's scaladoc, searching for `<table>` in its checkout I find 42 results in 36 files. Ideally all of these should be doable in the new syntax.

Example: http://doc.scalatest.org/3.0.0/index.html#org.scalatest.AsyncFunSuite

janekdb:
@lrytz — The identification of these pre-existing table is super useful. This will allow for some concrete feature validation. I will take a look over the next few days.

adriaanm:
@janekdb, which version would you like to schedule this for? For 2.12.4, the PR must be mergeable by Wed Sep 27

janekdb:
@adriaanm — 2.12.5 as it needs rethinking to accommodate,

* More compact way of introducing new rows
* Tables as seen in ScalaTest docs

janekdb:
Not forgotten. Making gradual progress on more compact row markdown.

Example of a three row table not using the `|+` new row markdown,
```
  /**
    * {|
    * |Orange|Apple|
    * |Bread|Pie|
    * |Butter|Ice cream|
    * |}
    */
```

84/84 tests enabled and passing with new more-GHFM compact row markdown using cleaner parsing approach.

☺️

janekdb:
Still working on a regression defect. Altering the way _closed_ rows are detected to be more straightforward.

janekdb:
Regressions fixed. Clean up and then some end-to-end testing next. ☕️ 🍰

janekdb:
scaladoc-table-java-convertors

GitHub flavoured table row markdown is working.

Dreadful colour scheme is temporary!

Some more TODOs listed in the this issue's description before I ask for reviews but always happy to get some feedback now.

lrytz:
Happy to see you getting back to this!

SethTisue:
can't wait to see what @bvenners does with this

janekdb:
Note to self: Proof paragraphs are parsed in cells:

image

The non-rendering of "and" as underline is a pre-existing defect not related to tables support.

janekdb:
I've decided to defer support for code blocks to a later PR.

The introduction of tables is mainly additive in the sense that only two existing methods in `CommentFactoryBase.scala` have been touched and then only to extend the cases for `table`. Everything else is achieved within the `table()` method.

Supporting code blocks and the other block elements within cells will need some reworking of `WikiParser` to avoid duplicating code inside the table parsing method. This might be accomplished by recursive parsing. Other block elements are horizontal rules, titles and lists. This work is very different to what has been done in adding tables so I want to cap the PR scope at this point.

With code block support as an immediate follower we won't be violating the principle of least surprise for too long.

Thoughts?

src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
```diff
@@ -544,6 +562,376 @@ trait CommentFactoryBase { this: MemberLookupBase =>
...
+
+      def checkAny(terminators: List[String]) = terminators.exists(check)
+
+      def endOfInlineWhenTableMarkOnNextLine = checkAny(TableAllMarksOnNextLine)
+
+      def isEndOfText = char == endOfText
+
+      def isNewline = char == endOfLine
+
+      def isNewlineAnyTableMarks = checkAny(TableAllMarksOnNextLine)
```
janekdb: `isNewlineAnyTableMarks` is identical to `endOfInlineWhenTableMarkOnNextLine`. There can only be one `table-mark-on-next-line` incumbent!!!!

janekdb:
Tables are now spaced with a top and bottom `1em` margin.

image

janekdb:
Table CSS has been subjected to a severe beating resulting in a new found and gladly embraced aversion to harmful individualistic tendencies.

image

janekdb:
This is ready for review. Please check the **Live Demo** section in the PR description for a useful link showing table markdown alongside the tables generated.

There are a number of very reasonable extensions to tables which I've listed in the description. I would prefer to follow these up in later PRs.

janekdb:
The regression test passed.
```
$ diff -r build-scaladoc-tables/scaladoc-quick/  build-scaladoc-2.12.x/scaladoc-quick/
diff -r build-scaladoc-tables/scaladoc-quick/lib/template.css build-scaladoc-2.12.x/scaladoc-quick/lib/template.css
822,847d821
< /* Author Content Table formatting */
< 
< .doctbl {
<   border-collapse: collapse;
<   margin: 1.0em 0em;
< }
< 
< table.doctbl caption {
<   text-align: center;
<   font-weight: bold;
<   font-size: 1.1em;
< }
< 
< table.doctbl th {
<   border: 1px dotted #364550;
<   background-color: #c2d2dc;
<   padding: 5px;
<   color: #103a51;
<   font-weight: bold;
< }
< 
< table.doctbl td {
<   border: 1px dotted #364550;
<   padding: 5px;
< }
< 
```
I built the Scaladoc for `library` using the `quick` binary from the `2.12.x` branch and this branch. The only differences was the expected new lines in `template.css`.

This PR is open for review!

lrytz:
@janekdb I'll take a look, but probably not this week..

janekdb:
Thanks @lrytz. I'll take your promissory note and bank it! Given this PR has been brewing for beyond the average gestation period of an [American Bison](https://en.wikipedia.org/wiki/List_of_mammalian_gestation_durations) next week or even next month is more than soon enough.

src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
```diff
@@ -544,6 +562,370 @@ trait CommentFactoryBase { this: MemberLookupBase =>
...
+        } else if (isEndOfText) {
+          // Case 1
+          storeContents()
+          reportError(pos, "unclosed table")
+        } else if (isNewlineAnyTableMarks) {
+          // Case 2
+          storeContents()
+          skipNewline()
+        }
+        else if (isNewline) {
```
janekdb: Fix new line

janekdb:
We are open 24-7 for all your reviewing comments needs!

lrytz:
Thanks @janekdb - I haven't forgotten, but got really busy with helping to integrate the 2.13 collections...

janekdb:
Build failure was: `The job exceeded the maximum time limit for jobs, and has been terminated`

lrytz:
I personally like this change and think it's good to go. Amazing work also on the testing front!

But I won't just decide on my own, because it has a long-term impact.

I guess the main concern people could have is the syntax. It is inspired by GitHub Markdown Tables (`| cell | cell | cell |`), but there are differences:
  - Cell content can be on multiple lines (GFM requires a single line per table row)
  - Support for captions
  - Table headers are different
  - `{|` / `|}` tokens for table start / end
  - plus some more

I asked on dotty's gitter (https://gitter.im/lampepfl/dotty?at=5b0db842b8abab25d37ea656). `dottydoc` supports scaladoc wiki, but also markdown. In fact it also supports some extensions by the flexmark-java implementation (https://github.com/vsch/flexmark-java/wiki/Tables-Extension). @smarter and @olafurpg argue that supporting plain markdown and deferring to html tables for more complex cases would be a better trade-off.

The drawback of falling back to html is that wiki syntax is gone entirely, so you end up writing a lot of html (`<a>`, `<li>`, `<code>`) within tables, which I dislike.

cc @adriaanm, @SethTisue.

olafurpg:
Markdown + HTML have the benefit of being easier to reproduce in external tools like IntelliJ. I think custom Scaladoc markup syntax should ideally be limited to functionality that is not possible to encode in standard markup languages.

adriaanm:
So, to summarize the gitter discussion: the preference is to use Github flavored markdown (which means: drop support multi-line tables from this PR). 

I'm a big +1 on dropping any non-standard wiki syntax and standardizing on github's flavor of markdown. In the mean time, this PR has been in the making for a long time, so I propose we merge as-is, and start the migration to pure GFM as a separate work item. Plenty of other changes will be required.

lrytz:
I wouldn't release someting in 2.12.7 and change it again in 2.12.8 though. If the end goal is GFM only, then we should not release additional features (optional header row, captions, multi-line cells) just to remove them again.

adriaanm:
True, my thinking was that the expected total pain is low (== usage of
multi-line table extension * pain to convert to html) compared to the
upside. Ideally, the multiline extension should be disabled in this PR,
assuming that’s the only delta compared to GFM

janekdb:
Thanks for feedback. @lrytz It was a nice chance to do some pure TDD!

Based on the observations this is the plan,

- Reduce differences to GHFM starting with dropping table open/close and reviewing header definitions
- Retain support for multiline content and make the case for this variation

lrytz:
Let's all agree on what to do before starting more work.

If I understand correctly, @adriaanm would really like GHFM only, which means
  - no captions
  - no multi-line contents
  - mandatory header row
  - (slightly) more verbose syntax than the current PR
  - no `||` cell separators, `\|` escape for pipe character inside a cell
  - add support for GHFM's column alignments (could come in a later PR)

janekdb:
Good to start the process of converging on agreed goals before restarting this.

Multiline content is important. It paves the way for allowing code blocks in cell content. With this in mind I assign limited importance to including the main deficiency of GHFM which is lack of support for multiline content other than via embedding `<br>`s. This can result in long lines that become inconvenient to edit which is contrary to the spirit of markdown. Some organisations use coding guidelines that impose a limit on line length.

Bailing out to HTML table tags when a cell needs to now have two paragraphs instead of one is fairly distracting and wasteful from an author's POV. With adequate markdown we can support the user in following the guidance to avoid HTML given on https://docs.scala-lang.org/style/scaladoc.html.

Let's not forget that GHFM is _currently_ a popular markdown variation but that status is no doubt a passing thing. My preference is to judge the solution by the intrinsic value. In this case support for multiline content outweighs the surprise factor.

Finally let's stand back: GHFM may already be an outgoing de facto standard. We could all be using [GitLab Flavored Markdown](https://docs.gitlab.com/ee/user/markdown.html) in the not too distant future.

lrytz:
I agree with, but it seems not everyone does. We need to decide somehow. Since this is scala/scala, @adriaanm could take the decision if he wants, otherwise we can take it to a SIP meeting.

adriaanm:
I don't think we need to SIP this. I would like compliance with GHFM as the default policy, but we can always discuss case by case where it makes sense to deviate. Supporting multiline content for tables is fine with me.

lrytz:
Cool! I agree to stick to GHFM and have extensions where we really think it adds significant and necessary value.

An issue is that dotty currently uses an off-the-shelve markdown parser, which wouldn't work anymore.

adriaanm:
Could we use some kind of pre-processing step for our extensions, and use the same markdown parser as dotty (or all agree on using one that perhaps supports the desired extensions?)

janekdb:
@adriaanm I'm happy to explore that before making changes here.

janekdb:
Would it be acceptable to implement a subset of the GHFM table syntax rules?

Namely not implementing the optional requirement on pipes at the start or end of a table line.

The reason for asking is if we require pipes on either side of every cell the parsing code would be somewhat different to the initial more expressive version that allows captions and multiline cells.

If optional leading and trailing pipes are supported then the parsing will need to change from the current modal greedy approach where the lookahead window is a few consecutive characters. The lookahead would need to be non-local to check for the existence a full table structure. This can be done efficiently by creating a suitable view of the raw markdown once at the beginning of the parse. I guess I've just about answered my own question in the stating of it!

Update: @lrytz +1'd requiring the surrounding pipes. @adriaanm - okay with this?

Here is a minimal GHFM table according to the [spec](https://help.github.com/articles/organizing-information-with-tables/),

```
Command
---
git status
git diff
```

Comments?

## Aside

Interestingly this valid table definition does not render correctly as shown below,

_Valid GHFM that does not render correctly_

image

However invalid GHFM markdown does render. The requirement for the header line to have at least three hyphens is skipped yet it renders,
```
Command
-|
git status
git diff
```

_Invalid GHFM that does render correctly_

image

It won't be practical to match the vagaries of the Github implementation of their spec so I would need to proceed by following the spec alone.

lrytz:
+1 on following the spec / ignoring the github implementation bugs :) for the optional pipes, you ask if it would be ok to require pipes in the beginning and end in scaladoc? i'd say that's fine.

janekdb:
The conversion to GHFM is progressing surprisingly straightforwardly now that we are working with the subset of all possible GHFM tables that have pipes at the start and end of each row. While I am sad face to see captions being removed we are gaining column alignment options.

When invalid markdown is parsed the plan is to

- Issue warnings
- Include the invalid markdown in the rendered page in a preformatted block

This ensures the content will be seen even when incorrectly marked down and prevents the markdown error from being ignored as easily as just warning.

So far this happy path case is now being parsed,
```
  /**
    * |First Header|Second Header|Third Header|
    * |:---|:---:|---:|
    * |Cell 1|Cell 2|Cell 3|
    */
  trait GHFM_ColumnOptionsAllTypes
```

adriaanm:
@janekdb The 2.12.7 window is closing next week (we may extend for max one week as needed). It sounds like this PR could make that release. What do you think?

janekdb:
@adriaanm Happy paths are going fine so Sept 14 is not impossible.

The main outstanding areas are,

1. Handle escaping of | with \\|
2. Ensure all multiline line support is removed
3. Tidy up

I'd have to defer work on negative cases and regard those as bug fixes. Would it be acceptable to also defer work on escaping | and get that in a bugfix?

adriaanm:
Sure, both sound fine as follow-ups! Since @lrytz has been more closely involved, I'll see what he thinks too. For the next few days, I suggest the priority would be to make sure there are no regressions on existing scaladoc out there. Should we run this through the community build?

lrytz:
Fixing these things in follow-ups sounds good to me too, sure!

janekdb:
Okay, good. I will concentrate on 2 & 3

- Ensure all multiline line support is removed
- Tidy up including squashing commits

Happy paths are done! Fixing up the tests for warnings and then the rendering code beckons.

lrytz:
@janekdb 
```
[test] [error]   - reflect/compile:doc failed: scala.NotImplementedError: an implementation is missing
[test] [error]   - compiler/compile:doc failed: scala.NotImplementedError: an implementation is missing
```

and

```
[test] !! 66 - run/tables-warnings.scala                 [compilation failed]
```

janekdb:
All positive and negative tests passing locally,

The parsing outcomes known to be incorrect are captured as-is in these test cases,
```
  trait CellMarkerEscaped
  trait CellMarkerEscapedTwice
  trait MissingInitialCellMark
  trait SplitCellContent
  trait MixedContentUnspaced
  trait LeadingWhitespaceNotSkipped
```
I'm intending to address these as fast follower bug fixes after this PR is merged.

Outstanding work,
- Remove dead code from `CommentFactoryBase`
- Update `HtmlPage.tableToHtml`

I'll need to do justice to those points tonight.

Update: The CI build found two cases of incompatibility with the Scaladoc in this project,
### TreeInfo.scala
```
[test] peek: 6: newline: '
[test]           | [id ’.’] ‘super’ [‘[’ id ‘]’] ‘.’ id
[test] '
```
### MarkupParsers.scala
```
[test] peek: 6: newline: '
[test]                      | `{` scalablock `}`
[test] '
```
Tables should only be recognized when the initial pipe is blocked to the left. There might be some whitespace trimming occurring that is causing this. For now I've removed the line breaks which I realise simply hides the problem.

Update: TreeInfo and MarkupParsers have been restored and initial ws trimming removed,
```
    def block(): Block = {
      if (checkSkipInitWhitespace("{{{"))
        code()
      else if (checkSkipInitWhitespace('='))
        title()
      else if (checkSkipInitWhitespace("----"))
        hrule()
      else if (checkList)
        listBlock
      else if (check(TableCellStart)) <- was checkSkipInitWhitespace
        table()
      else {
        para()
      }
    }
```
This implementation of tables assumed blocking to the left, which reduces the chances of parsing non-table markdown. Leading whitespace being allowed on the first row was inconsistent with that. Allowing leading whitespaces on any row appears to be allowed by the GHFM table spec but will cause incompatibilities as we've seen.

I will take this one as bug fix b/c it will require some thought and testing. One approach is to lookahead and require the line following the initial table row to be a delimiter row. This will head off any incompatibilities with Scaladoc found in the community I'm sure.

janekdb:
Seeing some encouraging HTML now...
image

Column alignment options looking good,
image

**Status**: All planned GHFM parsing done, initial end-to-end tests are good. Remaining: dead code removal, squash commits. I'll need to pick up that tomorrow morning and given that this should be good for merge decision.

src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
```diff
@@ -544,6 +568,340 @@ trait CommentFactoryBase { this: MemberLookupBase =>
...
+        val startMark = parseStartMark()
+
+        if (startMark.isEmpty) {
+          peek("Expected startMark")
+          sys.error("Precondition violated: Expected startMark.")
+        }
+
+        val inlineTerminators = makeInlineTerminators(startMark)
+
+        val content = Paragraph(inline(isInlineEnd = checkAny(inlineTerminators)))
```
lrytz: Is it true that all terminators are single chars? Then this could be made more efficient by just checking the buffer's current `char` instead of calling `check` many times. (We can do this in a follow-up PR). A first simple improvement would be to test in `check` if the `chars` string has length 1, and avoid calling `jump` in that case.

lrytz (commented):
LGTM! This is good to go once you're done with squashing and cleaning up.

src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala
janekdb: I'm doing the clean up now so will look at this.

janekdb:
@lrytz - This is the merge candidate.

Last actions,
* Removed most dead code
* Actioned some TODOs, consolidated remaining TODO comments
* Fixed incorrect parsing of empty leading cell
* Investigated your optimization suggestion to use `Char` checks now that the only delimiter is the pipe. It will work but not with the time I have today.
 * Squashed commits
 * Ran end-to-end test

Not done,
 * Make scala project Scaladoc
 * Make community Scaladoc

Every known instance of suboptimal parsing has a test case. This will make it easy to continue with the edge cases following merge.

Compared to the pre-GHFM version this had only light testing, however in balance the parsing code is much simpler thanks to the stricter requirements. The community build @SethTisue will give us more information about how well this copes with existing markdown.

I lay my offering down at the side of the lake and hope when the morning comes to see it gone.

adriaanm:
We gratefully and humbly accept your offerings! I have stood (by the sidelines) in awe at your diligence.

lrytz (approved):
🚀

janek-inferess:
Phew! Thanks for everyone's help and patience. I'll remind myself to get early community feedback before I plot a course the next time.

janekdb:
Phew! Thanks for everyone's help and patience. I'll remind myself to get early community feedback before I next plot a course.

SethTisue:
scala/community-builds@69bca5e2971e2d393476b5203fb42d65ac847644 moves to a Scala SHA that includes this PR, so we'll know by tomorrow if anything blows up

SethTisue:
@janekdb I would like to tweet about this feature from https://twitter.com/scala_lang (and credit you by name when we do so). do you have a tweet about it I could quote-tweet? (and/or could you propose a screenshot to include?)

janekdb:
@SethTisue How about this screenshot lifted from the forthcoming [blog post](https://github.com/janekdb/scala-lang/commits/topic/scaladoc-tables-part-one)

image

You can crop off the intitial text leaving just the markdown and the table if you like.

My tweet: https://twitter.com/janekdb/status/1045742653822955520

janekdb:
@SethTisue Here's a couple of alternative images,

image

image

SethTisue:
@janekdb I forgot about your blog post draft. rather than tweeting now, how about we publish the blog post and publicize the feature that way (including on Twitter)?

janekdb:
@SethTisue Blog post is done pending review: https://github.com/scala/scala-lang/pull/959

PR #6044, opened 2017-08-21
By default, only run DCE on methods with an ATHROW.

retronym:
GenBCode started to run a local DCE optimization on all methods to afford
simplicity in code generation of `Nothing`-typed expressions.
Benchmarks show this was costing 1-2% of compile times.

This commit selectively run DCE on methods with the problematic THROW, even
under `-opt:l:none`. It then changes `-opt:l:default` to the empty set of
optimizations.

An existing test for this code gen quirk is updated to show with either
explicitly enabled or selective targeted DCE, clean code is generated.


retronym:
(Based on top of #6012)

retronym:
[Comparative benchmark](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=58cfe9e76c%7C46bfecd54c&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40&from=1499584862985&to=1503285941893).

I'm looking into the test failure.

retronym:
ASM is inserting an unreachable frame into:

```
class C {
  def t(): Unit = {
    var x = ""
    while (x != null) {
      getClass
    }
  }
}
```

Which gets the `NOP;NOP;ATHROW` treatment.

```

  // access flags 0x1
  public t()V
   L0 @ L687685057
    LINENUMBER 2 L0 @ L687685057
    LINENUMBER 3 L0 @ L687685057
    LDC ""
   L1 @ L1604020967
    ASTORE 1
   L2 @ L277697988
    LINENUMBER 4 L2 @ L277697988
   FRAME APPEND [java/lang/String]
    ALOAD 1
    IFNULL L3 @ L1412612727
   L4 @ L367066629
    LINENUMBER 5 L4 @ L367066629
    ALOAD 0
    INVOKEVIRTUAL C.getClass ()Ljava/lang/Class;
    POP
    GOTO L2 @ L277697988
   L5 @ L287859212
   FRAME FULL [] [java/lang/Throwable]
    NOP
    NOP
    ATHROW
   L3 @ L1412612727
    LINENUMBER 4 L3 @ L1412612727
   FRAME APPEND [C java/lang/String]
    RETURN
   L6 @ L1810970264
    LOCALVARIABLE x Ljava/lang/String; L1 @ L1604020967 L3 @ L1412612727 1
    LOCALVARIABLE this LC; L0 @ L687685057 L6 @ L1810970264 0
    MAXSTACK = 1
    MAXLOCALS = 2
```

lrytz:
```
  def t(): Unit = {
    var x = ""
    while (x != null) {
      getClass
    }
  }
```

Code gen continues to generate code (the return statement) after the while loop, but that code is unreachable, so asm replaces it by `nop;..nop;athrow` during classfile writing. This happens for any method with unreachable code, so that's the change we'd have to live with if we don't enable DCE by default. Maybe we can identify some common special cases, like the above, and add those methods to `methodRequiringDCE`.

retronym:
Maybe its too ambitious, but we could modify (or extend) `asm.MethodWriter` to be able to call our DCE as a compensating action before resorting to resorting to `NOP; ATHROW`.

lrytz:
The replacing of unreachable code happens when computing stack map frames. First, frames (of reachable labels) are computed in a fixpoint algorithm. Then, code after unreachable labels is replaced (https://github.com/scala/scala-asm/blob/master/src/main/java/scala/tools/asm/MethodWriter.java#L1501).

It would be non-trivial to remove the code and fix up all indicies. They do something similar for handling long jumps (`GOTO_W`) in https://github.com/scala/scala-asm/blob/master/src/main/java/scala/tools/asm/MethodWriter.java#L2383, and it's hard.

We could remove the asm instructions and re-run the classfile writer, but that also sounds non-optimal.

As an alternative approach, we should try whether a home-grown DCE is faster the current [`removeUnreachableCodeImpl`](https://github.com/scala/scala/blob/v2.12.3/src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala#L487) which runs an Analyzer. Implementing that should not be hard, we already have our own "abstract" interpreter (not that abstract, but it could be made so..) in [`computeMaxLocalsMaxStack`](https://github.com/scala/scala/blob/v2.12.3/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala#L313-L330)

retronym:
> It would be non-trivial to remove the code and fix up all indicies.

I was thinking we'd abort writing that method, run our existing DCE, and restart the method writer. This assumes it is possible to do a "rollback" in the ClassWriter/MethodWriter. If methods with dead code are rare, the rollback need not be efficient.

retronym:
The DCE detection in ASM is probably faster than ours because it uses `Label.status` to mark reachable labels, whereas we build up a `Set`-s of `LabelNode`-s.

image

I think it safe to use `LabelNode.getLabel.status` to store flags so long as `InsnList.resetLabels` is called afterwards (either by us explicitly or by an enclosing `MethodNode.accept(MethodVisitor)`.

retronym:
I'll close this PR while we experiment with these alternative approaches. Let me know if you'd like to do this work, otherwise I'll come back to it next month.

retronym:
If the technique of using `Label.status` is feasible, we should also see if it helps performance elsewhere.

 `labelReferences` is another slow point. This isn't quite the same problem, but we might be able to replace:

```
val res = mutable.AnyRefMap[LabelNode, Set[AnyRef]]()
```

with use of `Label.info`

```
    /**
     * Field used to associate user information to a label. Warning: this field
     * is used by the ASM tree package. In order to use it with the ASM tree
     * package you must override the
     * {@link scala.tools.asm.tree.MethodNode#getLabelNode} method.
     */
    public Object info;
```

PR #6049, opened 2017-08-24
Remove deprecated view bounds in library

janekdb:
Replace view bounds with implicit parameter.

`evidence$1` is the name used by the parser when converting the view bound to a parameter but
`ev` is the conventional name. To ensure compatibility with code that might be using the
generated name parameter name deprecation is included.

## TODO

  - [x] Update the version following cherry-pick
  - [x] Use context bound and do not deprecate the implicit parameter name

### DIFF
--- a/src/library/scala/math/Ordering.scala
+++ b/src/library/scala/math/Ordering.scala
@@ -1,6 +1,6 @@
 /*                     __                                               *\
 **     ________ ___   / /  ___     Scala API                            **
-**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
+**    / __/ __// _ | / /  / _ |    (c) 2003-2017, LAMP/EPFL             **
 **  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
 ** /____/\___/_/ |_/____/_/ | |                                         **
 **                          |/                                          **
@@ -183,15 +183,19 @@ trait Ordering[T] extends Comparator[T] with PartialOrdering[T] with Serializabl
 }
 
 trait LowPriorityOrderingImplicits {
+
+  type AsComparable[A] = A => Comparable[A]
+
   /** This would conflict with all the nice implicit Orderings
    *  available, but thanks to the magic of prioritized implicits
    *  via subclassing we can make `Ordered[A] => Ordering[A]` only
    *  turn up if nothing else works.  Since `Ordered[A]` extends
    *  `Comparable[A]` anyway, we can throw in some Java interop too.
    */
-  implicit def ordered[A <% Comparable[A]]: Ordering[A] = new Ordering[A] {
+  implicit def ordered[A: AsComparable]: Ordering[A] = new Ordering[A] {
     def compare(x: A, y: A): Int = x compareTo y
   }
+
   implicit def comparatorToOrdering[A](implicit cmp: Comparator[A]): Ordering[A] = new Ordering[A] {
     def compare(x: A, y: A) = cmp.compare(x, y)
   }
--- a/src/library/scala/math/PartiallyOrdered.scala
+++ b/src/library/scala/math/PartiallyOrdered.scala
@@ -1,11 +1,10 @@
 /*                     __                                               *\
 **     ________ ___   / /  ___     Scala API                            **
-**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
+**    / __/ __// _ | / /  / _ |    (c) 2003-2017, LAMP/EPFL             **
 **  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
 ** /____/\___/_/ |_/____/_/ | |                                         **
 **                          |/                                          **
 \*                                                                      */
-
 
 
 package scala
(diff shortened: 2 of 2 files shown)


src/library/scala/math/Ordering.scala
```diff
@@ -189,7 +189,7 @@ trait LowPriorityOrderingImplicits {
    *  turn up if nothing else works.  Since `Ordered[A]` extends
    *  `Comparable[A]` anyway, we can throw in some Java interop too.
    */
-  implicit def ordered[A <% Comparable[A]]: Ordering[A] = new Ordering[A] {
+  implicit def ordered[A](implicit @deprecatedName('evidence$1, "2.12.3") ev: A => Comparable[A]): Ordering[A] = new Ordering[A] {
```
janekdb: Need to update the version following cherry-pick

som-snytt:
By definition or fiat or whatever, user code isn't supposed to use identifiers with dollars. A tool might interact with such idents, but shouldn't generate code with named params, which is a userland convenience.

The spec only says don't define idents with dollars.

The deprecation is sensitive in a good way, but also suggests the precedent that it's OK to rely on synthetic naming. I don't know if there's a policy when synthetics infect the API.  For example, if a function is lifted to some `foo$bar$baz()`, should I exploit that? I guess it matters what is visible to Scala code.

adriaanm:
I would rewrite instead into a context bound using `type FromComparable[A] = A => Comparable[A]`:  `implicit def ordered[A: FromComparable]: Ordering[A]`

Naming is hard, I welcome a better name than FromComparable :-)

janekdb:
Is the code now following your suggestion @adriaanm ? I have added `WIP` until I've understood this change is and why it is better.

Jasper-M:
> Naming is hard, I welcome a better name than FromComparable :-)

`ViewAsComparable`? `AsComparable`?

som-snytt:
`ComparableToComparable`.

adriaanm:
`AsComparable` sounds good to me. Ditto for `AsPartiallyOrdered`.

(Or `CanHazComparable`? 😁 )

janekdb:
Hmm, build error is,
```
[info] Failed: Total 759, Failed 0, Errors 1, Passed 758
[error] Error during tests:
[error] 	scala.reflect.quasiquotes.TypecheckedProps


[info] ! typechecked.tuple term: Exception raised on property evaluation.
[info] > Exception: java.lang.AssertionError: assertion failed: AnyRef
```

src/library/scala/math/PartiallyOrdered.scala
```diff
@@ -18,33 +17,22 @@ package math
...
-    }
+  def tryCompareTo[B >: A : AsPartiallyOrdered](that: B): Option[Int]
+
+  def <  [B >: A : AsPartiallyOrdered](that: B): Boolean = (this tryCompareTo that) exists (_ < 0)
+
+  def >  [B >: A : AsPartiallyOrdered](that: B): Boolean = (this tryCompareTo that) exists (_ > 0)
+
+  def <= [B >: A : AsPartiallyOrdered](that: B): Boolean = (this tryCompareTo that) exists (_ <= 0)
+
+  def >= [B >: A : AsPartiallyOrdered](that: B): Boolean = (this tryCompareTo that) exists (_ >= 0)
```
adriaanm: Let's keep this PR focused on removing deprecation. It doesn't look like the inliner is able to recover the code we had before. Not sure why /cc @lrytz 

From `wget https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots/org/scala-lang/scala-library/2.13.0-pre-889443e-SNAPSHOT/scala-library-2.13.0-pre-889443e-20170830.112331-1.jar ; unzip scala-library-2.13.0-pre-889443e-20170830.112331-1.jar scala/math/PartiallyOrdered.class ; cfr scala/math/PartiallyOrdered.class`

```
    default public <B> boolean $greater$eq(B that, Function1<B, PartiallyOrdered<B>> evidence$5) {
        return this.tryCompareTo(that, evidence$5).exists(x$4 -> {
            if (x$4 >= 0) {
                return true;
            }
            return false;
        }
        );
    }
```

adriaanm (changes requested):
Thanks for scouting the area, but I would prefer to restrict the changes to dealing with deprecations. As far as I can tell, the inliner does not inline the `exists` call.

janekdb:
Reverted to deprecation only.

lrytz (approved):

src/library/scala/math/PartiallyOrdered.scala
lrytz: That's a good question. If I compile https://github.com/scala/scala/commit/889443e locally using 2.13.0-M2 (the starr) and pass `-opt:l:inline '-opt-inline-from:**'`, then inlining works correctly.

lrytz: If I run `sbt 'setupPublishCore https://scala-ci.typesafe.com/artifactory/scala-pr-validation-snapshots' publishLocal`, then it's not inlined. I'll dig into this.

lrytz: Mistery solved: https://github.com/scala/scala-dev/issues/424

adriaanm:
Thanks @janekdb!

src/library/scala/math/PartiallyOrdered.scala
adriaanm: :+1:

PR #6050, opened 2017-08-29
Support by-name implicits with recursive dictionaries

milessabin:
**This Typelevel Scala collaboration sponsored by Lightbend ...**

This is an implementation of byname implicit parameters with recursive dictionaries, intended as a language-level replacement for shapeless's `Lazy` type. As of this PR, implicit arguments can be marked as byname and at call sites recursive uses of implicit values are permitted if they occur in an implicit byname argument.

There is a SIP describing this language change in greater detail [here](http://docs.scala-lang.org/sips/byname-implicits.html).

Consider the following example,

```scala
trait Foo { def next: Foo }
object Foo {
  implicit def foo(implicit rec: Foo): Foo = new Foo { def next = rec }
}
val foo = implicitly[Foo]
assert(foo eq foo.next)
```
In current Scala this would diverge due to the recursive implicit argument `rec` of method `foo`. Under the scheme implemented in this PR we can mark the recursive implicit parameter as byname,

```scala
trait Foo { def next: Foo }
object Foo {
  implicit def foo(implicit rec: => Foo): Foo = new Foo { def next = rec }
                              // ^^
}
val foo = implicitly[Foo]
assert(foo eq foo.next)
```
and recursive occurrences of this sort are extracted out as `val` members of a local synthetic object as follows,
```scala
val foo: Foo = scala.Predef.implicitly[Foo](
  {
    object LazyDefns$1 {
      val rec$1: Foo = Foo.foo(rec$1)
                       //      ^^^^^
                       // recursive knot tied here
    }
    LazyDefns$1.rec$1
  }
)
assert(foo eq foo.next)
```
The example now compiles with the assertion successful. Note that the recursive use of `rec$1` occurs within the byname argument of `foo` and is consequently deferred. The desugaring matches what a programmer would do to construct such a recursive value explicitly.

This general pattern is essential to the derivation of type class instances for recursive data types, one of shapeless's most common applications. Byname implicits have a number of benefits over the macro implementation of `Lazy` in shapeless,

+ the implementation of `Lazy` in shapeless is extremely delicate, relying on non-portable compiler internals. By contrast there is already an initial implementation of byname implicits in Dotty: lampepfl/dotty#1998.
+ the shapeless implementation is unable to modify divergence checking, so to solve recursive instances it effectively disables divergence checking altogether ... this means that incautious use of `Lazy` can cause the typechecker to loop indefinitely. The byname implicits implementation is able to both solve recursive occurrences and check for divergence.
+ the implementation of `Lazy` interferes with the heuristics for solving inductive implicits in #6481 because the latter depends on being able to verify that induction steps strictly reduce the size of the types being solved for; the additional `Lazy` type constructors make the type appear be non-decreasing in size. Whilst this could be special-cased, doing so would require some knowledge of shapeless to be incorporated into the compiler. Being a language-level feature, byname implicits can be accommodated directly in the induction heuristics.
+ in common cases more implicit arguments would have to be marked as `Lazy` than would have to be marked as byname under this PR due to restrictions on what the `Lazy` macro is able to do. Given that there is a runtime cost associated with capturing the thunks required for both `Lazy` and byname arguments, any reduction in the number is beneficial.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -90,7 +90,9 @@ trait Implicits {
     if (shouldPrint)
       typingStack.printTyping(tree, "typing implicit: %s %s".format(tree, context.undetparamsString))
     val implicitSearchContext = context.makeImplicit(reportAmbiguous)
-    val search = new ImplicitSearch(tree, pt, isView, implicitSearchContext, pos)
+    val dpt = if (isView) pt else dropByName(pt)
+    val isByName = dpt ne pt
+    val search = new ImplicitSearch(tree, dpt, isView, isByName, implicitSearchContext, pos)
     pluginsNotifyImplicitSearch(search)
     val result = search.bestImplicit
     pluginsNotifyImplicitSearchResult(result)
@@ -152,7 +154,7 @@ trait Implicits {
     val tvars = tpars map (TypeVar untouchable _)
     val tpSubsted = tp.subst(tpars, tvars)
 
-    val search = new ImplicitSearch(EmptyTree, functionType(List(tpSubsted), AnyTpe), true, context.makeImplicit(reportAmbiguousErrors = false))
+    val search = new ImplicitSearch(EmptyTree, functionType(List(tpSubsted), AnyTpe), true, false, context.makeImplicit(reportAmbiguousErrors = false))
 
     search.allImplicitsPoly(tvars)
   }
@@ -291,11 +293,16 @@ trait Implicits {
     private def isView_=(value: Boolean): Unit = _isView = value
 
     private[this] var _isView: Boolean = false
+
+    def isByName: Boolean = _isByName
+    private def isByName_=(value: Boolean): Unit = _isByName = value
+    private[this] var _isByName: Boolean = false
   }
   object OpenImplicit {
-    def apply(info: ImplicitInfo, pt: Type, tree: Tree, isView: Boolean): OpenImplicit = {
+    def apply(info: ImplicitInfo, pt: Type, tree: Tree, isView: Boolean, isByName: Boolean): OpenImplicit = {
       val result = new OpenImplicit(info, pt, tree)
       result.isView = isView
+      result.isByName = isByName
       result
     }
   }
@@ -382,7 +389,7 @@ trait Implicits {
    *                          (useful when we infer synthetic stuff and pass EmptyTree in the `tree` argument)
    *                          If it's set to NoPosition, then position-based services will use `tree.pos`
    */
-  class ImplicitSearch(val tree: Tree, val pt: Type, val isView: Boolean, val context0: Context, val pos0: Position = NoPosition) extends Typer(context0) with ImplicitsContextErrors {
+  class ImplicitSearch(val tree: Tree, val pt: Type, val isView: Boolean, val isByNamePt: Boolean, val context0: Context, val pos0: Position = NoPosition) extends Typer(context0) with ImplicitsContextErrors {
     val searchId = implicitSearchId()
     private def typingLog(what: String, msg: => String) = {
       if (printingOk(tree))
@@ -532,9 +539,7 @@ trait Implicits {
       val existsDominatedImplicit: Boolean =
         if(tree == EmptyTree) false
         else {
-          val ptByName = isByNameParamType(pt)
-          val dpt = if (ptByName) dropByName(pt) else pt
-          lazy val spt = stripped(core(dpt))
+          lazy val spt = stripped(core(pt))
           lazy val sptSyms = allSymbols(spt)
           // Are all the symbols of the stripped core of pt contained in the stripped core of tp?
           def coversPt(tp: Type): Boolean = {
@@ -547,43 +552,41 @@ trait Implicits {
             ois match {
               case Nil => false
               case (hd@OpenImplicit(info1, tp, tree1)) :: tl =>
-                val byName = isByNameParamType(tp)
-                val dtp = if (byName) dropByName(tp) else tp
                 (if (!info1.sym.isMacro && tree1.symbol == tree.symbol) {
-                  if(belowByName && (dtp =:= dpt)) Some(false) // if there is a byname argument between tp and pt we can tie the knot
-                  else if (dominates(dpt, dtp) && coversPt(dtp)) Some(true)
+                  if(belowByName && (tp =:= pt)) Some(false) // if there is a byname argument between tp and pt we can tie the knot
+                  else if (dominates(pt, tp) && coversPt(tp)) Some(true)
                   else None
                 } else None) match {
                   case Some(res) => res
-                  case None => loop(tl, byName || belowByName)
+                  case None => loop(tl, hd.isByName || belowByName)
                 }
             }
           }
-          loop(context.openImplicits, ptByName)
+          loop(context.openImplicits, this.isByNamePt)
         }
 
       if(existsDominatedImplicit) {
         //println("Pending implicit "+pending+" dominates "+pt+"/"+undetParams) //@MDEBUG
         DivergentSearchFailure
       } else {
-        val ref = context.refByNameImplicit(dropByName(pt))
+        val ref = context.refByNameImplicit(pt)
         if(ref != EmptyTree)
           new SearchResult(ref, EmptyTreeTypeSubstituter, Nil)
         else {
           val recursiveImplicit: Option[OpenImplicit] =
             if(tree == EmptyTree) None
             else context.openImplicits find {
-              case OpenImplicit(info, tp, tree1) =>
-                (isByNameParamType(tp) || isByNameParamType(pt)) && dropByName(tp) <:< dropByName(pt)
+              case oi @ OpenImplicit(info, tp, tree1) =>
+                (oi.isByName || isByNamePt) && oi.pt <:< pt
             }
 
           recursiveImplicit match {
             case Some(rec) =>
-              val ref = atPos(pos.focus)(context.linkByNameImplicit(dropByName(rec.pt)))
+              val ref = atPos(pos.focus)(context.linkByNameImplicit(rec.pt))
               new SearchResult(ref, EmptyTreeTypeSubstituter, Nil)
             case None =>
               try {
-                context.openImplicits = OpenImplicit(info, pt, tree, isView) :: context.openImplicits
+                context.openImplicits = OpenImplicit(info, pt, tree, isView, isByNamePt) :: context.openImplicits
                 //println("  "*context.openImplicits.length+"typed implicit "+info+" for "+pt) //@MDEBUG
                 val result = typedImplicit0(info, ptChecked, isLocalToCallsite)
                 if (result.isDivergent) {
@@ -591,7 +594,7 @@ trait Implicits {
                   if (context.openImplicits.tail.isEmpty && !pt.isErroneous)
                     DivergingImplicitExpansionError(tree, pt, info.sym)(context)
                   result
-                } else context.defineByNameImplicit(dropByName(pt), result)
+                } else context.defineByNameImplicit(pt, result)
               } finally {
                 context.openImplicits = context.openImplicits.tail
               }
@@ -688,7 +691,7 @@ trait Implicits {
         case NullaryMethodType(restpe)  => loop(restpe, pt)
         case PolyType(_, restpe)        => loop(restpe, pt)
         case ExistentialType(_, qtpe)   => if (fast) loop(qtpe, pt) else normalize(tp) <:< pt // is !fast case needed??
-        case _                          => if (fast) isPlausiblySubType(tp, if(isView) pt else dropByName(pt)) else tp <:< dropByName(pt)
+        case _                          => if (fast) isPlausiblySubType(tp, pt) else tp <:< pt
       }
       loop(tp0, pt0)
     }
@@ -758,7 +761,7 @@ trait Implicits {
       typingLog("considering", typeDebug.ptTree(itree1))
 
       @inline def fail(reason: => String): SearchResult = failure(itree0, reason)
-      def fallback = typed1(itree1, EXPRmode, dropByName(wildPt))
+      def fallback = typed1(itree1, EXPRmode, wildPt)
       try {
         val itree2 = if (!isView) fallback else pt match {
           case Function1(arg1, arg2) =>
@@ -796,7 +799,7 @@ trait Implicits {
         if (StatisticsStatics.areSomeColdStatsEnabled) statistics.incCounter(typedImplicits)
 
         val itree3 = if (isView) treeInfo.dissectApplied(itree2).callee
-                     else adapt(itree2, EXPRmode, dropByName(wildPt))
+                     else adapt(itree2, EXPRmode, wildPt)
 
         typingStack.showAdapt(itree0, itree3, pt, context)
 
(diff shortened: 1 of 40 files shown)


tpolecat:
This is great. Thank you!

tek:
yaaaaay awesome!

andyscott:
Yes please!

fommil:
Fantastic. I have some examples I'd like to try out when this is ready for general testing.

fommil:
One potential problem with this overload/behaviour change of the by-name language feature is that I was planning on using regular byname implicit parameters to avoid constructing the instance when it's not needed. This is necessary to get good memory allocation optimisation for implicits with conditional requirements... caching of `implicit def` (for runtime performance optimisation) is one such situation I'm currently working on, e.g. https://gitlab.com/fommil/stalactite/issues/35

Is there any requirement for `Strict`/`Cached` after this?

fommil:
It's there any requirement for Strict/Cached after this?

julienrf:
From what I understand the effective semantics is by-need, not by-name? So, what’s preventing us from replacing the `=>` with a `lazy`?

fommil:
@julienrf I would very much prefer what you're suggesting, and therefore not impact byname semantics.

lrytz:
@julienrf at least when provided explicitly, it behaves as by-name, not by-need.

```
Welcome to Scala 2.12.4-20170829-122239-87b912c (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).

scala> def f(implicit x: => Int) = { x + x }
f: (implicit x: => Int)Int

scala> f { println("hi"); 1 }
hi
hi
res1: Int = 2
```

milessabin:
Being implicit doesn't change the by name semantics. I think it would also make sense to have lazy arguments (both implicit and explicit), but that's somewhat orthogonal to this PR. See https://github.com/scala/bug/issues/240 and https://github.com/lampepfl/dotty/issues/3005 for more discussion.

retronym:
> This general pattern is essential to the derivation of type class instances for recursive data types, one of shapeless's most common applications. 

Could you share a small motivating example that currently diverges? (The example above is distilled too far for me to see the use case)

fommil:
@retronym https://gitlab.com/fommil/stalactite/tree/master/src/test/scala/xmlformat

Try removing the Lazy, Cached and Strict here in either the generic encoder or decoder and see the recursive tests fail.

I must admit, I'm confused about exactly what this means for my usecase... I'm unable to see the above in my example. Does this only replace Lazy? What about Cached and Strict? What's the implication to performance, because if these things really do become by-name it's going to kill performance when the lazy semantics are not necessary. I have a good usecase for byname implicits, which does help performance, but added across the board this will be terrible.

milessabin:
@retronym there are motivating examples in the tests. [`test/files/pos/byname-implicits-9.scala`](https://github.com/scala/scala/pull/6050/files#diff-ac49f3df7775e08c9544baacc25f633e) (a `Show` type class derivation exercised on `List`) is an example with explicit recursion in the ADT,
```
List[T] => Either[Nil.type, ::[T]]
::[T] => (T, List[T])
```
Without the ability to tie the knot here the compiler would diverge on seeing the occurrence of `List[T]` in `::[T]`.

milessabin:
@fommil `Cached` and `Strict` will go away too, as will `cachedImplicit`. There will be an alternative mechanism for handling caching.

adriaanm:
2.13 is a better target for this work. 2.12 is now at a point where we need to focus on stability and bug compatibility.

fommil:
I have an alternative mechanism (to shapeless Generic / LabelledGeneric) for typeclass derivation that also hits the recursive types problem. I'll be interested to see if this fixes it. In the meantime I can use Lazy and Cached, hopefully without typeclass authors or end users needing to know about it.

OlivierBlanvillain:
I've been playing around with this branch and found some variation between Dotty and this PR. [Here is](https://gist.github.com/OlivierBlanvillain/e2acd9bce49a871a86ba8ec84ea43c16) an isolated example (adapted from [shapeless-type-class-derivation-2015-demo](https://github.com/milessabin/shapeless-type-class-derivation-2015-demo)) that compiles using Dotty, but diverges on this PR.

I suspect the difference comes from the divergence checking.

milessabin:
@OlivierBlanvillain That's a very useful example. Thanks!

It wouldn't expect the `eqH` arguments of `eqHCons` and `eqCCons` to have to be byname, and if I make them strict then your example compiles with this PR. That said, I think it should also compile if they _are_ byname ... I'll investigate.

You're quite right that it's a difference in the divergence checker which is responsible for the difference here. I've implemented a slightly more sophisticated test than the user specified upper bound that you've  gone for in Dotty. I think that it would be better to avoid making the limit a compilation parameter if at all possible.

I'm putting together a SIP for this at the moment and I'll include the details of the divergence criteria along with a proof that implicit search either terminates successfully or fails reporting divergence. I'm fairly confident that you will be able to port it over to the Dotty implementation.

milessabin:
Rebased to 2.13.x.

@OlivierBlanvillain your example now works as is, but also with just the generic instance having a byname argument ... see [test/files/pos/byname-implicits-13.scala](https://github.com/scala/scala/pull/6050/files#diff-b56888e09205efffc4c2989275065116) and [test/files/pos/byname-implicits-14.scala](https://github.com/scala/scala/pull/6050/files#diff-6cd9a04bad934419ff43f1f98d165750).

OlivierBlanvillain:
@milessabin I just compiled all the pos & run tests from this PR on dotty master and it seams that both compilers agree as far as bymane implicits are concerned :+1:

milessabin:
@OlivierBlanvillain oh, awesome :-)

milessabin:
Jenkins wedged?

milessabin:
@SethTisue This seems to be wedged on the CLA bot?

milessabin:
Community build [here](https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/857/) although I think it might already be failing.

SethTisue:
/sync

EDIT: yup, that did it

milessabin:
Community build is green :-)

joroKr21:
@milessabin I'm able to fool the divergence check via bootstrapping. Try adding the code below to `test/files/run/byname-implicits-6.scala` and wait for it to compile. This in case you still intend to prove termination or divergence error like you mentioned above.
```scala
  case class Bootstrap[+A](head: A, tail: Option[Bootstrap[(A, A)]])
  object Bootstrap {
    type BootstrapRepr[+A] = A :: Option[Bootstrap[(A, A)]] :: HNil
    implicit def bootstrapGen[A]: Generic.Aux[Bootstrap[A], BootstrapRepr[A]] =
      new Generic[Bootstrap[A]] {
        type Repr = BootstrapRepr[A]
        def to(t: Bootstrap[A]): Repr = t.head :: t.tail :: HNil
        def from(r: Repr): Bootstrap[A] = Bootstrap(r.head, r.tail.head)
      }
  }

  class Tc[A]
  object Tc {
    implicit val tcInt: Tc[Int] = new Tc
    implicit def tcOpt[A: Tc]: Tc[Option[A]] = new Tc
    implicit def tcTuple[A: Tc, B: Tc]: Tc[(A, B)] = new Tc
    implicit val tcHNil: Tc[HNil] = new Tc
    implicit def tcHCons[H: Tc, T <: HList: Tc]: Tc[H :: T] = new Tc
    implicit def tcGen[A, R <: HList](
      implicit gen: Generic.Aux[A, R], tcR: => Tc[R]
    ): Tc[A] = new Tc
  }

  implicitly[Tc[Bootstrap[Int]]]
```

milessabin:
@joroKr21 thanks ... I'll investigate.

milessabin:
@joroKr21 thanks again for that ... there was a hole as big as a bus in the implementation which your counterexample helped me home in on. It's handled correctly now, and the algorithm is somewhat simpler and much more amenable to a termination proof.

milessabin:
Rebased.

lrytz:
Status summary for M3: there are no open questions to be addressed. The PR needs a review / approval. Assigning @adriaanm, please pass it on if someone else should do it.

milessabin:
SIP here: https://github.com/scala/docs.scala-lang/pull/1002.

milessabin:
Rebased.

erikerlandson:
This should be a big win for [coulomb](https://github.com/erikerlandson/coulomb/pull/24). Is it going to land in 2.13 soon?

milessabin:
Added [test case](https://github.com/scala/scala/pull/6050/commits/e7d547a0202d1c467860fe756eff46ed607149ec#diff-f9756a631b9f28fca6ef70d6f139d1a8) corresponding to @erikerlandson's [scenario](https://stackoverflow.com/questions/49088319/a-diverging-implicit-expansion-in-scala-involving-chained-implicits), rebased, squashed, and elaborated commit message.

test/files/pos/byname-implicits-22.scala
```diff
@@ -0,0 +1,39 @@
...
+  trait Y
+  implicit val declareY = Atomic[Y]
+  trait Z
+  implicit val declareZ = Atomic[Z]
+
+  trait Q
+  implicit val declareQ = Assign[Q, (X + Y) + Z]
+  trait R
+  implicit val declareR = Assign[R, Q + Z]
+}
```
erikerlandson: consider adding something like the following to exercise the example
```scala
implicitly[AsString[X]]
implicitly[AsString[X + Y]]
implicitly[AsString[Q]]
implicitly[AsString[R]]
```

milessabin:
@erikerlandson well ... duh ... yes, I omitted the business end of that test. Added now ... thanks :-)

milessabin:
Rebased.

milessabin:
Squashed, rebased, spec updated.

The latest version of the SIP is [here](https://github.com/milessabin/docs.scala-lang/blob/5d510d69dd56d4c7ce4bab63841e97b5521db9e1/_sips/sips/2017-11-20-byname-implicits.md), and @odersky is happy with the new covering-set based divergence criterion.

A lot of the additions to the spec are in fact bringing it in to line with the state of the divergence checking prior to this PR ... it hadn't kept up with changes that had been made to the implementation over the years.

milessabin:
Updated PR description to match the current state of play and link to the latest iteration of the SIP and the inductive implicits PR.

milessabin:
Rebased, also reinstated a change from 3cc2360b5e51e3d228799700d914f754291b8f36 which got lost in earlier rebasing.

adriaanm:
As discussed, we'll review in Berlin and give ourselves until M5 to refine.

milessabin:
Rebased.

milessabin:
@retronym I saw that you made some fairly major changes to `Contexts` ... this PR adds some bookkeeping state there ... please let me know if you think it now belongs elsewhere.

retronym:
@milessabin Would it fit into the existing `context.openImplicits`, by adding some extra state to `OpenImplicit`?

milessabin:
@retronym possibly ... I'll take a look if you think that's worth pursuing.

milessabin:
@retronym I did a quick and fairly disgusting spike along the lines that you suggested [on a branch in my fork](https://github.com/milessabin/scala/tree/topic/benchmarking-byname-implicits).

Then I used @rorygraves' `perf_tester` to benchmark the state prior to this PR vs. this PR vs. the spike. The results are [here](https://gist.github.com/milessabin/5385e70f9bbf2f613dd62ffb5f8adf25): corpus is `scala-library`, 50 compilations for each commit, and three JVM runs.

Looking at the top 20%,

```
                 Run Name       Cycle   samples            Wall time (ms)              All Wall time (ms)                         CPU(ms)                  Idle time (ms)                  Allocated(MBs)
                   pre-pr          1      10     8,426.70 [-1.35% +0.73%]        8,426.70 [-1.35% +0.73%]        8,342.60 [-1.59% +0.77%]              0.00 [ NaN%  NaN%]        2,736.66 [-0.12% +0.03%]
                   pre-pr          2      10     8,227.10 [-1.03% +0.64%]        8,227.10 [-1.03% +0.64%]        8,137.10 [-0.94% +0.58%]              0.00 [ NaN%  NaN%]        2,738.56 [-0.25% +0.04%]
                   pre-pr          3      10     8,326.00 [-0.67% +0.82%]        8,326.00 [-0.67% +0.82%]        8,238.40 [-0.62% +0.70%]              0.00 [ NaN%  NaN%]        2,738.08 [-0.01% +0.01%]
                       pr          1      10     8,307.10 [-0.57% +0.61%]        8,307.10 [-0.57% +0.61%]        8,214.20 [-0.72% +0.55%]              0.00 [ NaN%  NaN%]        2,726.58 [-0.02% +0.03%]
                       pr          2      10     8,313.40 [-1.10% +0.72%]        8,313.40 [-1.10% +0.72%]        8,223.70 [-0.99% +0.75%]              0.00 [ NaN%  NaN%]        2,756.78 [-0.02% +0.01%]
                       pr          3      10     8,411.30 [-0.67% +0.80%]        8,411.30 [-0.67% +0.80%]        8,324.50 [-0.80% +0.73%]              0.00 [ NaN%  NaN%]        2,758.38 [-0.03% +0.03%]
           jason-feedback          1      10     8,321.30 [-0.82% +0.63%]        8,321.30 [-0.82% +0.63%]        8,238.60 [-0.77% +0.66%]              0.00 [ NaN%  NaN%]        2,735.57 [-0.01% +0.01%]
           jason-feedback          2      10     8,321.10 [-0.65% +0.55%]        8,321.10 [-0.65% +0.55%]        8,229.40 [-0.89% +0.76%]              0.00 [ NaN%  NaN%]        2,771.65 [-0.27% +0.06%]
           jason-feedback          3      10     8,139.10 [-0.35% +0.44%]        8,139.10 [-0.35% +0.44%]        8,054.10 [-0.47% +0.46%]              0.00 [ NaN%  NaN%]        2,744.32 [-0.01% +0.01%]
```
my reaction was, "I read that as pr is faster than pre-pr and jason-feedback is faster than pr, but it's all pretty marginal and noisy between JVM runs" and @rorygraves' was "Yeah. I would read that as basically the same."

The spike is pretty gross. It could be cleaned up a little, but I think the benefits are marginal at best. So I'd prefer to merge this as-is.

milessabin:
Rebased.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -619,7 +688,7 @@ trait Implicits {
         case NullaryMethodType(restpe)  => loop(restpe, pt)
         case PolyType(_, restpe)        => loop(restpe, pt)
         case ExistentialType(_, qtpe)   => if (fast) loop(qtpe, pt) else normalize(tp) <:< pt // is !fast case needed??
-        case _                          => if (fast) isPlausiblySubType(tp, pt) else tp <:< pt
+        case _                          => if (fast) isPlausiblySubType(tp, if(isView) pt else dropByName(pt)) else tp <:< dropByName(pt)
```
retronym: I'll check this for performance.

retronym:
I'm in the process of reviewing some of the mechanics of this change in https://github.com/retronym/scala/tree/review/6050

retronym:
@milessabin I've finished proposing changes in https://github.com/retronym/scala/tree/review/6050.

The main simplification I found was to do the determine `isByName` and compute `dpt` before constructing the `ImplicitSearch`. Do you see any problems with that approach?

retronym (dismissed):
Suggested changes in https://github.com/retronym/scala/tree/review/6050

milessabin:
@retronym those changes look fine. I'll rebase and pull them in to this PR.

milessabin:
@retronym there are a couple of what look to be unrelated commits in there: https://github.com/retronym/scala/commit/e3cf13edd0dbc82fac6c021201d9fa722eacf010 and https://github.com/retronym/scala/commit/67a866510686823c2b9650781c47aa56d674bbbb.

I'll leave them out.

milessabin:
Rebased and incorporated @retronym's suggestions.

milessabin:
@adriaanm I think this is ready to merge now :-)

adriaanm:
🍰

milessabin:
:yum:

kubukoz:
Amazing work, thank you so much @milessabin and everyone involved! 2.13 is going to be so great!

PR #6051, opened 2017-08-30
Initialize statistics per global

jvican:
As described in the bug report, statistics are currently shared by
different instances of `Global` because they are located in objects.

This approach has an important disadvantage.  Statistics from different
global instances step on each other's toes, rendering the statistics
infrastructure unusable in most of the Scala developers' workflows
(think sbt running in parallel several compilations). When the data is
not useless, it produces races conditions and kills compilation.

This patch addresses the issue by creating a new object in `SymbolTable`
that is overriden in `Global` and hence available to most of the
compiler internals. This object contains all the statistics at the known
definition site (`Global` defines more statistics that are created in
`scala.tools.nsc` instead of `scala.reflect`).

The main goal of this implementation was to avoid moving the definition
of statistics from the places where they are used. Currently, they are
defined alongside the classes that have the call-sites (see, for
example, `Implicits.scala`, `Macros.scala` or `Types.scala`).

**NOTE**: This PR builds upon https://github.com/scala/scala/pull/6034, so it needs to be merged after it.
  
**EDIT**: Here's a summary of what this PR does:
  
1. It makes an overhaul of the previous statistics infrastructure that was
dependent on global objects shared by different instances of Global (making the
stats unusable in most of the cases).
1. It enables IOStats with a special trick that uses ThreadLocal to share
instances of quantities (timers/counters) as suggested by @lrytz.
1. It modifies the statistics framework to listen to changes in the mutable
settings and thus enable/disable according to what the driver of the compiler
does. This allows to run statistics in controlled and interactive environments,
enabling and disabling several times on the go.
1. It deduplicates calls to canEnable that could be hindering optimizations or
affecting the runtime.

Fix scala/bug#10460


src/reflect/scala/reflect/internal/SymbolTable.scala
```diff
@@ -51,6 +51,18 @@ abstract class SymbolTable extends macros.Universe
...
   val gen = new InternalTreeGen { val global: SymbolTable.this.type = SymbolTable.this }
 
+  trait ReflectStats extends BaseTypeSeqsStats
+                        with TypesStats
+                        with SymbolTableStats
+                        with TreesStats
+
+  /** Some statistics (normally disabled) set with -Ystatistics */
+  // This is a val so that we can override the `symbolTable` and `nodeCount` fields in `Global`
+  val statistics: ReflectStats = new {
```
jvican: If this were an `object`, the definition of `symbolTable` and `nodeCount` would be `final` and the override in `Global` would fail.

src/reflect/scala/reflect/internal/SymbolTable.scala
```diff
@@ -432,6 +444,9 @@ abstract class SymbolTable extends macros.Universe
   implicit val StringContextStripMarginOps: StringContext => StringContextStripMarginOps = util.StringContextStripMarginOps
 }
 
-object SymbolTableStats {
+trait SymbolTableStats {
+  self: scala.reflect.internal.TypesStats =>
   val phaseCounter = Statistics.newCounter("#phase calls")
+  // Defined here because `SymbolLoaders` needs it and has no access to global
+  val classReadNanos = Statistics.newSubTimer("time classfilereading", typerNanos)
```
jvican: Unfortunately, `SymbolLoaders`'s statistics comes from the `SymbolTable`, not `Global`, so this had to be moved here because `statistics` at this point hasn't been overridden and only contains the timers and counters from Scala reflect.

src/reflect/scala/reflect/internal/util/StatisticsInfo.scala
```diff
@@ -3,29 +3,29 @@
...
-  import global._
-  import scala.reflect.internal.TreesStats.nodeByType
+  import symbolTable._
+  import statistics.{treeNodeCount, nodeByType}
 
   val retainedCount  = Statistics.newCounter("#retained tree nodes")
   val retainedByType = Statistics.newByClass("#retained tree nodes by type")(Statistics.newCounter(""))
 
-  def print(phase: Phase) = if (settings.Ystatistics contains phase.name) {
+  def print(phase: Phase, units: Iterator[Global#CompilationUnit]) = {
```
jvican: `Global#CompilationUnit` is necessary because of a type mismatch at the call-site -- the `CompilationUnit` in `SymbolLoaders` is not the same as in `Global`.

jvican:
Oops, I've just realised this is not enough. `scala.reflect.internal.util.Statistics` also has a global `qs` with all the quantities and I need to change that. Unless there's a creative way to do it, that's a breaking change.

src/compiler/scala/tools/nsc/backend/jvm/BackendStats.scala
```diff
@@ -9,7 +9,7 @@ package backend.jvm
 import scala.reflect.internal.util.Statistics
 
 // Enable with `-Ystatistics:jvm`
-object BackendStats {
+trait BackendStats {
```
lrytz: Changing this to a trait will make the `timed` method defined in `BackendStats` non-inlineable, so it should be marked `final`. And it should also be moved to a different place, now that it's available in all statistics.

Please check for similar cases around.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -158,10 +158,18 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+  import scala.reflect.internal.util.StatisticsInfo
+  trait AllStats extends ReflectStats
+                    with TypersStats
+                    with ImplicitsStats
+                    with MacrosStats
+                    with BackendStats
+
+  override object statistics extends {
+    override type Global = Global.this.type
+    override val symbolTable: SymbolTable = Global.this
```
lrytz: maybe if you refine the type of `symbolTable` to `Global.this.type`, you don't need the `Global#CompilationUnit` type (and the `Global` abstract type) below (https://github.com/scala/scala/pull/6051/commits/08cd6c18aa290458efde61010e680636b18b48a2#r136072193)

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -158,10 +158,18 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+    override var nodeCount: Int = Global.this.nodeCount
```
lrytz: why is this override needed? isn't that the same var anyway?

src/reflect/scala/reflect/internal/Trees.scala
```diff
@@ -1910,11 +1910,11 @@ trait Trees extends api.Trees {
...
   implicit val ValOrDefDefTag         = ClassTag[ValOrDefDef](classOf[ValOrDefDef])
-
-  val treeNodeCount = Statistics.newView("#created tree nodes")(nodeCount)
 }
 
-object TreesStats {
-  // statistics
+trait TreesStats {
+  // Nodecount is only accessible from within `Trees.scala`, so this is mixed in at `SymbolTable`
+  var nodeCount: Int
```
lrytz: This could be an abstract getter and setter.

But how about moving the abstract `val symbolTable` into `ReflectStats` and using a `self: ReflectStats =>` here, then you could just use `symbolTable.nodeCount` on the next line, and get rid of the `override var nodeCount`.

src/reflect/scala/reflect/internal/SymbolTable.scala
lrytz: You mean it's not ideal because `SymbolLoaders` is defined in `scala.tools.nsc` and not `scala.reflect`? Maybe add that to the comment. But it's no big deal, really.

lrytz (commented):
For the binary incompatibility, you could add a concrete method `def hotStatisticsEnabled: Boolean = false` in `reflect.internal.settings.MutableSettings` and override it in `tools.nsc.settings.ScalaSettings` (using the existing `BooleanSetting`). The full change can be done in 2.13.x.

src/compiler/scala/tools/nsc/Global.scala
jvican: I tried this, and the compiler wasn't able to override it.

src/reflect/scala/reflect/internal/SymbolTable.scala
jvican: Yeah, exactly. I'll add it. But good to know it's not a big deal. My goal was to avoid moving the definition of statistics elsewhere, so I was a little bit disappointed with this :smile:.

src/compiler/scala/tools/nsc/backend/jvm/BackendStats.scala
jvican: Good catch. Didn't realise the effect on inlining that this change could bring. I'm checking for similar cases around and I'll move `timed` to `StatisticsInfo`.

src/reflect/scala/reflect/internal/Trees.scala
jvican: I fixed this in a similar way: I added an abstract val `symbolTable` in `Trees.scala` and use `symbolTable.nodeCount`. That allows me to keep `ReflectStats` defined in `SymbolTable`, avoid the cyclic dependency and only override that in the call-site. Thanks for the suggestion!

src/reflect/scala/reflect/internal/util/StatisticsInfo.scala
lrytz: How about moving the `print` method into a subclass where `symbolTable` is a `Global`? Sketch:

```scala
trait SymTab {
  val stats = new {
    val symTab: SymTab.this.type = SymTab.this
  } with SymTabStats
}

class Global extends SymTab {
  class CompilationUnit

  override val stats = new {
    val symTab: Global.this.type = Global.this
  } with GlobalStats

  stats.print(Iterator(new CompilationUnit))
}

trait SymTabStats {
  val symTab: SymTab
}

trait GlobalStats extends SymTabStats {
  val symTab: Global
  import symTab._

  def print(units: Iterator[CompilationUnit]) = println(units)
}
```

lrytz:
> global qs with all the quantities and I need to change that. Unless there's a creative way to do it, that's an extremely breaking change

In what sense is it breaking?

jvican:
It is breaking in that there are some stats that need to be shut down (e.g. `IOStats` that are used in `AbstractFile` because the call-sites have no connection to neither `Global` nor `SymbolTable`).
  
Also, I think I'm confused about the forwards and backwards binary compatible contract of Scalac. If we turn an `object` into a `trait` in Scala Reflect, isn't that a breaking binary change? A Scala compiler 2.12.2 that wants to use Scala reflect 2.12.4 will fail to link, no? Or is this not a supported use case? If so, in which cases do changes break bincompat in Scala Reflect?

src/reflect/scala/reflect/internal/util/StatisticsInfo.scala
jvican: I will try this. It's a good suggestion.

jvican:
Last commit turns the mother of all statistics, `scala.reflect.internal.util.Statistics`, from an object to a trait. This is a required change to initialize `qs` per `Global` and avoid conflicts.
  
Note that because of my change, bincompat is broken again because of the removal of `IOStats`. I have another question re bincompat for the Scala team. From the commit message:
  
> it is not clear to me whether this is a legit bincompat error since `IOStats` is `private[io]`, and I would expect the bincompat contract of Scalac to allow these "private" entities to change even if they could be accessible by third parties using the package `scala.reflect.io`.
  
Having said that and given the nature of this change, is it possible to whitelist it?

jvican:
Last commit turns the mother of all statistics, `scala.reflect.internal.util.Statistics`, from an object to a trait. This is a required change to initialize `qs` per `Global` and avoid conflicts.
  
Note that because of my change, bincompat is broken again because of the removal of `IOStats`. I have another question re bincompat for the Scala team. From the commit message:
  
> it is not clear to me whether this is a legit bincompat error since `IOStats` is `private[io]`, and I would expect the bincompat contract of Scalac to allow these "private" entities to change even if they could be accessible by third parties using the package `scala.reflect.io`.
  
Having said that and given the nature of this change, is it possible to whitelist it?

lrytz:
> some stats need to be shut down (e.g. IOStats that are used in AbstractFile because the call-sites have no connection to neither Global nor SymbolTable).

How about this hack: store the `statistics` instance in a thread local that's initialized when the `Global` is created? Or do you think these stats are not all that relevat anyway?

> If we turn an object into a trait in Scala Reflect, isn't that a breaking binary change

It is, but your change is in `scala.reflect.internal` which is whitelisted (https://github.com/scala/scala/blob/2.12.x/src/reflect/mima-filters/2.12.0.backwards.excludes#L1)

> bincompat is broken again because of the removal of IOStats

As you suggest, this can be whitelisted as it's `private[io]`

src/compiler/scala/tools/nsc/backend/jvm/analysis/AliasingFrame.scala
```diff
@@ -389,14 +389,17 @@ class AliasingFrame[V <: Value](nLocals: Int, nStack: Int) extends Frame[V](nLoc
...
-  val timer2 = newTimer("t2", "jvm")
-  val timer3 = newTimer("t3", "jvm")
-  val timers = List(timer1, timer2, timer3)
-  def reset(): Unit = for (t <- timers) { t.nanos = 0; t.timings = 0 }
+//  import scala.reflect.internal.util.Statistics._
+//  val timer1 = newTimer("t1", "jvm")
+//  val timer2 = newTimer("t2", "jvm")
+//  val timer3 = newTimer("t3", "jvm")
+//  val timers = List(timer1, timer2, timer3)
+//  def reset(): Unit = for (t <- timers) { t.nanos = 0; t.timings = 0 }
```
lrytz: I'm happy for you to just remove the `AliasingFrame` object entirely.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -159,16 +159,18 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
-                    with ImplicitsStats
-                    with MacrosStats
-                    with BackendStats
+  import scala.tools.nsc.transform.patmat.PatternMatchingStats
+  trait GlobalStats extends ReflectStats
+                       with TypersStats
+                       with ImplicitsStats
+                       with MacrosStats
+                       with BackendStats
+                       with PatternMatchingStats
```
lrytz: How do you feel about removing all those traits and just moving the stats into the `GlobalStats` / `ReflectStats`? I don't see much value in having these definitions spread around. Now that all stats are mixed into a single class, the definition site no longer acts as a namespace.

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -249,7 +249,7 @@ quant)
     }
   }
 
-  private val qs = new mutable.HashMap[String, Quantity]
+  private final val qs = new mutable.HashMap[String, Quantity]
```
lrytz: that's redundant

src/compiler/scala/tools/nsc/Global.scala
jvican: I kind of like the way it is right now not because they have their own namespace but because they are located right next to the place they're used -- so if I'm interested in macro statistics I can just grep through the file and get to their definition.
  
But if you think this is cumbersome, I have no problem to move all of them to those concrete trait. I'm really not a fan of the cake.

src/compiler/scala/tools/nsc/backend/jvm/analysis/AliasingFrame.scala
jvican: Cool, will remove it.

src/reflect/scala/reflect/internal/util/Statistics.scala
jvican: Didn't realize it is :smile:.

jvican:
> It is, but your change is in scala.reflect.internal which is whitelisted (https://github.com/scala/scala/blob/2.12.x/src/reflect/mima-filters/2.12.0.backwards.excludes#L1)
  
That makes a lot of sense. Wasn't sure that `internal` is whitelisted in Scala Reflect!
  
> How about this hack: store the statistics instance in a thread local that's initialized when the Global is created? Or do you think these stats are not all that relevat anyway?
  
I'm not sure these stats are really relevant, but I think that providing these these counters could in theory shed some light on the overhead of I/O work of the compiler, so I'll give your suggestion a go. The AB proposal asks for some of this data, so I think it's good to provide it anyway.

src/compiler/scala/tools/nsc/Global.scala
lrytz: I think it would make the whole thing easier to read, but I don't have very strong feelings about it either.

jvican: I propose to keep it the way it is and if other person complains I'll put it together. Personally, I don't find it hard to read, I just dislike the self-type dependencies... In any case, my last commit removed `StatisticsInfo`, so the last version is cleaner.

jvican:
All feedback is addressed and breaking change is whitelisted. Complexity when mixing in the trait has been removed in the last commit -- now it's clearer to read what's going on.
  
I think this PR is up for a final review, from https://github.com/scala/scala/pull/6051/commits/08cd6c18aa290458efde61010e680636b18b48a2 to https://github.com/scala/scala/pull/6051/commits/e787703fd70e4a864a8cf416e7e861ceb2f79632. Let's hope the dependent PR can be reviewed and merged first so that I can rebase this one and make it easier to review.

src/reflect/scala/reflect/internal/BaseTypeSeqs.scala
```diff
@@ -171,7 +171,7 @@ trait BaseTypeSeqs {
   /** A marker object for a base type sequence that's no yet computed.
    *  used to catch inheritance cycles
    */
-  val undetBaseTypeSeq: BaseTypeSeq = newBaseTypeSeq(List(), Array())
+  lazy val undetBaseTypeSeq: BaseTypeSeq = newBaseTypeSeq(List(), Array())
```
jvican, line 174: This is necessary because it is invoked in the initializer otherwise, calling `newBaseTypeSeq` which depends on `statistics`. When this class is initialized, `statistics` is null, so making it `lazy` sidesteps the `NullPointerException` that we would receive otherwise.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -158,19 +158,19 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
-  import scala.reflect.internal.util.StatisticsInfo
+  import scala.reflect.internal.util.Statistics
   import scala.tools.nsc.transform.patmat.PatternMatchingStats
   trait GlobalStats extends ReflectStats
                        with TypersStats
                        with ImplicitsStats
                        with MacrosStats
                        with BackendStats
-                       with PatternMatchingStats
+                       with PatternMatchingStats { self: Statistics => }
```
lrytz: just wondering, why is the self type here useful?

jvican:
Okay, I've squashed the commits related to this PR. I've left 4 independent PRs that summarize the changes in conceptual chunks, so that it's clear for future archeologists the order of the changes and the why (explained in the commit messages).
  
Here's a summary of what this PR does:
1. It makes an overhaul of the previous statistics infrastructure that was dependent on global objects shared by different instances of `Global` (making the stats unusable in most of the cases).
2. It enables `IOStats` with a special trick that uses `ThreadLocal` to share instances of quantities (timers/counters) as suggested by @lrytz.
3. It modifies the statistics framework to listen to changes in the mutable settings and thus enable/disable according to what the driver of the compiler does. This allows to run statistics in controlled and interactive environments, enabling and disabling several times on the go.
4. It deduplicates calls to `canEnable` that could be hindering optimizations or affecting the runtime.
  
This is ready for review. While this is being reviewed, I'm proceeding to benchmark this to compare results with https://github.com/scala/scala/pull/6034.

src/reflect/mima-filters/2.12.0.forwards.excludes
```diff
@@ -8,6 +8,7 @@ ProblemFilters.exclude[MissingClassProblem]("scala.reflect.io.PlainNioFile")
 
 ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.runtime.SynchronizedOps.newMappedBaseTypeSeq")
 ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.runtime.JavaUniverse.newMappedBaseTypeSeq")
+ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.runtime.JavaUniverse.statistics")
```
jvican, line 11: Whitelisted because it's `protected`. Same case as `newMappedBaseTypeSeq` which is above.

src/reflect/scala/reflect/runtime/JavaUniverseForce.scala
```diff
@@ -123,6 +123,7 @@ trait JavaUniverseForce { self: runtime.JavaUniverse  =>
     this.termNames
     this.nme
     this.sn
+    this.undetBaseTypeSeq
```
jvican, line 126: This was triggered by making `undetBaseTypeSeq` a lazy val. Explanation of why is above.

lrytz:
This PR needs a rebase on 2.12.x, there's a merge conflict in GenBCode.

lrytz:
For the commit messages (f9d88d10fd3837b8198d3735411e3818bbcb1075), don't include "Fix ..." in the title, but at the end of the message (https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#bug-fix).

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -1557,6 +1571,26 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
       }
       if (!pclazz.isRoot) resetPackageClass(pclazz.owner)
     }
+
+    import statistics.{treeNodeCount, nodeByType, retainedCount, retainedByType}
+    private final val ParserStats = Seq(treeNodeCount, nodeByType, retainedCount, retainedByType)
```
lrytz: Nits:
  - make it lower case
  - `final` is redundant
  - scope the import to the val

```
private val parserStats = {
  import statistics._
  Seq(treeNodeCount, nodeByType, retainedCount, retainedByType)
}
```

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
```diff
@@ -25,7 +26,10 @@ abstract class SymbolLoaders {
   val platform: backend.Platform {
     val symbolTable: SymbolLoaders.this.symbolTable.type
   }
+
   import symbolTable._
+  val statistics: Statistics with ReflectStats
```
lrytz: why is this needed? doesn't `symbolTable.statistics` work?

src/reflect/scala/reflect/internal/SymbolTable.scala
```diff
@@ -51,6 +51,16 @@ abstract class SymbolTable extends macros.Universe
...
   val gen = new InternalTreeGen { val global: SymbolTable.this.type = SymbolTable.this }
 
+  trait ReflectStats extends BaseTypeSeqsStats
+                        with TypesStats
+                        with SymbolTableStats
+                        with TreesStats
+                        with SymbolsStats { self: Statistics => }
+
+  /** Some statistics (normally disabled) set with -Ystatistics */
+  // This is a val so that it can be overridden by `Global.scala`
```
lrytz: nit: since it's an abstract val, i'd say it's not overridden, but implemented in subclasses

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -158,10 +158,19 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+  import scala.tools.nsc.transform.patmat.PatternMatchingStats
+  trait GlobalStats extends ReflectStats
+                       with TypersStats
+                       with ImplicitsStats
+                       with MacrosStats
+                       with BackendStats
+                       with PatternMatchingStats { self: Statistics => }
+
+  /** Redefine statistics to include all known global + reflect stats. */
+  override object statistics extends Statistics(Global.this, settings) with GlobalStats
```
lrytz: i think we usually don't put the `override` modifier when implementing an abstract member (here, and also in other places in this commit).

src/reflect/mima-filters/2.12.0.forwards.excludes
lrytz: In theory this breaks compatibility for external subclasses of `JavaUniverse`, but I guess we can accept it? @retronym introduced the similar cases in https://github.com/scala/scala/commit/777a0e5ae3.

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -17,9 +17,10 @@ import Variance._
 
 trait Symbols extends api.Symbols { self: SymbolTable =>
   import definitions._
-  import SymbolsStats._
+  import statistics._
 
   protected var ids = 0
+  def getSymbolId: Int = ids
```
lrytz: maybe something like `getCurrentSymbolIdCount`? I think something more lengthy is good here because it says "careful, you should know what you're doing when using this method".

src/reflect/scala/reflect/internal/settings/MutableSettings.scala
```diff
@@ -56,6 +56,11 @@ abstract class MutableSettings extends AbsSettings {
   def YpartialUnification: BooleanSetting
   def Yvirtpatmat: BooleanSetting
 
+  // Define them returning a `Boolean` to avoid breaking bincompat change
+  // TODO: Add these fields typed as `BooleanSetting` for 2.13.x
```
lrytz, line 60: could you prepare a follow-up PR in 2.13.x once this is merged?

src/reflect/scala/reflect/runtime/JavaUniverse.scala
```diff
@@ -5,6 +5,7 @@ package runtime
 import scala.reflect.internal.{TreeInfo, SomePhase}
 import scala.reflect.internal.{SymbolTable => InternalSymbolTable}
 import scala.reflect.runtime.{SymbolTable => RuntimeSymbolTable}
+import scala.reflect.internal.util.Statistics
```
lrytz, line 8: seems unused

build.sbt
```diff
@@ -365,6 +365,7 @@ lazy val reflect = configureAsSubproject(project)
   .settings(
     name := "scala-reflect",
     description := "Scala Reflection Library",
+    compileOrder := CompileOrder.Mixed,
```
lrytz: this is the default, no? so no need to set it.

jvican:
> This PR needs a rebase on 2.12.x, there's a merge conflict in GenBCode.
  
Yes, I'm waiting on the other PR to be merged so that I rebase this one. Otherwise, if I rebase this one the CI will trigger a build for every commit, and that could block other PR's validation for a while.
  
> For the commit messages (f9d88d1), don't include "Fix ..." in the title, but at the end of the message (https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#bug-fix).
  
Does GitHub close the issue that links if the "Fix ..." is at the end of the commit message instead of the title?

src/reflect/scala/reflect/runtime/JavaUniverse.scala
```diff
@@ -4,6 +4,7 @@ package runtime
 
 import scala.reflect.internal.{TreeInfo, SomePhase}
 import scala.reflect.internal.{SymbolTable => InternalSymbolTable}
+import scala.reflect.internal.util.Statistics
```
lrytz: one more unused import?

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
```diff
@@ -0,0 +1,39 @@
+package scala.reflect.internal.util;
+
+import scala.reflect.internal.util.Statistics;
+
+public class ThreadLocalStats {
```
lrytz: Why is this class defined in Java?

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final ThreadLocal<Statistics> statisticsLocal = new ThreadLocal<Statistics>();
+
+    // We can't use `Statistics.Counter` because of https://github.com/scala/bug/issues/10490
+    public static final ThreadLocal<Object> fileExistsCountLocal = new ThreadLocal<Object>();
+    public static final ThreadLocal<Object> fileIsFileCountLocal = new ThreadLocal<Object>();
+    public static final ThreadLocal<Object> fileIsDirectoryCountLocal = new ThreadLocal<Object>();
+
+    // Cache statistics when they are read from thread local for the first time
+    public static Statistics cachedStatistics = null;
```
lrytz: That seems to cause races when there are multiple compilers running in parallel.

src/compiler/scala/tools/nsc/backend/jvm/GenBCode.scala
```diff
@@ -329,17 +329,17 @@ abstract class GenBCode extends BCodeSyncAndTry {
      *
      */
     override def run() {
-      val bcodeStart = statistics.startTimer(bcodeTimer)
+      val bcodeStart = if (statistics.canEnable) statistics.startTimer(bcodeTimer) else null
```
lrytz: It seems to me that it would make more sense to do the opposite: make sure that all operations in `Statistics` are guarded by `canEnable` checks, and remove all the existing `canEnable` checks at the callsites. It's too easy to forget adding the check when introducing a new metric.

lrytz (commented):
Thanks for rebasing and cleaning up! Nice work.

lrytz:
> Yes, I'm waiting on the other PR to be merged so that I rebase this one

Sure, no problem

> Does GitHub close the issue that links if the "Fix ..." is at the end of the commit message instead of the title?

Yes

build.sbt
jvican: I think it is not, the build defaults to `JavaThenScala`.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
jvican: Because of statistcs :smile:.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
jvican: Good point, didn't realise this is the case. I'll think of a way to fix this.

src/reflect/scala/reflect/internal/SymbolTable.scala
jvican: Yeah, this comment is outdated.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
lrytz: I don't understand..

jvican: It's a Java class so that I can define `static` members. They are necessary so that the same instance is accessible from both `IOStats` and the `AbstractFile` and `Path` definitions.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
jvican: This is required because of a clash of path dependent types betwen `SymbolTable` and `Global`, IIRC.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
lrytz: Wouldn't a Scala `object` work?

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
lrytz: Would be nice to get rid of it if possible.. If you don't manage to convince the compiler that two paths are the same, you could also do a cast.

jvican: I can do a cast. I thought this solution would be nicer.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
jvican: Yes, but wouldn't the `static` members be faster?

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
jvican: I may need to remove `IOStats` altogether for this to be friendly to multiple compilers running in paralle.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
lrytz: Answering that question would require a benchmark. But by default we should write our code in Scala, unless a performance issue can be measured.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
lrytz: I would start by removing the cache here and see if there's a performance problem going through the `ThreadLocal`. Compared to IO, accessing a thread local is probably inexpensive.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
jvican: In my previous PR, the difference between accessing a `static` member and a normal `object` member was quite drastic re performance, this is why here I defaulted to `static`.

lrytz: Can you point me at the benchmark?

retronym: Maybe you're thinking about the special case of `MethodHandle`-s stored in final static vals. Invocations over these are inlinable by JIT.

jvican: Benchmarks [here](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=311a576030|bb3e997faa|6d006a5461|63953e1388|5254231683|b5a3aee4db&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1502073067834&to=1502882030576).
  
My first attempt for storing method handles [was modeled with an object](https://github.com/scalacenter/scala/blob/dbb772051fdc4b05af6f36c8f4501e6d96cd705b/src/reflect/scala/reflect/internal/util/Statistics.scala#L8-L11) and that made things slower, whereas using a static defined in Java made it significantly faster. This may be specific to method handles, though. However, I decided to follow the same logic in here.
  
I'll remove the cache and measure this concrete scenario.

jvican: > Maybe you're thinking about the special case of MethodHandle-s stored in final static vals. Invocations over these are inlinable by JIT.
  
Yeah. That was it.

src/reflect/scala/reflect/internal/util/ThreadLocalStats.java
retronym: I would advise to just remove these counters altogether. I think I added them to demonstrate that some optimizations in classpath scanning reduced the IO operations. It's easy enough to temporarily add a global counter for such things during future optimizations in this area.

jvican: Okay, I'll remove them then.

src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala
jvican: Oh no, I remember now why this was done. `statistics` in `symbolTable` is `protected`, so it's not accessible here. That's why the trait that mixes in `SymbolLoaders` and `SymbolTable` defines `statistics` using the knowledge from the inherited `SymbolTable`. As far as I see it, there's no other solution than this because abstract members cannot be `private[pkg]`.

lrytz: Let's make `statistics` public then - or is there a good reason to keep it protected?

jvican: It's protected so that this addition is binary compatible, otherwise it wouldn't because it leaks to public APIs. Unless you see a way to make bincompat, I don't see another solution.

lrytz: `scala.reflect.internal` is excluded from binary compatibility (https://github.com/scala/scala/blob/2.12.x/src/reflect/mima-filters/2.12.0.forwards.excludes#L1). In any case, Scala's `protected` translates to `public` in bytecode.

jvican: My first iteration made it public and I got a MiMa error somewhere in a public interface, but I'll try again to double-check -- I may have done something wrong.

src/compiler/scala/tools/nsc/Global.scala
jvican: Because:
```
[error] [E1] src/compiler/scala/tools/nsc/Global.scala
[error]      illegal inheritance;
[error]       self-type Global.this.GlobalStats does not conform to Global.this.ReflectStats's selftype Global.this.ReflectStats with scala.reflect.internal.util.Statistics
[error]      L164:  trait GlobalStats extends ReflectStats
[error]      L164:                            ^
```
  
and 5 more errors like this one.
  
It wasn't obvious to me this was required. I would have expected the trait without the self-type annotation to compile...

jvican:
This is ready for review again. Commits have been rebased, feedback addressed and [benchmarks have been scheduled](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=8a92ebdce5|1dff6d12a1|c91aa41a74&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=now-30d&to=now).

src/compiler/scala/tools/nsc/Global.scala
dragos: If the self-type is needed in `ReflectStats`, you could just make `GlobalStats extends Statistics`.

lrytz: Right, but that would require adding and forwarding the class parameters. I'm happy to keep it as it is right now.

lrytz:
I haven't finished the review yet, but gtg right now. In the meantime, here are some minor cleanups: https://github.com/lrytz/scala/tree/pr6051.

There are plenty of `if (statistics.canEnable) statistics.startTimer` calls, but `startTimer` also checks `canEnable`. Could we simplify that?

Could you update the commit message where it's out of date? E.g., "It enables IOStats with a special trick that uses ThreadLocal".

jvican:
> There are plenty of if (statistics.canEnable) statistics.startTimer calls, but startTimer also checks canEnable. Could we simplify that?
  
I know, this PR did this before, but I removed it because I saw some weird things re performance. I can add this, but we will need to check that this simplification is actually meaningful with benchmarks. That can slow down the process to get this merged. I'll try to do it tonight.
  
> Could you update the commit message where it's out of date? E.g., "It enables IOStats with a special trick that uses ThreadLocal".
  
Yes.

lrytz:
> I know, this PR did this before, but I removed it because I saw some weird things re performance.

I'm happy to have this in a follow-up PR, it doesn't need to hold up this one.

lrytz:
> benchmarks have been scheduled

If I understand correctly, there's a significant slowdown. It seems grafana is slightly confused about the two commits, it cannot tell them apart, probably because they have the same timestamp due to a rebase.

lrytz:
```
$ git log -n 2 c91aa41a74 --pretty=format:"%h %ct"
c91aa41a74 1506337846
1dff6d12a1 1506337846
```

src/compiler/scala/tools/nsc/backend/jvm/PostProcessor.scala
```diff
@@ -13,7 +13,7 @@ import scala.tools.nsc.backend.jvm.opt._
  * Implements late stages of the backend that don't depend on a Global instance, i.e.,
  * optimizations, post-processing and classfile serialization and writing.
  */
-abstract class PostProcessor extends PerRunInit {
+abstract class PostProcessor(statistics: Statistics with BackendStats) extends PerRunInit {
```
lrytz, line 16: I realize now that there's a problem here with thread safety, I assume `statistics` is not thread safe. So we'll have to come back to this when we parallelize classfile writing (and possibly other parts of the backend). But this PR doesn't introduce a new problem, before we were just accessing the static `Statistics` object, which was not thread safe either. cc @mkeskells.

lrytz (commented):
I went through the diff again, I didn't see any other problems than the ones I pointed out earlier today. So this is almost good to go, but we need to figure out the performance impact.

jvican:
I'm trying different things to massage the performance impact. From what I can see in the benchmark results, the benchmark machine is not quite stable.

jvican:
Removing `canEnable` in the call-sites has a significant hit on performance, as per https://github.com/scala/scala/pull/6051/commits/249f9e242f96642b64d6f7b25100a4d21647aa23. I guess this is due to the fact that `canEnable` is not inlined by the JVM, and we always pay the cost of the call to the statistics instance.
  
I've reverted the change and added a new commit that removes `canEnable` in the definition sites, to check if there's a performance degradation. In theory, this should give the same or faster running time.
  
**EDIT**: [Updated link to benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=8a92ebdce5|1dff6d12a1|c91aa41a74|c818f40|249f9e2|25c0d0d&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=now-30d&to=now).

jvican:
I think the main performance hit here comes from the fact that before `Statistics.canEnable` was static because `Statistics` was an object, but now it's not.
  
I don't think this can be addressed since it's fundamental to have statistics being dependent on the global instance (as motivated in the README), but I think that removing the `canEnable` calls in the definition site may yield some performance benefit.
  
In the next hours, I'll update this commit with a detailed table and a reading of the performance results from Grafana.

jvican:
It seems the candidate to be merged is https://github.com/scala/scala/pull/6051/commits/c818f40cb7976fbd456f4eb4c56b8b91981bc345. This is the performance degradation recorded by Grafana:
  
| better-files | scala | scalap | vector |
| ------------- | ------- | -------- | -------- |
| +2.4% | +3.1% | +2.6% | +2.0% |

[Data source](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=8a92ebdce5|1dff6d12a1|c91aa41a74|c818f40|249f9e2|25c0d0d&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1505924607857&to=1506530529968).
  
I find it difficult to reason about the performance impact since the machine does not provide consistent results across different repos and same commits. Some of the commits I would expect to behave exactly the same and yet they provide sometimes different results (an example of this is https://github.com/scala/scala/pull/6051/commits/25c0d0d7666582a083c1417706b5c34289baa01b, which should not have any impact unless statistics are enabled). I don't think it's a big deal, it just hints that the machine is not 100% stable.
  
@lrytz I did the experiment you asked me, and the result is that it's significantly slower. See the commit  https://github.com/scala/scala/pull/6051/commits/27973c5c45e81df30c6e423422b0dadc9c72d212 and the [results](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=8a92ebdce5|1dff6d12a1|c91aa41a74|c818f40|249f9e2|25c0d0d&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1505924607857&to=1506530529968). I provide an hypothesis for this behaviour in one of my commits above.

lrytz:
2-3% is significant.. Not sure what to do here. For comparison, could you benchmark a version where all calls to statistics are commented out?

retronym:
Perhaps we could still have a static canenable that is the disjunction of
the canenable of the per global statistics instances.

jvican:
> Perhaps we could still have a static canenable that is the disjunction of the canenable of the per global statistics instances.
  
I can quickly whip up something like this, but note that there's a clear problem here: if a `Global` instance does enable statistics, they will be enabled for all the global instances, meaning that in a multi-project environment like sbt the enabling of statistics in one project will make other projects compile with statistics enabled too.
  
> 2-3% is significant.
  
It is significant, but I'm afraid it's the price to pay for having a working statistics infrastructure. Having statistics enabled globally for all globals makes statistics run into thread-safety issues, as we've experienced in the past.
  
Please, let me know your position on this, I can quickly act on this.

retronym:
> I can quickly whip up something like this, but note that there's a clear problem here: if a Global instance does enable statistics, they will be enabled for all the global instances, meaning that in a multi-project environment like sbt the enabling of statistics in one project will make other projects compile with statistics enabled too.

My idea was to have a fast, static check (using the `AlmostFinalField` technique) that means `someGlobalHasEnabledStatistics`. If that is true, we can consult the `canEnable` of the `Statistics` instance for the current `Global`.

The overwhelmingly common case is that no statistics will ever be enabled, and we should strive to make this run as fast as if all the stats collection code was commented out.

retronym:
Sorry, meant to hit "comment", not "close and comment"

jvican:
My last commit massages the performance of this change, and as [reported by the benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=8a92ebdce5|1dff6d12a1|c91aa41a74|c818f40|249f9e2|25c0d0d|66b5fb1|1e09de17a3&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1505939731620&to=1506676767085) the previous performance before this change has been restored. In some benchmarks, it seems it's *slightly* faster but not sure if this is due to machine unstability.
  
The major impact in performance is due to the direct use of a static call in the if checks, instead of `statistics.canEnable`. This static is just a flag that tells us if some global instance has enabled statistics. If it has, then we double-check if statistics have been enabled for that concrete global. Therefore, we only pay the price of checking the non-static fields (+2/3%) only when one global has enabled statistics.
  
@retronym @lrytz Can you review my last commit?

lrytz (approved):
Great! Really glad that worked out.

jvican:
Thank you @lrytz @retronym.

adriaanm:
this broke scalameta (https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2002/consoleFull):

code patterns
```
import TypersStats._

if (Statistics.canEnable) Statistics.incCounter
if (Statistics.canEnable) Statistics.startTimer
if (Statistics.canEnable) Statistics.stopTimer
```

counters/timers
```
appStart
failedApplyNanos
failedOpEqNanos
opeqStart
start
typedApplyCount
typedSelectCount
```

Lines:
```
scala/meta/internal/SemanticdbAnalyzer.scala:18 
scala/meta/internal/SemanticdbAnalyzer.scala:86 
scala/meta/internal/SemanticdbAnalyzer.scala:86 
scala/meta/internal/SemanticdbAnalyzer.scala:89 
scala/meta/internal/SemanticdbAnalyzer.scala:89 
 scala/meta/internal/SemanticdbAnalyzer.scala:150
 scala/meta/internal/SemanticdbAnalyzer.scala:150
 scala/meta/internal/SemanticdbAnalyzer.scala:151
 scala/meta/internal/SemanticdbAnalyzer.scala:151
 scala/meta/internal/SemanticdbAnalyzer.scala:181
 scala/meta/internal/SemanticdbAnalyzer.scala:181
 scala/meta/internal/SemanticdbAnalyzer.scala:193
 scala/meta/internal/SemanticdbAnalyzer.scala:193
 scala/meta/internal/SemanticdbAnalyzer.scala:201
 scala/meta/internal/SemanticdbAnalyzer.scala:201
 scala/meta/internal/SemanticdbAnalyzer.scala:212
 scala/meta/internal/SemanticdbAnalyzer.scala:212
 scala/meta/internal/SemanticdbAnalyzer.scala:520
 scala/meta/internal/SemanticdbAnalyzer.scala:520

```

adriaanm:
Is there a way to provide a source compat shim via `import TypersStats._`? I have not looked at the details, just a quick suggestion.

jvican:
No, it's not possible since Scalameta is accessing `Statistics` directly, which doesn't exist anymore. Compiler plugins will need special cross-compiling for 2.12.4 onwards (i.e. `src/main/scala-2.12.4/`).

adriaanm:
Ok, I am fiddling with a patch for our fork at scalacommunitybuild/scalameta branch community-build-2.12:


```diff
➜  scalameta git:(community-build-2.12) ✗ diff -u scalameta/semanticdb-scalac/src/main/scala-2.12.{3,4}/scala/meta/internal/SemanticdbAnalyzer.scala
--- scalameta/semanticdb-scalac/src/main/scala-2.12.3/scala/meta/internal/SemanticdbAnalyzer.scala	2017-09-29 10:59:27.000000000 -0700
+++ scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala	2017-09-29 11:05:10.000000000 -0700
@@ -9,6 +9,8 @@
 import scala.reflect.internal.Flags._
 import scala.meta.internal.ReflectionToolkit
 
+import scala.reflect.internal.util.StatisticsStatics
+
 // NOTE: grep for "scalac deviation" to see what exactly has been changed
 // I had to copy/paste a lot of code from Namers and Typers, but only a small percentage is actually relevant
 
@@ -16,6 +18,7 @@
   import global._
   import definitions._
   import TypersStats._
+  import statistics.{appStart, failedApplyNanos, failedOpEqNanos, opeqStart, start, typedApplyCount, typedSelectCount}
 
   override def newNamer(context: Context) = new SemanticdbNamer(context)
   class SemanticdbNamer(context0: Context) extends Namer(context0) {
@@ -41,6 +44,8 @@
     import TyperErrorGen._
     import typeDebug.ptTree
     import runDefinitions._
+    
+    import StatisticsStatics.areSomeColdStatsEnabled
 
     // TODO: No idea whether this is gonna work well in the batch compilation mode.
     override def canTranslateEmptyListToNil = false
@@ -83,10 +88,10 @@
        * insert an implicit conversion.
        */
       def tryTypedApply(fun: Tree, args: List[Tree]): Tree = {
-        val start = if (Statistics.canEnable) Statistics.startTimer(failedApplyNanos) else null
+        val start = if (areSomeColdStatsEnabled) statistics.startTimer(failedApplyNanos) else null
 
         def onError(typeErrors: Seq[AbsTypeError], warnings: Seq[(Position, String)]): Tree = {
-          if (Statistics.canEnable) Statistics.stopTimer(failedApplyNanos, start)
+          if (areSomeColdStatsEnabled) statistics.stopTimer(failedApplyNanos, start)
 
           // If the problem is with raw types, convert to existentials and try again.
           // See #4712 for a case where this situation arises,
@@ -147,8 +152,8 @@
         // TODO: replace `fun.symbol.isStable` by `treeInfo.isStableIdentifierPattern(fun)`
         val stableApplication = (fun.symbol ne null) && fun.symbol.isMethod && fun.symbol.isStable
         val funpt = if (mode.inPatternMode) pt else WildcardType
-        val appStart = if (Statistics.canEnable) Statistics.startTimer(failedApplyNanos) else null
-        val opeqStart = if (Statistics.canEnable) Statistics.startTimer(failedOpEqNanos) else null
+        val appStart = if (areSomeColdStatsEnabled) statistics.startTimer(failedApplyNanos) else null
+        val opeqStart = if (areSomeColdStatsEnabled) statistics.startTimer(failedOpEqNanos) else null
 
         def isConversionCandidate(qual: Tree, name: Name): Boolean =
           !mode.inPatternMode && nme.isOpAssignmentName(TermName(name.decode)) && !qual.exists(_.isErroneous)
@@ -178,7 +183,7 @@
           case Select(qual, name) if isConversionCandidate(qual, name) =>
             val qual1 = typedQualifier(qual)
             if (treeInfo.isVariableOrGetter(qual1)) {
-              if (Statistics.canEnable) Statistics.stopTimer(failedOpEqNanos, opeqStart)
+              if (areSomeColdStatsEnabled) statistics.stopTimer(failedOpEqNanos, opeqStart)
               val erred = qual1.exists(_.isErroneous) || args.exists(_.isErroneous)
               if (erred) reportError(error) else {
                 val convo = convertToAssignment(fun, qual1, name, args)
@@ -190,7 +195,7 @@
                 }
               }
             } else {
-              if (Statistics.canEnable) Statistics.stopTimer(failedApplyNanos, appStart)
+              if (areSomeColdStatsEnabled) statistics.stopTimer(failedApplyNanos, appStart)
               val Apply(Select(qual2, _), args2) = tree
               val erred = qual2.exists(_.isErroneous) || args2.exists(_.isErroneous)
               reportError {
@@ -198,7 +203,7 @@
               }
             }
           case _ =>
-            if (Statistics.canEnable) Statistics.stopTimer(failedApplyNanos, appStart)
+            if (areSomeColdStatsEnabled) statistics.stopTimer(failedApplyNanos, appStart)
             reportError(error)
         }
         val silentResult = silent(
@@ -209,7 +214,7 @@
         silentResult match {
           case SilentResultValue(fun1) =>
             val fun2 = if (stableApplication) stabilizeFun(fun1, mode, pt) else fun1
-            if (Statistics.canEnable) Statistics.incCounter(typedApplyCount)
+            if (areSomeColdStatsEnabled) statistics.incCounter(typedApplyCount)
             val noSecondTry = (
                  isPastTyper
               || context.inSecondTry
@@ -517,7 +522,7 @@
           // the qualifier type of a supercall constructor is its first parent class
           typedSelect(tree, typedSelectOrSuperQualifier(qual), nme.CONSTRUCTOR)
         case Select(qual, name) =>
-          if (Statistics.canEnable) Statistics.incCounter(typedSelectCount)
+          if (areSomeColdStatsEnabled) statistics.incCounter(typedSelectCount)
           val qualTyped = checkDead(typedQualifier(qual, mode))
           val qualStableOrError = (
             if (qualTyped.isErrorTyped || !name.isTypeName || treeInfo.admitsTypeSelection(qualTyped))
```

adriaanm:
https://github.com/adriaanm/scalameta/tree/scalameta-statistics pushed to https://github.com/scalacommunitybuild/scalameta/tree/community-build-2.12
https://github.com/scala/community-builds/commit/c12dccafdfbed09299b461e052c09281810268ef

retronym:
Hate to be the bearer of bad news, but our benchmarks have regressed. 
We only run the `scala` benchmark daily, so I can't say for sure if this PR, or something else in #6066  #6097 #6027 #6101 #6082 #6098 #6089 is to blame. I've scheduled more fine grained runs to try to isolate the problem.

image

lrytz:
Jorge did run the "scala" benchmark on the final solution of this PR (https://github.com/scala/scala/pull/6051#issuecomment-332814321) and there was no regression. 

Maybe it was in https://github.com/scala/scala/pull/6067? There Jorge also ran the benchmarks without regression (https://github.com/scala/scala/pull/6067#issuecomment-332949297), but then pushed a few more changes / added more commits.

retronym:
#6067 was merged after the regression, if I'm reading things correctly.

Last known good was 95142900feeefafd587313ebabb7521c00ce813b, first known bad is d41c97f13ea2348ec6f59c24ae098d83a96f30dc

lrytz:
OK, thanks! I was just thinking of PR I touched on :)

jvican:
@retronym The baseline of my benchmarks was 39301 for scala/scala, and the results for this PR yielded 39250. In the attached image, it seems that the regression has been from ~3700s, so it looks like this PR is not the responsible for the regression.

retronym:
@jvican Agreed, #6082 appears to be the point of regression.

retronym:
1e09de17a3473efb26db535a71f9ec8b03018ac2 ("Massage performance with statistics optimizations  …") appears to increase the overhead of having statistics enabled.

```
⚡ for extra in "-Ystatistics:_" ""; do for v in 2.12.4-bin-1e09de1-SNAPSHOT 2.12.4-bin-4eab2f7-SNAPSHOT; do echo $v $extra; sbt 'set scalaVersion in compilation := "'$v'"' 'hot -psource=scalap -f1 -wi 5 -i 3 -pextraArsgs='$extra | egrep 'HotScalacBenchmark.compile\s'; done; done

⚡ for extra in "-Ystatistics:_" ""; do for v in 2.12.4-bin-1e09de1-SNAPSHOT 2.12.4-bin-4eab2f7-SNAPSHOT; do echo $v $extra; sbt 'set scalaVersion in compilation := "'$v'"' 'hot -psource=scalap -f1 -wi 5 -i 3 -pextraArgs='$extra | egrep 'HotScalacBenchmark.compile\s'; done; done
2.12.4-bin-1e09de1-SNAPSHOT -Ystatistics:_
[info] HotScalacBenchmark.compile                          a8c43dc  -Ystatistics:_       false    scalap  sample   12  2939.508 ± 203.203  ms/op
2.12.4-bin-4eab2f7-SNAPSHOT -Ystatistics:_
[info] HotScalacBenchmark.compile                          a8c43dc  -Ystatistics:_       false    scalap  sample   33   951.249 ± 16.147  ms/op
2.12.4-bin-1e09de1-SNAPSHOT
2.12.4-bin-1e09de1-SNAPSHOT 
[info] HotScalacBenchmark.compile                          a8c43dc                    false    scalap  sample   36   858.492 ± 42.606  ms/op
2.12.4-bin-4eab2f7-SNAPSHOT 
[info] HotScalacBenchmark.compile                          a8c43dc                    false    scalap  sample   35   894.675 ± 69.345  ms/op
2.12.4-bin-1e09de1-SNAPSHOT

```

@jvican Any idea why? I'll keep digging into it.

jvican:
If I had to venture a guess, I would say that the fact that `statistics` is not final anymore makes the access to it more expensive; but honestly, I don't know why. The reason why `statistics` couldn't be final anymore was because it's overridden in both `JavaUniverse` and `Global`.
  
I'm happy to look into this over the weekend, let me know if you'd like me to.

retronym:
I've collected profiles in https://scala-ci.typesafe.com/view/scala-bench/job/compiler-benchmark/931/artifact/target/

It looks like JIT is working harder. The profile-async profile only collects after the 10x10s warmup, but the JIT threads [dominate](https://scala-ci.typesafe.com/view/scala-bench/job/compiler-benchmark/931/artifact/target/profile-async/flame-graph-cpu.svg) the profile.

I'm removing a few moving parts in https://github.com/retronym/scala/tree/perf/stats to see if I can pinpoint the problem.

retronym:
It's possible that I've bisected this incorrectly and am witnessing a problem that was was already fixed by 1c47b2ffad20b6f589255b3c935c8538fd714403 ("don't try to re-disabled the almost final value")

retronym:
Ha, the problem is actually redundantly pushing `true` into the almost final value.

I think this is the fix:

```diff
diff --git a/src/reflect/scala/reflect/internal/util/StatisticsStatics.java b/src/reflect/scala/reflect/internal/util/StatisticsStatics.java
index a7a2e02f71..3670af2058 100644
--- a/src/reflect/scala/reflect/internal/util/StatisticsStatics.java
+++ b/src/reflect/scala/reflect/internal/util/StatisticsStatics.java
@@ -50,3 +50,4 @@ public final class StatisticsStatics extends BooleanContainer {
   public static void enableColdStats() {
-    COLD_STATS.setValue(new TrueContainer());
+    if (!areSomeColdStatsEnabled())
+      COLD_STATS.setValue(new TrueContainer());
   }
@@ -58,3 +59,4 @@ public final class StatisticsStatics extends BooleanContainer {
   public static void enableHotStats() {
-    HOT_STATS.setValue(new TrueContainer());
+    if (!areSomeHotStatsEnabled())
+      HOT_STATS.setValue(new TrueContainer());
   }
```

This avoids invalidating all the call sites that have (JIT) inlined `areAnyColdStatsEnabled`. I'll submit the patch tomorrow.

jvican:
That makes sense — this is an oversight, I thought I had done this. Thanks for looking into this, I appreciate it.

PR #6052, opened 2017-08-30
Provide capacity for ProcessBuilder.lineStream

juliuszsompolski:
ProcessBuilder.lineStream will create a Stream[String] backed by a LinkedBlockingQueue with a default constructor that gives it a default capacity of Integer.MAX_VALUE.
If the produced stream is very large, and the consumer cannot keep up with consuming it, such a large buffer will cause OOMs.

Implement an option to specify buffer capacity, and reduce the default one to a safer value.

### DIFF
--- a/src/library/scala/sys/process/ProcessBuilder.scala
+++ b/src/library/scala/sys/process/ProcessBuilder.scala
@@ -173,6 +173,17 @@ trait ProcessBuilder extends Source with Sink {
     */
   def lineStream: Stream[String]
 
+  /** Starts the process represented by this builder.  The output is returned as
+   * a Stream that blocks when lines are not available but the process has not
+   * completed.
+   * The producer process will block if the given capacity of lines if filled
+   * without being consumed from the stream.
+   * Standard error is sent to the console.  If the process exits
+   * with a non-zero value, the Stream will provide all lines up to termination
+   * and then throw an exception.
+   */
+  def lineStream(capacity: Integer): Stream[String]
+
   /** Deprecated (renamed). Use `lineStream` instead. */
   @deprecated("use lineStream instead", "2.11.0")
   def lines: Stream[String] = lineStream
@@ -184,6 +195,17 @@ trait ProcessBuilder extends Source with Sink {
     * to termination and then throw an exception.
     */
   def lineStream(log: ProcessLogger): Stream[String]
+
+  /** Starts the process represented by this builder.  The output is returned as
+   * a Stream that blocks when lines are not available but the process has not
+   * completed.
+   * The producer process will block if the given capacity of lines if filled
+   * without being consumed from the stream.
+   * Standard error is sent to the provided ProcessLogger.  If the
+   * process exits with a non-zero value, the Stream will provide all lines up
+   * to termination and then throw an exception.
+   */
+  def lineStream(log: ProcessLogger, capacity: Integer): Stream[String]
 
   /** Deprecated (renamed).  Use `lineStream(log: ProcessLogger)` instead. */
   @deprecated("use lineStream instead", "2.11.0")
@@ -197,6 +219,17 @@ trait ProcessBuilder extends Source with Sink {
     */
   def lineStream_! : Stream[String]
 
+  /** Starts the process represented by this builder.  The output is returned as
+   * a Stream that blocks when lines are not available but the process has not
+   * completed.
+   * The producer process will block if the given capacity of lines if filled
+   * without being consumed from the stream.
+   * Standard error is sent to the console. If the process exits
+   * with a non-zero value, the Stream will provide all lines up to termination
+   * but will not throw an exception.
+   */
+  def lineStream_!(capacity: Integer): Stream[String]
+
   /** Deprecated (renamed).  Use `lineStream_!` instead. */
   @deprecated("use lineStream_! instead", "2.11.0")
   def lines_! : Stream[String] = lineStream_!
@@ -208,6 +241,17 @@ trait ProcessBuilder extends Source with Sink {
     * to termination but will not throw an exception.
     */
   def lineStream_!(log: ProcessLogger): Stream[String]
+
+  /** Starts the process represented by this builder.  The output is returned as
+   * a Stream that blocks when lines are not available but the process has not
+   * completed.
+   * The producer process will block if the given capacity of lines if filled
+   * without being consumed from the stream.
+   * Standard error is sent to the provided ProcessLogger. If the
+   * process exits with a non-zero value, the Stream will provide all lines up
+   * to termination but will not throw an exception.
+   */
+  def lineStream_!(log: ProcessLogger, capacity: Integer): Stream[String]
 
   /** Deprecated (renamed).  Use `lineStream_!(log: ProcessLogger)` instead. */
   @deprecated("use lineStream_! instead", "2.11.0")
--- a/src/library/scala/sys/process/BasicIO.scala
+++ b/src/library/scala/sys/process/BasicIO.scala
@@ -42,8 +42,8 @@ object BasicIO {
   )
 
   private[process] object Streamed {
-    def apply[T](nonzeroException: Boolean): Streamed[T] = {
-      val q = new LinkedBlockingQueue[Either[Int, T]]
+    def apply[T](nonzeroException: Boolean, capacity: Integer): Streamed[T] = {
+      val q = new LinkedBlockingQueue[Either[Int, T]](capacity)
       def next(): Stream[T] = q.take match {
         case Left(0)    => Stream.empty
         case Left(code) => if (nonzeroException) scala.sys.error("Nonzero exit code: " + code) else Stream.empty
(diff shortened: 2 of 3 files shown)


juliuszsompolski:
please sign the Scala CLA by clicking on 'Details' --> done

juliuszsompolski:
New library methods change binary:
```
[info] scala-library: found 14 potential binary incompatibilities while checking against org.scala-lang:scala-library:2.12.0  (filtered 20)
[error]  * method apply(Boolean)scala.sys.process.BasicIO#Streamed in object scala.sys.process.BasicIO#Streamed does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.BasicIO#Streamed.apply")
[error]  * abstract method lineStream_!(java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.sys.process.ProcessBuilder.lineStream_!")
[error]  * abstract method lineStream_!(scala.sys.process.ProcessLogger,java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.sys.process.ProcessBuilder.lineStream_!")
[error]  * abstract method lineStream(java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.sys.process.ProcessBuilder.lineStream")
[error]  * abstract method lineStream(scala.sys.process.ProcessLogger,java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.sys.process.ProcessBuilder.lineStream")
[error]  * method apply(Boolean,java.lang.Integer)scala.sys.process.BasicIO#Streamed in object scala.sys.process.BasicIO#Streamed does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.BasicIO#Streamed.apply")
[error]  * method lineStream_!(java.lang.Integer)scala.collection.immutable.Stream in class scala.sys.process.ProcessBuilderImpl#AbstractBuilder's type is different in other version, where it is (scala.sys.process.ProcessLogger)scala.collection.immutable.Stream instead of (java.lang.Integer)scala.collection.immutable.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("scala.sys.process.ProcessBuilderImpl#AbstractBuilder.lineStream_!")
[error]  * method lineStream_!(scala.sys.process.ProcessLogger,java.lang.Integer)scala.collection.immutable.Stream in class scala.sys.process.ProcessBuilderImpl#AbstractBuilder does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.ProcessBuilderImpl#AbstractBuilder.lineStream_!")
[error]  * method lineStream(java.lang.Integer)scala.collection.immutable.Stream in class scala.sys.process.ProcessBuilderImpl#AbstractBuilder's type is different in other version, where it is (scala.sys.process.ProcessLogger)scala.collection.immutable.Stream instead of (java.lang.Integer)scala.collection.immutable.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("scala.sys.process.ProcessBuilderImpl#AbstractBuilder.lineStream")
[error]  * method lineStream(scala.sys.process.ProcessLogger,java.lang.Integer)scala.collection.immutable.Stream in class scala.sys.process.ProcessBuilderImpl#AbstractBuilder does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.ProcessBuilderImpl#AbstractBuilder.lineStream")
[error]  * method lineStream_!(java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder's type is different in other version, where it is (scala.sys.process.ProcessLogger)scala.collection.immutable.Stream instead of (java.lang.Integer)scala.collection.immutable.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("scala.sys.process.ProcessBuilder.lineStream_!")
[error]  * abstract method lineStream_!(scala.sys.process.ProcessLogger,java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.ProcessBuilder.lineStream_!")
[error]  * method lineStream(java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder's type is different in other version, where it is (scala.sys.process.ProcessLogger)scala.collection.immutable.Stream instead of (java.lang.Integer)scala.collection.immutable.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("scala.sys.process.ProcessBuilder.lineStream")
[error]  * abstract method lineStream(scala.sys.process.ProcessLogger,java.lang.Integer)scala.collection.immutable.Stream in interface scala.sys.process.ProcessBuilder does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.ProcessBuilder.lineStream")
```
What are the guidelines to when new methods are allowed to be added to library?

som-snytt:
There's an issue for specifying encoding for this API implicitly. Possibly capacity could also be implicit; instead of a queue size, it could be a target consumption rate and falling below that throws.

IIRC you can't have a reactive stream where it tells the producer process to slow build.

hrhino:
@juliuszsompolski you can only add/remove methods between major versions; effectively, this would need to go into 2.13.x instead of 2.12.x. If you rebase it onto the HEAD of 2.13.x, you should be able to retarget the PR on github.

juliuszsompolski:
@hrhino Thanks! I rebased to 2.13.x

juliuszsompolski:
@som-snytt 
> IIRC you can't have a reactive stream where it tells the producer process to slow build.

It worked for me in my use case in https://github.com/databricks/spark-sql-perf/pull/115.
I had an issue with Spark being too slow to consume, parse & put to S3 data coming from TPCDS data generator, eventually leading to OOMs. After limiting the queue size, I see the dsdgen generating processes working at avg. 20% cpu, blocked when the queue becomes full.

From a Scala user perspective, I expected a Stream[String] returned by lineStream to be lazy for the consumer, and not overflow me with unconsumed data. The docs indeed mentions only that the consumer will block and wait if the buffer is empty, and doesn't mention blocking the producer, but I think it's reasonable to have a reasonably sized blocking buffer there.

A POSIX pipe in linux has a 64kB buffer by default, so I think that even my initial proposition of 64k lines that could lead to about several megabytes of buffering is too high, so I lowered it to 4k lines.

som-snytt:
The javadoc for `java.lang.Process` warns to read promptly or risk deadlock; I think it was on Windows that I saw that happen years ago. But I don't have a strong opinion here. They mentioned they're putting this code into a separate module, https://github.com/scala/bug/issues/10461

src/library/scala/sys/process/ProcessBuilder.scala
```diff
@@ -173,6 +173,17 @@ trait ProcessBuilder extends Source with Sink {
     */
   def lineStream: Stream[String]
 
+  /** Starts the process represented by this builder.  The output is returned as
+   * a Stream that blocks when lines are not available but the process has not
+   * completed.
+   * The producer process will block if the given capacity of lines if filled
+   * without being consumed from the stream.
```
lrytz, line 180: Can you provide more documentation what happens in this case? Is the child process suspended until the stream is being consumed? Or is the output buffered somewhere else? Or a mix of these?

juliuszsompolski: The thread in scala that consumes from the child will get blocked on inserting to LinkedBlockingQueue, until the application consumes the stream.

The behaviour of the child process is OS dependent, but similar anywhere.
On a POSIX system the child and parent process communicate through a PIPE, which on Linux has a 64kb buffer by default. If that buffer is full, the child process will be put to sleep by the kernel and made runnable again after the parent process consumption makes space in the buffer.
This is the standard behaviour of interprocess communication in any OS, so in my opinion it seems out of scope here.

lrytz: OK, thanks for the background!

lrytz (approved):

som-snytt:
FSR this change says `Integer` instead of `Int`

hrhino:
@SethTisue and @dwijnand are running around checking in on old open PRs, but it's @som-snytt's job to check in on old closed PRs.

som-snytt:
My ears were burning, and then the github app chimed. FTR there's a new open PR in the vicinity, which I looked at even though I swore never to crack open `sys.process` until it was spun off into ["its own module,"](https://github.com/scala/scala/pull/6052#issuecomment-326344249) which it turns out was just a metaphor.

hrhino:
Oh, that's fixed in Dotty.

PR #6053, opened 2017-09-02
Fix scala/bug#10490: Allow Java nested class selection

jvican:
Java type selections, for example `Foo.Bar`, can represent two different
trees depending on the context. On the one hand, it can be a type
selection over an object `Foo`. On the other hand, it can be a type
selection over a type (that is, a selection of a nested class).

However, when we parse the Scala signatures is not possible to know
which case the user meant. For example, `Foo.Bar` could be either:

1. `Select(Ident(TermName("Foo")), TypeName("Bar"))`; or
2. `SelectFromTypeTree(Ident(TypeName("Foo")), TypeName("Bar"))`.

By default, the Scala compiler parses `Foo.Bar` in a type position as
the first option, but it could very well be the case that the user meant
the second one, as specified by the tests in this commit.

To address the issue, then, there's nothing but the ugly solution of
injecting special logic into typer so that we try to typecheck both
trees for those different scenarios.

To handle this specific logic and try to make it as isolated as
possible, I've introduced a new context mode called `JavaTypeSelection`.
This context mode is set by `typedSelectOrSuperCall` when the
compilation unit is Java, the tree at hands selects a type name and the
qualifier is an identifier. `typedIdent` detects when this mode is on,
and if the default term symbol lookup fails it tries to find the type
symbol. If that succeeds, it then delegates to `typedType`. The
different returning trees are handled in `typedSelectOrSuperCall`.

Run tests in this commit illustrate both accesses. They are independent
because if t10490 had the `Foo` companion, the bug that this commit
fixes wouldn't have been required (i.e. it's a very subtle workaround).

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4087,15 +4087,6 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
       def acceptsApplyDynamic(tp: Type) = tp.typeSymbol isNonBottomSubClass DynamicClass
 
-      /** Returns `Some(t)` if `name` can be selected dynamically on `qual`, `None` if not.
-       * `t` specifies the type to be passed to the applyDynamic/selectDynamic call (unless it is NoType)
-       * NOTE: currently either returns None or Some(NoType) (scala-virtualized extends this to Some(t) for selections on staged Structs)
-       */
-      def acceptsApplyDynamicWithType(qual: Tree, name: Name): Option[Type] =
-        // don't selectDynamic selectDynamic, do select dynamic at unknown type,
-        // in scala-virtualized, we may return a Some(tp) where tp ne NoType
-        if (!isApplyDynamicName(name) && acceptsApplyDynamic(qual.tpe.widen)) Some(NoType)
-        else None
 
       def isDynamicallyUpdatable(tree: Tree) = tree match {
         case DynamicUpdate(qual, name) =>
@@ -4147,7 +4138,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
        *  - simplest solution: have two method calls
        *
        */
-      def mkInvoke(context: Context, tree: Tree, qual: Tree, name: Name): Option[Tree] = {
+      def mkInvoke(context: Context, tree: Tree, qual: Tree, name: Name): Tree = {
         val cxTree = context.enclosingNonImportContext.tree // scala/bug#8364
         debuglog(s"dyna.mkInvoke($cxTree, $tree, $qual, $name)")
         val treeInfo.Applied(treeSelection, _, _) = tree
@@ -4158,7 +4149,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
             case _                              => false
           }
         }
-        acceptsApplyDynamicWithType(qual, name) map { tp =>
+        if (!isApplyDynamicName(name) && acceptsApplyDynamic(qual.tpe.widen)) {
           // If tp == NoType, pass only explicit type arguments to applyXXX.  Not used at all
           // here - it is for scala-virtualized, where tp will be passed as an argument (for
           // selection on a staged Struct)
@@ -4194,8 +4185,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
             case _ =>
               setError(tree)
           }
-        }
+        } else EmptyTree
       }
+
       def wrapErrors(tree: Tree, typeTree: Typer => Tree): Tree = silent(typeTree) orElse (err => DynamicRewriteError(tree, err.head))
     }
 
@@ -4207,15 +4199,6 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       // Lookup in the given qualifier.  Used in last-ditch efforts by typedIdent and typedSelect.
       def lookupInRoot(name: Name): Symbol  = lookupInOwner(rootMirror.RootClass, name)
       def lookupInEmpty(name: Name): Symbol = rootMirror.EmptyPackageClass.info member name
-
-      def lookupInQualifier(qual: Tree, name: Name): Symbol = (
-        if (name == nme.ERROR || qual.tpe.widen.isErroneous)
-          NoSymbol
-        else lookupInOwner(qual.tpe.typeSymbol, name) orElse {
-          NotAMemberError(tree, qual, name)
-          NoSymbol
-        }
-      )
 
       def typedAnnotated(atd: Annotated): Tree = {
         val ann = atd.annot
@@ -4877,10 +4860,6 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           if (tree.isInstanceOf[SelectFromTypeTree]) TypeSelectionFromVolatileTypeError(tree, qual)
           else UnstableTreeError(qual)
         else {
-          def asDynamicCall = dyna.mkInvoke(context, tree, qual, name) map { t =>
-            dyna.wrapErrors(t, (_.typed1(t, mode, pt)))
-          }
-
           val sym = tree.symbol orElse member(qual, name) orElse inCompanionForJavaStatic(qual.tpe.prefix, qual.symbol, name)
           if ((sym eq NoSymbol) && name != nme.CONSTRUCTOR && mode.inAny(EXPRmode | PATTERNmode)) {
             // symbol not found? --> try to convert implicitly to a type that does have the required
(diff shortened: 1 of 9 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4996,9 +4996,20 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
           if (Statistics.canEnable) Statistics.incCounter(typedSelectCount)
-          val qualTyped = checkDead(typedQualifier(qual, mode))
+          def typeQual = checkDead(typedQualifier(qual, mode))
+          val (qualTyped, isJavaTypeSelection) = {
+            if (name.isTermName) (typeQual, false)
+            else {
+              qual match {
+                case id: Ident if context.unit.isJava =>
+                  val qualTyped1 = context.withinJavaTypeSelection(typeQual)
+                  (qualTyped1, qualTyped1.isType)
```
jvican: It's really interesting that this is all we need to do in `typedSelectOrSuperCall`, since the whole method will return `Select(qualTyped1, name)` whereas the more correct tree would have been `SelectFromTypeTree(qualTyped1, name)`. I have not done this change because of simplicity. But it does make me wonder what is the crucial difference between `Select` and `SelectFromTypeTree`... It seems that `Select(Ident(TypeName("Foo")), _) == SelectFromTypeTree(Ident(TypeName("Foo")), _)`, no?

lrytz: I think for consistency it would be better to generate the `SelectFromTypeTree`, even if a `Select` doesn't violate any of the current assumptions in the rest of the compiler.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4996,9 +4996,20 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           typedSelect(tree, typedSelectOrSuperQualifier(qual), nme.CONSTRUCTOR)
         case Select(qual, name) =>
           if (Statistics.canEnable) Statistics.incCounter(typedSelectCount)
-          val qualTyped = checkDead(typedQualifier(qual, mode))
+          def typeQual = checkDead(typedQualifier(qual, mode))
+          val (qualTyped, isJavaTypeSelection) = {
+            if (name.isTermName) (typeQual, false)
+            else {
+              qual match {
+                case id: Ident if context.unit.isJava =>
```
lrytz: A comment would be good here, e.g.

```
scala/bug#10490: given a Scala class A { class B }, in mixed compilation, a reference A.B in a Java source
would cause a "not found: value A". The `JavaTypeSelection` mode makes `typedIdent` type check A as
a type if no value is found. In case A has a companion, it works thanks to the fix of scala/bug#3120 in
typedSelectInternal.
```

jvican: I'll add it 👍

src/compiler/scala/tools/nsc/typechecker/Typers.scala
jvican: I can do that, but I'm afraid to modify too much of the logic of this method since it's really hot. I also think that for consistency it's better to generate a `SelectFromTypeTree`. Doing that requires me to wrap almost all the rest of the body that follows this statement with an `if`.

retronym:
It would be worth looking at #5762, which overhauls typechecking of selections to remove inconsistencies. I wasn't brave enough to merge that into 2.12.3, but will likely be merged soon.

smarter:
> To address the issue, then, there's nothing but the ugly solution of
injecting special logic into typer so that we try to typecheck both
trees for those different scenarios.

Isn't there already logic in the typer to do this in https://github.com/scala/scala/blob/e1e8d050deb643ca56db1549e2e5a3114572a952/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L4919-L4926 ? Dotty does something very much inspired by that (https://github.com/lampepfl/dotty/blob/851c3c2212b7457dedc9ad1e59ff688b17458db5/compiler/src/dotty/tools/dotc/typer/Typer.scala#L418-L421) and both of your test cases compile with it.

jvican:
I saw that logic, I don't know why it was added but it doesn't seem to work; this path in `typedSelect` is never hit because it is `typedIdent` the one that fails to typecheck `Ident(TypeName("foo"))`. An error is registered right away in the reporter and the symbol and name are marked as errors when they hit `typedSelect`.
  
I would also argue that aside from working, my fix is more elegant and correct. The real fix should happen in `typedIdent`, not `typedSelect`, and `typedSelect` should not try to reverse errors that are reported by `typedIdent`. The less communication and trial and error, the better IMO.

jvican:
@retronym I've read the code in that PR. It doesn't seem to interact with the bug that I fix here, let me know if I'm wrong. I'm happy to rebase this when that gets merged.

lrytz:
The existing logic in `typedSelect` is for a selection `A.B` in a Java source file, where `A` and `B` are java-defined classes.

```
class A { class B { } }        // v1
class A { static class B { } } // v2
class C { void foo(A.B) { } }
```

For each package-level Java class, the Scala compiler creates a class and a module symbol, static members are added to the module symbol.

When the Scala compiler type-checks `A.B`, it doesn't know if `B` is a static class or not. That's what the existing fix is for. (If `A.B` was written in Scala, we'd now it's the static member, as the non-static one would be `A#B`.)

The fix in this PR is when selecting `A.B` in a Java source with `A` defined in Scala. `A` is type checked as a term ident. This fails if `A` doesn't have a companion object, as there's no value `A`, only a type.

So there must be some other code path than https://github.com/lampepfl/dotty/blob/851c3c2212b7457dedc9ad1e59ff688b17458db5/compiler/src/dotty/tools/dotc/typer/Typer.scala#L418-L421 in dotty that makes the example work.

jvican:
@adriaanm Any chance this can be merged for 2.12.4? I'll update it as soon as https://github.com/scala/scala/pull/5762 lands...

adriaanm:
I don't see why not. I'm making my way through the PR queue. Will look at this one next (in the next few hours).

adriaanm:
I assume you meant whether it can be merged once the review feedback has been addressed? I +1 the more comments as requested by Lukas -- this is tricky stuff and there's some good knowledge around this in this PR that should end up in the code. 

I'll look at the actual code after lunch, but I would also second the other request in directly emitting the right tree (SelectFromType), if we can find an elegant way to do it.

adriaanm:
When you rebase, please also drop "Fix scala/bug#10490: " from the commit title. That should go into the PR description. Thanks for the great commit message.

adriaanm:
I've pushed a [tree](https://github.com/adriaanm/scala/tree/rebase-6053) with a sketch that rebases your work on my earlier PR and also boyscouts a bit in the area. Could you verify that it makes sense and absorb this into your PR? Now that I've looked at the code, I feel this area has hit a critical point where we need to do a bit of reinforcement before we tack on more stuff.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
adriaanm: This should also get a comment that explains why `context.unit.isJava` here needs to be reified as a `context.withinJavaTypeSelection` for consumption down the line in `typedIdent`.

jvican:
> I assume you meant whether it can be merged once the review feedback has been addressed? 
  
Indeed.
  
> When you rebase, please also drop "Fix scala/bug#10490: " from the commit title. That should go into the PR description. Thanks for the great commit message.
  
Deal.
  
> I've pushed a tree with a sketch that rebases your work on my earlier PR and also boyscouts a bit in the area. Could you verify that it makes sense and absorb this into your PR? Now that I've looked at the code, I feel this area has hit a critical point where we need to do a bit of reinforcement before we tack on more stuff.
  
Excellent. I'm absorbing these commits into this PR.

jvican:
This PR has been updated according to feedback and is ready for another pass. I haven't scheduled benchmarks because I'm not sure this PR requires them!

jvican:
Weird -- I cannot reproduce the error locally. I'm trying tomorrow again.

adriaanm:
I'll take a look as well

adriaanm:
I think we should just update the check file. It doesn't make sense to me to expand to `selectDynamic("<error>")` when we encounter a mall-formed tree. This is why I pulled [that condition out](https://github.com/scala/scala/pull/6053/commits/7b8450b4db91d9a01217b1000989eaa83f614a3e#diff-4eab1aad4533a31c10565971e90f73eaR4921) as `if (name == nme.ERROR || qual.tpe.widen.isErroneous) EmptyTree`.

jvican:
@adriaanm Makes sense. PR has been updated with the updated check file, agreed that the new error message is better than exposing the last try with the dynamic call typechecking.

lrytz:
- I'd like to see a summary of https://github.com/scala/scala/pull/6053#issuecomment-328479876 in the commit message.
- Adriaan's comment https://github.com/scala/scala/pull/6053#discussion_r139848142 was not addressed, I agree we should document why a mode is necessary
- I also proposed a comment here: https://github.com/scala/scala/pull/6053#discussion_r136821444, I don't see it added.
- We're still building a `Select` tree instead of a `SelectFromTypeTree`. I'd like to see that changed, @adriaanm wdyt?

jvican:
> I'd like to see a summary of #6053 (comment) in the commit message.
  
I'll add it.
  
> Adriaan's comment #6053 (comment) was not addressed, I agree we should document why a mode is necessary
  
Sorry, I missed this one.

> I also proposed a comment here: #6053 (comment), I don't see it added.
  
It's been added, but it's shorter:
https://github.com/scala/scala/pull/6053/files#diff-4eab1aad4533a31c10565971e90f73eaR5047

adriaanm:
> We're still building a Select tree instead of a SelectFromTypeTree. I'd like to see that changed, @adriaanm wdyt?

yes please -- let's stick to the convention (although the original design decision is not immediately clear to me)

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4987,8 +4967,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         }
       }
 
-      def typedTypeSelectionQualifier(tree: Tree, pt: Type = AnyRefTpe) =
-        context.withImplicitsDisabled { typed(tree, MonoQualifierModes | mode.onlyTypePat, pt) }
+      def typedTypeSelectionQualifier(tree: Tree, pt: Type = AnyRefTpe) = {
+        def typedQual = context.withImplicitsDisabled { typed(tree, MonoQualifierModes | mode.onlyTypePat, pt) }
+        tree match {
+          case id: Ident if context.unit.isJava => context.withinJavaTypeSelection { typedQual }
```
adriaanm, line 4973: Talking this through with @retronym, it occurred to us that this `withinJavaTypeSelection` context bit would also be useful in @hrhino's quest to improve error messages in #6042. With that insight, should it just be `withinTypeSelection` (and should we have one for `withinPattern`?)

Also, why is this is only done when we reach an `Ident`? What are the varying combos of `A`, `A.B` and `A.B.C`.

lrytz: > why is this is only done when we reach an Ident

For selections there's an existing fix / workaround, see my comment here https://github.com/scala/scala/pull/6053#issuecomment-328479876.

adriaanm: Would it be feasible to combine the two fixes?

jvican:
@adriaanm Let's delay this change to 2.12.5. I'll update this PR and try to integrate it with other efforts in this area, as you've pointed out to @hrhino's efforts.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
jvican: Yes, I think so.

jvican: @adriaanm i was having a look at rebasing this PR and addressing your comment. Do you want me to combine the two fixes on this PR or how should we operate? /cc @hrhino

adriaanm: combining sounds good to me

jvican: Alright, I'll cherry-pick @hrhino's changes into this PR :+1:

SethTisue:
@jvican interested in trying to finish this up in time for 2.12.5...? I've changed the milestone to 2.12.6, but only tentatively.

jvican:
@SethTisue Let's leave it for 2.12.6, since I believe that integrating Harrison's changes will take me a little bit of time and there are things with higher priority on my scala/scala queue :smile:.

adriaanm:
At this point I think we should move this work to 2.13.0-M5. I would also encourage you to integrate the other efforts in this area.

jvican:
Can i reopen this PR targeting 2.13.0-M5?

SethTisue:
sure, you can retarget and reopen

retronym:
I'm taking a look at reviving this in https://github.com/scala/scala/compare/2.13.x...retronym:topic/java-inherited-type-ident

PR #6055, opened 2017-09-03
Initial attempt at limiting *only* the compensating threads of the global EC

viktorklang:
(This PR is for discussion with the reporter of the issue, do not merge yet)

### DIFF
--- a/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
+++ b/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
@@ -10,9 +10,9 @@ package scala.concurrent.impl
 
 import java.util.concurrent.{ ForkJoinPool, ForkJoinWorkerThread, ForkJoinTask, Callable, Executor, ExecutorService, ThreadFactory, TimeUnit }
 import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.Semaphore
 import java.util.Collection
 import scala.concurrent.{ BlockContext, ExecutionContext, CanAwait, ExecutionContextExecutor, ExecutionContextExecutorService }
-import scala.annotation.tailrec
 
 
 private[scala] class ExecutionContextImpl private[impl] (val executor: Executor, val reporter: Throwable => Unit) extends ExecutionContextExecutor {
@@ -23,28 +23,19 @@ private[scala] class ExecutionContextImpl private[impl] (val executor: Executor,
 
 
 private[concurrent] object ExecutionContextImpl {
-
   // Implement BlockContext on FJP threads
   final class DefaultThreadFactory(
     daemonic: Boolean,
-    maxThreads: Int,
+    maxExtraThreads: Int,
     prefix: String,
-    uncaught: Thread.UncaughtExceptionHandler) extends ThreadFactory with ForkJoinPool.ForkJoinWorkerThreadFactory {
+    uncaught: Thread.UncaughtExceptionHandler)
+    extends Semaphore(maxExtraThreads) with ThreadFactory with ForkJoinPool.ForkJoinWorkerThreadFactory {
 
     require(prefix ne null, "DefaultThreadFactory.prefix must be non null")
-    require(maxThreads > 0, "DefaultThreadFactory.maxThreads must be greater than 0")
+    require(maxExtraThreads >= 0, "DefaultThreadFactory.maxExtraThreads must be greater than or equal to 0")
 
-    private final val currentNumberOfThreads = new AtomicInteger(0)
-
-    @tailrec private final def reserveThread(): Boolean = currentNumberOfThreads.get() match {
-      case `maxThreads` | Int.`MaxValue` => false
-      case other => currentNumberOfThreads.compareAndSet(other, other + 1) || reserveThread()
-    }
-
-    @tailrec private final def deregisterThread(): Boolean = currentNumberOfThreads.get() match {
-      case 0 => false
-      case other => currentNumberOfThreads.compareAndSet(other, other - 1) || deregisterThread()
-    }
+    override def toString: String =
+      s"DefaultThreadFactory(daemonic = $daemonic, maxExtraThreads = $maxExtraThreads, prefix = $prefix, uncaught = $uncaught)"
 
     def wire[T <: Thread](thread: T): T = {
       thread.setDaemon(daemonic)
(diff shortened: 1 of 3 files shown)


test/files/jvm/future-spec/FutureTests.scala
```diff
@@ -817,6 +817,24 @@ class FutureTests extends MinimalScalaTest {
...
+        val all = Future.sequence(Vector.range(1, nTasks + 1).map {
+          result => Future {
+            if (blocking) scala.concurrent.blocking { Thread.sleep(time) }
+            else Thread.sleep(time)
+            //log(s"$result - sleep was: $time - ${java.util.Calendar.getInstance().getTime()}")
+            result
+          }
+        })
+        //log("waiting")
+        Await.result(all, Duration(s"${10 + (time.toLong * nTasks)}s"))
```
avkonst: Duration for this waiting should be smaller. 
- For the blocking case, it is: `(nTasks / (scala.concurrent.context.maxExtraThreads + scala.concurrent.context.maxExtraThreads) + 1) * sleepTime + minorAdditionForTestStability`. (minorAdditionForTestStability can be well below 10 seconds, eg. 1-2 should be heaps)
- For the non-blocking case it is: `(nTasks / scala.concurrent.context.maxThreads + 1) * sleepTime + minorAdditionForTestStability`.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +52,55 @@ private[concurrent] object ExecutionContextImpl {
...
-                  } finally {
-                    isdone = true
-                  }
-
-                true
-              }
-              override def isReleasable = isdone
-            })
-            result
+    private[this] final def allocateNormalThread(): Boolean =
```
avkonst: I am not able to review this logic carefully. I understood 'scala.concurrent.blocking' clause now has got only effect when current number of already blocked threads is < scala.concurrent.context.maxExtraThreads. Is that right? 
I achieved the same logic by placing a semaphore before invoking scala.concurrent.blocking. I think applying a semaphore would help to make this code easier, but up to you..

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +52,55 @@ private[concurrent] object ExecutionContextImpl {
...
+
+      final class ExtraThread(fjp: ForkJoinPool) extends ForkJoinWorkerThread(fjp) with BlockContext {
+        final override def onTermination(exception: Throwable): Unit = deregisterExtraThread()
+        final override def blockOn[T](thunk: =>T)(implicit permission: CanAwait): T =
+          BlockContext.defaultBlockContext.blockOn(thunk)
+      }
+
+      final class WorkerThread(fjp: ForkJoinPool) extends ForkJoinWorkerThread(fjp) with BlockContext {
+        @volatile var isBlocking: Boolean = false
+        final override def blockOn[T](thunk: =>T)(implicit permission: CanAwait): T = if (isBlocking) {
```
avkonst: Also, just an observation. I noticed that scala.concurrent.blocking has got only effect when global thread pool is used as an executor context. When a client implements (or sources 3rd party) another executor context, there is no way to instruct executor context about blocking call. As a result, the only correctly working with `blocking` executor context is the global one (or the one created from `null` executor). I see it as incomplete API case or unfair favoring for scala internal implementation of an executor context.

avkonst (commented):
Thanks. My observations are in the comments.

test/files/jvm/future-spec/FutureTests.scala
viktorklang: @avkonst It is as it is because it doesn't look at number of available threads. Can have a look at that later, but it is a bit delicate because we cannot rely on specific pool behavior (since it uses the JDK FJP)

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: @avkonst No, what the code does is that it only spawns compensating threads up to maxExtraThreads.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: @avkonst No, anyone who writes an ExecutionContext, and anyone who wraps one, and anyone who wants to install some other BlockContext within a scope is free to do so. So it is most definitely not favoring any implementation. See: http://viktorklang.com/blog/Futures-in-Scala-protips-4.html

avkonst:
Thanks. I have seen BlockContext before but did not know how to use it.

lrytz:
To fix the build failure, rebase on current 2.12.x

viktorklang:
@lrytz I'm experimenting with minimizing solution. will push updated PR when available.

viktorklang:
@avkonst I think I have created a solution for this problem. But I think I'd like some more eyes on this, and especially field feedback before I think it is safe to go into the stdlib.

avkonst:
@viktorklang Hi, thanks. I can try to integrate it to the product and run some tests. The problem was random reproducing, until I figured out the stable reproducer (which I reported). So, not sure I can provide field feedback. It also may take time. I can have a look into it, but not sure how much I can be helpful... do you have a review / PR?

viktorklang:
@avkonst This is the PR :) (https://github.com/scala/scala/pull/6055/files)

test/files/jvm/t10484.scala
```diff
@@ -0,0 +1,32 @@
...
+              }
+            } else {
+              Thread.sleep(time)
+              log(result.toString)
+            }
+
+            result
+          }).map(_ => ())
+      }
+      Await.result(spawn(600, "stage1", true), Duration(s"${1 + (time.toLong * 600)}s"))
```
avkonst, line 28: Since tasks run in parallel, await duration should be lower.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -96,27 +102,22 @@ private[concurrent] object ExecutionContextImpl {
...
-    // scala/bug#8955 Deadlocks can happen if maxNoOfThreads is too low, although we're currently not sure
-    //         about what the exact threshold is. numThreads + 256 is conservatively high.
-    val maxNoOfThreads = getInt("scala.concurrent.context.maxThreads", "x1")
+    def range(floor: Int, desired: Int, ceiling: Int): Int = scala.math.min(scala.math.max(floor, desired), ceiling)
 
     val desiredParallelism = range(
       getInt("scala.concurrent.context.minThreads", "1"),
-      numThreads,
-      maxNoOfThreads)
+      getInt("scala.concurrent.context.numThreads", "x1"),
```
avkonst, line 109: what does it mean "x1"?

avkonst: please, ignore this. I have found what it means.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +44,54 @@ private[concurrent] object ExecutionContextImpl {
...
-
-                true
+    private[this] final def shouldAllocateUnlimited(): Boolean =
+      Thread.currentThread match {
+        case w: WorkerThread if w.isBlocking => false
+        case _ => true
+      }
+
+    override def newThread(runnable: Runnable): Thread =
+      if (shouldAllocateUnlimited()) wire(new Thread(runnable))
```
avkonst, line 54: Should it be really unlimited? Can it cause uncontrolled number of threads?

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +44,54 @@ private[concurrent] object ExecutionContextImpl {
...
-            result
+            } finally {
+              isBlocking = false
+            }
           }
-        })
-      } else null
+      }
+    final class Block[T](thunk: =>T) extends ForkJoinPool.ManagedBlocker {
+      @volatile var done: Boolean = _ // TODO: validate whether @volatile is required or not
```
avkonst: should it be private?

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +44,54 @@ private[concurrent] object ExecutionContextImpl {
...
+      var result: T = null.asInstanceOf[T]
+      override def isReleasable = done
+      override def block(): Boolean =
```
avkonst: if this function is not invoked in a loop like 'while(!block()){}', done does not have to be volatile. Also trunk call likely does at least one memory barrier anyway.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +44,54 @@ private[concurrent] object ExecutionContextImpl {
...
-                    // When we block, switch out the BlockContext temporarily so that nested blocking does not created N new Threads
-                    BlockContext.withBlockContext(BlockContext.defaultBlockContext) { thunk }
-                  } finally {
-                    isdone = true
-                  }
-
-                true
+    private[this] final def shouldAllocateUnlimited(): Boolean =
+      Thread.currentThread match {
+        case w: WorkerThread if w.isBlocking => false
```
avkonst, line 49: sorry, I do not understand the logic behind this. Why blocking worker thread is the only case for not allowing unlimited threads?

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -53,39 +44,54 @@ private[concurrent] object ExecutionContextImpl {
...
+      }
+    final class Block[T](thunk: =>T) extends ForkJoinPool.ManagedBlocker {
+      @volatile var done: Boolean = _ // TODO: validate whether @volatile is required or not
+      var result: T = null.asInstanceOf[T]
+      override def isReleasable = done
+      override def block(): Boolean =
+        if (!done) {
+          result = try thunk finally { done = true }
+          true
+        } else true
```
avkonst: always returns true - is it by design?

test/files/jvm/t10484.scala
viktorklang: Yes, but I do not know what the underlying CPU is, so I'm being conservative.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: :+1:

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: So if we are in `newThread` and the current thread is a WorkerThread which is currently blocking, then it is more than likely that this is a request for a compensating thread.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: This is actually harmless in this case since this thread factory is only going to be used for newThread(ForkJoinPool). The ThreadFactory is legacy compliance (when the fallback was a ThreadPoolExecutorService)

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: Actually I might just rename it to `isReleasable`.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: @avkonst Yes and no, technically it is unspecified, because the pool could share the ManagedBlocker instance and have multiple threads waiting for it to be releaseable.
So it is volatile to be conservative to the FJP implementation used.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: Yes, according to contract set forth by ManagedBlocker.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
avkonst: ok, there is historical knowledge, which I do not know about, so do not take my review as a serious check.

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
avkonst: in this case, I would move true out of if block and would delete else branch.

viktorklang: done

src/library/scala/concurrent/impl/ExecutionContextImpl.scala
viktorklang: done

adriaanm:
Note that the 2.12.4 window will close in 7 days.

viktorklang:
I don't think this change is mature enough to ship at this point. There is
a workaround (increase extra threads) if one hits the edge case. I guess we
could have a flag to switch between impls but that would complicate things
even further.

-- 
Cheers,
√

adriaanm:
√, thanks for the update.

viktorklang:
You're most welcome!

-- 
Cheers,
√

SethTisue:
@viktorklang could you rebase this against current 2.12.x? I did a community build run but I got some strange failures I don't think are related to your changes. rather than investigate, I'd like to try again against fresher code.

adriaanm:
I'm closing this PR because it has been dormant for a while now. We'd be very happy to receive a new PR that addresses points raised by the reviewer.

NthPortal:
@viktorklang is this superseded by #6529?

viktorklang:
@NthPortal Correct. Thanks!

PR #6056, opened 2017-09-04
WIP: Add test for note order

janekdb:


janekdb:
Cannot get this test to fail locally. Checking the build fails here. Will close PR after build.

PR #6057, opened 2017-09-05
Pipeline code gen and post-processing if global optimizations disabled

lrytz:
If no global optimizations are enabled (inlining, closure rewriting - these
require building the call graph of the code being compiled), we can run the
backend (local optimizations, classfile writing) directly after code gen
without storing the ClassNodes.

Motivation is to reduce memory pressure and the required heap space.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/CodeGen.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/CodeGen.scala
@@ -1,6 +1,7 @@
 package scala.tools.nsc
 package backend.jvm
 
+import scala.collection.mutable
 import scala.tools.asm.tree.ClassNode
 
 abstract class CodeGen[G <: Global](val global: G) extends PerRunInit {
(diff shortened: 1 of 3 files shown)


lrytz:
Benchmarks should [show up here](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=a28305b676%7Cdb24e3761f&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1504122566185&to=1504604620225)

src/compiler/scala/tools/nsc/backend/jvm/CodeGen.scala
```diff
@@ -44,6 +45,7 @@ abstract class CodeGen[G <: Global](val global: G) extends PerRunInit {
     }
 
     genClassDefs(unit.body)
+    res.toList
```
jvican, line 48: Naive question: could it be possible to avoid this conversion by returning `Iterator` and only converting to `List` or `Iterable` when optimizations are enabled and the code path hits `postProcessAndSendToDisk`? It seems that the happy path we're optimizing for here only needs one traversal over this data structure, in `postProcessor.generatedClasses ++= generated`.

lrytz: `ListBuffer.toList` is [almost a no-op](https://github.com/scala/scala/blob/v2.12.3/src/library/scala/collection/mutable/ListBuffer.scala#L313-L316).

jvican: Cool, didn't know!

retronym (approved):
Very lean and clean.

PR #6058, opened 2017-09-05
Classes can access `protected static` members of their Java superclasses

hrhino:
This more-or-less reverts the fix added in 1ebbe029dd7ba. Instead of wantonly treating `protected static` Java members like they're public, it suffices to modify `isSubClassOrCompanion` to consider a class a "subclass or companion" of the companions of its parent classes, provided the parent is a Java-defined class.

It's a bit wacky, but comparing the Java and Scala definitions of `protected` access:

[SLS 5.2.8](https://www.scala-lang.org/files/archive/spec/2.12/05-classes-and-objects.html#protected):

> Protected members of a class can be accessed from within
>
> - the template of the defining class,
> - all templates that have the defining class as a base class,
> - the companion module of any of those classes.

[JLS 6.6.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.6.1):

> A member (class, interface, field, or method) of a reference type ... is accessible only if the type is accessible and the member or constructor is declared to permit access:
> ...
> - if the member or constructor is declared protected, then access is permitted only when one of the following is true:
>   - Access to the member or constructor occurs from within the package containing the class in which the protected member or constructor is declared.
>   - Access is correct as described in §6.6.2.

[JLS 6.6.2.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.6.2.1):

> Let C be the class in which a protected member is declared. Access is permitted only within the body of a subclass S of C.

The important differences are that Java protected members are indeed accessible from the package that they're declared in (this behavior is implemented by `accessBoundary`, and that `static` methods are treated the same as instance methods for purposes of access control. Therefore, I feel comfortable adding the third case to `isSubClassOrCompanion`, thus taking the Java access control model when dealing with Java-defined types.

I also scouted a grammar tweak, which makes this have a few more files in the diff than would be expected.

Fixes scala/bug#6934; review by @adriaanm and (of course) whoever feels like swinging by and saying hi.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -1,5 +1,5 @@
 /* NSC -- new Scala compiler
- * Copyright 2005-2013 LAMP/EPFL
+ * Copyright 2005-2017 LAMP/EPFL
  * @author  Martin Odersky
  */
 
@@ -647,10 +647,16 @@ trait Contexts { self: Analyzer =>
     // Accessibility checking
     //
 
-    /** Is `sub` a subclass of `base` or a companion object of such a subclass? */
+    /** True iff...
+      * - `sub` is a subclass of `base`
+      * - `sub` is the module class of a companion of a subclass of `base`
+      * - `base` is a Java-defined module class (containing static members),
+      *   and `sub` is a subclass of its companion class. (see scala/bug#6394)
+      */
     private def isSubClassOrCompanion(sub: Symbol, base: Symbol) =
       sub.isNonBottomSubClass(base) ||
-    sub.isModuleClass && sub.linkedClassOfClass.isNonBottomSubClass(base)
+        (sub.isModuleClass && sub.linkedClassOfClass.isNonBottomSubClass(base)) ||
+        (base.isJavaDefined && base.isModuleClass && sub.isNonBottomSubClass(base.linkedClassOfClass))
 
     /** Return the closest enclosing context that defines a subclass of `clazz`
      *  or a companion object thereof, or `NoContext` if no such context exists.
@@ -702,10 +708,10 @@ trait Contexts { self: Analyzer =>
         val c = enclosingSubClassContext(sym.owner)
         if (c == NoContext)
           lastAccessCheckDetails =
-            "\n Access to protected "+target+" not permitted because"+
-            "\n "+"enclosing "+this.enclClass.owner+
-            this.enclClass.owner.locationString+" is not a subclass of "+
-            "\n "+sym.owner+sym.owner.locationString+" where target is defined"
+            sm"""
+                | Access to protected $target not permitted because
+                | enclosing ${this.enclClass.owner}${this.enclClass.owner.locationString} is not a subclass of
+                | ${sym.owner}${sym.owner.locationString} where target is defined"""
         c != NoContext &&
         {
           target.isType || { // allow accesses to types from arbitrary subclasses fixes #4737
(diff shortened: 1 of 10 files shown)


lrytz:
Nice!

Curious: where is it implemented that Scalac allows an access to a java-defined protected member in a non-subclass in the same package (basically your test-case that says "double-checking that we can still do this")?

hrhino:
That's just because both [`JavaParsers`](https://github.com/scala/scala/blob/0b1eacfe21a49e2f8684bc5c2b81d931a05fc027/src/compiler/scala/tools/nsc/javac/JavaParsers.scala#L391-L393) and [`ClassfileParser`](https://github.com/scala/scala/blob/0b1eacfe21a49e2f8684bc5c2b81d931a05fc027/src/reflect/scala/reflect/internal/PrivateWithin.scala#L23-L25) treat `protected` in Java as `protected[containingPackage]`.

lrytz:
Thanks!

lrytz (approved):
LGTM, I'll let it sit in the queue so @adriaanm can take a look.

adriaanm (approved):
Excellent, thanks!

retronym:
Does this PR fix https://github.com/scala/bug/issues/9272 ?

hrhino:
I don't think so:

```scala
hhoughton:tmp hhoughton$ scala-launch 2.12.4-bin-c2a5883 -cp android-4.1.1.4.jar
Welcome to Scala 2.12.4-bin-c2a5883 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> import android._, content._, provider._, ContactsContract._
import android._
import content._
import provider._
import ContactsContract._

scala> :paste

val ops: java.util.ArrayList[ContentProviderOperation] = new java.util.ArrayList[ContentProviderOperation]
    ops.add(ContentProviderOperation.newInsert(RawContacts.CONTENT_URI)
            .withValue(RawContacts.ACCOUNT_TYPE, AccountType)
            .withValue(RawContacts.ACCOUNT_NAME, AccountName).build)

<pastie>:27: error: value ACCOUNT_TYPE is not a member of object android.provider.ContactsContract.RawContacts
            .withValue(RawContacts.ACCOUNT_TYPE, AccountType)
                                   ^
<pastie>:27: error: not found: value AccountType
            .withValue(RawContacts.ACCOUNT_TYPE, AccountType)
                                                 ^
<pastie>:28: error: value ACCOUNT_NAME is not a member of object android.provider.ContactsContract.RawContacts
            .withValue(RawContacts.ACCOUNT_NAME, AccountName).build)
                                   ^
<pastie>:28: error: not found: value AccountName
            .withValue(RawContacts.ACCOUNT_NAME, AccountName).build)
                                                 ^

scala>
```

I'm surprised that this works in Java, but it does.

```java
hhoughton:tmp hhoughton$ cat Ax.java Sax.java Wax.java
// Ax.java
package a;

public class Ax {
  protected static interface Max {
    static int Bax = 1;
  }
}

// Sax.java
package a;

public class Sax implements Ax.Max {}


// Wax.java
package b;

import a.*;

public class Wax {
  int hax = Sax.Bax;
}


hhoughton:tmp hhoughton$ javac Ax.java Sax.java Wax.java
hhoughton:tmp hhoughton$ echo $?
0
```

(but of course:)

```scala
hhoughton:tmp hhoughton$ scala-launch 2.12.4-bin-c2a5883 -cp .
Welcome to Scala 2.12.4-bin-c2a5883 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> val flax = Sax.Bax
<console>:11: error: value Bax is not a member of object Sax
       val flax = Sax.Bax
                      ^
```

PR #5906, opened 2017-05-15
WIP mapOver/traverse/transform: refactor pattern match into virtual call [ci: last-only]

retronym:
Pushing the WIP branch here for the record.

The performance gains of ~2% need further attribution to individual commits
and analysis of the underlying reasons.

```
2.12.3-bin-2f52c61-SNAPSHOT

[info] 
[info] Benchmark                                   (extraArgs)  (source)    Mode  Cnt     Score   Error  Units
[info] HotScalacBenchmark.compile                                         sample  365  1146.958 ± 6.225  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00                           sample       1088.422          ms/op
[info] HotScalacBenchmark.compile:compile·p0.50                           sample       1147.142          ms/op
[info] HotScalacBenchmark.compile:compile·p0.90                           sample       1172.308          ms/op
[info] HotScalacBenchmark.compile:compile·p0.95                           sample       1234.174          ms/op
[info] HotScalacBenchmark.compile:compile·p0.99                           sample       1289.077          ms/op
[info] HotScalacBenchmark.compile:compile·p0.999                          sample       1317.011          ms/op
[info] HotScalacBenchmark.compile:compile·p0.9999                         sample       1317.011          ms/op
[info] HotScalacBenchmark.compile:compile·p1.00                           sample       1317.011          ms/op
[success] Total time: 869 s, completed 15/05/2017 5:41:30 PM

2518bc6e2c Spread Tree transform/traverse into virtual methods of Tree
2.12.3-bin-c26c401-SNAPSHOT: Spread mapOver into invokevirtual of Type.mapOver

[info] 
[info] Benchmark                                   (extraArgs)  (source)    Mode  Cnt     Score   Error  Units
[info] HotScalacBenchmark.compile                                         sample  361  1138.783 ± 4.662  ms/op
[info] HotScalacBenchmark.compile:compile·p0.00                           sample       1086.325          ms/op
[info] HotScalacBenchmark.compile:compile·p0.50                           sample       1136.656          ms/op
[info] HotScalacBenchmark.compile:compile·p0.90                           sample       1165.597          ms/op
[info] HotScalacBenchmark.compile:compile·p0.95                           sample       1184.471          ms/op
[info] HotScalacBenchmark.compile:compile·p0.99                           sample       1255.187          ms/op
[info] HotScalacBenchmark.compile:compile·p0.999                          sample       1272.971          ms/op
[...]

PR #6060, opened 2017-09-06
Spread transform/traverse/mapOver into virtual methods of Tree/Type

retronym:
Analysis of JIT logs shows that a handful of very hot pattern matches in tree and type traversal run through ~10 deoptimizations before finding a steady state. Even once JITted, they show up in profiles as consuming a few percent of time, even though they are mostly just doing dispatching.

In this PR, I have changed them to virtual calls. Analysis of the JIT logs shows better stability for these methods, and the benchmarks suggest that the compiler has better peak performance.


retronym:
Benchmarks scheduled to [appear soon](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=6d29764%7C556a56b897%7Cac0d7f0680%7C72c8868f18%7C6d2a6c2c1a%7Cb1f0a40f79&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1349846730990&to=1504675530990). I measured approx 4% speedup for the `scalap` benchmark locally.

retronym:
Reviving #5906, although I'd forgotten about the test failures that need investigation ☹️

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -1129,6 +1131,7 @@ trait Types
     override def safeToString: String = "<error>"
     override def narrow: Type = this
     override def kind = "ErrorType"
+    override def mapOver(map: TypeMap): Type = this
```
lrytz: There are many `override def mapOver(map: TypeMap): Type = this` added in this commit, and they are unnecessary in principle. I suggest to either make `mapOver` in `Type` abstract or throw, or remove all redundant overrides.

lrytz (commented):
Looks good! Can you provide some motivation in the message of the commit that reverts c2dc346?

sjrd:
I see a lot of
```
-          super.transform(tree)
+          tree.transform(this)
```
in transforms. Are these mandatory with the new scheme? Will this likely break compiler plugins?

lrytz:
They continue to work, transformers still have a `transform` method that forwards to `tree.transform`.

sjrd:
Cool thanks :)

retronym:
> Looks good! Can you provide some motivation in the message of the commit that reverts c2dc346?

I was aiming for cleaner stack traces in the transformers in order to make it a little easier to navigate through profiles. It didn't seem to make a difference to performance.

src/reflect/scala/reflect/internal/Types.scala
retronym: Hmm, I can't remember the sequence of how I arrived at this. I do remember running into problems when `Type` became a SAM and the inliner (prior to #5895) was overzealous. Maybe I made this non-abstract as a workaround.

retronym:
Anything more needed by the reviewers?

adriaanm:
There's a (trivial) conflict. I resolved and moved the revert to the last commit to make it easier to adapt the commit message. I didn't want to push -f to your branch. My branch is at https://github.com/adriaanm/scala/tree/faster/virtual-transform-minimal

adriaanm:
After that, LGTM

retronym:
Added a merge commit to resolve the conflict with #6089

retronym:
Something was wonky with Jenkins in https://scala-ci.typesafe.com/job/scala-2.13.x-validate-main/567/console, the build flow claimed it was waiting for the integrate-ide job, but that has already finished.

retronym:
Merging on Adriaan's conditional LGTM.

PR #6062, opened 2017-09-06
Use AnyRefMap in pickler, we don't actually need a linked map

retronym:
AnyRefMap is quite a bit faster as it avoids cooperative equality and doesn't maintain the linked list of entries in insertion order.

The linked-ness of this particular map dates back to everyone's favourite commit, "Massive check-in for IDE." a205b6b0. I don't see any motivation for it, as the contents are never iterated.

Given that a number of maps were made linked in that commit, I conclude that this one was done speculatively while developing that change, but wasn't needed.

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
@@ -9,10 +9,11 @@ package classfile
 
 import java.lang.Float.floatToIntBits
 import java.lang.Double.doubleToLongBits
+
 import scala.io.Codec
-import scala.reflect.internal.pickling.{ PickleBuffer, PickleFormat }
+import scala.reflect.internal.pickling.{PickleBuffer, PickleFormat}
 import scala.reflect.internal.util.shortClassOfInstance
-import scala.collection.mutable.LinkedHashMap
+import scala.collection.mutable
 import PickleFormat._
 import Flags._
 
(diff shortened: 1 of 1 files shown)


retronym:
`LinkedHashMap` suffers from the performance problems that `mutable.HashMap` used to have (at least, before #5528 et al), and incurs the cooperative equality overhead that `AnyRefMap` doesn't.

The linked-ness of this particular map dates back to everone's favourite commit, ["Massive check-in for IDE."](https://github.com/scala/scala/commit/a205b6b06e705711308c9ad3abac74ba66b6266f#diff-c455cb229f5227b1bcaa1544478fe3acL59). I don't see an obvious motivation for it, the contents never appear to be iterated.

So we might just be able to switch to `AnyRefMap` here.

milessabin:
It's possible that at one time that map was passed to Eclipse plugin code and was wanted to be more Java friendly. My memory is (thankfully) very hazy ...

retronym:
It was private at the time, and not obviously passed outside of the class. `LinkedHashMap` featured heavily in the other parts of the commit.

retronym:
[Benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=a28305%7Cebe72d73%7Cdea7d27&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1503533376290&to=1504742976290)

retronym:
The force-pushed commit here is identical to dea7d27 in the benchmark results.

lrytz (approved):

PR #6063, opened 2017-09-06
allow custom indices start in zipWithIndex

Dveim:
Motivation -- sometimes I want to use other than 0 start index in `zipWithIndex`. Some alternatives -- implicit class or manual addition, like `for ((x, _i) <- list.zipWithIndex; i = _i + start) ...`. However, probably would be useful to have this functionality built in stdlib, to reduce overhead and simplify user code.

Example:
```
scala> List(1, 2, 3).zipWithIndex(42)
res0: List[(Int, Int)] = List((1,42), (2,43), (3,44))
```

Goals:
1. Already existing syntax `list.zipWithIndex ...` must be preserved, so no user code should be changed.
2. New method must have the same name, so it looks like `list.zipWithIndex(42) ...`.
3. `list.zipWithIndex() ...` must behave exactly the same as `list.zipWithIndex ...`.

In current implementation, there are few debatable decisions:
1. To achieve first goal, I had to create explicit overloaded `zipWithIndex` method. That slightly complicates collection's code. However, I don't know any alternative.
2. `zipWithIndex` calls `zipWithIndex(0)`, thus +1 method call. That can be overcome with copy-pasted implementation of `zipWithIndex` in both cases, however, runtime should optimize it away.

Any comments and suggestions are welcome.

### DIFF
--- a/src/library/scala/collection/GenIterableLike.scala
+++ b/src/library/scala/collection/GenIterableLike.scala
@@ -96,7 +96,7 @@ trait GenIterableLike[+A, +Repr] extends Any with GenTraversableLike[A, Repr] {
    *                 result class `That` from the current representation type `Repr`
    *                 and the new element type `(A1, Int)`.
    *  @return        A new collection of type `That` containing pairs consisting of all elements of this
-   *                 $coll paired with their index. Indices start at `0`.
+   *                 $coll paired with their index. Indices start at `indicesStart`.
    *
    *  @usecase def zipWithIndex: $Coll[(A, Int)]
    *    @inheritdoc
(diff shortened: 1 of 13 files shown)


src/library/scala/collection/GenIterableLike.scala
```diff
@@ -96,20 +96,23 @@ trait GenIterableLike[+A, +Repr] extends Any with GenTraversableLike[A, Repr] {
    *                 result class `That` from the current representation type `Repr`
    *                 and the new element type `(A1, Int)`.
    *  @return        A new collection of type `That` containing pairs consisting of all elements of this
-   *                 $coll paired with their index. Indices start at `0`.
+   *                 $coll paired with their index. Indices start at `indicesStart`.
```
arbitrary-dev, line 99: I'd suggest renaming `indicesStart` to just `start`.

arbitrary-dev (changes requested):

src/library/scala/collection/GenIterableLike.scala
Dveim: I bluntly copied that name from docs, was said `Indices start at 0`.

Don't mind to change it to `start` or `startFrom` or similar, if community prefer that.

hrhino:
Sadly you're going to have to make this change in 2.13.x; it's not forward [binary compatible](https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#code) with 2.12.x (code that relies on this method can't be used with an older 2.12 library).

The 2.13 collections are being overhauled in [scala/collection-strawman](https://github.com/scala/collection-strawman); you should contribute this change over there instead.

Dveim:
@hrhino Missed point about forward binary compatibility. Also, https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/5520/console wasn't run locally, how can I run full test suite before submitting?

Well, I was going to submit PR with same functionality in new collections anyway.

hrhino:
The full set of tests (which takes an hour or so on the rather intimidating Jenkins box) can be run with [`sbt testAll`](https://github.com/scala/scala/blob/be961f9a96bd69f7befea28e5319021606a9f6ae/build.sbt#L811-L830).

If you're only touching library things, you can usually get away with
```
;test/junit:test;test/scalacheck:test;library/mimaReportBinaryIssues
```
which will run the unit tests and the MiMa tests.

Technically the PR validation builds the compiler with a stable scala version, and then re-builds and tests everything with _that_ compiler, but if you're only fiddling with the library it shouldn't matter.

PR #6064, opened 2017-09-07
Don't serve stale JAR metadata from cache in classpath implementation

retronym:
Use the last modified timestamp and the file inode to help detect
when the file has been overwritten (as is common in SBT builds
with `exportJars := true`, or when using snapshot dependencies).

Fixes scala/bug#10295

In addition to the enclosed unit test, I manually tested that
the manifestation of this problem reported in sbt/zinc#282
is solved.

```
% git clone sbt/zinc; cd zinc

% git co 3e9f23ae158240c4077b2d9a50cb2b5eafa8cab7~1

% sbt

> ++2.12.4-bin-a966b7d-SNAPSHOT

> compile

> # edit types.json

> compile
[error] one error found
[error] /Users/jz/code/zinc/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala:438: not enough arguments for constructor Projection: (x$1: xsbti.api.Type, x$2: String, x$3: String)xsbti.api.Projection.
[error] Unspecified value parameter x$3.
[error]       case None => new api.Projection(Empty, cls)
[error]                    ^
```

### DIFF
--- a/src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
+++ b/src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
@@ -5,38 +5,32 @@ package scala.tools.nsc.classpath
 
 import java.io.File
 import java.net.URL
+import java.nio.file.Files
+import java.nio.file.attribute.{BasicFileAttributes, FileTime}
 
 import scala.annotation.tailrec
-import scala.reflect.io.{ AbstractFile, FileZipArchive, ManifestResources }
+import scala.reflect.io.{AbstractFile, FileZipArchive, ManifestResources}
 import scala.tools.nsc.util.{ClassPath, ClassRepresentation}
 import scala.tools.nsc.Settings
 import FileUtils._
 
 /**
  * A trait providing an optional cache for classpath entries obtained from zip and jar files.
- * It's possible to create such a cache assuming that entries in such files won't change (at
- * least will be the same each time we'll load classpath during the lifetime of JVM process)
- * - unlike class and source files in directories, which can be modified and recompiled.
  * It allows us to e.g. reduce significantly memory used by PresentationCompilers in Scala IDE
  * when there are a lot of projects having a lot of common dependencies.
  */
 sealed trait ZipAndJarFileLookupFactory {
-  private val cache = collection.mutable.Map.empty[AbstractFile, ClassPath]
+  private val cache = new FileBasedCache[ClassPath]
 
   def create(zipFile: AbstractFile, settings: Settings): ClassPath = {
-    if (settings.YdisableFlatCpCaching) createForZipFile(zipFile)
+    if (settings.YdisableFlatCpCaching || zipFile.file == null) createForZipFile(zipFile)
     else createUsingCache(zipFile, settings)
   }
 
   protected def createForZipFile(zipFile: AbstractFile): ClassPath
 
-  private def createUsingCache(zipFile: AbstractFile, settings: Settings): ClassPath = cache.synchronized {
-    def newClassPathInstance = {
-      if (settings.verbose || settings.Ylogcp)
-        println(s"$zipFile is not yet in the classpath cache")
-      createForZipFile(zipFile)
-    }
-    cache.getOrElseUpdate(zipFile, newClassPathInstance)
+  private def createUsingCache(zipFile: AbstractFile, settings: Settings): ClassPath = {
+    cache.getOrCreate(zipFile.file.toPath, () => createForZipFile(zipFile))
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
```diff
@@ -5,38 +5,32 @@ package scala.tools.nsc.classpath
...
 
-  private def createUsingCache(zipFile: AbstractFile, settings: Settings): ClassPath = cache.synchronized {
-    def newClassPathInstance = {
-      if (settings.verbose || settings.Ylogcp)
-        println(s"$zipFile is not yet in the classpath cache")
-      createForZipFile(zipFile)
-    }
-    cache.getOrElseUpdate(zipFile, newClassPathInstance)
+  private def createUsingCache(zipFile: AbstractFile, settings: Settings): ClassPath = {
+    cache.getOrCreate(zipFile.file.toPath, () => createForZipFile(zipFile))
```
retronym, line 33: Removed needlessly noisy logging here.

src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
```diff
@@ -181,3 +175,27 @@ object ZipAndJarSourcePathFactory extends ZipAndJarFileLookupFactory {
 
   override protected def createForZipFile(zipFile: AbstractFile): ClassPath = ZipArchiveSourcePath(zipFile.file)
 }
+
+final class FileBasedCache[T] {
+  private case class Stamp(lastModified: FileTime, fileKey: Object)
+  private val cache = collection.mutable.Map.empty[java.nio.file.Path, (Stamp, T)]
+
+  def getOrCreate(path: java.nio.file.Path, create: () => T): T = cache.synchronized {
+    val lastModified = Files.getLastModifiedTime(path)
```
lrytz: it seems you can avoid calling `getLastModifiedTime ` here (which calls `Files.readAttributes`internally) and use `attrs.lastModifiedTime()` instead.

src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
```diff
@@ -181,3 +175,27 @@ object ZipAndJarSourcePathFactory extends ZipAndJarFileLookupFactory {
...
+    val attrs = Files.readAttributes(path, classOf[BasicFileAttributes])
+    val stamp = Stamp(lastModified, attrs.fileKey())
```
lrytz: It seems `attrs.fileKey` is null on windows, at least on the version i checked. I guess that's fine, `lastModified` should be enough? Should we check still keep the `fileKey` check for other platforms?

retronym: Yes, on other platforms we fall back to just the timestamp check. I could declare it as `Option[Object]` to make it clearer that we've thought about this.

lrytz: It's fine without an Option, a commend will do.

retronym: pushed a comment

src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
lrytz: this could still be addressed :)

retronym: The latest commit also uses `attrs.lastModifiedTime()`

lrytz: Sorry, my bad, I mis-read.

lrytz (approved):

lrytz:
Very nice to have this fixed, thanks!

lrytz:
At some point in the future we can hopefully remove the `-YdisableFlatCpCaching` setting.

retronym:
> At some point in the future we can hopefully remove the `-YdisableFlatCpCaching` setting.

Yep, that would be nice. I can think of two reasons that folks might still want to disable the caching:
  - With caching enabled, a long lived classloader for `Global` will retain strong references to JAR files, which prevents even `File.finalize` from closing the file descriptor. This causes problems with inability to overwrite the JAR on windows, and file descriptor leaks for use cases that use lots of different JARs over time.
  - In addition to the file descriptor leak, we also leak the datastructures we are caching.

So we need to have a better story for resource management here, maybe involving reference counting ("are any instance of `Global` actually using this right now?") and timed eviction ("when was the last time this was used?").

Those things aren't straightforward to retrofit onto a `Global` and `Run` that don't currently have a `close()` method, but are within the purview of scala/scala-dev#416.

gkossakowski:
We've hit this bug at Stripe while testing 2.12 with bazel. The incremental compilation didn't work for us. Bazel always puts all class files in a jar before passing them to a dependent target (bazel's name for something like "subproject").

Bazel makes a heavy use of symlinks and I'm wondering if cache invalidation implemented in the PR will handle them properly. I'll see whether I can get a nightly distribution build to work with bazel and verify whether 2.12's HEAD fixes this issue for us.

gkossakowski:
I just verified with the latest Scala nightly: http://www.scala-lang.org/files/archive/nightly/2.12.x/scala-2.12.4-bin-c1c9c7c.tgz

The bug with under compilation in bazel goes away. This is great.

PR #6066, opened 2017-09-11
Avoid an AST traversal in the erasure phase

retronym:
A separate traversal between preErase and re-typechecking was
responsible for filling in super-init calls in trait constructors
and adjusting the template parents in the AST to align with the
info-transformed class info.

This commit removes this traversal and handles those concerns in
the typechecking traversal

This was probably harder to do before the changes to the trait encoding
in Scala 2.12.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Erasure.scala
+++ b/src/compiler/scala/tools/nsc/transform/Erasure.scala
@@ -398,48 +398,32 @@ abstract class Erasure extends InfoTransform
 
   class UnknownSig extends Exception
 
-  // TODO: move to constructors?
-  object mixinTransformer extends Transformer {
-    /** Add calls to supermixin constructors
-      *    `super[mix].$init$()`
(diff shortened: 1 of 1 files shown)


retronym:
The traversal I've removed was reported at around 0.5% of compile time in my profiler.

Let's see if we can see any of that as an improvement in the [benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=e1e8d050d%7Cd9ebf76&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1504043350594&to=1505118187562)

retronym:
The benchmarks results appear to be either very slightly positive or neutral. Ideally, we'd be able to benchmark the erasure phase in isolation to be able to resolve the impact of small changes like this.

adriaanm (approved):
Regardless of performance, this is a nice cleanup.

PR #6067, opened 2017-09-11
Add hooks, counters and refine the statistics infrastructure

jvican:
This is a continuation of my work for SCP-010. This pull request is the responsible of polishing the statistics framework, adding missing hooks for our Scala plugin scalac-profiling and double-checking the general state of statistics.

As a result, this pull request ends up doing several independent tasks:
1. It instruments scopes so that we can analyze them later on.
1. It adds timers for all the phases (`-Ystatistics` will report how much time do phases take).
1. Adds hooks to `AnalyzerPlugins` to inspect implicit search data and therefore allow external plugins to turn it into something meaningful.
1. Polishes some parts of the user interface of statistics (flags, descriptions).
1. Fixes/cleans some parts of the infrastructure that is unstable or buggy.

Depends on #6051 and #6034.


src/reflect/scala/reflect/internal/Scopes.scala
```diff
@@ -8,9 +8,12 @@ package reflect
 package internal
 
 import scala.annotation.tailrec
+import scala.reflect.internal.util.Statistics
 
 trait Scopes extends api.Scopes { self: SymbolTable =>
 
+  private[scala] var scopeCount = 0
```
lrytz: Should this be reset to 0 after each run? That could be achieved with `perRunCaches.newGeneric`.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -1431,26 +1431,31 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
       warnDeprecatedAndConflictingSettings()
       globalPhase = fromPhase
 
+      val timePhases = statistics.canEnable || settings.verbose
+      val startTotal = if (timePhases) statistics.startTimer(totalCompileTime) else null
+
       while (globalPhase.hasNext && !reporter.hasErrors) {
-        val startTime = currentTime
         phase = globalPhase
+        val phaseTimer = statistics.newSubTimer(s"  ${phase.name}", totalCompileTime)
```
lrytz: Can we make this `null` if `! timePhases`?

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -1431,26 +1431,31 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
...
+
         val profileBefore=profiler.beforePhase(phase)
-        globalPhase.run()
+        try globalPhase.run()
+        finally if (timePhases) statistics.stopTimer(phaseTimer, startPhase) else ()
         profiler.afterPhase(phase, profileBefore)
 
         // progress update
-        informTime(globalPhase.description, startTime)
+        informTime(globalPhase.description, phaseTimer.nanos)
```
lrytz: should this be guarded by `if (timePhases)`?

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -1503,7 +1508,13 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
       }
       symSource.keys foreach (x => resetPackageClass(x.owner))
 
-      informTime("total", startTime)
+      if (timePhases) {
+        statistics.stopTimer(totalCompileTime, startTotal)
+        informTime("total", totalCompileTime.nanos)
+        inform("*** Cumulative timers for phases")
+        for (q <- statistics.allQuantities if q.phases.nonEmpty && q.showAt(GlobalPhaseName))
```
lrytz: Nit: the guard `q.phases.nonEmpty && q.showAt(GlobalPhaseName)` is non-obvious to understand - how about the more explicit `q.phases == Seq(GlobalPhaseName)`?

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1578,6 +1581,7 @@ trait ImplicitsStats {
   val matchingImplicits   = newSubCounter("  #matching", implicitSearchCount)
   val typedImplicits      = newSubCounter("  #typed", implicitSearchCount)
   val foundImplicits      = newSubCounter("  #found", implicitSearchCount)
+  val macroImplicits      = newSubCounter("  #macros", implicitSearchCount)
```
lrytz: Can we give it a more telling name, e.g. `#triggered during macro expansion`?

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1581,4 +1591,8 @@ trait ImplicitsStats {
   val matchesPtNanos      = newSubTimer  ("  matchesPT", typerNanos)
   val implicitCacheAccs   = newCounter   ("implicit cache accesses", "typer")
   val implicitCacheHits   = newSubCounter("implicit cache hits", implicitCacheAccs)
+
+  import scala.reflect.internal.util.Position
+  final val implicitSearchesByType = symbolTable.perRunCaches.newMap[global.Type, Int]()
+  final val implicitSearchesByPos = symbolTable.perRunCaches.newMap[Position, Int]()
```
lrytz: Instead of collecting this data within the compiler (this data is collected when ever statistics are enabled, but only used by your compiler plugin), I'd prefer some mechanism to allow a compiler plugin to listen for certain events. See [`AnalyzerPlugins`](https://github.com/scala/scala/blob/v2.12.3/src/compiler/scala/tools/nsc/typechecker/AnalyzerPlugins.scala#L281-L288) for an example. Here it would be simpler than with analyzer plugins, because it's only about listening for notifications and not changing the result of any method call.

I'm open for other suggestions.

src/reflect/scala/reflect/internal/Scopes.scala
jvican: Hm, yes, it should be reset to 0 after every run. I wanted to emulate trees ids here, that's why I used just one counter.

src/compiler/scala/tools/nsc/Global.scala
jvican: We can use this because `GlobalPhaseName` will always be the only phase, but it's more restrictive semantically.

src/reflect/scala/reflect/internal/Scopes.scala
jvican: I didn't figure out how to do it with `newGeneric`, so I implemented with a clearable. See updated PR. Thanks for the suggestion.

src/reflect/scala/reflect/internal/Scopes.scala
```diff
@@ -8,12 +8,21 @@ package reflect
...
 import scala.annotation.tailrec
+import scala.collection.generic.Clearable
 import scala.reflect.internal.util.Statistics
 
 trait Scopes extends api.Scopes { self: SymbolTable =>
 
   private[scala] var scopeCount = 0
 
+  // Reset scopeCount per every run
+  perRunCaches.recordCache(new Clearable {
```
lrytz: you could use a lambda, `Clearable` is a SAM type

src/reflect/scala/reflect/internal/Scopes.scala
lrytz: OK, `newGeneric` was probably not a good suggestion here

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
lrytz: This was not addressed. The goal here is to provide a hook for compiler plugins, so I really feel that it's not great to collect some very specific data inside the compiler itself.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
lrytz: This was not addressed in your latest update.

lrytz (commented):
For some reason two of my earlier comments showed up collapsed on the PR discussion even though they were up to date. Maybe a bug in GitHub.
  - https://github.com/scala/scala/pull/6067/commits/69ef9a50671544b501dd5393fe8b83e42b808622#r140470518
  - https://github.com/scala/scala/pull/6067/commits/5e57b299f721cc42303eabbd9fb2a7f00502d24d#r140473690

jvican:
Indeed, it seems I missed completely these two comments.

src/reflect/scala/reflect/internal/Scopes.scala
jvican: I tried -- it didn't work because the compiler failed to create a SAM type because `recordCache`'s signature expects `T <: Clearable`. It seems SAM creation doesn't work for these cases (or so I think).

lrytz: ah right, you'd need a type annotation `(() => { scopeCount = 0 }): Clearable`

jvican: What I ended up doing locally is creating a new private val with the closure. IMO it's better than the type annotation. If you don't have any objection, I'll continue this way.

lrytz: Either way is fine. A local val would be better than a private val in `Scopes`, as the latter creates a field.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
jvican: Hey Lukas, I don't have a clear idea how we could engineer a general interface to listen for general events in the compiler that adds more value than this current implementation.
  
The problem I see here is that these hooks are very explicit and require instrumentation per ever kind of event (in different parts of the compiler).
  
In `AnalyzerPlugins`, I see a more general interface that allows plugins to refine type signatures in certain places, giving a feeling that analyzer plugins can complement typechecking. I don't see a way we can do this for explicit event listening because this requires:
  
* The creation of the timers and subtimers.
* Exposing these timers and subtimers explicitly in the `statistics` field, so that they are accessible from any call-site of the compiler.
* Instrumenting one call-site per ever event you want to listen.
  
Maybe I'm missing something?

lrytz:
Some of the commits have the following test failure:
```
!! 1278 - pos/t8531                                 [compilation failed]
##### Log file '/home/jenkins/workspace/scala-2.12.x-validate-test/test/files/pos/t8531-pos.log' from failed test #####

error: null
```

The fact that it happens on some commits but not on all suggests it could be a race condition (AFAIK, partest runs multiple `Global` instances in parallel).

adriaanm:
Looks more likely this will slip to 2.12.5 then.

jvican:
@lrytz I would be really surprised if this had any thread safety issues as it just adds some counters and timers. If this code had any thread safety issues, all the statistics infrastructure would, which I find pretty unlikely given the code and my experience using/hacking on it.
  
I think the problem here is that some commits were rebased and there were some issues with this rebase. Let me look into this and quickly push the new changes.
  
> Looks more likely this will slip to 2.12.5 then.
  
I would really love if this can be merged after my changes have been done and all commits are green. This change is non controversial -- it's not like my other PR that makes a fundamental change to the infrastructure.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
retronym: I think that implicit search should be instrumented here in the compiler to keep track of some important statistics and aggregate these into `SearchResult`. The most important stat is the overall number of `typedImplicit1` calls made within an implicit search tree. Perhaps the number of implicit macros expanded is also interesting. We should also have an easy way for the tools which process the data to distinguish "top level" implicit searches from recursive ones.

A new extension point in `AnalyzerPlugin`-s could be added to listen on the result of `inferImplicit`, and then the work to slice, dice, and present this data could be left in the hands of the plugin.

lrytz: Thanks for the input. If this PR can still make it for 2.12.4 I'm happy to have a minimal solution here for now and implement a more structured approach later. Would that work for you, @retronym? Would you be OK with the two hash maps?

src/reflect/scala/reflect/internal/Scopes.scala
```diff
@@ -229,36 +239,41 @@ trait Scopes extends api.Scopes { self: SymbolTable =>
     /** Lookup a symbol.
      */
     def lookup(name: Name): Symbol = {
+      val startTime = if (statistics.canEnable) statistics.startTimer(statistics.scopeLookupTime) else null
```
retronym: I don't think this is really a suitable place to instrument, as it misses a lot of important lookups that use alternative scope lookup APIs. E.g. `Context.isQualifyingImplicit` uses `lookupEntry` and `Context.lookupSymbol` uses `lookupUnshadowedEntries`.

retronym: Oh, I see you've instrumented `lookupEntry` as well below. Should we then start this timer below the call to `lookupEntry` to avoid double counting?

lrytz:
I didn't spot any issues from rebasing when I looked at the commits. The error message in the test failure is as unhelpful as it gets (`error: null` without *any* context), that's really annoying.

`t8531` is about the pattern matcher / exhaustivity analysis, but I'm not sure that is a hint in the right direction.

retronym:
> that's really annoying.

I can get that deficient error message by throwing an NPE in the compiler, or any exception with a null message.

I'm guessing that partest is culpable for the old favourite, [`println(throwable.getMessage)`](https://github.com/scala/scala-partest/blob/4d2b327b48da98553f8efd2e3bbc55052d68642d/src/main/scala/scala/tools/partest/nest/DirectCompiler.scala#L124). But I'm not 100% sure, we've got similarly horrifying exception handling in the [compiler itself](https://github.com/scala/scala/blob/a74f0c7a1457b8a17db36db25d068dc1583c575e/src/compiler/scala/tools/nsc/Global.scala#L1508).

A quick fix to diagnose this would be to add a commit that changes `Run.compileInternal` to catch `Throwable` and print the results to system out. That should show up in the partest log.

src/reflect/scala/reflect/internal/Scopes.scala
jvican: I instrumented both so that `lookupEntry` also records the time spent in executing its whole body, which relies on `lookupNextEntry`. I think that the statistics infrastructure prevents double counting by design (nested calls do not record the times if there are other timers wrapping them) -> a fresh pair of eyes to confirm this would be good.
  
However, maybe measuring `lookup` is uninteresting... so I've removed it in my last commit.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
jvican: @retronym @lrytz This sounds like a good idea. I'm implementing it real quick. However, I'll add two hooks into implicit search, one before the search and the other after the search. This will help me add timers and not only counters.

jvican: Implemented in https://github.com/scala/scala/pull/6067/commits/41d60f836be54a7010c3656e5e28107451bfea54.

jvican:
@lrytz @retronym Can you have a look at the new `AnalyzerPlugin` hooks?
  
Everything is green and ready for a review.

jvican:
Benchmark scheduled [here](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=8a92ebdce5|1be1eb7bee&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=now-30d&to=now).

adriaanm:

jvican:
@adriaanm Sounds good. I'll run the community build before you do to make sure all my PRs are ok.

retronym:
@jvican Did you figure out the root cause of that null-messaged exception? If not, I'm a bit worried that either this PR or something recent has introduced an non-deterministic error that we need to track down before 2.12.4.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -90,7 +90,9 @@ trait Implicits {
     if (shouldPrint)
       typingStack.printTyping(tree, "typing implicit: %s %s".format(tree, context.undetparamsString))
     val implicitSearchContext = context.makeImplicit(reportAmbiguous)
-    val result = new ImplicitSearch(tree, pt, isView, implicitSearchContext, pos).bestImplicit
+    val search0 = new ImplicitSearch(tree, pt, isView, implicitSearchContext, pos)
+    val search = pluginsImplicitSearch(search0)
+    val result = pluginsImplicitSearchResult(search.bestImplicit)
```
retronym: I'm not really comfortable letting plugins change the results of an implicit search. A less powerful extension point would just have `pluginsImplicitSearchResult(search: ImplicitSearch): Unit`.

The standard `ImplicitSearch` can itself contain a timer, which is conditionally computed if stats are on. We don't need to leave that part to the plugin.

jvican: > I'm not really comfortable letting plugins change the results of an implicit search. A less powerful extension point would just have pluginsImplicitSearchResult(search: ImplicitSearch): Unit.
  
I've adapted it to be a simple hook. `AnalyzerPlugins` is itself a powerful tool that refines types from the typechecker, so I wasn't sure that such a powerful extension point would not be accepted since it aligns with the overall idea of the plugin.
  
> The standard ImplicitSearch can itself contain a timer, which is conditionally computed if stats are on. We don't need to leave that part to the plugin.
  
We already have a timer for this, e.g. `implicitNanos`, and I can access it statically in my compiler plugin, so I think this is not necessary.
  
However, I didn't realize that we have another `ImplicitSearch` for views. They should mostly be unused, but just in case I've added a timer for them too.

src/compiler/scala/tools/nsc/typechecker/AnalyzerPlugins.scala
```diff
@@ -349,6 +369,18 @@ trait AnalyzerPlugins { self: Analyzer =>
     def accumulate = (tpe, p) => p.pluginsTypedReturn(tpe, typer, tree, pt)
   })
 
+  /** @see AnalyzerPlugin.pluginsImplicitSearch */
+  def pluginsImplicitSearch(search: ImplicitSearch): ImplicitSearch = invoke(new CumulativeOp[Unit] {
```
lrytz: Should have return type `Unit` (also below)

jvican: Yes, I pushed a bad commit.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1585,6 +1590,7 @@ trait ImplicitsStats {
   val improvesCachedCount = newSubCounter("  #implicit improves cached ", implicitSearchCount)
   val inscopeImplicitHits = newSubCounter("  #implicit inscope hits", implicitSearchCount)
   val oftypeImplicitHits  = newSubCounter("  #implicit oftype hits ", implicitSearchCount)
+  val viewsNanos          = newSubTimer  ("time spent in views", typerNanos)
```
lrytz: I think this description is a bit misleading: `allViewsFrom` seems to be used only by scaladoc, the compiler uses `inferImplicit` with `isView = true`.

lrytz (commented):
It looks good to me now. I'm also still a bit worried about the null-message. But looking through the commits I really cannot see anything suspicious.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
jvican: Intersting, didn't know this. Then I'm removing it.

jvican:
> It looks good to me now. I'm also still a bit worried about the null-message. But looking through the commits I really cannot see anything suspicious.
  
Me neither... I've gone through my previous PRs too and didn't find anything suspicious re thread safety. I'll go through them again, just in case there's something hidden.
  
I'm integrating all the changes with my scala-profiling integration, and double-checking that everything works. I'll be running the community build with only my changes to check that there's no issue. In a few hours, I'll have some results to show.

jvican:
Community build is still running on my cluster, but so far I'm getting no errors.

jvican:
Today I've spent almost all day double-checking this PR and [integrating it with scalac-profiling](https://github.com/scalacenter/scalac-profiling/pull/7). This process has given me confidence that I have everything I need from the compiler after the exhaustive feedback that I received in this and the previous PRs.
  
These are the tasks I've done to double-check this PR is in good shape:
  
* Compile both the library and the compiler with the binaries and the following flags: `-Ystatistics`, `-Yhot-statistics-enabled`. Everything ok.
* Enabled `-Ystatistics` in one project in an sbt multi-build and double-checked that only that project outputs statistics (the other one doesn't, despite the fact that they share the same classloader). Everythink ok.
* Executed different projects with statistics enabled in parallel. Everything ok.
* Run the community build in our cluster (it's still running but so far everything passes).
  
As a result of this process, I've updated this PR with minor fixes to the UI and the semantics of the whole statistics framework. I'm fairly confident this PR doesn't represent a problem for the upcoming release.
  
I would appreciate if someone can go through it again and tell me if it looks good to you.
  
Ping me during the weekend if you have any question regarding my PRs.

adriaanm:
Thanks for the due diligence, @jvican. I'll let @retronym take a look his Monday morning and hopefully get this merged by my Sunday night :-)

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -384,7 +384,7 @@ trait ScalaSettings extends AbsScalaSettings
   }
   override def YstatisticsEnabled = Ystatistics.value.nonEmpty
 
-  val YhotStatistics = BooleanSetting("-Yhot-statistics", "Print hot compiler statistics for all relevant phases")
+  val YhotStatistics = BooleanSetting("-Yhot-statistics-enabled", s"Enable `${Ystatistics.name}` to print hot statistics.")
```
lrytz, line 387: You could make `-Ystatistics` a `ChoiceSetting` (`-Ystatistics:{none|cold|all}`), but that's not a blocker for merging this PR.

lrytz (approved):

rorygraves:
Woot!

PR #6068, opened 2017-09-11
Only cache ImplicitInfo erroneous status in isCyclicOrErroneous.

milessabin:
4aeb8acf99493a5b3f7c2e012796ded77ae40a7f slightly overreached as the cyclic error is context dependent and should not be globally cached.

Fixes scala/bug#9122


milessabin:
Community build output [here](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1906/) looks good apart from the cats-effect failure which also appears to occur on the vanilla 2.12.x nightly build.

retronym (approved):

PR #6069, opened 2017-09-11
Higher-kinded type variable unification.

TomasMikula:
Currently, when checking whether `[α]Foo[Bar[α]]` is a subtype of (higher-kinded) type variable `?F`, the current implementation eta-expands `F` to `[β]F[β]`, substitutes the same type parameters `[α]F[α]` and then checks whether `Foo[Bar[α]] <: F[α]`. Then (I think) `F` is unified with `Foo` and `α` fails to unify with `Bar[α]`.

This PR changes that to immediately unify `F` with `[α]Foo[Bar[α]]` if `α` is compatible with `F`'s type parameter.

This change can cause ambiguous implicits, so is under a compiler flag
~`-Yhk-typevar-unification`~ `-Xsource:2.13`.

Fixes scala/bug#10185,  scala/bug#10195, scala/bug#10197, scala/bug#10213, scala/bug#10238, scala/bug#10372.
Also fixes scala/bug#6895 in a different way than `-Ypartial-unification`.


TomasMikula:
The build failed due to binary compatibility issues. How can I run the binary compatibility test locally?

lrytz:
If you want to run all tests you can use `testAll` in sbt. For MiMa specifically, use `mimaReportBinaryIssues`.

TomasMikula:
Thanks. I get

```
...
[info] scala-compiler: previous-artifact not set, not analyzing binary compatibility
...
```

lrytz:
That's expected, we only check binary compatibility for scala-library and scala-reflect. Minor versions of the scala-compiler jar are not binary compatible.

TomasMikula:
Oh, right, the incompatible change was in `scala-reflect` and my speculative "fix"

```scala
ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.runtime.Settings.YhkTypevarUnification")
```

worked, so hopefully the current build will succeed.

It also took me a while to figure out why `mimaReportBinaryIssues` kept crashing with

```
java.nio.charset.MalformedInputException: Input length = 1
```

Turns out I cannot have the file `src/reflect/mima-filters/2.12.0.forwards.excludes` open in `vim`, because it creates a temporary `.2.12.0.forwards.excludes.swp` file in the same directory, and `mimaReportBinaryIssues` apparently doesn't like that.

non:
Great work @TomasMikula! This is really exciting! ✨

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,9 +365,23 @@ trait TypeComparers {
...
+      case (tp1, tv2 @ TypeVar(_, _)) if settings.YhkTypevarUnification =>
+        val ntp1 = tp1.normalize
+        (tv2.params corresponds ntp1.typeParams)(methodHigherOrderTypeParamsSubVariance) &&
+        { tv2.addLoBound(ntp1); true }
+      case (tv1 @ TypeVar(_, _), tp2) if settings.YhkTypevarUnification =>
+        val ntp2 = tp2.normalize
+        (ntp2.typeParams corresponds tv1.params)(methodHigherOrderTypeParamsSubVariance) &&
+        { tv1.addHiBound(ntp2); true }
+      case (tp1, tp2) => isSub2(tp1.normalize, tp2.normalize)  // @M! normalize reduces higher-kinded case to PolyType's
+    }
```
retronym: The implementations of `<:<` and `=:=` usually follow the same pattern. I wonder if a similar change needed in:

```scala
  private def isSameHKTypes(tp1: Type, tp2: Type) = (
       tp1.isHigherKinded
    && tp2.isHigherKinded
    && (tp1.normalize =:= tp2.normalize)
  )
```

It would be great to setup a unit test in the same vein as https://github.com/scala/scala/blob/9acab45aeeadef2f63da69faf81465cc15599789/test/junit/scala/reflect/internal/TypesTest.scala  to directly demonstrate the new unification (rather than indirectly via compilation), and to explore whether the change preserves transitivity of subtyping and consistency with `=:=`.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,9 +365,23 @@ trait TypeComparers {
 
   // @assume tp1.isHigherKinded || tp2.isHigherKinded
   def isHKSubType(tp1: Type, tp2: Type, depth: Depth): Boolean = {
-    def isSub(ntp1: Type, ntp2: Type) = (ntp1.withoutAnnotations, ntp2.withoutAnnotations) match {
+    def isSub(tp1: Type, tp2: Type) = (tp1.withoutAnnotations, tp2.withoutAnnotations) match {
       case (TypeRef(_, AnyClass, _), _)                                     => false                    // avoid some warnings when Nothing/Any are on the other side
       case (_, TypeRef(_, NothingClass, _))                                 => false
```
retronym: Type aliases for `Any` and `Nothing` would evade the short circuit now that the caller doesn't normalize.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,9 +365,23 @@ trait TypeComparers {
...
+      case (tv1 @ TypeVar(_, _), tv2 @ TypeVar(_, _)) if settings.YhkTypevarUnification =>
+        (tv2.params corresponds tv1.params)(methodHigherOrderTypeParamsSubVariance) &&
+        { tv1.addHiBound(tv2); tv2.addLoBound(tv1); true }
+      case (tp1, tv2 @ TypeVar(_, _)) if settings.YhkTypevarUnification =>
+        val ntp1 = tp1.normalize
+        (tv2.params corresponds ntp1.typeParams)(methodHigherOrderTypeParamsSubVariance) &&
+        { tv2.addLoBound(ntp1); true }
```
retronym: The existing code in `TypeVar.registerBound` that responsible for unifying applied hk type vars walks up the base type sequence to find an arity match. What do we know about `tv2.params.length == ntp1.typeParams.length` by virtue of making it to this line of code?

TomasMikula: We know `tv2.params.length == ntp1.typeParams.length` thanks to a successful `corresponds` on the previous line.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
TomasMikula: Thanks, will move these 2 lines down to `isSub2`.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
TomasMikula: > I wonder if a similar change needed in:
>
> ```scala
> private def isSameHKTypes(tp1: Type, tp2: Type)
> ```

If `=:=` also performs unification as a side effect, then yes. Otherwise, I think it is not necessary.

TomasMikula:
Added some tests, some of them currently failing. Pending resolution of #6080 before I continue on this one.

Atry:
Does this PR solve https://stackoverflow.com/questions/32182694/how-to-define-an-existential-higher-kinded-type-in-scala?

TomasMikula:
@Atry That is not solved by this PR.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,7 +365,23 @@ trait TypeComparers {
 
   // @assume tp1.isHigherKinded || tp2.isHigherKinded
   def isHKSubType(tp1: Type, tp2: Type, depth: Depth): Boolean = {
-    def isSub(ntp1: Type, ntp2: Type) = (ntp1.withoutAnnotations, ntp2.withoutAnnotations) match {
+
+    def isSub(tp1: Type, tp2: Type) = (tp1.withoutAnnotations, tp2.withoutAnnotations) match {
+      case (tv1 @ TypeVar(_, _), tv2 @ TypeVar(_, _)) if settings.YhkTypevarUnification =>
```
adriaanm: I'd rather not add more individual feature flags, they end up sticking around and complicating the user experience. Let's just use -Xsource:2.13, and let's test that flag only once, instead of in each guard (pull out the more advanced type var unification into another method?)

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,7 +365,23 @@ trait TypeComparers {
...
+        (tv2.params corresponds tv1.params)(methodHigherOrderTypeParamsSubVariance) &&
+        { tv1.addHiBound(tv2); tv2.addLoBound(tv1); true }
```
adriaanm: Let's debug this and get some statistics on how (often) we actually get here during a bootstrap/test suite. As Jason said, inference that compares TVs on both ends is likely to fail in the end.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,7 +365,23 @@ trait TypeComparers {
 
   // @assume tp1.isHigherKinded || tp2.isHigherKinded
   def isHKSubType(tp1: Type, tp2: Type, depth: Depth): Boolean = {
-    def isSub(ntp1: Type, ntp2: Type) = (ntp1.withoutAnnotations, ntp2.withoutAnnotations) match {
+
+    def isSub(tp1: Type, tp2: Type) = (tp1.withoutAnnotations, tp2.withoutAnnotations) match {
+      case (tv1 @ TypeVar(_, _), tv2 @ TypeVar(_, _)) if settings.YhkTypevarUnification =>
+        (tv2.params corresponds tv1.params)(methodHigherOrderTypeParamsSubVariance) &&
```
adriaanm: It's kind of confusing that we check `methodHigherOrderTypeParamsSubVariance(tp2.typeParams, tp1.typeParams)` to compute `isSub(tp1, tp2)` (what's the logic behind going in the opposite direction?)

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
adriaanm: It does, via `mutateNonTypeConstructs` and then `registerTypeEquality`

adriaanm (changes requested):
Exciting! I have a few questions/suggestions. Thanks for taking this on.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
TomasMikula: I took this from [isPolySubType](https://github.com/scala/scala/blob/3b9feeafcd5a7e63fb79ca47fda24719d532ab7e/src/reflect/scala/reflect/internal/tpe/TypeComparers.scala#L344-L347), assuming the reason for opposite direction was that type parameters in `[X, Y] => T` appear in contravariant position. Looking at the [implementation of `methodHigherOrderTypeParamsSubVariance`](https://github.com/scala/scala/blob/3b9feeafcd5a7e63fb79ca47fda24719d532ab7e/src/reflect/scala/reflect/internal/tpe/TypeComparers.scala#L179-L180), from 2.11 onwards it always returns `true`, so could now be removed altogether?

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
TomasMikula: Yeah. Nevertheless, `mutateNonTypeConstructs` already seems to handle the higher-kinded `TypeVar` case correctly.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
TomasMikula: Done.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
TomasMikula: Will investigate.

TomasMikula: I tried to add `assert(false)` to this branch and it is not triggered by tests or bootstrap. What do you suggest? Should I just change that branch to `assert(false)`?

adriaanm:
Needs a mima exception:
```
ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.runtime.Settings.isScala213")
```

TomasMikula:
Thanks! Fixed.

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
adriaanm: I wouldn't risk a crash, but we could consider a warning to test our theory since this codepath isn't enabled by default. I haven't thought about it much, but there's precedent (grep "please report" for another warning).

TomasMikula: Done.

TomasMikula:
Just realized that this PR also fixes some problems with pattern matching:

 - scala/bug#10185
 - scala/bug#10195

Added positive tests for those to this PR.

TomasMikula:
As per comments in #6080, seems like the two failing tests added in the latter commit can be ignored, since those cases should be unreachable through compilation.

adriaanm:
Our policy is not to merge PRs that contain failing commits (so, let's  `testHigherKindedTypeVarUnification`). Also, on second thought (sorry), I'd rather not ignore MiMa's advice (I've come to regret that before). Let's inline `isScala213`. 

TomasMikula:
Ideas how I can inline `isScala213`? In `scala.tools.nsc.settings.ScalaSettings` it is defined as `source.value >= version213`, but `source` is not a member of `scala.reflect.internal.settings.MutableSettings`.

TomasMikula:
Would making it `private[scala]` be any better? It would still require MiMa exception, but at least user's of the API couldn't use it.

lrytz:
Given that `scala.reflect.internal` is [excluded from binary compatibility](https://github.com/scala/scala/blob/2.12.x/src/reflect/mima-filters/2.12.0.backwards.excludes#L1), shouldn't sub-packages be excluded as well?

TomasMikula:
There's `scala.reflect.runtime.Settings` which extends and therefore inherits the method from `scala.reflect.internal.settings.MutableSettings`.

lrytz:
Ah yes, I overlooked that, thanks. You could make `def isScala213` concrete in `internal.settings. MutableSettings` and do a follow-up PR for 2.13 that cleans it up. We did something similar here: https://github.com/scala/scala/pull/6051/files#diff-acaf4d66546a9cc80bec831144320695. But `private[scala]` is simpler, and I can't see a problem with going that way.

TomasMikula:
Made it

```scala
private[scala] def isScala213
```

adriaanm (approved):
Nice! Thanks for pushing this all the way through!

adriaanm:
PS: for the "fixes" command in the PR description to be picked up (and bugs to be auto-closed on merge), the syntax for the bug reference is `scala/bug#NNNNN` (I updated the description)

TomasMikula:
Thanks! I thought it was sufficient that the "fixes" were spelled out this way in the commit message.

joroKr21:
Regression: scala/bug#11303

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
joroKr21: We can get here thanks to https://github.com/scala/scala/blob/3a76f41d531a2aefba5c605f5b1fb738c33c5fa6/src/compiler/scala/tools/nsc/typechecker/Infer.scala#L538-L550

src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
```diff
@@ -365,7 +365,32 @@ trait TypeComparers {
...
+    def isSubHKTypeVar(tp1: Type, tp2: Type) = (tp1, tp2) match {
+      case (tv1 @ TypeVar(_, _), tv2 @ TypeVar(_, _)) =>
+        reporter.warning(tv1.typeSymbol.pos,
+          sm"""|compiler bug: Unexpected code path: testing two type variables for subtype relation:
+               |  ${tv1} <:< ${tv2}
+               |Please report bug at https://github.com/scala/bug/issues
+            """.trim)
+        false
+      case (tp1, tv2 @ TypeVar(_, _)) =>
+        val ntp1 = tp1.normalize
```
joroKr21, line 378: I don't understand, what was the reason for calling `normalize` here on the to-be-bound?

PR #6070, opened 2017-09-12
Optimize use of hash maps in makeLocal in the back end

retronym:
- Use an AnyRefMap rather than a general hash map
 - Avoid a map lookup in an assertion


retronym:
[Benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=02d1723eb2%7C554dfb2229&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40&from=1489300110561&to=1505197710561) scheduled.

lrytz (approved):

PR #6074, opened 2017-09-12
Resolve implicit instances for abstract types, according to the spec.

TomasMikula:
Having an abstract type `Foo[_]` in

```scala
trait Module {
  type Foo[_]
}
```
and

```scala
val m: Module = ???
```

the following rules from [SLS 2.12, section 7.2](https://scala-lang.org/files/archive/spec/2.12/07-implicits.html#implicit-parameters) all apply when collecting all companion modules associated with type `m.Foo[Bar]`:

> - if _T_ is an abstract type, the parts of its upper bound;
> - if _T_ is a parameterized type _S[T1,…,Tn]_, the union of the parts of _S_ and _T1,…,Tn_;
> - if _T_ is a type projection _S#U_, the parts of _S_ as well as _T_ itself;

(The last one applies because `m.Foo[Bar]` expands to `m.type#Foo[Bar]`.)

Thanks to @Atry for citing the relevant parts of the spec in scala/bug#10283.

Currently, the compiler only considers the upper bound of the abstract type (i.e. the first of the rules above).

This PR includes the other two rules.

Since this can cause ambiguous implicits in existing code, it is put under compiler flag `-Xsource:2.13`.

Fixes scala/bug#5818, scala/bug#10283.


jvican:
This is cool. If I'm not mistaken, this is a bug that was pointed out by Martin O. in a private discussion and that Dotty handles correctly; and fixing it will help me in my current implementation of opaque alias types since I'll not need to modify the implicit search to look for members on the type companion of the type alias (I still need to double check this, I'm not sure I understand all the implications of this change regarding implicit search).

TomasMikula:
> to look for members on the type companion of the type alias

I like this idea. I considered it for a while, but I figured there was a better chance of being merged if I stick to the spec.

jvican:
Yeah, I agree.
  
Opaque type aliases are for now a SIP. There have been lots of changes, so I'll be updating the public on the current design decisions soon. But the basic idea is that type aliases won't be transparent outside the type companion (i.e. `betaReduce` and `dealias` will return the existing type symbol, that is, do nothing, if `!isPastTyper && typeSymbol.isAliasType && typeSymbol.isOpaque`). Of course, this introduces the notion of type companion for opaque type aliases. [This example](https://github.com/scalacenter/scala/blob/b11dcd8eb39501a178efe06329d64af9f6ebde71/test/files/run/opaque-type/Logarithm.scala) gives a very rough idea of how the feature should work.

retronym:
> Since this can cause ambiguous implicits in existing code, it is put under compiler flag '-Yabstype-implicits'.

I'd prefer `-Xsource:2.13` to control this. You could structure this PR to make the change unconditionally in the first commit, and then guard it with `-Xsource:2.13` in a following commit. We can then test the first commit in the Scala community build to get a feel for how often the bug fix will cause trouble for existing code.

retronym:
We should consider this change in concert with #5867 (/cc @hrhino)

TomasMikula:
Incorporated @retronym's suggestions.

hrhino:
cc successful. I'm trying to think of reasons they'd clash and not coming up with any (other than both of us scouting that spare `-`). 

This one looks pretty great, too. Nice job!

adriaanm (approved):
Let's run the benchmarks on the merge commit. If there's an issue, we'll back out the change on Monday.

Jasper-M:
```
trait Module[Companion] { type Foo <: Bar[Companion]; type Bar[C] }
object MyCompanion { implicit def foo(f: m.Foo): String = "ok" }
val m: Module[MyCompanion.type] = null
```

Should `MyCompanion` be part of the implicit scope of `m.Foo` according to the spec? I would think so, but you say that "the parts of its upper bound" are already part of the implicit scope, and that snippet still isn't working in 2.12.3.

update:
I just verified that it does work in Dotty. So is this also fixed by this pull request? I am suspecting it isn't since you said that part of the spec was already implemented. (Sorry, can't check out your branch and see for myself at the moment)

TomasMikula:
I can check, but can you complete your example so that it contains the code that triggers the implicit search?

Jasper-M:
Of course. This is a valid dotty snippet: https://scastie.scala-lang.org/Jasper-M/xsaFmpj7TBOhvSG0sn3uYQ
The program should print `ok`.
```scala
object Main {
  trait Module[Companion] { type Foo >: Null <: Bar[Companion]; type Bar[C] }
  object MyCompanion { implicit def foo(f: m.Foo): String = "ok" }
  val m: Module[MyCompanion.type] = null  

    val foo: m.Foo = null
    println(foo: String)
  }
}
```

I just added the `Null` to avoid any casting, but that shouldn't make a difference.

adriaanm:
~~Only classes have companions as the specs stands. Abstract types don't. (This may change with the opaque types SIP.)~~ looks like I replied too quickly.

TomasMikula:
@Jasper-M It compiles with minor changes to your snippet: removed the `Null` lower bound and replaced the second `null` with `???`:

```scala
object Main {
  trait Module[Companion] { type Foo <: Bar[Companion]; type Bar[C] }
  object MyCompanion { implicit def foo(f: m.Foo): String = "ok" }
  val m: Module[MyCompanion.type] = null  

    val foo: m.Foo = ???
    println(foo: String)
  }
}
```

Jasper-M:
@TomasMikula Right sorry, Scalac and Dotty handle bounds differently. Great that it works though! You're on fire.

TomasMikula:
> Should `MyCompanion` be part of the implicit scope of `m.Foo` according to the spec? I would think so, but you say that "the parts of its upper bound" are already part of the implicit scope, and that snippet still isn't working in 2.12.3.

The upper bound of `Foo` in `m.Foo` doesn't seem relevant for finding the conversion. It is found through `m`'s type, `Module[MyCompanion.type]`. That is what was added in this PR.

PR #6075, opened 2017-09-13
Performance improvements for DCE

lrytz:
DCE is enabled by default. This PR eliminates some performance issues. Continuation of #6044.

### DIFF
--- a/src/compiler/scala/tools/asm/LabelAccess.java
+++ b/src/compiler/scala/tools/asm/LabelAccess.java
@@ -1,0 +1,18 @@
+package scala.tools.asm;
+
+/**
+ * Temporary class to allow access to the package-private status field of class Label.
+ */
+public class LabelAccess {
+    public static boolean isLabelFlagSet(Label l, int f) {
+        return (l.status & f) != 0;
+    }
+
+    public static void setLabelFlag(Label l, int f) {
+        l.status |= f;
+    }
+
+    public static void clearLabelFlag(Label l, int f) {
+        l.status &= ~f;
+    }
+}
--- a/src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala
@@ -113,14 +113,13 @@ import scala.tools.nsc.backend.jvm.opt.BytecodeUtils._
  *
  *
  * empty local variable descriptors (removes unused variables from the local variable table)
- *   + enables downstream:
- *     - stale labels (labels that the entry points to, if not otherwise referenced)
  *
  * empty line numbers (eliminates line number nodes that describe no executable instructions)
- *   + enables downstream:
- *     - stale labels (label of the line number node, if not otherwise referenced)
  *
- * stale labels (eliminate labels that are not referenced, merge sequences of label definitions)
+ * At this point, we used to filter out redundant label nodes (sequences of labels without any
+ * executable instructions in between). However, this operation is relatively expensive, and
+ * unnecessary: labels don't exist in the classfile, they are lowered to bytecode offsets, so
+ * redundant labels disappear by design.
  *
  *
  * Note on a method's maxLocals / maxStack: the backend only uses those values for running
@@ -385,11 +384,11 @@ abstract class LocalOpt {
       else false
     traceIfChanged("localVariables")
 
+    // The asm.MethodWriter writes redundant line numbers 1:1 to the classfile, so we filter them out
+    // Note that this traversal also cleans up `LABEL_REACHABLE_STATUS` flags that were added to Label's
+    // `stats` fields during `removeUnreachableCodeImpl` (both are guarded by `optUnreachableCode`).
     val lineNumbersRemoved = if (compilerSettings.optUnreachableCode) removeEmptyLineNumbers(method) else false
     traceIfChanged("lineNumbers")
-
-    val labelsRemoved = if (compilerSettings.optUnreachableCode) removeEmptyLabelNodes(method) else false
-    traceIfChanged("labels")
 
     // assert that local variable annotations are empty (we don't emit them) - otherwise we'd have
     // to eliminate those covering an empty range, similar to removeUnusedLocalVariableNodes.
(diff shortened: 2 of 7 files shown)


lrytz:
Profile at the end:
image

Compare with the one before: https://github.com/scala/scala/pull/6044#issuecomment-325882798

I'll start the benchmarks once the build looks green.

lrytz:
Benchmark results to [appear here](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=02d1723eb2%7C46aebd4440%7C741d5892a3%7C4813f3bd8c%7C6e3c5fcada%7C11363a54b0%7C2a703a4922&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1505082758398&to=1505312469085)

lrytz:
Benchmark results are maybe a bit less clear / good than I hoped, but overall it looks positive I'd say. I'll try to write a microbenchmark tomorrow for the two versions.

lrytz:
Uh, I also ran the `scala` benchmark, it seems to help more with larger codebases.

retronym:
DCE is still taking noticeable time. I think it is still worthwhile to carve out a fast path here. In #6044, I tried to use the absence of an `ATHROW` as reason enough to skip DCE. That wasn't enough, as certain while loops had the same problem.

We could also detect back-edges during code gen. In `genApply`, we'd can to see if the call to `programPoint` for the label application found an existing entry in the `jumpDest` map.

```scala
if (sym.isLabel) { // jump to a label
  genLoadLabelArguments(args, labelDef(sym), app.pos)
  bc goTo programPoint(sym)
}
```

So the question is: are there any methods without back-edges or throws for which our DCE still would find something to do?

retronym:
Here's an implementation of my comment above: https://github.com/scala/scala/compare/2.12.x...retronym:review/fasterDCE?expand=1

[Benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=02d1723eb2%7C46aebd4440%7C741d5892a3%7C4813f3bd8c%7C6e3c5fcada%7C11363a54b0%7C2a703a4922%7Ccd71109b&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40&from=1505082758398&to=1505361023905).

src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala
```diff
@@ -612,16 +613,21 @@ object LocalOptImpls {
...
-        removedHandlers += handler
+        handlerRemoved = true
+        if (!liveHandlerRemoved && BackendUtils.isLabelReachable(handler.start))
+          liveHandlerRemoved = true
         handlersIter.remove()
       }
     }
-    removedHandlers
+
+    (handlerRemoved, liveHandlerRemoved)
```
retronym: The benchmarks suggest this commit made performance go slightly backwards.

Assuming the common case is that nothing is removed, it could help to reuse an instance of `(false, false)` here (and above on line 323), which would be allocation free.

retronym: The code will probably look a bit cleaner if you introduce a case class to represent that result.

```
case class RemoveEmptyExceptionHandlersResult(handlerRemoved: Boolean, liveHandlerRemoved: Boolean)
object RemoveEmptyExceptionHandlersResult {
   val Empty = RemoveEmptyExceptionHandlersResult(false, false)
   def get(handlerRemoved: Boolean, liveHandlerRemoved: Boolean) = if (handlerRemoved || liveHandlerRemoved) RemoveEmptyExceptionHandlersResult(handlerRemoved, liveHandlerRemoved) else Empty
}
```

retronym:
There are some opportunities to improve the performance of the ASM Analyzer framework (in a followup change).

In https://github.com/scala/scala-asm/blob/51ec7ee2a82fbf7ffc6bf8c9af7e555eace9140e/src/main/java/scala/tools/asm/tree/analysis/Frame.java#L623-L632, the calls to:

`Type.getArgumentTypes(desc).length` and `Type.getReturnType(desc) == Type.VOID_TYPE` could be more directly implemented. We could also use this in `invokeConsProd`.

In [https://github.com/scala/scala-asm/blob/51ec7ee2a82fbf7ffc6bf8c9af7e555eace9140e/src/main/java/scala/tools/asm/Type.java](`Type.java`), a number of `String` to `char[]` conversions are performed to call into `getType`. If this method was duplicated to operator on `String` directly, those conversions could be avoided.

```
public static Type getReturnType(final String methodDescriptor) {
        char[] buf = methodDescriptor.toCharArray();
        return getType(buf, methodDescriptor.indexOf(')') + 1);
    }
```

Could use `lastIndexOf` to search from the end of the descriptor.

src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala
lrytz: I can also encode the two booleans in an int

lrytz:
If we manage to identify methods that need DCE during code gen that's great! We should
  - jardiff
  - Gather statistics: how many DCEs do we save? It will probably mostly affect very small methods.
  - Maybe run `removeEmptyLineNumbers` on all methods by default? I think redundant line number entries are pretty common in our code gen.

Should we keep this PR focus on performance improvements for DCE?

I'll create tickets on scala-asm for your suggested improvements to the Analyzer.

src/compiler/scala/tools/nsc/backend/jvm/opt/LocalOpt.scala
lrytz: done

src/compiler/scala/tools/asm/LabelAccess.java
```diff
@@ -0,0 +1,18 @@
+package scala.tools.asm;
+
+/**
+ * Temporary class to allow access to the package-private status field of class Label.
+ */
+public class LabelAccess {
```
lrytz, line 6: Should I do a new asm release with the `status` made public before we merge this PR?

retronym: I don't really mind if we merge this first.

retronym (approved):

PR #6076, opened 2017-09-13
@implicitNotFound messages can access type args from prefixes

hrhino:
Motivation:

```scala
class HMap[Rel[_, _]] {
  def get[K, V](k: K)(implicit rel: Rel[K, V]): V
  def add[K, V](k: K, v: V)(implicit rel: Rel[K, V]): HMap[Rel]
}

trait ListOf[Upper] {
  @implicitNotFound("Cannot add (${K}, ${V}) to a HMap (is ${V} =:= List[${K}] and <: ${Upper}?)")
  trait Lambda[K, V]
}
object ListOf {
  implicit def rel[Upper, T <: Upper]: ListOf[Upper]#Lambda[T,List[T]] = ...
}
```

and various other type-level weirdnesses.

The implementation is pretty simple: traverse the type, grabbing `(sym.tparams zip args)` for each `TypeRef` we can get our hands on. For `validate`, of course, we have the symbol of the type definition to go off of, and just grab the `tparams` from all of the types in the `ownerChain`.

I also buffed up the validation message since I was in the area anyways.

PR note: there are things like `TypeCollector` which seem almost what I want, but I couldn't get them to work exactly as expected. If that'd be a better solution I'm all for using it, but I could use a pointer on which one is what I want.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -1498,8 +1498,8 @@ trait Implicits {
   }
 
   class ImplicitAnnotationMsg(f: Symbol => Option[String], clazz: Symbol, annotationName: String) {
-    def unapply(sym: Symbol): Option[(Message)] = f(sym) match {
-      case Some(m) => Some(new Message(sym, m, annotationName))
+    def unapply(sym: Symbol): Option[Message] = f(sym) match {
+      case Some(msgStr) => Some(new Message(sym, msgStr, annotationName))
       case None if sym.isAliasType =>
         // perform exactly one step of dealiasing
         // this is necessary because ClassManifests are now aliased to ClassTags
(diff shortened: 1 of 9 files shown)


src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1511,11 +1511,11 @@ trait Implicits {
     }
 
     // check the message's syntax: should be a string literal that may contain occurrences of the string "${X}",
-    // where `X` refers to a type parameter of `sym`
+    // where `X` refers to a type parameter of `sym` or any of its prefixen
```
som-snytt: `Prefixen` is the German spelling.

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1423,7 +1423,7 @@ abstract class RefChecks extends Transform {
           applyChecks(sym.annotations)
 
           def messageWarning(name: String)(warn: String) =
-            reporter.warning(tree.pos, f"Invalid $name message for ${sym}%s${sym.locationString}%s:%n$warn")
+            reporter.warning(tree.pos, f"Invalid @$name message for $sym%s${sym.locationString}%s:%n$warn")
```
som-snytt, line 1426: might as well rm extra `%s` or two.

hrhino:
It occurs to me that this should probably go into 2.13 instead; anyone who relies on this to write an `@inf` error message in 2.12.x will wind up with incorrect messages for anyone using an earlier version of 2.12.x.

adriaanm:
Maybe we should turn this map from all possible names to their types inside out, and parse the contents of the spliced code and type check it? This would be more robust for name clashes etc.

adriaanm:
I'm sorry to postpone this to the next milestone. I will do my best to be more on top of PR reviews once we get situated in our new place.

SethTisue:
@hrhino is this one you mean to revive at some point...?

som-snytt:
This is the other one he meant to revive. But maybe Adriaan isn't situated yet?

adriaanm:
That excuse is no longer applicable -- now I have to blame our 1yo.

PR #6078, opened 2017-09-14
WIP By default, only run DCE on methods that throw or backwards branch

retronym:
For discussion in #6075


adriaanm:
Moving this to WIP since it's for discussion

PR #6079, opened 2017-09-14
[WIP] Skip over contexts with the same our outer nested scopes during name lookup.

milessabin:
This avoids researching the scope of an enclosing context if the scope is identical or if the inner scope was nested in the outer (which would mean that we have searched the outer scopes definitions already).

This does result in fewer searches, but if anything appears to very marginally increase compile times ... I'm mainly submitting this PR to see what peoples opinions are on why this doesn't result in any sort of speedup.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -1081,7 +1081,7 @@ trait Contexts { self: Analyzer =>
             else newOverloaded(cx.owner, pre, entries)
         }
         if (!defSym.exists)
-          cx = cx.outer // push further outward
+          cx = outerContextWithDistinctScope(cx) // push further outward
       }
       if (symbolDepth < 0)
         symbolDepth = cx.depth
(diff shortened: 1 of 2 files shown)


milessabin:
I can't see what's causing the test failures here ... partest and junit all run to completion successfully locally.

som-snytt:
I was just rubbernecking (in case there was some interesting interaction with context and the named arg stuff). It also works for me but I noticed that I'm not running the same thing as jenkins, because jenkins doesn't show the extra warning for the deprecated opt flag.

```
$ sbt setupPublishCore testAll

[info] Compiling 583 Scala sources and 120 Java sources to /home/amarki/projects/scala/build/quick/classes/library...
[warn] -opt:l:classpath is deprecated, use -opt:l:inline and -opt-inline-from
[warn] there were 37 deprecation warnings (since 2.10.0)
[warn] there were 24 deprecation warnings (since 2.11.0)
[warn] there were 46 deprecation warnings (since 2.12.0)
[warn] there were 107 deprecation warnings in total; re-run with -deprecation for details
[warn] there were 368 inliner warnings; re-run enabling -opt-warnings for details, or try -help
[warn] 6 warnings found

```

lrytz:
The difference may be in the bootstrap, i.e., the resulting compiler has a bug and cannot compile the scala repo anymore. Run `sbt setupPublishCore publishLocal` then `sbt -Dstarr.version=2.12.....`

milessabin:
@lrytz you're right ... it does fail to bootstrap.

adriaanm:
Tentatively assigning to 2.12.5

milessabin:
@adriaanm this is really more of an experiment to see if the the redundant searches actually are a performance issue. The benchmarks seem to suggest not.

adriaanm:
Ok, thanks -- assigning to WIP then.

SethTisue:

PR #6082, opened 2017-09-15
Cleanup in aisle patmat, someone spilled a few `<unapply-selector>`s

adriaanm:
mostly ready for review, TODO: commit messages :-)

Fixes scala/bug#6130


src/compiler/scala/tools/nsc/transform/patmat/MatchTranslation.scala
```diff
@@ -487,51 +476,57 @@ trait MatchTranslation {
...
-    class ExtractorCallRegular(aligner: PatternAligned, extractorCallIncludingDummy: Tree, val args: List[Tree]) extends ExtractorCall(aligner) {
-      val Unapplied(fun) = extractorCallIncludingDummy
+    /**
+      *
+      * @param fun reference to the unapply method
+      * @param args the subpatterns
+      * @param unapplyAppliedToDummy an application of the unapply method to the (dummy) unapply selector
+      */
+    class ExtractorCallRegular(fun: Tree, args: List[Tree])(unapplyAppliedToDummy: Tree) extends ExtractorCall(fun, args) {
+      override def unapplySelector =
```
adriaanm: TODO:  make lazy val? (can't be val)

adriaanm:
I intend to squash the initial few commits after review.

retronym:
All LGTM so far.

test/files/neg/t4425b.check
```diff
@@ -24,15 +24,9 @@ Note: def unapply(x: String)(y: String): Nothing exists in object X, but it cann
...
     println(      "" match { case _ X _   => "ok" ; case _ => "fail" })
-                                  ^
-t4425b.scala:18: error: too many patterns for object X offering Boolean: expected 0, found 2
-    println(      "" match { case _ X _   => "ok" ; case _ => "fail" })
                                     ^
 t4425b.scala:19: error: The result type of an unapply method must contain a member `get` to be used as an extractor pattern, no such member exists in Nothing
     println((X: Any) match { case _ X _   => "ok" ; case _ => "fail" })
-                                  ^
-t4425b.scala:19: error: too many patterns for object X offering Boolean: expected 0, found 2
-    println((X: Any) match { case _ X _   => "ok" ; case _ => "fail" })
```
retronym, line 35: Do you know why these errors went away?

```
object Test2 {
  object X { def unapply(x : String) = throw new Exception }

  def f1() {
    println(      "" match { case _ X _   => "ok" ; case _ => "fail" }) // line 18
    println((X: Any) match { case _ X _   => "ok" ; case _ => "fail" }) // line 19
    println(      "" match { case X(_)    => "ok" ; case _ => "fail" })
    println((X: Any) match { case X(_)    => "ok" ; case _ => "fail" })
    println(      "" match { case X(_, _) => "ok" ; case _ => "fail" }) // Just reporting one error is now consistent with these, so looks like a progression
    println((X: Any) match { case X(_, _) => "ok" ; case _ => "fail" }) // ...
  }
}
```

adriaanm: Yes, I consolidated the old `noGetError` into the error checks at https://github.com/scala/scala/pull/6082/files#diff-46bedd6c71916de4f4963e301828852eR246. Before, this error was emitted independently of the other ones.

adriaanm:
I'll investigate today how this PR regressed our benchmarks. We do intrinsically need to do more substitutions.

PR #6083, opened 2017-09-15
Remove sorting of implicit candidates by use frequency and divergent-prone Ordering implicit

retronym:
This performance optimization leads to compilation order dependent variations
in whether an implicit search will diverge or will find a successful result.

Let's try turning off this sorting and measuring the impact on compilation
speed.

It is worth noting that I have observed cases in the past where this sorting
was actually counterproductive, as it undermines the attempt of an macro
author to control the order of implicit searches with the subclass rule of
static overload resolution. The subclass rule is still applied correctly,
but the superclass implicit can be tried _before_ the subclass implicit 
which can explore more expensive parts of the search space that end
up being out-prioritized.

Refererences: scala/bug#10222, scala/bug#8541, scala/bug#10080

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -957,8 +957,12 @@ trait Implicits {
           result
         }
 
-        // most frequent one first
-        matches sortBy (x => if (isView) -x.useCountView else -x.useCountArg)
+        if (settings.isScala213) matches
+        else {
+          // most frequent one first under Scala 2.12 mode. We've turned this optimization off to avoid
+          // compilation order variation in whether a search succeeds or diverges.
+          matches sortBy (x => if (isView) -x.useCountView else -x.useCountArg)
+        }
       }
       if (eligible.nonEmpty)
         printTyping(tree, eligible.size + s" eligible for pt=$pt at ${fullSiteString(context)}")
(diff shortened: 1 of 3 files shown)


retronym:
[Benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=12b8cfb69f1f41c9f5bc1653e9549421f497e682%7C12b8cfb69f1f41c9f5bc1653e9549421f497e682&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1504015529102&to=1505437792751) scheduled.

odersky:
But even without sorting you still have compilation order-dependency, right?

retronym:
What do you have in mind? Without sorting, I’d expect the in scope and
companion scope implicits to have a deterministic ordering for a given
implicit search.

There an order-of-typechecking dependency around the way that scalac
excludes inferred implicits that succeed the position of the search, but
that’s orthogonal to this sorting.

SethTisue:
@retronym is this ready for review? I assumed not and added WIP label.

retronym:
I've rebased, and removed [the last commit](https://github.com/retronym/scala/tree/topic/divergey-ordering) that tried to make the divergent-prone implicit in `Ordering` non-implicit. The first change helps our sanity and doesn't seem to hurt performance (I belieive it can help performance in some cases by honouring the implicit prioritization in the search order).

Ready to review.

adriaanm (approved):

PR #6087, opened 2017-09-18
Add bridges for overridden methods in lambda indy call

hrhino:
If a SAM trait's abstract method overrides a method in a supertrait while changing the return type, the generated invokedynamic instruction needs to pass the types of the overridden methods to `LambdaMetaFactory` so that bridge methods can be added to the generated lambda.

Java does this differently: it generates a default method in the subinterface overriding the superinterface method. Theoretically we could also generate the default bridges, but that is a binary-incompatible change, so I think it's safer to just add the bridges in the invokedynamic.

Honestly I'm surprised that this hasn't come up already.

Fixes scala/bug#10512.

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -626,7 +626,7 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
         case Apply(fun, args) if app.hasAttachment[delambdafy.LambdaMetaFactoryCapable] =>
           val attachment = app.attachments.get[delambdafy.LambdaMetaFactoryCapable].get
           genLoadArguments(args, paramTKs(app))
-          genInvokeDynamicLambda(attachment.target, attachment.arity, attachment.functionalInterface, attachment.sam, attachment.isSerializable, attachment.addScalaSerializableMarker)
+          genInvokeDynamicLambda(attachment)
           generatedType = methodBTypeFromSymbol(fun.symbol).returnType
 
         case Apply(fun, List(expr)) if currentRun.runDefinitions.isBox(fun.symbol) =>
@@ -1332,7 +1332,9 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
     def genSynchronized(tree: Apply, expectedType: BType): BType
     def genLoadTry(tree: Try): BType
 
-    def genInvokeDynamicLambda(lambdaTarget: Symbol, arity: Int, functionalInterface: Symbol, sam: Symbol, isSerializable: Boolean, addScalaSerializableMarker: Boolean) {
+    def genInvokeDynamicLambda(canLMF: delambdafy.LambdaMetaFactoryCapable) = {
+      import canLMF._
+
       val isStaticMethod = lambdaTarget.hasFlag(Flags.STATIC)
       def asmType(sym: Symbol) = classBTypeFromSymbol(sym).toASMType
 
@@ -1349,23 +1351,36 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
       val constrainedType = MethodBType(lambdaParams.map(p => typeToBType(p.tpe)), typeToBType(lambdaTarget.tpe.resultType)).toASMType
       val samMethodType = methodBTypeFromSymbol(sam).toASMType
       val markers = if (addScalaSerializableMarker) classBTypeFromSymbol(definitions.SerializableClass).toASMType :: Nil else Nil
-      visitInvokeDynamicInsnLMF(bc.jmethod, sam.name.toString, invokedType, samMethodType, implMethodHandle, constrainedType, isSerializable, markers)
+      val overriddenMethods = bridges.map(b => methodBTypeFromSymbol(b).toASMType)
+      visitInvokeDynamicInsnLMF(bc.jmethod, sam.name.toString, invokedType, samMethodType, implMethodHandle, constrainedType, overriddenMethods, isSerializable, markers)
       if (isSerializable)
         addIndyLambdaImplMethod(cnode.name, implMethodHandle)
     }
   }
 
   private def visitInvokeDynamicInsnLMF(jmethod: MethodNode, samName: String, invokedType: String, samMethodType: asm.Type,
-                                        implMethodHandle: asm.Handle, instantiatedMethodType: asm.Type,
-                                        serializable: Boolean, markerInterfaces: Seq[asm.Type]) = {
+                                        implMethodHandle: asm.Handle, instantiatedMethodType: asm.Type, overriddenMethodTypes: Seq[asm.Type],
+                                        serializable: Boolean, markerInterfaces: Seq[asm.Type]): Unit = {
     import java.lang.invoke.LambdaMetafactory.{FLAG_BRIDGES, FLAG_MARKERS, FLAG_SERIALIZABLE}
     // scala/bug#10334: make sure that a lambda object for `T => U` has a method `apply(T)U`, not only the `(Object)Object`
     // version. Using the lambda a structural type `{def apply(t: T): U}` causes a reflective lookup for this method.
-    val needsBridge = samMethodType != instantiatedMethodType
-    val bridges = if (needsBridge) Seq(Int.box(1), instantiatedMethodType) else Nil
+    val needsGenericBridge = samMethodType != instantiatedMethodType
+    // scala/bug#10512: any methods which `samMethod` overrides need bridges made for them
+    // this is done automatically during erasure for classes we generate, but LMF needs to have them explicitly mentioned
+    // so we have to compute them at this relatively late point.
+    val bridges = (
+      if (needsGenericBridge)
+        instantiatedMethodType +: overriddenMethodTypes
+      else overriddenMethodTypes
+    ).distinct.filterNot(_ == samMethodType)
+
+    /* We're saving on precious BSM arg slots by not passing 0 as the bridge count */
+    val bridgeArgs = if (bridges.nonEmpty) Int.box(bridges.length) +: bridges else Nil
+
     def flagIf(b: Boolean, flag: Int): Int = if (b) flag else 0
-    val flags = FLAG_MARKERS | flagIf(serializable, FLAG_SERIALIZABLE) | flagIf(needsBridge, FLAG_BRIDGES)
-    val bsmArgs = Seq(samMethodType, implMethodHandle, instantiatedMethodType, Int.box(flags), Int.box(markerInterfaces.length)) ++ markerInterfaces ++ bridges
+    val flags = FLAG_MARKERS | flagIf(serializable, FLAG_SERIALIZABLE) | flagIf(bridges.nonEmpty, FLAG_BRIDGES)
+    val bsmArgs = Seq(samMethodType, implMethodHandle, instantiatedMethodType, Int.box(flags), Int.box(markerInterfaces.length)) ++ markerInterfaces ++ bridgeArgs
+
     jmethod.visitInvokeDynamicInsn(samName, invokedType, lambdaMetaFactoryAltMetafactoryHandle, bsmArgs: _*)
   }
 
--- a/src/compiler/scala/tools/nsc/transform/Delambdafy.scala
+++ b/src/compiler/scala/tools/nsc/transform/Delambdafy.scala
@@ -28,7 +28,7 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
   /** the following two members override abstract members in Transform */
   val phaseName: String = "delambdafy"
 
-  final case class LambdaMetaFactoryCapable(target: Symbol, arity: Int, functionalInterface: Symbol, sam: Symbol, isSerializable: Boolean, addScalaSerializableMarker: Boolean)
+  final case class LambdaMetaFactoryCapable(lambdaTarget: Symbol, arity: Int, functionalInterface: Symbol, sam: Symbol, bridges: List[Symbol], isSerializable: Boolean, addScalaSerializableMarker: Boolean)
 
   /**
     * Get the symbol of the target lifted lambda body method from a function. I.e. if
@@ -59,7 +59,10 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
     private[this] lazy val methodReferencesThis: Set[Symbol] =
       (new ThisReferringMethodsTraverser).methodReferencesThisIn(unit.body)
 
-    private def mkLambdaMetaFactoryCall(fun: Function, target: Symbol, functionalInterface: Symbol, samUserDefined: Symbol, isSpecialized: Boolean): Tree = {
+    private def mkLambdaMetaFactoryCall(fun: Function, target: Symbol, functionalInterface: Symbol, samUserDefined: Symbol, userSamCls: Symbol, isSpecialized: Boolean): Tree = {
+      /* user-defined SAM types should have gotten a class symbol made for them in `typer` */
+      assert(isFunctionType(fun.tpe) || (samUserDefined.exists && userSamCls.isClass), s"$fun / ${fun.symbol} / ${fun.tpe}")
+
       val pos = fun.pos
       def isSelfParam(p: Symbol) = p.isSynthetic && p.name == nme.SELF
       val hasSelfParam = isSelfParam(target.firstParam)
@@ -98,13 +101,19 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
       val isSerializable = samUserDefined == NoSymbol || samUserDefined.owner.isNonBottomSubClass(definitions.JavaSerializableClass)
       val addScalaSerializableMarker = samUserDefined == NoSymbol
 
+      val samBridges = logResultIf[List[Symbol]](s"will add SAM bridges for $fun", _.nonEmpty) {
+        userSamCls.fold[List[Symbol]](Nil) {
+          _.info.findMembers(excludedFlags = 0L, requiredFlags = BRIDGE).toList
+        }
+      }
+
       // The backend needs to know the target of the lambda and the functional interface in order
       // to emit the invokedynamic instruction. We pass this information as tree attachment.
       //
       // see https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/LambdaMetafactory.html
       //   instantiatedMethodType is derived from lambdaTarget's signature
       //   samMethodType is derived from samOf(functionalInterface)'s signature
-      apply.updateAttachment(LambdaMetaFactoryCapable(lambdaTarget, fun.vparams.length, functionalInterface, sam, isSerializable, addScalaSerializableMarker))
+      apply.updateAttachment(LambdaMetaFactoryCapable(lambdaTarget, fun.vparams.length, functionalInterface, sam, samBridges, isSerializable, addScalaSerializableMarker))
 
       apply
     }
@@ -254,8 +263,11 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
           (functionalInterface, isSpecialized)
         }
 
-      val sam = originalFunction.attachments.get[SAMFunction].map(_.sam).getOrElse(NoSymbol)
-      mkLambdaMetaFactoryCall(originalFunction, target, functionalInterface, sam, isSpecialized)
+      val (sam, synthCls) = originalFunction.attachments.get[SAMFunction] match {
+        case Some(SAMFunction(_, sam, synthCls)) => (sam,      synthCls)
+        case None                                => (NoSymbol, NoSymbol)
+      }
+      mkLambdaMetaFactoryCall(originalFunction, target, functionalInterface, sam, synthCls, isSpecialized)
     }
 
     // here's the main entry point of the transform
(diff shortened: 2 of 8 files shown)


src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2385,7 +2385,13 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
...
+    /** Equivalent to allOverriddenSymbols.nonEmpty, but more efficient.
+      *
+      * Drive-by warning: the result of `allOverriddenSymbols` can go from
+      * non-empty to empty following `erasure`; it seems that this is always
+      * initialized pre-erasure, so post-erasure this method is no longer
+      * equivalent to `allOverriddenSymbols.nonEmpty` like it claims.
+      */
     lazy val isOverridingSymbol = (
          canMatchInheritedSymbols
       && owner.ancestors.exists(base => overriddenSymbol(base) != NoSymbol)
```
retronym: Another drive-by warning: this cache can serve up outdated values in multi-Run compilation if the ancestors change! We should store either the `Run` or maybe the full `Period` alongside this boolean and recompute if needed, as is done for other caches in `*Symbol`

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1349,23 +1349,36 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
       val constrainedType = MethodBType(lambdaParams.map(p => typeToBType(p.tpe)), typeToBType(lambdaTarget.tpe.resultType)).toASMType
       val samMethodType = methodBTypeFromSymbol(sam).toASMType
       val markers = if (addScalaSerializableMarker) classBTypeFromSymbol(definitions.SerializableClass).toASMType :: Nil else Nil
-      visitInvokeDynamicInsnLMF(bc.jmethod, sam.name.toString, invokedType, samMethodType, implMethodHandle, constrainedType, isSerializable, markers)
+      val overriddenMethods = enteringErasure(sam.allOverriddenSymbols).map(o => methodBTypeFromSymbol(o).toASMType)
```
retronym: It would be ideal to unify the computation of the required bridges with the normal bridge logic in `Erasure`. We would need to change it to recognize that `Function` nodes are classes-in-waiting, and as such should have `ComputeBridges` applied. The listing of required bridges could be attached to the function symbol (probably inside the existing attachment, [`SAMFunction`](https://github.com/scala/scala/blob/f18b2494059599633f5afbefd57be8080ac21b0b/src/reflect/scala/reflect/internal/StdAttachments.scala#L57).)

retronym: The implementation of `ComputeBridges` as a `SymbolPair`-s walk has an efficiency argument over finding overriden members `decl`-wise. For SAM functions that won't make a difference, though, as we've only decl.

So the argument for unifying the logic would be more about keeping responsibilities clear, which might help curtail bugs down the track.

lrytz:
LGTM

sjrd:
> Theoretically we could also generate the default bridges, but that is a binary-incompatible change

Why would it be binary incompatible to add a default method to an interface that already inherits a signature for that method (although abstract)?

sjrd:
So I checked with Scala.js, and the test case also causes a linking error in Scala.js:
```
[error] Referring to non-existent method helloworld.JsonEncoderInstances$$$Lambda$1.encode(java.lang.Object)helloworld.JsonValue
[error]   called from helloworld.Test$.delayedEndpoint$helloworld$Test$1()scala.Unit
[error]   called from helloworld.Test$delayedInit$body.apply()java.lang.Object
[error]   called from scala.Function0.apply$mcZ$sp()scala.Boolean
[error]   called from scala.runtime.AbstractFunction0.apply$mcZ$sp()scala.Boolean
...
```
It seems that, with the fix in this PR, we'll have to duplicate this logic with `overriddenSymbols` in the Scala.js back-end. I would definitely be nice if the appropriate computation could be done earlier in a unified way.

hrhino:
@sjrd thanks for the heads up. I guessed without checking, since this showed up with the 2.12 LMF-based lambda encoding, that it was JVM-specific. Now I guess not. The signature of `computeBridges` was a little awkward for me (IIRC it takes a symbol and a tree and mutates them? I'm on my phone...) but that's a good argument for a different fix.

sjrd:
> I guessed without checking, since this showed up with the 2.12 LMF-based lambda encoding, that it was JVM-specific. Now I guess not.

For all LMF-based lambdas, Scala.js has to replicate whatever the JVM back-end + LMF itself do in the Scala.js back-end. It's mostly around [`synthesizeSAMWrapper`](https://github.com/scala-js/scala-js/blob/50b5f908e415535cd21e9239c7ffae9f7a2f8169/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L5187). It's possible that the solution is really back-end specific, but that we both have the same issue, which seems to be the case here. I'm not saying the back-end solution in this PR is not the correct one; but that if we can somehow compute the relevant information in a unified way, that would help keep the back-ends in sync and avoid duplicated efforts. It has also happened in the past the solution was lying entirely earlier in the pipeline (I can't find the PR/issue at the moment).

src/reflect/scala/reflect/internal/Symbols.scala
hrhino: I'll put that in a separate commit once I deal with the `ComputeBridges` suggestions.

hrhino:
@sjrd my bad w/r/t "binary compatibility", I think; it shouldn't matter with the 2.12 trait encoding.

However, there is another consideration here, which is that if the traits are defined in library A, and project B wants to use the lambda syntax, with this change the project B only needs to bump to a newer version of the compiler, whereas with the Java-style default method fix B would need to wait for A to recompile with the newer version and release. 

Perhaps this could be revisited for 2.13, which would simplify things for SJS (it wouldn't be a backend change anymore) but it seems awkward for 2.12. If that sounds better I can slap a [nomerge] on this and make that change against 2.13.x.

sjrd:
Ah, good point.

IMO your plan sounds good :+1:

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2843,65 +2843,94 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+            samCls.info.decls.enter {
+              sam.cloneSymbol(samCls, sam.flags & ~DEFERRED).setInfo(samTp memberInfo sam)
+            }
+
+            fun.setType(samTp)
+
+            /* Arguably I should do `fun.setSymbol(samCls)` rather than leaning
+             * on an attachment, but doing that confounds lambdalift's free var
+             * analysis in a way which does not seem to be trivially reparable. */
+            fun.updateAttachment(SAMFunction(samTp, sam, samCls))
```
hrhino: @retronym it was plan "B", I'm afraid...

src/reflect/scala/reflect/internal/Symbols.scala
hrhino: I wound up just caching `allOverriddenSymbols`...

It appears that `isOverridingSymbol` gets called (after `typer`) mainly in [`constructors`](https://github.com/scala/scala/blob/e100de3e8c82b7e1812c2ea47cd31fc17be5539a/src/compiler/scala/tools/nsc/transform/Constructors.scala#L168) and [`explicitouter`](https://github.com/scala/scala/blob/e100de3e8c82b7e1812c2ea47cd31fc17be5539a/src/compiler/scala/tools/nsc/transform/ExplicitOuter.scala#L456), and probably at most once per symbol per each of those phases, so I don't feel _that_ bad about moving the cache to `allOverriddenSymbols`.

hrhino:
FWIW: with a `Statistics.incCounter` in the non-cached branch of `allOverriddenSymbols`:

Before:
```
(library)
typer:   [info] #overridden symbol calculations: 3413
erasure: [info] #overridden symbol calculations: 20789

(reflect)
typer:   [info] #overridden symbol calculations: 22097
erasure: [info] #overridden symbol calculations: 43493

(compiler)
typer:   [info] #overridden symbol calculations: 45744
erasure: [info] #overridden symbol calculations: 77901
```

After:
```
(library)
typer:   [info] #overridden symbol calculations: 965
erasure: [info] #overridden symbol calculations: 18251

(reflect)
typer:   [info] #overridden symbol calculations: 18478
erasure: [info] #overridden symbol calculations: 39577

(compiler)
typer:   [info] #overridden symbol calculations: 40197
erasure: [info] #overridden symbol calculations: 72164
```

(cumulative stats, of course)

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2371,23 +2371,35 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
...
+          val builder = List.newBuilder[Symbol]
+          for (o <- owner.ancestors) {
+            overriddenSymbol(o).andAlso(builder += _)
+          }
+          builder.result()
+        } else Nil
+        overridingPeriod = currentPeriod
+        overridingCache = result
+        result
+      }
```
retronym: I'm reluctant to introduce this cache. This method is called on each decl in each class symbol (source or classpath based) during specialization, and these caches will hang around until the end of the compiler. It appears as though the caches would also consume O(N^2) space wrt the length of the override chain (the N-th deepest symbol would have a cache of N-1 elements).

I forget the exact reason that the cache in `isOverridingSymbol` got in the way of this bugfix, could you refresh my memory? My changes to `isOverridingSymbol` in #6197 will conflict here.

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
```diff
@@ -1349,23 +1351,37 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
       val constrainedType = MethodBType(lambdaParams.map(p => typeToBType(p.tpe)), typeToBType(lambdaTarget.tpe.resultType)).toASMType
       val samMethodType = methodBTypeFromSymbol(sam).toASMType
       val markers = if (addScalaSerializableMarker) classBTypeFromSymbol(definitions.SerializableClass).toASMType :: Nil else Nil
-      visitInvokeDynamicInsnLMF(bc.jmethod, sam.name.toString, invokedType, samMethodType, implMethodHandle, constrainedType, isSerializable, markers)
+      val overriddenMethods =
+        bridges.map(b => methodBTypeFromSymbol(b).toASMType)
```
retronym: I'm pretty sure that `methodBTypeFromSymbol` is sufficient to ensure that the backend include types within to `InnerClasses`, @lrytz can you confirm this.

src/compiler/scala/tools/nsc/transform/Delambdafy.scala
```diff
@@ -98,13 +101,21 @@ abstract class Delambdafy extends Transform with TypingTransformers with ast.Tre
       val isSerializable = samUserDefined == NoSymbol || samUserDefined.owner.isNonBottomSubClass(definitions.JavaSerializableClass)
       val addScalaSerializableMarker = samUserDefined == NoSymbol
 
+      val samBridges = logResultIf[List[Symbol]](s"will add SAM bridges for $fun", _.nonEmpty) {
+        userSamCls.fold[List[Symbol]](Nil) {
+          _.info.decls.collect {
+            case bridge: MethodSymbol if bridge hasFlag BRIDGE => bridge
+          }.toList
+        }
+      }
```
retronym: `userSamCls.findMembers(0L, BRIDGE)` might also work here (we know the only base class is `Object` without bridges).

src/reflect/scala/reflect/internal/StdAttachments.scala
```diff
@@ -51,10 +51,11 @@ trait StdAttachments {
     *
     * @param samTp the expected type that triggered sam conversion (may be a subtype of the type corresponding to sam's owner)
     * @param sam the single abstract method implemented by the Function we're attaching this to
+    * @param samCls the *synthetic* class representing the eventual implementation class (spun at runtime by LMF on the JVM)
     *
     * @since 2.12.0-M4
     */
-  case class SAMFunction(samTp: Type, sam: Symbol) extends PlainAttachment
+  case class SAMFunction(samTp: Type, sam: Symbol, samCls: Symbol) extends PlainAttachment
```
retronym: Perhaps we should call this `syntheticSamClass` here and in vals the reference it (e.g. `userSamCls` below) to make it really clear that it isn't the functional interface.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -2843,65 +2843,94 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+
+          if (samTp eq NoType) false
+          else {
+            /* Make a synthetic class symbol to represent the synthetic class that
+             * will be spun up by LMF for this function. This is necessary because
+             * it's possible that the SAM method might need bridges, and they have
+             * to go somewhere. Erasure knows to compute bridges for these classes
+             * just as if they were real templates extending the SAM type. */
+            val samCls = fun.symbol.owner.newClassWithInfo(
+              name = TypeName(s"${samTp.typeSymbol.name}$$${sam.nameString}$$samImpl"),
```
retronym: Do we need a name for this? I'd probably just use `tpnme.ANON_FUN_NAME`.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -934,17 +950,24 @@ abstract class Erasure extends InfoTransform
...
         val (bridges, toBeRemoved) = bridgeDefs(base)
         if (bridges.isEmpty) stats
         else (stats filterNot (stat => toBeRemoved contains stat.symbol)) ::: bridges
       }
 
+    def addBridgesToLambda(lambdaClass: Symbol): Unit = {
+      assert(phase == currentRun.erasurePhase, phase)
+      assert(lambdaClass.isClass, lambdaClass)
+      val eb = new EnterBridges(unit, lambdaClass)
+      eb.computeAndEnter(includeDeferred = true)
```
retronym: Could you add a comment on the reason for `includeDeferred = true`? It isn't clear to me.

src/reflect/scala/reflect/internal/Symbols.scala
hrhino: Yeah, I saw those changes and they're probably better. I did this because I read the "we should" in your comment to mean "I should, and now". I'm just going to drop that commit from this PR.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
hrhino: We don't. That's better.

src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
hrhino: I'll come up with a test to be sure.

src/compiler/scala/tools/nsc/transform/Delambdafy.scala
hrhino: Seems to work, and you're right that our base types are only the SAM trait and `Object`. I'll change it.

src/compiler/scala/tools/nsc/transform/Erasure.scala
hrhino: Ugh. I recall needing this, but I didn't comment it and now I can't remember. I'll poke at it until I find out why.

hrhino: Ah. I think it is that at one point I was cloning the SAM's `MethodSymbol` and not changing its flags, so it would show up deferred here. Good catch.

retronym (approved):
The end result LGTM!

@adriaanm Could you please review as well, in particular, the changes around `inferSamType`?

@sjrd I believe this version should be relatively straight forward for you to support in ScalaJS. Out of curiousity, would your life be easier if, under ScalaJS mode, we just desugared user defined SAMs to anonymous classes after typechecking? Or do you have a runtime component analagous to `LambdaMetafactory`?

hrhino:
@retronym thanks :D Sorry for biting off a bit much. 

I'm willing to jump into the scala.js code this weekend to update it for this, assuming that this PR is in vaguely final shape.

sjrd:
> Out of curiousity, would your life be easier if, under ScalaJS mode, we just desugared user defined SAMs to anonymous classes after typechecking?

It would, but only ever so slightly. It would allow us to remove [`synthesizeSAMWrapper`](https://github.com/scala-js/scala-js/blob/8e9ae3be4f923cf6f9294b184331640e81cc4a84/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L5106), which I mentioned above. But it's a pretty small thing, and pretty clean too. I suspect the degradation in terms of compile time performance (more trees through the pipeline) is not worth it.

What is more annoying for us is actually the early desugaring of SAMs for `js.FunctionN`s and `js.ThisFunctionN`s. Those are not JVM SAMs because they inherit from classes, so scalac desugars them. But then we *have to* reverse-engineer them in the back-end as they must become JavaScript functions. That's done in [`tryGenAnonFunctionClassGeneric`](https://github.com/scala-js/scala-js/blob/8e9ae3be4f923cf6f9294b184331640e81cc4a84/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala#L4845), which is really ugly and fragile, as it can be broken by small encoding changes that would be done in the earlier phases of scalac.

> Or do you have a runtime component analagous to LambdaMetafactory?

We might eventually have a link-time component for that, but I'm not sure. For 0.6.x and even until 1.0.0, that is very unlikely to happen.

> I'm willing to jump into the scala.js code this weekend to update it for this, assuming that this PR is in vaguely final shape.

Cool :) Make sure to work on the 0.6.x branch if you do. We regularly forward merge 0.6.x into master.

retronym:
@sjrd Just mentioning you to highlight that this is (finally) merged and you can plan the downstream changes.

sjrd:
Thanks. I was keeping an eye on this one ;)

sjrd:
FTR, here is the Scala.js PR: https://github.com/scala-js/scala-js/pull/3260

adriaanm:
I haven't fully gotten to the bottom of this, but is this missing a check that the bridges are indeed for the SAM method (i.e. find member should also pass in `sam.name`)? This looks to have regressed when bridges for other methods are seen as needed, causing https://github.com/scala/bug/issues/11373

smarter:
> Java does this differently: it generates a default method in the subinterface overriding the superinterface method.
> ...
> Perhaps this could be revisited for 2.13, which would simplify things for SJS

Looks like it wasn't as 2.13 contains code from this PR, was this explored at all? Also what happens if I write a lambda in Java that calls a Scala trait with such an overridden method, does Java emits code that crashes at runtime? Asking since this is affecting Dotty too: https://github.com/lampepfl/dotty/issues/10068

lrytz:
> was this explored at all?

I don't think so...

> what happens if I write a lambda in Java that calls a Scala trait with such an overridden method

It seems `javac` knows how to handle this. Quick test (lmk if I got it wrong):

```
$> cat A.scala
trait A {
  def f(): Object
}
trait B extends A {
  def f(): String
}
$> cat Test.java
public class Test {
  public interface JA {
    Object foo();
  }
  public interface JB extends JA {
    String foo();
  }

  static B  b  = () -> "";
  static JB jb = () -> "";

    System.out.println(Test.b);
    System.out.println(Test.jb);
  }
}
$> sc A.scala
$> javac -cp . Test.java
$> java Test
Test$$Lambda$1/321001045@1b28cdfa
Test$$Lambda$2/1705736037@eed1f14
$> asmc Test.class

...

    INVOKEDYNAMIC f()LB; [
      // handle kind 0x6 : INVOKESTATIC
      java/lang/invoke/LambdaMetafactory.altMetafactory(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite;
      // arguments:
      ()Ljava/lang/String;, 
      // handle kind 0x6 : INVOKESTATIC
      Test.lambda$static$0()Ljava/lang/String;, 
      ()Ljava/lang/String;, 
      4, 
      1, 
      ()Ljava/lang/Object;
    ]

...

    INVOKEDYNAMIC foo()LTest$JB; [
      // handle kind 0x6 : INVOKESTATIC
      java/lang/invoke/LambdaMetafactory.metafactory(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;
      // arguments:
      ()Ljava/lang/String;, 
      // handle kind 0x6 : INVOKESTATIC
      Test.lambda$static$1()Ljava/lang/String;, 
      ()Ljava/lang/String;
    ]

...
```

The indy-lambda for `B` asks for a bridge method, the one for `JB` not.

smarter:
Interesting, so it seems that no matter what we do we still need to be able to generate indy-lambdas that ask for bridges in the backend. I assume javac does that for backward-compatibility with pre-java8 interfaces?

smarter:
Thinking about this more, it seems that we cannot assume that a sub-interface coming from Java implements the necessary bridge(s) and so we always need to do the altMetafactory dance: the problem is that under separate compilation we can easily check if the subinterface contains bridges, but under joint compilation of java/scala sources we don't have this information, and we cannot assume that the bridge is always there since it's going to depend on the version of javac (and presumably isn't something mandated by the spec in any version although I haven't checked), to keep the compilation output stable we cannot do something different under joint or separate compilation, therefore we have to be pessimistic and never assume that the bridge is present.

PR #6088, opened 2017-09-19
Translate a virtual call on sealed trait/class with a single concrete subclass to a virtual call on that subclass.

TomasMikula:
Translate a virtual call on sealed trait/class with a single concrete subclass to a virtual call on that subclass.

This enables the call to later be inlined.

As an example, consider

```scala
sealed trait Foo {
  def foo(s: String): Unit
}

class FooImpl extends Foo {
  @inline
  final def foo(s: String): Unit = ()
}

class Test extends {
  def go(f: Foo): Unit = f.foo("Hi")
}
```

With Scala 2.12.3,
```
scalac -opt:l:inline '-opt-inline-from:**' -opt-warnings test.scala
```

compiles the `Test#go` method to

```
 0: aload_1
 1: ldc           #13                 // String Hi
 3: invokeinterface #19,  2           // InterfaceMethod Foo.foo:(Ljava/lang/String;)V
 8: return
```

With this PR, `f` in `f.foo("Hi")` is first cast to `FooImpl` and then a (virtual) call to `FooImpl#foo` is made. However, this virtual call to a `final` method is later inlined, resulting in the following bytecode:

```
 0: aload_1
 1: checkcast     #13                 // class FooImpl
 4: ifnonnull     9
 7: aconst_null
 8: athrow
 9: return
```

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala
@@ -662,7 +662,6 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
               else InvokeStyle.Virtual
 
             if (invokeStyle.hasInstance) genLoadQualifier(fun)
-            genLoadArguments(args, paramTKs(app))
 
             val Select(qual, _) = fun // fun is a Select, also checked in genLoadQualifier
             if (sym == definitions.Array_clone) {
@@ -679,12 +678,13 @@ abstract class BCodeBodyBuilder extends BCodeSkelBuilder {
               //   http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/87ee5ee27509/src/share/vm/interpreter/linkResolver.cpp#l439
               // Example: `class C { override def clone(): Object = "hi" }`
               // Emitting `def f(c: C) = c.clone()` as `Object.clone()` gives a VerifyError.
+              assert(args.isEmpty)
               val target: String = tpeTK(qual).asRefBType.classOrArrayType
               val methodBType = methodBTypeFromSymbol(sym)
               bc.invokevirtual(target, sym.javaSimpleName.toString, methodBType.descriptor, app.pos)
               generatedType = methodBType.returnType
             } else {
-              val receiverClass = if (!invokeStyle.isVirtual) null else {
+              val (receiverClass, method) = if (!invokeStyle.isVirtual) (null, sym) else {
                 // receiverClass is used in the bytecode to as the method receiver. using sym.owner
                 // may lead to IllegalAccessErrors, see 9954eaf / aladdin bug 455.
                 val qualSym = qual.tpe.typeSymbol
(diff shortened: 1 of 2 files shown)


TomasMikula:
I don't quite understand why I had to add the following checks. Help is appreciated with understanding them.

1. Check that the single concrete subclass actually has the method from the superclass

   ```scala
   getMemberIfDefined(s.head, sym.name).hasFlag(Flags.METHOD)
   ```

   If omitted, script/REPL tests fail with errors like

   ```
   scala.reflect.internal.FatalError: class iw does not have a member $line5$$read$$iw$$iw$$$outer
   ```

2. Check that the single concrete subclass is not a dummy local child

   ```scala
   s.head.rawname != tpnme.LOCAL_CHILD
   ```

sjrd:
Unfortunately, I do not think this has the expected properties in terms of binary compatibility and separate compilation, especially for non-public subclasses.

If my `FooImpl` is private to the file (which, given the pattern, is likely), I expect to be able to rename it and/or add/remove subclasses without impacting binary compatibility and separate compilation outside of this file. But with your change, that is not the case anymore: such changes will actually break binary compatibility and separate compilation.

TomasMikula:
Yeah, keeping implementation private is the intended use case. Bummer. 😞 

I could add a check that the concrete implementation is also visible, but that limits the use case.

Or maybe this logic could be implemented in the inliner and then it would be OK?

Blaisorblade:
First, thanks for attaching this problem!

> Or maybe this logic could be implemented in the inliner and then it would be OK?

That's interesting but tricky.

The JIT inliner has this logic without binary compatibility issues because it's at runtime, but it's not easy to anticipate when it triggers. But we know that leaving everything to the JIT doesn't always work, so I think this optimization is still worthwhile.

Otherwise, you can't improve binary compatibility, just have a usage model with different expectations. [Dotty-linker](https://github.com/dotty-linker/dotty) avoids all these problems, simply because it's supposed to be used on a whole program.

The Scalac inliner seems subject to the same binary compatibility limitations, but IIUC there is one difference with `@inline`: who uses `@inline` explicitly makes the implementation part of the ABI. A new attribute which documents such semantics would have the same effect, and existing code would stay unoptimized unless it requests such an attribute.
If *that* were accepted, though, the attribute couldn't be added to the `scala-library` JAR before 2.13, because of forward compatibility.
You'd need a separate library with different guarantees.

Worst-case, for 2.12 you could have a compiler plugin that monkey-patches Scalac, like Miles Sabin did with https://github.com/milessabin/si2712fix-plugin. It risks being a much worse maintenance tradeoff, since 2.12 is much younger than 2.10. Not sure if Typelevel Scala's compatibility policy allows for such scenarios.

TomasMikula:
If I put the `@inline` annotation on the abstract method in the public interface, and the inliner then inlines the (single) implementation, wouldn't the compatibility guarantees be the same as with any other `@inline` annotated method?

lrytz:
We can handle this case in the optimizer / inliner, it falls within the same binary compatibility (non-) guarantees that we already have. When the inliner is enabled, the default heuristics may (and does) pick methods for inlining that don't have the `@inline` annotation. This means that you cannot control the ABI with the optimizer enabled (except for `-opt-inline-from`). So adding this case is fine.

lrytz:
I think the starting place is to extract the necessary bits of information when building the involved `ClassBType`s: https://github.com/scala/scala/blob/07310fa675304cfce82dc830e980758cbcec3292/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala?utf8=%E2%9C%93#L585

Blaisorblade:
@TomasMikula I assumed `@inline` doesn't expose the class name—but it seems I missed the memo on the inliner's ABI guarantees. Sorry. I'd still like to have the ABI guarantees spelled out somewhere, beyond "use MIMA", but maybe I'm indeed asking too much.
IIUC, this way, if a library suggests inlining Scalaz into your app, it can't expose only parts of its implementation. OTOH, not sure if the intended users have ABI guarantees.

lrytz:
From the [2.12.0 release notes](http://www.scala-lang.org/news/2.12.0/):

> When inlining code from the classpath, you must ensure that all dependencies have exactly the same versions at compile time and run time.
> If you are building a library to publish on Maven Central, you should not inline code from dependencies. Users of your library might have different versions of those dependencies on the classpath, which breaks binary compatibility.

lrytz:
Status of this PR: the idea is nice, the implementation should be changed to represent the necessary data in the `ClassBType`'s `inlineInfo`.

adriaanm:
I'm closing this PR because it has been dormant for a while now. We'd be very happy to receive a new PR (targeting 2.13.0-M5) that addresses points raised by the reviewer.

PR #6089, opened 2017-09-19
Deprecate assignments in argument position

lrytz:
Issue a deprecation warning when an assignment expression (`x = e`)
in a parameter position is treated as an assignment, not a named
argument.

In 2.13 this will be an error, `f(x = e)` will be restricted to
named arguments and never be treated as an assignment to `x`.

See scala/scala-dev#426


lrytz:
Implemented the 2.13 behavior under `-Xsource:2.13`

adriaanm (approved):

SethTisue:
seen in the community build, https://github.com/scala/community-builds/issues/612

it looks to me like the warning is being issued appropriately

PR #6090, opened 2017-09-19
Future.firstCompletedOf introduces a memory leak when at least one of the futures is long-lived

eyalfa:
introduce a specialized function implementation that clears the reference to the result promise on its first execution.

Fixes scala/bug#10513

cc: @axel22 , @viktorklang

### DIFF
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -9,12 +9,11 @@
 package scala.concurrent
 
 import scala.language.higherKinds
-
 import java.util.concurrent.{CountDownLatch, TimeUnit}
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
 
 import scala.util.control.NonFatal
-import scala.util.{Try, Success, Failure}
+import scala.util.{Failure, Success, Try}
 import scala.concurrent.duration._
 import scala.collection.generic.CanBuildFrom
 import scala.reflect.ClassTag
@@ -677,8 +676,13 @@ object Future {
    */
   def firstCompletedOf[T](futures: TraversableOnce[Future[T]])(implicit executor: ExecutionContext): Future[T] = {
     val p = Promise[T]()
-    val completeFirst: Try[T] => Unit = p tryComplete _
-    futures foreach { _ onComplete completeFirst }
+    val firstCompleteHandler = new AtomicReference[Promise[T]](p) with (Try[T] => Unit) {
+      override def apply(v1: Try[T]): Unit = getAndSet(null) match {
+        case null => ()
+        case some => some tryComplete v1
+      }
+    }
+    futures foreach { _ onComplete firstCompleteHandler }
     p.future
   }
 
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -9,12 +9,11 @@
 package scala.concurrent
 
 import scala.language.higherKinds
-
 import java.util.concurrent.{CountDownLatch, TimeUnit}
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
 
 import scala.util.control.NonFatal
-import scala.util.{Try, Success, Failure}
+import scala.util.{Failure, Success, Try}
 import scala.concurrent.duration._
 import scala.collection.generic.CanBuildFrom
 import scala.reflect.ClassTag
@@ -677,8 +676,15 @@ object Future {
    */
   def firstCompletedOf[T](futures: TraversableOnce[Future[T]])(implicit executor: ExecutionContext): Future[T] = {
     val p = Promise[T]()
-    val completeFirst: Try[T] => Unit = p tryComplete _
-    futures foreach { _ onComplete completeFirst }
+    val firstCompleteHandler = new AtomicReference[Promise[T]](p) with (Try[T] => Unit) {
+      override def apply(v1: Try[T]): Unit = {
+        val p = getAndSet(null)
+        if( null != p ){
+          p tryComplete v1
+        }
+      }
+    }
+    futures foreach { _ onComplete firstCompleteHandler }
     p.future
   }
 
--- a/src/library/scala/concurrent/Future.scala
+++ b/src/library/scala/concurrent/Future.scala
@@ -677,11 +677,9 @@ object Future {
   def firstCompletedOf[T](futures: TraversableOnce[Future[T]])(implicit executor: ExecutionContext): Future[T] = {
     val p = Promise[T]()
     val firstCompleteHandler = new AtomicReference[Promise[T]](p) with (Try[T] => Unit) {
-      override def apply(v1: Try[T]): Unit = {
-        val p = getAndSet(null)
-        if( null != p ){
-          p tryComplete v1
-        }
+      override def apply(v1: Try[T]): Unit = getAndSet(null) match {
+        case null => ()
+        case some => some tryComplete v1
       }
     }
     futures foreach { _ onComplete firstCompleteHandler }
--- a/test/files/run/t10513.scala
+++ b/test/files/run/t10513.scala
@@ -6,27 +6,28 @@ import scala.util.{Random, Try}
 import ExecutionContext.Implicits.global
 
 /** This test uses recursive calls to Future.flatMap to create arrays whose
- *  combined size is slightly greater than the JVM heap size. A previous
- *  implementation of Future.flatMap would retain references to each array,
- *  resulting in a speedy OutOfMemoryError. Now, each array should be freed soon
- *  after it is created and the test should complete without problems.
- */
+  * combined size is slightly greater than the JVM heap size. A previous
+  * implementation of Future.flatMap would retain references to each array,
+  * resulting in a speedy OutOfMemoryError. Now, each array should be freed soon
+  * after it is created and the test should complete without problems.
+  */
 object Test {
- val arrSz = 50 * 10000
-  val numFutures = 4000
-
-  val rng = new Random()
-
 
   def main(args: Array[String]) {
+    val arrSz = 50 * 10000
+    val numFutures = 4000
+
+    val rng = new Random()
     val longStandingPromise = Promise[Nothing]
 
-    val futures = List.tabulate(numFutures){ i =>
+    val futures = List.tabulate(numFutures) { i =>
       val arr = Array.tabulate(arrSz)(identity)
       val idx = rng.nextInt(arrSz)
-      val f1 = Future{ arr }
+      val f1 = Future {
+        arr
+      }
       val f2 = Future.firstCompletedOf(List(longStandingPromise.future, f1))
-      f2.map( arr => arr(idx))
+      f2.map(arr => arr(idx))
     }
     val fSeq = Future.sequence(futures)
     val finalF = fSeq.map(_.sum)
--- a/test/files/run/t10513.scala
+++ b/test/files/run/t10513.scala
@@ -1,5 +1,6 @@
 //package scala.concurrent
 import scala.concurrent._
+import duration._
 
 import scala.util.{Random, Try}
 import ExecutionContext.Implicits.global
@@ -29,6 +30,6 @@ object Test {
     }
     val fSeq = Future.sequence(futures)
     val finalF = fSeq.map(_.sum)
-    val res = Await.result(finalF, duration.Duration.Inf)
+    val res = Await.result(finalF, 2.minutes)
   }
 }
--- a/test/files/run/t10513.scala
+++ b/test/files/run/t10513.scala
@@ -12,7 +12,7 @@ import ExecutionContext.Implicits.global
  */
 object Test {
  val arrSz = 50 * 10000
-  val numFutures = 10000
+  val numFutures = 4000
 
   val rng = new Random()
 
--- a/test/files/run/t10513.scala
+++ b/test/files/run/t10513.scala
@@ -1,0 +1,34 @@
+//package scala.concurrent
+import scala.concurrent._
+
+import scala.util.{Random, Try}
+import ExecutionContext.Implicits.global
+
+/** This test uses recursive calls to Future.flatMap to create arrays whose
+ *  combined size is slightly greater than the JVM heap size. A previous
+ *  implementation of Future.flatMap would retain references to each array,
+ *  resulting in a speedy OutOfMemoryError. Now, each array should be freed soon
+ *  after it is created and the test should complete without problems.
+ */
+object Test {
+ val arrSz = 50 * 10000
+  val numFutures = 10000
+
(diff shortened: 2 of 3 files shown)


test/files/run/t10513.scala
```diff
@@ -0,0 +1,35 @@
...
+object Test {
+ val arrSz = 50 * 10000
+  val numFutures = 10000
+
+  val rng = new Random()
+
+
+  def main(args: Array[String]) {
+    val longStandingPromise = Promise[Nothing]
+    //val shortTermFuture = Promise[Int]
```
axel22: For consistency and cleanliness, I suggest that you remove the commented line.

eyalfa:
signed CLA

test/files/run/t10513.scala
```diff
@@ -0,0 +1,35 @@
...
+      val idx = rng.nextInt(arrSz)
+      val f1 = Future{ arr }
+      val f2 = Future.firstCompletedOf(List(longStandingPromise.future, f1))
+      f2.map( arr => arr(idx))
+    }
+    val fSeq = Future.sequence(futures)
+    val finalF = fSeq.map(_.sum)
+    val res = Await.result(finalF, duration.Duration.Inf)
+  }
+}
```
axel22: Not sure if empty lines at the end of file are mandatory, but I think they are recommended.

axel22 (approved):
LGTM

eyalfa:
guys, any idea what does this failure stand for?

> [error]   - library/mimaReportBinaryIssues failed: java.lang.RuntimeException: scala-library: Binary compatibility check failed!

found in [build log](https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/5627/console)

src/library/scala/concurrent/Future.scala
```diff
@@ -677,11 +676,20 @@ object Future {
...
   def firstCompletedOf[T](futures: TraversableOnce[Future[T]])(implicit executor: ExecutionContext): Future[T] = {
     val p = Promise[T]()
-    val completeFirst: Try[T] => Unit = p tryComplete _
-    futures foreach { _ onComplete completeFirst }
+    val firstCompleteHandler = new FirstCompleteHandler(p)
+    futures foreach { _ onComplete firstCompleteHandler }
     p.future
   }
 
+  private  class FirstCompleteHandler[T](p : Promise[T]) extends AtomicReference[Promise[T]](p) with (Try[T] => Unit) {
```
retronym: The addition of this inner class is triggering the binary compatibility error you saw. In bytecode, a private class is actually public, so a Java source file could see it and reference it when compiling against 2.12.4, and then fail to link if run against 2.12.3. (meaning we haven't kept our forwards binary compatibility promise.)

We can either add a [whitelist entry](https://github.com/scala/scala/blob/2.12.x/src/library/mima-filters/2.12.0.forwards.excludes), or more simply, define this as an anonymous class inside `firstCompletedOf`

eyalfa: since I'm the new kid on the block :sunglasses: I'd appreciate guidance here, which path is preferable? the white list entry or the anonymous class?

I'm also not so sure about the issue introduced by this class, code running with previous versions gets the original impl (lambda), code running with my patch gets the new specialized function impementation... or is it a worry about code compiled against my patch and executed with a previous version of scala? this code would have to make quite an 'effort' to get to this situation as it'd have to either use reflection or Java code in order to access the newly added nested class...

eyalfa:
guys, seems mima issue is fixed, what's this 'combined' thingy?

SethTisue:
> guys, seems mima issue is fixed, what's this 'combined' thingy?

in this repo, we ask that very commit in a PR be green before we merge it, to assist with `git bisect` and such, and as an invariant that reviewers and maintainers can rely on when looking at the history

while the PR is still being reviewed, it can be helpful to preserve the series of commits showing how the PR progressed. but before final merge, we ask for a squash and/or rebase with green commits only.

eyalfa:
So am I supposed to squash now? I know that other projects (i.e.
spark)squash on merge, I think github allows you to set this as a policy
for your project.

Bottom line, I'd appreciate guidance on the next steps here.

SethTisue:
we ask that you do the squashing and force-pushing, so that Jenkins can validate the squashed commits before final merge

eyalfa:
Ok, will do once near a computer

eyalfa:
that's it, pushed -f a squashed commit

test/files/run/t10513.scala
```diff
@@ -0,0 +1,34 @@
...
+    val futures = List.tabulate(numFutures){ i =>
+      val arr = Array.tabulate(arrSz)(identity)
+      val idx = rng.nextInt(arrSz)
+      val f1 = Future{ arr }
+      val f2 = Future.firstCompletedOf(List(longStandingPromise.future, f1))
+      f2.map( arr => arr(idx))
+    }
+    val fSeq = Future.sequence(futures)
+    val finalF = fSeq.map(_.sum)
+    val res = Await.result(finalF, duration.Duration.Inf)
```
viktorklang: This is a bit of a recipe for a hanging buildserver, we need a sensible timeout.

test/files/run/t10513.scala
```diff
@@ -0,0 +1,34 @@
...
+ *  combined size is slightly greater than the JVM heap size. A previous
+ *  implementation of Future.flatMap would retain references to each array,
+ *  resulting in a speedy OutOfMemoryError. Now, each array should be freed soon
+ *  after it is created and the test should complete without problems.
+ */
+object Test {
+ val arrSz = 50 * 10000
+  val numFutures = 10000
+
+  val rng = new Random()
```
viktorklang: Why are these outside of `main`?

src/library/scala/concurrent/Future.scala
```diff
@@ -677,8 +676,15 @@ object Future {
    */
   def firstCompletedOf[T](futures: TraversableOnce[Future[T]])(implicit executor: ExecutionContext): Future[T] = {
     val p = Promise[T]()
-    val completeFirst: Try[T] => Unit = p tryComplete _
-    futures foreach { _ onComplete completeFirst }
+    val firstCompleteHandler = new AtomicReference[Promise[T]](p) with (Try[T] => Unit) {
+      override def apply(v1: Try[T]): Unit = {
```
viktorklang:
```scala
override def apply(v1: Try[T]): Unit = getAndSet(null) match {
  case null => ()
  case some => some tryComplete v1
}
```

test/files/run/t10513.scala
eyalfa: first two are simply constants, rng can be either static class member or local variable, can't really see how it matters for a test class being executed just once :smile: .
I actually recall now, this used to be a scalatest suite with one failing test and a second one with the fix implemented inline, so it made sense to stay DRY.

test/files/run/t10513.scala
eyalfa: checked on my 4 core 16GB laptop, original failing test fails after 20 seconds, same test with the fix applied completed after 74 seconds.
so few conclusions here:
1. we can definitely decrease the number of iterations required for this test to make its point.
2. we can calculate this magic number of iteration based on the amount of memory allocated for this test, @viktorklang  can you please point me to the section in SBT that determines that?

src/library/scala/concurrent/Future.scala
eyalfa: matter of style I guess, will adopt to the local style :sunglasses:

test/files/run/t10513.scala
eyalfa: just took a closer look at partest's output:
> Java options are:    -Xmx1024M -Xms64M

so I guess I can set the number of iteration such that total memory consumption is ~2gb, seems 4000 iterations are enough.

4000 iterations are enough to fail the original test and takes ~35 seconds to complete on my LT, @viktorklang do you think a 1 minute timeout suffice here?

viktorklang: @eyalfa 10mins would be ok too :)

test/files/run/t10513.scala
viktorklang: I'd prefer if they were contained within `main` :)

src/library/scala/concurrent/Future.scala
viktorklang: Awesome, thank you :+1: :)

test/files/run/t10513.scala
som-snytt: partest can use a `.javaopts` file, there is a `./test/files/run/bridges.javaopts`. Also wondered if this can be written as a junit test; `test/junit/scala/collection/mutable/MutableListTest.scala` uses `assertNotReachable`, but maybe that's not practical here. There were previous tests around memory leaks, but I don't remember which offhand.

eyalfa:
This is basically a smoke test, I think making this part if a junit suit
makes it very sensitive to env modifications, if for some reason there's a
need to increase memory/cores whatever, this test becomes meaningless.
Btw, I took this testing approach from a blog post by @viktorklang where he
explains about promise linking (currently on cell phone,can't find the
link), his test also tested for OOM (when using flatMap) so I think it's
reasonable to take the same approach here.

som-snytt:
The test in this gist fails before your change and succeeds after, FWIW.

https://gist.github.com/som-snytt/f0cb3468662de2977e635c8ffebff028

Also at https://github.com/som-snytt/scala/tree/review/10513.

Apologies in advance if I missed something subtle and it's not useful to you.

eyalfa:
@som-snytt , this is actually a much better approach to test this, I definitely prefer it over the 'driving the jvm to edge and observe the exception' approach :smiley: 
@axel22 , @viktorklang , is it acceptable to add this to the existing `DefaultPromiseTest`? or should I place it in a different test suite?

test/files/run/t10513.scala
```diff
@@ -0,0 +1,35 @@
...
+import ExecutionContext.Implicits.global
+
+/** This test uses recursive calls to Future.flatMap to create arrays whose
+ *  combined size is slightly greater than the JVM heap size. A previous
+ *  implementation of Future.flatMap would retain references to each array,
+ *  resulting in a speedy OutOfMemoryError. Now, each array should be freed soon
+ *  after it is created and the test should complete without problems.
+ */
+object Test {
+ val arrSz = 50 * 10000
```
axel22: Please make sure to indent this with 2 spaces.

axel22:
@eyalfa Maybe a different file is a better option, since it's a different test type.

som-snytt:
Since it is a test of `Future.firstCompletedOf`, I'd suggest the file I suggested.

I found the old PR where retronym preferred following references: https://github.com/scala/scala/pull/4133

He also linked to another issue around GC and reachability. I was looking to see if there was interest in a separate set of tests that might be sensitive to settings etc.

axel22:
sgtm!

test/files/run/t10513.scala
som-snytt: Thanks I was also irked. I think those vals are scheduled to be moved anyway. But it makes me miss a formatting option, scalafmt etc. It's the little things...

eyalfa:
@som-snytt, the fix is very specific to Default promise's Impl, so saying this is a 'Future.xxx' test has a stronger meaning as it sets the bar for other implementations and kind of extends the specs for future Future implementations... Though it's quite possible other currently existing futures also support it (fulfilled,never).

Guys, what do you think?

som-snytt:
Yes, you're saying `firstCompletedOf` must exhibit a certain behavior, and a passing test shows that it now fulfills that expectation. If, in the future, the test fails, then that is a regression. How the method is implemented (in terms of `Promise`) is not relevant. In other words, this is not a test of `Promise`.

eyalfa:
@som-snytt , with your permission, I'd like to merge your review branch into mine so I can use your test-code while maintaining your credit for the this code.

guys, do you find @som-snytt's test sufficient or do you want me to keep my original test as well? (in my original use-case, this ended up with an OOM)

som-snytt:
@eyalfa Help yourself, and thanks for the consideration. My own opinion is to also keep the original test, since it's the pudding in which lies the proof. The `run` tests still get isolated JVMs. I didn't track down if that previous issue (which also had an OOM) got a run test to show OOM, though I suspect not. Maybe later the community can come up with a scheme for running such tests, for example under `test/files/flaky` (kidding, almost) if the test proves unstable. They retire flaky tests all the time; did they get rid of `pending`? There used to be an attic for sad tests.

eyalfa:
guys, seems like an unrelated failure to me can anyone please have a look?

src/library/scala/concurrent/Future.scala
```diff
@@ -9,12 +9,11 @@
...
 
 import scala.language.higherKinds
-
 import java.util.concurrent.{CountDownLatch, TimeUnit}
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
 
 import scala.util.control.NonFatal
-import scala.util.{Try, Success, Failure}
+import scala.util.{Failure, Success, Try}
```
viktorklang, line 16: Is this change really warranted?

viktorklang:
LGTM

eyalfa:
@viktorklang , thank you very much. what's the next step? how will this get merged?

viktorklang:
@eyalfa Hmm, @SethTisue?

src/library/scala/concurrent/Future.scala
som-snytt: +1 intellij or scalafmt needs an option for that, if they don't have it. But it would need heuristics to decide Success before Failure, maybe inspecting the implementations of `toEither` or `toOption` etc.

eyalfa: must say I didn't modify this line, Intellij does its magic for me :sunglasses:

eyalfa:
well, @SethTisue ? anyone else?

eyalfa:
@SethTisue , @viktorklang , @axel22 (and anyone else I might forgot),
what's the next step here?

axel22:
@eyalfa If everything is addressed, then I think that you just have to wait until the someone with adequate rights merges this PR.

eyalfa:
@axel22 , as far as I can see everything is addressed according to the review comments I had on this PR, I'm just new to this repo and I don't know what are the procedures to get things merged here, aka who has write permissions and alike...
in other repos I've contributed to, there's a group of maintainers (spark calls them committers) that have write permissions and once the PR is approved by one of them, this person also merges it, it seems this repo is managed a bit differently so I'd like to know who's the person taking care of the merges and how is it managed.

thanks.

axel22:
Everything seems addressed to me too - and thank you for your contribution!
I think that the responsible maintainer involved in this PR is @SethTisue - I'm sure he'll merge this as soon as he gets a chance.

SethTisue:
checking with @adriaanm to see if we've missed the cutoff for 2.12.4

SethTisue:
yeah. I'll merge this soon after 2.12.4 actually gets cut

eyalfa:
@SethTisue ,
I think it's out: https://github.com/scala/scala/milestone/64

PR #6091, opened 2017-09-21
Drop useless failing assert in `mkCast`

adriaanm:
I added it in e1c732d to encode an assumption, but it's
not an invariant that must be enforced (erasure will erase anyway).

More generally / longer term, we should reevaluate the use of `dealias`
for types that are meant to approximate their equivalent in the erased
type system.

Specifically, for `mkCastPreservingAnnotations` I think we should
either `normalize` or do nothing, but the half measure of `dealias`
does not make sense, as the logic behind a cast operates on the fully
normalized type, not just on a dealiased type (think refinements with
type aliases). It would be ok to do nothing here, because erasure will
convert the type to something that can be casted anyway.

Fixes scala/bug#10088

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/TreeGen.scala
+++ b/src/compiler/scala/tools/nsc/ast/TreeGen.scala
@@ -146,7 +146,6 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
     debuglog("casting " + tree + ":" + tree.tpe + " to " + pt + " at phase: " + phase)
     assert(!tree.tpe.isInstanceOf[MethodType], tree)
     assert(!pt.isInstanceOf[MethodType], tree)
-    assert(pt eq pt.normalize, tree +" : "+ debugString(pt) +" ~>"+ debugString(pt.normalize))
     atPos(tree.pos) {
       mkAsInstanceOf(tree, pt, any = !phase.next.erasedTypes, wrapInApply = isAtPhaseAfter(currentRun.uncurryPhase))
     }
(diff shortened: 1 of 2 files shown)


src/compiler/scala/tools/nsc/ast/TreeGen.scala
```diff
@@ -156,6 +155,9 @@ abstract class TreeGen extends scala.reflect.internal.TreeGen with TreeDSL {
   // let's assume for now annotations don't affect casts, drop them there, and bring them back using the outer Typed tree
   def mkCastPreservingAnnotations(tree: Tree, pt: Type) =
     Typed(mkCast(tree, pt.withoutAnnotations.dealias), TypeTree(pt))
+    // ^^^ I think we should either normalize or do nothing, but the half measure of dealias does not make sense,
+    // as the logic behind a cast operates on the fully normalized type, not just on a dealiased type (think refinements with type aliases).
+    // It would be ok to do nothing here, because erasure will convert the type to something that can be casted anyway.
```
retronym, line 160: I was in the neighbourhood in https://github.com/scala/scala/pull/5722/files#diff-36cc56dc4b99ae4c41d8f122c5e9dce9R150 and the sky didn't fall when I added auto-dealiasing.

retronym (approved):

src/compiler/scala/tools/nsc/ast/TreeGen.scala
adriaanm: I'll follow up with cleanup in 2.13.x

PR #6092, opened 2017-09-21
Remove logic to check if a named arg could be an assignment (and enable `-Xsource:2.13` by default)

lrytz:
An `x = e` expression in parameter position is now a named argument,
syntactically, unconditionally.

Fixes scala/scala-dev#426


lrytz:
Hehe, our current 2.13.x branch doesn't compile with `-Xsource:2.13`
```
[error] /home/jenkins/workspace/scala-2.13.x-validate-test/src/library/scala/sys/process/BasicIO.scala:165: type mismatch;
[error]  found   : String
[error]  required: () => String
[error]     try processLinesFully(processLine)(reader.readLine)
[error]                                               ^
```

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -133,7 +135,7 @@ trait ScalaSettings extends AbsScalaSettings
   val sourceReader       = StringSetting       ("-Xsource-reader", "classname", "Specify a custom method for reading source files.", "")
   val reporter           = StringSetting       ("-Xreporter", "classname", "Specify a custom reporter for compiler messages.", "scala.tools.nsc.reporters.ConsoleReporter")
   val strictInference    = BooleanSetting      ("-Xstrict-inference", "Don't infer known-unsound types")
-  val source             = ScalaVersionSetting ("-Xsource", "version", "Treat compiler input as Scala source for the specified version, see scala/bug#8126.", initial = ScalaVersion("2.12"))
+  val source             = ScalaVersionSetting ("-Xsource", "version", "Treat compiler input as Scala source for the specified version, see scala/bug#8126.", initial = ScalaVersion("2.13"))
```
adriaanm, line 138: good catch!

src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
```diff
@@ -46,9 +46,10 @@ trait ContextErrors {
...
   /**
-   * Marks a TypeError that was constructed from a CyclicReference (under silent).
-   * This is used for named arguments, where we need to know if an assignment expression
-   * failed with a cyclic reference or some other type error.
+   * Marks a TypeError that was constructed from a CyclicReference (under silent). This was used
+   * up to 2.12 to identify cyclic references when type checking a named argument as if it were an
+   * assignment (to figure out if the expression was a valid assignment, a4a892fb01). Since 2.13 we
+   * could use a normal NormalTypeError, but I'll leave it in for now, it might come in handy.
    */
   class NormalTypeErrorFromCyclicReference(underlyingTree: Tree, errMsg: String)
```
adriaanm: I'm always weary of leaving unused stuff in, because some future dev may misunderstand its purpose and wreak havoc with it :-)

lrytz: OK, I'm happy to remove it.

test/files/run/t7187.scala
```diff
@@ -0,0 +1,6 @@
+object Test extends App {
+  def foo(): () => String = () => "hi there"
+  val f: () => Any = foo
```
lrytz, line 3: Note that in 2.13 (`-Xsource:2.13`) this compiles without a warning, `foo` is applied to the empty parameter list.

In 2.12, eta-expansion comes in before, and it compiles with a deprecation warning:
```
➜  sandbox git:(syntacticNamedArgs-213) scalac Test.scala -deprecation
Test.scala:3: warning: Eta-expansion of zero-argument method values is deprecated. Did you intend to write Test.this.foo()?
  val f: () => Any = foo
                     ^
one warning found
```

So the program compiles in both versions, but differently.

adriaanm (approved):

adriaanm:
Sorry for the delay, Lukas.

lrytz:
This merge broke the build, as noted in https://github.com/scala/scala/pull/6143#issuecomment-348407519

These are most likely the relevant commits that went into 2.13.x after this PRs parent:
  - https://github.com/scala/scala/pull/5983
  - https://github.com/scala/scala/pull/6069

PR #6093, opened 2017-09-22
Intrinsfy raw and s String interpolators

retronym:
Boxing and varargs incur a significant penalty for string
interpolation, as explored in:

  https://medium.com/@dkomanov/scala-string-interpolation-performance-21dc85e83afd

This pull request takes an initial, big step back towards decent performance
by translating calls to these interpolators to string concatenation, which
itself is already translated to string builder appends in the JVM backend.

I have also tried to be a little smarter with the way we size the string
builder.

Longer term, we should build atop of JEP-280 to extract even better
performance with less code for us to maintain. But that's not available
until Java 9.


lrytz:
```
JavaUniverseForce.scala must be updated.
===========================================================
 DIFF:
===========================================================
--- a
+++ b
@@ -321,2 +321,3 @@
     definitions.StringContextClass
+    definitions.StringContextModule
     definitions.QuasiquoteClass
```

Also the following, which I cannot pin down by just looking at the jenkins log:

```
[error] - junit/test
[error]   - junit/test:compileIncremental failed: scala.StringContext$InvalidEscapeException: invalid escape '\x' not one of [\b, \t, \n, \f, \r, \\, \", \'] at index 0 in "\x". Use \\ for literal \.
```

retronym:
Fixed up the test failure and refactored the implementation some.

I'm happy with the 2.12.5 milestone for this one, seems too risky to rush in to 2.12.4.

src/reflect/scala/reflect/internal/StdNames.scala
```diff
@@ -689,6 +689,8 @@ trait StdNames {
     val ex: NameType                   = "ex"
     val experimental: NameType         = "experimental"
     val f: NameType                    = "f"
+    val s: NameType                    = "s"
+    val raw_ : NameType                = "raw"
```
dwijnand: alphabetical order?

plokhotnyuk:
Next optimization step whould be reusing of StringBuilder instance limited by some resonable size:
```scala
package utils.macros
/*
   Super fast interpolator for string concat.
   Most of code borrowed from: https://github.com/dkomanov/scala-string-format
 */
object StringConcat {
  private val pool = new ThreadLocal[java.lang.StringBuilder] {
    override def initialValue(): java.lang.StringBuilder = new java.lang.StringBuilder(65536)

    override def get(): java.lang.StringBuilder = {
      val sb = super.get()
      if (sb.capacity > 65536) {
        val newSb = initialValue()
        set(newSb)
        newSb
      } else {
        sb.setLength(0)
        sb
      }
    }
  }

  def stringBuilder(): java.lang.StringBuilder = pool.get()

  implicit class SuperFastInterpolator(val sc: StringContext) extends AnyVal {
    def sfi(args: Any*): String = macro sfiImpl
  }

  def sfiImpl(c: whitebox.Context)(args: c.Expr[Any]*): c.Expr[String] = {
    import c.universe._

    val constantParts = extractConstantPartsAndTreatEscapes(c)
    if (args.isEmpty) c.Expr(Literal(Constant(constantParts.mkString(""))))
    else {
      val (valDeclarations, arguments: Seq[c.universe.Tree]) = args.zipWithIndex.map {
        case (e, index) =>
          val name = TermName("__local" + index)
          e.actualType match {
            case tt if tt.typeSymbol.isClass && tt.typeSymbol.asClass.isPrimitive  =>
              (Nil, e.tree)
            case tt if tt <:< typeOf[CharSequence] =>
              val expr = q"""
                val $name = {
                  val tmp = $e
                  if (tmp ne null) tmp else "null"
                }"""
              (List(expr), Ident(name))
            case _ =>
              val expr = q"""
                val $name = {
                  val tmp = $e
                  if (tmp != null) tmp.toString else "null"
                }"""
              (List(expr), Ident(name))
          }
      }.unzip
      val allParts = getAllPartsForAppend(c)(constantParts, arguments)
      // code generation
      val stringBuilderWithAppends = newStringBuilderWithAppends(c)(q"utils.macros.StringConcat.stringBuilder()", allParts.reverse)
      val stats = valDeclarations.flatten.toList ++ (q"$stringBuilderWithAppends.toString" :: Nil)
      c.Expr(c.typecheck(q"..$stats"))
    }
  }

  private def getAllPartsForAppend(c: whitebox.Context)(rawParts: Seq[String], arguments: Seq[c.universe.Tree]) = {
    import c.universe._

    def nilOrConst(s: String) = if (s.isEmpty) Nil else List(Literal(Constant(s)))

    rawParts.zipAll(arguments, null, null).flatMap {
      case (raw, null) => nilOrConst(raw)
      case (raw, arg) if raw != null => nilOrConst(raw) :+ arg
    }.toList
  }

  private def extractConstantPartsAndTreatEscapes(c: whitebox.Context): List[String] = {
    import c.universe._

    val rawParts = c.prefix.tree match {
      case Apply(_, List(Apply(_, list))) => list
    }
    rawParts.map { case Literal(Constant(rawPart: String)) => StringContext.treatEscapes(rawPart) }
  }

  private def newStringBuilderWithAppends(c: whitebox.Context)(newStringBuilderExpr: c.universe.Tree, expressions: List[c.universe.Tree]): c.universe.Tree = {
    import c.universe._

    if (expressions.isEmpty) newStringBuilderExpr
    else Apply(Select(newStringBuilderWithAppends(c)(newStringBuilderExpr, expressions.tail), TermName("append")), List(expressions.head))
  }
}
```
Or even better: a stack of them to allow recursive fast string interpolation.

retronym:
@lrytz Are you happy with the final state of this to target 2.12.5?

src/reflect/scala/reflect/internal/Definitions.scala
```diff
@@ -1451,6 +1452,10 @@ trait Definitions extends api.StandardDefinitions {
       def isStringAddition(sym: Symbol) = sym == String_+ || sym == StringAdd_+
 
       lazy val StringContext_f = getMemberMethod(StringContextClass, nme.f)
+      lazy val StringContext_s = getMemberMethod(StringContextClass, nme.s)
+      lazy val StringContext_raw = getMemberMethod(StringContextClass, nme.raw_)
+      lazy val StringContext_constructor = getMemberMethod(StringContextClass, nme.CONSTRUCTOR)
```
lrytz: this is unused

src/reflect/scala/reflect/internal/StdNames.scala
lrytz: no strong opinion, but it seems the others here are alphabetical indeed.

src/reflect/scala/reflect/internal/TreeGen.scala
```diff
@@ -322,6 +322,7 @@ abstract class TreeGen {
 
   def mkLiteralUnit: Literal = Literal(Constant(()))
   def mkUnitBlock(expr: Tree): Block = Block(List(expr), mkLiteralUnit)
+  def mkLiteralString(s: String): Literal = Literal(Constant(s)).setType(StringTpe)
```
lrytz: unused

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1510,6 +1529,39 @@ abstract class RefChecks extends Transform {
         currentApplication = tree
         tree
     }
+
+    private object StringContextIntrinsic {
+      def unapply(t: Apply): Option[(List[Tree], List[Tree])] = {
```
lrytz, line 1534: Could return an `Option[(List[Literal], List[Tree])]` to avoid the casts at the use site, though it would require another `map` in the `raw` case. Up to you, both ways is fine.

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -1499,7 +1499,26 @@ abstract class RefChecks extends Transform {
             isIrrefutable(pat1, tpt.tpe) && (qual.tpe <:< tree.tpe)) =>
 
           transform(qual)
-
+      case StringContextIntrinsic(treated, args) =>
+        var tree: Tree = treated.head
```
lrytz: maybe call it `result` to avoid shadowing (and confusion)

lrytz (changes requested):
Looks good, only a few cosmetic cleanups.

lrytz (approved):
pushed my review suggestions

retronym:
> pushed my review suggestions

Thanks!

SethTisue:
questions: this is intrinsics and not a macro in order to get it into 2.12.x, because bincompat...? and the motivation for getting it into 2.12.x instead of just waiting for 2.13.x is compiler performance...?

retronym:
Yes, bin compat rules out the macro in 2.12. Source compat, taken strictly,
also makes it had to turn a regular method into a macro in a major release,
but we still plan to do that in 2.13:
https://github.com/scala/scala-dev/issues/475

The compiler already avoided this iniefficiency in really hot code paths,
we didn’t expect nor measure a discernible important time improvement after
this change. But we expected expect application code to benefit (example
where it would have helped: https://github.com/sbt/util/pull/152), hence
the desire to deliver earlier than 2.13

SethTisue:
Nice, forgot about 475! Thx /cc @hrhino

xuwei-k:
https://github.com/scala/bug/issues/10870 😢

eed3si9n:
https://github.com/scala/bug/issues/11196

PR #6095, opened 2017-09-23
Reverse order of repeated tags

Philippus:
closes https://github.com/scala/bug/issues/10325
alternatively the *RegEx cases of the parse0-method could be changed from prepending to appending the tag bodies


Philippus:
closed in favour of https://github.com/scala/scala/pull/6111 which is a superset of this PR.

PR #6096, opened 2017-09-24
No warn unused import on macro expansion

som-snytt:
Creating an import context registers the import for the unused
warning. However, if the use site is already typechecked,
then on re-typechecking a macro expansion, the use won't be
registered.

As a quick fix, if there are open macros, don't register an
import for unused checking.

Fixes scala/bug#10270


som-snytt:
@adriaanm Hoping this easy one makes it into 2.12.4 because it's super annoying and fix is low risk.

adriaanm:
Sure, check file update needed?

som-snytt:
Updated the check file. There was a previous fix not to crash when linting a macro expansion (due to lack of position), but this fix is to skip that linting altogether.

adriaanm (approved):

PR #6097, opened 2017-09-25
Fix runtime reflection of empty package members under Java 9.

retronym:
We used to rely on `cls.getPackage == null` for `cls` defined in the
empty package. Under Java 9, we actually get the empty package back
from that call.

This commit ensures we use the one true empty package symbol on
either Java 8 or 9.

Fixes scala/scala-dev#304


retronym:
@adriaanm Optimistically assigning to 2.12.4 as the fix is straight forward and it removes an impediment to Java 9 experimentation.

Unfortunately, without a backport, it won't help with the original bug report, which involves SBT hitting this problem in runtime reflection in Scala 2.10.

adriaanm (approved):

SethTisue:
ticket on having sbt move to 2.12.4 to get the fix: https://github.com/sbt/sbt/issues/3587

PR #6099, opened 2017-09-25
No need to dealias for cast tree

adriaanm:
Also drop asserts that prevent casting to/from MethodType.
(The assert on `pt eq pt.normalize` is no longer true,
because it need not be.)

(Follow up for #5722 aka uncurry-ohne-dealias and #6091.)

With apologies for the future merge conflict.


SethTisue:
rebase for new Jenkins run...?

SethTisue:
ping @adriaanm

retronym (approved):

PR #6100, opened 2017-09-26
Make Lazy* classes serializable

lrytz:
The instances may be captured by closures, which should
be serializable.

Fixes scala/bug#10522


lrytz:
Used the wrong branch in my fork

PR #4167, opened 2014-11-27
SI-8911  scala.collection.convert.Wrappers$MapWrapper not serializable

Ichoran:
Added `with Serializable` to `MapWrapper` and `SetWrapper`.

Test verifies that serialziation works in the simplest case.

PR #6101, opened 2017-09-26
Make Lazy* classes serializable

lrytz:
The instances may be captured by closures, which should
be serializable.

Fixes scala/bug#10522


lrytz:
```
[info] scala-library: found 10 potential binary incompatibilities while checking against org.scala-lang:scala-library:2.12.0  (filtered 20)
[error]  * the type hierarchy of class scala.runtime.LazyRef is different in other version. Missing types {scala.Serializable,java.io.Serializable}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("scala.runtime.LazyRef")
...
```

Hmm.. should we ignore them?

adriaanm:
I've been digging around for our prior experience with making stuff serializable. See also https://github.com/scala/scala/pull/4167, https://github.com/scala/scala-dev/issues/237

Is there any way the difference in extending Serializable is visible to the casual user, except when serializing?

retronym:
A client compiling against 2.12.4 could write:

```scala
def foo(s: Serializable) = ()
def bar(l: LazyRef) = foo(l)
```

But this would fail when using the 2.12.3 library. That's okay IMO, as these classes aren't intended for direct usage.

src/library/scala/runtime/LazyRef.scala
```diff
@@ -10,7 +10,7 @@ package scala.runtime
 
 /** Classes used as holders for lazy vals defined in methods. */
 
-class LazyRef[T] {
+class LazyRef[T] extends Serializable {
```
retronym: Let's add the `@SerialVersionUID` annotations on these in case we need to change the representation while having the ability to be serialization backwards compatible.

lrytz: ah yes, thanks

adriaanm:
Yeah, I wouldn't expect the casual user to venture into `scala.runtime`

adriaanm (approved):

PR #6104, opened 2017-09-28
Return in PartialFun relative to right enclosing method

adriaanm:
Where the right enclosing method is the lexically enclosing one,
not the one the body ends up being type checked in (`applyOrElse`).
Since the current owner impacts more than just type checking
`Return` trees, we adjust the currently logically enclosing method
as we do for lazy getters (the owner also governs e.g., skolemization).

Fixes https://github.com/scala/bug/issues/10291.

I'm proposing to sneak this one into 2.12.4 since it's pretty embarrassing and the fix is pretty safe, I believe.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -214,7 +214,12 @@ trait Contexts { self: Analyzer =>
     /** Is this context in all modes in the given `mask`? */
     def apply(mask: ContextMode): Boolean = contextMode.inAll(mask)
 
-    /** The next outer context whose tree is a method */
+    /** The next (logical) outer context whose tree is a method.
+      *
+      * NOTE: this is the "logical" enclosing method, which may not be the actual enclosing method when we
+      * synthesize a nested method, such as for lazy val getters (scala/bug#8245) or the methods that
+      * implement a PartialFunction literal (scala/bug#10291).
+      */
     var enclMethod: Context = _
 
     /** Variance relative to enclosing class */
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -2649,6 +2649,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
           CaseDef(Bind(nme.DEFAULT_CASE, Ident(nme.WILDCARD)), body)
         }
 
+      def synthMethodTyper(methodSym: MethodSymbol) = {
+        val ctx = context.makeNewScope(context.tree, methodSym)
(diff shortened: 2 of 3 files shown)


src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -484,7 +484,10 @@ trait Contexts { self: Analyzer =>
       c(ConstructorSuffix) = !isTemplateOrPackage && c(ConstructorSuffix)
 
       // scala/bug#8245 `isLazy` need to skip lazy getters to ensure `return` binds to the right place
-      c.enclMethod         = if (isDefDef && !owner.isLazy) c else enclMethod
+      // scala/bug#10291 ditto for the applyOrElse synthetic method in a PartialFunction
+      // (a PF could also contain an `apply` instead under the @cps plugin, but this is no longer supported)
+      def skipAsEnclosingMethod = owner.isLazy || (owner.name == nme.applyOrElse && owner.owner.isAnonymousFunction)
+      c.enclMethod         = if (isDefDef && !skipAsEnclosingMethod) c else enclMethod
```
retronym: Could this inadvertently catch user-defined `applyOrElse`?

Is it possible to drive this instead with:

```diff
diff --git a/src/compiler/scala/tools/nsc/typechecker/Typers.scala b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
index 57e46d95d7..32cd15347c 100644
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -2667,7 +2667,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         val paramSyms = List(x, default)
         methodSym setInfo genPolyType(List(A1, B1), MethodType(paramSyms, B1.tpe))

-        val methodBodyTyper = newTyper(context.makeNewScope(context.tree, methodSym))
+        val nestedContext = context.makeNewScope(context.tree, methodSym)
+        nestedContext.enclMethod = context.enclMethod
+        val methodBodyTyper = newTyper(nestedContext)
         if (!paramSynthetic) methodBodyTyper.context.scope enter x

         // First, type without the default case; only the cases provided
```

adriaanm:
The second part checks the synthetic flag

adriaanm:
That, said it does look cleaner to manipulate the context here.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
adriaanm: good spot! nice opportunity for cleanup -- now, if someone returns from the guard in isDefinedAt, it'll also be the right kind of return :-)

retronym (approved):

lrytz:
LGTM, maybe add a comment on `var enclMethod`, as this might trip up other future users of that variable.

adriaanm:
Added comment and `push -f`'ed

lrytz (approved):

PR #6105, opened 2017-09-29
Avoid repr call in ArrayOps

cryptom-dev:
This PR is part of fixes for issue: rorygraves#23. It means to improve the performance of ArrayOps @mkeskells spotted.

This PR replaced the ofDim call, which involved Manifest resolving, for :+ and +: method in ArrayOps with constructor of scala Array.
According to jprofiler, the call percentage reduced from 14.4% (ofDim) to 3.5% (constructor of Array)

### DIFF
--- a/src/library/scala/collection/mutable/ArrayOps.scala
+++ b/src/library/scala/collection/mutable/ArrayOps.scala
@@ -63,7 +63,7 @@ sealed trait ArrayOps[T] extends Any with ArrayLike[T, Array[T]] with CustomPara
 
   def :+[B >: T: ClassTag](elem: B): Array[B] = {
     val currentLength = repr.length
-    val result = Array.ofDim[B](currentLength + 1)
+    val result = new Array[B](currentLength + 1)
     Array.copy(repr, 0, result, 0, currentLength)
     result(currentLength) = elem
     result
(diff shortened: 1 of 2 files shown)


ackratos:
This PR is for issue: https://github.com/rorygraves/scalac_perf/issues/23

cryptom-dev:
jprofiler comparation (ofDim call in first image has been removed):

Baseline call tree:
arrayopsofdim_before

Improved call tree:
arrayopsofdim_after

mkeskells:
I think that we can go further in this pr

I think that to common case is non widening so for some operations we can use Arrays.copyof via delegation.

The jvm is sensitive to the gap between array creation and fill as  to meet his spec the content is defined to be default values. I think that the scala calls will not be recognised as a safe combination so this will be an allocate then fill then copy then add sequence when in general the fill can be removed. IM me directly is this doesn't make sense

This will require additional test cases and we should include specific widening and non widening cases and some widening will move from primitive types to wrappers

All of this is subject to it performing better though :-)

mkeskells:
Can you share the result of the JMH benchmark here

lrytz (approved):
Even though this can be refined, let's get the current improvement in. In 2.13 ArrayOps will really be only extension methods (https://github.com/scala/collection-strawman/pull/479)

PR #6106, opened 2017-09-30
Revert "Add infer Product with Serializable linter flag"

adriaanm:
Reverts scala/scala#5990

Fix scala/bug#10535


adriaanm:
Let’s come back to this after 2.12.4

PR #6107, opened 2017-10-02
[SUPERSEDED] No illegal stars in typed select

### LINKED ISSUE #8923: imported function forgets varargness
Why?

```scala
object Foo { def foo[A]: (A*) => A = as => as.head }
```

ok:
```scala
scala> Foo.foo(1,2,3)
res1: Int = 1
```

not ok:
```scala
scala> import Foo.foo; foo(1,2,3)
<console>:10: error: too many arguments for method apply: (v1: Seq[A])A in trait Function1
        foo(1,2,3)
           ^
```

som-snytt:
`typedIdent` converts repeated params as required,
and `typedSelect` should do the same.

There is a more generic issue in scala/bug#6929

Fixes scala/bug#8923

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -962,7 +962,7 @@ self =>
      *  called: either in a pattern context or not.  Formerly, this was
      *  threaded through numerous methods as boolean isPattern.
      */
-    trait PatternContextSensitive {
+    sealed trait PatternContextSensitive {
       /** {{{
        *  ArgType       ::=  Type
        *  }}}
@@ -1921,16 +1921,12 @@ self =>
     /** Methods which implicitly propagate whether the initial call took
      *  place in a context where sequences are allowed.  Formerly, this
      *  was threaded through methods as boolean seqOK.
+     *  @param isSequenceOK is a sequence pattern _* allowed?
+     *  @param isXML are we in an XML pattern?
      */
-    trait SeqContextSensitive extends PatternContextSensitive {
-      // is a sequence pattern _* allowed?
-      def isSequenceOK: Boolean
-
-      // are we in an XML pattern?
-      def isXML: Boolean = false
-
-      def functionArgType(): Tree = argType()
-      def argType(): Tree = {
+    final class SeqContextSensitive(val isSequenceOK: Boolean, val isXML: Boolean) extends PatternContextSensitive {
+      final def functionArgType(): Tree = argType()
+      final def argType(): Tree = {
         val start = in.offset
         in.token match {
           case USCORE =>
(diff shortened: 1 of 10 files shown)


som-snytt:
The first cut failed because accessors were star-typed, or star-crossed.
```
scala> class C(val xs: Int*) { def f = xs }
defined class C

scala> val c = new C(1,2,3)
c: C = C@7ddd84b5

scala> c.f //print
   $line4.$read.$iw.$iw.c.f // : Seq[Int]

scala> c.xs //print
   $line4.$read.$iw.$iw.c.xs // : Int*


```
I tested bootstrap build locally thanks to @lrytz instructions on another PR, maybe he can also say how far off base this tweak is. I didn't follow type completers in namers far enough to get a sense of it.

som-snytt:
The tweak probably isn't tweaky enough, but one test failure is because it relies on previous star in function:
```
val testAllS: (Int, =>Int, Int*) => Int = testAll _
test("all s", 8, testAllS(1, 5, 78, 89))

```
It also re-broke varargs in ctors. I guess that relies on the accessors having star type.

retronym:
This area fills me with dread, but I've been here before so I'll assign myself to this ticket. Let's move the PR to 2.13.x though.

som-snytt:
It makes me see stars.

som-snytt:
@retronym your change for looking at param aliases after typechecking obviates one hack from the previous version of this PR.

SethTisue:
ping @retronym

som-snytt:
I'm so sorry, @retronym .

lrytz:
hmm i don't see it, which part of the diff fixes the bug?

som-snytt:
Was there a bug? I see I took a couple of days this year to refresh my memory. What year is this again? I'll take another look. Oh my, it's been fast-tracked. I think 2.13.1 means after everyone is back from vacation and before mid-terms.

som-snytt:
I see the fix allows star in the signature, but requires `Seq` in the application.  Essentially, instead of disallowing it, the star becomes a kind of alternative syntax for `Seq`. The spec language would say that the star type means `Seq` everywhere _except_ in a method application.

Probably that made more sense back in the days of `-Yeta-expand-keeps-star`.  But since that is no longer an option (literally), it should be disallowed except for method parameters.

I kind of like the current fix, but it's rather eccentric.

som-snytt:
Added a commit that disallows `T*` in `(T*) => U`.

Maybe that should be deprecated until `-Xsource:2.14`?

The previous commit still does something, compare to [last year](https://github.com/scala/scala/pull/6107#issuecomment-333462229):

```
$ skala
Welcome to Scala 2.13.0 (OpenJDK 64-Bit Server VM 1.8.0_181)

scala> class C(val xs: Int*) { def f = xs }
defined class C

scala> val c = new C(1,2,3)
c: C = C@7cea0110

scala> c.f //print
   c.f // : Seq[Int]

scala> c.xs //print
   c.xs // : Seq[Int]

```
The original report, that `import X.f ; f(1,2,3)` fails but `X.f(1,2,3)` succeeds, is fixed by stripping star in `typedSelect`. But with the parser change, the star type is not expressible, so you can't import something of star type. The print-tab in REPL might see it if I reverted that change; I haven't tried that yet.

It's probably the aliasing bit that retronym found dreadful. I don't see a new test, and I don't remember which tests fail without that change, which I'll have to puzzle over a minute.

You may have noticed 38 Down in the Friday NY Times crossword, "rotary phone", @retronym so-called. I wonder if "compiler engineer" could be considered a retronym. Then is retronym a compiler engineer or the other way around?

SethTisue:
closing for inactivity, can be revived & reopened whenever

som-snytt:
Since retronym is dreadfully hung up, I moved the other commits to separate PRs.

SethTisue:
new PR: #7399

PR #6108, opened 2017-10-03
Move compilation daemon portfile to homedir

adriaanm:
Store the compilation daemon's administrativia (port file, redirection)
under `~/.scalac/`, instead of the less standard
`/tmp/scala-devel/${USER:shared}/scalac-compile-server-port`.

On creation, remove group- and other-permissions from these
private files, ditto for the repl's history file.

On Java 6 on Windows, opt in to compilation daemon using `-nc:false`.

### DIFF
--- a/src/compiler/scala/tools/nsc/CompileSocket.scala
+++ b/src/compiler/scala/tools/nsc/CompileSocket.scala
@@ -5,12 +5,16 @@
 
 package scala.tools.nsc
 
-import java.io.{ FileNotFoundException, PrintWriter, FileOutputStream }
+import java.math.BigInteger
 import java.security.SecureRandom
-import io.{ File, Path, Directory, Socket }
-import scala.tools.util.CompileOutputCommon
+
+import scala.reflect.internal.util.OwnerOnlyChmod
 import scala.reflect.internal.util.StringOps.splitWhere
 import scala.sys.process._
+import scala.tools.nsc.Properties.scalacDir
+import scala.tools.nsc.io.{File, Socket}
+import scala.tools.util.CompileOutputCommon
+import scala.util.control.NonFatal
 
 trait HasCompileSocket {
   def compileSocket: CompileSocket
@@ -46,14 +50,11 @@ trait HasCompileSocket {
 class CompileSocket extends CompileOutputCommon {
   protected lazy val compileClient: StandardCompileClient = CompileClient
   def verbose = compileClient.verbose
-  
+  def verbose_=(v: Boolean) = compileClient.verbose = v
+
   /* Fixes the port where to start the server, 0 yields some free port */
   var fixPort = 0
 
-  /** The prefix of the port identification file, which is followed
-   *  by the port number.
-   */
-  protected lazy val dirName = "scalac-compile-server-port"
   protected def cmdName = Properties.scalaCmd
 
   /** The vm part of the command to start a new scala compile server */
@@ -67,22 +68,10 @@ class CompileSocket extends CompileOutputCommon {
 
   /** The class name of the scala compile server */
   protected val serverClass     = "scala.tools.nsc.CompileServer"
-  protected def serverClassArgs = (if (verbose) List("-v") else Nil) ::: (if (fixPort > 0) List("-p", fixPort.toString) else Nil) 
-
-  /** A temporary directory to use */
-  val tmpDir = {
-    val udir  = Option(Properties.userName) getOrElse "shared"
-    val f     = (Path(Properties.tmpDir) / ("scala-devel" + udir)).createDirectory()
-
-    if (f.isDirectory && f.canWrite) {
-      info("[Temp directory: " + f + "]")
-      f
-    }
-    else fatal("Could not find a directory for temporary files")
-  }
+  protected def serverClassArgs = (if (verbose) List("-v") else Nil) ::: (if (fixPort > 0) List("-p", fixPort.toString) else Nil)
 
   /* A directory holding port identification files */
-  val portsDir = (tmpDir / dirName).createDirectory()
+  private lazy val portsDir = mkDaemonDir("fsc_port")
 
   /** The command which starts the compile server, given vm arguments.
     *
@@ -104,7 +93,7 @@ class CompileSocket extends CompileOutputCommon {
   }
 
   /** The port identification file */
-  def portFile(port: Int) = portsDir / File(port.toString)
+  def portFile(port: Int): File = portsDir / File(port.toString)
 
   /** Poll for a server port number; return -1 if none exists yet */
   private def pollPort(): Int = if (fixPort > 0) {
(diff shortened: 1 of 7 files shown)


src/reflect/scala/reflect/internal/util/OwnerOnlyChmod.scala
```diff
@@ -0,0 +1,110 @@
...
+ * Copyright 2017 LAMP/EPFL
+ * @author  Martin Odersky
+ */
+package scala.reflect.internal.util
+
+import java.io.{File, FileOutputStream, IOException}
+
+
+trait OwnerOnlyChmod {
+  /** Remove group/other permisisons for `file`, it if exists */
```
SethTisue: typo (permissions)

src/compiler/scala/tools/nsc/GenericRunnerSettings.scala
```diff
@@ -45,5 +45,13 @@ class GenericRunnerSettings(error: String => Unit) extends Settings(error) {
 
   val nc = BooleanSetting(
       "-nc",
-      "do not use the fsc compilation daemon") withAbbreviation "-nocompdaemon"
+      "do not use the fsc compilation daemon") withAbbreviation "-nocompdaemon" withPostSetHook((x: BooleanSetting) => {_useCompDaemon = !x.value })
+
+  private def defaultUseCompdaemon = {
+    // can't reliably lock down permissions on the portfile in this environment => disable by default.
+    !scala.util.Properties.isWin || scala.util.Properties.isJavaAtLeast("7")
```
SethTisue: maybe we should print a warning here rather than silently giving different behavior

src/compiler/scala/tools/nsc/CompileSocket.scala
```diff
@@ -138,19 +127,19 @@ class CompileSocket extends CompileOutputCommon {
...
   }
 
   /** Set the port number to which a scala compile server is connected */
-  def setPort(port: Int) {
+  def setPort(port: Int): Unit = {
     val file    = portFile(port)
-    val secret  = new SecureRandom().nextInt.toString
+    val secretBytes = new Array[Byte](16)
+    new SecureRandom().nextBytes(secretBytes)
+    val secretDigits = new BigInteger(secretBytes).toString().getBytes("UTF-8")
```
SethTisue: these 3 lines of code are quite mystifying. some comments explaining the motivation would be good

src/reflect/scala/reflect/internal/util/OwnerOnlyChmod.scala
```diff
@@ -0,0 +1,110 @@
...
+}
+
+
+/** Adjust permissions with `File.{setReadable, setWritable}` */
+object Java6UnixChmod extends OwnerOnlyChmod {
+
+  def chmod(file: File): Unit = if (file.exists()) {
+    def clearAndSetOwnerOnly(f: (Boolean, Boolean) => Boolean): Unit = {
+      def fail() = throw new IOException("Unable to modify permissions of " + file)
+      // attribute = false, ownerOwnly = false
```
SethTisue: typo here and 2 lines down: ownly -> only

src/reflect/scala/reflect/internal/util/OwnerOnlyChmod.scala
```diff
@@ -0,0 +1,110 @@
...
+    val userPrinciple_class = Class.forName("java.nio.file.attribute.UserPrincipal")
+    val setPrincipal = aclEntryBuilder_class.getMethod("setPrincipal", userPrinciple_class)
+    val setPermissions = aclEntryBuilder_class.getMethod("setPermissions", Class.forName("[Ljava.nio.file.attribute.AclEntryPermission;"))
+    val aclEntryType_class = Class.forName("java.nio.file.attribute.AclEntryType")
+    val setType = aclEntryBuilder_class.getMethod("setType", aclEntryType_class)
+    val aclEntryPermission_class = Class.forName("java.nio.file.attribute.AclEntryPermission")
+    val aclEntryPermissionValues = aclEntryPermission_class.getDeclaredMethod("values")
+    val aclEntryType_ALLOW = aclEntryType_class.getDeclaredField("ALLOW")
+  }
+  private val reflectors = try { new Reflectors } catch { case ex: Throwable => null }
```
SethTisue: should we log the swallowed exception?

src/repl-jline/scala/tools/nsc/interpreter/jline/FileBackedHistory.scala
```diff
@@ -8,15 +8,37 @@ package scala.tools.nsc.interpreter.jline
...
 
 /** TODO: file locking.
   */
 trait FileBackedHistory extends JLineHistory with PersistentHistory {
   def maxSize: Int
 
-  protected lazy val historyFile: File = FileBackedHistory.defaultFile
+  // For a history file in the standard location, always try to restrict permission,
+  // creating an empty file if none exists.
+  // For a user-specified location, only lock down permissions on if we're the ones
```
SethTisue: missing word in this sentence

src/repl-jline/scala/tools/nsc/interpreter/jline/FileBackedHistory.scala
```diff
@@ -8,15 +8,37 @@ package scala.tools.nsc.interpreter.jline
...
+      case Some(p) => if (!p.exists) secure(p) else p
+      case None => secure(Path(userHome) / FileBackedHistory.defaultFileName)
+    }
+  }
+
+  private def secure(p: Path): Path = {
+    try OwnerOnlyChmod().chmodOrCreateEmpty(p.jfile)
+    catch { case NonFatal(e) =>
+      if (interpreter.isReplDebug) e.printStackTrace()
+      interpreter.replinfo(s"Warning: history file ${p}'s permissions could not be restricted to owner-only.")
```
SethTisue, line 36: is warning and continuing considered sufficiently secure?

adriaanm:
@retronym, ready for final review

retronym (approved):

adriaanm:
@szeiger could you give this a go on your surface and any other windows desktops at your disposal ?

src/reflect/scala/reflect/internal/util/OwnerOnlyChmod.scala
```diff
@@ -0,0 +1,107 @@
...
+    val fos = new FileOutputStream(file)
+    fos.close()
+    chmod(file)
+    val fos2 = new FileOutputStream(file)
+    try {
+      fos2.write(contents)
+    } finally {
+      fos2.close()
+    }
+  }
```
setharnold, line 26: Note that creating a file _then_ changing the permissions on the file allows a race condition where an attacker on the system can open the file between the file create and the permissions change. It is far safer to use open(2) to set the permissions restrictive while creating the file.

Furthermore if this code is ever run on a directory with shared writing permissions, it's possible for an attacker to unlink(file), set a symlink in place, and have the process change modes of any target that the process has privileges to modify. fchmod(2) is the safe way to change permissions on a file that is already open, though that won't fix the above race condition. So consider this just an aside.

Thanks

adriaanm: Thank you. Could you suggest how to do this on Java 6? I agree the race condition is a security hole, but couldn't figure out how to do this atomically with the API we have available. I think our NIO-based implementation on Java 8 is closer to what you're suggesting: https://github.com/scala/scala/pull/6120/files#diff-3578ac76088d22b5d9912e984dee3affR46.

retronym: Thank you for the review.

We implemented things that way (or at least, tried to) in the 2.12 branch where we could link against the Java 8 NIO API. Scala 2.11 still supports Java 6.

https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/util/OwnerOnlyChmod.scala

retronym: > adriaanm a minute ago
> retronym 29 seconds ago  

Speaking of race conditions :)

setharnold: The #diff-3578ac7.. link above looks like it may have the same race condition.

I don't know Java particularly well any longer, so it's hard to suggest a replacement. Two possibilities come to mind:

- set the umask to 0077 (octal), create the file, set the modes, and then reset umask to whatever the user wanted
- chmod 0700 (octal) the containing directory, create the file, set the file modes, then reset the directory to whatever permissions it should have.

I hope this helps. Thanks.

huntc: How about: create the file with some temporary name/location, do what you need to do and then move it. file moves are atomic.

PR #6112, opened 2017-10-05
lengthen timeout on some Future tests

SethTisue:
these have repeatedly (though intermittently) failed on Jenkins,
e.g. https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-windows/795/

not *that* often, but just often enough to be annoying

perhaps longer timeouts will make it happen less often


viktorklang (approved):
LGTM

PR #6115, opened 2017-10-06
[WIP] October grab bag of compiler optimizations [ci: last-only]

retronym:
Pushed here to facilitate benchmarking.

Locally I measured 6-7% aggregate improvement on the scala and scalap
corpi.


retronym:
[Benchmarking](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=097c047d3e%7Cdfb4975434%7C8c2bf41a56%7C120a721a32%7Cd893e7ca56%7Ceb1457884f%7Cdc4dbe32fe%7Cd4b090d6b2%7Cb5d54c6419%7Cf4bb266f7d%7C25009543ee%7C8ac07d37d0%7Cd3f346c089%7C76513a76bf%7C7c8cb7bb89%7Cb22b2a1f7b%7C792e03fde3%7C54a7449134%7Cc153e96543%7C49297ebf0e%7Cc5e6a18836%7Cfbf26bafc5%7Ca479afdd66%7Cde408c2292%7Cec4460712a%7C92d4261fb6%7Ce1c70ee798%7C3a9739bd25%7C0daa171cb0%7C8638155824%7Ced08c77d55%7C63af08da15%7Ca589051477%7C65a4df0214%7C0092783af5%7Cb6ee3c923d%7C66ad50fd1d%7C0ef9bcb74a%7Ceb6f50a0b8%7Ce695a3b1cd&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1503634037049&to=1507275610843)

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -56,22 +68,25 @@ abstract class BTypes {
...
-      case ClassBType(internalName) => builder.append('L').append(internalName).append(';')
-      case ArrayBType(component)    => builder.append('['); component.buildString(builder)
-      case MethodBType(args, res)   =>
-        builder.append('(')
-        args.foreach(_.buildString(builder))
-        builder.append(')')
-        res.buildString(builder)
+      case _: PrimitiveBType =>
+        this match {
+          case UNIT => builder.append('V')
```
viktorklang, line 73: Did you consider using a HashMap for this lookup?

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
```diff
@@ -324,8 +324,11 @@ abstract class BackendUtils extends PerRunInit {
       bTypesFromClassfile.classBTypeFromParsedClassfile(internalName).info.get.nestedClasses.force
 
     def getClassIfNested(internalName: InternalName): Option[ClassBType] = {
-      val c = bTypesFromClassfile.classBTypeFromParsedClassfile(internalName)
-      if (c.isNestedClass.get) Some(c) else None
+      if (internalName.indexOf('$') < 0) None
```
viktorklang, line 327: Do $-signs occur primarily close to start or close to end, in the latter case you might want to check `lastIndexOf`.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
retronym: The common case is `! PrimitiveBType`, so the main thing is to get to that case quickly. A reasonable and performant alternative would be to make it a virtual method.

src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala
retronym: Depends if the inner or outer class has a longer name. The common case is probably a name with no `$`, so we have to look at the whole name most of the time.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: could make the descriptor a field of the PrimativeType

i.e. case p: PrimitiveBType => builder.append(p.descriptorChar)

Is that what you meant?

src/compiler/scala/tools/nsc/transform/patmat/Logic.scala
```diff
@@ -174,8 +174,8 @@ trait Logic extends Debugging  {
     def simplify(f: Prop): Prop = {
 
       // limit size to avoid blow up
-      def hasImpureAtom(ops: Seq[Prop]): Boolean = ops.size < 10 &&
-        ops.combinations(2).exists {
+      def hasImpureAtom(ops: Traversable[Prop]): Boolean =
+        ops.toSeq.combinations(2).exists {
```
mkeskells, line 178: this looks tunable
It looks to be O n squared, and will generate lots of local garbage

could you scan for all Not(_) and if you find any scan for matching content?
Or something similar to avoid all of the list generation and unapplySeq

src/reflect/scala/reflect/internal/AnnotationInfos.scala
```diff
@@ -58,7 +58,7 @@ trait AnnotationInfos extends api.Annotations { self: SymbolTable =>
     @tailrec private
     def dropOtherAnnotations(anns: List[AnnotationInfo], cls: Symbol): List[AnnotationInfo] = anns match {
```
mkeskells, line 59: isnt this just a while loop, that can avoid the unapplySeq and pattern match?

src/compiler/scala/tools/nsc/transform/patmat/Logic.scala
retronym: The garbage problem could be fixed with a custom iterator:

```
class PairsIterator[A](as: Seq[A]) {
  private val asArray = as.toArray
  var indexA, indexB: Int = 0
  def hasNext: Boolean = ...
  def advance(): Unit = ...
  def _1: A = asArray(indexA)
  def _2: B = asArray(indexA)
}
```

src/reflect/scala/reflect/internal/AnnotationInfos.scala
retronym: There is no `unapplySeq` call for case class constructor patterns, and tailrec emits a loop for this:

```
scala> class Tail { final def count[A](as: List[A], z: Int = 0): Int = as match { case a :: bs => count(bs, z + 1); case _ => z }}
defined class Tail

scala> :javap -c Tail#count
  public final <A> int count(scala.collection.immutable.List<A>, int);
    Code:
       0: aload_1
       1: astore        5
       3: aload         5
       5: instanceof    #19                 // class scala/collection/immutable/$colon$colon
       8: ifeq          35
      11: aload         5
      13: checkcast     #19                 // class scala/collection/immutable/$colon$colon
      16: astore        6
      18: aload         6
      20: invokevirtual #23                 // Method scala/collection/immutable/$colon$colon.tl$access$1:()Lscala/collection/immutable/List;
      23: astore        7
      25: aload         7
      27: iload_2
      28: iconst_1
      29: iadd
      30: istore_2
      31: astore_1
      32: goto          0
      35: goto          38
      38: iload_2
      39: istore        4
      41: goto          44
      44: iload         4
      46: ireturn
```

But a while loop might be just as readable.

src/compiler/scala/tools/nsc/transform/patmat/Logic.scala
retronym: I agree this is an area ripe for tuning. A comparative benchmark with `-Xno-patmat-analysis` shows the overhead of patmat analysis. I think it is worth 2% last I measured.

retronym:
Perf results aren't very usable on Grafana as the commit dates aren't in the same order as the git lineage. Here's the same data in a spreadsheet. https://docs.google.com/spreadsheets/d/1nUDDrpdMVKHtfWQOCWZ76TKbGJvq_vsVTSwbbzdfxts/edit?usp=sharing

som-snytt:
A spreadsheet is not as compelling as grafana? Maybe I'm not a numbers guy. The Las Vegas lunatic is reported to have been a numbers guy.

retronym:
Continuing this over in #6209

PR #6117, opened 2017-10-07
Fix for scala/bug#10540 - AnyRefMap dropped entries with one hash code.

Ichoran:
Changed the internal hash code calculation to not produce zero values.

Fix scala/bug#10540


som-snytt:
I used this one to explain to my daughter what "love of truth" means.

retronym (approved):

retronym:
Backport submitted in #6119

PR #6118, opened 2017-10-08
Fix wrong parenthesis order around creation of mutable.BitSet

YuvalItzchakov:
Exploring the implementation of `mutable.BitSet`, I've noticed that the parenthesis seemed to be misplaced, causing the expression `6 max 1` to be evaluated, instead of what I believe was meant to first calculate the initial bit-shifted size, and then max that with 1 in case we're in the lower 64 bits.

### DIFF
--- a/src/library/scala/collection/mutable/BitSet.scala
+++ b/src/library/scala/collection/mutable/BitSet.scala
@@ -50,7 +50,7 @@ class BitSet(protected final var elems: Array[Long]) extends AbstractSet[Int]
    *
    *  @param initSize    initial size of the bitset.
    */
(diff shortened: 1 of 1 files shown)


som-snytt:
Precedence of letter-name is lowest.

One of many ways to check is print-tab in REPL:

```
scala> def f(initSize: Int) = (initSize + 63) >> 6 max 1  //print
   def f(initSize: scala.Int) = scala.Predef.intWrapper(initSize.+(63).>>(6)).max(1) // : <notype>
```

If this behavior were actually broken, of course, a test would be welcome.

YuvalItzchakov:
Thanks for clarifying @som-snytt.

PR #6120, opened 2017-10-09
Move compilation daemon portfile under `~/.scalac/`

adriaanm:
Store the compilation daemon's administrativia (port file, redirection)
under `~/.scalac/`, instead of the less standard
`/tmp/scala-devel/${USER:shared}/scalac-compile-server-port`.

On creation, remove group- and other-permissions from these
private files, ditto for the repl's history file.

Based on b64ad85

### DIFF
--- a/src/compiler/scala/tools/nsc/CompileServer.scala
+++ b/src/compiler/scala/tools/nsc/CompileServer.scala
@@ -6,11 +6,12 @@
 package scala.tools.nsc
 
 import java.io.PrintStream
-import io.Directory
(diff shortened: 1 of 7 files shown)


retronym (approved):

adriaanm:
Let's make sure to play around with the nightly build on different platforms before we finally tag the release. This should hopefully be the last commit.

adriaanm:
(For the record, this needed a quickfix in c0bcc296dff008c0815b98e4f64452f63810a62f.)

PR #6121, opened 2017-10-10
Merge 2.12.x to 2.13.x [ci:last-only]

retronym:
I've included a larger-than-usual number of merge
commits to give us a better chance at bisecting
functional or performance issues that occur after
this merge (e.g things predicated on -Xsource:2.13).
 for regexes.


retronym:
This merges up to, but excluding #6120, as that needs a merge conflict resolved.

adriaanm:
I've resolved that conflict, which should catches this PR up what I hope to be 2.12.4

PR #6122, opened 2017-10-11
When testing for presence of a Symbol in a Type don't skip aliases.

### LINKED ISSUE #10545: Incorrect implementation of Type containment of Symbol frustrates implicit selection
The following does not compile in Scala,
```scala

class D[T]

class C[F[_]](val i: Int)
object C {
  def apply[F[_]](implicit cf: C[F]): Int = cf.i

  implicit def c0[F[_]]: C[F] = new C[F](0)
  implicit def c1: C[D] = new C[D](1)
}

object Test extends App {
  assert(C[D] == 1) // Works in Dotty ...
}
```
because `c0` and `c1` are considered to be ambiguous. The spec and intuitions suggest that `c1` is more specific than the quantified `c0` and that the former should be selected over the latter. This is the behaviour in Dotty. 

scalac gives the wrong result because in the specificity test the call `isStrictlyMoreSpecific` uses `existentialAbstraction` before comparing the result types of the two methods. The latter _ought_ to take `[F[_]]C[F]` on to `C[F] forSome { type F[_] }` however due to a bug in the implementation of `ContainsCollector` (which backs `Type#contains`) it fails to identify the occurrence of `F` in `C[F]` as a variable requiring existential quantification. Neither `C[D]` nor `C[F]` is then a subtype of the other so neither is preferred over the other, hence ambiguity.

PR with a fix incoming ...

milessabin:
The implementation of `ContainsCollector` overly aggressively normalizes away aliases resulting in symbols corresponding to higher-kinded type variables being missed. This shows up in `existentialAbstraction` where a type like `[F[_]]C[F]` which should be taken to `C[F] forSome { type F[_] }` is instead taken to `C[F]` (where `F` has escaped its quantifier). This is responsible for errors of the form reported in https://github.com/scala/bug/issues/10545 because implicit selection is driven by a specificity test which in turn relies on `existentialAbstraction`.

I've made the smallest possible change that fixes this problem, however I think that `ContainsCollector` is ambivalent about its use of `normalize` and should probably be reworked to either fully normalize (in which case `Type#contains` will see through all aliases) or to not normalize at all apart from the case where it's required to handle refined types correctly.

Fixes https://github.com/scala/bug/issues/10545. All (par)tests pass.

### DIFF
--- a/src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
+++ b/src/reflect/scala/reflect/internal/tpe/TypeMaps.scala
@@ -964,11 +964,13 @@ private[internal] trait TypeMaps {
             // We can just map over the components and wait until we see the underlying type before we call
             // normalize.
             tp.mapOver(this)
+          case TypeRef(_, sym1, _) if (sym == sym1) => result = true // catch aliases before normalization
           case _ =>
             tp.normalize match {
               case TypeRef(_, sym1, _) if (sym == sym1) => result = true
               case refined: RefinedType =>
-                tp.prefix.mapOver(this)
+                tp.prefix.mapOver(this) // Assumption is that tp was a TypeRef prior to normalization so we should
+                                        // mapOver its prefix
                 refined.mapOver(this)
               case SingleType(_, sym1) if (sym == sym1) => result = true
               case _ => tp.mapOver(this)
--- a/test/files/neg/t10545.check
+++ b/test/files/neg/t10545.check
@@ -1,0 +1,7 @@
+t10545.scala:32: error: ambiguous implicit values:
+ both method barF0 in object Bar of type [F[_]](implicit fooF: Foo[F])Bar[F]
+ and method barF1 in object Bar of type [F[_]](implicit fooF: Foo[F])Bar[F]
+ match expected type Bar[Option]
+  implicitly[Bar[Option]]
+            ^
+one error found
--- a/test/files/neg/t10545.scala
+++ b/test/files/neg/t10545.scala
@@ -1,0 +1,33 @@
+class Foo[F[_]]
+object Foo {
+  // Prior to this fix these two are ambiguous
+  implicit def fooF0[F[_]]: Foo[F] = new Foo[F]
+  implicit def fooF1: Foo[Option] = new Foo[Option]
+}
+
+class Bar[F[_]]
+object Bar extends Bar0 {
+  // Prior to this fix these two aren't selected because there is no
+  // Foo[F] due to the ambiguity above
+  // After this fix these two are ambiguous
+  implicit def barF0[F[_]](implicit fooF: Foo[F]): Bar[F] = new Bar[F]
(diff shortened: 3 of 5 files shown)


milessabin:
Relevant prior commit: https://github.com/scala/scala/commit/a56afcd6192ae69633df23137576505015d34992.

Nb. I based the unit test on `BytecodeTesting` rather than include it in `TypesTest` because the `contains` failure doesn't show up in the latter case ... a limitation of `SymbolTableForUnitTesting` I don't understand. The test also exercises `existentialAbstraction` and `isStrictlyMoreSpecific` directly which are only accessible given an instance of `Inferencer`.

retronym:
See also https://github.com/scala/scala-dev/issues/207

milessabin:
Thanks for the pointer ... I'll run a community build when this goes green.

milessabin:
The [community build](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2033/) fails with scalaz: the change produces some new implicit ambiguities ... similar situation to the SI-2712 fix.

retronym:
I'm going to review this shortly by:
 - sketching out a spectrum of ways we could implement `Type.contains` (in addition to the status quo and this patch, I could imagine a version that unpeels alias chains)
 - cataloging the callers of `contains`
 - showing how correctness of those callers would by improved or broken by the different implementations.

milessabin:
I've added a neg test which illustrates the sort of revealed ambiguity failures we get with this fix in place.
```scala
class Foo[F[_]]
object Foo {
  // Prior to this fix these two are ambiguous
  implicit def fooF0[F[_]]: Foo[F] = new Foo[F]
  implicit def fooF1: Foo[Option] = new Foo[Option]
}

class Bar[F[_]]
object Bar extends Bar0 {
  // Prior to this fix these two aren't selected because there is no
  // Foo[F] due to the ambiguity above
  // After this fix these two are ambiguous
  implicit def barF0[F[_]](implicit fooF: Foo[F]): Bar[F] = new Bar[F]
  implicit def barF1[F[_]](implicit fooF: Foo[F]): Bar[F] = new Bar[F]
}

trait Bar0 {
  // Prior to this fix we fall back to here
  implicit def barF2[F[_]]: Bar[F] = new Bar[F]
}

object Test {
  // Prior to this fix Bar.barF2[Option]
  // After this fix,
  // error: ambiguous implicit values:
  //   both method barF0 in object Bar of type [F[_]](implicit fooF: Foo[F])Bar[F]
  //   and method barF1 in object Bar of type [F[_]](implicit fooF: Foo[F])Bar[F]
  //   match expected type Bar[Option]
  //    implicitly[Bar[Option]]
  //              ^
  // one error found
  implicitly[Bar[Option]]
}
```

milessabin:
@adriaanm I'd be happy to target this at 2.13.x too.

milessabin:
Rebased to 2.13.x ...

retronym (approved):

PR #6123, opened 2017-10-11
Fix bounds of higher-kinded existentials

### LINKED ISSUE #9325: existential with higher kinded quantifier seems to lose lower bound during typedExistentialTypeTree or packSymbols
```scala
% cat sandbox/test.scala; scalac sandbox/test.scala && scala Test
object Test {

    println(typeOf[A forSome {type A >: Int <: String}])
    println(typeOf[A[Int] forSome {type A[_] >: Int <: String}])
  }
}
warning: there were three feature warnings; re-run with -feature for details
one warning found
A forSome { type A >: Int <: String }
A[Int] forSome { type A[_] <: String }
```

### LINKED ISSUE #9445: Existential higher kinded type does not work
```scala
Welcome to Scala version 2.11.7 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_45).

scala> :paste

trait H[F[_, _]]
trait T[A, B]
val h:H[T] = null
val e:H[F] forSome { type F[A, B] } = h


 found   : H[T]
 required: H[_[A, B] <: Any]
Note: T <: Any, but trait H is invariant in type F.
You may wish to define F as +F instead. (SLS 4.5)
       val e:H[F] forSome { type F[A, B] } = h
                                             ^
```

### LINKED ISSUE #10258: Foo[F] not a subtype of Foo[F] forSome { type F[_] }
Tested with Scala 2.12.1.

I have a type `Foo` parameterized by type `F[_]`. In some contexts, the type `F[_]` doesn't matter and I want to be able to talk about `Foo` without specifying what `F` is. In other contexts, `F` needs to be known.

One attempt to model this is to have `F[_]` as a type parameter of `Foo`, and use an existential type alias `type Foo0 = Foo[F] forSome { type F[_] }` for when `F` doesn't matter:

```scala
trait Foo[F[_]]

object Foo {
  type Foo0 = Foo[F] forSome { type F[_] }
}
```

(Another (failed) attempt at modeling this situation is in #10257, using a type member.)

**The problem** is that `Foo[F]` does not satisfy a requirement for an implicit `Foo0`. The type checker cannot even prove that `Foo[F] <:< Foo.Foo0`.

```scala
import Foo._

def f(implicit foo: Foo0): Unit =
  ()

def test[F[_]](implicit foo: Foo[F]): Unit = {
  // It is not even the case that
  implicitly[Foo[F] <:< Foo0] // error: Cannot prove that Foo[F] <:< Foo.Foo0

  // so then of course this gives an
  // error: could not find implicit value for parameter foo: Foo.Foo0
  f
}
```

joroKr21:
Which in turn restores the correct subtyping relation.
See the test for an example.

Fixes scala/bug#9325, fixes scala/bug#9445 and fixes scala/bug#10258

### DIFF
--- a/src/reflect/scala/reflect/internal/ExistentialsAndSkolems.scala
+++ b/src/reflect/scala/reflect/internal/ExistentialsAndSkolems.scala
@@ -62,8 +62,8 @@ trait ExistentialsAndSkolems {
     // Hanging onto lower bound in case anything interesting
     // happens with it.
     mapFrom(hidden)(s => s.existentialBound match {
-      case TypeBounds(lo, hi) => TypeBounds(lo, hiBound(s))
-      case _                  => hiBound(s)
+      case GenPolyType(tparams, TypeBounds(lo, _)) => GenPolyType(tparams, TypeBounds(lo, hiBound(s)))
+      case _ => hiBound(s)
     })
   }
 
--- a/test/files/neg/hk-existential-lb.check
+++ b/test/files/neg/hk-existential-lb.check
@@ -1,0 +1,8 @@
+hk-existential-lb.scala:3: error: type mismatch;
+ found   : Functor[Option]
+ required: Functor[_[x] >: List[x]]
+Note: Option <: [x]Any, but class Functor is invariant in type F.
+You may wish to define F as +F instead. (SLS 4.5)
+  val someF: Functor[F] forSome { type F[x] >: List[x] } = new Functor[Option]
+                                                           ^
+one error found
--- a/test/files/neg/hk-existential-lb.flags
+++ b/test/files/neg/hk-existential-lb.flags
@@ -1,0 +1,1 @@
+-language:higherKinds,existentials -Xfatal-warnings
--- a/test/files/neg/hk-existential-lb.scala
+++ b/test/files/neg/hk-existential-lb.scala
@@ -1,0 +1,4 @@
+class Functor[F[_]]
+object Functor {
+  val someF: Functor[F] forSome { type F[x] >: List[x] } = new Functor[Option]
+}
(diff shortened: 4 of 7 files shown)


milessabin:
I also spotted this recently.

milessabin:
I've scheduled a community build for this.

milessabin:
This results in a [community build fallure](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2041/) in pprint.

I think a compensating change there should be fairly straightforward ... @joroKr21 want to take a look?

joroKr21:
I found that @retronym has fixed this two years ago (retronym/scala@2614d1d73953dfa6f0a343cc74bd299eb100d2fb) but it wasn't picked up. I didn't know about `GenPolyType`. Should I update my PR to make use of it?

@milessabin I looked at PPrint and looks like it tries to make some sense of the malformed bounds [here](https://github.com/lihaoyi/PPrint/blob/005a51a509bae088dc69c4427a27280249ce8cf6/pprint/shared/src/main/scala-2.10%2B/TPrintImpl.scala#L122-L128). Should be an easy fix. @lihaoyi wdyt?

Is the passing community build a prerequisite for merging?

milessabin:
>  Is the passing community build a prerequisite for merging?

That's @adriaanm's call. Given that this is targetting 2.12.x I would have thought probably yes. It might make sense to rebase it against 2.13.x instead.

milessabin:
@joroKr21 I recommend rebasing this against 2.13.x and then backport to 2.12.x guarded by `-Xsource:2.13`.

joroKr21:
* Rebased to 2.13.x
* Now using `GenPolyType`
* Added test for preserved lower bound

adriaanm:
Excellent first contribution. I apologize for the delay. I promise we'll do better next time. This quarter has been super busy.

adriaanm (approved):

PR #6125, opened 2017-10-11
simplify LRU for parsedClasses in the ByteCodeRepository

mkeskells:
should reduce CPU and memory usage when optimising


src/compiler/scala/tools/nsc/backend/jvm/opt/ByteCodeRepository.scala
```diff
@@ -35,68 +35,33 @@ abstract class ByteCodeRepository {
...
   }
 
-  private object lruCounter extends AtomicLong(0l) with collection.generic.Clearable {
-    def clear(): Unit = { this.set(0l) }
-  }
-  recordPerRunCache(lruCounter)
-
-  /**
-   * Prevent the code repository from growing too large. Profiling reveals that the average size
-   * of a ClassNode is about 30 kb. I observed having 17k+ classes in the cache, i.e., 500 mb.
```
lrytz, line 67: could you move this comment up to the definition of `private val maxCacheSize`?

src/compiler/scala/tools/nsc/backend/jvm/opt/LruMap.scala
```diff
@@ -0,0 +1,21 @@
...
+  def apply[K,V](maxSize:Int, threadsafe:Boolean): Map[K,V] = {
+    require (maxSize > 0)
+    val basic = new LruMapImpl[K,V](maxSize)
+    val threaded = if (threadsafe) Collections.synchronizedMap(basic) else basic
+
+    threaded.asScala
+  }
+
+  private class LruMapImpl[K,V](maxSize: Int) extends LinkedHashMap[K,V] {
+    override def removeEldestEntry(eldest: JMap.Entry[K, V]): Boolean = {
```
lrytz, line 17: TIL! Thanks.

lrytz (commented):
Looks good otherwise.

lrytz:
LGTM, please squash the changes into a single commit.

mkeskells:
@lrytz will squash but there is an outstanding query for you here though - see "I dont see why this was concurrent, but retained the thread safety. @lrytz what are the required semantics"

lrytz:
I guess it depends how far we go with parallelizing the backend. The bytecode repo is used by the inliner to get bytecode, but also to obtain ClassBTypes. For all classes referenced in the source code, CodeGen creates a ClassBType from the corresponding Symbol. However, for references to "new" classes introduced by inlining, we don't have a symbol and create the ClassBType from the classfile. I think this doesn't happen eagerly in the inliner, but can happen at a later point when traversing the bytecode, for example local optimizations (`eliminateRedundantCasts`) or collecting references to nested classes (`NestedClassesCollector`).

During classfile writing, ClassBTypes are also accessed to compute stack map frames (`ClassWriterWithBTypeLub`). However, at this point the implementation expects the ClassBTypes to already exist in the cache (it calls `cachedClassBType(internalName).get`).

mkeskells:
@lrytz OK - will add a comment that concurrency is an enabler for future work

src/compiler/scala/tools/nsc/backend/jvm/opt/ByteCodeRepository.scala
mkeskells: done in e28ad86

mkeskells:
@lrytz squashed and added a comment regarding concurrency for parsedClasses

lrytz (approved):

PR #6127, opened 2017-10-12
Improvements to the type variable solver.

milessabin:
This is an attempt to fix some issues and make some small improvements in the type variable solver. It's not complete, but I'm pushing to get feedback and so that I can run a community build against it.

The main change is that I've added a more sophisticated test for the currently-being-solved `TypeVar` being used in the bounds of another type variable, allowing more information to be propagated. This also fixes a bug where the test `` case TypeRef(_, `tparam`, _) `` allows a fully applied type to be added to the bounds of a higher kinded type resulting in a blow up in glb/lub.

I've left various open issues in comments ... these should be pursued if this is followed up on.

Nb. the test cases compile in Dotty, so this is a move in the right direction.

### DIFF
--- a/src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
+++ b/src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
@@ -195,48 +195,52 @@ private[internal] trait TypeConstraints {
     def solveOne(tvar: TypeVar, tparam: Symbol, variance: Variance) {
       if (tvar.constr.inst == NoType) {
         val up = if (variance.isContravariant) !upper else upper
-        tvar.constr.inst = null
+        tvar.constr.inst = null // mark tvar as being solved
         val bound: Type = if (up) tparam.info.bounds.hi else tparam.info.bounds.lo
         //Console.println("solveOne0(tv, tp, v, b)="+(tvar, tparam, variance, bound))
         var cyclic = bound contains tparam
-        foreach3(tvars, tparams, variances)((tvar2, tparam2, variance2) => {
-          val ok = (tparam2 != tparam) && (
-            (bound contains tparam2)
-              ||  up && (tparam2.info.bounds.lo =:= tparam.tpeHK)
-              || !up && (tparam2.info.bounds.hi =:= tparam.tpeHK)
-            )
-          if (ok) {
-            if (tvar2.constr.inst eq null) cyclic = true
+
+        foreach3(tvars, tparams, variances){ (tvar2, tparam2, variance2) =>
+          val contributes = (tparam2 != tparam) && ((bound contains tparam2) || {
+            val bound2 = if (up) tparam2.info.bounds.lo else tparam2.info.bounds.hi
+            bound2 =:= tparam.typeConstructor
+          })
+          if (contributes) {
+            if (tvar2.constr.inst eq null) cyclic = true // came back to a tvar that's being solved --> cycle! (note that we capture the `cyclic` var)
             solveOne(tvar2, tparam2, variance2)
           }
-        })
+        }
+
         if (!cyclic) {
           if (up) {
             if (bound.typeSymbol != AnyClass) {
               debuglog(s"$tvar addHiBound $bound.instantiateTypeParams($tparams, $tvars)")
               tvar addHiBound bound.instantiateTypeParams(tparams, tvars)
             }
-            for (tparam2 <- tparams)
-              tparam2.info.bounds.lo.dealias match {
-                case TypeRef(_, `tparam`, _) =>
-                  debuglog(s"$tvar addHiBound $tparam2.tpeHK.instantiateTypeParams($tparams, $tvars)")
-                  tvar addHiBound tparam2.tpeHK.instantiateTypeParams(tparams, tvars)
-                case _ =>
-              }
           } else {
             if (bound.typeSymbol != NothingClass && bound.typeSymbol != tparam) {
               debuglog(s"$tvar addLoBound $bound.instantiateTypeParams($tparams, $tvars)")
               tvar addLoBound bound.instantiateTypeParams(tparams, tvars)
             }
-            for (tparam2 <- tparams)
-              tparam2.info.bounds.hi.dealias match {
+          }
+
+          // can we derive more constraints for `tvar` (and `tparam`) from its occurrences in the bounds of the other tparams?
+          foreach2(tparams, tvars) { (tparamOther, tvarOther) =>
+            if (tparamOther ne tparam) {
+              val boundOther =
+                if (up) tparamOther.info.bounds.lo else tparamOther.info.bounds.hi
+
+              boundOther.dealias match {
+                // `tparam` is the lower/upper bound of `tvarOther`. Flip that, and add `tvarOther` as an upper/lower bound for `tvar`
                 case TypeRef(_, `tparam`, _) =>
-                  debuglog(s"$tvar addLoBound $tparam2.tpeHK.instantiateTypeParams($tparams, $tvars)")
-                  tvar addLoBound tparam2.tpeHK.instantiateTypeParams(tparams, tvars)
-                case _ =>
+                  debuglog(s"$tvar propagate bound from $tvarOther")
+                  if (up) tvar.addHiBound(tvarOther) else tvar.addLoBound(tvarOther)
+                case _                       =>
               }
+            }
           }
         }
+
         tvar.constr.inst = NoType // necessary because hibounds/lobounds may contain tvar
 
         //println("solving "+tvar+" "+up+" "+(if (up) (tvar.constr.hiBounds) else tvar.constr.loBounds)+((if (up) (tvar.constr.hiBounds) else tvar.constr.loBounds) map (_.widen)))
--- a/src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
+++ b/src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
@@ -200,10 +200,15 @@ private[internal] trait TypeConstraints {
         //Console.println("solveOne0(tv, tp, v, b)="+(tvar, tparam, variance, bound))
         var cyclic = bound contains tparam
 
+        // This is attempting to reorder the solving of TypeVars such that tvs which are
+        // mentioned in the active bound are solved first.
+        // 1. Is this reordering desirable? Should solving always be left to right instead?
+        // 2. Should this take account of the variance of tparam2 and flip the bounds if
+        //    necessary?
         foreach3(tvars, tparams, variances){ (tvar2, tparam2, variance2) =>
           val contributes = (tparam2 != tparam) && ((bound contains tparam2) || {
             val bound2 = if (up) tparam2.info.bounds.lo else tparam2.info.bounds.hi
-            bound2 =:= tparam.typeConstructor
+            bound2 contains tparam
           })
           if (contributes) {
             if (tvar2.constr.inst eq null) cyclic = true // came back to a tvar that's being solved --> cycle! (note that we capture the `cyclic` var)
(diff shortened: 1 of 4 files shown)


adriaanm:
Changes to type inference and other core parts of the compiler should go to 2.13 at this point.

milessabin:
@adriaanm gotcha ... I'll rebase.

milessabin:
Green community build [here](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2040/).

milessabin:
Rebased to 2.13.x ...

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
```diff
@@ -210,31 +215,62 @@ private[internal] trait TypeConstraints {
...
+                case _ => NoType
+              }
+
+              // Try harder to propagate information we can discover from this type variables occurence in
+              // bounds on other type variable. Also avoid some ill-kinding issues.
+              // Open issue: should we track variance and and chose whether to add a high or low bound accordingly?
+              def propagate(bound0: Type, instantiate: => Type): Unit = bound0 match {
+                case TypeRef(_, `tparam`, _) if !tvar.isHigherKinded && !bound0.isHigherKinded =>
+                  addBound(instantiate)
+                case TypeRef(_, `tparam`, _) if tvar.isHigherKinded && bound0.isHigherKinded && tparam.typeParams.size == bound0.typeParams.size =>
```
joroKr21: I'm confused. When can this happen? If `bound0.isHigherKinded` shouldn't it be a `PolyType` and not a `TypeRef`? Also checking the size of type parameters might not be enough. Basically we care only if the bound is `F[x, y, z[_]] <: G[x, y, z]`, but it could be `F[x, y, z[_]] <: G[z, x, y]` or `F[x, y, z[_]] <: G[x, y, List]`.

milessabin: I can't remember the scenario now. It's possible that it's something that came up with incorrect code.

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
```diff
@@ -198,6 +198,11 @@ private[internal] trait TypeConstraints {
         val bound: Type = if (up) tparam.info.bounds.hi else tparam.info.bounds.lo
         //Console.println("solveOne0(tv, tp, v, b)="+(tvar, tparam, variance, bound))
         var cyclic = bound contains tparam
+        // This is attempting to reorder the solving of TypeVars such that tvs which are
+        // mentioned in the active bound are solved first.
+        // 1. Is this reordering desirable? Should solving always be left to right instead?
+        // 2. Should this take account of the variance of tparam2 and flip the bounds if
+        //    necessary?
```
joroKr21, line 207: When would it be necessary to flip the bounds?

milessabin: Flip the bounds depending on whether the param is co/inv vs. contra variant.

joroKr21: `up` depends on the variance of `tparam` and the variance of `tparam2` shouldn't matter I think.

milessabin: I'd like to be convinced of that :-)

joroKr21:
@milessabin do you need help with this? What more needs to be done?

milessabin:
I wanted to get things to a state where `t7289_status_quo.scala` was consistent between the covariant and bounded existential cases and all around reasonable, _without_ breaking a ton of other stuff. I ran out of time ... it'd be great if you could try and make some more progress on it.

joroKr21:
I think I made some progress on [my fork](https://github.com/joroKr21/scala/tree/bounds-propagation) that improves the status quo test. It also shows that the reordering of type variables is necessary to make that happen (otherwise `A` would be inferred as `Nothing`). But there are still some things I am concerned about:

* Blindly propagating the type constructor - think `Nil.type <:< List[x]` but `Nil.type` doesn't have a type constructor. However, I couldn't come up with a counterexample that compiled before.
* Blindly propagating the type arguments - think `Map[k, v] <:< Traversable[(k, v)]`. Again, couldn't find a counterexample.
* Respecting variance when propagating type arguments - this is trickier. What if we have bounds like this: `[X, Y, F <: X => Y]`?

joroKr21:
I tried to take care of the points above and it looks correct to me now. What would be the next step?

milessabin:
This looks great! Presumably all (par)tests pass?

I suggest PR'ing your work and I'll close this in favour of that.

adriaanm:
I added a few comments on the ~~PR~~ commit. Looks like it's heading in the right direction, but I want to be clear that there's a high bar for changes to type inference -- especially in a library release cycle (language changes should normally go to 2.14, unless needed now for some other reason, e.g. collections rework).

Can you elaborate on why it's safe/needed to check kind conformance from `solveOne`?  Do we risk adding constraints? How about cycles? Performance?

adriaanm:
That said I can be convinced to merge this, of course. Will also need a bit more comments, with motivation / brain dump, while it's fresh. Green community build.

milessabin:
I'd like to add this to https://github.com/scala/scala/pull/6378 and test with the new collections. So, far there haven't been any problems.

@joroKr21 how do you want to proceed? I'm happy for you to make a PR from your branch, or I could push your changes here ... you choose :-)

joroKr21:
@adriaanm I don't feel strongly about checking kind conformance. I was mostly worried about the first two points from https://github.com/scala/scala/pull/6127#issuecomment-374049713, but like I said I couldn't find any counterexamples. I'm inclined to agree that it's not worth the cost, hence I will remove the check.

> * Blindly propagating the type constructor - think Nil.type <:< List[x] but Nil.type doesn't have a type constructor. However, I couldn't come up with a counterexample that compiled before.
> * Blindly propagating the type arguments - think Map[k, v] <:< Traversable[(k, v)]. Again, couldn't find a counterexample.

@milessabin this is still mostly your work, so I can PR my additions to your branch.

milessabin:
Rebased and merged @joroKr21's updates.

milessabin:
@joroKr21 good work on that :-)

This must fix a bunch of type inference- and hkt-related tickets ... do we know which?

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
```diff
@@ -199,44 +199,75 @@ private[internal] trait TypeConstraints {
...
+            def propagate(variance: Variance, bound: Type, instantiate: => Type): Unit = bound match {
+              // Propagate bounds of the same kind: F[A, B, C[_]] <: G[A, B, C] => F <: G
+              // Undoing eta-expansion to ensure the propagated bound is correct.
+              case GenPolyType(tparams, TypeRef(_, `tparam`, targs))
+                if tparams.corresponds(targs) { (tp, ta) => ta.typeArgs.isEmpty && ta.typeSymbolDirect == tp }
+                => addBound(variance, instantiate)
+              // Propagate type constructor: F[A] <: B => F <: B.typeConstructor
+              // Propagate type arguments: F[+A, -B] <: C => A <: C.typeArgs(0) && B >: C.typeArgs(1)
+              case TypeRef(_, tsym, targs) =>
+                val inst = toInst(instantiate)
```
adriaanm, line 245: Check for `NoType`?

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
```diff
@@ -199,44 +199,75 @@ private[internal] trait TypeConstraints {
...
+                if (tsym == tparam) addBound(variance, inst.typeConstructor)
+                foreach3(tsym.typeParams, targs, inst.typeArgs) {
```
adriaanm: I'm not convinced of this part. Subtyping only moves to arguments once the type constructors are *equal*, not just in some subtype relation. You'd have to do something like relating F's type args with `C.baseType(F).typeArgs`, but we don't support doing that stuff on type vars. Can you show with an example that this code path is crucial, and how it works?

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
joroKr21: `addBound` checks for `NoType`

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
joroKr21: I feel like we're talking about the same thing over and over. This was my concern to begin with. For example:
* `Nil.type <:< List[x]` but they don't have the same type constructor
* `Map[k, v] <:< Traversable[(k, v)]` but they don't have the same type arguments

I tried to be on the safer side by doing a kind conformance check, but as you noted it's not a 100% guarantee and it's costly.

Propagating the type arguments is necessary for `t7289_status_quo`. Otherwise `implicitly[Ext[_ <: List[List[Int]]]]` and `implicitly[ExtCov[List[List[Int]]]]` are ambiguous.

joroKr21: In any case we have type variables on both sides so in the worst case we will infer something that doesn't respect the bounds and it will fail to compile when we check them. I couldn't find an example that compiled before and would fail with the more aggressive propagation, but if you can find one, that would be awesome.

adriaanm: Thanks for bearing with me, this is tricky stuff and it's core to the type checker, so we need to take our time to understand the change and make sure it's right. 

I don't see how this step propagates the constraints to the right type args, since `addBound` will still be constraining the type var that was originally bounded by the applied type constructor, and now we are recursing on its args without changing the type var that `addBound` works on. 

Even if we did propagate those args to the right type var somehow, I'm not sure it would help. When evaluating the progress made in the test, it's important to keep in mind the actual types inferred using `-Xprint:typer`. In the covariant case (and equivalents) we just end up with `Nothing`.  In the example, this seems to be an accident, because the type args in bounds of CC end up constraining CC itself. (Turn the `debugLog` in `addBound` into a `println` to see the trace.)

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
milessabin: Just added a check for `NoType` returned from `toInst`.

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
milessabin: @adriaanm you're right, we _are_ propagating constraints to the type var currently under consideration. We're getting those constraints from bounds on _other_ type variables that it appears in.

For instance, consider `def foo[A <: List[B], B](x: A): Option[B] = None` where `A` is solved as `List[Int]` (see, `test/files/pos/bounds-propagation.scala`). Now when we come to solve for `B` we inspect the bounds that `B` appears in, here the bound on `A`. From that we can infer an `Int` bound on `B` and so solve `B` as `Int`. With the status quo we don't consider `B`'s occurrence in other bounds so the best we get is `Nothing`.

Apropos the use of `baseType` in this context, at least one of the participants will be an unsolved type var, so we'll not get anything useful out of it. As @joroKr21 says we'll be checking for within bounds later anyway, so I don't think there's a problem here.

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
```diff
@@ -199,44 +199,77 @@ private[internal] trait TypeConstraints {
...
+                  foreach3(tsym.typeParams, targs, inst.typeArgs) {
+                    (tparam, targ, inst) => propagate(tparam.variance * variance, targ, inst)
+                  }
+                }
+              case _ =>
+            }
+
+            val variance = if (up) Variance.Covariant else Variance.Contravariant
+            val bound = if (up) tparam2.info.bounds.lo else tparam2.info.bounds.hi
+            propagate(variance, bound.dealias, tparam2.tpeHK.instantiateTypeParams(tparams, tvars))
```
adriaanm: Why not do a `foreach2(tparams, tvars)(propagateBounds)` below, so that `tparam2.tpeHK.instantiateTypeParams(tparams, tvars)` can simply become `tvar2`? The `tpeHK` of a type parameter `tparam2` is `TypeRef(NoPrefix, tparam2, Nil)`, upon which we subst that to the corresponding tvar...

adriaanm: I realize you didn't write this code, but now that we're all on this crazy ride down the rabbit hole...

adriaanm: (This one is [13 years deep](https://github.com/scala/scala/commit/d5dd908#diff-559832f08ca444d5c9c85bdebb909831R101)!)

milessabin: @adriaanm I'm not quite following what you're suggesting ... `propagateBounds` takes a single argument, so it doesn't fit as an argument to `foreach2`. We do already have `tparams.foreach(propagateBounds)` and `propagateBounds` closes over `tvars`.

joroKr21: I agree, also in that case `instantiate` doesn't need to be a by-name parameter.

adriaanm: I pushed a cleanup of the existing code that I see as uncontroversial: https://github.com/adriaanm/scala/tree/review/6127. I would be grateful if you could rebase your PR on top of this. I know this has already been a lot of work, but I'd be more comfortable if we move slowly here.

milessabin:
Looks like Jenkins fell over ...

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
milessabin: Ahh ... OK, got it. Trying that now.

adriaanm:
Artifactory has been crashing recently. Not sure why. Jenkins itself has been pretty stable.

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
adriaanm: My commit has that refactoring already in place. The remaining bit would be to dive into higher-order args, and peel off eta-expansion.

milessabin: Flippin' 'eck ... the solver has barely been touched in 13 years!?!

milessabin:
`testAll` passes for me locally. rebuild again or leave it for a bit?

milessabin:
@adriaanm your `foreach2` suggestion works out just fine. I noticed you've been playing around a bit too ...   shall I push or are you working on an alternative?

src/reflect/scala/reflect/internal/tpe/TypeConstraints.scala
adriaanm: Yup... All the more reason to be weary of users having come to depend on its precise (mal)functioning.... Experience over the years has shown that touching old code is bound to upset users. As I said before, 2.13 is focused on library changes, so we must be especially careful as we're counting on spending our "breakage budget" on the collections rework.

adriaanm:
I've pushed my work to https://github.com/adriaanm/scala/tree/review/6127, planning to leave it at  https://github.com/adriaanm/scala/commit/dc392562fe200782e4f3e6d15b9b10875a4dcf7a. Would appreciate if you could rebase on top of that as our baseline, so that the review can focus on the core of the contribution.

milessabin:
Squashed and rebased onto @adriaanm's cleanup.

adriaanm:
To avoid causing more back-and-forth, I'll adopt the part of this PR I agree with. I don't think we should go look in the type arguments. Closing in favor of https://github.com/scala/scala/pull/6492

milessabin:
That's disappointing. We'll continue in Typelevel Scala.

PR #6130, opened 2017-10-13
Improve error on absent import selector

som-snytt:
It's easy to provide extra guidance to newbies on missing dot.

Also, when reporting on newline as statement separator, indicate that it is an inferred semi.

Fixes scala/bug#10550

### DIFF
--- a/test/files/neg/badimport.check
+++ b/test/files/neg/badimport.check
@@ -1,0 +1,4 @@
+badimport.scala:2: error: . expected
+import collection
+                 ^
+one error found
--- a/test/files/neg/macro-deprecate-idents.check
+++ b/test/files/neg/macro-deprecate-idents.check
@@ -55,7 +55,7 @@ macro-deprecate-idents.scala:3: error: '=' expected but '}' found.
 macro-deprecate-idents.scala:7: error: '=' expected but '}' found.
 }
 ^
-macro-deprecate-idents.scala:42: error: '{' expected but ';' found.
+macro-deprecate-idents.scala:42: error: '{' expected.
 package foo {
 ^
 macro-deprecate-idents.scala:45: error: '{' expected but '}' found.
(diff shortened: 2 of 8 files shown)


test/files/neg/badimport.check
```diff
@@ -0,0 +1,4 @@
+badimport.scala:2: error: missing import selector collection._
+import collection
+       ^
```
retronym: I think I'd prefer to keep the existing ". expected" message, and just work to improving its position.

This position is an better than the status quo, but I think the best spot would be at the end of "collection".

test/files/neg/macro-deprecate-idents.check
```diff
@@ -55,7 +55,7 @@ macro-deprecate-idents.scala:3: error: '=' expected but '}' found.
 macro-deprecate-idents.scala:7: error: '=' expected but '}' found.
 }
 ^
-macro-deprecate-idents.scala:42: error: '{' expected but ';' found.
+macro-deprecate-idents.scala:42: error: '{' expected but (inferred) ';' found.
```
retronym: I'm surprised that so few test cases needed updating. I suppose that in most cases semi inference is disabled when waiting for a matching brace/paren.

WDYT about removing the " but (inferred) ';' found." from the error message altogether? The position of the error communicates the problem clearly.

test/files/neg/badimport.check
som-snytt: Sure, that's helpful.

test/files/neg/macro-deprecate-idents.check
som-snytt: I was also surprised. Yes, `(inferred) ';'` was not felicitous. I like both your suggestions to let the caret speak for itself. More caret than shtick.

som-snytt:
I accidentally omitted the period after `. expected`. One remedy would be to change it to:
```
. expected!
```

lrytz (approved):

PR #6131, opened 2017-10-13
Don't assume that class literals of value classes wrap TypeRefs.

hrhino:
When erasure comes around to munge a `TypeRef` to a derived value class, it swaps it out for an `ErasedValueType` that will get posterased to the underlying type. This behavior is wholly inappropriate for class literals, which one would prefer remain referencing the derived value class itself, and so such `TypeRef`s are rightfully exempted from the `specialScalaErasure`.

However, `TypeRef`s aren't the only way to reference derived value classes: the type may be an `ExistentialType` or some other wrapper of a `TypeRef` to a derived value class. The easy way to check is to call `typeSymbol.isDerivedValueClass`, and use the `javaErasure` (which erases derived value class `TypeRef`s the same as normal `TypeRef`s).

There are three cases where this matters:
  - erasing `classOf` during `erasure` (scala/bug#10551)
  - erasing the element type of `ArrayValue`s during `erasure` (scala/bug#10646)
  - erasing `classOf` inside Java annotations during `jvm`

fixes scala/bug#10551; fixes scala/bug#10646,and closes scala-js/scala-js#2801 (not like github honors this these days...)


src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1216,14 +1216,23 @@ abstract class Erasure extends InfoTransform
...
 
           if (ct.tag == ClazzTag && ct.typeValue.typeSymbol != definitions.UnitClass) {
-            val erased = ct.typeValue.dealiasWiden match {
-              case tr @ TypeRef(_, clazz, _) if clazz.isDerivedValueClass => scalaErasure.eraseNormalClassRef(tr)
+            /* scala/bug#10551: derived value class refs wrapped in existentials
+             * need to be erased the same as non-existential value class refs. */
+            val erasureTarget = ct.typeValue.dealiasWiden match {
+              case ExistentialType(_, underlying) => underlying
+              case tpe => tpe
+            }
```
hrhino: This is wrong: something like
```scala
class AV[X, Y](val xy: (X, Y)) extends AnyVal
type A[X] = AV[X, _]
type Z = A[_]
```
produces types that re-produce the bug. Running partests while I sleep; code to follow coffee tomorrow.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1216,14 +1216,24 @@ abstract class Erasure extends InfoTransform
...
+             * a derived value class, so classOf[it] doesn't get erased to the
+             * underlying type.
+             *
+             * Derived value classes can't be nested inside of other (non-static)
+             * classes, so rather than fiddling around with the erasure map, just
+             * make a vanilla `TypeRef` to the underlying symbol. */
+            val erased =
+              if (typeValue.typeSymbol.isDerivedValueClass)
+                typeRef(NoPrefix, typeValue.typeSymbol, Nil)
+              else specialScalaErasure(typeValue)
```
dwijnand: What happened to dealiasWiden?

hrhino: It went the way of the unpartested dodo. Good catch. I wonder how I could test that.

hrhino: I poked at it a bit, and I'm not experienced enough to come up with a reason why the `dealiasWiden` should matter (after this patch).

I'm not sure what it fixed before, so I've put it back.

retronym: Bikeshedding a bit.

We could just unconditionally call `specialScalaErasure` and check for the value classity by seeing if we get an `ErasedValueType` back:

```
scala> erasure.specialScalaErasure(typeOf[C]) match { case ErasedValueType(valueClass, _) => valueClass.tpeHK; case tp => tp }
res5: $r.intp.global.Type = C
```

We can also use `valueClassSymbol.tpeHK` rather than `typeRef(NoPrefix, ..., Nil)`. That is cached internally so we save a few cycles.

retronym: Or just use:

```
scala> erasure.javaErasure(typeOf[C])
res6: $r.intp.global.erasure.global.Type = C
```

I'd suggest to put in the different options and assert they give the same results (as I think they would).

hrhino: Poking at those suggestions (thanks!), a corner case arises (arrayses?):

My (current) version, my version with `.tpeHK`, and matching on `specialScalaErasure`: 
```scala
scala> :pa -raw

class A[T](val a: T) extends AnyVal

scala> classOf[Array[A[_]]]
res0: Class[Array[A[_]]] = class [Ljava.lang.Object;

scala> classOf[Array[A[Int]]]
res1: Class[Array[A[Int]]] = class [LA;
```

Using `javaErasure`:
```scala
scala> :pa -raw

class A[T](val a: T) extends AnyVal

scala> classOf[Array[A[_]]]
res0: Class[Array[A[_]]] = class [LA;

scala> classOf[Array[A[Int]]]
res1: Class[Array[A[Int]]] = class [LA;
```

But... that's not the end of the array problems...

hrhino (commented):
Here's a related problem, now known as scala/bug#10646:

```scala
scala> val aInt: A[Int] = new A(1)
aInt: A[Int] = A@1

scala> val aStr: A[String] = new A("2")
aStr: A[String] = A@32

scala> :t Array(aInt, aStr)
Array[A[_ >: String with Int]]

scala> Array(aInt, aStr)
java.lang.ArrayStoreException: java.lang.Integer
  at scala.runtime.ScalaRunTime$.array_update(ScalaRunTime.scala:71)
  at scala.Array$.$anonfun$apply$1(Array.scala:122)
  at scala.Array$.$anonfun$apply$1$adapted(Array.scala:122)
  at scala.collection.Iterator.foreach(Iterator.scala:938)
  at scala.collection.Iterator.foreach$(Iterator.scala:938)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1426)
  at scala.Array$.apply(Array.scala:122)
  ... 28 elided
```

The `ClassTag` we give to `Array.apply` is that for `[LA;`, and that means we'd need to box the elements of the value class in `A` to put them in the array, like [the overview claims](https://docs.scala-lang.org/overviews/core/value-classes.html#allocation-details).

Amusingly, the opposite problem exists on current 2.12:

```scala
scala> class A[T](val a: T) extends AnyVal
defined class A

scala> val aInt: A[Int] = new A(1)
aInt: A[Int] = A@1

scala> val aStr: A[String] = new A("2")
aStr: A[String] = A@32

scala> Array(aInt, aStr).head
java.lang.ClassCastException: java.lang.Integer cannot be cast to A
  ... 27 elided
```

So something needs to change. SIP-15 itself is unclear on what is expected. I'm not sure if this is particular to this PR, but it's worth noting.

hrhino:
So I think this fixes the array boxing problem (by always boxing value classes when put in arrays):

```diff
       def typedArrayValue(tree: ArrayValue) = {
         val elemtpt1 = typedType(tree.elemtpt, mode)
-        val elems1   = tree.elems mapConserve (elem => typed(elem, mode, elemtpt1.tpe))
+        val elemtpe1 =
+          if (phase.erasedTypes)
+            postErasure.elimErasedValueType(elemtpt1.tpe)
+          else elemtpt1.tpe
+        val elems1   = tree.elems mapConserve (elem => typed(elem, mode, elemtpe1))
         // see run/t6126 for an example where `pt` does not suffice (tagged types)
-        val tpe1     = if (isFullyDefined(pt) && !phase.erasedTypes) pt else arrayType(elemtpt1.tpe)
+        val tpe1     = if (isFullyDefined(pt) && !phase.erasedTypes) pt else arrayType(elemtpe1)

-        treeCopy.ArrayValue(tree, elemtpt1, elems1) setType tpe1
+        treeCopy.ArrayValue(tree, elemtpt1 setType elemtpe1, elems1) setType tpe1
       }
```

I'm not going to push it until I have test cases for all the corners I can think of, but it seems reasonable given the phrasing from the overview.

src/reflect/scala/reflect/internal/transform/Erasure.scala
```diff
@@ -170,9 +170,26 @@ trait Erasure {
...
+     * `javaErasure`, which treats refs to value classes the same as any other
+     * `TypeRef`. This used to be done by matching on `tr@TypeRef(_,sym,_)`, and
+     * checking whether `sym.isDerivedValueClass`, but there are more types with
+     * `typeSymbol.isDerivedValueClass` than just `TypeRef`s (`ExistentialType`
+     * is one of the easiest to bump into (`classOf[VC[_]]` for instance).
+     *
+     * tl;dr if you're trying to erase a value class ref to the value class itself
+     * and not going through this method, you're inviting trouble into your life.
+     */
+    def applyInArray(tp: Type): Type = {
```
hrhino, line 190: Amazon's new drones lost my shipment of creative juice and I haven't yet come up with a good name for this method. I'll think on it, but suggestions welcome.

test/files/run/t10551.scala
```diff
@@ -0,0 +1,48 @@
...
+  println(classOf[Biddy[_, _]])
+
+  println(classOf[Biddouble[Int]])
+  println(classOf[Biddouble[_]])
+
+  println(classOf[Bixt[Int]])
+  println(classOf[Bixt[_]])
+
+  println(classOf[Bixty])
+}
```
retronym: Old vs new:

```diff
⚡ diff -y <(cat -n /tmp/old) <(cat -n /tmp/new)
     1  class test.NotNoPrefix$Id                                    1  class test.NotNoPrefix$Id
     2  class java.lang.Object                                |      2  class test.NotNoPrefix$Id
     3  class test.NotNoPrefix$Ids                                   3  class test.NotNoPrefix$Ids
     4  interface scala.collection.Seq                        |      4  class test.NotNoPrefix$Ids
     5  class test.NotNoPrefix$Bid                                   5  class test.NotNoPrefix$Bid
     6  interface scala.collection.immutable.Map              |      6  class test.NotNoPrefix$Bid
     7  interface scala.collection.immutable.Map              |      7  class test.NotNoPrefix$Bid
     8  interface scala.collection.immutable.Map              |      8  class test.NotNoPrefix$Bid
     9  class test.NotNoPrefix$Id                                    9  class test.NotNoPrefix$Id
    10  class java.lang.Object                                |     10  class test.NotNoPrefix$Id
    11  class test.NotNoPrefix$Ids                                  11  class test.NotNoPrefix$Ids
    12  interface scala.collection.Seq                        |     12  class test.NotNoPrefix$Ids
    13  class test.NotNoPrefix$Bid                                  13  class test.NotNoPrefix$Bid
    14  interface scala.collection.immutable.Map              |     14  class test.NotNoPrefix$Bid
    15  interface scala.collection.immutable.Map              |     15  class test.NotNoPrefix$Bid
    16  interface scala.collection.immutable.Map              |     16  class test.NotNoPrefix$Bid
    17  class test.NotNoPrefix$Bid                                  17  class test.NotNoPrefix$Bid
    18  interface scala.collection.immutable.Map              |     18  class test.NotNoPrefix$Bid
    19  interface scala.collection.immutable.Map              |     19  class test.NotNoPrefix$Bid
    20  interface scala.collection.immutable.Map              |     20  class test.NotNoPrefix$Bid
    21  interface scala.collection.immutable.Map              |     21  class test.NotNoPrefix$Bid
```

retronym: I manually checked that the new, correct output also holds for `classTag`, rather than `classOf`.

retronym: That suggests that [class tag materialization](https://github.com/scala/scala/blob/71e76e8c056f0dad2447356847433ad33f3321a6/src/compiler/scala/reflect/reify/package.scala#L63-L71) was already doing the right thing, as I don't think it calls into code paths changed here.

I think we should at least expand this test case to assert identical results with `classTag`/`classOf`, and ideally create a link between the different parts of the code by changing `tpe.dealiasWiden.erasure` in `reifyRuntimeClass` to use the new method, `applyInArray`.

retronym (dismissed):
Marking review as "request changes" to address the overlap with class tag reification.

test/files/run/t10551.scala
retronym: Here's the expanded test: https://github.com/scala/scala/commit/da003548092c10f03f985ed6fbde1db300a4e76d

I tried unifying the implementation of `reifyRuntimeClass` with your code and it eluded me. I'm happy to leave that work aside and merge this PR once you add my commit with the class tag tests.

hrhino: So I did a bit of poking at that, too. I don't think we actually need to talk about erasures in `reifyRuntimeClass` at all; this seems to work perfectly well:

```diff
@@ -65,9 +68,7 @@ package object reify {
         val componentErasure = reifyRuntimeClass(global)(typer0, componentTpe, concrete)
         gen.mkMethodCall(currentRun.runDefinitions.arrayClassMethod, List(componentErasure))
       case _ =>
-        var erasure = tpe.erasure
-        if (tpe.typeSymbol.isDerivedValueClass && global.phase.id < global.currentRun.erasurePhase.id) erasure = tpe
-        gen.mkNullaryCall(currentRun.runDefinitions.Predef_classOf, List(erasure))
+        gen.mkClassOf(tpe)
     }
   }
```

Generating the call to `classOf` requires calculating the erasure, since `classOf`'s type argument isn't poly-kinded (so we need to make sure we pass it a `*`-kinded type) and since `typedClassOf` will only accept class types. However, all that `typedClassOf` does (if it succeeds) is replace the `TypeApply` with a `TypeTree` wrapping the argument, so we can just make the `TypeTree` here!

The generated `TypeTree` will be subject to the same treatment in `erasure` as if the user had written `ClassTag[T](classOf[T])` directly.

I'll push that up as a separate commit and squash if it makes sense.

hrhino: I also want to consider renaming `applyInArray`; it's not the most explanatory in some of its new usages. As I said, though, I can't think of something terse and better.

lrytz:
@retronym this looks ready to be merged.

retronym (approved):

PR #6133, opened 2017-10-14
Replace calls to sym.name with sym.nameString in ScaladocAnalyzer

Philippus:
For some reason `sym.name` adds a space to the identifier, giving "a " instead of "a", could not really figure out where that happens. `sym.nameString` doesn't.

Fixes scala/bug#10357


SethTisue:
the space is used to distinguish the field itself from its getter. background: https://github.com/scala/bug/issues/5736

adriaanm:
`getterName` would be a more robust way to rewrite both the underlying field and the setter to the corresponding getter name. Also, please format your commit message according to https://github.com/scala/scala/blob/2.12.x/CONTRIBUTING.md#clean-commits-clean-history

Philippus:
@adriaanm I changed it, is this what you had in mind?

lrytz (approved):

PR #6134, opened 2017-10-15
Default getter is never assignment

som-snytt:
Don't check whether supplied `f(x = default$getter)`
is an assignment to `x`.

Fixes scala/bug#10336

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala
@@ -596,6 +596,10 @@ trait NamesDefaults { self: Analyzer =>
       var positionalAllowed = true
       def stripNamedArg(arg: AssignOrNamedArg, argIndex: Int): Tree = {
         val AssignOrNamedArg(Ident(name), rhs) = arg
+        def invokesDefault: Boolean = rhs match {
+          case Select(_, f) => f.startsWith(nme.DEFAULT_GETTER_INIT_STRING) || f.indexOf(nme.DEFAULT_GETTER_STRING) >= 0
+          case _ => false
+        }
         params indexWhere (p => matchesName(p, name, argIndex)) match {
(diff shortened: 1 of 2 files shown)


lrytz:
Ah, I closed the ticket before seeing this PR. LGTM, but could you mark the commit `[nomerge]`? IIUC it won't apply to 2.13 once https://github.com/scala/scala/pull/6092 is in.

som-snytt:
Thanks, this is a marginal issue but it gave me something to do on the weekend while trapped indoors by raging wildfires, or by the smoke of it.

lrytz (approved):

martingd:
Scala compiler reinterpretation of assignment operator `l op= r` to `l = l op r` does not seem to work in Scala 2.12.3 and 2.12.4 unless both the “getter” and “setter” are defined *directly* on the same class/trait.

    trait ReadOnly {
      def value : Int
    }

    trait ReadWrite extends ReadOnly {
      def value_=(i: Int) : Unit
    }

    class V(private var v: Int) extends ReadWrite {
      override def value = v
      override def value_=(newValue: Int) = { v = newValue }
    }

    val v = new V(42)
    v.value += 2   // <== OK

    val rw : ReadWrite = v
    rw.value += 2  // <== Error

Adding `override def value = v` directly to trait `ReadWrite` makes the code compile without errors.

This bug is fixed in Scala 2.12.5 – the `def value = v` inherited from trait `ReadOnly` is sufficient to trigger the reinterpretation of `+=` and the original example compiles without errors.

I can find no trace of such a fix *neither* in the release note *nor* by searching the issues closed. However, this issue (#6134) seems to be the best guess of what has changed the compiler's behaviour for the better. Is that the case?

I am asking because it would be a pity if the bug reappears in the future.

som-snytt:
@martingd The relevant fix was https://github.com/scala/scala/pull/6156 and it looks like the associated ticket is as you describe.

PR #6136, opened 2017-10-15
Allow monotonic group annotations in Scaladoc

Lacaranian:
In the special case that every member of a package used the same group
annotation, Scaladoc would ignore the group. This was due to filtering
that checked the number of extant groups (including the default
"Ungrouped" group), skipping groups if only one group was found.
Unfortunately, the "Ungrouped" group could be missing if all templates
used the same group!

Fix this by filtering out the "Ungrouped" group entirely, and checking
for the existence of any other groups. Make the solution a bit less
fragile by referencing the default group constant defined in
ModelFactory, rather than the default group's string directly.

Fixes scala/bug#10078

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/html/page/Entity.scala
@@ -237,7 +237,7 @@ trait EntityPage extends HtmlPage {
             <span class="filtertype">Ordering</span>
             <ol>
               {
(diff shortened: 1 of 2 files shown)


Lacaranian:
Ready for review! There might be another place to put those shared constants used in both ModelFactory and now pages/Entity.scala? @VladUreche @adriaanm

adriaanm (approved):
Nice, thanks!

PR #6139, opened 2017-10-17
Disentangle implicit-selection logic from specificity rules and make …

### LINKED ISSUE #10526: Implicit argument list modifier affects overload resolution with explicit arguments
The following fails with an ambiguous overload,
```scala
object Test extends App {
  class Foo
  class Bar extends Foo

  def overload(implicit foo: Foo): Unit = {}
  def overload(implicit bar: Bar): Unit = {}

  overload(new Bar)
}
/*
implicit-overload.scala:8: error: ambiguous reference to overloaded definition,
both method overload in object Test of type (implicit bar: Test.Bar)Unit
and  method overload in object Test of type (implicit foo: Test.Foo)Unit
match argument types (Test.Bar)
  overload(new Bar)
  ^
one error found
*/
```
However, remove the `implicit` argument list modifier and it compiles as expected,
```scala
object Test extends App {
  class Foo
  class Bar extends Foo

  def overload(foo: Foo): Unit = {}
  def overload(bar: Bar): Unit = {}

  overload(new Bar)
}
```
This is because the logic in `Inferences.isAsSpecific` only compares the result types of method types with implicit argument lists.

I think that at least part of this logic is intended to handle implicit method selection rather than overload resolution per se, but is misplaced and should actually be moved to `ImplicitSearch.improves` where, as well as fixing this bug, we have opportunities to improve implicit method selection based on implicit argument lists.

This bug is also present in Dotty.

milessabin:
…implicit selection more consistent with normal overload resolution.

This is an attempt to make implicit selection rules align more closely with people's intuitions and the spec's aspirations that it be consistent with normal overload resolution rules. It also fixes https://github.com/scala/bug/issues/10526.

Candidate implicits are ordered via `Infer#isStrictlyMoreSpecific` which delegates to `Infer#isAsSpecific`. Prior to this PR the latter treats all methods with an implicit argument list as if they were nullary methods, leading to https://github.com/scala/bug/issues/10526 ... I believe that the motivation for this behaviour was to allow methods which are intended to appear to users to be nullary to have implicit arguments without affecting their overload resolution behaviour wrt same-named methods which are _actually_ nullary. Whilst this does the right thing in some circumstances, it leads to oddities such as https://github.com/scala/bug/issues/10526 and also means that the specificity test used to order and select implicits is very crude, in effect _only_ considering the result type of an implicit method, not its implicit arguments.

This PR addresses these issues by making the "treat implicit methods as nullary" logic explicitly flagged and enabling it only at the relevant call site in `Inferencer#inferExprAlternative`. Normal overloading rules are unchanged, and the implicit selection logic is extracted out to `ImplicitSearch#Improves`. There the relative ordering of a pair of candidate implicits is computed as follows,
1. The candidates are compared by their result type, the most specific by normal overload rules being selected (this is consistent with the behaviour prior to this PR).
2. If (1) is a tie then we compare the lengths of the implicit argument lists of the candidates (if any). The candidate with the longest argument list wins (this matches users intuitions that methods with more implicit arguments are "more demanding" hence more specific).
3. If (2) is a tie then we have a pair of candidates with implicit argument lists of equal length and which are both equally specific wrt their result type/the expected type. We now apply the full normal overloading resolution rules and choose the most specific.

This change makes combining implicits from multiple implicit scopes a great deal more straightforward, since we can now rely on more than just the result type of the implicit definition to guide implicit selection. After this PR the following works as expected,
```scala
class Show[T](val i: Int)
object Show {
  def apply[T](implicit st: Show[T]): Int = st.i

  implicit val showInt: Show[Int] = new Show[Int](0)
  implicit def fallback[T]: Show[T] = new Show[T](1)
}

class Generic
object Generic {
  implicit val gen: Generic = new Generic
  implicit def showGen[T](implicit gen: Generic): Show[T] = new Show[T](2)
}

object Test extends App {
  assert(Show[Int] == 0)
  assert(Show[String] == 1)
  assert(Show[Generic] == 2) // showGen beats fallback due to longer argument list
}
```
Thanks to the use of normal overload resolution rules we also have the ability to use a "phantom" implicit parameter with arguments ordered by subtyping to explicitly prioritize implicit definitions, something for which language extensions of one form or another have occasionally been proposed,
```scala
class Low
object Low {
  implicit val low: Low = new Low
}
class Medium extends Low
object Medium {
  implicit val medium: Medium = new Medium
}
class High extends Medium
object High {
  implicit val high: High = new High
}

class Foo[T](val i: Int)
object Foo {
  def apply[T](implicit fooT: Foo[T]): Int = fooT.i

  implicit def foo[T](implicit priority: Low): Foo[T] = new Foo[T](0)
  implicit def foobar[T](implicit priority: Low): Foo[Bar[T]] = new Foo[Bar[T]](1)
  implicit def foobarbaz(implicit priority: Low): Foo[Bar[Baz]] = new Foo[Bar[Baz]](2)
}
class Bar[T]
object Bar {
  implicit def foobar[T](implicit priority: Medium): Foo[Bar[T]] = new Foo[Bar[T]](3)
  implicit def foobarbaz(implicit priority: Medium): Foo[Bar[Baz]] = new Foo[Bar[Baz]](4)
}
class Baz
object Baz {
  implicit def baz(implicit priority: High): Foo[Bar[Baz]] = new Foo[Bar[Baz]](5)
}

object Test extends App {
  assert(Foo[Int] == 0)
  assert(Foo[Bar[Int]] == 3)
  assert(Foo[Bar[Baz]] == 5)
}
```
This PR changes the outcome of only one partest (`test/files/neg/t7289_status_quo.scala`) which is intended to exactly capture the current failure behaviour in a specific implicit resolution scenario ... I believe the new (successful) behaviour is an improvement on the old.

With that exception all (par)tests pass. One shapeless test fails to compile after this change but is trivially fixable. I'll schedule a community build when this goes green ...

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -418,6 +418,46 @@ trait Implicits {
     /** Is implicit info `info1` better than implicit info `info2`?
      */
     def improves(info1: ImplicitInfo, info2: ImplicitInfo) = {
+      def restpe(tpe: Type): Type =
+        if(isView || isFunctionType(pt)) tpe
+        else tpe match {
+          case NullaryMethodType(restpe) => restpe
+          case MethodType(_, restpe) => restpe
+          case PolyType(tparams, NullaryMethodType(restpe)) => PolyType(tparams, restpe)
+          case PolyType(tparams, MethodType(_, restpe)) => PolyType(tparams, restpe)
+          case _ => tpe
+        }
+
+      val restpe1 = restpe(info1.tpe)
+      val restpe2 = restpe(info2.tpe)
+
+      def compute = {
+        // 1. Do we win on result type specificity?
+        val result0 = isStrictlyMoreSpecific(restpe1, restpe2, info1.sym, info2.sym, nullaryImplicitArgs = false)
+        if(result0) true
+        else {
+          val result1 = isStrictlyMoreSpecific(restpe2, restpe1, info2.sym, info1.sym, nullaryImplicitArgs = false)
+          if(result1) false
+          else {
+            // 2. If (1) is a tie, take the one with the most implicit arguments
+            def depoly(tpe: Type): Type = tpe match {
+              case PolyType(_, tp) => tp
+              case tp => tp
+            }
+            val imp1 = isImplicitMethodType(depoly(info1.tpe)) || info1.tpe.params.isEmpty
+            val imp2 = isImplicitMethodType(depoly(info2.tpe)) || info2.tpe.params.isEmpty
+            if(imp1 && imp2) {
+              val params1 = info1.tpe.paramTypes
+              val params2 = info2.tpe.paramTypes
+              if(params1.length > params2.length) true
+              else if(params2.length > params1.length) false
+              else // 3. If (2) is a tie, take the most specific by normal overloading rules
+                isStrictlyMoreSpecific(info1.tpe, info2.tpe, info1.sym, info2.sym, nullaryImplicitArgs = false)
+            } else imp1
+          }
+        }
+      }
+
       if (StatisticsStatics.areSomeColdStatsEnabled) statistics.incCounter(improvesCount)
       (info2 == NoImplicitInfo) ||
       (info1 != NoImplicitInfo) && {
@@ -425,11 +465,11 @@ trait Implicits {
           improvesCache get ((info1, info2)) match {
             case Some(b) => if (StatisticsStatics.areSomeColdStatsEnabled) statistics.incCounter(improvesCachedCount); b
             case None =>
-              val result = isStrictlyMoreSpecific(info1.tpe, info2.tpe, info1.sym, info2.sym)
+              val result = compute
               improvesCache((info1, info2)) = result
               result
           }
-        } else isStrictlyMoreSpecific(info1.tpe, info2.tpe, info1.sym, info2.sym)
+        } else compute
       }
     }
     def isPlausiblyCompatible(tp: Type, pt: Type) = checkCompatibility(fast = true, tp, pt)
--- a/test/files/neg/t7289_status_quo.scala
+++ b/test/files/neg/t7289_status_quo.scala
@@ -12,7 +12,10 @@ object Test1 {
 
   // Making Ext[+T] should incur the same behavior as these. (so says @paulp)
   implicitly[Ext[_ <: List[Int]]]               // compiles
-  implicitly[Ext[_ <: List[List[Int]]]]         // fails - ambiguous
+  implicitly[Ext[_ <: List[List[Int]]]]         // compiles due to f and m having equally specific result types due
+                                                // to scala/bug#10545 but f now being selected over m because it has
+                                                // the longer implicit argument list. When the fix for 10545 is
+                                                // merged this should compile, selecting m.
   implicitly[Ext[_ <: List[List[List[Int]]]]]   // compiles
 
   // But, we currently get:
--- a/spec/07-implicits.md
+++ b/spec/07-implicits.md
@@ -90,11 +90,21 @@ The _parts_ of a type $T$ are:
 Note that packages are internally represented as classes with companion modules to hold the package members.
 Thus, implicits defined in a package object are part of the implicit scope of a type prefixed by that package.
 
-If there are several eligible arguments which match the implicit
-parameter's type, a most specific one will be chosen using the rules
-of static [overloading resolution](06-expressions.html#overloading-resolution).
-If the parameter has a default argument and no implicit argument can
-be found the default argument is used.
+If there are several eligible definitions which match an implicit parameters type, a most specific one will be
+chosen using the rules of static [overloading resolution](06-expressions.html#overloading-resolution) modified to
+accommodate the fact if any of these definitions themselves have implicit parameters then those parameters must be
+resolved during the application of the overloading rules.
+
+The relative ordering of a pair of candidate implicits is computed as follows,
+
+1. The candidates are compared by their result type, the most specific by normal overload rules being selected.
+2. If (1) is a tie then we compare the lengths of the implicit argument lists of the candidates (if any). The
+   candidate with the longest argument list wins.
+3. If (2) is a tie then we have a pair of candidates with implicit argument lists of equal length and which are both
+   equally specific wrt their result type/the expected type. We now apply the full normal overloading resolution
+   rules and choose the most specific according to those rules.
+
+If a parameter has a default argument and no implicit argument can be found the default argument is used.
 
 ###### Example
 Assuming the classes from the [`Monoid` example](#example-monoid), here is a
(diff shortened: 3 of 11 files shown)


milessabin:
As expected the [community build](https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/729/) fails due to a single shapeless test failure. Everything not downstream of shapeless builds successfully however, which is very promising.

test/files/neg/t7289_status_quo.scala
```diff
@@ -12,12 +12,15 @@ object Test1 {
 
   // Making Ext[+T] should incur the same behavior as these. (so says @paulp)
   implicitly[Ext[_ <: List[Int]]]               // compiles
-  implicitly[Ext[_ <: List[List[Int]]]]         // fails - ambiguous
+  implicitly[Ext[_ <: List[List[Int]]]]         // compiles due to f and m having equally specific result types due
+                                                // to scala/bug#10545 but f now being selected over m because it has
+                                                // the longer implicit argument list. When the fix for 10545 is
```
lrytz, line 17: shouldn't the bug reference be 10526 here?

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -401,18 +401,58 @@ trait Implicits {
...
+          case PolyType(tparams, MethodType(_, restpe)) => PolyType(tparams, restpe)
+          case _ => tpe
+        }
+
+      val restpe1 = restpe(info1.tpe)
+      val restpe2 = restpe(info2.tpe)
+
+      def compute = {
+        // 1. Do we win on result type specificity?
+        val result0 = isStrictlyMoreSpecific(restpe1, restpe2, info1.sym, info2.sym, false)
```
lrytz: style nit: use a named argument when passing a literal for a boolean param (also in the other invocations of `isStrictlyMoreSpecific`)

lrytz (commented):
The fix for scala/bug#10526 looks good to me. About the new ranking rules, should that come with a spec update? I see that the current spec is already vague, but there are new rules now that are quite different from what we have in overloading resolution. It looks like we should bring that proposal to people's attention first - what is dotty doing?

milessabin:
@lrytz I've pointed @smarter and @odersky at this ... some interest, but no very definite response as yet. I think this is consistent with @odersky's intentions to align implicit selection with normal overload resolution.

I'd be happy to add some language to the spec and put together a write up for general consumption.

test/files/neg/t7289_status_quo.scala
milessabin: No, that's the right bug reference ... scala/bug#10526 and scala/bug#10545 intersect here.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
milessabin: I'll fix.

test/files/neg/t7289_status_quo.scala
milessabin: See https://github.com/scala/scala/pull/6122.

milessabin:
Nb. shapeless has been updated to compile once this is merged.

milessabin:
@lrytz rebased and issues addressed.

test/files/neg/t7289_status_quo.scala
lrytz: Ah got it, sorry for the noise.

milessabin:
Rebased.

milessabin:
Rebased, fixed broken JUnit test, squashed, added useful commit message.

lrytz:
To me, this is still in need of a wider discussion - maybe @adriaanm can bring in a new perspective.

milessabin:
Arguably this is just aligning the implementation with the spec's aspirations, but whatever ... I'd be happy to write a short SIP up for this and have it discussed more widely. I guess I could also port to Dotty.

milessabin:
Rebased.

adriaanm:
We've discussed this in depth in our team meeting, and ran it by Martin. While we agree implicit prioritization can get tedious, we feel this PR introduces too much magic (e.g., longest argument list wins). The overall complexity of implicit resolution is already high enough. (The implicit argument list is skipped on purpose to avoid costly recursive searches.)

Ultimately, we are hopeful that we can replace type-level computation via implicit search prolog with a much more elegant, functional mechanism: [type-level meta-programming](https://github.com/dotty-staging/dotty/blob/10fee4d263de7c86bbcaf70f07a55fd2a8e14588/docs/docs/typelevel.md).

fommil:
This pull request solves a real problem that I've seen countless times across a variety of projects (from "better java" Play web servers to highfalutin typelevel magic) and I welcome and celebrate it!

@adriaanm I'd like to encourage you to reconsider closing this. The problem that is claimed to be fixed is huge and impacts a lot more codebases than I believe you to realise.

It feels unfair that the decision has been made behind closed doors. Don't we have a process for this so the community can decide how they want the language to evolve? If so, I encourage both Miles and the reviewers to proceed with this route.

SethTisue:
> It feels unfair that the decision has been made behind closed doors. Don't we have a process for this so the community can decide how they want the language to evolve?

When a PR closes, no door has shut forever. Existing PRs can be reopened. New PRs can be created.

Consider starting a thread about it on https://contributors.scala-lang.org and try and rally support there. You might be able to change minds that way directly, or perhaps it would lead to someone writing a SIP on it.

> Don't we have a process for this so the community can decide how they want the language to evolve?

We have two of them. One is a more informal process consisting of overlapping discussions that happen in lots of places both online and offline involving varying groups of people, sometimes smaller groups, sometimes larger.

The other is a more formal one, the SIP process. If a decision is sufficiently important that there must be an on-the-record discussion and a formal up-or-down vote, that's what the SIP process is for.

fommil:
@SethTisue https://contributors.scala-lang.org hasn't really worked out for me in the past so I shall leave the task of politicking to the more eloquent and diplomatic.

jvican:
@fommil https://contributors.scala-lang.org/ is precisely for this kind of discussions. It is not for bike shedding, and it's not only incorrect but disrespectful you say so for everyone investing their time in answering people's language proposals and creating proposals of their own.
 
This change needs to go through the SIP process in order to be merged: it's a language change of significant proportions. I don't see the point of raising a Lightbend support ticket. It's up to the original author of this proposal to create SIP, or to others to get involved in making that happen (or get the discussion started), to see further action.

adriaanm:
> It feels unfair that the decision has been made behind closed doors.

and

> I'm happy continuing the discussion via the private Lightbend portal

seem somewhat contradictory to me.

I'm also baffled that you keep complaining that you are somehow "forced into using a website that I don't enjoy using". We're perfectly happy to have a *technical* discussion on github, but please keep it on topic (complaining about discourse is not).

milessabin:
I'm not finding the non-technical commentary on this PR enlightening or helpful in any way. If noone has a good objection I'll delete it.

adriaanm:
You can also hide as off-topic, but either way is fine with me.

dwijnand:
:+1: for minimising as off-topic (and not deleting)

fommil:
@milessabin can we expect this branch to work as well as #6582 with `dist/mkPack` ? If so, I'd be interested in trying this out.

kailuowang:
Cats encounter this issue when migrating to 2.13. Without this PR it's yet to see how to fix without breaking our API. 
Detail in https://github.com/typelevel/cats/pull/2335#issuecomment-408236030

PR #6140, opened 2017-10-19
When solving type variables propagate solutions to bounds of variables left undetermined

### LINKED ISSUE #10528: implicit resolution error for Aux pattern with parametrized type
## Description ##

The error shows itself when implicit resolution tried to find Aux pattern for a type that has parameter. The same case without Aux pattern or without type parameter compiles fine. I expect that more complicated example should also compile, but it fails.

### Error reproduction ###

The error could be reproduced with the following code:

    trait Holder[A]
    trait NilHolder[A] extends Holder[A]

    trait Solve[A, H <: Holder[A]] {
      type Output <: Holder[A]
    }
    type SolveAux[A, H <: Holder[A], O <: Holder[A]] = Solve[A, H] {type Output = O}

    implicit def nilSolve[A] = new Solve[A, NilHolder[A]] {
      override type Output = NilHolder[A]
    }

    trait WrapSolve[A, H <: Holder[A]] {
      type Output <: Holder[A]
    }

    implicit def wrapAux[A, H <: Holder[A], O <: Holder[A]](implicit one : SolveAux[A, H, O]) =
      new WrapSolve[A, H] {
        override type Output = O
      }

    val wrapped = implicitly[WrapSolve[String, NilHolder[String]]]

### Error message ###

    OutsideImplicitTree.scala:30: error: could not find implicit value for parameter e: Parametrized.WrapSolve[String,Parametrized.NilHolder[String]]
      val wrapped = implicitly[WrapSolve[String, NilHolder[String]]]
                              ^
    one error found

I tried to `-Xlog-implicits` and it showed quite useful info. `nilSolve` was successfully adopted as implicit argument, but typechecker suddenly failed to recognized its type:

    OutsideImplicitTree.scala:30: nilSolve is not a valid implicit value for Parametrized.SolveAux[String,Parametrized.NilHolder[String],O] because:
    type parameters weren't correctly instantiated outside of the implicit tree: inferred type arguments [this.Output] do not conform to method wrapAux's type parameter bounds [O <: Parametrized.Holder[A]]
      val wrapped = implicitly[WrapSolve[String, NilHolder[String]]]
                              ^

### Workaround ###

If you chose to abandon Aux pattern compiler recognizes types immediately:

    implicit def wrapNoAux[A, H <: Holder[A]](implicit one : Solve[A, H]) =
      new WrapSolve[A, H] {
        override type Output = one.Output
      }

### When workaround does not work ###

This workaround works but is not universally applicable. For example, what If I need to perform some transformation twice?

    trait TwoTimes[A, H <: Holder[A]] {
      type Output <: Holder[A]
    }
  
    implicit def twoTimes[A, H <: Holder[A], O <: Holder[A]](implicit
      one : SolveAux[A, H, O],
      two : Solve[A, O]
    ) = new TwoTimes[A, H] {
      override type Output = two.Output
    }

### Comparing to simple type ###

If the `Holder` type is not parametrized the entire thing works flawlessly:

    trait Holder
    trait NilHolder extends Holder
  
    trait Solve[A, H <: Holder] {
      type Output <: Holder
    }
    type SolveAux[A, H <: Holder, O <: Holder] = Solve[A, H] {type Output = O}
  
    implicit def nilSolve[A] = new Solve[A, NilHolder] {
      override type Output = NilHolder
    }
  
    trait WrapSolve[A, H <: Holder] {
      type Output <: Holder
    }
  
    implicit def wrapAux[A, H <: Holder, O <: Holder](implicit one : SolveAux[A, H, O]) =
      new WrapSolve[A, H] {
        override type Output = O
      }
  
    val wrapped = implicitly[WrapSolve[String, NilHolder]]

milessabin:
When instantiating type parameters and retracting unsolved ones via `adjustTypeArgs` and friends we end up with a list of solved type params/variables and a list of still-undetermined type parameters. Typically this is followed by a substitution of the solved variables into the positions of the corresponding parameters, and type checking proceeds with the substitutions eliminated from the list of undetermined type parameters. 

However, the substituted parameters might occur as components of the bounds of the type parameters which are yet to be determined and, prior to this PR, these occurrences are _not_ substituted into with the solved variables. This leads to issues of the form seen in https://github.com/scala/bug/issues/10528. This PR performs the substitution similarly to the way it is done in `enhanceBounds` in `typedAppliedTypeTree` and fixes https://github.com/scala/bug/issues/10528.

I didn't attempt to move this functionality into `AdjustedTypeArgs` where it probably belongs, because the way the latter is structured would mean that we'd be performing side effects in a pattern match, albeit an irrefutable one. I think some restructuring of this logic would be worthwhile. I think that it should also be possible now to reinstate a commented out call to `checkBounds` [here](https://github.com/milessabin/scala/blob/e7179bda5c0b0194e01f318eb303e72d38ad781b/src/compiler/scala/tools/nsc/typechecker/Infer.scala#L1053-L1055) with something similar, but I think that would also benefit from some reworking of `AdjustedTypeArgs`.

All (par)tests pass.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -712,6 +712,17 @@ trait Infer extends Checkable {
         argtpes
     }
 
+    // This is primarily a duplicte of enhanceBounds in typedAppliedTypeTree
+    // modified to use updateInfo rather than setInfo to avoid wiping out 
+    // type history.
+    def enhanceBounds(okparams: List[Symbol], okargs: List[Type], undets: List[Symbol]): Unit =
+      undets.foreach { undet =>
+        val bounds = undet.info.bounds
+        val substBounds = bounds.subst(okparams, okargs)
+        if(bounds ne substBounds)
+          undet.updateInfo(substBounds)
+      }
+
     private def isApplicableToMethod(undetparams: List[Symbol], mt: MethodType, argtpes0: List[Type], pt: Type): Boolean = {
       val formals          = formalTypes(mt.paramTypes, argtpes0.length, removeByName = false)
       def missingArgs      = missingParams[Type](argtpes0, mt.params, x => Some(x) collect { case NamedType(n, _) => n })
@@ -728,6 +739,7 @@ trait Infer extends Checkable {
       def tryInstantiating(args: List[Type]) = falseIfNoInstance {
         val restpe = mt resultType args
         val AdjustedTypeArgs.Undets(okparams, okargs, leftUndet) = methTypeArgs(EmptyTree, undetparams, formals, restpe, args, pt)
+        enhanceBounds(okparams, okargs, leftUndet)
         val restpeInst = restpe.instantiateTypeParams(okparams, okargs)
         // #2665: must use weak conformance, not regular one (follow the monomorphic case above)
         exprTypeArgs(leftUndet, restpeInst, pt, useWeaklyCompatible = true) match {
@@ -936,11 +948,14 @@ trait Infer extends Checkable {
         List()
       } else {
         val AdjustedTypeArgs.Undets(okParams, okArgs, leftUndet) = adjustTypeArgs(tparams, tvars, targsStrict)
+        enhanceBounds(okParams, okArgs, leftUndet)
+
         def solved_s = map2(okParams, okArgs)((p, a) => s"$p=$a") mkString ","
         def undet_s = leftUndet match {
           case Nil => ""
           case ps  => ps.mkString(", undet=", ",", "")
         }
+
         printTyping(tree, s"infer solved $solved_s$undet_s")
         substExpr(tree, okParams, okArgs, pt)
         leftUndet
@@ -982,6 +997,7 @@ trait Infer extends Checkable {
 
           val AdjustedTypeArgs.AllArgsAndUndets(okparams, okargs, allargs, leftUndet) =
             methTypeArgs(fn, undetparams, formals, restpe, argtpes, pt)
+          enhanceBounds(okparams, okargs, leftUndet)
 
           if (checkBounds(fn, NoPrefix, NoSymbol, undetparams, allargs, "inferred ")) {
             val treeSubst = new TreeTypeSubstituter(okparams, okargs)
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -737,6 +737,7 @@ trait Implicits {
             // we must be conservative in leaving type params in undetparams
             // prototype == WildcardType: want to remove all inferred Nothings
             val AdjustedTypeArgs(okParams, okArgs) = adjustTypeArgs(undetParams, tvars, targs)
+            enhanceBounds(okParams, okArgs, undetParams)
 
             val subst: TreeTypeSubstituter =
               if (okParams.isEmpty) EmptyTreeTypeSubstituter
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -5188,6 +5188,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
               val asym = arg.symbol
               def abounds = asym.info.bounds
               def tbounds = tparam.info.bounds
+              // TODO investigate whether this should be merged with the near duplicate in Inferencer
+              // and whether or not we should avoid using setInfo here as well to avoid potentially
+              // trampling on type history.
               def enhanceBounds(): Unit = {
                 val TypeBounds(lo0, hi0) = abounds
                 val TypeBounds(lo1, hi1) = tbounds.subst(tparams, argtypes)
--- a/test/files/pos/t10528.scala
+++ b/test/files/pos/t10528.scala
@@ -1,0 +1,35 @@
+object Test {
+  trait Holder[A]
+  trait NilHolder[A] extends Holder[A]
+
+  trait Solve[A, H <: Holder[A]] {
+    type Output <: Holder[A]
+  }
+  type SolveAux[A, H <: Holder[A], O <: Holder[A]] = Solve[A, H] {type Output = O}
+
+  implicit def nilSolve[A] = new Solve[A, NilHolder[A]] {
+    override type Output = NilHolder[A]
+  }
+
+  trait WrapSolve[A, H <: Holder[A]] {
+    type Output <: Holder[A]
+  }
+
+  implicit def wrapAux[A, H <: Holder[A], O <: Holder[A]](implicit one : SolveAux[A, H, O]) =
+    new WrapSolve[A, H] {
+      override type Output = O
+    }
+
+  val wrapped = implicitly[WrapSolve[String, NilHolder[String]]]
+}
+
+object Test2 {
+  class Inv[T]
+  class Foo[T, U <: Inv[T]]
+
+  implicit def foo[T]: Foo[T, Inv[T]] = new Foo[T, Inv[T]]
+
+  def bar[T, U <: Inv[T]](implicit foo: Foo[T, U]): Inv[T] = new Inv[T]
+
+  val baz: Inv[Int] = bar
+}


lrytz (approved):
LGTM, passing to @adriaanm for a final look.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -712,6 +712,15 @@ trait Infer extends Checkable {
         argtpes
     }
 
+    // cp. enhanceBounds in typedAppliedTypeTree
+    def enhanceBounds(okparams: List[Symbol], okargs: List[Type], undets: List[Symbol]): Unit =
+      undets.foreach { undet =>
+        val bounds = undet.info.bounds
+        val substBounds = bounds.subst(okparams, okargs)
+        if(bounds ne substBounds)
+          undet.setInfo(substBounds)
```
adriaanm: I think we should use `updateInfo` instead of `setInfo`. The latter is problematic here because it wipes out `undet`'s type history. Since we didn't create the `undet` symbol, we can't be sure that's ok.. In practice, this isn't a problem if we're early on in the compiler and there is no type history to wipe out anyway (which is already a tenuous assumption in `typedAppliedTypeTree`). However, type inference definitely can be triggered in later phases.

I seem to remember fixing bugs resulting of similar abuse of setInfo in the GADT arena. It's also possible, as always, that I'm mistaken, but I'd like you to investigate this (e.g., by trying to construct a run where the type history is non-empty and dumping it before calling setInfo), and hear what @lrytz and @retronym think before merging.

adriaanm (changes requested):
The required change is likely small, but I'd like you to explore the potential regression outlined in my comment.

milessabin:
This appears to be fine with `updateInfo` in place of `setInfo`.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
milessabin: I'm pretty much convinced that `updateInfo` is the right thing to do rather than `setInfo`, but I've not been able to construct an example with a non-trivial `TypeHistory`. Prior to the `setInfo` call above we have a single entry either for namer or typer, and afterwards the entry is overwritten yielding exactly the same result as if `updateInfo` was called. Clearly this could go wrong if the `TypeHistory` was more interesting, but I can't see a way to get that to happen. I think that might be because these are all undetermined params, and will have been freshly cloned (eg. in `Inferencer.isApplicable`).

I'm happy to follow up on this if anyone can suggest a way of accumulating a more interesting `TypeHistory`, but for now I'm just going to switch to `updateInfo` because that appears to be the right thing to do in any case.

milessabin:
Also rebased.

src/compiler/scala/tools/nsc/typechecker/Infer.scala
adriaanm: The TypeHistory would become more interesting (and relevant) if this method is called later during compilation. Let's say specialization. It's certainly far from trivial to come up with an example (or I would have included it in my comment). If nothing comes up after playing around some more in a debugger and observing what happens, I'm happy to conclude this thread with  the side effect of leaving this breadcrumb in the mind of future contributors :-)

adriaanm:
Thanks for the quick updates to the code. Please also take into account the style guide for commit messages, as explained in the other PR. Squash the commits, and add a comment to the code that makes it clear `enhanceBounds` is duplicated, albeit with a deviation to avoid blowing away infos. Add a TODO to the original one to investigate whether `setInfo` is the right choice there too.

milessabin:
Added comments, squashed and commit message massaged.

adriaanm (approved):
❤️

retronym:
This one seems to have caused a regression (maybe, by unearthing an latent bug)

https://github.com/scala/scala-java8-compat/issues/97

```scala

trait BaseStream[T, S <: BaseStream[T, S]]
trait Stream[T] extends BaseStream[T, Stream[T]]
trait IntStream extends BaseStream[Integer, IntStream]

sealed trait SS[T, S <: BaseStream[_, S]]
object SSImplicits extends Low {
  implicit val IntValue: SS[Int, IntStream] = null
}
trait Low {
  implicit def anyStreamShape[T]: SS[T, Stream[T]] = null
}

import SSImplicits.{IntValue, anyStreamShape}

class Test {
  implicit def f[A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A, S]): S = ???

  // switch these lines and typechecking the body of `def x` fails in 2.13.x
  x
  y

  def x = f(0): IntStream
  def y = f[String, Stream[String], Vector[String]]("")

}

```

src/compiler/scala/tools/nsc/typechecker/Infer.scala
```diff
@@ -712,6 +712,17 @@ trait Infer extends Checkable {
...
 
+    // This is primarily a duplicte of enhanceBounds in typedAppliedTypeTree
+    // modified to use updateInfo rather than setInfo to avoid wiping out 
+    // type history.
+    def enhanceBounds(okparams: List[Symbol], okargs: List[Type], undets: List[Symbol]): Unit =
+      undets.foreach { undet =>
+        val bounds = undet.info.bounds
+        val substBounds = bounds.subst(okparams, okargs)
+        if(bounds ne substBounds)
+          undet.updateInfo(substBounds)
```
retronym, line 723: The problem is likely that these sharpened bounds are only intended to be seen for a single candidate in `typedImplicit0`, but end up leaking into later candidates.

retronym: Yep, seems to be the case:

```
|    |    |    |    [search #7] start `[A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A,S])S`, searching for adaptation to pt=Stream[Int] => IntStream (silent: method x in Test) implicits disabled
|    |    |    |    [search #7] considering f
|    |    |    |    |-- f BYVALmode-EXPRmode-FUNmode-POLYmode (silent: method x in Test) implicits disabled
|    |    |    |    |    [adapt] [A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A,S])S adapted to [A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A,S])S
|    |    |    |    |    \-> (a: A)(implicit ss: SS[A,S])S
|    |    |    |    solving for (A: ?A, S: ?S, CC: ?CC)
enhanceBounds(type S#21256) updated  <: BaseStream#6553[_, S#21256] to  <: BaseStream#6553[_, S#21256]
|    |    |    |    [search #8] start `[A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A,S])S` inferring type S, searching for adaptation to pt=SS[Stream[Int],S] (silent: method x in Test) implicits disabled
|    |    |    |    [search #8] considering anyStreamShape
|    |    |    |    solving for (T: ?T)
|    |    |    |    [adapt] anyStreamShape adapted to [T]=> SS[T,Stream[T]] based on pt SS[Stream[Int],S]
|    |    |    |    [search #8] solve tvars=?S, tvars.constr= >: Stream[Stream[Int]] <: Stream[Stream[Int]]
|    |    |    |    solving for (S: ?S)
enhanceBounds(type S#21256) updated  <: BaseStream#6553[_, S#21256] to  <: BaseStream#6553[_, Stream#6580[Stream#6580[Int#1192]]]
|    |    |    |    [search #8] success inferred value of type SS[Stream[Int],=?Stream[Stream[Int]]] is SearchResult(SSImplicits.anyStreamShape[Stream[Int]], TreeTypeSubstituter(List(type S),List(Stream[Stream[Int]])))
|    |    |    |    [adapt] f adapted to [A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A,S])S based on pt Stream[Int] => IntStream
|    |    |    |    [search #9] start `[A, S <: BaseStream[_, S], CC](a: A)(implicit ss: SS[A,S])S`, searching for adaptation to pt=(=> Stream[Int]) => IntStream (silent: method x in Test) implicits disabled
sandbox/test.scala:23: error: type mismatch;
 found   : Stream[Int]
 required: IntStream
  def x = f(0): IntStream
           ^
```

```diff
⚡ git diff
diff --git a/src/compiler/scala/tools/nsc/typechecker/Infer.scala b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
index 481307467c..4c1994ef61 100644
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -723,6 +723,9 @@ trait Infer extends Checkable {
         val substBounds = bounds.subst(okparams, okargs)
         if(bounds ne substBounds)
           undet.updateInfo(substBounds)
+        settings.uniqid.value = true
+        println(s"enhanceBounds($undet) updated $bounds to $substBounds")
+        settings.uniqid.value = false
       }
```

milessabin: I'll take a look.

adriaanm: Hmm, sounds like we need to clone the undetermined params for each run of the search? Hopefully not too expensive

milessabin: Yup, that was my thought too. I'm surprised I wasn't doing that already.

milessabin: Do you think `undoLog` could be used for this?

milessabin:
Bug identified by @retronym reported here: https://github.com/scala/bug/issues/10708. PR fixing it here: https://github.com/scala/scala/pull/6308.

PR #6141, opened 2017-10-19
Emit serialVersionUID fields as private.

hrhino:
The JVM serialization code doesn't care either way, and leaving them as public leads to annoying (but mostly harmless) MiMa errors if the annotation is added. Since the generated field is synthetic, it can only be referenced from Java, so privatizing it is unlikely to break anyone's use of it, or so I'd hope.

See scala/scala#6101 and scala/scala#6138 for examples of places where this has caused troubles before.

Also, because the diff was too small otherwise, have a warning for putting the annotation where it cannot possibly do any good. No one asked for it but me, but maybe it'll save someone's skin, at some time in the future.


lrytz (approved):
Thanks! It's really nice to review your PRs, everything's in place: code, commit messages, comments / docs, tests.

PR #6145, opened 2017-10-23
optimise GenBCode flag generation

mkeskells:
GenBCode.mkFlags is not an efficient way to OR explicit Ints

Not sure if the correct fix is this, or to replace GenBCode.mkFlags with a macro, both would yield the same results, efficient OR.

existing implementation is vararg based so generates an array, ArrayWrapper, and performs a foldLeft with a lambda 

Modest but simple gains in reduced CPU and allocation

timings for 50 cycles JVM phase after 10 warmup, removing top 10%, no GC
```
      00_backend-baseline	  28	 4,695.13 [-5.42% +5.85%]	 4,488.84 [-5.67% +6.17%]	   388.84 [-0.06% +0.12%]
               00_mkFlags	  32	 4,521.65 [-4.65% +9.38%]	 4,344.24 [-3.97% +7.90%]	   377.68 [-0.24% +0.08%]

```

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeSkelBuilder.scala
@@ -569,12 +569,12 @@ abstract class BCodeSkelBuilder extends BCodeHelpers {
 
       val isNative         = methSymbol.hasAnnotation(definitions.NativeAttr)
       val isAbstractMethod = rhs == EmptyTree
-      val flags = GenBCode.mkFlags(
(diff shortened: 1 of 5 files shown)


mkeskells:
what us the preferred form, | or macro?
or something else

lrytz:
I'm happy to just use `|`. Can we remove the definition of `mkFlags`?

mkeskells:
Ok will emove . Didn't want to remove if you wanted to macroise it, or had some other solution

mkeskells:
@lrytz squashed and removed mkFlags def

lrytz (approved):

PR #6148, opened 2017-10-25
SIP-36 - Adding prefix types

soronpo:
Implements SIP-36- Adding prefix types.
https://docs.scala-lang.org/sips/adding-prefix-types.html

Status: SIP is pending. Will be merged into TLS to get more response from users.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -708,6 +708,8 @@ self =>
       case _ => false
     }
 
+    def isSimpleTypeIntro: Boolean = isTypeIntroToken(in.token)
+
     def isStatSeqEnd = in.token == RBRACE || in.token == EOF
 
     def isCaseDefEnd = in.token == RBRACE || in.token == CASE || in.token == EOF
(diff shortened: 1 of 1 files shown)


SethTisue:
what did the SIP committee conclude about this...? let's have a link here

soronpo:
> what did the SIP committee conclude about this...? let's have a link here

Updated the head of the thread.

SethTisue:
thanks. closing, can always be reopened later should the situation change.

soronpo:
I'm unsure about closing. TLS acceptance policy is to have an (open?) PR against LBS https://github.com/typelevel/scala#how-to-contribute-to-typelevel-scala

SethTisue:
well, our motivation is that we don’t want lots of open stale PRs sitting around in the queue. dauntingly large PR queue causes neglect cause even more dauntingly large PR queue, which is the cycle we’ve been in lately. it’s just bookkeeping, I’m sure the TLS folks can clarify and/or adapt their rules if needed

soronpo:
@milessabin your thoughts?

milessabin:
I think that we'll have to revise the TLS policy ... we've had quite a few important PRs closed due to slow progress rather than outright rejection and we don't want to drop them from TLS on just those grounds.

Nb. I think it's perfectly reasonable to want to clean up the scala/scala PR queue.

PR #6149, opened 2017-10-26
Check bounds on implicit macro applications

dragos:
If the tree that comes back during implicit search is the result of a macro expansion, check if the expandee was a type application and check bounds on it as well, just as we do for normal inferred types.

Current state: during implicit search type arguments may be inferred, but not checked for bounds until the last step. Along the way a call to `adapt` may expand the macro with inferred types even though they don't fit the bounds. When the actual bounds check is performed, the original tree was replaced with the result of the macro, so no error will be issued and things will go along their merry way (unless this is actually wrong and leads to an ambiguous implicit that would have been otherwise discarded).

I added a call to `checkBounds` on the original (unexpanded) tree at the point where we perform the usual checkBounds. However, I'm open to a better fix if there could be one that does NOT expand the macro at all if bounds don't fit. However, since the inferred type is added and then the macro expanded during [adapt](https://github.com/dragos/scala/blob/issue/8351-implicits-bounds-check/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L1154) I'm not sure where to add the check.

Ref scala/bug#8351

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -727,6 +727,16 @@ trait Implicits {
 
             // #2421: check that we correctly instantiated type parameters outside of the implicit tree:
             checkBounds(itree3, NoPrefix, NoSymbol, undetParams, targs, "inferred ")
+
+            // In case we stepped on a macro along the way, the macro was expanded during the call to adapt. Along the way,
+            // any type parameters that were instantiated were NOT yet checked for bounds, so we need to repeat the above
+            // bounds check on the expandee tree
+            itree3.attachments.get[MacroExpansionAttachment] match {
+              case Some(MacroExpansionAttachment(exp @ TypeApply(fun, targs), _)) =>
+                checkBounds(exp, NoPrefix, NoSymbol, fun.symbol.typeParams, targs.map(_.tpe), "inferred ")
+              case _ => ()
+            }
+
             context.reporter.firstError match {
               case Some(err) =>
                 return fail("type parameters weren't correctly instantiated outside of the implicit tree: " + err.errMsg)
--- a/test/files/pos/macro-bounds-check/MacroImpl_1.scala
+++ b/test/files/pos/macro-bounds-check/MacroImpl_1.scala
@@ -1,0 +1,48 @@
+package m
+
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+trait HList
+trait Coproduct
+
+abstract class Encoder[T]
+
+object Macros {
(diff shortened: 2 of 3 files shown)


adriaanm:
Thanks for the great analysis and description! Maybe we could annotate the expected type `wildPt` passed to `adapt` to signal that `adaptType` should run a full bounds check?

dragos:
I have a test now so I can try alternative implementations.

dragos:
I'm not sure how to proceed though. I'm aware of tree attachments, but *annotations* seem to be really about source-level annotations and I would avoid polluting the symbol table with one that would be so "niche". Is there a some place in the compiler where annotations are used as internal carriers of information I could look at?

adriaanm:
Good point, the attachment goes to the tree. Maybe we can check that we're in implicit search mode during macro expansion and do a full bounds check? Not sure about the impact.

lrytz:
@adriaanm / @dragos let's try to get this over the finish line! It's a valuable fix, and it looks like we're close. If we can avoid the expansion it would be a fix for https://github.com/scala/bug/issues/8351, right?

dragos:
I can try adding `checkBounds` before expanding the macro, not sure yet if that's going to be easy. I'm also afraid this fix will break some code out there.

lrytz:
> this fix will break some code

Is this a macro application with an explicit type parameter that doesn't conform to the bounds, but then after macro expansion the resulting tree conforms, so no error is reported?

Maybe 2.13 is a better target then?

dragos:
My PR  is only about implicit search, but the ticket you linked to is about the more general case. Either way, the code shouldn't have type-checked in the first place. What's the policy for such fixes?

lrytz:
I think there's no fixed policy. To me this looks like something we can fix in a minor release, I would not expect a lot of users to depend on type arguments not matching the bounds. We can run a community build. @adriaanm?

adriaanm:
"This is clear enough of a bug that we don't need to worry about people relying on this," he said, and then a bunch of (weirdly-hooded) macros jumped from behind the shrubberies and demanded compensation.

adriaanm:
(Sorry, I spent the last two hours trying to configure a play setting in scabot. Nevermind me.)
Yeah, let's not worry about breaking the "use case" of a macro application that doesn't respect type bounds. I do agree it's tricky to fix.

dragos:
>"This is clear enough of a bug that we don't need to worry about people relying on this," he said, and then a bunch of (weirdly-hooded) macros jumped from behind the shrubberies and demanded compensation.

I know you're not joking

SethTisue:
moved to 2.12.6, but correct me if that's wrong

adriaanm (approved):
On N-th thought, I say let's go for it! :-)

PR #6150, opened 2017-10-26
When checking for a non-empty type intersection don't treat Singleton as final

milessabin:
`inferTypedPattern` checks for a non-empty intersection between the static type of the scrutinee and the inferred pattern type using `Type#isPopulated`. The latter has a shortcut execution path for final classes which accidentally picks up `Singleton` (it's a class and has the `FINAL` flag set, so that's understandable). However all singleton types are at least conceptually subtypes of `Singleton` so this leads to incorrect behaviour in cases of the form,
```scala
Tuple1[Foo.type](Foo) match {
  case Tuple1(foo: Singleton) => foo
}
```
```
error: pattern type is incompatible with expected type;
 found   : Singleton
 required: Test.Foo.type
    case Tuple1(foo: Singleton) => foo
       ^
one error found
```
This PR fixes that by adding a check that the shortcut is only followed if the type isn't `Singleton`.

This turned out to be the cause of https://github.com/typelevel/scala/issues/152 reported against the SIP-23 literal types extension, but as can be seen the example here it's an issue for singleton types generally.

### DIFF
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -1017,7 +1017,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
     /** Is this symbol effectively final? I.e, it cannot be overridden */
     final def isEffectivelyFinal: Boolean = (
-         (this hasFlag FINAL | PACKAGE)
+         ((this hasFlag FINAL | PACKAGE) && this != SingletonClass)
       || isModuleOrModuleClass && (isTopLevel || !settings.overrideObjects)
       || isTerm && (isPrivate || isLocalToBlock || (hasAllFlags(notPRIVATE | METHOD) && !hasFlag(DEFERRED)))
       || isClass && originalOwner.isTerm && children.isEmpty // we track known subclasses of term-owned classes, use that infer finality
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -4133,12 +4133,13 @@ trait Types
         throw new MatchError((tp1, tp2))
     }
 
-    def check(tp1: Type, tp2: Type) = (
-      if (tp1.typeSymbol.isClass && tp1.typeSymbol.hasFlag(FINAL))
-        tp1 <:< tp2 || isNumericValueClass(tp1.typeSymbol) && isNumericValueClass(tp2.typeSymbol)
+    def check(tp1: Type, tp2: Type) = {
+      val sym1 = tp1.typeSymbol
+      if (sym1.isClass && sym1.hasFlag(FINAL) && sym1 != SingletonClass)
+        tp1 <:< tp2 || isNumericValueClass(sym1) && isNumericValueClass(tp2.typeSymbol)
       else tp1.baseClasses forall (bc =>
         tp2.baseTypeIndex(bc) < 0 || isConsistent(tp1.baseType(bc), tp2.baseType(bc)))
-    )
+    }
 
     check(tp1, tp2) && check(tp2, tp1)
   }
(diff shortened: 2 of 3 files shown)


src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4035,7 +4035,7 @@ trait Types
     }
 
     def check(tp1: Type, tp2: Type) = (
-      if (tp1.typeSymbol.isClass && tp1.typeSymbol.hasFlag(FINAL))
+      if (tp1.typeSymbol.isClass && tp1.typeSymbol.hasFlag(FINAL) && tp1 != SingletonClass.tpe)
```
retronym: A few similar checks (e.g `CheckabilityChecker` are predicated on `isEffectivelyFinal`. I think we should also change that, too.

I'd also recommend here:

```scala
val sym1 = tp1.typeSymbol
if (sym1.isClass && sym1.hasFlag(FINAL) && sym1 != SingletonClass)
```

retronym:
It would be simpler to remove the `FINAL` modifier from the synthetic class.

I suppose the intention is to prevent inheritance:

```
scala> class C extends Singleton
<console>:11: error: illegal inheritance from final trait Singleton
       class C extends Singleton
                       ^
```

We could achieve that with `SEALED`, although this would still result in`isEffectivelyFinal` holding true. If we can find another bug like the one you mentioned that requires `isEffectivelyFinal` to be false for `SingletonClass`, that solution wouldn't work.

And just removing `FINAL` upsets the typerchecker:

```
scala> symbolOf[Singleton].resetFlag(Flag.FINAL)
res4: $r.intp.global.TypeSymbol = trait Singleton

scala> class C extends Singleton
exception when typing class C extends Object with Object {
  def <init>(): C = {
    C.super.<init>();
    ()
  }
}/class scala.reflect.internal.Trees$ClassDef
class Object is inherited twice in file <console>
scala.reflect.internal.Types$TypeError: class Object is inherited twice
	at scala.tools.nsc.typechecker.Contexts$ThrowingReporter.handleError(Contexts.scala:1405)
	at scala.tools.nsc.typechecker.Contexts$ContextReporter.issue(Contexts.scala:1257)
```

Eeek.

On second thoughts, maybe the spot you've changed is sufficient, and we should leave the can of worms closed until we identify further problems.

src/reflect/scala/reflect/internal/Types.scala
milessabin: Consider it done ...

milessabin:
Updated in response to @retronym's feedback ...

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -1014,7 +1014,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
 
     /** Is this symbol effectively final? I.e, it cannot be overridden */
     final def isEffectivelyFinal: Boolean = (
-         (this hasFlag FINAL | PACKAGE)
+         ((this hasFlag FINAL | PACKAGE) && this != SymbolClass)
```
lrytz: did you mean `SingletonClass` here?

lrytz (commented):
Looks good otherwise, the commits can be squashed.

adriaanm (changes requested):
See @lrytz's comment

src/reflect/scala/reflect/internal/Symbols.scala
milessabin: Yes I did.

milessabin:
Fixed up after @lrytz's comment and rebased.

Would you like this squashed?

adriaanm:
Yes, please squash, consolidate and reformat commit messages. Then we're ready to merge.

milessabin:
Squashed.

adriaanm:
Fixes https://github.com/scala/bug/issues/10569

PR #6151, opened 2017-10-27
Merge 2.10.x + PR 6113 to 2.11.x [ci: last-only]

retronym:
This brings in backports of basic support for Java 9.

When running on Java 9, the command line `scala` and `scalac` runner scripts must be run with `-nobootcp` to allow the compiler access to `javax.tools.ToolProvider`, which is the basis of our implementation the platform classpath.


src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1025,6 +1025,7 @@ trait Implicits {
                 getParts(bts(i))
                 i += 1
               }
+              Array("") :+ "2"
```
adriaanm: stray line?

src/compiler/scala/tools/util/PathResolver.scala
```diff
@@ -268,9 +269,20 @@ abstract class PathResolverBase[BaseClassPathType <: ClassFileLookup[AbstractFil
 
     import classPathFactory._
 
+    private def javaBootClasspath = {
+      val cp = classesInPath(javaBootClassPath)
+      val okay = cp.exists(_.findClass("java/lang/Object.class").isDefined)
```
adriaanm: can we avoid this scan unless we're under java 9?

adriaanm (commented):
Minor comments. I forget the TODO you mentioned yesterday re: documenting bootcp or something?

src/compiler/scala/tools/util/PathResolver.scala
retronym: done

retronym:
> Minor comments. I forget the TODO you mentioned yesterday re: documenting bootcp or something?

I've noted the limitation in the PR description.

retronym:
I've dropped the commit that used `jrt://`/NIO for the platform classpath impl: I'd forgotten that 2.11 also needs to build on Java 6!

PR #6152, opened 2017-10-27
Update -Ywarn-unused:params description

marconilanna:
In 2.12.4, `-Ywarn-unused:params` changed to be an expanding option that enables both `explicits` and `implicits`.

This has implications that may break builds that use `-Ywarn-unused:-params`.

This PR updates the flag description to make the relationship clear and avoid confusion.

Relates to https://github.com/scala/bug/issues/10572


som-snytt:
LGTM

PR #6154, opened 2017-10-28
Document negative inputs for `take` and `drop`

markus1189:
Adds documentation for `take` and `drop` what happens if the input is negative.


retronym (approved):
Thanks for this, and please accept our apologies for getting stuck in our PR queue for so long.

PR #6155, opened 2017-10-28
SingletonTypes can have Any as their typeSymbol, confusing kinding and bounds checking logic.

milessabin:
This turned out to be the underlying cause of https://github.com/typelevel/scala/issues/139 reported against the SIP-23 implementation in Typelevel Scala.

It turns out that in a handful of places it's assumed that `tpe.typeSymbol(Direct) == AnyClass` tells us that tpe is `Any` ... this messes up bounds and kind checking in circumstances which only really practically arise where singleton types of values of type `Any` are expressible (ie. with literal types enabled) because these `SingleTypes` delegate `typeSymbol(Direct)` to their underlying, which is `Any`. Mutatis mutandis for `Nothing`.

After discussion with @retronym and @adriaanm we concluded that the various instances of these tests should be consolidated under `typeIsAny` and `typeIsNothing` in `Types` with an implementation of the form `tp.dealias match { case TypeRef(_, XXXClass, _) => true ; case _ => false }`.

I've added a unit test here ... once the SIP 23 work is merged we can add a pos test similar to the one included here: https://github.com/milessabin/scala/commit/5bfbd2f2d79ed7dfffb107349f890d86ceb4959f#diff-479c21233a4b966839a6800f0997a975.

### DIFF
--- a/src/reflect/scala/reflect/internal/Types.scala
+++ b/src/reflect/scala/reflect/internal/Types.scala
@@ -3252,12 +3252,14 @@ trait Types
        *  }}}
        */
       def unifySimple = {
-        val sym = tp.typeSymbol
-        if (sym == NothingClass || sym == AnyClass) { // kind-polymorphic
-          // scala/bug#7126 if we register some type alias `T=Any`, we can later end
-          // with malformed types like `T[T]` during type inference in
-          // `handlePolymorphicCall`. No such problem if we register `Any`.
-          addBound(sym.tpe)
+        // scala/bug#7126 if we register some type alias `T=Any`, we can later end
+        // with malformed types like `T[T]` during type inference in
+        // `handlePolymorphicCall`. No such problem if we register `Any`.
+        if (typeIsNothing(tp)) { // kind-polymorphic
+          addBound(NothingTpe)
+          true
+        } else if(typeIsAny(tp)) { // kind-polymorphic
+          addBound(AnyTpe)
           true
         } else if (params.isEmpty) {
           addBound(tp)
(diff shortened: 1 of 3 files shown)


milessabin:
It turns out that https://github.com/scala/scala/pull/6123 results in bounds being wrapped in a `PolyType` which interferes with the revised tests. This manifests itself as a change in the output of the neg test added in #6123, adding a redundant `<: [x]Any` being rendered.

The latest commit here looks through the `PolyType` and restores the expected test output.

milessabin:
Squashed to eliminate the earlier commit invalidated by #6123.

milessabin:
@SethTisue how do I make the "continuous-integration/travis-ci/push" failure go away?

SethTisue:
I'm not sure why Travis CI is confused here — it doesn't seem to understand that your branch is in your fork, not in scala/scala. I'm not inclined to investigate unless it becomes a recurring problem. We can always `/nothingtoseehere` this one later, at merge time.

milessabin:
@SethTisue I think the problem is possibly that I accidentally pushed my branch to the scala/scala repo instead of milessabin/scala and then immediately deleted it and pushed to the correct repo. Perhaps rebasing will fix it?

SethTisue:
maybe. Travis only builds the spec anyway, so it's not worth worrying much about.

milessabin:
Looks like rebasing has done the trick.

lrytz:
Status for M3: looks fine to me, review by @adriaanm.

src/reflect/scala/reflect/internal/Types.scala
```diff
@@ -4851,8 +4853,26 @@ trait Types
   private[scala] val boundsContainType = (bounds: TypeBounds, tp: Type) => bounds containsType tp
   private[scala] val typeListIsEmpty = (ts: List[Type]) => ts.isEmpty
   private[scala] val typeIsSubTypeOfSerializable = (tp: Type) => tp <:< SerializableTpe
-  private[scala] val typeIsNothing = (tp: Type) => tp.typeSymbolDirect eq NothingClass
-  private[scala] val typeIsAny = (tp: Type) => tp.typeSymbolDirect eq AnyClass
+  private[scala] val typeIsNothing = (tp: Type) => {
```
adriaanm: Boyscout opportunity: turn these into methods. It seems backwards to me that we're creating function values for use in `toString` (as argument for `filterNot`), while all other call sites are direct invocations which would be more efficient if this is defined as a method.

adriaanm (changes requested):
If no time before M3, I'm okay merging without boyscouting.

milessabin:
@adriaanm this is actually included in the SIP-23 PR, so if you think you'll be able to merge that for M3 then I should make that change there.

src/reflect/scala/reflect/internal/Types.scala
milessabin: Just these two? Or all of them?

adriaanm: I haven't looked at the callsites for all of them, and I may be missing another reason these are here (undocumented optimization), so let's be conservative and only do these two. /cc @retronym on thoughts re: performance impact

adriaanm:
Would it be feasible to rebase the bigger PR on top of the smaller ones that it includes, if we merge them first? This PR is at the right granularity for a change like this -- would be a pity to throw that away.

milessabin:
Very much so ... I'd be more than happy to do that.

milessabin:
Boyscounted and rebased.

adriaanm:
Sorry, but the commit message does not adhere to our style guide. The first line should be under ca 60 chars, and should ready as a positive action on the code base. One possible title: "More precise check for Any/Nothing, ruling out singleton types", with follow up description. While you're at it, please squash and include the reasoning for moving the checks to a method as note in the commit message

milessabin:
Squashed and commit message massaged.

adriaanm (approved):

PR #6156, opened 2017-10-29
Assignment operator rewrite finds setter

som-snytt:
Previously, it looked where the getter was, instead of
where the getter was selected from.

Fixes scala/bug#10375


adriaanm (approved):
Thanks!

PR #6157, opened 2017-10-30
Limit the values that can be added to HashSet to Int.Max.

kunal642:
If the length of the hash table is Int.Max then throw an exception to notify that the element cannot be added. 
Earlier it used to throw NegativeArraySizeException.

Fixes: scala/bug#10409

### DIFF
--- a/src/library/scala/collection/mutable/FlatHashTable.scala
+++ b/src/library/scala/collection/mutable/FlatHashTable.scala
@@ -219,7 +219,9 @@ trait FlatHashTable[A] extends FlatHashTable.HashUtils[A] {
     tableSize = 0
     nnSizeMapReset(table.length)
     seedvalue = tableSizeSeed
-    threshold = newThreshold(_loadFactor, table.length)
+    // If the table length reaches MaximumTableCapacity of Int.MaxValue / 2 then set the
+    // threshold to Int.MaxValue - 8 as a safe value
+    threshold = if (table.length >= MaximumTableCapacity) Int.MaxValue - 8 else newThreshold(_loadFactor, table.length)
     var i = 0
     while (i < oldtable.length) {
       val entry = oldtable(i)
@@ -383,6 +385,8 @@ private[collection] object FlatHashTable {
   final def seedGenerator = new ThreadLocal[scala.util.Random] {
     override def initialValue = new scala.util.Random
   }
+
+  private val MaximumTableCapacity = 1 << 30
 
   private object NullSentinel {
     override def hashCode = 0
--- a/src/library/scala/collection/mutable/OpenHashMap.scala
+++ b/src/library/scala/collection/mutable/OpenHashMap.scala
@@ -21,6 +21,8 @@ object OpenHashMap {
   def apply[K, V](elems : (K, V)*) = new OpenHashMap[K, V] ++= elems
   def empty[K, V] = new OpenHashMap[K, V]
 
+  private val MaximumTableCapacity = 1 << 30
+
   /** A hash table entry.
     * 
     * The entry is occupied if and only if its `value` is a `Some`;
@@ -56,7 +58,7 @@ extends AbstractMap[Key, Value]
    with Map[Key, Value]
    with MapLike[Key, Value, OpenHashMap[Key, Value]] {
 
-  import OpenHashMap.OpenEntry
+  import OpenHashMap._
   private type Entry = OpenEntry[Key, Value]
 
   /** A default constructor creates a hashmap with initial size `8`.
@@ -97,7 +99,7 @@ extends AbstractMap[Key, Value]
     */
   private[this] def growTable() = {
     val oldSize = mask + 1
-    val newSize = 4 * oldSize
+    val newSize = if (oldSize >= MaximumTableCapacity) Int.MaxValue - 8 else 4 * oldSize
     val oldTable = table
     table = new Array[Entry](newSize)
     mask = newSize - 1
--- a/src/library/scala/collection/mutable/HashTable.scala
+++ b/src/library/scala/collection/mutable/HashTable.scala
@@ -269,7 +269,9 @@ trait HashTable[A, Entry >: Null <: HashEntry[A, Entry]] extends HashTable.HashU
       }
       i = i - 1
     }
-    threshold = newThreshold(_loadFactor, newSize)
+    // If the table length reaches MaximumTableCapacity of Int.MaxValue / 2 then set the
+    // threshold to Int.MaxValue - 8 as a safe value
+    threshold = if (table.length >= MaximumTableCapacity) Int.MaxValue - 8 else newThreshold(_loadFactor, table.length)
   }
 
   /* Size map handling code */
@@ -404,6 +406,8 @@ private[collection] object HashTable {
   private[collection] final def newThreshold(_loadFactor: Int, size: Int) = ((size.toLong * _loadFactor) / loadFactorDenum).toInt
 
   private[collection] final def sizeForThreshold(_loadFactor: Int, thr: Int) = ((thr.toLong * loadFactorDenum) / _loadFactor).toInt
+
+  private val MaximumTableCapacity = 1 << 30
 
   private[collection] final def capacity(expectedSize: Int) = nextPositivePowerOfTwo(expectedSize)
 


kunal642:
@SethTisue Please review.

SethTisue:
I don't think `table.length * 2 > Int.MaxValue` is the correct way to do the test, because `*` silently overflows

SethTisue:
also `assert(oldtable.length == Int.MaxValue` doesn't seem to make sense

SethTisue:
you can't use `table.length * 2 > Int.MaxValue` as the test, because `*` can silently overflow

also `assert(oldtable.length == Int.MaxValue` doesn't seem to make sense. I think that's probably why many tests are failing on CI

even once you have the fix in correct form and the existing tests are passing, in order for this to be accepted you'll also need to add one or more new tests with coverage for the fix

SethTisue:
```
[error] /home/jenkins/workspace/scala-2.12.x-validate-publish-core/src/library/scala/collection/mutable/FlatHashTable.scala:420: not found: value byteswap32
[error]     protected final def improve(hcode: Int, seed: Int) = rotateRight(byteswap32(hcode), seed)
[error]                                                                      ^
```

kunal642:
@SethTisue i did not get time to write the test cases..
Will update soon

ascii766164696D:
A safer value for arrays is a bit smaller due to specifics of the JVM implementation.
Both OpenJDK and HotSpot have `Int.MaxValue - 2` as the max size for arrays.

OpenJDK's `java.util.Hashtable` uses `Int.MaxValue - 8` as a safe value.
http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/Hashtable.java#l373

kunal642:
@SethTisue Can you help with the following failure

partest --update-check \
  /home/jenkins/workspace/scala-2.12.x-validate-test/test/files/run/t8549.scala

[error] Failed: Total 1849, Failed 1, Errors 0, Passed 1848
[error] Failed tests:
[error] 	partest
Partest version:

SethTisue:
@kunal642 you should be able to reproduce the failure locally by running `partest test/files/run/t8549.scala` from within sbt

offhand, it looks to me like the `val` you added ought to be in the companion object, not in the trait; there's no need to store a copy of the val in every instance. (in part because it affects serialization, which I assume is then leading to the test failure)

style note: in Scala we use `ThisNamingConvention` for constants, not `THE_JAVA_NAMING_CONVENTION`

kunal642:
@SethTisue I have moved the constant to the companion object but now it cannot be overridden. 
If i try to use the original values then OutOfMemory exception is thrown.

Please suggest some approach on how to write the test case.

SethTisue:
> Please suggest some approach on how to write the test case

ah.... heh, I didn't see that coming. sorry to send you on an unnecessary detour.

something's gotta give, here. I (somewhat sadly) suggest deleting the test. sorry to send you on a detour there. it's more important to save storage and avoid serialization problems, than for the code to be testable in this way.

I'd suggest commenting the `Int.MaxValue - 8` thing since it will seem rather mysterious otherwise.

retronym:
I believe we should make this change uniformly across all of our `Array` based collections in a single pass:

  - immutable.HashMap
  - immutable.HashSet (this PR)
  - mutable.ArrayBuilder
  - mutable.AnyRefMap
  - mutable.FlatHashTable
  - mutable.IntMap
  - mutable.LongMap
  - mutable.OpenHashMap
  - mutable.StringBuilder
  - mutable.ResizableArray
  - ParArray
  - ParHashSet

This might make sense to do in the 2.13.x branch as part of the current rework of the collections implementation. @szeiger is leading that work.

kunal642:
@szeiger can you please verify the PR.
i have added the change for HashMap and HashTable. All others were incrementing the value of the array by one so i do not think this change is needed for them.

kunal642:
@szeiger Hi can you start the review

src/library/scala/collection/mutable/FlatHashTable.scala
```diff
@@ -219,7 +219,9 @@ trait FlatHashTable[A] extends FlatHashTable.HashUtils[A] {
     tableSize = 0
     nnSizeMapReset(table.length)
     seedvalue = tableSizeSeed
-    threshold = newThreshold(_loadFactor, table.length)
+    // If the table length reaches MaximumTableCapacity of Int.MaxValue / 2 then set the
+    // threshold to Int.MaxValue - 8 as a safe value
+    threshold = if (table.length >= MaximumTableCapacity) Int.MaxValue - 8 else newThreshold(_loadFactor, table.length)
```
szeiger, line 224: Isn't it too late at this point? This is only the threshold. The new table has already been allocated a few lines earlier (or not, if `table.length * 2` overflows). The same goes for `HashTable` as far as I can tell. Only `OpenHashMap` limits the actual size.

kunal642: Yeah but the logic is that if the table length reaches (Int.MaxValue / 2) then set the threshold to Int.Max so that next time the table can store Int.Max number of elements before trying to grow the table.

szeiger: Ah, yes, you're right.

szeiger (approved):

src/library/scala/collection/mutable/FlatHashTable.scala
szeiger: I was just looking again at the possibility of already being over the size limit before recomputing the new threshold but it doesn't seem like this is possible. There's no way to instantiate the class with a larger initial size. Nevertheless I think it would be cleaner to do this computation inside of `newThreshold`.

SethTisue:
@Ichoran hey Ol' Eagle Eye, want to take a look?

src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -97,7 +99,7 @@ extends AbstractMap[Key, Value]
     */
   private[this] def growTable() = {
     val oldSize = mask + 1
-    val newSize = 4 * oldSize
+    val newSize = if (oldSize >= MaximumTableCapacity) Int.MaxValue - 8 else 4 * oldSize
```
Ichoran, line 102: You can't do this; the mask (see 3 lines down) requires the size to be a power of two.

src/library/scala/collection/mutable/FlatHashTable.scala
```diff
@@ -219,7 +219,9 @@ trait FlatHashTable[A] extends FlatHashTable.HashUtils[A] {
     tableSize = 0
     nnSizeMapReset(table.length)
     seedvalue = tableSizeSeed
-    threshold = newThreshold(_loadFactor, table.length)
+    // If the table length reaches MaximumTableCapacity of Int.MaxValue / 2 then set the
+    // threshold to Int.MaxValue - 8 as a safe value
```
Ichoran, line 223: Why is this "safe"?  If you never try to resize your table when getting near the maximum size, your performance will get worse and worse and worse as you add elements and eventually terminate in an infinite loop when there are no more free slots available in the flat table.

src/library/scala/collection/mutable/OpenHashMap.scala
```diff
@@ -21,6 +21,8 @@ object OpenHashMap {
   def apply[K, V](elems : (K, V)*) = new OpenHashMap[K, V] ++= elems
   def empty[K, V] = new OpenHashMap[K, V]
 
+  private val MaximumTableCapacity = 1 << 30
```
Ichoran, line 24: Why don't you ever check that the _new_ size isn't greater than this?  If you start with the default of 8 = 1<<3, and the table increases by *4 each time, you'll jump straight from 1<<29 to 1<<31.

Ichoran (changes requested):
I am not sure that this PR solves the problems it is intended to, and I am also not sure that the cure isn't worse than the disease.

Also, there should be a unit test, even if it is left commented out normally (to save time/memory), so it is at least possible to check whether the code is behaving as desired.

adriaanm:
@kunal642 Thanks for raising this and providing a fix! It looks like we have a few remaining questions to answer. I'm going to close this PR because I think we target 2.13 here. @szeiger please resubmit for 2.13.x, including addressing Ichoran's concerns.

szeiger:
So it looks like there isn't anything we can do about `OpenHashMap` because the size needs to be a power of 2. @Ichoran what do you suggest for `FlatHashTable`? Leave as-is to make it overflow instead of silently degrading performance?

Ichoran:
@szeiger - Probably, or maybe put in a different limit for the last overflow so you can pack up to, I don't know, maybe 80% capacity.

PR #6158, opened 2017-10-31
Add overflow protection to the range method

stevorobs3:
The range method was not dealing with Int.MaxValue in a well
defined way. For the first test case originally 30 samples
were being created with the last ones being negative. For the
second example this would previously produce an infitite sequence.

This is to fix scala/bug#10557

### DIFF
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -107,9 +107,18 @@ object Iterator {
   def range(start: Int, end: Int, step: Int): Iterator[Int] = new AbstractIterator[Int] {
     if (step == 0) throw new IllegalArgumentException("zero step")
     private var i = start
-    def hasNext: Boolean = (step <= 0 || i < end) && (step >= 0 || i > end)
+    private var hasOverflowed = false
+    def hasNext: Boolean = {
+      (step <= 0 || i < end) && (step >= 0 || i > end) && !hasOverflowed
+    }
     def next(): Int =
-      if (hasNext) { val result = i; i += step; result }
+      if (hasNext) {
+        val result = i
+        val nextValue = i + step
+        hasOverflowed = (step > 0) == nextValue < i
+        i = nextValue
+        result
+      }
       else empty.next()
   }
 
--- a/test/junit/scala/collection/IteratorTest.scala
+++ b/test/junit/scala/collection/IteratorTest.scala
@@ -105,6 +105,54 @@ class IteratorTest {
     assertFalse(r3 contains 5)
     assertTrue(r3.isEmpty)
   }
+  @Test def rangeOverflow(): Unit = {
+    val step = 100000000
+    val numExpectedSamples = 22
+    def createIterator = Iterator.range(0, Int.MaxValue, step)
+    assertEquals(createIterator.size, numExpectedSamples)
+    assertEquals(createIterator.min, 0)
+    assertEquals(createIterator.max, (numExpectedSamples - 1) * step)
+  }
+  @Test def rangeOverflow2() : Unit = {
+    val step = (Int.MaxValue / 2) + 1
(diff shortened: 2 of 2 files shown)


stevorobs3:
Added a couple of tests for the underflow test case.

src/library/scala/collection/Iterator.scala
```diff
@@ -107,7 +107,11 @@ object Iterator {
   def range(start: Int, end: Int, step: Int): Iterator[Int] = new AbstractIterator[Int] {
     if (step == 0) throw new IllegalArgumentException("zero step")
     private var i = start
-    def hasNext: Boolean = (step <= 0 || i < end) && (step >= 0 || i > end)
+    def hasNext: Boolean = {
+      import scala.util.Try
```
xuwei-k: unused import

test/junit/scala/collection/IteratorTest.scala
```diff
@@ -105,6 +105,42 @@ class IteratorTest {
     assertFalse(r3 contains 5)
     assertTrue(r3.isEmpty)
   }
+  @Test def rangeOverflow(): Unit = {
+    val step = 100000000
+    val numSamples = 30
+    val numExpectedSamples = 22
+    def createInerator = Iterator.range(0, Int.MaxValue, step).take(numSamples)
```
michniewicz: Is this a typo? Shouldn't be `createIterator` ? 😜

Ichoran:
After this fix, the algorithm still looks buggy: try `Iterator.range(Int.MinValue+10, Int.MaxValue-10, 1000000000)`.

You need to detect overflow when making the step, not statically after the new position is computed.  You need a second var for this; let me know if you need advice on how to write it.

stevorobs3:
Ah yes, Ive got a new idea for this using math.addExact. I'll update the pr tomorrow.

Ichoran:
You can't use `addExact` as that throws an exception, and exceptions are extremely slow.  You have to detect the overflow yourself.  For instance, if `b` is nonzero, `a + b` overflows from `a` if `(a < a+b) != (b < 0)`.

stevorobs3:
This seems like a premature optimisation. I'd be happy to write some performance based tests to verify that the new solution is fast enough.

stevorobs3:
alternatively I could copy the code from the addExact method minus the exception part?

Ichoran:
Exceptions take multiple microseconds as they fill a stack trace (important when the behavior is truly exceptional, but slow).  Traversing a range on a small collection (e.g. one element) takes a few nanoseconds.  Don't waste your time writing a benchmark; it's slower by on the order of 1,000x.

Just use math to find the bound.  Java uses a license for its source code that is not compatible with Scala's.  You may not copy the code.  Do it from scratch, or use what I wrote above.  If you're not comfortable with verifying that numeric code like this is correct, thank you for your interest in fixing this, but I'm afraid someone else will have to do it!

stevorobs3:
Thanks for the advice about exceptions, didn't realise they were that much slower, thats really helpful to know, thanks!

I am happy with the math and will put a fix together for this.

This sort of overflow addition protection code seems like it would be helpful in other parts of the code base - would it be helpful to create a method, e.g. doesAddOverflow(a : Int, b : Int) : Boolean? Or would the extra overhead of a method call be too slow?

Ichoran:
It might be helpful, but we can't change any public methods within 2.12 or we'll lose binary compatibility, and this is the sort of thing that is most useful if public.  I'd just fix this bug with the check manually inlined.

stevorobs3:
OK great I have added those changes plus a couple of test cases for the example you provided.

stevorobs3:
Actually I have just noticed that the Iterator.from() method will have the same issue so I will add the test cases for this plus a fix.

Ichoran (approved):
LGTM

PR #6159, opened 2017-11-01
TaskSupport respects specific FJP

som-snytt:
ForkJoinTaskSupport forks if it seems to be running on
a ForkJoinPool, but it should only fork on its own FJP.
This change was added to the default FJP executor service
but not backported to parallel collections.

Fixes scala/scala-parallel-collections#284


SethTisue:
will need a forward-port to https://github.com/scala/scala-parallel-collections for Scala 2.13, too

retronym (approved):

PR #6161, opened 2017-11-05
Add missing parens

Philippus:
reported here: https://contributors.scala-lang.org/t/small-correction-to-an-example-in-the-scala-spec/1288
interestingly, they are present in the Scala Overview paper: http://scala-lang.org/docu/files/ScalaOverview.pdf


adriaanm (approved):
Thanks

PR #6162, opened 2017-11-05
optimise Lazy value generation and calculation

mkeskells:
to reduce memory and lock contention in multi-threaded jvm phase

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -891,52 +891,114 @@ abstract class BTypes {
   def isCompilingPrimitive: Boolean
 
   // The [[Lazy]] and [[LazyVar]] classes would conceptually be better placed within
-  // PostProcessorFrontendAccess (they access the `frontendLock` defined in that class). However,
+  // PostProcessorFrontendAccess (they may access the `frontendLock` defined in that class). However,
   // for every component in which we define nested classes, we need to make sure that the compiler
   // knows that all component instances (val frontendAccess) in various classes are all the same,
   // otherwise the prefixes don't match and we get type mismatch errors.
   // Since we already do this dance (val bTypes: GenBCode.this.bTypes.type = GenBCode.this.bTypes)
   // for BTypes, it's easier to add those nested classes to BTypes.
 
-  object Lazy {
-    def apply[T <: AnyRef](t: => T): Lazy[T] = new Lazy[T](() => t)
+  abstract sealed class Lazy[+T] {
+    /** get the result of the lazy value, calculating the result and performing the additional actions if the value
+      * is not already known.
+      */
+    def force: T
+
+    /** add an accumulating action, which is performed when the result is forced.
+      * If the result is already known at the time of this call then perform the action immediately
+      *
+      * If the result is not known the action will be performed after the value is force.
+      * The order of application of multiple onForce definitions is undefined
+      */
+    def onForce(f: T => Unit): Unit
   }
+  object Lazy {
 
-  /**
-   * A lazy value that synchronizes on the `frontendLock`, and supports accumulating actions
-   * to be executed when it's forced.
-   */
-  final class Lazy[T <: AnyRef](t: () => T) {
-    @volatile private var value: T = _
+    /**
+      * create a Lazy, whose calculation is performed with `frontendLock`
+      */
+    def withLock[T <: AnyRef](t: => T): Lazy[T] = new LazyWithLock[T](() => t)
 
-    private var initFunction = {
-      val tt = t // prevent allocating a field for t
-      () => { value = tt() }
+    /**
+      * create a Lazy, whose calculation is conditionally performed with `frontendLock` or eagerly evaluated
+      */
+    def withLockOrEager[T <: AnyRef](beLazy:Boolean, t: => T): Lazy[T] =
+      if (beLazy) new LazyWithLock[T](() => t)
+      else eager(t)
+    /**
+      * create a Lazy where the result is pre-determined, typically a constant, e.g. Nil None etc
+      */
+    def eager[T <: AnyRef](value: T): Lazy[T] = new Eager[T](value)
+
+    /**
+      * create a Lazy, whose calculation is performed on demand
+      */
+    def withoutLock[T <: AnyRef](t: => T): Lazy[T] = new LazyWithoutLock[T](() => t)
+
+    val eagerNil = eager(Nil)
+    val eagerNone = eager(None)
+
+    private final class Eager[T](val force: T) extends Lazy[T] {
+      def onForce(f: T => Unit): Unit = f(force)
+
+      override def toString = force.toString
     }
 
-    override def toString = if (value == null) "<?>" else value.toString
+    private abstract class AbstractLazy[T <: AnyRef](private var t: () => T) extends Lazy[T] {
+      // value need be volatile to ensure that init doesn't expose incomplete results
+      // due to JVM inlining of init
+      @volatile protected var value: T = _
 
-    def onForce(f: T => Unit): Unit = {
-      if (value != null) f(value)
-      else frontendSynch {
+      override def toString = if (value == null) "<?>" else value.toString
+
+      private var postForce: List[T => Unit] = Nil
+
+      def onForce(f: T => Unit): Unit = {
         if (value != null) f(value)
-        else {
-          val prev = initFunction
-          initFunction = () => {
-            prev()
-            f(value)
-          }
+        else this.synchronized {
+          if (value != null) f(value)
+          else postForce = f :: postForce
         }
+      }
+
+      def force: T = {
+        // assign to local var to avoid volatile reads and associated memory barriers
+        var result = value
+        if (result != null) result
+        else {
+          result = init(t)
+          t = null
+          this.synchronized {
+            postForce foreach (_.apply (result))
+            postForce = Nil
+          }
+          result
+        }
+      }
+
+      def init(t: () => T): T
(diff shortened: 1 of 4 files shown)


lrytz:
Test failures:
```
[error] Test scala.tools.nsc.backend.jvm.BTypesTest.lazyLockForceTest failed: org.junit.ComparisonFailure: expected:<1-[2:hi-3:hi]:hi> but was:<1-[3:hi:hi-2]:hi>, took 0.003 sec
[error]     at scala.tools.nsc.backend.jvm.BTypesTest.lazyLockForceTest(BTypesTest.scala:91)
```

```
[error] Test scala.tools.nsc.backend.jvm.BTypesTest.lazyNoLockForceTest failed: org.junit.ComparisonFailure: expected:<1-[2:hi-3:hi]:hi> but was:<1-[3:hi:hi-2]:hi>, took 0.001 sec
[error]     at scala.tools.nsc.backend.jvm.BTypesTest.lazyNoLockForceTest(BTypesTest.scala:102)
```

mkeskells:
Yes – fixing and enhancing the test now

lrytz:
Thanks - I was just pointing it out because it's a bit hard to find in the logs

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -891,52 +891,113 @@ abstract class BTypes {
...
-      else frontendSynch {
-        if (value == null) {
-          initFunction()
-          initFunction = null
-        }
+    /**
+      * A lazy value that synchronizes on the `frontendLock`, and supports accumulating actions
+      * to be executed when it's forced.
+      */
+    private final class LazyWithLock[T <: AnyRef](t: () => T) extends AbstractLazy[T](t) {
```
lrytz, line 986: In the previous implementation the init function (`t` here) was set to `null` after being used. Maybe we should keep that to release object references?

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
```diff
@@ -422,12 +422,12 @@ abstract class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
 
     val nestedClasses = {
       val ph = phase
-      Lazy(enteringPhase(ph)(nestedClassSymbolsNoJavaModuleClasses.map(classBTypeFromSymbol)))
+      Lazy.withLock(enteringPhase(ph)(nestedClassSymbolsNoJavaModuleClasses.map(classBTypeFromSymbol)))
```
lrytz: I wonder if we can identify certain `Lazy`s that we know will be forced, depending on the callsite where the `ClassBType` / `ClassInfo` is created from (code generation in particular)? Maybe you could experiment a bit (track from where the `ClassBType`s are created and see which ones will be completed)?

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -891,52 +891,113 @@ abstract class BTypes {
...
   // otherwise the prefixes don't match and we get type mismatch errors.
   // Since we already do this dance (val bTypes: GenBCode.this.bTypes.type = GenBCode.this.bTypes)
   // for BTypes, it's easier to add those nested classes to BTypes.
 
-  object Lazy {
-    def apply[T <: AnyRef](t: => T): Lazy[T] = new Lazy[T](() => t)
+  abstract sealed class Lazy[T] {
+    /** get the result of the lazy value, calculation the result and performing the additional actions if the value
+      * is not already known.
+      * @return the
```
lrytz: Let's keep the doc comments concise; the main comment is enough, no need for the annotations (also in the rest of the PR). Also, the comment here is a bit hard to read.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -891,52 +891,113 @@ abstract class BTypes {
...
+      // value need be volatile to ensure that init doesn't expose incomplete results
+      // due to JVM inlining of init
+      @volatile protected var value: T = _
 
-    def onForce(f: T => Unit): Unit = {
-      if (value != null) f(value)
-      else frontendSynch {
+      override def toString = if (value == null) "<?>" else value.toString
+
+      private var postForce: List[T => Unit] = Nil
```
lrytz, line 954: The previous implementation had the advantage of not requiring a field for those actions - could we still use that approach now?

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
mkeskells: I did have a brief try with making it more eager. I agree that if we know its going to be needed it better to do it more eagerly though.

I didn't spend much time on that though, but didnt see any obvious pattern

mkeskells:
I have fixed and added some test cases, clarified the execution order of accumulated functions (slightly different to pre-existing), and reduced the comments

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: done

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: I don't think that was a true saving. You had a inner class that wrapped the action via a dynamic class and it also extended the lock duration

If we are really concerned with these field counts we have save much more by making these top level classes and not using cake pattern everywhere

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: done

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
mkeskells: in my testing of the 3 call sites that use the lock about 70% of them were expanded.

Another thought that I had was that the background threads could expand the non symbol Lazy values more eagerly if we know some that we will need ( the withoutLock values). This doesn't occur without optimisation as I see it and that isn't my focus really

mkeskells: @lrytz I went back and had another look. Some improvements added and I think that there are more to come.Not sure if they are part of this PR though.Will discuss via email

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -891,52 +891,105 @@ abstract class BTypes {
...
+        if (result != null) result
         else {
-          val prev = initFunction
-          initFunction = () => {
-            prev()
-            f(value)
+          result = init(t)
+          t = null
+          this.synchronized {
+            postForce foreach (_ (result))
```
lrytz: the syntax here could be more explicit `postForce.foreach(_.apply(result))`

lrytz (commented):
Benchmarks are scheduled, results should [come up here](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=38628d11fc0c67b552e9c640a7f544dd7563235c%7C6afe450f8e22ebe36fe04bf7b92f1a6f78f18ccf%7C28f9105763936c2da7ffe10ebd79561b35ac4f6a&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@&from=1504445910328&to=1510322201039)

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
mkeskells: I did consider that.... Have changed to be more explicit

I would not expect to see any material change to the performance from this PR, until we get a parallel genBCode

Maybe a slight memory allocation reduction, but I don't think that your benchmarks include that measure

lrytz:
LGTM, but could you adjust the commit messages so that the first lines are not too long (72, https://github.com/scala/scala/blob/2.13.x/CONTRIBUTING.md#clean-commits-clean-history)?

mkeskells:
commit messages adjusted

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -891,52 +891,114 @@ abstract class BTypes {
...
-      if (value != null) f(value)
-      else frontendSynch {
+      override def toString = if (value == null) "<?>" else value.toString
+
+      private var postForce: List[T => Unit] = Nil
+
+      def onForce(f: T => Unit): Unit = {
         if (value != null) f(value)
+        else this.synchronized {
+          if (value != null) f(value)
```
jbgi, line 959: > assign to local var to avoid volatile reads and associated memory barriers

not done above.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -891,52 +891,114 @@ abstract class BTypes {
...
-        }
+    /**
+      * A lazy value that synchronizes on the `frontendLock`, and supports accumulating actions
+      * to be executed when it's forced.
+      */
+    private final class LazyWithLock[T <: AnyRef](t: () => T) extends AbstractLazy[T](t) {
+
+      def init(t: () => T): T = frontendSynch {
+        if (value == null) value = t()
+        value
```
jbgi, line 990: > assign to local var to avoid volatile reads and associated memory barriers

not done above.

PR #6163, opened 2017-11-06
[REJECTED] Optionally shutdown fork join pool on interrupt

som-snytt:
A system flag `scala.concurrent.context.interruptible`
to mean shut down pool on any interrupted worker thread.

Fixes scala/bug#10587

### DIFF
--- a/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
+++ b/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
@@ -95,6 +95,9 @@ private[concurrent] object ExecutionContextImpl {
       case s if s.charAt(0) == 'x' => (Runtime.getRuntime.availableProcessors * s.substring(1).toDouble).ceil.toInt
       case other => other.toInt
     }
+    def getBool(name: String) = try java.lang.Boolean.getBoolean(name) catch {
+      case e: SecurityException => false
+    }
 
     def range(floor: Int, desired: Int, ceiling: Int) = scala.math.min(scala.math.max(floor, desired), ceiling)
     val numThreads = getInt("scala.concurrent.context.numThreads", "x1")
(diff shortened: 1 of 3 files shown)


src/library/scala/concurrent/impl/ExecutionContextImpl.scala
```diff
@@ -111,16 +114,24 @@ private[concurrent] object ExecutionContextImpl {
     // The thread factory must provide additional threads to support managed blocking.
     val maxExtraThreads = getInt("scala.concurrent.context.maxExtraThreads", "256")
 
+    val interruptible = getBool("scala.concurrent.context.interruptible")
+
     val uncaughtExceptionHandler: Thread.UncaughtExceptionHandler = new Thread.UncaughtExceptionHandler {
-      override def uncaughtException(thread: Thread, cause: Throwable): Unit = reporter(cause)
+      override def uncaughtException(thread: Thread, cause: Throwable): Unit = {
+        if (interruptible) (thread, cause) match {
+          case (fjw: ForkJoinWorkerThread, _: InterruptedException) => fjw.getPool.shutdownNow()
```
viktorklang, line 122: Can someone please explain how this is desirable behavior, with precedent?

som-snytt: @viktorklang it's totally dubious. The use case was a maven plugin to run a `main` and attempt to ensure daemon threads terminate with shutdown-like semantics (without extra code). I don't think that's a thing, but the linked ticket on scala-js says it is a thing they contend with. This PR is in case someone thinks the use case is useful. The DIY code is not onerous, even for casual usage.

viktorklang: Yeah, but this PR is what you absolutely don't want to have. :)

som-snytt: I just saw the other comment about vanilla FJP.

PR #6164, opened 2017-11-06
Restrict library and reflect to compact1 profile

adriaanm:
The scala library (core and reflect) now depends only on the compact1 profile.

```
jdeps -s -profile /Users/adriaan/.ivy2/local/org.scala-lang/scala-{library,reflect}/2.13.0-bin-local-07ff7ac/jars/scala-*.jar                                                                                                               
scala-library.jar -> java.base (compact1)
scala-reflect.jar -> java.base (compact1)
scala-reflect.jar -> /Users/adriaan/.ivy2/local/org.scala-lang/scala-library/2.13.0-bin-local-07ff7ac/jars/scala-library.jar
```

See scala/bug#10559


src/library/scala/sys/process/package.scala
```diff
@@ -201,12 +201,6 @@ package scala.sys {
     *   associated with it.
     */
   package object process extends ProcessImplicits {
-    /** The arguments passed to `java` when creating this process */
-    def javaVmArguments: List[String] = {
-      import scala.collection.JavaConverters._
-
-      java.lang.management.ManagementFactory.getRuntimeMXBean.getInputArguments.asScala.toList
```
lrytz, line 208: Since this method is not deprecated in 2.12, we could do a reflective implementation
```
scala> import scala.collection.JavaConverters._

scala> val mxb = Class.forName("java.lang.management.ManagementFactory").getMethod("getRuntimeMXBean").invoke(null)
mxb: Object = sun.management.RuntimeImpl@4447c594

scala> Class.forName("java.lang.management.RuntimeMXBean").getMethod("getInputArguments").invoke(mxb).asInstanceOf[java.util.List[String]].asScala.toList
res19: List[String] = List(-Xmx1536m, -Xms1536m, -Xss5m, -XX:+UseCompressedOops, -XX:+UseParallelGC, -Xbootclasspath/a:/Users/luc/scala/scala-2.12.3/lib/jline-2.14.4.jar:/Users/luc/scala/scala-2.12.3/lib/scala-compiler.jar:/Users/luc/scala/scala-2.12.3/lib/scala-library.jar:/Users/luc/scala/scala-2.12.3/lib/scala-parser-combinators_2.12-1.0.6.jar:/Users/luc/scala/scala-2.12.3/lib/scala-reflect.jar:/Users/luc/scala/scala-2.12.3/lib/scala-swing_2.12-2.0.0.jar:/Users/luc/scala/scala-2.12.3/lib/scala-xml_2.12-1.0.6.jar:/Users/luc/scala/scala-2.12.3/lib/scalap-2.12.3.jar, -Dscala.boot.class.path=/Users/luc/scala/scala-2.12.3/lib/jline-2.14.4.jar:/Users/luc/scala/scala-2.12.3/lib/scala-compiler.jar:/Users/luc/scala/scala-2.12.3/lib/scala-library.jar:/Users/luc/scala/...
```

This would fail with `java.lang.ClassNotFoundException` if the classes are not available, which is probably comparable to what currently happens.

We can deprecate the method at the same time.

lrytz (commented):
cc @nilskp for the removal of BeanInfo, since you recently voiced your interest in that feature (https://users.scala-lang.org/t/beaninfo-alternative/1301). I guess the same functionality could be implemented with a compiler plugin.

@adriaanm how did you identify implementations that use classes outside the compact1 profile? How can we prevent new references from sneaking in?

lrytz:
Ah, `jdeps`!

src/library/scala/sys/process/package.scala
lrytz: I see you're deprecating it for 2.12.5; I guess either way is fine.

dwijnand:
> How can we prevent new references from sneaking in?

:+1: `jdeps` should be used to enforce nothing sneaks in (I don't mean necessarily in this PR).

adriaanm:
I can't figure out how to add a test to our build that checks that `jdeps -s -P build/pack/lib/scala-{library,reflect}.jar  | grep -v build/pack | perl -pe 's/.*\((.*)\)$/$1/' | sort | uniq` outputs (only) `compact1`. @szeiger, could you take a stab at that please?

So far, I have something like:
```scala
lazy val jdeps = taskKey[Unit]("Run jdeps to check dependencies")
lazy val restrictDeps = project
  .dependsOn(library, reflect)
  .settings(commonSettings)
  .settings(disableDocs)
  .settings(disablePublishing)
  .settings(
    Keys.test in Test := (Keys.test in Test).dependsOn(jdeps).value,
    jdeps := {
      import Process._
      val profilePart = ".*\\((.*)\\)$".r
      val jdepsOut = s"jdeps -s -P ${(packageBin in library in Compile).value.getPath} ${(packageBin in reflect in Compile).value.getPath}".lines_!
      val profiles = jdepsOut.flatMap(line => line.split(" -> ", 2) match { case Array(_, profilePart(profile)) => Some(profile) case s => None }).toSet
      if (profiles != Set("compact1")) streams.value.log.error("Detected dependency outside of compact1:\n"+jdepsOut.mkString("\n")) // TODO how to report error?
    }
  )
```

Not sure how to run this as a test and report the error

dwijnand:
Task failure is reported by throwing an exception.

szeiger:
I suggest the following:

```scala
    jdeps := {
      val profilePart = ".*\\((.*)\\)$".r
      val jdepsOut = Process("jdeps", Seq("-s", "-P", (packageBin in library in Compile).value.getPath, (packageBin in reflect in Compile).value.getPath)).lines_!
      val profiles = jdepsOut.flatMap(line => line.split(" -> ", 2) match { case Array(_, profilePart(profile)) => Some(profile) case s => None }).toSet
      if (profiles != Set("compact1"))
        throw new RuntimeException("Detected dependency outside of compact1:\n"+jdepsOut.mkString("\n"))
    },
```

Errors are reported as Exceptions. I've also changed the `Process` call to avoid command line parsing. The other version would fail with spaces in the base directory. IMHO this is bad API design in sbt: doing the wrong thing is easier than doing the right thing.

I don't see any reason for the special `restrictDeps` project. I simply put it into `root` for testing. I'd rather call this kind of "extra" test from `testAll` instead of running it with the standard `test` command.

@dwijnand Does sbt have any code for localizing `javac` that could be co-opted for `jdeps` or do you simply call whatever `javac` is on the path?

dwijnand:
By default it uses what it finds on the PATH, with support for a custom `javaHome` setting. But I wouldn't bother getting fancy until you need it.

adriaanm:
I won't have time to implement the test for this. Tracking as https://github.com/scala/scala-dev/issues/460

PR #6166, opened 2017-11-07
Fix glb of deeply structural types.

keikonakata:
- Perform `dealias` when calling glb0
- Don't decrement the depth when `proto. isTerm` in glb1. (Is this potentially dangerous?)
- Increase `globalGlbLimit`

Fixes  [scala/bug#10592](https://github.com/scala/bug/issues/10592).

According to my experiments, `globalGlbLimit` has to increase in proportion to the depth of nesting. The value `8` in this PR is chosen arbitrary (according to our use cases) and the supplemented positive and negative tests delineate the limit. 
Alternatively one may compute the nesting depth, for instance, on entering `glb1` -- this could be a less ad hoc and arguably more complicated solution (but I did not implement this alternative solution).

### DIFF
--- a/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
+++ b/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
@@ -415,7 +415,7 @@ private[internal] trait GlbLubs {
     *  If the recursion is broken, no member is added to the glb.
     */
   private var globalGlbDepth = Depth.Zero
-  private final val globalGlbLimit = Depth(2)
+  private final val globalGlbLimit = Depth(8)
 
   /** The greatest lower bound of a list of types (as determined by `<:<`). */
   def glb(ts: List[Type]): Type = elimSuper(ts) match {
@@ -496,7 +496,7 @@ private[internal] trait GlbLubs {
               val symtypes = syms map glbThisType.memberInfo
               assert(!symtypes.isEmpty)
               proto.cloneSymbol(glbRefined.typeSymbol).setInfoOwnerAdjusted(
-                if (proto.isTerm) glb(symtypes, depth.decr)
+                if (proto.isTerm) glb(symtypes, depth)
                 else {
                   def isTypeBound(tp: Type) = tp match {
                     case TypeBounds(_, _) => true
@@ -544,7 +544,7 @@ private[internal] trait GlbLubs {
     }
     // if (settings.debug.value) { println(indent + "glb of " + ts + " at depth "+depth); indent = indent + "  " } //DEBUG
     if (StatisticsStatics.areSomeColdStatsEnabled) statistics.incCounter(nestedLubCount)
-    glb0(ts)
+    glb0(ts.map(_.dealias))
     // if (settings.debug.value) { indent = indent.substring(0, indent.length() - 2); log(indent + "glb of " + ts + " is " + res) }//DEBUG
   }
 
--- a/test/files/neg/t10592.check
+++ b/test/files/neg/t10592.check
@@ -1,0 +1,13 @@
+t10592.scala:24: error: type mismatch;
+ found   : O.U[O.A]
+    (which expands to)  O.U[AnyRef{def X: AnyRef{def X: AnyRef{def X: AnyRef{def X: AnyRef{def X: AnyRef{def X: AnyRef{def X: AnyRef{def a: Int}}}}}}}}]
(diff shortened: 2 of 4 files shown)


keikonakata:
I did sign the CLA and the legal authority within my employer approved.

adriaanm:
I apologize we haven't provided you with faster feedback. We've been a bit overloaded recently. We'll do our best to take a look soon, though!

As a first question, could you elaborate a bit on the reasoning behind not decrementing the depth when `proto.isTerm`?

keikonakata:
I noticed that if `proto.isTerm` is decremented, we will immediately hit [line 486](https://github.com/scala/scala/blob/2.13.x/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala#L486), returning `AnyRef` and discarding refinement body.

The `depth` is used as a termination measure, right? How the `depth` is chosen/computed initially (e.g., the size of the shape of input types, whatever the size means, etc)?

Atry:
`glb` ignores members in classes and traits.

``` scala
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_152).

scala> :power

scala> trait Foo {
     |   def myMethod: Option[String]
     | }
defined trait Foo

scala> trait Bar {
     |   def myMethod: Option[AnyRef]
     | }
defined trait Bar

scala> glb(List(typeOf[Foo], typeOf[Bar]))
res0: $r.intp.global.Type = Foo with Bar

scala> res0.member(TermName("myMethod"))
res1: $r.intp.global.Symbol = method myMethod

scala> res1.info
res2: $r.intp.global.Type = => Option[AnyRef]

```
I expect the return type of `res1.info` is `Option[String]` instead of `Option[AnyRef]`.

It would be better if `glb` returns the refinement type `Foo with Bar { def def myMethod: Option[String] }`.

keikonakata:
I have observed @Atry's issue, but it is out of the scope of this PR. 

This PR only recursively merges structural members of refinement types on computing `glb`, but it does not mean to change existing behaviour with respect to non-structural members.

keikonakata:
@Atry By the way, if you change the order of `Foo` and `Bar` when calling `glb`, you get a different result:
```
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).

scala> :power

scala> trait Foo { def myMethod: Option[String] }
defined trait Foo

scala> trait Bar { def myMethod: Option[AnyRef] }
defined trait Bar

scala> glb(List(typeOf[Bar], typeOf[Foo]))
res0: $r.intp.global.Type = Bar with Foo

scala> res0.member(TermName("myMethod"))
res1: $r.intp.global.Symbol = method myMethod

scala> res1.info
res2: $r.intp.global.Type = => Option[String]
```
It is a bit confusing that I expect  that`glb` to be symmetric (i.e., the argument order should not matter) and `with` is asymmetric, nevertheless the former uses the latter.

Atry:
`with` is not symmetric.
What symmetric is [intersection type](http://dotty.epfl.ch/docs/reference/intersection-types.html).

keikonakata:
Yes, `with` is asymmetric, so it feels incorrect (in some situations) that `glb` uses `with`. (`glb` stands for greatest lower bounds, right?)

This is not exactly the issue that is addressed by this PR, though it would be great if it is fixed. (My gut feeling is that it would require more substantial changes than my current PR.)

SethTisue:
this and #6170 should probably be reviewed together?

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -544,7 +544,7 @@ private[internal] trait GlbLubs {
     }
     // if (settings.debug.value) { println(indent + "glb of " + ts + " at depth "+depth); indent = indent + "  " } //DEBUG
     if (StatisticsStatics.areSomeColdStatsEnabled) statistics.incCounter(nestedLubCount)
-    glb0(ts)
+    glb0(ts.map(_.dealias))
```
adriaanm, line 547: I've been poking around a bit here. The dealias is needed because `typeDepth` considers type aliases to have depth 1, regardless of their RHS. Trying to figure out whether we can avoid this `dealias`, but no luck so far

adriaanm:
I'm afraid we won't be able to include this change in the 2.12 series. I'm worried about termination issues due to the dropped `.decr`, and the arbitrary constant bump from `2` -> `8`. There's a lot of exciting stuff coming in Scala 3 that will have more principled solutions for the use case that prompted this PR.

PR #6168, opened 2017-11-08
Fix lub of structural types

keikonakata:
- don't decrement the depth when proto.isTerm
- add refinement symbols which are refined by every types

Fixes [scala/bug#10593](https://github.com/scala/bug/issues/10593)

This PR does not enable the following code, which I would consider well-typed, hence it is consciously omitted from the test.
```
type A = AnyRef { def x: Int; def a : Int }
type B = AnyRef { def x: Int; def b : Int }
type C = AnyRef { def x: Int }

class U[+X]
def f_bound[X <: A, Y <: B](a: U[X], b: U[Y]): Seq[U[C]] = Seq(a, b)
```


keikonakata:
I am sorry, I have to take a look at the failed test.

SethTisue:
replaced by #6170

PR #6169, opened 2017-11-08
Fixes #10587 by removing adaptations to accommodate for <JDK8 FJP

viktorklang:
**PR for discussing, do not merge yet.**

Should address https://github.com/scala/bug/issues/10587

Rationale: Prior-to-jdk8 FJP wasn't being smart about when to fork() and when to submit to the global submission system, so the existing scala.concurrent-code (the default EC implementation specifically) worked around this by wrapping the ForkJoinTask for Runnables and injecting the UncaughtExceptionHandler 

@rkuhn @ktoso Please chime in if you suspect that this change is too bold.

### DIFF
--- a/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
+++ b/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
@@ -120,32 +120,7 @@ private[concurrent] object ExecutionContextImpl {
                                                                       prefix = "scala-execution-context-global",
                                                                       uncaught = uncaughtExceptionHandler)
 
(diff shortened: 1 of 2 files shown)


adriaanm:
Would be good to get a review -- ping @ktoso. If not in the next few days, we'll postpone this to M4.

SethTisue:
ping @ktoso @rkuhn?

ktoso:
Would have to benchmark this. Will try to squeeze it in this week

adriaanm:
I'm closing this PR because it has been dormant for a while now. Happy to receive a resubmit!

PR #6170, opened 2017-11-08
Fix lub of structural types

keikonakata:
1. add refinement symbols which are refined by every types
2. introduce glabalLubDepth and globalLubLimit as in glb
   - Without this (due to the change in 3.), test/files/pos/lub-from-hell.scala will loop.
   - According to my experiments, `globalLubLimit` has to increase in proportion to the depth of 
      nesting.  The value of `8` is chosen arbitrary.
3. don't decrement the depth when proto.isTerm

The change forces computation of lub to unfold (recursive) types deeper, which I think results in different outputs in test/files/neg/lub-from-hell-2.scala. (But please correct me, if I'm supposing wrong.)

Fixes [scala/bug#10593](https://github.com/scala/bug/issues/10593)

This PR does not enable the following code, which I would consider well-typed.
```
type A = AnyRef { def x: Int; def a : Int }
type B = AnyRef { def x: Int; def b : Int }
type C = AnyRef { def x: Int }

class U[+X]
def f_bound[X <: A, Y <: B](a: U[X], b: U[Y]): Seq[U[C]] = Seq(a, b)
```


SethTisue:
this PR seems to have fallen through the cracks. can someone review it...?

@keikonakata since quite some time has passed (sorry about that), would you mind rebasing against current 2.12.x?

SethTisue:
this and #6166 should probably be reviewed together?

lrytz:
lub / glb are very subtle operations, and at the same time at the core of the type system, so i re-assign this to 2.12.6, we cannot review it in time.

dwijnand:
Something's happened to the tooling here..

/sync

SethTisue:
this probably needs a rebase in order to get PR-validated on the current Jenkins infra

adriaanm:
I'm afraid we won't be able to include this change in the 2.12 series, for similar reasons to #6166. There's a lot of exciting stuff coming in Scala 3 that will have more principled solutions for the use case that prompted this PR.

PR #6171, opened 2017-11-09
Backport of fix to 10587

viktorklang:
Do not merge yet.

### DIFF
--- a/src/library/mima-filters/2.12.0.backwards.excludes
+++ b/src/library/mima-filters/2.12.0.backwards.excludes
@@ -6,4 +6,6 @@ ProblemFilters.exclude[DirectMissingMethodProblem]("scala.collection.immutable.V
 ProblemFilters.exclude[DirectMissingMethodProblem]("scala.collection.mutable.OpenHashMap.nextPositivePowerOfTwo")
 ProblemFilters.exclude[DirectMissingMethodProblem]("scala.collection.mutable.HashTable.powerOfTwo")
 
+ProblemFilters.exclude[MissingClassProblem]("scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask")
+
 ProblemFilters.exclude[DirectMissingMethodProblem]("scala.sys.process.ProcessImpl#CompoundProcess.getExitValue")
--- a/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
+++ b/src/library/scala/concurrent/impl/ExecutionContextImpl.scala
@@ -120,32 +120,7 @@ private[concurrent] object ExecutionContextImpl {
                                                                       prefix = "scala-execution-context-global",
                                                                       uncaught = uncaughtExceptionHandler)
 
-    new ForkJoinPool(desiredParallelism, threadFactory, uncaughtExceptionHandler, true) {
-      override def execute(runnable: Runnable): Unit = {
-        val fjt: ForkJoinTask[_] = runnable match {
-          case t: ForkJoinTask[_] => t
-          case r                  => new ExecutionContextImpl.AdaptedForkJoinTask(r)
-        }
-        Thread.currentThread match {
-          case fjw: ForkJoinWorkerThread if fjw.getPool eq this => fjt.fork()
-          case _                                                => super.execute(fjt)
-        }
-      }
-    }
-  }
-
-  final class AdaptedForkJoinTask(runnable: Runnable) extends ForkJoinTask[Unit] {
-    final override def setRawResult(u: Unit): Unit = ()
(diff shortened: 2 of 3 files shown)


SethTisue:
```
[info] scala-library: found 1 potential binary incompatibilities while checking against org.scala-lang:scala-library:2.12.0  (filtered 40)
[error]  * class scala.concurrent.impl.ExecutionContextImpl#AdaptedForkJoinTask does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask")
```

the whole `scala.concurrent.impl` package could just be exempted from MiMa, I imagine...?

viktorklang:
@SethTisue The entire *impl cannot (unfortunately) be exempted because the scala-java-8-compat jar messes around with the DefaultPromise.

I'll add an exemption to the removed class.

viktorklang:
@ktoso @rkuhn Sorry for pinging you, just need a sanity check for this :)

ktoso:
No idea, could you explain the rationale and show why we're safe to revert this now, what's the assumption on JDK version?
Would have to release this and test against akka to be sure it does not cause regressions -- perhaps you can do that?

viktorklang:
@ktoso Assumption of JDK8 (Scala 2.12+). I believe historically FJP wouldn't escalate exceptions on wrapped Runnables to the UncaughtExceptionHandler of the executing thread, which is why we wrapped it ourselves and delegated. The only observable difference that I can think of right now is that uncaught exceptions will now lead to worker death (which arguably is more correct).

retronym:
@ktoso @viktorklang What's the status of? Could you summarize the risks/benefits?

viktorklang:
@retronym AFAIK: Only binary compatibility problem would be if anyone is instantiating this class (AdaptedForkJoinTask) themselves, which does seem unlikely given its `private[concurrent]` visibility. Semantically, the change is that now we're going to rely more on the FJP implementation's termination semantics, which we overrode with the AdaptedForkJoinTask, to intercept uncaught exceptions, which at least IIRC on Java7 didn't behave the same.

What does the community build say about this change?

SethTisue:
> What does the community build say about this change?

there is some breakage in the 2.12 build I need to fix first, but once that's done I'll return to this ticket and trigger a test run.

SethTisue:
@viktorklang could you rebase this against current 2.12.x? I did a community build run but I got some strange failures I don't think are related to your changes. rather than investigate, I'd like to try again against fresher code.

lrytz:
@SethTisue rebased and push-f-d to viktor's branch. @viktorklang assuming the community build is fine, can this be merged?

viktorklang:
@lrytz AFAIK: yes.

SethTisue:
community build queued: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2623/ (404 til Jenkins catches up)

lrytz:
Compared to the previous run, scala-async and cats-effect failed.

scala-async shows a number of errors like this:

```
[scala-async] [info] Test scala.async.TreeInterrogation.a minimal set of vals are lifted to vars started
[scala-async] [error] Test scala.async.TreeInterrogation.a minimal set of vals are lifted to vars failed: java.lang.RuntimeException: output directory /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.10/project-builds/scala-async-df3d202214fa948565640ce4232068882df4314c/target/scala-2.12.5-SNAPSHOT/classes does not exist., took 1.116 sec
[scala-async] [error]     at scala.sys.package$.error(package.scala:27)
[scala-async] [error]     at scala.async.package$.toolboxClasspath(package.scala:80)
[scala-async] [error]     at scala.async.TreeInterrogation.a minimal set of vals are lifted to vars(TreeInterrogation.scala:16)
```
Not sure what to make from this, but it seems unrelated.

For cats-effect

```
[cats-effect] [error] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.10/project-builds/cats-effect-7e331c3aae06fb0e9bb84cead7cad7c024916bdc/laws/shared/src/test/scala/cats/effect/BaseTestsSuite.scala:38:9: local val ruleSet in method checkAllAsync is never used
[cats-effect] [error]     val ruleSet = f(context)
[cats-effect] [error]         ^
[cats-effect] [error] one error found
```

Not sure why, it used the same revision as in the previous build.

SethTisue:
@lrytz the community build failures seem unrelated, so I think we can just hit "merge" on this?

lrytz:
👍

PR #6172, opened 2017-11-10
Improve Scaladoc

Atry:


adriaanm:
"Scaladoc" is a bit ambiguous in this project since it could refer to the tool as well, for next time, I suggest: "Improve  comments".

PR #6175, opened 2017-11-16
Keep path-dependent types when typing HOF args in overload resolution

szeiger:
Here's an attempt at fixing https://github.com/scala/bug/issues/10608:
- Keep path-dependent types by starting with `pre.memberType(alt)` instead of `alt.info.paramTypes` (which has already lost them)
- Don't drop them afterwards in `funArgTypes` by not running them through `approximateAbstracts`

I assume there was a good reason for approximating but no test case fails when we don't do it, so I'm not sure what corner cases I may have missed.


adriaanm (approved):
Good catch. Could use some more explanation: the problem is that the type parameter that occurs in the type of a prefix is approximated to wildcard, which then keeps the types from matching.

PR #6176, opened 2017-11-16
Fix corner case in preErase's treatment of Apply trees.

hrhino:
`preEraseApply` can return another `Apply` that's a subtree of the tree being transformed; if it does, then that tree needs `preErase`d itself, so just send it through the wash one more time.

fixes scala/bug#10611.

### DIFF
--- a/src/compiler/scala/tools/nsc/transform/Erasure.scala
+++ b/src/compiler/scala/tools/nsc/transform/Erasure.scala
@@ -1268,6 +1268,11 @@ abstract class Erasure extends InfoTransform
               finally tpt setType specialErasure(tree1.symbol)(tree1.symbol.tpe).resultType
             case ApplyDynamic(qual, Literal(Constant(bootstrapMethodRef: Symbol)) :: _) =>
               tree
+            case _: Apply if tree1 ne tree =>
+              /* some Apply trees get replaced (in `preEraseApply`) with one of
+               * their subtrees, which needs to be `preErase`d in its entirety,
+               * not just recursed over by super.transform(). */
+              transform(tree1)
             case _ =>
               super.transform(tree1).clearType()
           }
--- a/test/files/run/t10611.scala
+++ b/test/files/run/t10611.scala
@@ -1,0 +1,7 @@
+object Test extends App {
+  /* crashes at jvm with AssertionError */
+  def q[T: reflect.ClassTag] = Array.empty[T].length.##
+
+  /* crashes at runtime with NoSuchMethodError */
+  this.##.##
+}


lrytz (approved):

som-snytt:
Thanks @lrytz .  I took a look yesterday, and I found the adhoc recursion awkward. I'll add a note if I see a better way. (I thought I might have the weekend before the pros stepped in. Maybe Lukas realized Met Opera's Exterminating Angel is live in HD this weekend.)

lrytz:
I didn't merge it yet because I thought maybe someone else might want to take a look, so thanks!

hrhino:
Thanks for the opinion, @som-snytt. I'm inclined to agree with you that it's a little ad-hoc; there are a few other ways of going about it. The other one I was considering is (in `preErase.transform`):
```scala
        if (tree.symbol == ArrayClass && !tree.isType) tree
        else {
          val tree1 = preErase(tree)
          tree1 match {
            case TypeApply(fun, targs @ List(targ)) if (fun.symbol == Any_asInstanceOf  || fun.symbol == Object_synchronized) && targ.tpe == UnitTpe =>
              // scala/bug#9066 prevent transforming `o.asInstanceOf[Unit]` to `o.asInstanceOf[BoxedUnit]`.
              // adaptMember will then replace the call by a reference to BoxedUnit.UNIT.
              treeCopy.TypeApply(tree1, transform(fun), targs).clearType()
            case EmptyTree | TypeTree() =>
              tree1 setType specialScalaErasure(tree1.tpe)
            case ArrayValue(elemtpt, trees) =>
              treeCopy.ArrayValue(
                tree1, elemtpt setType specialScalaErasure.applyInArray(elemtpt.tpe), trees map transform).clearType()
            case DefDef(_, _, _, _, tpt, _) =>
              // TODO: move this in some post-processing transform in the fields phase?
              if (fields.symbolAnnotationsTargetFieldAndGetter(tree.symbol))
                fields.dropFieldAnnotationsFromGetter(tree.symbol)

              try super.transform(tree1).clearType()
              finally tpt setType specialErasure(tree1.symbol)(tree1.symbol.tpe).resultType
            case ApplyDynamic(qual, Literal(Constant(bootstrapMethodRef: Symbol)) :: _) =>
              tree
+           case Apply(_, _) if tree1 ne tree =>
+             transform(tree1)
            case _ =>
              super.transform(tree1).clearType()
          }
 ```

I'd justify that by pointing out that `Apply` trees can turn into other `Apply` trees, and since we treat such nodes quite heavily in `preErase`, it's important that they don't slip under the radar. I like this solution for its parsimony, but I feel a little weird about the `ne` (trees are reference-compared elsewhere, but they're still mutable, so...). I count a handful of other places where wiser people than I do a similar thing, though.

WDYT?

_EDIT_: also, I tagged you in here b/c you showed up on the ticket; I hope it didn't seem like I'm demanding your time. Thanks!

File: src/compiler/scala/tools/nsc/transform/Erasure.scala [hunk 1]
```diff
@@ -1268,6 +1268,8 @@ abstract class Erasure extends InfoTransform
               finally tpt setType specialErasure(tree1.symbol)(tree1.symbol.tpe).resultType
             case ApplyDynamic(qual, Literal(Constant(bootstrapMethodRef: Symbol)) :: _) =>
               tree
+            case Apply(_, _) if tree1 ne tree =>
```
som-snytt: `case _: Apply`, just to say something.

som-snytt (approved):
That's so simple. I was thinking along similar lines, but I don't quite grasp the structure of the pre-pre and then the pre- and then the re-typecheck. Probably if I understood the comment about typechecking the "staticsCall", I'd understand better why things have to happen in the order they happen in. I was inclined to move the `##` code out of pre-pre: you transform an Apply then do the qual and args as usual. That seems regular to me, but maybe I just need to spend some time with it. I touched something and got `BoxAndUnbox` in a stack trace.

But I tried out this minimal change and couldn't break it.

hrhino:
Yeah, I think I just needed to sleep on it for one more night. Thanks for calling a hack a hack.

Replacing this fix with 
```scala
            case Apply(fun @ Select(qual, _), Nil) if poundPoundMethods.contains(fun.symbol) =>
              transform(postPreEraseHash(qual))
```
(where `postPreEraseHash` is the relevant code in `preEraseHash`) seems to work decently, but it's basically the same code moved into a different part of the traversal. You've still gotta call `transform` rather than `super.transform` like that, though. 

`case _: Apply` isn't allowed by Scalazzi, of course.

src/compiler/scala/tools/nsc/transform/Erasure.scala
```diff
@@ -1268,6 +1268,8 @@ abstract class Erasure extends InfoTransform
               finally tpt setType specialErasure(tree1.symbol)(tree1.symbol.tpe).resultType
             case ApplyDynamic(qual, Literal(Constant(bootstrapMethodRef: Symbol)) :: _) =>
               tree
+            case _: Apply if tree1 ne tree =>
+              transform(tree1)
```
lrytz: So this calls `preErase` recursively until it stabilizes. I'm not sure if it's a bit too clever for my taste.. There should at least be some comment.

hrhino: Mmhm. So, there are [a few](https://github.com/scala/scala/blob/v2.12.4/src/compiler/scala/tools/nsc/transform/CleanUp.scala#L87-L89) [places](https://github.com/scala/scala/blob/v2.12.4/src/compiler/scala/tools/nsc/transform/CleanUp.scala#L365-L367) [already](https://github.com/scala/scala/blob/v2.12.4/src/compiler/scala/tools/nsc/transform/Fields.scala#L726) where we recurse on a tree that's either freshly made, or a sub-tree of the tree that we're currently transforming.

My double-original idea, that didn't see the light of day, was something like
```diff
             else if (args.isEmpty && interceptedMethods(fn.symbol)) {
               if (poundPoundMethods.contains(fn.symbol)) {
+                val qualSym = qual.tpe.typeSymbol
+                val qual1 = transform(qual)
                 // This is unattractive, but without it we crash here on ().## because after
                 // erasure the ScalaRunTime.hash overload goes from Unit => Int to BoxedUnit => Int.
                 // This must be because some earlier transformation is being skipped on ##, but so
                 // far I don't know what.  For null we now define null.## == 0.
                 def staticsCall(methodName: TermName): Tree = {
-                  val newTree = gen.mkMethodCall(RuntimeStaticsModule, methodName, qual :: Nil)
+                  val newTree = gen.mkMethodCall(RuntimeStaticsModule, methodName, qual1 :: Nil)
                   global.typer.typed(newTree)
                 }

-                qual.tpe.typeSymbol match {
+                qualSym match {
                   case UnitClass | NullClass                    => LIT(0)
-                  case IntClass                                 => qual
-                  case s @ (ShortClass | ByteClass | CharClass) => numericConversion(qual, s)
-                  case BooleanClass                             => If(qual, LIT(true.##), LIT(false.##))
+                  case IntClass                                 => qual1
+                  case s @ (ShortClass | ByteClass | CharClass) => numericConversion(qual1, s)
+                  case BooleanClass                             => If(qual1, LIT(true.##), LIT(false.##))
                   case LongClass                                => staticsCall(nme.longHash)
                   case FloatClass                               => staticsCall(nme.floatHash)
                   case DoubleClass                              => staticsCall(nme.doubleHash)
```
wherein we fall somewhat in line with the other transformations, in recursing manually over a sub-tree. However, since this is buried two-method-calls-deep from the `traverse` method, there's not any way I know of to say "no, this tree's fine, I took care of it already" like we can elsewhere, so we'll wind up going over this tree again in the `super.traverse` call. I don't see it being harmful at this exact moment but it introduces a level of fragility. The other version of this, deemed ad-hoc, recursed only over `preEraseApply`, since the scope is smaller and therefore less code for me to reason about.

I tried very hard to justify the use of recursion here guarded only by `tree1 ne tree`; nowhere do we return an `Apply` wrapping the `Apply` that we just made in a way that will make it get changed and re-recurse (other than in cases like `##.##.##`). It's a bit strange, though. I'll gladly add a comment, I guess.

som-snytt: I admire the efforts, and I'll try to donate some more time to see if even I can reason about one of the solutions.

retronym (approved):

PR #1109, opened 2012-08-09
Warn when Any or AnyVal is inferred.

paulp:
For the very small price of annotating types as Any/AnyVal in those
cases where we wish to use them, we can obtain useful warnings.
I made trunk clean against this warning and found several bugs
or at least suboptimalities in the process.

I put the warning behind -Xlint for the moment, but I think this
belongs on by default, even for this alone:

  scala> List(1, 2, 3) contains "a"
  <console>:8: warning: a type was inferred to be `Any`; this may indicate a programming error.
                List(1, 2, 3) contains "a"
                                       ^
  res0: Boolean = false

Or this punishment meted out by SI-4042:

  scala> 1l to 5l contains 5
  <console>:8: warning: a type was inferred to be `AnyVal`; this may indicate a programming error.
                1l to 5l contains 5
                                  ^
  res0: Boolean = false

A different situation where this arises, which I have seen variations
of many times:

  scala> class A[T](default: T) {
    def get(x: => Option[T]) = x getOrElse Some(default)
  }
  <console>:7: warning: a type was inferred to be `Any`; this may indicate a programming error.
         class A[T](default: T) { def get(x: => Option[T]) = x getOrElse Some(default) }
                                                                             ^
  // Oops, this was what I meant
  scala> class A[T](default: T) {
    def get(x: => Option[T]) = x getOrElse default
  }
  defined class A

Harder to avoid spurious warnings when "Object" is inferred.

PR #6178, opened 2017-11-16
Warn when Object/AnyRef is inferred.

julienrf:
Extends -Xlint:warn-any to also emit a warning when Object/AnyRef
is inferred.


hrhino:
Is this intended?

```scala
Welcome to Scala 2.13.0-20171116-213949-1e28571 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> class A; class B
defined class A
defined class B

scala> val as = List(new A)
as: List[A] = List(A@14b95942)

scala> val b: AnyRef = new B
b: AnyRef = B@4fa5cc73

scala> as contains new B
          ^
       warning: a type was inferred to be `Object`; this may indicate a programming error.
res0: Boolean = false

scala> as contains b
res1: Boolean = false

scala>
```

adriaanm (approved):
Thanks!

SethTisue:
Note that this is a sequel to https://github.com/scala/scala/pull/1109

PR #6181, opened 2017-11-19
fix currentRun.compiles(Symbol)

mkeskells:
before GenBCode the owner of classes change, so the `enclosingTopLevelClass` misses the top level class (probably changed in delambify) 
this allows the `compiles` to check the `originalOwner`, which seems to work in my test cases

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1330,7 +1330,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     def compiles(sym: Symbol): Boolean =
       if (sym == NoSymbol) false
       else if (symSource.isDefinedAt(sym)) true
-      else if (!sym.isTopLevel) compiles(sym.enclosingTopLevelClassOrDummy)
+      else if (!sym.isTopLevel) compiles(sym.originalEnclosingTopLevelClassOrDummy)
       else if (sym.isModuleClass) compiles(sym.sourceModule)
       else false
 
--- a/src/compiler/scala/tools/nsc/transform/Flatten.scala
+++ b/src/compiler/scala/tools/nsc/transform/Flatten.scala
@@ -57,8 +57,9 @@ abstract class Flatten extends InfoTransform {
   private val flattened = new TypeMap {
     def apply(tp: Type): Type = tp match {
       case TypeRef(pre, sym, args) if isFlattenablePrefix(pre) =>
-        assert(args.isEmpty && sym.enclosingTopLevelClass != NoSymbol, sym.ownerChain)
-        typeRef(sym.enclosingTopLevelClass.owner.thisType, sym, Nil)
+        val top = sym.enclosingTopLevelClass
+        assert(args.isEmpty && top != NoSymbol, sym.ownerChain)
+        typeRef(top.owner.thisType, sym, Nil)
       case ClassInfoType(parents, decls, clazz) =>
         var parents1 = parents
         val decls1 = scopeTransform(clazz) {
--- a/src/reflect/scala/reflect/internal/PrivateWithin.scala
+++ b/src/reflect/scala/reflect/internal/PrivateWithin.scala
@@ -20,8 +20,9 @@ trait PrivateWithin {
   // protected in java means package protected. #3946
   // See ticket #1687 for an example of when the enclosing top level class is NoSymbol;
   // it apparently occurs when processing v45.3 bytecode.
-  def setPackageAccessBoundary(sym: Symbol): Symbol = (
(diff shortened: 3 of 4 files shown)


mkeskells:
@lrytz @adriaanm as discussed via email

mkeskells:
an alternate implementation for compiles might be
```
def compiles(sym: Symbol): Boolean =compiledFiles contains sym.pos.source.file
```
not sure why this would not work, and it seems simpler and faster (not yet benchmarked though)

retronym:
I don't believe that comparing `SourceFile` will have the right semantics, as `SourceFile.equals` just compares file paths.

The test below shows this problem, the symbol's positon's sourcefile from two runs are considered equal.

```
scala> import scala.tools.nsc._, scala.tools.nsc.settings._
import scala.tools.nsc._
import scala.tools.nsc.settings._

scala> val g = new Global(new Settings)
g: scala.tools.nsc.Global = scala.tools.nsc.Global@25a52a60

scala> g.settings.usejavacp.value = true
g.settings.usejavacp.value: Boolean = true

scala> val C_sym_1 = { val r = new g.Run; r.compileSources(file :: Nil); r1.units.toList.head.body.children.last}
C_sym_1: g.Tree =
class C extends Object {
  def <init>(): C = {
    C.super.<init>();
    ()
  }
}

scala> val C_sym_2 = { val r = new g.Run; r.compileSources(file :: Nil); r1.units.toList.head.body.children.last}
C_sym_2: g.Tree =
class C extends Object {
  def <init>(): C = {
    C.super.<init>();
    ()
  }
}

scala> C_sym_1.pos.source == C_sym_2.pos.source
res7: Boolean = true
```

mkeskells:
@retronym hadn't thought of repl...

should this PR be targeted at 2.12 or 2.13?

I would hope that we could target this at 2.12 as its needed for the GenBCode changes to correctly predict BType expansion

retronym:
Benchmarks [arriving](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?var-scalaSha=e12ac748cd%7Ccbd596339f%7Cc47ab156b1%7Cc929228698&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@). (I'm a bit cautious about all the extra map lookups, but let's see what the benchmarks say)

retronym:
I'd be happy for this to target 2.12

mkeskells:
discussing on other channels how to avoid the map lookups
maybe compiles should be 
```
    /** Does this run compile given class, module, or case factory?
      * We look up to the outermost class, and see if that is compiled, but in the later phases the owner is adjusted
      * So we need to look at the originalOwner, which is more expensive, so we have a couple of shortcuts
      * 1. if the symbol has a position then we can use that to check if the source file is compiled, but if it isnt, it may still not be compiled
      *    but it avoids to navigation in a common case
      * 2. If the owner is not top level then we can navigate there as it will still be within a compiled symbol
      *
      * This reduced to number of expensive originalOwner lookups, but at least (2) could be removed if the originalOwner becomes cheaper to lookup
      */
    // NOTE: Early initialized members temporarily typechecked before the enclosing class, see typedPrimaryConstrBody!
    //       Here we work around that wrinkle by claiming that a pre-initialized member is compiled in
    //       *every* run. This approximation works because this method is exclusively called with `this` == `currentRun`.
    def compiles(sym: Symbol): Boolean =
      if (sym == NoSymbol) false
      else if (symSource.contains(sym)) true
      else if ({
        val posFile = sym.pos.source.file
        posFile != NoAbstractFile && !compiledFiles.contains(posFile.path)
      }) false
      else if (!sym.isTopLevel) {
        var parent = sym
        while(!parent.owner.isTopLevel) parent = parent.owner
        var originalOwner = parent.originalOwner
        while(!originalOwner.isTopLevel) {
          parent = originalOwner
          originalOwner = parent.originalOwner
        }
        compiles(parent.originalEnclosingTopLevelClassOrDummy)
      }
      else if (sym.isModuleClass) compiles(sym.sourceModule)
      else false

```
and remove the method from Symbols. This is unchecked. ANy work on this would be a separate PR though. https://github.com/rorygraves/scalac_perf/issues/31

mkeskells:
changed to target 2.12

mkeskells:
@lrytz @retronym still need answers on other owner walking APIs like associatedFile

lrytz:
The PR looks fine to me in its current form, the benchmarks seem flat / within noise. The commits could be squashed into one.

mkeskells:
squashed as requested

retronym (approved):
LGTM, too. Thanks!

Jasper-M:
I think this is assigned to the wrong milestone.

PR #6124, opened 2017-10-11
Enable parallel optimizing and writing of classes by GenBCode

mkeskells:
Introduces a compiler flag `-Ybackend-parallelism N` which allows specifying the number of threads for optimizing and writing classes in parallel.

The granularity of parallelization is a compilation unit (a source file).

We measured speedup of 4-5% when compiling the scala/scala sources with 8 backend threads ([graph](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?from=1519356411351&to=1519465862679&orgId=1&var-scalaSha=9fc13560ff%7C0b6d7c317d&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench@scalabench@)).

rework ClassBType to enable parallelism, and move logic in the companion
rewrite ClasfileWriter, specialising for JAR/dir, and providing wrappers for the less common cases
rework directory classfile writing to be async NIO based. 
Async the directory creation
make PerRunInit theadsafe

BackendUtils - make some data structure/APIs theadsafe (indyLambdaMethods)
add extra parameter -YmaxAdditionalWriterThreads .. "maximum additional threads to write class files"
add a class handler as a delegate that allows the minimal set in post-processing steps to be performed

PR #6182, opened 2017-11-19
improve benchmarking of multi-threaded compilation 

mkeskells:
an enabler for GenBCode (and other future) optimisations

factored out of https://github.com/scala/scala/pull/6124 to aid review


lrytz:
I'm happy to merge this without much of a review - or @retronym do you want to take a deeper look?

In general, it would be nice to have some documentation how to use (how you are using) the classes in nsc.profile, it might be helpful to other parties working on performance.

mkeskells:
We can add some docs on this. They are in wiki atm on other projects but don't want it to sound like an advert for another site

Would you prefer to see details documented usage in the code comments or a brief summary and a reference to an external wiki

I would prefer the latter but it's your call

Either way can this be a separate PR as we are upgrading the tooling in the next couple of weeks

retronym (approved):

src/compiler/scala/tools/nsc/profile/Profiler.scala
```diff
@@ -120,6 +89,10 @@ private [profile] object RealProfiler {
 }
 
 private [profile] class RealProfiler(reporter : ProfileReporter, val settings: Settings) extends Profiler with NotificationListener {
+  def completeBackground(threadRange: ProfileRange): Unit = {
+    reporter.reportBackground(this, threadRange)
```
lrytz, line 93: @mkeskells looking at this while reviewing #6124. Is this thread-safe?

PR #6183, opened 2017-11-19
use java concurrent hash map for speed in GenBCode

mkeskells:
minor improvement factored out of https://github.com/scala/scala/pull/6124 to aid review

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -6,7 +6,6 @@
 package scala.tools.nsc
 package backend.jvm
 
-import scala.collection.concurrent.TrieMap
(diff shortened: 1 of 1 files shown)


retronym:
I've done something similar in a WIP performance branch: https://github.com/retronym/scala/commit/a4f7018418956797297ea05fdeeabbc87629a2cc

Your patch looks simpler than mine. I tried to use the Java API directly rather than involving ` `JavaConverters`, but that involved some extra work to make plug it into the per-run cache clearing mechanism.

retronym:
Another variation I considered, but haven't implemented, is using a non-concurrent map if the backend is run in single threaded mode (currently the only mode!)

mkeskells:
I know it similar - just trying to get the review of the parallelisation PR moving by moving bits out into separate PRs

retronym (approved):

PR #2601, opened 2013-05-27
SI-7516 Revert "SI-7234 Make named args play nice w. depmet types"

retronym:
This reverts commit 83c9c764b528a7a1c1d39c480d22c8e3a71d5a58.

The tests are shunted to 'pending'.

Whitespace fixes and identifier renaming were not reverted.
Why revert this seemingly innocous commit? 83c9c76 generates a ValDef whose
tpt TypeTree has no original; this contains a reference to the symbol for `d`.
resetAttrs and the retypecheck assigns a new symbol for d and leaves a the
reference to the prior symbol dangling. The real bug is the resetAttrs concept.

Review by @paulp

PR #6186, opened 2017-11-21
Fixup symbol ownership in eta expansion

retronym:
The fix for scala/bug#6274 did this selectively when function symbols
were introduced, but we really ought to insert the eta$n temp vals into
the owner chain of all lifted expressions.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/EtaExpansion.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/EtaExpansion.scala
@@ -56,16 +56,22 @@ trait EtaExpansion { self: Analyzer =>
         else {
           val vname: Name = freshName()
           // Problem with ticket #2351 here
+          val valSym = typer.context.owner.newValue(vname.toTermName, tree.pos.focus, SYNTHETIC)
           defs += atPos(tree.pos) {
             val rhs = if (byName) {
-              val res = typer.typed(Function(List(), tree))
-              new ChangeOwnerTraverser(typer.context.owner, res.symbol) traverse tree // scala/bug#6274
-              res
-            } else tree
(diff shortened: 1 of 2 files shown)


retronym:
The risk of this change here is that we run into the same problems that led to the reversion (#2601) of 83c9c764b528a7a1c1d39c480d22c8e3a71d5a58.

src/compiler/scala/tools/nsc/typechecker/EtaExpansion.scala
```diff
@@ -55,16 +55,22 @@ trait EtaExpansion { self: Analyzer =>
...
+              val funSym = valSym.newAnonymousFunctionValue(tree.pos.focus)
+              val tree1 = tree.changeOwner(typer.context.owner -> funSym)
+              val funType = definitions.functionType(Nil, tree1.tpe)
+              funSym.setInfo(funType)
+              Function(List(), tree1).setSymbol(funSym).setType(funType)
+            } else {
+              tree.changeOwner(typer.context.owner -> valSym)
+            }
+            valSym.setInfo(rhs.tpe)
+            ValDef(valSym, rhs)
```
lrytz: looks good; `def etaExpand` could now take a `owner: Symbol` parameter instead of the full `typer: Typer`.

adriaanm: Pushed a commit with that refactoring.

adriaanm (approved):

SethTisue:
assuming not critical for M5. retarget if you disagree

adriaanm:
I think this one is ready to go, actually. If the resetAttrs interaction proves problematic, we can refine.

PR #6187, opened 2017-11-21
Document stability of sortBy

DanMoss:
I've assumed that the stability of `sortBy` is guaranteed due to such a guarantee existing in the documentation for `sorted` and `sortWith`. I was hoping this could be confirmed explicitly by documenting this information. alexnixon previously raised this in issue [10280](https://github.com/scala/bug/issues/10280).

### DIFF
--- a/src/library/scala/collection/SeqLike.scala
+++ b/src/library/scala/collection/SeqLike.scala
@@ -603,8 +603,13 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
 
   /** Sorts this $Coll according to the Ordering which results from transforming
    *  an implicitly given Ordering with a transformation function.
-   *  @see [[scala.math.Ordering]]
    *  $willNotTerminateInf
+   *
+   *  The sort is stable. That is, elements that are equal (as determined by
+   *  `ord.compare`) appear in the same order in the sorted sequence as in the original.
+   *
+   *  @see [[scala.math.Ordering]]
+   *
    *  @param   f the transformation function mapping elements
    *           to some other domain `B`.
    *  @param   ord the ordering assumed on domain `B`.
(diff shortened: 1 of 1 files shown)


DanMoss:
I've now signed the CLA.

src/library/scala/collection/SeqLike.scala
```diff
@@ -603,8 +603,13 @@ trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[
 
   /** Sorts this $Coll according to the Ordering which results from transforming
    *  an implicitly given Ordering with a transformation function.
-   *  @see [[scala.math.Ordering]]
    *  $willNotTerminateInf
+   *
+   *  The sort is stable. That is, elements that are equal (as determined by
+   *  `lt`) appear in the same order in the sorted sequence as in the original.
```
mkeskells: `lt` is not in scope. Do you mean ord.compare/lt etc

DanielMoss: Yes - I'm admittedly not aware of a different `lt` unless you're referring to the parameter passed to `sortWith`? I noted that both `sorted` and `sortWith` had this exact comment in their documentation and from there this felt like the appropriate choice of words.

lrytz: It's a incautious copy paste error in the doc of `sorted` and should be fixed.

adriaanm (dismissed):
Please fix the copy/paste error identified by @mkeskells and @lrytz

src/library/scala/collection/SeqLike.scala
DanielMoss: Ah, my bad then. I've updated the scaladoc to refer to `ord.compare` for both `sortBy` and `sorted`, which I believe should now make the docs correct.

SethTisue:
@DanMoss this looks ready to merge to me. before we do that, can you squash the two commits into a single commit?

DanMoss:
Cheers @SethTisue - that should be done now. Apologies for the delay, had some unexpected trouble with git that I've only had the time for tonight.

SethTisue:
@DanMoss no worries! thanks for catching this, and for submitting the fix.

SethTisue:
I've made a to-do item locally to make sure this doesn't get lost when https://github.com/scala/collection-strawman lands here.

SethTisue:
it didn't get lost.

PR #6188, opened 2017-11-21
 Fix @inheritdoc on vals.

hrhino:
The field symbol doesn't override the superclass members; it's the responsibility of the getter to do that.

~The deprecated `getter` method in `Symbols.scala` stymied my ability to say `sym.getter.orElse(sym)`, so I got rid of it (and it's sibling `setter`, so it wouldn't be lonely). It's in `reflect.internal`, and it's been deprecated for quite a few releases, so I can't imagine anyone's using it.~ Nope, `compiler-interface` does. Woops.

I couldn't resist the urge to shoo away a gratuitous `var` in a related method. Apologies for the bloated diff.

Fixes scala/bug#10621.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/DocComments.scala
+++ b/src/compiler/scala/tools/nsc/ast/DocComments.scala
@@ -70,9 +70,10 @@ trait DocComments { self: Global =>
    *  the doc comment of the overridden version is copied instead.
    */
   def cookedDocComment(sym: Symbol, docStr: String = ""): String = cookedDocComments.getOrElseUpdate(sym, {
-    var ownComment = if (docStr.length == 0) docComments get sym map (_.template) getOrElse ""
-                       else DocComment(docStr).template
-    ownComment = replaceInheritDocToInheritdoc(ownComment)
+    val ownComment = replaceInheritDocToInheritdoc {
+      if (docStr.length == 0) docComments get sym map (_.template) getOrElse ""
+      else DocComment(docStr).template
+    }
 
     superComment(sym) match {
       case None =>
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/ast/DocComments.scala
```diff
@@ -133,8 +134,12 @@ trait DocComments { self: Global =>
     mapFind(sym :: allInheritedOverriddenSymbols(sym))(docComments get _)
 
   /** The cooked doc comment of an overridden symbol */
-  protected def superComment(sym: Symbol): Option[String] =
-    allInheritedOverriddenSymbols(sym).iterator map (x => cookedDocComment(x)) find (_ != "")
+  protected def superComment(sym: Symbol): Option[String] = {
+    val getter = sym.getterIn(sym.owner)
```
lrytz: There's `sym.getter` as a shorthand

hrhino: Sadly that doesn't work (see the stricken text in the PR description): evidently Scala can't decide whether to eta-expand the deprecated one-arg `getter` method or select from the one you mention.

lrytz: Oh, overloading resolution..!
```
scala> :pa -raw

class A { def f = 0 }
object C extends A { def f(x: Int) = 1 }

scala> C.f
<console>:12: error: ambiguous reference to overloaded definition,
both method f in object C of type (x: Int)Int
and  method f in class A of type => Int
match expected type ?
       C.f
         ^
```

It works if you provide the `: Symbol` ascription.

hrhino: Yeah. I think I'm going to pop over to sbt and see if I can get rid of them using the deprecated one. It's been years, really.

retronym (approved):

SethTisue:
thank you!

PR #6189, opened 2017-11-21
allow PerRunInit to operate in multi-threaded environment

mkeskells:
this is an enabling change spun out of https://github.com/scala/scala/pull/6124 to aid review

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/PerRunInit.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/PerRunInit.scala
@@ -9,10 +9,12 @@ import scala.collection.mutable.ListBuffer
  * The trait provides an `initialize` method that runs all initializers added through `perRunLazy`.
  */
 trait PerRunInit {
+  // We have to synchronize on inits, as many of the initializers are themselves lazy,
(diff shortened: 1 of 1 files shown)


mkeskells:
I think that the perRunInit model is wrong, for issues outside this PR

It takes an initialiser that runs on first use or on reinitialisation, which takes a lock and is not documented to do so

This is inefficient both for memory retention and multi-threading - 

1. many of the operations don't require a frontEndLock

1. the data stuctures are not cleared when not in use 

I think that working out the better API will take some work and could be the subject of a separate PR, and probably not in 2.12. 

I do wonder if the reinit should be done in the end of phase rather than the start of phase @lrytz what do you think

lrytz:
I'm happy to work on a a better API here. I think we can safely work on the 2.12 branch, as there's (currently) no parallelism.

> reinit should be done in the end of phase

We cannot eagerly initialize at the end of a phase, the point of per-run inits is to re-compute values that depend on symbols, and symbols may change from one run to another (not the symbols themselves, they stay the same, but their information like types). So we need to do the computation when the new symbols are available.

However we can change to a model where we clear eagerly at the end of a run and compute all values lazily (during the next run). The advantage I guess is that references would be released.

lrytz (approved):

PR #6190, opened 2017-11-22
Warn-unused uses type alias in nesting check

som-snytt:
A reference from the body of a class is not a usage
of the class, but a reference to a local type alias of it
is a usage of the type alias.

Also don't warn about filter for refutable pattern.

Also `classOf[C]` is a usage of `C`.

Simplify message for a `var` that isn't mutated.

Fixes scala/bug#10623
Fixes scala/bug#10394
Fixes scala/bug#9058
Fixes scala/bug#10282
Fixes scala/bug#10448

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
@@ -540,6 +540,9 @@ trait TypeDiagnostics {
                 case NullaryMethodType(_) =>
                 case MethodType(_, _)     =>
                 case SingleType(_, _)     =>
+                case ConstantType(Constant(k: Type)) =>
+                  log(s"classOf $k referenced from $currentOwner")
+                  treeTypes += k
                 case _                    =>
                   log(s"${if (isAlias) "alias " else ""}$tp referenced from $currentOwner")
                   treeTypes += tp
--- a/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
@@ -79,7 +79,11 @@ trait TypeDiagnostics {
     prefix + name.decode
   }
 
+  // Bind of pattern var was `x @ _`
   private def atBounded(t: Tree) = t.hasAttachment[AtBoundIdentifierAttachment.type]
+
+  // ValDef was a PatVarDef `val P(x) = ???`
+  private def wasPatVarDef(t: Tree) = t.hasAttachment[PatVarDefAttachment.type]
 
   /** Does the positioned line assigned to t1 precede that of t2?
    */
@@ -478,7 +482,6 @@ trait TypeDiagnostics {
         val targets   = mutable.Set[Symbol]()
         val setVars   = mutable.Set[Symbol]()
         val treeTypes = mutable.Set[Type]()
-        val atBounds  = mutable.Set[Symbol]()
         val params    = mutable.Set[Symbol]()
         val patvars   = mutable.Set[Symbol]()
 
@@ -503,16 +506,19 @@ trait TypeDiagnostics {
           val sym = t.symbol
           var bail = false
           t match {
-            case m: MemberDef if qualifies(t.symbol)   =>
-              defnTrees += m
+            case m: MemberDef if qualifies(sym)   =>
               t match {
+                case ValDef(mods@_, name@_, tpt@_, rhs@_) if wasPatVarDef(t) =>
+                  if (!atBounded(t)) patvars += sym
                 case DefDef(mods@_, name@_, tparams@_, vparamss, tpt@_, rhs@_) if !sym.isAbstract && !sym.isDeprecated && !sym.isMacro =>
                   if (sym.isPrimaryConstructor)
                     for (cpa <- sym.owner.constrParamAccessors if cpa.isPrivateLocal) params += cpa
                   else if (sym.isSynthetic && sym.isImplicit) bail = true
                   else if (!sym.isConstructor)
                     for (vs <- vparamss) params ++= vs.map(_.symbol)
+                  defnTrees += m
                 case _ =>
+                  defnTrees += m
               }
             case CaseDef(pat, guard@_, rhs@_) if settings.warnUnusedPatVars    =>
               pat.foreach {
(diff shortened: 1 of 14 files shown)


src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -527,8 +527,10 @@ trait TypeDiagnostics {
           // Only record type references which don't originate within the
           // definition of the class being referenced.
           if (t.tpe ne null) {
-            for (tp <- t.tpe if !treeTypes(tp) && !currentOwner.ownerChain.contains(tp.typeSymbol)) {
-              tp match {
+            for (tp <- t.tpe if !treeTypes(tp)) {
+              val alias = tp.isInstanceOf[AliasTypeRef]
```
som-snytt: Is this the best one can do?

hrhino: `tp.typeSymbolDirect.isAliasType`?

som-snytt: That pulls in too much, it turns out. I tried various `isNonClassType` etc, so mostly I don't actually understand the API.

lrytz: > That pulls in too much

Can you give an example?

som-snytt: In constructors you see:
```
[log typer] ALIAS C.super.type/class scala.reflect.internal.Types$UniqueSuperType/class C/type AnyRef primary constructor C
[log typer] C.super.type referenced from primary constructor C
```
with the false negative for `object X { private class C }`. So the `AnyRef` is the `Direct.isAliasType`, defined as an alias of `Object`.

I guess it's sufficient to also check `isLocal || isPrivate`.

SethTisue:
this seems to have been forgotten about. can you suggest a reviewer?

som-snytt:
@SethTisue I believe the community that wants the fix will step up to review it! I believe in the community!

src/reflect/scala/reflect/internal/TreeGen.scala
```diff
@@ -732,11 +732,16 @@ abstract class TreeGen {
   def mkPatDef(pat: Tree, rhs: Tree)(implicit fresh: FreshNameCreator): List[ValDef] =
     mkPatDef(Modifiers(0), pat, rhs)
 
+  private def cpAtBoundAttachment(from: Tree, to: ValDef): to.type =
+    if (from.hasAttachment[AtBoundIdentifierAttachment.type]) to.updateAttachment(AtBoundIdentifierAttachment) else to
+  private def cpPatVarDefAttachments(from: Tree, to: ValDef): to.type =
```
hrhino: I assume this is UNIX for `copy`? It seems strange that `cpAtBoundAttachment` only adds the attachment if `from` has it but `cpPatVarDefAttachments` unconditionally adds `PatVarDefAttachment`.

hrhino (approved):
Just because you asked.

It might be cool to add the ticket references in somewhere for future retrospective capabilities.

src/reflect/scala/reflect/internal/TreeGen.scala
som-snytt: One propagates whether the syntax was `x @ _`, the other adds that this is a patvar. There is doc over in TypeDiag.

I would justify `cp` by persistence of `mk` for make and similar, which are not due to a subculture. I do resist long names, and these are long.

I think I will shave my beard now and maybe go get a hair cut.

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
```diff
@@ -10,7 +10,7 @@ import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 import scala.util.control.Exception.ultimately
 import symtab.Flags._
-import PartialFunction._
+import PartialFunction.{condOpt => whenever}
```
SethTisue, line 13: Seth's nit: I know, `condOpt` isn't the greatest name, but it's the name we have, I don't think we should indulge ourselves in arbitrary local renamings like this

SethTisue (approved):
this all looks plausible to me, but maybe one more pair of eyes on the attachments stuff.

src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala
som-snytt: I know what you mean. Paulp once said those methods where the most underlooked of his contributions, and I think it's partly the names. Maybe we can fix that.

som-snytt:
I need another session to exercise the last commit. While debugging the underlying issue, I realized I know nothing.

lrytz (approved):
🏅

hrhino:
And we're back under 2,100 issues in scala/bug!

lrytz:
🙀

som-snytt:
I'll create a few new issues to address @SethTisue 's nits. Is there a "nit" label?

SethTisue:
very glad to see this crossing the finish line! 🎉  thanks Som, greatly appreciate your ongoing commitment to this unused warnings work, it's a real boon for code quality community-wide

som-snytt:
I won't rest until all the code is used.

SethTisue:
no child (in any abstract syntax tree) left behind

som-snytt:
"I speak for the trees." - The Lorax.

adriaanm:
Looks like this regressed for a local val captured in a CBN closure:

```
class Test {
  def timed[T](body: => T) = body

  def run(): Boolean = {
    val grouped = List(("bla","bla"))


    val (_, _) = timed {
      for ((kind, paths) <- grouped) {
        println("")
      }
      (1, 2)
    }
    
    false
  }
}
```

nok:
```
➜  Desktop sbt 'set resolvers += "pr-integration" at "https://scala-ci.typesafe.com/artifactory/scala-integration"' 'set scalaVersion := "2.12.5-bin-56918e4"' 'set scalacOptions ++= Seq("-Xfatal-warnings", "-Xlint:unused")' compile
[info] Compiling 1 Scala source to /Users/adriaan/Desktop/target/scala-2.12/classes...
[error] /Users/adriaan/Desktop/unused_captured.scala:5: local val grouped in method run is never used
[error]     val grouped = List(("bla","bla"))
[error]         ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
[error] Total time: 2 s, completed Mar 8, 2018 11:25:25 AM
```

ok:
```
➜  Desktop sbt 'set resolvers += "pr-integration" at "https://scala-ci.typesafe.com/artifactory/scala-integration"' 'set scalaVersion := "2.12.5-bin-f78f517"' 'set scalacOptions ++= Seq("-Xfatal-warnings", "-Xlint:unused")' compile
[info] Compiling 1 Scala source to /Users/adriaan/Desktop/target/scala-2.12/classes...
[success] Total time: 3 s, completed Mar 8, 2018 11:25:49 AM
```

fommil:
thanks @som-snytt ... ~~I'm still getting false positives in synthetic methods. I don't know how to create a minimisation that doesn't depend on the scalaz-deriving compiler plugin.~~

(Actually this might have been a bug in my compiler plugin...)

PR #6194, opened 2017-11-23
cache the result of an expensive canonical file lookup

mkeskells:
for abstract files that can have a canonicalPath cache the result as this may be frequently called

Not expected this to be a significant change, but will benchmark

### DIFF
--- a/src/reflect/scala/reflect/io/PlainFile.scala
+++ b/src/reflect/scala/reflect/io/PlainFile.scala
@@ -22,6 +22,9 @@ class PlainFile(val givenPath: Path) extends AbstractFile {
   assert(path ne null)
 
   val file = givenPath.jfile
+
+  override lazy val canonicalPath = super.canonicalPath
+
   override def underlyingSource = Some(this)
 
   private val fpath = givenPath.toAbsolute
@@ -102,6 +105,8 @@ private[scala] class PlainNioFile(nioPath: java.nio.file.Path) extends AbstractF
   } catch {
     case _: UnsupportedOperationException => null
   }
+
+  override lazy val canonicalPath = super.canonicalPath
 
   override def underlyingSource  = Some(this)
 
--- a/src/reflect/scala/reflect/io/ZipArchive.scala
+++ b/src/reflect/scala/reflect/io/ZipArchive.scala
@@ -66,6 +66,8 @@ import ZipArchive._
 abstract class ZipArchive(override val file: JFile) extends AbstractFile with Equals {
   self =>
 
+  override lazy val canonicalPath = super.canonicalPath
+
   override def underlyingSource = Some(this)
   def isDirectory = true
   def lookupName(name: String, directory: Boolean) = unsupported()


mkeskells:
results on windows for akka compile, based on 150 cycles eliminating the first 20 (warmup) and worst 10% (outliers)
** Updated stats **

```
               01_2.12.x        3183      9,495.64 [-12.65% +4.40%]			9,337.21 [-12.98% +3.58%]           1,993.53 [-0.05% +0.04%]
               02_2.12.x        3183      9,293.90 [-4.64% +5.38%]			9,145.97 [-3.65% +4.55%]            2,000.94 [-0.05% +0.05%]

              00_CanFile        3183      9,383.89 [-5.10% +7.60%]			9,221.55 [-5.11% +5.73%]            1,985.32 [-0.10% +0.12%]
              01_CanFile        3183      9,257.88 [-6.28% +7.47%]			9,108.17 [-6.85% +6.19%]            1,977.94 [-0.08% +0.13%]
              02_CanFile        3183      9,225.93 [-4.87% +5.45%]			9,078.39 [-5.85% +4.64%]            1,981.27 [-0.06% +0.05%]
```

retronym:
What call site presents the performance problem?

Is it:

```scala
byteCodeRepository.add(c.classNode, Some(c.sourceFile.canonicalPath))
```

Or:

```
    /** Is this symbol defined in the same scope and compilation unit as `that` symbol? */
    def isCoDefinedWith(that: Symbol) = (
         !rawInfoIsNoType
      && (this.effectiveOwner == that.effectiveOwner)
      && (   !this.effectiveOwner.isPackageClass
          || (this.associatedFile eq NoAbstractFile)
          || (that.associatedFile eq NoAbstractFile)
          || (this.associatedFile.path == that.associatedFile.path)  // Cheap possibly wrong check, then expensive normalization
          || (this.associatedFile.canonicalPath == that.associatedFile.canonicalPath)
         )
    )
```

mkeskells:
@retronym this change was motivated by code inspection rather than profiling, and I would be suspicious of the result - I will post up a rerun in a few hours when it completes. If you like I can send a phase-by-phase breakdown.
I was not expecting this to yield noticeable change. Lets see what the rerun yields as the baseline was before the machine crash, maybe there was some other effect on the numbers for that run

mkeskells:
@retronym I have updated updated the stats - closer, but still better than I expected

taking the averages of the runs cpu time reduced by 1.13% and allocation by 0.79%

**note** the iimprovement is platform neutral, but its scale is platform specific, canonicalisation is a function of the file system, and therefore on Unix system the improvements may be different

retronym:
Down the track I'd like to move to a model more like `JavacFileManager` / `FSInfo` used in Javac, in which the caching of file stats and derived information has a more clearly defined lifetime. But in our current architecture, I beleive this new cache is acceptable.

retronym (approved):

mkeskells:
Maybe when we get the new classpath implementation we can just be canonical :-)

PR #6196, opened 2017-11-27
add comonadic comprehensions (keyword 'cofor')

shimib:
This is a proposal (POC) to provide a new keyword (cofor) for co-monadic comprehensions in Scala.
The work is based over Dominic Orchard and Alan Mycroft paper: https://www.cl.cam.ac.uk/~dao29/publ/codo-notation-orchard-ifl12.pdf

If you find the implementation flawed, it is me to blame and not them!

the 'cofor' expression returns a function: (T[A] => B).
The function's type MUST be stated explicitly in the call-site. e.g.:
val result: Zipper[Person] => Boolean = cofor...

the syntax:
cofor(inputpatten) {
pattern1 <- gen1
pattern2 <- gen2
...
} yield body

the 'cofor' and generator must use types that provide:
def map(A => B)
def extract: A
def coflatMap(T[A] => B)

implementation is much more complex than 'for' desugaring as stated in the paper.

current implementation supports full Scala patterns in the input and generator patterns.
missing:

desugared tree positioning validation (should not affect surrounding code)
quasiquotes/macros/reification -- this is the next phase in the implementation.
the flag to control the keyword: -Ycofor-extension
if the flag is not enabled, current Scala code should not be affected.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -1522,6 +1522,24 @@ self =>
             tree setPos tree.pos.withStart(start)
           else tree
         adjustStart(parseFor)
+        case COFOR if settings.YcoforExtension =>
+        val start = in.skipToken()
+        def parseCoFor = atPos(start) {
+
+          val inputPattern = inParens(gen.patvarTransformer.transform(noSeq.pattern1()))
+          val enums =
+            inBraces(cofor_enumerators())
+          newLinesOpt()
+          accept(YIELD)
+
+          gen.mkCoFor(inputPattern, enums, expr())
+
+        }
+        def adjustStart(tree: Tree) =
+          if (tree.pos.isRange && start < tree.pos.start)
+            tree setPos tree.pos.withStart(start)
+          else tree
+        adjustStart(parseCoFor)
       case RETURN =>
         def parseReturn =
           atPos(in.skipToken()) {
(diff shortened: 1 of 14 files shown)


adriaanm:
I apologize if the conclusion of the SIP meeting was not communicated to you properly, but the SIP for this change was [rejected](http://docs.scala-lang.org/sips/minutes/2017-05-08-sip-minutes.html). I agree with @odersky's [review](https://github.com/scala/docs.scala-lang/pull/709): before we bake this into the language, we first want to see adoption of these ideas, which themselves do not depend (much) on having syntactic support. Honestly, I am skeptical this adoption will happen at a scale that we'd like to see.

Our focus for Scala is on simplification and better tooling, and this complicates both those efforts. We have a limited budget to spend on language changes, and I'd rather see work on non-null types, simpler implicit resolution, direct support for deriving and HLists, ["monadic-if"](https://github.com/scala/scala/pull/5995#issuecomment-347971536) or even type-level computation (perhaps using Martin's [staging macro proposal](https://gist.github.com/odersky/f91362f6d9c58cc1db53f3f443311140#gistcomment-2242950)).

I'm always willing to discuss, but the venue for that is the SIP meeting. Given the SIP committee's current stance on this, I am closing this PR.

Some relevant tickets for the direction we have in mind for Scala: https://github.com/scala/scala-dev/milestone/18, https://github.com/scala/scala-dev/issues/430

shimib:
Thanks for the detailed response Adriaan. 
I'm quite aware of the SIP resolution, i submitted this PR for TLS sake (as it may be included there).
Sorry for any inconvenience caused by the submission.

PR #6197, opened 2017-11-27
Reduce footprint of compiler Symbols

retronym:
Overall diff in JOL reported layouts:

https://gist.github.com/retronym/1a0d9f14802eb5ec1112781b074c6ec8

TermSymbol goes from 88 to 72 bytes.

### DIFF
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -2857,16 +2857,14 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
       this
     }
 
-    private val validAliasFlags = SUPERACCESSOR | PARAMACCESSOR | MIXEDIN | SPECIALIZED
-
     override def alias: Symbol =
-      if (hasFlag(validAliasFlags)) initialize.referenced
+      if (hasFlag(ValidAliasFlags)) initialize.referenced
       else NoSymbol
 
     def setAlias(alias: Symbol): TermSymbol = {
       assert(alias != NoSymbol, this)
       assert(!alias.isOverloaded, alias)
-      assert(hasFlag(validAliasFlags), this)
+      assert(hasFlag(ValidAliasFlags), this)
 
       referenced = alias
       this
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -217,7 +217,9 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     // makes sure that all symbols that runtime reflection deals with are synchronized
     private def isSynchronized = this.isInstanceOf[scala.reflect.runtime.SynchronizedSymbols#SynchronizedSymbol]
     private def isAprioriThreadsafe = isThreadsafe(AllOps)
-    assert(isCompilerUniverse || isSynchronized || isAprioriThreadsafe, s"unsafe symbol $initName (child of $initOwner) in runtime reflection universe")
+
+    if (!(isCompilerUniverse || isSynchronized || isAprioriThreadsafe))
+      throw new AssertionError(s"unsafe symbol $initName (child of $initOwner) in runtime reflection universe")
 
     type AccessBoundaryType = Symbol
     type AnnotationType     = AnnotationInfo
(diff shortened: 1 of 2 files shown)


xeno-by:
Very curious to see the performance impact of this change!

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2386,11 +2386,23 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
...
     )
+    final def isOverridingSymbol: Boolean = {
+      val curRunId = currentRunId
+      if (isOverridingSymbolCache == curRunId) true
+      else if (isOverridingSymbolCache == -curRunId) false
+      else {
+        val result = computeIsOverridingSymbol
+        isOverridingSymbolCache = (if (result) 1 else -1) * curRunId
+        result
+      }
```
retronym, line 2407: While this fixes a bug in the cache under multi-run compilation, we might still have a bug here if the result of this method is phase dependent. I'm thinking of a case where something doesn't override at typer, but does override at erasure.

Surveying the the call heirarchy:

`allOverridingSymbols` uses this, and it has hot callsites in `typedBlock => isPossibleInRefinement`; and in `SpecializeTypes.needsSpecialOverride`.

`nextOverridingSymbol` also uses this, and it has hot callsites in `Namer.methodSig` (for return type and default method inheritance).

So we wouldn't want to drop the caching altogether, but we could bypass the cache after we get to a certain phase.

To motivate this more precisely, we could try to craft an example that is vulnerable to the bug.

retronym: Concrete examples.

## Incorrect caching across runs.

Prior to this PR:
```
⚡ scala-pr 6197
scala> :power

scala> :paste -raw

package p1; class A; class B extends A { def foo = 42 }

scala> typeOf[p1.B].decl(TermName("foo")).isOverridingSymbol
res0: Boolean = false

scala> :paste -raw

package p1; class A { def foo = 42 }

scala> typeOf[p1.B].decl(TermName("foo")).isOverridingSymbol
res1: Boolean = false
```

After, we correctly recompute the override status of `B.a` after the parent class was separately redefined.

```
⚡ scala-pr 6197
scala> :power

scala> :paste -raw

package p1; class A; class B extends A { def foo = 42 }

scala> typeOf[p1.B].decl(TermName("foo")).isOverridingSymbol
res0: Boolean = false

scala> :paste -raw

package p1; class A { def foo = 42 }

scala> typeOf[p1.B].decl(TermName("foo")).isOverridingSymbol
res1: Boolean = true
```

## Incorrect caching across phases

This problem has _not_ been addressed by this PR.

```scala
trait A[X] { def a(x: X) }; trait B extends A[String] { def a(s: String) }

scala> def B_a = typeOf[B].decl(TermName("a"))
B_a: $r.intp.global.Symbol

```

`exitingErasure(B_a.isOverridingSymbol)` should give `false`, but will still yield `true` if `isOverridingSymbol` was called pre-erasure:


```
scala> { println(exitingErasure(B_a.isOverridingSymbol)) }
false
```

```
scala> { println(B_a.isOverridingSymbol); println(exitingErasure(B_a.isOverridingSymbol)) }
true
true
```

I propose to leave this problem as TODO.

retronym:
Ping @adriaanm for review, or @lrytz alternatively.

src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -217,7 +217,9 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     // makes sure that all symbols that runtime reflection deals with are synchronized
     private def isSynchronized = this.isInstanceOf[scala.reflect.runtime.SynchronizedSymbols#SynchronizedSymbol]
     private def isAprioriThreadsafe = isThreadsafe(AllOps)
-    assert(isCompilerUniverse || isSynchronized || isAprioriThreadsafe, s"unsafe symbol $initName (child of $initOwner) in runtime reflection universe")
+
+    if (!(isCompilerUniverse || isSynchronized || isAprioriThreadsafe))
```
adriaanm, line 221: Add a comment to avoid future boyscouting back to an `assert`

src/reflect/scala/reflect/internal/Symbols.scala
adriaanm: Ok, but do add the `// TODO`

adriaanm (changes requested):
After a few added comments, LGTM

retronym:
Comments added.

adriaanm (approved):
Sorry about the delay

PR #6198, opened 2017-11-27
Drop PagedSeq

adriaanm:
Let's see if we can get this done for M3. We'll need a scala-parser-combinators 1.1.x release for 2.13.

Fixes scala/scala-dev#298

Original deprecation in scala/bug#9503


SethTisue:
> We'll need a scala-parser-combinators 1.1.x release for 2.13

I don't think so, actually. (Despite my past remarks at https://github.com/scala/scala-dev/issues/298.) Scala no longer depends on scala-parser-combinators and (since #5790) we no longer bundle it, either.

We do still _want_ such a release to happen, since the library is still in wide use. But we can roll M3 regardless.

PR #6199, opened 2017-11-29
don't evaluate the head of a traversable twice in last

martijnhoekstra:
don't call head before looping over the traversable
fixes scala/bug#10631

### DIFF
--- a/src/library/scala/collection/TraversableLike.scala
+++ b/src/library/scala/collection/TraversableLike.scala
@@ -428,13 +428,10 @@ trait TraversableLike[+A, +Repr] extends Any
     * @throws NoSuchElementException If the $coll is empty.
     */
   def last: A = {
-    var lst: A = null.asInstanceOf[A]
-    var hasElements = false
-    for (x <- this){
-      hasElements = true
+    var lst = head
+    for (x <- this)
       lst = x
-    }
-    if (hasElements) lst else throw new NoSuchElementException("last of empty traversable")
+    lst
   }
 
   /** Optionally selects the last element.
--- a/src/library/scala/collection/TraversableLike.scala
+++ b/src/library/scala/collection/TraversableLike.scala
@@ -428,10 +428,13 @@ trait TraversableLike[+A, +Repr] extends Any
     * @throws NoSuchElementException If the $coll is empty.
     */
   def last: A = {
-    var lst = head
-    for (x <- this)
+    var lst: A = null.asInstanceOf[A]
+    var hasElements = false
+    for (x <- this){
+      hasElements = true
       lst = x
-    lst
+    }
+    if (hasElements) lst else throw new NoSuchElementException("last of empty traversable")
   }
 
   /** Optionally selects the last element.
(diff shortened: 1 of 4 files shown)


martijnhoekstra:
I don't entirely understand why (at least the `lastOption` part of) 32d2b15 targets `TraversableViewLike` rather than `TraversableLike` directly. If that's not the right thing to do, maybe it's better to move the `lastOption` implementation override to `TraversableLike` or if it is, to move this `last` implementation to `TraversableViewLike`.

Maybe @Ichoran would have an opinion on that?

martijnhoekstra:
After some thinking about this (I'm a slow thinker), I came to the conclusion that since View is the only known lazy collection root above TraverableLike, it makes *some* sense to have the fast version with isEmpty on TraversableLike and the one that gives correct behaviour for lazy collections overridden on View, which means you don't have to override everywhere else for performance.

For the case of `lastOption` I can see that making some sense, since you shortcircuit the empty case by never doing the foreach. In case of `last` however the empty case is already pathalogical in terms of performance, since it throws. That means that that can't be a performance-critical path.

While I think that morally it makes more sense to have an always correct base implementation and override for performance if possible, I can see the practical consideration of having the "usually" correct and in that case faster implementation on TraversableLike and override just for View, rather than overriding in each collection.

Also, this PR targets 2.13.x, but may be superseded by the collections rework. 2.12.x is a more natural home for it I think.

szeiger:
In 2.8 to 2.12 the default implementations are usually geared towards strict collections and need to be overwritten in lazy collections, so it makes sense to have the faster version in `TraversableLike` and implement your fix in `TraversableViewLike` instead.

As you already mentioned it should also target 2.12.x instead of 2.13.x because all of this will be replaced by the new collections in 2.13.0-M4. Additionally you may want to implement and benchmark a similar optimization for `StrictOptimizedIterableOps` in the collection-strawman repo. There we have a default lazy implementation in `IterableOps` which is based on an `Iterator`. Without benchmarking I wouldn't want to guess whether adding an up-front `isEmpty` check makes it faster or slower on average.

lrytz:
@martijnhoekstra as far as I understand, the TODO here is to move your new implementation to `TraversableViewLike`, then this PR can be merged.

martijnhoekstra:
I forgot all about this @lrytz - I'll get back on it, rebase on 2.12.x, and move the impl to TraversableViewLike.

SethTisue (changes requested):
"rebase on 2.12.x, and move the impl to TraversableViewLike"

martijnhoekstra:
re-basing and re-targeting seems to have confused CI, as it's running 2.13.x validate main, and it seems to be seeing earlier commits that are not there.

@SethTisue is that a problem, and if so, can you tell it these are not the droids it's looking for?

SethTisue (approved):

SethTisue:
I hit "approve", but the real signoff should come from Lukas or Stefan.

As for Jenkins... the commits are green now. The fact that 2.13.x-validate-main ran them rather than 2.12.x-validate-main may not actually matter, but I'm not sure.

lrytz (approved):

SethTisue:
thank you Martijn!

PR #6201, opened 2017-11-29
improve construction of ClassBTypes

mkeskells:
change spun out of #6124 to aid review

initialise ClassBType info field as part of construction
ban external mutation of info
provide memory barriers for late initialised info field

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -36,12 +36,12 @@ abstract class BTypes {
    * name. The method assumes that every class type that appears in the bytecode exists in the map
    */
   def cachedClassBType(internalName: InternalName): Option[ClassBType] =
-    classBTypeCacheFromSymbol.get(internalName).orElse(classBTypeCacheFromClassfile.get(internalName))
+    classBTypeCache.get(internalName)
 
   // Concurrent maps because stack map frames are computed when in the class writer, which
   // might run on multiple classes concurrently.
-  val classBTypeCacheFromSymbol: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(FlatConcurrentHashMap.empty)
-  val classBTypeCacheFromClassfile: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(FlatConcurrentHashMap.empty)
+  // Note usage should be private to this file, except for tests
+  val classBTypeCache: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(FlatConcurrentHashMap.empty)
 
   /**
    * A BType is either a primitive type, a ClassBType, an ArrayBType of one of these, or a MethodType
@@ -607,7 +607,8 @@ abstract class BTypes {
    * a missing info. In order not to crash the compiler unnecessarily, the inliner does not force
    * infos using `get`, but it reports inliner warnings for missing infos that prevent inlining.
    */
-  final class ClassBType private (val internalName: InternalName) extends RefBType {
+  sealed abstract class ClassBType protected(val internalName: InternalName) extends RefBType {
+    def fromSymbol: Boolean
     /**
      * Write-once variable allows initializing a cyclic graph of infos. This is required for
      * nested classes. Example: for the definition `class A { class B }` we have
@@ -814,17 +815,27 @@ abstract class BTypes {
     )
     def unapply(cr:ClassBType) = Some(cr.internalName)
 
-    def apply(internalName: InternalName, cache: mutable.Map[InternalName, ClassBType])(init: (ClassBType) => Either[NoClassBTypeInfo, ClassInfo]) = {
-      val res = new ClassBType(internalName)
+    def apply(internalName: InternalName, fromSymbol: Boolean)(init: (ClassBType) => Either[NoClassBTypeInfo, ClassInfo]) = {
+      val newRes = if (fromSymbol) new ClassBTypeFromSymbol(internalName) else new ClassBTypeFromClassfile(internalName)
       // synchronized s required to ensure proper initialisation if info.
       // see comment on def info
-      res.synchronized {
-        cache(internalName) = res
-        res._info = init(res)
-        res.checkInfoConsistency()
+      newRes.synchronized {
+        classBTypeCache.putIfAbsent(internalName, newRes) match {
+          case None =>
+            newRes._info = init(newRes)
+            newRes.checkInfoConsistency()
+            newRes
+          case Some(old) =>
+            old
+        }
       }
-      res
     }
+  }
+  private final class ClassBTypeFromSymbol(internalName: InternalName) extends ClassBType(internalName) {
+    override def fromSymbol: Boolean = true
+  }
+  private final class ClassBTypeFromClassfile(internalName: InternalName) extends ClassBType(internalName) {
+    override def fromSymbol: Boolean = false
   }
 
   /**
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
@@ -607,7 +607,7 @@ abstract class BTypes {
    * a missing info. In order not to crash the compiler unnecessarily, the inliner does not force
    * infos using `get`, but it reports inliner warnings for missing infos that prevent inlining.
    */
-  final case class ClassBType(internalName: InternalName)(cache: mutable.Map[InternalName, ClassBType]) extends RefBType {
+  final class ClassBType private (val internalName: InternalName) extends RefBType {
     /**
      * Write-once variable allows initializing a cyclic graph of infos. This is required for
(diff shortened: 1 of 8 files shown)


src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -804,6 +812,19 @@ abstract class BTypes {
...
+    def apply(internalName: InternalName, cache: mutable.Map[InternalName, ClassBType])(init: (ClassBType) => Either[NoClassBTypeInfo, ClassInfo]) = {
+      val res = new ClassBType(internalName)
+      // synchronized s required to ensure proper initialisation if info.
+      // see comment on def info
+      res.synchronized {
+        cache(internalName) = res
+        res._info = init(res)
+        res.checkInfoConsistency()
+      }
+      res
```
lrytz: This code doesn't ensure that there's a unique `ClassBType` per `InternalName`. For example, if there are two concurrent executions of `classBTypeFromParsedClassfile` for the same `InternalName`, both could get `None` from `cachedClassBType` and create a new instance.

mkeskells: Yes- need to improve that to make the new implementation more tolerent to MT issues than the existing as part of the PR

simplest fix is to use concurrent APIs for the cache, but this still exposes concurrent creation between bTypesFromSymbols and bTypesFromClassFiles

I can't see a logical reason for the separation of the these 2 caches - they perform the same function and are unified for read access. 

@lrytz  Do you have any objection to unifying these caches? Leaving a boolean field (or child class) to signify the origin to satisfy the assert in ` def classBTypeFromSymbol` apply method then uses putIfAbsent to ensure uniqueness. (Ideally we could use computeIfAbsent and remove the whole dummy synchronization, but I think that this would fail due to recursive access)
Added a commit to aid discussion with unified caches and a child class for the breadcrumb

At a later time we could remove the scala wrapping of these caches to save a few cpu cycles

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -607,29 +607,29 @@ abstract class BTypes {
...
+    // volatile is required to ensure no early initialisation in apply
+    // like classic double checked lock in java
+    @volatile private var _info: Either[NoClassBTypeInfo, ClassInfo] = null
 
     def info: Either[NoClassBTypeInfo, ClassInfo] = {
+      if (_info eq null)
+        // synchronization required to ensure the apply is finished
+        // which populates info. ClassBType doesnt escape apart from via the map
+        // and the object mutex is locked prior to insertion. See apply
+        this.synchronized()
```
retronym, line 628: What's the intent of the empty synchronized block here? I'm not sure this should have any effect under the Java Memory Model.

https://stackoverflow.com/a/23961401

mkeskells: it requires the acquire and releaseof the lock - see the comment above the statement

The lock is acquired by apply. I could make this an explicit lock but that would demand another field

The JVM cannot elide the call as it is escaping in the apply.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -783,6 +783,15 @@ abstract class BTypes {
       } while (fcs == null)
       fcs
     }
+
```
mkeskells: hashcode and equals are needed by one of the unit tests, and as this is no longer a case class needed to be explicit

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
retronym: Thanks for the clarification. I hadn't seen this idiom before :)

mkeskells: It does seem however that we could model it better to avoid the synchronization, info and Either/BiasedEither and exception as control flow if the info field is one of a pending result or one of the two actual results, via sealed traits

It may be too big a change to include in the PR though

mkeskells: @retronym I do think that the empty synchronized block feels dirty though :-(

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
```diff
@@ -96,17 +96,15 @@ abstract class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
...
+          // The new ClassBType is added to the map via its apply, before we set its info. This
           // allows initializing cyclic dependencies, see the comment on variable ClassBType._info.
-          val res = ClassBType(internalName)(classBTypeCacheFromSymbol)
-          if (completeSilentlyAndCheckErroneous(classSym)) {
-            res.info = Left(NoClassBTypeInfoClassSymbolInfoFailedSI9111(classSym.fullName))
-            res
-          } else {
-            setClassInfo(classSym, res)
+          ClassBType(internalName, true) { res:ClassBType =>
+            if (completeSilentlyAndCheckErroneous(classSym))
```
retronym, line 105: I would assume that we need to take `frontEndLock` here, as this calls into `Symbol.info`.

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
```diff
@@ -36,12 +36,12 @@ abstract class BTypes {
...
   def cachedClassBType(internalName: InternalName): Option[ClassBType] =
-    classBTypeCacheFromSymbol.get(internalName).orElse(classBTypeCacheFromClassfile.get(internalName))
+    classBTypeCache.get(internalName)
 
   // Concurrent maps because stack map frames are computed when in the class writer, which
   // might run on multiple classes concurrently.
-  val classBTypeCacheFromSymbol: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(FlatConcurrentHashMap.empty)
-  val classBTypeCacheFromClassfile: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(FlatConcurrentHashMap.empty)
+  // Note usage should be private to this file, except for tests
+  val classBTypeCache: concurrent.Map[InternalName, ClassBType] = recordPerRunCache(FlatConcurrentHashMap.empty)
```
retronym, line 44: This unification seems okay to me, @lrytz split them in 67c151f1821eb2124ac70c70bf1d974ca0e5b364 but your approach seems to satisfy the same goal.

src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
mkeskells: As I understand it if we have access to the Symbol then we have the lock, so the caller already has the lock, We take the lock when access via some API that uses Symbol or other global data indirectly

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
lrytz: 👍

src/compiler/scala/tools/nsc/backend/jvm/BTypes.scala
lrytz: I'm in favor of removing the overrides, or is it too hard to adjust the tests?

mkeskells: probably not hard - will find out which tests failed

mkeskells: failed test was 
```
[info] Test scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.compareClassBTypes started
[error] Test scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.compareClassBTypes failed: java.lang.AssertionError: assertion failed: Ljava/lang/Object; != Ljava/lang/Object;, took 0.014 sec
[error]     at scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.sameBType(BTypesFromClassfileTest.scala:39)
[error]     at scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.$anonfun$check$1(BTypesFromClassfileTest.scala:106)
[error]     at scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.duringBackend(BTypesFromClassfileTest.scala:26)
[error]     at scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.check(BTypesFromClassfileTest.scala:94)
[error]     at scala.tools.nsc.backend.jvm.opt.BTypesFromClassfileTest.compareClassBTypes(BTypesFromClassfileTest.scala:113)
[error]     ...

```

mkeskells: and should be fixed by https://github.com/scala/scala/pull/6201/commits/261dde55d1b3b4ac87bfa948cd6960feba698bec

lrytz (approved):

PR #6206, opened 2017-12-01
Fix tests that failed due to -Xsource:2.13

lrytz:
In #6092, `-Xsource:2.13` was enabled by default. Between this PRs
parent and current 2.13.x, there were some changes that depended on
this flag (#5983, #6069).


lrytz:
Review by @retronym, @TomasMikula. For the `t6895` test, I only kept a single version without any `.flags` file in `pos/` from all the variants that exist in 2.12.x.

test/files/neg/t7187-2.13.scala
```diff
@@ -1,4 +1,6 @@
 class EtaExpandZeroArg {
   def foo() = ""
-  val f: () => Any = foo
+  val f: () => Any = foo   // error (not eta-expanded)
+  val g: () => Any = foo _ // ok
+  val h = foo _            // error (not eta-expanded)
```
lrytz, line 5: @adriaanm is the above as expected?

adriaanm: That looks like a bug (`foo _` should type check), will include that in my eta-expansion work. Hope to work on that today.

```
scala -Xsource:2.13
Welcome to Scala 2.13.0-20170927-190913-5cf58c5 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_152).

scala> def foo() = ""
foo: ()String

scala> val h = foo _ 
               ^
       error: type mismatch;
        found   : String
        required: () => ?
```

lrytz:
should we merge this pr anyway to unblock the queue? it's only adding a
test for the current behavior, not introducing a bug.

adriaanm:
Yes, let’s unblock. I’ll roll this into my PR

TomasMikula:
👍  regarding `pos/t6895`.

test/files/neg/t7187-2.13.scala
lrytz: https://github.com/scala/bug/issues/10612

PR #6210, opened 2017-12-04
Upgrade to partest 1.1.2

retronym:
Second part of the fix for scala/scala-dev#453.

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -19,6 +19,6 @@ scala.binary.version=2.13.0-M2
 #  - jline: shaded with JarJar and included in scala-compiler
 #  - partest: used for running the tests
 scala-xml.version.number=1.0.6
-partest.version.number=1.1.1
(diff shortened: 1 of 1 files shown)


retronym:
The failure is due to the (temporary? or did I miss a step in publishing?) unavailability of https://repo1.maven.org/maven2/org/scala-lang/modules/scala-partest_2.13.0-M2/1.1.2/scala-partest_2.13.0-M2-1.1.2.pom

In any case, this partest upgrade turned out to be insufficient, it only gets as far as a (fatal) warning about the deprecated setting `-opt:l:classpath` in the bootstrap build.

#6211 comes at the problem from the other direction: disabling `-Xfatal-warnings`.

retronym:
Closing in favour of #6211, given that AFAIK the 1.1.2 version doesn't have anything functional changes we need.

PR #6213, opened 2017-12-05
Elide prefixes in printed types uniformly in runtime reflection

retronym:
The logic that decides to print `Function`, rather than `scala.Function`
did not account for the multiplicity of symbols for a given package
in the JavaMirrors universe.

Fixes scala/bug#5941

References scala/scala-partest#75


adriaanm (approved):

SethTisue:
@retronym the failures at https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-windows/414/consoleFull look like they might be the result of this?

retronym:
Probably just a semantic merge conflict with https://github.com/scala/scala/pull/6143, rather than a windows specific problem.

PR #6215, opened 2017-12-05
Narrow scope of sensibility check for equals

adriaanm:
I have my doubts whether it's correct, but at least it's less ambitious in its erring.

Fixes scala/bug#10644


adriaanm:
Weird, I don't see how this could have broken scala.tools.nsc.backend.jvm.opt.MethodLevelOptsTest

retronym:
This probably just needs a rebase.

retronym (approved):

lrytz:
Rebased and pushed to `adriaanm`

PR #6216, opened 2017-12-06
Revert "Type Alternative patterns as the lub of their branches."

retronym:
This reverts commit 31210e7f6306b4a9fce27a7f2dcb687cdbbcbf2d.

Fixes scala/scala-dev#455


retronym:
Reverting for now, @hrhino.

test/files/run/sd455.scala
```diff
@@ -0,0 +1,10 @@
+object Const { final val VAL = 1 ; final val VAR = 2 }
+import Const._
+object Test {
+	def test(i: Int) = i match { case v @ (VAR | VAL) => v == VAR case _ => "default" }
+	def main(args: Array[String]) {
+		println(test(VAR))
+		println(test(VAL))
+		println(test(-1))
+	}
+}
```
retronym: A test for the regression so we catch it before the bootstrap integration test.

hrhino (approved):
Yikes. Sounds good.

PR #6217, opened 2017-12-06
Backtick underscore in REPL

som-snytt:
Since underscore can be a member, it must be backticked
in a path such as `X._`.

Probably the underscore should be disallowed in the related ticket.

Also note the behavior in an import path, where underscore is a special identifier even if backticked.

The underscore was specifically not backticked; possibly there was a change in behavior?

Fixes scala/bug#10645

### DIFF
--- a/src/repl/scala/tools/nsc/interpreter/IMain.scala
+++ b/src/repl/scala/tools/nsc/interpreter/IMain.scala
@@ -287,7 +287,7 @@ class IMain(initialSettings: Settings, protected val out: JPrintWriter) extends 
 
   def backticked(s: String): String = (
     (s split '.').toList map {
-      case "_"                               => "_"
+      case "_"                               => "`_`"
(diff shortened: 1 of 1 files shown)


dwijnand:
Oh! Went to https://github.com/scala/bug/issues/10645 to see if I could fix it and saw that you'd submitted a PR already. Wow.

I think this could do with a REPL test, including the variant when using class-based mode. That way we can verify it will continue to work in the future, particularly when 2.12.x is later merged into 2.13.x where there have been big REPL changes.

retronym:
We'll probably won't merge this one forward and will need another PR to 2.13.x. But this change looks okay to me to merge as is, feel free to submit a test as a followup though.

som-snytt:
I don't want to get anyone in trouble, but this one seems to have made its way forward and won't need a forward port.

PR #6218, opened 2017-12-07
Import selector respects backquotes [ci: last-only]

### LINKED ISSUE #10641: Illegal import rename of a wildcard
scalac allows one to rename wildcard to an arbitrary identifier:

```scala
scala> object Foo { class A }
defined object Foo

scala> import Foo.{_ => Bla}
import Foo._

// A is available because the import is a wildcard one despite the rename!
scala> new A
res1: Foo.A = Foo$A@687a0e40

// Bla is not available
scala> new Bla
<console>:16: error: not found: type Bla
       new Bla
```

This should be ruled out by the parser. dottyc does rule it out at parsing time.

som-snytt:
An import selector from WILDCARD to WILDCARD is taken as importing member underscore, and WILDCARD to anything else not null as renaming it. Renaming unquoted wildcard is not allowed.

was:
To disambiguate backquoted underscore, the convention
for ImportSelector is tweaked by using EMPTY instead of
WILDCARD as the special value. So EMPTY => EMPTY is
wildcard import, nonempty => EMPTY is hidden.

This encoding also avoids a null value for the rename field.

A rename is detected more simply by name != rename.

This change would require communication with anyone who builds ImportSelector manually. It would be nice to use ImportSelector.wild instead of relying on the encoding. See the test that does that.

Fixes scala/bug#6426
Fixes scala/bug#10641
Fixes scala/bug#7691

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -73,16 +73,12 @@ trait Contexts { self: Analyzer =>
     for (imps <- allImportInfos.remove(unit)) {
       for (imp <- imps.distinct.reverse) {
         val used = allUsedSelectors(imp)
-        for (sel <- imp.tree.selectors if !isMaskImport(sel) && !used(sel))
+        for (sel <- imp.tree.selectors if !sel.isMask && !used(sel))
           reporter.warning(imp.posOf(sel), "Unused import")
       }
       allUsedSelectors --= imps
     }
   }
-
-  def isMaskImport(s: ImportSelector): Boolean = s.name != nme.WILDCARD && s.rename == nme.WILDCARD
-  def isIndividualImport(s: ImportSelector): Boolean = s.name != nme.WILDCARD && s.rename != nme.WILDCARD
-  def isWildcardImport(s: ImportSelector): Boolean = s.name == nme.WILDCARD
 
   var lastAccessCheckDetails: String = ""
 
@@ -1021,16 +1017,16 @@ trait Contexts { self: Analyzer =>
       def collect(sels: List[ImportSelector]): List[ImplicitInfo] = sels match {
         case List() =>
           List()
-        case List(ImportSelector(nme.WILDCARD, _, _, _)) =>
+        case sel :: Nil if sel.isWildcard =>
           // Using pre.implicitMembers seems to exposes a problem with out-dated symbols in the IDE,
           // see the example in https://www.assembla.com/spaces/scala-ide/tickets/1002552#/activity/ticket
           // I haven't been able to boil that down the an automated test yet.
           // Looking up implicit members in the package, rather than package object, here is at least
           // consistent with what is done just below for named imports.
           collectImplicits(qual.tpe.implicitMembers, pre, imported = true)
-        case ImportSelector(from, _, to, _) :: sels1 =>
-          var impls = collect(sels1) filter (info => info.name != from)
-          if (to != nme.WILDCARD) {
+        case (sel @ ImportSelector(from, _, to, _)) :: sels1 =>
+          var impls = collect(sels1).filter(info => info.name != from)
+          if (!sel.isMask) {
             withQualifyingImplicitAlternatives(imp, to, pre) { sym =>
               impls = new ImplicitInfo(to, pre, sym) :: impls
             }
@@ -1634,8 +1630,7 @@ trait Contexts { self: Analyzer =>
     }
 
     /** Is name imported explicitly, not via wildcard? */
-    def isExplicitImport(name: Name): Boolean =
-      tree.selectors exists (_.rename == name.toTermName)
+    def isExplicitImport(name: Name): Boolean = tree.selectors.exists(_.introduces(name.toTermName))
 
     /** The symbol with name `name` imported from import clause `tree`. */
     def importedSymbol(name: Name): Symbol = importedSymbol(name, requireExplicit = false, record = true)
@@ -1656,12 +1651,13 @@ trait Contexts { self: Analyzer =>
       var selectors = tree.selectors
       def current = selectors.head
       while ((selectors ne Nil) && result == NoSymbol) {
-        if (current.rename == name.toTermName)
-          result = qual.tpe.nonLocalMember( // new to address #2733: consider only non-local members for imports
-            if (name.isTypeName) current.name.toTypeName else current.name)
-        else if (current.name == name.toTermName)
+        if (current.introduces(name.toTermName))
+          result = qual.tpe.nonLocalMember( // #2733: consider only non-local members for imports
+            if (name.isTypeName) current.name.toTypeName else current.name
+          )
+        else if (!current.isWildcard && current.name == name.toTermName)
           renamed = true
-        else if (current.name == nme.WILDCARD && !renamed && !requireExplicit)
+        else if (current.isWildcard && !renamed && !requireExplicit)
           result = qual.tpe.nonLocalMember(name)
 
         if (result == NoSymbol)
@@ -1680,19 +1676,19 @@ trait Contexts { self: Analyzer =>
       else NoSymbol
     }
     private def selectorString(s: ImportSelector): String = {
-      if (s.name == nme.WILDCARD && s.rename == null) "_"
-      else if (s.name == s.rename) "" + s.name
-      else s.name + " => " + s.rename
+      if (s.isWildcard) "_"
+      else if (s.isRename) s.name + " => " + s.rename
+      else "" + s.name
     }
 
     def allImportedSymbols: Iterable[Symbol] =
       importableMembers(qual.tpe) flatMap (transformImport(tree.selectors, _))
 
     private def transformImport(selectors: List[ImportSelector], sym: Symbol): List[Symbol] = selectors match {
-      case List() => List()
-      case List(ImportSelector(nme.WILDCARD, _, _, _)) => List(sym)
-      case ImportSelector(from, _, to, _) :: _ if from == sym.name =>
-        if (to == nme.WILDCARD) List()
+      case Nil => Nil
+      case sel :: Nil if sel.isWildcard => List(sym)
+      case (sel @ ImportSelector(from, _, to, _)) :: _ if from == sym.name =>
+        if (sel.isMask) Nil
         else List(sym.cloneSymbol(sym.owner, sym.rawflags, to))
       case _ :: rest => transformImport(rest, sym)
     }
--- a/src/compiler/scala/tools/nsc/typechecker/Namers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Namers.scala
@@ -561,7 +561,7 @@ trait Namers extends MethodSynthesis {
         def isValid(original: Name) =
           original.bothNames forall (x => (base nonLocalMember x) == NoSymbol)
 
-        if (from != nme.WILDCARD && base != ErrorType) {
+        if (!s.isWildcard && base != ErrorType) {
           if (isValid(from)) {
             // for Java code importing Scala objects
             if (!nme.isModuleName(from) || isValid(from.dropModule)) {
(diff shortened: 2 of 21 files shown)


som-snytt:
A couple of fixes for reflection. I was inclined to add API for `ImportSelector` such as `isWildcard` and `isHidden`. The conventional construction is distributed far and wide, for testing and making wildcards etc. Note that underscore remains _ in WildcardSelectorRepr and friends, because `pq"_"`.

dwijnand:
Btw, you need to say "fixes xx/yy#zz" for each xx/yy#zz, otherwise it'll just close the first one.

som-snytt:
Also fix `ImportInto.isExplicitImport` by adding that API to `ImportSelector`.

retronym:
I'd be most worried about compatibility with Scala IDE and Scala refactoring.

```
⚡ git grep WILDCARD | grep -i import
src/main/scala/scala/tools/refactoring/analysis/ImportAnalysis.scala:      imp.expr.toString() + "." + nme.WILDCARD.decoded
src/main/scala/scala/tools/refactoring/analysis/ImportAnalysis.scala:      val symImpStr = (nme.WILDCARD.decoded :: parts).reverse.mkString(".")
src/main/scala/scala/tools/refactoring/analysis/ImportAnalysis.scala:      if (sel.name == nme.WILDCARD)
src/main/scala/scala/tools/refactoring/analysis/ImportsToolbox.scala:        if (sel.name == nme.WILDCARD) List(mkName(thatQual))
src/main/scala/scala/tools/refactoring/implementations/ImportsHelper.scala:                          selector.name == nme.WILDCARD || {
src/main/scala/scala/tools/refactoring/implementations/UnusedImportsFinder.scala:  def wildcardImport(i: ImportSelector) = i.name == nme.WILDCARD
src/main/scala/scala/tools/refactoring/implementations/UnusedImportsFinder.scala:    if (s.rename == nme.WILDCARD) return true
src/main/scala/scala/tools/refactoring/implementations/oimports/ImportParticipants.scala:        importSelector.rename == nme.WILDCARD
src/main/scala/scala/tools/refactoring/implementations/oimports/ImportParticipants.scala:      importSelector.name == nme.WILDCARD || (allTypeNamesFromSelect(select) & allTypeNamesFromImportSelector(importSelector)).nonEmpty
src/main/scala/scala/tools/refactoring/implementations/oimports/ImportParticipants.scala:      val wild = imp.selectors.find(_.name == nme.WILDCARD)
src/main/scala/scala/tools/refactoring/implementations/oimports/ImportParticipants.scala:        val (wild, rest) = imports.partition(_.selectors.exists(_.name == nme.WILDCARD))
src/main/scala/scala/tools/refactoring/implementations/oimports/ImportParticipants.scala:            Some(imp.copy(qual, List(ImportSelector(nme.WILDCARD, -1, nme.WILDCARD, -1))).copyAttrs(imp))
src/main/scala/scala/tools/refactoring/implementations/oimports/RegionTransformations.scala:        importName.endsWith(selector.name.decoded) && selector.rename == nme.WILDCARD
src/main/scala/scala/tools/refactoring/implementations/oimports/RegionTransformations.scala:          selector.name == nme.WILDCARD || importsNames.exists { importName =>
src/main/scala/scala/tools/refactoring/implementations/oimports/RegionTransformations.scala:              case imp @ RegionImport(expr, sels) if sels.exists { _.name == nme.WILDCARD } =>
src/main/scala/scala/tools/refactoring/implementations/oimports/TreeToolbox.scala:              rsels.exists { _.name == global.nme.WILDCARD } && (treeComparables.isSameWithSymbols(true)(expr.symbol, rexpr.symbol) || treeComparables.isSameExprByName(true)(expr, rexpr)) ||
```

I wonder if we can find a way to leave the encoding as is, but sneak an extra boolean onto `ImportSelector` to identify the problematic (and esoteric) case of `import foo.{`_` => bar}`. (Are there more problematic cases than that?)

som-snytt:
Yes, for sure, I meant tooling needs communication, but that it would nicer if there were an API to make the change easier and more uniform, too. I could experiment on refactoring.

I think my first thought was a tree attachment.

adriaanm:
Let's give this some time for communication. Moving to M4.

SethTisue:
needs rebase

som-snytt:
@SethTisue Copy that. (On Flashpoint, a Canadian police procedural with all-Canadian cast, they half-smirk and say, "Copy that," all the time. "Joe won't make the BBQ at 6." "Copy that.")

SethTisue:
ping @som-snytt

dwijnand:
Refactoring requested and new commits pushed. Re-review by @retronym, perhaps?

adriaanm:
I like using `EMPTY` for the target of a rename that intends to hide a name (`a => _`), but don't understand why we can't represent a final `_` import internally as (`_ => _`). It seems that would yield a smaller delta in the internal representation.

The caution about breaking tooling remains, but I'm fine moving forward as long as scalafix etc is not affected. (Which seems unlikely since they reimplemented all this ;-))

src/reflect/scala/reflect/api/Trees.scala
```diff
@@ -802,13 +802,13 @@ trait Trees { self: Universe =>
...
-   *  The last (and maybe only name) may be a nme.WILDCARD. For instance:
+   *  The last (and maybe only name) may be a nme.EMPTY. For instance:
    *
-   *    import qual.{x, y => z, _}
+   *    import qual.{w => _, x, y => z, _}
    *
    *  Would be represented as:
    *
-   *    Import(qual, List(("x", "x"), ("y", "z"), (WILDCARD, null)))
+   *    Import(qual, List(("w", EMPTY), ("x", "x"), ("y", "z"), (EMPTY, EMPTY)))
```
adriaanm: As mentioned in the discussion, could we use `(WILDCARD, WILDCARD)` for the last element?

adriaanm (changes requested):

SethTisue:
@som-snytt do you plan to resume work on this?

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -574,7 +574,7 @@ trait Namers extends MethodSynthesis {
         def loop(xs: List[Name]): Unit = xs match {
           case Nil      => ()
           case hd :: tl =>
-            if (hd == nme.WILDCARD || !(tl contains hd)) loop(tl)
+            if (hd == nme.EMPTY || !(tl contains hd)) loop(tl)
             else DuplicatesError(tree, hd, check)
         }
         loop(names filterNot (x => x == null || x == nme.WILDCARD))
```
som-snytt: I think you missed this one.

som-snytt:
I unwound the change in representation and, as I suggested in December, added `selector.isMask` and friends to encapsulate.

Although the previous delta was most regular, and I regret that it wasn't adopted even though Scala-IDE is in a vegetative state, now I'll just add support for what is currently unsupported,
```
import x.`_`, y.{`_` => Underscore}
```
where `ImportSelector(WILDCARD, _, Underscore, _)`, i.e., wildcard import has null rename, anything else imports a member named `_`. Maybe that will be disallowed soon, like Unicode escapes.

src/reflect/scala/reflect/api/Trees.scala
som-snytt: Now (WILD, WILD) means import backticked underscore, possibly with rename. Otherwise, the encoding is unchanged.

src/compiler/scala/tools/nsc/typechecker/Namers.scala
```diff
@@ -580,21 +580,22 @@ trait Namers extends MethodSynthesis {
...
+      def noDuplicates(): Unit = {
+        def loop(xs: List[ImportSelector]): Unit = xs match {
           case Nil      => ()
           case hd :: tl =>
-            if (hd == nme.WILDCARD || !(tl contains hd)) loop(tl)
-            else DuplicatesError(tree, hd, check)
+            if (!hd.isWildcard && tl.exists(x => !x.isWildcard && x.name == hd.name))
+              DuplicatesError(tree, hd.name, RenamedTwice)
+            else if (hd.isRename && tl.exists(x => x.isRename && x.rename == hd.rename))
+              DuplicatesError(tree, hd.name, AppearsTwice)
```
som-snytt: `hd.rename`

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -1656,12 +1651,13 @@ trait Contexts { self: Analyzer =>
...
       while ((selectors ne Nil) && result == NoSymbol) {
-        if (current.rename == name.toTermName)
-          result = qual.tpe.nonLocalMember( // new to address #2733: consider only non-local members for imports
-            if (name.isTypeName) current.name.toTypeName else current.name)
-        else if (current.name == name.toTermName)
+        if (current.introduces(name.toTermName))
+          result = qual.tpe.nonLocalMember( // #2733: consider only non-local members for imports
+            if (name.isTypeName) current.name.toTypeName else current.name
+          )
+        else if (!current.isWildcard && current.name == name.toTermName)
```
som-snytt, line 1658: i.e., if name is underscore, don't count wildcard as rename.

som-snytt:
Someone may have mentioned that REPL doesn't preserve backticks.
```
scala> object X { def `_` = 42 }
defined object X

scala> import X.{`_` => _, _}
import X.{_, _}

scala> `_`

       import X.{_, _}
                 ^
<synthetic>:4: error: Wildcard import must be in last position

(To diagnose errors in synthetic code, try adding `// show` to the end of your input.)
res0: Int = 42

scala> `_` // show

       import X.{_, _}
                 ^
<synthetic>:4: error: Wildcard import must be in last position
```
I'm not sure how `res0` completes successfully after erroring. "I'm not sure" is how academics and engineers convey, "I have no idea." If they have any inkling at all, they feel quite sure of themselves.

som-snytt:
The previous commits went red after rebasing, because of test/macro-annot.

I see that's macro-paradise, which has more hacky breakage of encapsulation.

som-snytt:
Squashed, and added tweak to paradise namer, analogous to Context. I think that fixes one of the three paradise test failures.

This PR was totally not worth it.

som-snytt:
@adriaanm This is totally incrementable and useful and unoffensive and could be moved up to 2.13.0 cycle. I was looking at imports today thinking I thought I'd changed that code.

adriaanm (approved):
💯 the code is much, much nicer! encapsulation FTW
🏅 for moxie
🃏 because wild card

SethTisue:
`implicit val _ = ...` used to work and now doesn't. that's as intended, but [it would be nice if the compiler issued a warning](https://github.com/scala/bug/issues/11618)

eed3si9n:
Thanks for fixing this. Maybe we could mention in https://github.com/scala/scala/releases/tag/v2.13.0 that this allows multiple dummy binding.

What do you think about the following edit?

```diff
  * Underscore is no longer a legal identifier unless backquoted ([bug\#10384](https://github.com/scala/bug/issues/10384))
+   * Multiple dummy bindings can now be used within a scope: `{ val _ = 3; val _ = 5; 7 }`
    * `val _ =` is now a pattern match
    * `implicit val _ =` is also now a pattern match which does not introduce an identifier and therefore does not add anything to implicit scope
```

SethTisue:
@eed3si9n hmm... not against adding some wording, but I think we should make it very clear to the reader why we're telling them this. (which the existing text isn't necessarily doing an ideal job of, either)

what is the recommended way to value-discard, warning-free? is it `(...: Unit)`, or `val _ = ...`? I'm not sure which I like better, style-wise.

eed3si9n:
Between the two, `val _ = ....` I think is less surprising to the person reading the code:

```scala
val _ = compile.value
val _ = copyResources.value
```

`(...: Unit)` has similar vibe to `final val` in a sense that all the tokens make sense, but the intent is opaque unless you're in the close circle of people who's been watching the development of 2.13 closely.

som-snytt:
> what is the recommended way to value-discard, warning-free? 

`@nowarn` or `-Wconf`

"Warn-free, as free as the wind blows..." :notes:

SethTisue:
okay, I changed that section of the notes to:

* Underscore is no longer a legal identifier unless backquoted ([bug\#10384](https://github.com/scala/bug/issues/10384))
  * `val _ =` is now a pattern match (and discards the value without incurring a warning)
  * `implicit val _ =` is also now a pattern match (and is useless, because it no longer adds to implicit scope)

(The fact that as an additional consequence it is now allowed to `val _ = ...` multiple times seems like unnecessary extra detail to me.)

som-snytt:
Also, from now on, we should call them the release :notes: or is that already the name of Seth's band?

dwijnand:
> what is the recommended way to value-discard, warning-free? is it `(...: Unit)`, or `val _ = ...`? I'm not sure which I like better, style-wise.

#7563 makes `(...: Unit)` the official way to avoid warnings.

PR #6219, opened 2017-12-07
emit parameter names for static forwarders

ghik:
Fixes https://github.com/scala/bug/issues/10650


ghik:
Test failures seem to be unrelated to my changes, they fail on `2.12.x` as well when I run them on my local machine.

SethTisue:
test failures reported at https://github.com/scala/scala-dev/issues/457

test/files/run/t10650b/Test_2.scala
```diff
@@ -0,0 +1,13 @@
+object Test extends App {
```
lrytz: I don't see any additional value in also testing this under separate compilation - maybe I'm missing something?

ghik: You may be right. I simply wrote tests similar to the ones for https://github.com/scala/bug/issues/9437

ghik: So should I remove this test? If so, then it probably also makes sense to remove `t9437b`.

lrytz: Ah I see. I don't see the value in separate compilation in the existing tests either. When using a named paramter (``a = 1, `_` = "2"`` etc) in Scala code on a Scala-defined method, the Scala compiler doesn't look at the classfile attribute holding the parameter names. The information is obtained from the "pickle", the serialized chunk of the symbol table. So I'd say let's remove the additional test in this PR.

lrytz: Yes, let's remove those tests.

lrytz:
You can squash all in one commit, and also rebase on current 2.12.x to fix the build failure.

ghik:
All seems good.

lrytz (approved):
Thanks!

PR #6220, opened 2017-12-08
Propagate fatal errors during macro expansion. Fixes #10552.

olafurpg:
Previously, fatal errors got swallowed during macro expansion.  In the
case of implicit blackbox macros, a fatal error (for example
OutOfMemoryException) got reported with the message "exception during
macro expansion". For implicit whitebox macros the fatal error is not
even reported unless -Xlog-implicits is enabled. By default, the user
only sees a cryptic "implicit not found" error message.  See #10649.

This commit changes the error handling of exceptions during macro
expansion to propagate fatal errors. Now fatal errors are left uncaught
and crash compilation with a full stack trace instead of getting
swallowed.

```
error: java.lang.OutOfMemoryError
    at Macros$BlackBox$.materializeImpl(so.scala:8)
```

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Macros.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Macros.scala
@@ -13,6 +13,7 @@ import scala.reflect.internal.util.ListOfNil
 import scala.reflect.macros.runtime.{AbortMacroException, MacroRuntimes}
 import scala.reflect.macros.compiler.DefaultMacroCompiler
 import scala.tools.reflect.FastTrack
+import scala.util.control.NonFatal
 import Fingerprint._
 
 /**
@@ -815,7 +816,8 @@ trait Macros extends MacroRuntimes with Traces with Helpers {
               case ex: AbortMacroException => MacroGeneratedAbort(expandee, ex)
               case ex: ControlThrowable => throw ex
               case ex: TypeError => MacroGeneratedTypeError(expandee, ex)
-              case _ => MacroGeneratedException(expandee, realex)
+              case NonFatal(_) => MacroGeneratedException(expandee, realex)
+              case fatal => throw fatal
             }
         } finally {
           expandee.removeAttachment[MacroRuntimeAttachment]
--- a/test/files/neg/t10552.check
+++ b/test/files/neg/t10552.check
@@ -1,0 +1,1 @@
+error: OOM
--- a/test/files/neg/t10552/Macros_1.scala
+++ b/test/files/neg/t10552/Macros_1.scala
@@ -1,0 +1,7 @@
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+object A {
+  def f: Unit = macro f_impl
+  implicit def f_impl(c: whitebox.Context): c.Expr[Unit] =
+    throw new OutOfMemoryError("OOM") with scala.util.control.NoStackTrace
+}
--- a/test/files/neg/t10552/Test_2.scala
+++ b/test/files/neg/t10552/Test_2.scala
@@ -1,0 +1,3 @@
+object B {
+  def t = A.f
+}
--- a/test/files/run/macro-sip19-revised/Impls_Macros_1.scala
+++ b/test/files/run/macro-sip19-revised/Impls_Macros_1.scala
@@ -4,6 +4,10 @@ object Macros {
   def impl(c: Context) = {
     import c.universe._
 
+    val thisMacro = c.macroApplication.symbol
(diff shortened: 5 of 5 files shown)


olafurpg:
Is the CI is failing due to this patch or is it spurious? I was able to see 
```
[error] 1 of 12 test tasks failed:
[error] - partest pos neg jvm
[error]   - test/it:testOnly failed: sbt.TestsFailedException: Tests unsuccessful
java.lang.RuntimeException
```
from the log https://scala-ci.typesafe.com/job/scala-2.12.x-validate-test/6094/consoleText but I can't find the failing test.

hrhino:
Your error is:

```
!!  321 - run/macro-sip19-revised                   [compilation failed]
##### Log file '/home/jenkins/workspace/scala-2.12.x-validate-test/test/files/run/macro-sip19-revised-run.log' from failed test #####

error: null
```

which indicates an ~NPE~ SOE somewhere. 

I can make it happen locally, with `partest test/files/run/macro-sip19-revised`. 

Maybe that test regularly throws ~NPEs~ SOEs, which were being accidentally caught?

olafurpg:
Thanks @hrhino I can reproduce the error locally with `partest`. The macro triggers implicit search on `SourceLocation` which recursively calls the macro and causes a stack overflow error. I think the new behavior makes more sense, compilation crashes with a SOE with a trace pointing to the culprit line that triggers the recursive implicit search
```
	at scala.reflect.macros.contexts.Context.inferImplicitValue(Context.scala:6)
	at Macros$.impl(Impls_Macros_1.scala:7)
```
https://github.com/scala/scala/blob/a45586102d9ff839a6c3a037ad1eaf6ebeaa87d6/test/files/run/macro-sip19-revised/Impls_Macros_1.scala#L7

This test is for a rejected SIP https://docs.scala-lang.org/sips/source-locations.html I wonder if the best step is to remove it?

joroKr21:
Looks like the test shows how source locations can be implemented with a macro. But there is a later commit c539ae2f56fe9f565cffb4afd6ab131bda89acb7 that disables the divergence check for macros. It's probably at this point that the reason for `c.inferImplicitValue` failing changed from a divergent implicit error to a stack overflow :laughing:

joroKr21:
Anyway, you can add this at the beginning of the macro to break out of the diverging expansion:
```scala
val thisMacro = c.macroApplication.symbol
val depth = c.enclosingMacros.count(_.macroApplication.symbol == thisMacro)
if (depth > 1) c.abort(c.enclosingPosition, "") // avoid SO
```

olafurpg:
Thanks @joroKr21 I updated the test following your advice and it passes now for me locally. I think it might still be worth removing this test.

olafurpg:
Ping. Any chance to get a review on this PR? This could be considered a breaking change for macros that rely on stack overflows for control flow. However, I think incorrect handling of fatal errors should not protected by the public macro API.

lrytz (commented):
Could you add a test? There are existing `neg` tests for macros, for example `test/files/neg/t6931`.

Suggested two files:

```
import scala.language.experimental.macros
import scala.reflect.macros.blackbox.Context
object A {
  def f: Unit = macro f_impl
  def f_impl(c: Context): c.Expr[Unit] = throw new OutOfMemoryError("E") with scala.util.control.NoStackTrace
}
```

```
object B {
  def t = A.f
}
```

This should give a deterministic output.

olafurpg:
Thank you for the review @lrytz I added a neg test but I'm unable to get it to pass

```
> partest test/files/neg/t10552 --update-check --verbose --show-log --show-diff

# starting 1 test in neg
% scalac t10552/Macros_1.scala
% scalac t10552/Test_2.scala
!! 1 - neg/t10552                                [compilation failed]
##### Log file '/Users/ollie/dev/scala/test/files/neg/t10552-neg.log' from failed test #####

error: OOM

# 0/1 passed, 1 failed in neg

##### Transcripts from failed tests #####

# partest /Users/ollie/dev/scala/test/files/neg/t10552
% scalac t10552/Macros_1.scala
% scalac t10552/Test_2.scala
error: OOM
```

I tried to follow the structure of the other neg tests. I expected the full stack trace to appear to appear in the .check file but the only output is `error: OOM`.

joroKr21:
It's expected that there is no stack trace because of `scala.util.control.NoStackTrace`

lrytz:
So it seem partest treats a crashing compiler as a failed test, also in `neg`...

Here's a solution: https://github.com/olafurpg/scala/compare/10552...lrytz:pr6220?expand=1

lrytz:
Submitted with the updated test as https://github.com/scala/scala/pull/6344

PR #6221, opened 2017-12-08
Revert "optimise use of indyLamdaMethods map"

lrytz:
This reverts commit fe0165c6863a64accea4c6c87c2af5fc1c79d368.
See scala/scala-dev#457


lrytz:
@mkeskells I think the easiest is to revert until you sumbit a revised version.

PR #6222, opened 2017-12-08
Compare positions in tryTypedApply by focus.

hrhino:
Synthetic trees usually get offset positions, even with range positions enabled. The comparison previously used by `errorInResult` did not consider the error issued by `AdaptTypeError` on an `ApplyToImplicitArgs` to be part of the result expression, meaning that an implicit view wouldn't be sought on a second try.

Fixes scala/bug#10643.

### DIFF
--- a/src/reflect/scala/reflect/internal/util/Position.scala
+++ b/src/reflect/scala/reflect/internal/util/Position.scala
@@ -94,6 +94,8 @@ sealed abstract class UndefinedPosition extends Position {
   override def start           = fail("start")
   override def point           = fail("point")
   override def end             = fail("end")
+
+  override def samePointAs(that: Position) = false
 }
 
 private[util] trait InternalPositionImpl {
@@ -199,6 +201,10 @@ private[util] trait InternalPositionImpl {
     else if (isDefined) s"[$point]"
     else "[NoPosition]"
   )
+
+  /* Same as `this.focus == that.focus`, but less allocation-y. */
+  def samePointAs(that: Position): Boolean =
+    that.isDefined && this.point == that.point && this.source.file == that.source.file
 
   private def asOffset(point: Int): Position = Position.offset(source, point)
   private def copyRange(source: SourceFile = source, start: Int = start, point: Int = point, end: Int = end): Position =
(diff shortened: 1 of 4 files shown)


hrhino:
This makes me suspect that `==` on positions is rarely desired in the presence of `-Yrangepos`. Not sure how many bugs are out there (most seem to be `somePos == NoPosition`), but a quick survey shows two comparisons in `normalTypeApply` (to decide what advice to show) as using `==` on `Position`. Accordingly, `neg/t9834.scala`, `neg/t0903.scala`, and `neg/t1215.scala` all fail with range positions enabled.

som-snytt:
This was a good one, sorry I didn't notice. This year I learned that I have to bootstrap, otherwise I don't know anything. Now I know I have to rangepos, too.

hrhino:
I [learned](https://github.com/scala/scala/pull/6216) you need to bootstrap with `-Xsource:2.13`, too. Yay.

I hate the thing that happens when my PRs lead me down a rabbit hole of vaguely-related issues. Possibly I'll fix those test cases (with rangepos) in a separate PR once this is merged.

hrhino:
Apropos of scala/scala-dev#430: Those three tests are the only tests that fail for legitimate reasons with `-Yrangepos` enabled (by my reckoning). A few more fail due to differences in printing range positions, and the remaining few fail due to differently-ordered error messages.

som-snytt:
`partest` has a crude filtering mechanism for check files. A line `#partest flag` says include the subsequent block if flag is set. It was originally for java version and `-optimise`, not used much now.

src/reflect/scala/reflect/internal/util/Position.scala
```diff
@@ -94,6 +94,8 @@ sealed abstract class UndefinedPosition extends Position {
   override def start           = fail("start")
   override def point           = fail("point")
   override def end             = fail("end")
+
+  override def samePointAs(that: Position) = !that.isDefined
```
hrhino: Coming back to this, I waffle slightly on this definition. I'm willing to be convinced that the RHS should be `false`.

lrytz: `false` sounds more reasonable.

lrytz (commented):
Looks good otherwise!

src/reflect/scala/reflect/internal/util/Position.scala
hrhino: You've convinced me.

lrytz (approved):

PR #6223, opened 2017-12-11
allow per-run init to be used with java collections

mkeskells:
change spun out of https://github.com/scala/scala/pull/6124 to aid review

add tests for per run caches, for java and scala collections

add fixes to java collections used in per-run, as detected in https://github.com/scala/scala-dev/issues/457

### DIFF
--- a/src/reflect/scala/reflect/internal/SymbolTable.scala
+++ b/src/reflect/scala/reflect/internal/SymbolTable.scala
@@ -365,9 +365,15 @@ abstract class SymbolTable extends macros.Universe
     // letting us know when a cache is really out of commission.
     import java.lang.ref.WeakReference
     private var caches = List[WeakReference[Clearable]]()
+    private var javaCaches = List[JavaClearable[_]]()
 
     def recordCache[T <: Clearable](cache: T): T = {
-      caches ::= new WeakReference(cache)
+      cache match {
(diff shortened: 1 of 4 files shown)


src/reflect/scala/reflect/internal/SymbolTable.scala
```diff
@@ -363,11 +363,17 @@ abstract class SymbolTable extends macros.Universe
 
     // Weak references so the garbage collector will take care of
     // letting us know when a cache is really out of commission.
-    import java.lang.ref.WeakReference
+    import java.lang.ref.{WeakReference, Reference}
```
lrytz: unused import?

lrytz (commented):
looks good - i was keeping it in the back of my mind for a week but couldn't come up with a "nicer" solution. could you squash the commits into one and update the PR title and description?

src/reflect/scala/reflect/internal/SymbolTable.scala
mkeskells: fixed in https://github.com/scala/scala/pull/6223/commits/abef11199b9e140e7241fa17b3f5e08ea117c453

mkeskells:
squashed and updated as requested

lrytz (approved):

PR #6224, opened 2017-12-12
use ":" as separator for sbt ScalaVersionSetting

martijnhoekstra:


martijnhoekstra:
Review @SethTisue who alerted me as to where the issue was I think

SethTisue:
thanks! this makes e.g. `scalac -Xsource:2.13` work from the sbt prompt

SethTisue:
hmm, this could have targeted 2.12, I think, seconds too late...

martijnhoekstra:
should it have? I thought 2.13 was the default PR target with stuff getting backported to 2.12 - I actually had it against 2.12 originally, and then went back to fix it D:

SethTisue:
> should it have?

yes. safe changes can still target 2.12.x directly. we continue to merge everything forward from 2.12.x to 2.13.x. Adriaan just made 2.13.x the default to signal that's where actual new work should go.

SethTisue:
backport (#6225) merged

PR #6226, opened 2017-12-14
Experiment: don't consider prefix for implicit scope

adriaanm:
Experiment for https://github.com/scala/scala-dev/issues/446

Let's run a community build and benchmarks (Seth and Jason, please start those jobs)

### DIFF
--- a/src/compiler/scala/tools/nsc/backend/jvm/BCodeIdiomatic.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BCodeIdiomatic.scala
@@ -28,6 +28,8 @@ abstract class BCodeIdiomatic {
   import bTypes._
   import coreBTypes._
   import genBCode.postProcessor.callGraph.callsitePositions
+  import BackendReporting.RightBiasedEither
+
 
   lazy val JavaStringBuilderClassName = jlStringBuilderRef.internalName
 
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromClassfile.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromClassfile.scala
@@ -15,6 +15,8 @@ abstract class BTypesFromClassfile {
   import bTypes._
   import coreBTypes._
   import frontendAccess.compilerSettings
+  import BackendReporting.RightBiasedEither
+
 
   /**
    * Obtain the BType for a type descriptor or internal name. For class descriptors, the ClassBType
--- a/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
+++ b/src/compiler/scala/tools/nsc/backend/jvm/BTypesFromSymbols.scala
@@ -24,6 +24,7 @@ abstract class BTypesFromSymbols[G <: Global](val global: G) extends BTypes {
   import genBCode._
   import codeGen.CodeGenImpl._
   import postProcessor.{bTypesFromClassfile, byteCodeRepository}
+  import BackendReporting.RightBiasedEither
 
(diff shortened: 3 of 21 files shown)


retronym:
This doesn't bootstrap.

```
[error] /home/jenkins/workspace/bootstrap-benchmark/src/reflect/scala/reflect/internal/Definitions.scala:836: type mismatch;
[error]  found   : scala.reflect.internal.settings.MutableSettings#BooleanSetting
[error]  required: Boolean
[error]     private[this] val doSam = settings.isScala212 || (settings.isScala211 && settings.Xexperimental)
```
(etc)

This is an interesting case.

```scala
abstract class MutableSettings extends AbsSettings {

  type Setting <: SettingValue
  type BooleanSetting <: Setting { type T = Boolean }
  type IntSetting <: Setting { type T = Int }
  type MultiStringSetting <: Setting { type T = List[String] }
...
}

object MutableSettings {
  import scala.language.implicitConversions
  /** Support the common use case, `if (settings.debug) println("Hello, martin.")` */
  @inline implicit def reflectSettingToBoolean(s: MutableSettings#BooleanSetting): Boolean = s.value
}

```

`BooleanSetting` can't have a companion module (as it isn't a class, it is an abstract type.)

I can run the performance tests if we modify this PR to predicate the change in behaviour on a `-Y` options.

adriaanm:
Thanks. I had to run. Will see about a better fix tomorrow

som-snytt:
I think there's an issue, probably closed, about this intersection of type aliases and conversions. The boolean setting conversion was one of my first contributions. Sorry it's taking your time years later.

adriaanm:
It's a great test case! I think the problem lies with this code:
https://github.com/scala/scala/blob/78d3de8e46811a79c324f853743a777bcf4cea67/src/compiler/scala/tools/nsc/typechecker/Implicits.scala#L1067-L1074

It doesn't consider companion object implicits unless the prefix is stable, which it isn't in this case

adriaanm:
The prefix does have to be stable (or static), because we need to use it to select the implicit member...

adriaanm:
Repro:

```scala
abstract class MutableSettings {
  object Setting {
    import scala.language.implicitConversions
    /** Support the common use case, `if (settings.debug) println("Hello, martin.")` */
    @inline implicit def reflectSettingToBoolean(s: BooleanSetting): Boolean = s.value
  }

  class Setting[T](val value: T)

  class BooleanSetting extends Setting[Boolean]
}


object Test {
  object settings extends MutableSettings

  import settings._
  val debug: MutableSettings#BooleanSetting = new BooleanSetting
  if (debug) println("woele")
}

```

adriaanm:
Now more boo-strappy

SethTisue:
@adriaanm if you rebase it, I can do the community build run

adriaanm:
Rebased

SethTisue:
I'll do the community build run after we :shipit: 2.12.6

SethTisue:
in progress: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/2781/consoleFull

SethTisue:
```
[info] >>> The dbuild result is----------------: FAILED (failed: akka-actor, scalaz, scalameta-2, scalameta-1, scallop)
```

note that having scalaz fail knocks out a lot of the build, primarily because of specs2's scalaz dependency. (as an aside, it would be nice to push the specs2 3.9->4.0 upgrade through the 2.12 build, since 4.0 drops the scalaz dependency. the 2.13 build is already on 4.0.)

SethTisue:
> (as an aside, it would be nice to push the specs2 3.9->4.0 upgrade through the 2.12 build

(as an aside to an aside, this happened)

PR #6231, opened 2017-12-17
Extra caution in namer treating parents

som-snytt:
Parents have not been validated yet, so only use `addChild`
when `isClass`.

Fixes scala/bug#10661


retronym (approved):

retronym:
Thanks!

PR #6235, opened 2017-12-20
There is a bug in JavaMirror.makeScalaPackage 

roman0yurin:
A conflict was fixed when searching for a symbol of subpackage when there is a collision between the names of the nested package and the method in the package object. For example there is package a.b and method b inside package object a. The search found any symbols inside the package, although only the symbol of the package was needed here. Because of this, there was an error (throw new ReflectError(opkg+" is not a package"))

### DIFF
--- a/src/reflect/scala/reflect/runtime/JavaMirrors.scala
+++ b/src/reflect/scala/reflect/runtime/JavaMirrors.scala
@@ -949,18 +949,19 @@ private[scala] trait JavaMirrors extends internal.SymbolTable with api.JavaUnive
         if (split > 0) packageNameToScala(fullname take split) else this.RootPackage
       val owner = ownerModule.moduleClass
       val name = TermName(fullname) drop split + 1
-      val opkg = owner.info decl name
-      if (opkg.hasPackageFlag)
+      // Be tolerant of clashes between, e.g. a subpackage and a package object member. These could arise
+      // under separate compilation.
+      val opkg = owner.info.decl(name).filter(_.hasPackageFlag)
+      if (opkg != NoSymbol)
         opkg.asModule
-      else if (opkg == NoSymbol) {
+      else {
         val pkg = owner.newPackage(name)
         pkg.moduleClass setInfo new LazyPackageType
         pkg setInfoAndEnter pkg.moduleClass.tpe
         markFlagsCompleted(pkg)(mask = AllFlags)
         info("made Scala "+pkg)
         pkg
-      } else
-        throw new ReflectError(opkg+" is not a package")
+      }
     }
 
     private def scalaSimpleName(jclazz: jClass[_]): TypeName = {
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -5238,6 +5238,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         val pdef = treeCopy.PackageDef(pdef0, pdef0.pid, pluginsEnterStats(this, pdef0.stats))
         val pid1 = typedQualifier(pdef.pid).asInstanceOf[RefTree]
         assert(sym.moduleClass ne NoSymbol, sym)
+        if(pid1.symbol.ne(NoSymbol) && !(pid1.symbol.hasPackageFlag || pid1.symbol.isModule ))
+          reporter.error(pdef.pos, s"There is name conflict between the ${pid1.symbol.fullName} and the package ${sym.fullName}.")
         val stats1 = newTyper(context.make(tree, sym.moduleClass, sym.info.decls))
           .typedStats(pdef.stats, NoSymbol)
         treeCopy.PackageDef(tree, pid1, stats1) setType NoType
--- a/test/files/neg/names-package-method-conflict.check
+++ b/test/files/neg/names-package-method-conflict.check
@@ -1,0 +1,4 @@
+names-package-method-conflict.scala:7: error: There is name conflict between the foo.bar and the package foo.bar.
+	package bar{
+                ^
+one error found
--- a/test/files/neg/names-package-method-conflict.scala
+++ b/test/files/neg/names-package-method-conflict.scala
@@ -1,0 +1,10 @@
+package object foo {
+	def bar: Unit = {}
+}
+
+package foo {
+
+	package bar{
+		class Baz
+	}
+}
--- a/test/files/run/reflection-package-name-conflict.check
+++ b/test/files/run/reflection-package-name-conflict.check
@@ -1,0 +1,4 @@
+class c
+class c
+class c
+class c
--- a/test/files/run/reflection-package-name-conflict/A_1.scala
+++ b/test/files/run/reflection-package-name-conflict/A_1.scala
@@ -1,0 +1,8 @@
+package a {
+  object `package` {
+    val b2 = 42
+  }
+  package b1 {
+    class c
+  }
+}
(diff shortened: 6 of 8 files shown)


retronym:
Definitions in package objects are intended to shadow over symbols in the package, including subpackages. Such shadowing is a poor practice but not illegal.

```
scala> :paste -raw

package a { object `package` { case class b() }; package b { class C } }
package test { class Test { new a.b.C  }}

<pastie>:2: error: type C is not a member of object a.b
package test { class Test { new a.b.C  }}
                                    ^
```

I'm not sure that runtime reflection should have a different semantic in this case.

Could you share the motivating example that you're trying to fix?

roman0yurin:
The situation is not so simple in my real example. The duplicating method is one module (in the package object), and a duplicate packet in the other module. An error occurs when both modules work together. An error is raised during runtime reflection of type, which not directly associated with either the package or the method, through a long call chain.
    I will try to create a simple test for this problem

roman0yurin:
For the test, I need to load the method's symbol (from the package object) into the scope (defs) of the owner package symbol before the symbol for the nested package (with the same name) is called. But how do I get the runtime reflection system to load the symbol of method (inside a package object)?

roman0yurin:
```
package object foo {
  def bar: Unit = {}
}

package foo {

  package bar{
    class Baz
  }
}
```
This code causes an internal compiler error:
ClassCastException inside scala.reflect.internal.Trees.itransform because the method shadows the packet and the reference to the package (PackageDef) turns out to be incorrect. Typer use lookupUnshadowedEntries to fine nested package and can not find correct package symbol. Because of this there are no easy ways to make this code work.
   I added a diagnostic for this error (in next part of PR) so that the application programmer gets an explicit instruction to fix it in the code.

roman0yurin:
Due to the aforementioned compiler limitation, this problem can not be reproduced within the single module. You need three modules to write the motivation test: two parallel modules (one with the method in package object, the second with the nested package) and the third one, which uses them and activates the error by using runtime reflection.

As a matter of fact, this is a confection between two different libraries and it is very difficult to understand it at the application level (the message ReflectError(opkg+" is not a package") does not provide information about the location of the problem in the code).

roman0yurin:
Do I need to create a project of three modules to demonstrate the conflict by name between the subpackage and the method in package object?

retronym:
[Sorry about the slow response.]

> Do I need to create a project of three modules to demonstrate the conflict by name between the subpackage and the method in package object?

If you can show the problem in a small SBT project with the structure you describe (but I can't quite follow), I can help to turn that into a test case.

retronym:
Okay, I think I understand the instructions now.

```
⚡ scalac -d /tmp $(f "package object p { def foo = 42 }"); scalac -d /tmp $(f "package p.foo; class C") && scala -cp /tmp
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_152).

scala> import reflect.runtime.universe._

scala> new p.foo.C
<console>:15: error: type C is not a member of Int
       new p.foo.C
                 ^

scala> Class.forName("p.foo.C")
res1: Class[_] = class p.foo.C

scala> rootMirror.classSymbol(res1)
scala.reflect.runtime.ReflectError: method foo is not a package
  at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$scala$reflect$runtime$JavaMirrors$$makeScalaPackage$1(JavaMirrors.scala:963)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.scala$reflect$runtime$JavaMirrors$$makeScalaPackage(JavaMirrors.scala:946)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$packageToScala$1(JavaMirrors.scala:928)
  at scala.reflect.runtime.TwoWayCaches$TwoWayCache.$anonfun$toScala$1(TwoWayCaches.scala:38)
  at scala.reflect.runtime.TwoWayCaches$TwoWayCache.toScala(TwoWayCaches.scala:34)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.packageToScala(JavaMirrors.scala:928)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.sOwner(JavaMirrors.scala:856)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.classToScala1(JavaMirrors.scala:993)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$classToScala$1(JavaMirrors.scala:986)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$toScala$1(JavaMirrors.scala:98)
  at scala.reflect.runtime.TwoWayCaches$TwoWayCache.$anonfun$toScala$1(TwoWayCaches.scala:38)
  at scala.reflect.runtime.TwoWayCaches$TwoWayCache.toScala(TwoWayCaches.scala:34)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.toScala(JavaMirrors.scala:96)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.classToScala(JavaMirrors.scala:986)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.classSymbol(JavaMirrors.scala:197)
  at scala.reflect.runtime.JavaMirrors$JavaMirror.classSymbol(JavaMirrors.scala:55)
```

src/reflect/scala/reflect/runtime/JavaMirrors.scala
```diff
@@ -949,18 +949,16 @@ private[scala] trait JavaMirrors extends internal.SymbolTable with api.JavaUnive
         if (split > 0) packageNameToScala(fullname take split) else this.RootPackage
       val owner = ownerModule.moduleClass
       val name = TermName(fullname) drop split + 1
-      val opkg = owner.info decl name
-      if (opkg.hasPackageFlag)
-        opkg.asModule
-      else if (opkg == NoSymbol) {
+      val opkg = owner.info declPackage name
```
retronym: I'd prefer `owner.info.decl(name).filter(_.hasFlag(PACKAGE))` here rather than adding the specialized `declPackage`.

retronym:
I am sympathetic to the argument that runtime reflection should be more tolerant of the double defintion here, we just need to be sure the change doesn't have unwanted consequences.

src/reflect/scala/reflect/runtime/JavaMirrors.scala
roman0yurin: As I understand it, the original search method (findDecl) contains a combination of symbol traversing, filtering and overriding. I just increased the abstraction of the filtering method in the "findDeclPredecate" method, adding a filter function (condition parameter) there instead of checking for the symbol flags. If the symbol is not excluded by the filter in line 955 (on the edited version of the Types.scala), then the override principle in line 1005 will be applied further to it.
 "baseClasses.head.newOverloaded(this, alts)"
This can lead to an undesirable side effect in our case.

roman0yurin:
And what about the second part of the problem with the explicit collision of nested package names and methods (when the problem can be found at the compilation stage)? Motivation test test / files / neg / names-package-method-conflict.scala

retronym:
Reporting the name clash during compilation makes sense, especially as the failure mode before as an compiler crash later on. There is some risk that this new error is reported in some case that used to _not_ crash, but I think that's acceptable on the 2.13.x branch as there is time for any problems to be flushed out.

I have added a test for runtime reflection (like in my comment above), and tweaked the commit messages and implementation a little:

https://github.com/scala/scala/compare/2.13.x...retronym:review/6235#diff-f85ad677e0fe479ae3f4e10a08a2a4ec

If you're happy that this fully addresses the bug, could you force push this commit to this PR and we'll be ready to merge.

roman0yurin:
I am happy)

roman0yurin:
What's wrong with this PR?

SethTisue:
was this closed without merging on purpose...?

retronym:
I thought we'd keep this one alive with my small tweaks, but seeing as this is closed I've resubmitted my review branch as #6274.

PR #6236, opened 2017-12-21
Tail-recursive Tseitin model solver.

hrhino:
Large (hopefully computer-generated) matches can lead to a search to become deep enough to send `findTseitinModelFor` into a `StackOverflowError`. The change here is a faithful reification of the call stack as a `List`.

This fixes scala/bug#10387. Annoyingly, the test case in that ticket can't be reduced much in size before it starts slipping within the max stack depth, so this test takes a minute or so to compile.

Additionally, some `Set[Int]`s in which the elements won't be negative are changed to use `BitSet`s instead, to maybe help performance. My wholly unscientific benchmark against the attached test case yields [these numbers](https://gist.github.com/hrhino/758a71d789aa838c5753ec9a73475279).

I took a swing at using a `BitSet` (or pair thereof) to represent clauses, but the result winds up being far more complicated and only a handful of percentage points faster (on the DPLL statistics counter). I think it could be made faster still by writing it as an imperative `while` loop, but that makes its correctness far less obvious.


src/compiler/scala/tools/nsc/transform/patmat/Solving.scala
```diff
@@ -468,50 +468,60 @@ trait Solving extends Logic {
...
+      state match {
+        case Nil => NoTseitinModel
+        case (clauses, extraLits) :: rest =>
+          if (clauses.isEmpty) extraLits
+          else if (clauses exists (_.isEmpty)) findTseitinModel0(rest)
+          else clauses.find(_.size == 1) match {
+            case Some(unitClause) =>
+              val unitLit = unitClause.head
+              findTseitinModel0((dropUnit(clauses, unitLit), extraLits + unitLit) :: rest)
+            //withLit(findTseitinModelFor(dropUnit(clauses, unitLit)), unitLit)
```
lrytz: stale comment

src/compiler/scala/tools/nsc/transform/patmat/Solving.scala
```diff
@@ -468,50 +468,60 @@ trait Solving extends Logic {
...
+              // appearing only in either positive/negative positions
+              val pures = pos ^ neg
+
+              if (pures.nonEmpty) {
+                val pureVar = pures.head
+                // turn it back into a literal
+                // (since equality on literals is in terms of equality
+                //  of the underlying symbol and its positivity, simply construct a new Lit)
+                val pureLit = Lit(if (neg(pureVar)) -pureVar else pureVar)
+                // debug.patmat("pure: "+ pureLit +" pures: "+ pures +" impures: "+ impures)
```
lrytz: `impures` is no longer there, might be confusing

lrytz (commented):
Wihtout knowing the code LGTM. cc @gbasler, for another review?

hrhino:
Thanks for that, @lrytz. I've removed those and added a quick overview of DPLL as a doc comment.

It occurs to me that the partest could be replaced by a unit test in `SolvingTest`, if I write out (or generate) the formula. Dunno if that would be preferred over the partest.

gbasler:
LGTM, very nice.

som-snytt:
Running partest just now, the `pos` section paused waiting for the test to complete.

Either partest could be tweaked not to pause between test kinds, or tweaked to start long-running tests early.

hrhino:
Ooh, so it does. I'm tempted to start by moving it to `test/files/pos/a10387.scala` as a 0-line hack, but which PR to piggy-back that onto?

Maybe if retronym's [slight optimization](https://gist.github.com/retronym/6ba30b3bbfd76715cfd1821f77c6075a) of that code that I noticed while poking at this (but couldn't find the hubris to cherry-pick in) makes it in somehow, that would be a good place.

Or hey, how many 0-line PRs does scala have? Other than the already-resolved mergebacks, I guess.

PR #6238, opened 2017-12-23
Patternly apply reduces apparent parens

som-snytt:
When typing an apply in pattern mode,
if a method type results, use the method
result directly. This adjustment was
previously applied in typedCase, but
not in typedBind. Now it should happen
in any pattern context. This occurs
when a case class has more than two
parameter lists.

Fixes scala/bug#10667


lrytz (commented):
LGTM, but worth another 👀 by @adriaanm

retronym (approved):
LGTM, too.

PR #6240, opened 2017-12-29
Enforce maxerrs more robustly

som-snytt:
`-Xmaxerrs` is currently advisory and depends on the good will
of the reporter. The standard console reporter is well-advised,
but others are not, such as sbt, which uses an sbt setting to
control logging.

This commit adds a simple filter in front of any reporter
deemed untrustworthy, if `-Xmaxerrs` has been set to a value
lower than the default.

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -33,7 +33,7 @@ import scala.tools.nsc.ast.{TreeGen => AstTreeGen}
 import scala.tools.nsc.classpath._
 import scala.tools.nsc.profile.Profiler
 
-class Global(var currentSettings: Settings, var reporter: Reporter)
+class Global(var currentSettings: Settings, reporter0: Reporter)
     extends SymbolTable
     with CompilationUnits
     with Plugins
@@ -74,6 +74,17 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
   // alternate constructors ------------------------------------------
 
   override def settings = currentSettings
+
+  private[this] var currentReporter: Reporter = { reporter = reporter0 ; currentReporter }
+
+  def reporter: Reporter = currentReporter
+  def reporter_=(newReporter: Reporter): Unit =
+    currentReporter = newReporter match {
+      case _: reporters.ConsoleReporter | _: reporters.LimitingReporter => newReporter
+      case _ if settings.maxerrs.isSetByUser && settings.maxerrs.value < settings.maxerrs.default =>
+        new reporters.LimitingReporter(settings, newReporter)
+      case _ => newReporter
+    }
 
   /** Switch to turn on detailed type logs */
   var printTypings = settings.Ytyperdebug.value
--- a/src/compiler/scala/tools/nsc/reporters/AbstractReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/AbstractReporter.scala
@@ -9,11 +9,8 @@ package reporters
 import scala.collection.mutable
 import scala.tools.nsc.Settings
 import scala.reflect.internal.util.Position
-// TODO
-//import scala.reflect.internal.Reporter
 
-/**
- * This reporter implements filtering.
+/** This reporter implements filtering by severity and position.
  */
 abstract class AbstractReporter extends Reporter {
   val settings: Settings
--- a/src/compiler/scala/tools/nsc/reporters/AbstractReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/AbstractReporter.scala
@@ -9,6 +9,8 @@ package reporters
 import scala.collection.mutable
 import scala.tools.nsc.Settings
 import scala.reflect.internal.util.Position
+// TODO
+//import scala.reflect.internal.Reporter
 
 /**
  * This reporter implements filtering.
--- a/src/compiler/scala/tools/nsc/reporters/StoreReporter.scala
+++ b/src/compiler/scala/tools/nsc/reporters/StoreReporter.scala
@@ -8,18 +8,14 @@ package reporters
 
 import scala.collection.mutable
 import scala.reflect.internal.util.Position
-// TODO
-//import scala.reflect.internal.Reporter
 
-/**
- * This class implements a Reporter that stores its reports in the set `infos`.
- */
+/** This class implements a Reporter that stores its reports in the set `infos`. */
 class StoreReporter extends Reporter {
   case class Info(pos: Position, msg: String, severity: Severity) {
     override def toString() = s"pos: $pos $msg $severity"
   }
   val infos = new mutable.LinkedHashSet[Info]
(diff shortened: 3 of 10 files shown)


src/compiler/scala/tools/nsc/reporters/Filter.scala
```diff
@@ -0,0 +1,54 @@
...
+ */
+package scala.tools.nsc
+package reporters
+
+import reflect.internal.util.Position
+import reflect.internal.settings.MutableSettings
+
+/** A `Reporter` that filters events before delegating.
+ */
+abstract class Filter(delegate: Reporter) extends Reporter {
```
lrytz: Could you instead create a class `DelegatingReporter`, place it next to `Reporter`, and maybe put a comment in `Reporter` somewhere mentioning the existence of the delegate? This would increase the likelyhood of the delegate to stay up to date.

som-snytt:
`reporters.Reporter` is scheduled for removal. I added the currently useless `DelegatingReporter` at `internal.Reporter`, where it will be useful once `global` takes an `internal.Reporter`.

I also added comments on the other reporter implementations.

src/reflect/scala/reflect/internal/Reporting.scala
```diff
@@ -116,6 +116,42 @@ abstract class Reporter {
     }
 }
 
+/** A `Reporter` that filters events before delegating.
+ *
+ *  Concrete subclasses should implement just the abstract `filter` method.
+ */
+abstract class DelegatingReporter(delegate: Reporter) extends Reporter {
```
lrytz: I think keeping this unused copy of `DelegatingReporter` around is more confusing than helpful - instead a comment in `reporters.DelegatingReporter` that the class should move together with `Reporter`?

src/compiler/scala/tools/nsc/reporters/DelegatingReporter.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.reflect.internal.settings.MutableSettings
+
+/** A `Reporter` that filters events before delegating.
+ *
+ *  Concrete subclasses should implement just the abstract `filter` method.
+ */
+abstract class DelegatingReporter(delegate: Reporter) extends Reporter {
+
+  /* True to permit the message. */
+  protected def filter(pos: Position, msg: String, severity: Severity): Boolean
```
lrytz: Should we move this method to `LimitingReporter` and make `DelegatingReporter` a pure delegate?

som-snytt: OK, I didn't overthink. It was Filter before. I noticed the java tool convention is `Forwarding` as in `ForwardingJavaFileManager`. Thx for the second look.

src/reflect/scala/reflect/internal/Reporting.scala
som-snytt: I started to transition `Reporter` but it's nontrivial. I didn't consider just making this a mix-in. I'll try that.

som-snytt:
Changed to traits `internal.{ForwardingReporter, FilteringReporter}` to avoid hierarchy lock-in.

som-snytt:
I wanted to use `-Xreporter` to run vi with a list of files to edit, since I don't have an IDE with red squiggles set up, but that doesn't work under sbt.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -75,6 +75,17 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
 
   override def settings = currentSettings
 
+  private[this] var currentReporter: Reporter = { reporter = reporter0 ; currentReporter }
+
+  def reporter: Reporter = currentReporter
+  def reporter_=(newReporter: Reporter): Unit =
+    currentReporter = newReporter match {
+      case _: reporters.ConsoleReporter | _: reporters.LimitingReporter => newReporter
+      case _ if settings.maxerrs.isSetByUser && settings.maxerrs.value < settings.maxerrs.default =>
```
lrytz, line 84: why do we check `settings.maxerrs.value < settings.maxerrs.default` here?

src/compiler/scala/tools/nsc/reporters/AbstractReporter.scala
```diff
@@ -9,6 +9,8 @@ package reporters
 import scala.collection.mutable
 import scala.tools.nsc.Settings
 import scala.reflect.internal.util.Position
+// TODO
+//import scala.reflect.internal.Reporter
```
lrytz: I actually think those comments are more noise than help; could you remove them across the PR? When we finally get around to removing class `nsc.reporters.Reporter`, it will be simple enough to find the places where we need to import `reflect.internal.Reporter`.

src/compiler/scala/tools/nsc/reporters/StoreReporter.scala
```diff
@@ -8,22 +8,23 @@ package reporters
...
  * This class implements a Reporter that stores its reports in the set `infos`.
  */
 class StoreReporter extends Reporter {
   case class Info(pos: Position, msg: String, severity: Severity) {
-    override def toString() = "pos: " + pos + " " + msg + " " + severity
+    override def toString() = s"pos: $pos $msg $severity"
   }
   val infos = new mutable.LinkedHashSet[Info]
-  protected def info0(pos: Position, msg: String, severity: Severity, force: Boolean) {
+  override protected def info0(pos: Position, msg: String, severity: Severity, force: Boolean): Unit = {
```
lrytz: I think we usually don't put `override` when only implementing an abstract method.

lrytz (changes requested):
Just some minor cleanups.

src/compiler/scala/tools/nsc/reporters/StoreReporter.scala
som-snytt: OK. I'll also normalize NoReporter, above. :)

src/compiler/scala/tools/nsc/Global.scala
som-snytt: That is a back door. Instead of a sentinel value like maxerrs=-1, if they're explicitly asking for a "huge" maxerrs, then trust their reporter to do whatever filtering and limiting it wants. Only if the user asks to "restrict" the maxerrs do we take this extraordinary step of intervening.

Probably this will never happen IRL, but for example: a reporter that removes similar errors might be trusted to respect maxerrs. Put another way, this change is really because sbt reporter doesn't respect maxerrs and when I set it to 5, I want that to work and not have to look in the sbt docs for their special setting.

lrytz: OK, doppelt genäht hält besser :-)

som-snytt: 💯

lrytz (approved):

PR #6241, opened 2017-12-29
[SIWIP] Accept case guard without trivial underscore

som-snytt:
`case if condition =>` seems easy and natural.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -1786,8 +1786,10 @@ self =>
      */
     def block(): Tree = makeBlock(blockStatSeq())
 
-    def caseClause(): CaseDef =
-      atPos(in.offset)(makeCaseDef(pattern(), guard(), caseBlock()))
+    def caseClause(): CaseDef = {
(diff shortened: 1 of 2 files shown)


SethTisue:
Does the spec already allow this and you’re just bringing the implementation up to date, or what?

If it isn’t already spec’ed this way, it isn’t clear to me that adding flexibility here is worth the loss of uniformity?

som-snytt:
This would be a spec change to make "abuse of match" (as an if-alternative) look prettier.

The underscore doesn't add information and makes for a longer line, especially with slightly longer conditional expressions. Omitting underscore makes it easier to keep the alternatives on one line.

```
v match {
  case p => x
  case _ => if (b1) y else if (b2) z else ???
}
v match {
  case p => x
  case if b1 => y
  case if b2 => z
  case _ => ???
}
```
The syntax is straightforward.

lrytz:
Since the proposal changes the syntax it affects various tools (scala meta, IntelliJ, etc) and needs some wider communication / discussion. Would you like to write a mini-SIP so the proposal and its cost/beneift can be discussed in a meeting?

SethTisue:
closing for inactivity. can be reopened if activity resumes

nafg:
I think it makes things less regular and orthogonal

PR #6242, opened 2017-12-31
Enable issue navigation for IntelliJ.

hrhino:
- Cmd-click (Ctrl-click) on scala/bug#1234 goes to that issue page on GitHub.

I've got these in other projects and I just realized that they're vanishing from scala because I keep running `intellijFromSample` when I go to/from `2.12.x`. 

Dunno if this has been considered and rejected, but I've found it pretty convenient.

- SD-1234 ticket references changed to scala/scala-dev#1234. 9acab45aee already normalized all bug references to this form.


retronym (approved):
Neat!

lrytz:
Nice! FTR, I personally keep separate checkouts for the major branches.

PR #6246, opened 2018-01-04
Make updateDynamic work with context bounds

howtonotwin:
Avoid inferring any type parameters to `updateDynamic` too early, and allow them to be inferred from the RHS.

Fixes scala/bug#10406

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4341,7 +4341,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
       def typedAssign(lhs: Tree, rhs: Tree): Tree = {
         // see scala/bug#7617 for an explanation of why macro expansion is suppressed
-        def typedLhs(lhs: Tree) = typed(lhs, EXPRmode | LHSmode)
+        def typedLhs(lhs: Tree) = typed(lhs, EXPRmode | LHSmode | POLYmode)
         val lhs1    = unsuppressMacroExpansion(typedLhs(suppressMacroExpansion(lhs)))
         val varsym  = lhs1.symbol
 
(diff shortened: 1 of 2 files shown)


howtonotwin:
This causes an uglier error message in some cases, however

```scala
scala> class Dyn extends Dynamic { def updateDynamic(f: String)(i: Int) = () }
scala> new Dyn().foo = ("a": Int)
// old
 found   : String("a")
 required: Int
       new Dyn().foo = ("a": Int)
                        ^
// new
 found   : String("a")
 required: Int
error after rewriting to new Dyn().updateDynamic("foo")(("a": Int))
possible cause: maybe a wrong Dynamic method signature?
       new Dyn().foo = ("a": Int)
                        ^
```

The previous version of the code typed the rhs before typing the application as a whole, which I think was an attempt at solving this problem, but it had the side effect of breaking inference. I can get everything to work just fine by manually managing `context.undetparams`, but that seems ugly.

som-snytt:
Elsewhere I tried to show both the first error and the error on the rewrite. Maybe a similar effort would be useful here. Also worth noting that after this change, there will be other differences in typing the arg, probably desirable.

https://github.com/scala/scala/blob/6815129b8d82615b249a084ab70af2288a0b19eb/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L4700

howtonotwin:
Do you mean leaving notes like the following?

```scala
scala> class Dyn extends Dynamic { var foo: Int = 0; def updateDynamic(f: String)(s: String) = () }
scala> new Dyn().foo = "a"
error: type mismatch;
  found   : String("a")
  required: Int
  // new v
  expression does not convert to Dynamic application because member foo exists on type Dyn
scala> new Dyn().bar = 1
error: type mismatch;
  found   : Int(1)
  required: String
  error after rewriting to new Dyn().updateDynamic("bar")(1)
  possible cause: maybe a wrong Dynamic method signature?
  // new v
  expression is rewritten because there is no member named bar on type Dyn
```

The choice between dynamic application and normal application is based on whether or not members exist, regardless of type. In contrast, choosing between `v = v.+(x)` and `v.+=(x)` when compiling `v += x` depends on the type of `+=` and `+`. I think this makes adding errors like the above rather pointless in this case, because there will never be messages more interesting than "there is/isn't a member named _ on _".

lrytz (approved):
Thanks!

PR #6247, opened 2018-01-04
Deprecate subtype notation for trait extends

som-snytt:
Fixes scala/bug#10678


retronym (approved):

SethTisue:
briefly pondered adding "release-notes" label, but my hunch is that nobody is actually doing this

som-snytt:
@SethTisue thanks, I deleted that usage from my hobby project.

PR #6248, opened 2018-01-05
Improve scaladoc link refs and syntax

ashawley:
- A couple of links to release notes are to non-existent RC tags
- Some links are not active in HTML because they're missing `[[` and `]]` brackets
- Some web sites links are broken or redirects
- There are unnecessary http apidoc links which Scaladoc could make
- Miscellaneous bogus syntax

### DIFF
--- a/src/library/scala/DelayedInit.scala
+++ b/src/library/scala/DelayedInit.scala
@@ -43,7 +43,7 @@ package scala
  *
  *  @author  Martin Odersky
  */
-@deprecated("DelayedInit semantics can be surprising. Support for `App` will continue. See the release notes for more details: https://github.com/scala/scala/releases/tag/v2.11.0-RC1", "2.11.0")
+@deprecated("DelayedInit semantics can be surprising. Support for `App` will continue. See the release notes for more details: https://github.com/scala/scala/releases/tag/v2.11.0", "2.11.0")
 trait DelayedInit {
   def delayedInit(x: => Unit): Unit
 }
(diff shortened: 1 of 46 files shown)


lrytz:
Still needs an `--update-check` on `test/files/neg/delayed-init-ref.scala`

ashawley:
Ok, done:  Amended and pushed the new commit.

src/library/scala/DelayedInit.scala
```diff
@@ -43,7 +43,7 @@ package scala
  *
  *  @author  Martin Odersky
  */
-@deprecated("DelayedInit semantics can be surprising. Support for `App` will continue. See the release notes for more details: https://github.com/scala/scala/releases/tag/v2.11.0-RC1", "2.11.0")
+@deprecated("DelayedInit semantics can be surprising. Support for `App` will continue. See the release notes for more details: [[https://github.com/scala/scala/releases/tag/v2.11.0]]", "2.11.0")
```
lrytz: I'm sorry to see this only now, but could you remove the `[[` / `]]` link markers here? It's not in a Scaladoc comment and will show up on the command line (deprecation warning). Then we also don't need the change to  test/files/neg/delayed-init-ref.check.

ashawley: No worries!  I'm pretty sure the links **do** show up [in the Scaladoc](http://www.scala-lang.org/api/2.12.3/scala/DelayedInit.html), but sure, I agree the brackets shouldn't show up on the console output!  I'll ~~undo both the change and the check~~ _remove the brackets_.

lrytz (approved):
Thank you!

PR #6249, opened 2018-01-06
Make Typer#dyna.mkInvoke more accurate

howtonotwin:
When searching the context for the tree in question, make sure to
actually check that the tree under scrutiny matches the one we're
looking for. This means that the check for varargs won't give a false
positive if the context is too large. Fixes scala/bug#7420.

Adjust the "vararg not supported" error so it doesn't hide other errors.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -622,8 +622,10 @@ trait ContextErrors {
         NormalTypeError(tree, fun.tpe+" does not take parameters")
 
       // Dynamic
-      def DynamicVarArgUnsupported(tree: Tree, name: Name) =
-        issueNormalTypeError(tree, name+ " does not support passing a vararg parameter")
+      def DynamicVarArgUnsupported(tree: Tree, name: Name) = {
+        issueNormalTypeError(tree, name + " does not support passing a vararg parameter")
+        setError(tree)
+      }
 
       def DynamicRewriteError(tree: Tree, err: AbsTypeError) = {
         issueTypeError(PosAndMsgTypeError(err.errPos, err.errMsg +
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -4105,14 +4105,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         else None
 
       def isDynamicallyUpdatable(tree: Tree) = tree match {
-        case DynamicUpdate(qual, name) =>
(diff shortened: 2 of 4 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -4184,22 +4179,26 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
           def findSelection(t: Tree): Option[(TermName, Tree)] = t match {
-            case Apply(fn, args) if hasStar(args) => DynamicVarArgUnsupported(tree, applyOp(args)) ; None
-            case Apply(fn, args) if matches(fn)   => Some((applyOp(args), fn))
-            case Assign(lhs, _) if matches(lhs)   => Some((nme.updateDynamic, lhs))
-            case _ if matches(t)                  => Some((nme.selectDynamic, t))
-            case _                                => (t.children flatMap findSelection).headOption
+            case Apply(fn, args) if matches(fn) =>
+              val op = if(args.exists(_.isInstanceOf[AssignOrNamedArg])) nme.applyDynamicNamed else nme.applyDynamic
+              // not supported: foo.bar(a1,..., an: _*)
+              val fn1 = if(treeInfo.isWildcardStarArgList(args)) DynamicVarArgUnsupported(fn, op) else fn
```
lrytz, line 4185: `fn1` is unused

src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
```diff
@@ -622,8 +622,10 @@ trait ContextErrors {
         NormalTypeError(tree, fun.tpe+" does not take parameters")
 
       // Dynamic
-      def DynamicVarArgUnsupported(tree: Tree, name: Name) =
-        issueNormalTypeError(tree, name+ " does not support passing a vararg parameter")
+      def DynamicVarArgUnsupported(tree: Tree, name: Name) = {
+        issueNormalTypeError(tree, name + " does not support passing a vararg parameter")
+        setError(tree)
```
lrytz, line 627: The new errors (reported in the test case) are not helpful / confusing, I'd say let's remove this again.

lrytz (commented):
Looks great, just two minor comments.

src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
howtonotwin: I just found it rather irregular/strange that the "varargs not supported" error masked this (IMO more important) one. Besides, there *are* two separate errors here: you can't use varargs and you are missing the method itself, and they may as well be given up front.

lrytz: OK, you convinced me :)

lrytz (approved):
Thank you!

PR #6251, opened 2018-01-08
Case also ends block stats in special cases

som-snytt:
In a few special cases, block statement sequence would finish on `}` but not on `case`.

Possibly other statement sequences would also necessarily terminate on `case` keyword, but that was not tested here.

Fixes scala/bug#10684


retronym (approved):
LGTM, agrees with IntellIJ's parser now :)

PR #6252, opened 2018-01-08
Improve error message when using a Java class as a value.

hrhino:
As pointed out in gitter, it's kinda unfair to our users to refer to a Java class as an `object` in error messages, despite that being what the compiler sees.

Also, the `enum object` error emitted by `ClassfileParser` sounds kinda strange to my picky ears. `enum Foobar` is perfectly fine IM(V)HO.

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
@@ -970,7 +970,10 @@ abstract class ClassfileParser {
           val s = module.info.decls.lookup(n)
           if (s != NoSymbol) Some(LiteralAnnotArg(Constant(s)))
           else {
-            warning(s"""While parsing annotations in ${in.file}, could not find $n in enum $module.\nThis is likely due to an implementation restriction: an annotation argument cannot refer to a member of the annotated class (scala/bug#7014).""")
+            warning(
+              sm"""While parsing annotations in ${in.file}, could not find $n in enum ${module.nameString}.
+                  |This is likely due to an implementation restriction: an annotation argument cannot refer to a member of the annotated class (scala/bug#7014)."""
+            )
             None
           }
 
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -750,7 +750,12 @@ trait ContextErrors {
 
       // def stabilize
       def NotAValueError(tree: Tree, sym: Symbol) = {
-        issueNormalTypeError(tree, sym.kindString + " " + sym.fullName + " is not a value")
+        /* Give a better error message for `val thread = java.lang.Thread`. */
+        val betterKindString =
+          if (sym.isJavaDefined && sym.isTrait) "Java interface"
+          else if (sym.isJavaDefined && (sym.isClass || sym.isModule)) "Java class"
+          else sym.kindString
+        issueNormalTypeError(tree, s"$betterKindString ${sym.fullName} is not a value")
         setError(tree)
       }
 
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -197,7 +197,9 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     def paramLists: List[List[Symbol]] = paramss
   }
 
-  private[reflect] case class SymbolKind(accurate: String, sanitized: String, abbreviation: String)
+  private[reflect] final case class SymbolKind(accurate: String, sanitized: String, abbreviation: String) {
+    def skolemize: SymbolKind = copy(accurate = s"$accurate skolem", abbreviation = s"$abbreviation#SKO")
+  }
 
   protected def newStubSymbol(owner: Symbol,
                               name: Name,
@@ -2579,7 +2581,8 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
       else ""
 
     private def symbolKind: SymbolKind = {
-      var kind =
+      implicit val triple2SK = (SymbolKind.apply _).tupled
+      val kind: SymbolKind =
         if (isTermMacro)                         ("term macro",           "macro method",    "MACM")
         else if (isInstanceOf[FreeTermSymbol])   ("free term",            "free term",       "FTE")
         else if (isInstanceOf[FreeTypeSymbol])   ("free type",            "free type",       "FTY")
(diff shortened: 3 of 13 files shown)


src/reflect/scala/reflect/internal/Symbols.scala
```diff
@@ -2565,31 +2565,32 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
...
+        if (isTermMacro)                         ("term macro",           "macro method",       "MACM" )
+        else if (isInstanceOf[FreeTermSymbol])   ("free term",            "free term",          "FTE"  )
+        else if (isInstanceOf[FreeTypeSymbol])   ("free type",            "free type",          "FTY"  )
+        else if (isPackageClass)                 ("package class",        "package",            "PKC"  )
+        else if (hasPackageFlag)                 ("package",              "package",            "PK"   )
+        else if (isPackageObject)                ("package object",       "package",            "PKO"  )
+        else if (isPackageObjectClass)           ("package object class", "package",            "PKOC" )
+        else if (isAnonymousClass)               ("anonymous class",      "anonymous class",    "AC"   )
+        else if (isRefinementClass)              ("refinement class",     "",                   "RC"   )
+        else if (isModule && isJavaDefined)      ("Java static module",   "Java-defined class", "MODJ" )
```
hrhino: This is the addition. Everything else is the required vertical alignment fixes.

hrhino:
Updated to be slightly less ambitious. Saying "Java class" everywhere for Java-defined classes is maybe worth it, but doing so only for Java statics modules is silly, and enabling it for classes changes a lot, and not always in good ways (e.g. "superJava class" showing up as part of a message).

For now let's just change it in this specific message (gotten from `stabilize`, where the un-valueness of Java-defined classes is uncovered).

src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
```diff
@@ -750,7 +750,15 @@ trait ContextErrors {
 
       // def stabilize
       def NotAValueError(tree: Tree, sym: Symbol) = {
-        issueNormalTypeError(tree, sym.kindString + " " + sym.fullName + " is not a value")
+        /* Give a better error message for `val thread = java.lang.Thread`.
+         * In general, saying "Java-defined {class,interface}" everywhere looks
+         * a bit ugly (usually we don't care), but here it avoids mentioning the
+         * compiler fiction that is Java companion modules. */
```
som-snytt: Just a personal opinion: I wonder if this code comment works better as a commit comment?

It's kind of tabs-vs-spaces, where the question is should I see this explanation inline or when I consult the commit after git blame? There are precedents for explaining inline.

I'd also love it if tooling made it easy to discover the discussion on the PR. Everything is forever.

lrytz: I can follow your reasoning, but personally I like (reasonable) explanations in line. Commit messages cannot be updated over time if the code changes, so you often end up having to track down multiple commits. (In genearl - this is a small example of course, the code is not likely to change often over time).

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -970,7 +970,10 @@ abstract class ClassfileParser {
           val s = module.info.decls.lookup(n)
           if (s != NoSymbol) Some(LiteralAnnotArg(Constant(s)))
           else {
-            warning(s"""While parsing annotations in ${in.file}, could not find $n in enum $module.\nThis is likely due to an implementation restriction: an annotation argument cannot refer to a member of the annotated class (scala/bug#7014).""")
+            warning(
+              sm"""While parsing annotations in ${in.file}, could not find $n in enum ${module.nameString}.
+                  |This is likely due to an implementation restriction: an annotation argument cannot refer to a member of the annotated class (scala/bug#7014).""".stripMargin
```
lrytz: I think you don't need to invoke `stripMargin` when using the `sm` interpolator

lrytz (commented):
Looks good otherwise.

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
hrhino: Ach, you're right. I'm so used to that tailing triple-quoted strings that it's easy to ignore.

Is there a reason why `sm` couldn't grace the 2.13 stdlib?

som-snytt: +1 Together with intrinsic standard interpolators.

hrhino:
Updated with a shorter comment, a longer commit message, and less extraneous margin stripping.

som-snytt:
For the record, "I value Java classes." (Says my t-shirt.)

src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
hrhino: The latter is in a WIP PR. I'll pull over `sm` today.

src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
```diff
@@ -750,7 +750,12 @@ trait ContextErrors {
 
       // def stabilize
       def NotAValueError(tree: Tree, sym: Symbol) = {
-        issueNormalTypeError(tree, sym.kindString + " " + sym.fullName + " is not a value")
+        /* Give a better error message for `val thread = java.lang.Thread`. */
+        val betterKindString =
+          if (sym.isJavaDefined && sym.isTrait) "Java interface"
+          else if (sym.isJavaDefined && (sym.isClass || sym.isModule)) "Java class"
```
retronym, line 756: Can Java Enums / Annotations arrive here? They could warrant custom descriptions.

"Non-Scala" would really be more accurate than "Java", but that doesn't really roll off the tongue.

retronym: I suppose we could just say "class/interface/enum/annotation java.util.Xxx is not a value", and leave off the "Java " part.

hrhino: Alright. I tried a change (again) to `symbolKind` and here are the results:

https://github.com/scala/scala/compare/2.12.x...hrhino:java-kinds-2

I'm okay with this, but wdyt?

hrhino: So, I've come up with an even more extensive list of `symbolKind` strings than that, and it turns out that some occurrences of `object` come from elsewhere (`Type#objectString` being the most recent one I found).

I'm sure I'll miss some. Do you think that the potential inconsistency from places I've missed is too likely to cause confusion worse than this patch is meant to fix?

lrytz: I think not, and we can fix them in the future. The change here looks good to me, and I'm happy to also include the other commit from your branch linked above.

lrytz:
@hrhino this is almost ready, just include the changes from your other branch (https://github.com/scala/scala/compare/2.12.x...hrhino:java-kinds-2).

hrhino:
Thanks for the ping. Changes squashed. I had tried to be a bit cleverer but I've given up for now.

lrytz (approved):

PR #6257, opened 2018-01-10
Eliminate TypeVars and Wildcards early in glb/lub

### LINKED ISSUE #10686: Order of TypeVar constraints affects unification
Contrived example, but situations like that _can_ arise during implicit search and they make the difference between success and failure:
```scala
def foo[A](implicit ev: (Int, String) =:= (Int, A)): (Int, A) = null
def bar[A](implicit ev: (Int, A) =:= (Int, String)): (Int, A) = null
foo // nope
bar // ok
```
The cause for this?
```scala
scala> val wild = appliedType(typeOf[(_,_)].typeConstructor, IntTpe, WildcardType)
wild: $r.intp.global.Type = (Int, ?)

scala> lub(wild :: typeOf[(Int, String)] :: Nil)
res2: $r.intp.global.Type = (Int, ?)

scala> lub(typeOf[(Int, String)] :: wild :: Nil)
res3: $r.intp.global.Type = (Int, String)
```
It doesn't look right to me. What if `TypeVar` constraints were not added in reverse order?

### LINKED ISSUE #10740: Bad type inference when type parameter as existential bound has invariant position
Compiles in dotty:
```scala
  def diff1[A](xs: Set[A], ys: Set[_ <: A]) = xs -- ys
  def diff2[A](xs: Set[_ >: A], ys: Set[A]) = xs -- ys
  def diff3[A](xs: Set[_ >: A], ys: Set[_ <: A]) = xs -- ys
  val xs = Set(Option(42))
  val ys = Set(Some(42))
  diff1(xs, ys)
  diff2(xs, ys) // type mismatch: A is inferred as Option[Int]
  diff3(xs, ys)
```
Adapted from https://github.com/scala/bug/issues/5559#issuecomment-292403446

Cause is similar to #10686 - when solving type variables the order of constraints matters.
Variables being solved here come from `ExistentialType.withTypeVars`.

joroKr21:
Type variables and wildcards can act both as subtypes and
supertypes depending on the direction of the test. This means
types with variables and/or wildcards can shadow other types in
`glb/lub`.

Consider the following examples:
```scala
lub((Int, ? ) :: (Int, String) :: Nil)
glb((Int, ?A) :: (Int, String) :: Nil)
```

Intuitively both should return `(Int, String)`, but this depends
on the order of types in the list passed to `glb/lub`.
This commit incorporates the simple heuristic of moving types that
contain variables and/or wildcards to the end of the list.

Fixes scala/bug#10686 and fixes scala/bug#10740

### DIFF
--- a/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
+++ b/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
@@ -158,31 +158,30 @@ private[internal] trait GlbLubs {
         rest filter (t => !first.typeSymbol.isSubClass(t.typeSymbol)))
   }
 
+  /** From a list of types, retain only maximal types as determined by the partial order `po`. */
+  private def maxTypes(ts: List[Type])(po: (Type, Type) => Boolean): List[Type] = {
+    def loop(ts: List[Type]): List[Type] = ts match {
+      case t :: ts1 =>
+        val ts2 = loop(ts1.filterNot(po(_, t)))
+        if (ts2.exists(po(t, _))) ts2 else t :: ts2
+      case Nil => Nil
+    }
+
+    loop(ts)
+  }
+
   /** Eliminate from list of types all elements which are a supertype
     *  of some other element of the list. */
-  private def elimSuper(ts: List[Type]): List[Type] = ts match {
-    case List() | List(_) => ts
-    case t :: ts1 =>
-      val rest = elimSuper(ts1 filter (t1 => !(t <:< t1)))
-      if (rest exists (t1 => t1 <:< t)) rest else t :: rest
-  }
+  private def elimSuper(ts: List[Type]): List[Type] =
+    maxTypes(ts)((t1, t2) => t2 <:< t1)
 
   /** Eliminate from list of types all elements which are a subtype
     *  of some other element of the list. */
-  private def elimSub(ts: List[Type], depth: Depth): List[Type] = {
-    def elimSub0(ts: List[Type]): List[Type] = ts match {
-      case List() => ts
-      case List(t) => ts
-      case t :: ts1 =>
-        val rest = elimSub0(ts1 filter (t1 => !isSubType(t1, t, depth.decr)))
-        if (rest exists (t1 => isSubType(t, t1, depth.decr))) rest else t :: rest
-    }
-    val ts0 = elimSub0(ts)
-    if (ts0.isEmpty || ts0.tail.isEmpty) ts0
-    else {
-      val ts1 = ts0 mapConserve (t => elimAnonymousClass(t.dealiasWiden))
-      if (ts1 eq ts0) ts0
-      else elimSub(ts1, depth)
+  @tailrec private def elimSub(ts: List[Type], depth: Depth): List[Type] = {
+    val ts1 = maxTypes(ts)(isSubType(_, _, depth.decr))
+    if (ts1.lengthCompare(1) <= 0) ts1 else {
+      val ts2 = ts1.mapConserve(t => elimAnonymousClass(t.dealiasWiden))
+      if (ts1 eq ts2) ts1 else elimSub(ts2, depth)
     }
   }
 
--- a/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
+++ b/src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
@@ -167,7 +167,14 @@ private[internal] trait GlbLubs {
       case Nil => Nil
     }
 
-    loop(ts)
+    // The order here matters because type variables and
+    // wildcards can act both as subtypes and supertypes.
+    val (ts2, ts1) = ts.partition(_ exists {
+      case tv: TypeVar => !tv.isGround
+      case t => t.isWildcard
+    })
+
+    loop(ts1 ::: ts2)
   }
 
   /** Eliminate from list of types all elements which are a supertype
--- a/test/files/pos/t10686.scala
+++ b/test/files/pos/t10686.scala
@@ -1,0 +1,6 @@
+object Test {
+  def ltr[A](implicit ev: (Int, A) =:= (Int, String)) = null
+  def rtl[A](implicit ev: (Int, String) =:= (Int, A)) = null
+  ltr
+  rtl
+}
--- a/test/files/pos/t10740.scala
+++ b/test/files/pos/t10740.scala
@@ -1,0 +1,11 @@
+object Test {
+  case class Inv[A](x: A)
+  def diff1[A](i: Inv[A], j: Inv[_ <: A]) = 1
+  def diff2[A](i: Inv[_ >: A], j: Inv[A]) = 2
+  def diff3[A](i: Inv[_ >: A], j: Inv[_ <: A]) = 3
+  val i = Inv(Option(42))
+  val j = Inv(Some(42))
+  diff1(i, j)
+  diff2(i, j)
+  diff3(i, j)
+}


joroKr21:
I'm not sure how this would affect performance

joroKr21:
Is it possible that there is a deadlock somewhere in `concurrent.TrieMap` that is sometimes triggered by the scalacheck tests?

joroKr21:
@retronym how does this look?

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -20,6 +20,10 @@ private[internal] trait GlbLubs {
   /** In case anyone wants to turn off lub verification without reverting anything. */
   private final val verifyLubs = true
 
+  private object wildcardExists extends FindTypeCollector(_.isWildcard) {
```
joroKr21: Should I check for `isGround` instead? I'm not sure.

retronym:
To help me understand your patch, I've [attempted to split](https://github.com/scala/scala/compare/2.13.x...retronym:review/6257?expand=1) it into a refactoring and a behavior change.

Does that split look faithful to your intent?

I still need a bit more time to understand the change.

joroKr21:
Your split is almost correct. Apologies for not doing this myself.

The semantic change is just the addition of this switch of the current type from `t` to `t1`:
```scala
if (lteq(t, t1) && t.exists(_.isWildcard)) t1 else t
```
Note that if we're not going to check for wildcards the switch is pointless and there would be no need to compare `lteq(t, t1)`.

The (tail)-recursive version of the "find all maximal elements in a list given a partial order" (here `maxTypes`) made it easier for me to understand what it's actually doing. Here the worst running time is inherently O(n^2). There is a difference in the order that types are compared. In `maxTypes` if we find a bigger element than the current one, we immediately switch to it and go back to the beginning of the list. The old implementation always uses the head of the list to eliminate smaller elements and then recurses into the tail.

joroKr21:
Hold on, it looks to me that `maxTypes` doesn't do enough comparisons, so there must be a bug.
The most annoying part is to preserve the original order of the list. This shouldn't matter right?

joroKr21:
Should be good now

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -182,31 +182,46 @@ private[internal] trait GlbLubs {
...
+  /** From a list of types, retain only maximal types as determined by the partial order `po`. */
+  private def maxTypes(ts: List[Type])(po: (Type, Type) => Boolean): List[Type] = {
+    /** Outer loop: fold over `ts` to calculate maximal types, preserving the original order. */
+    @tailrec def outer(ts: List[Type], max: List[Type], rev: Boolean): List[Type] = ts match {
+      case t1 :: ts1 =>
+        /** Inner loop: combine `t1` and the maximal list `max` into the next maximal list. */
+        @tailrec def inner(work: List[Type], done: List[Type]): List[Type] = work match {
+          case t2 :: ts2 if po(t1, t2) => // found a bigger type, we're done here
+            if (!t2.exists(_.isWildcard) || !po(t2, t1)) max
+            else t1 :: (ts2 reverse_::: done)
```
joroKr21: Could also do `ts2 reverse_::: t1 :: done` here, but it doesn't seem to matter.

joroKr21: Huh, I guess it does matter after all - so frustrating that `with` is not symmetric.

joroKr21:
@retronym It looks like the list passed to `lub`/`glb` can contain `TypeVar`s which depend on the exact order of comparisons as being called in the current implementation to be constrained :sob: so I can't change it in any way whatsoever. All I can do is push the `partition` on `?` inside.

joroKr21:
Updated to fix scala/bug#10740

An alternative would be to use a `Queue` for type constraints and prepend / append as needed.

Note that this also compiles in Dotty, but neither on Scala 2.13.x nor with this PR:
```scala
def foo[A, B](implicit ev: (Int, A) =:= (B, String)) = 42
foo
```

This would require a redesign of how type variables and constraints are handled.

**Edit:** Should I add it as a negative test case?

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -158,31 +158,30 @@ private[internal] trait GlbLubs {
         rest filter (t => !first.typeSymbol.isSubClass(t.typeSymbol)))
   }
 
+  /** From a list of types, retain only maximal types as determined by the partial order `po`. */
+  private def maxTypes(ts: List[Type])(po: (Type, Type) => Boolean): List[Type] = {
+    def loop(ts: List[Type]): List[Type] = ts match {
```
adriaanm, line 163: `@tailrec`?

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -158,31 +158,30 @@ private[internal] trait GlbLubs {
...
+      case t :: ts1 =>
+        val ts2 = loop(ts1.filterNot(po(_, t)))
+        if (ts2.exists(po(t, _))) ts2 else t :: ts2
+      case Nil => Nil
```
adriaanm, line 167: I assume we special-cased `Nil` and the singleton list based on profiling before? Before we drop these cases, let's gather some statistics on the frequency with which they occur (from casual inspection, not all callers ensure the list's length is >= 2).

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
```diff
@@ -167,7 +167,14 @@ private[internal] trait GlbLubs {
       case Nil => Nil
     }
 
-    loop(ts)
+    // The order here matters because type variables and
+    // wildcards can act both as subtypes and supertypes.
+    val (ts2, ts1) = ts.partition(_ exists {
```
adriaanm: I don't think this is the right solution. We shouldn't be calling `lub`/`glb` with a list of types that includes type variables or wildcards. These test cases are a great help in debugging this.

joroKr21: Any idea how we could avoid that?

joroKr21: Recalling my experiments here, the current tests exercise this code path, i.e. type variables being inferred by virtue of participating in the `lub/glb`.

The complete solution would probably come with Dotty but in the mean time it's unsatisfactory explaining to people that they have to swap the arguments to `=:=` to make their code compile :smile: (due to scala/bug#10686). How about approximating just the wildcards?

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
joroKr21: It would be a good exercise for me - any pointers how to do this?

adriaanm: Check out `Statistics.incCounter`

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
adriaanm: I agree we should improve, but I'd like to take a look at the actual code paths that lead here, and how to remove those non-sensical types from the beginning. The type checker already approximates the expected type using wildcards (as `wildPt`), which shares a similar motiviation.

adriaanm:
Also, I was poking around here and was reminded of `stripExistentialsAndTypeVars`. Maybe that method is the right target for this PR?

joroKr21:
In its current form it doesn't help, because it handles only top-level type variables and it just bails out, but I will try out something similar with a `TypeMap`.

adriaanm:
Yes, I meant this PR could "target" that method so that it also handles the general case. Maybe it was a mistake that it only handles top-level ones?

src/reflect/scala/reflect/internal/tpe/GlbLubs.scala
joroKr21: One code path is via `ExistentialType.withTypeVars` in subtyping checks when the other type also has type vars. Try this and see a few of the pos partests fail with "trying to do lub/glb of typevar":

```diff
--- a/src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
+++ b/src/reflect/scala/reflect/internal/tpe/TypeComparers.scala
@@ -529,7 +529,7 @@ trait TypeComparers {
         (rt2.parents forall (isSubType(tp1, _, depth))) &&
           (rt2.decls forall (specializesSym(tp1, _, depth)))
       case et2: ExistentialType =>
-        et2.withTypeVars(isSubType(tp1, _, depth), depth) || fourthTry
+        suspendingTypeVars(typeVarsInType(tp1))(et2.withTypeVars(isSubType(tp1, _, depth), depth)) || fourthTry
       case mt2: MethodType =>
         tp1 match {
           case mt1 @ MethodType(params1, res1) =>
```

I'm not sure what is the correct thing to do here. Since we only care if there is any solution to the existential skolems but not what that solution is, if feels unexpected that this would affect the inference of some unrelated type variables.

adriaanm: Agreed that those unrelated type vars should not be constrained -- they should act like skolems, I guess. I'm reluctant to change this until I fully understand how this works, which is not likely to happen anymore for 2.13.0, sadly.

adriaanm: My initial comment was wrong -- I see now from your ltr/rtl test case that we get into the situation with wildcards in types. I can't think of a better way to make this more consistent, so, let's go with yours.

adriaanm (approved):
Agreed this is a problem, and I can't think of a better way to solve it. So, let's book the progress. Would like to make sure performance is not affected before merging /cc @retronym.

adriaanm:
Since we have a lot going on, optimistically merging -- if performance is affected, we can fix before RC1.

PR #6261, opened 2018-01-13
Flesh out the interfaces of <:< and =:=

howtonotwin:
Make `=:=` a subtype of `<:<`.

Have both "Co" and "Contra" variants of every method on `<:<` and `=:=`
(except `apply`).

Flesh out the documentation of `<:<` and `=:=`, and some other `Predef` methods.


howtonotwin:
Ideally, I'd like to separate `$conforms` from `<:<`, as I feel like it'd be more sensible to have `def $conforms[A]: A => A` as *just* producing implicit views and all the `<:<` machinery contained in the `<:<` object, but it appears that doing that breaks the tests and I am not experienced enough to know why.

SethTisue:
any reviewers with an opinion on this...?

SethTisue:
one further rebase should fix CI (because #6557)

howtonotwin:
I think that should do it.

adriaanm:
@johnynek, any thoughts on this, as you were in these woods most recently?

johnynek:
I think these changes look nice. Thanks for sending them!

src/library/scala/Predef.scala
```diff
@@ -725,6 +712,7 @@ object Predef extends LowPriorityImplicits {
       override def andThen[C](r: Any =:= C) = r
       override def liftCo    [F[_]] = asInstanceOf[F[Any] =:= F[Any]]
       override def liftContra[F[_]] = asInstanceOf[F[Any] =:= F[Any]]
+      override def toString = "generalized constraint"
```
johnynek: can we include `=:=` in this toString?

howtonotwin: I thought about that and decided against it because the *only* generalized constraint you should ever see is this one, and it might be confusing when you have a `Int <:< Any` but it’s telling you that it’s a `=:=`. If there’s strong reason for it to say `=:=`, I’ll do it.

howtonotwin:
Not to be cliched, but... works on my machine.

/rebuild

howtonotwin:
`run/t2873` runs fine on my machine but it's been failing on CI. Does anyone know why? I think I remember it being something Java 9 related, but I can't fathom why these changes would affect it.

SethTisue:
@howtonotwin it's not your PR's fault. it's a test that depends on exactly what Java 8 update you're running. if you rebase onto the current head of 2.13.x, the failure will go away — you'll have a commit that makes the test less picky.

howtonotwin:
A bit of history rewriting later and all should be well again.

adriaanm:
Thanks for your patience, @howtonotwin!

PR #6263, opened 2018-01-15
Documentation and cleanups around the pickle encoding

lrytz:


src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala
```diff
@@ -345,7 +354,8 @@ abstract class ClassfileParser {
             if (index <= 0 || ConstantPool.this.len <= index) errorBadIndex(index)
             val start = firstExpecting(index, CONSTANT_UTF8)
             val len   = (in getChar start).toInt
-            in.buf drop start + 2 take len
+            val s     = start + 2
+            in.buf.slice(s, s + len)
```
retronym, line 358: This could be `.view.slice` to avoid creating the intermediate arrays just to pass into `flatMap`.

Going further, we might be able to correctly size the result array by summing the lengths.

```scala
    def getBytes(indices: List[Int]): Array[Byte] = {
      val head = indices.head
      values(head) match {
        case xs: Array[Byte] => xs
        case _               =>
          val starts = new Array[Int](indices.length)
          val lengths = new Array[Int](indices.length)
          var totalLength = 0
          foreachWithIndex(indices) { (index, i) =>
            if (index <= 0 || ConstantPool.this.len <= index) errorBadIndex(index)
            val start = firstExpecting(index, CONSTANT_UTF8)
            starts(i) = start
            val length = (in getChar start).toInt
            lengths(i)  = length
            totalLength += length
          }
          val result = new Array[Byte](totalLength)
          var resultPos = 0
          for (i <- starts.indices) {
            System.arraycopy(in.buf, starts(i) + 2, result, resultPos, lengths(i))
            resultPos += lengths(i)
          }
          recordAtIndex(getSubArray(result), head)
      }
    }
```

But AFAIK, this isn't a hot path so long as we still store most pickles in a single string, rather than array, so perhaps this is a premature optimization.

src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala
```diff
@@ -1034,21 +1034,34 @@ object BCodeHelpers {
...
-   *  Details about the storage format of pickles at the bytecode level (classfile annotations) can be found in SIP-10.
+   * Helpers for encoding a Scala signature (array of bytes) into a String or, if too large, an
+   * array of Strings.
+   *
+   * The encoding is as described in [[scala.reflect.internal.pickling.ByteCodecs]]. However, the
+   * special encoding of 0x00 as 0xC0 0x80 is not done here, as the resulting String(s) are passed
+   * as annotation argument to ASM, which will perform this step.
    */
-  class ScalaSigBytes(bytes: Array[Byte]) {
+  final class ScalaSigBytes(bytes: Array[Byte]) {
```
retronym, line 1044: We could generalize this to take `bytes` and `len` to avoid an intermediate array creation in:

https://github.com/scala/scala/blob/2f3791c3079d998d29788d121552c27517f58a6c/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala#L405

retronym:
A quick retrospective unit test for this, e.g. showing decode/encode round tripping and the `String`/`Array[String]` encoding boundary would also be welcome.

lrytz:
There's https://github.com/scala/scala/blob/2.13.x/test/files/run/bytecodecs.scala - it could be moved to JUnit.

Thanks for the other suggestions, it'll have to stay in my backlog for now though..

retronym (approved):

PR #6268, opened 2018-01-17
Remove uses of the @bridge annotation in compiler.

hrhino:
It's `private[scala]` and unused, so I doubt this will break anyone.

edited: obviously this wouldn't bootstrap. I should have known.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Infer.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Infer.scala
@@ -58,7 +58,6 @@ trait Infer extends Checkable {
     def improves(sym1: Symbol, sym2: Symbol) = (
          (sym2 eq NoSymbol)
       || sym2.isError
-      || (sym2 hasAnnotation BridgeClass)
       || isBetter(sym1, sym2)
     )
 
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -550,7 +550,7 @@ abstract class RefChecks extends Transform {
         }
 
         def checkOverrideDeprecated() {
-          if (other.hasDeprecatedOverridingAnnotation && !(member.hasDeprecatedOverridingAnnotation || member.ownerChain.exists(x => x.isDeprecated || x.hasBridgeAnnotation))) {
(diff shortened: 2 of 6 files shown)


SethTisue:
assuming re-STARRing is needed, and if you don't already know the drill there, let us know

hrhino:
So I'm kinda aware of how STARR works. I was intending to try to get this merged for M3 and delete it in M4, since someone (you? :D) would change `starr.version` to M3.

Is there a way to get a reSTARR between milestones? I know I can update `versions.properties`, and that your Jenkins publishes versions tagged with commit numbers, but I didn't think they were resolvable from local scala builds.

Anyhow, I'd love to know how this works. Up until now I've just guessed from PR and commit comments.

Thanks, Seth!!

retronym:
It is probably easiest to just defer removal of the annotation until after the next restarr (that looks like what you've done in the latest commit.)

We have restarr-ed between releases before, but it requires a bit of manual work to cut a tag and a release, so where possible we try to use the natural milestone cycle for such changes.

hrhino:
@retronym thanks. I'm inclined to just run nature run its course. There will be a M4, according to github, and a second PR will take a single file removal (I hope). 

Basically, if this gets merged in a milestone/RC before 2.13, it's all good :D

SethTisue:
oh, I was confused because I hadn't seen the earlier version of the code, before you changed it not to remove the annotation, apparently.

so, you can save learning the between-release re-STARR-ing dance for another day, then. (basically it means rolling a release like normal, just giving it a special in-between version number so nobody thinks it's something they should actually use for any other purpose... see e.g. scala/scala@db6c152b981622adb2106ac4812e1d8edf87b41d)

> your Jenkins publishes versions tagged with commit numbers, but I didn't think they were resolvable from local scala builds

(aside: well, I guess we *could* have STARR point to a selected nightly build, by adding a resolver? I'm not 100% sure why we don't do it that way, but I guess actually publishing something to Maven Central just seems safer... surely _is_ safer. there was that one time we accidentally deleted all of our nightly builds 😱😱😱)

SethTisue:
@hrhino time for followup PR...?

soronpo:
This PR seems to cause a regression for Scala reflection (hoping I tracked the regression correctly). See https://github.com/scala/bug/issues/10758

hrhino:
.....really? I'll take a look.

soronpo:
Hmmm... Maybe I'm wrong. Unfortunately, there is no direct correlation between the bad snapshot and the bad commit merge that caused the regression. After looking at the merge history in the vicinity, there might be other PR candidates to cause this.

PR #6271, opened 2018-01-18
Speed up creation of [Abs]TypeErrors.

hrhino:
`Throwable#fillInStackTrace` can be expensive, so minimize the number of calls we make to it.

- `AbsTypeError` and its subclasses are never thrown, so they don't need to extend `Throwable` at all.
- `TypeError`s are thrown, but if the user sees them it's a compiler bug anyhow, so only populate the stack trace if we're in `-Ydebug` mode.

This also adds a minute bit of clarity to the distinction between `TypeError` and `AbsTypeError`: you can't throw the ~former~ latter.

Contribution note: retronym did this independently last October, because all good ideas are already had. Found this out when asking for permission on gitter rather than forgiveness on github. Still committing, though, because I've noticed that smaller changes tend to get merged sooner.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala
@@ -21,7 +21,7 @@ trait ContextErrors {
   import global._
   import definitions._
 
-  sealed abstract class AbsTypeError extends Throwable {
+  sealed abstract class AbsTypeError {
     def errPos: Position
     def errMsg: String
     override def toString() = "[Type error at:" + errPos + "] " + errMsg
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -3693,7 +3693,7 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
   /** An exception for cyclic references of symbol definitions */
   case class CyclicReference(sym: Symbol, info: Type)
   extends TypeError("illegal cyclic reference involving " + sym) {
-    if (settings.debug.value) printStackTrace()
+    if (settings.debug) printStackTrace()
   }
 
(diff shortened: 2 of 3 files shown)


som-snytt:
Can't throw the latter.

If `Abs` prefix means unthrowable, maybe we can choose what it stands for. I'd help, but only absinthe comes to mind.

som-snytt:
`Abs` = All but stacktrace.

hrhino:
Thanks. I hope Jenkins doesn't mind rebuilding the same source code all over again so I can pretend that never happened.

I'm still not sure what `Abs` should stand for. I notice that `ContextErrors` doesn't follow the scalac tradition along the lines of `Symbols`, `Typers`, `Contexts`, and so on. `AbsTypeError` should just be `ContextError`, and then all will be well. Alternatively, since it has members named `TyperContextErrors`, `InferencerContextErrors`, and similar, maybe it should be renamed to follow that _other_ scalac naming tradition: `ContextErrorss`.

lrytz (approved):

retronym:
This PR has delivered a discernable performance boost to our compiler benchmarks in the `scala` corpus, in the 1-2% range. Thanks!

https://scala-ci.typesafe.com/grafana/dashboard/db/scala-benchmark?orgId=1&from=1493919850104&to=1520304489345&var-branch=2.12.x&var-source=scala&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40

hrhino:
Oh, wow, that's more than I was expecting! Glad to know it helps out ^_^

SethTisue:
@hrhino nice!!!

som-snytt:
Amazing! I wonder if @SethTisue will bring back the "low-hanging fruit" label.

rorygraves:
very nice!

mkeskells:
@hrhino 👍

PR #6272, opened 2018-01-19
Use partest 1.1.3

lrytz:
This should fix the failing bootstrap.


lrytz:
Ah: first, 1.1.2 was not built against 2.13.0-M2 (that would be easy to fix). And second, the commit that fixes the bootstrap is not actually in 1.1.2 (I misread the link on the release page, which goes to commit *since* a release, not those that went *into* a release).

Going to release 1.1.3.

PR #6273, opened 2018-01-21
Check for error in qual of select in type

som-snytt:
Typing select of terms and types was unified, with
the change that implicits are disabled in
`typedTypeSelectionQualifer`, with the side effect
or lack thereof that `instantiateToMethodType` no
longer issued an error.

Not sure if it's better to figure out from context whether to emit the 
failure to adapt, or whether a nicer message at this location
is warranted. @adriaanm probably has an opinion.

Also, updated a test that shows a cycle on imports; the new message
additionally shows what triggered the lookup, which is arguably
a feature.

Fixes scala/bug#10695

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -895,7 +895,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
 
         def cantAdapt =
           if (context.implicitsEnabled) MissingArgsForMethodTpeError(tree, meth)
-          else setError(tree)
+          else UnstableTreeError(tree)
 
         def emptyApplication: Tree = adapt(typed(Apply(tree, Nil) setPos tree.pos), mode, pt, original)
 
--- a/test/files/neg/t10474.check
+++ b/test/files/neg/t10474.check
@@ -1,0 +1,7 @@
+t10474.scala:8: error: stable identifier required, but Test.this.Foo found.
+    case Foo.Bar ⇒ true
+         ^
+t10474.scala:15: error: stable identifier required, but hrhino.this.Foo found.
+  val Foo.Crash = ???
+      ^
+two errors found
--- a/test/files/neg/t10474.scala
+++ b/test/files/neg/t10474.scala
@@ -1,0 +1,16 @@
+
+object Test {
+  def Foo(a: Int): Char = ???
+
+  object Bar
+
+  def crash[A](): Boolean = Bar match {
+    case Foo.Bar ⇒ true
+    case _ ⇒ false
+  }
+}
+
+trait hrhino {
+  def Foo(i: Int) = i
+  val Foo.Crash = ???
+}
--- a/test/files/neg/t10695.check
+++ b/test/files/neg/t10695.check
@@ -1,0 +1,4 @@
+t10695.scala:6: error: stable identifier required, but X.raw found.
+  val node: raw.Node = null
+            ^
+one error found
--- a/test/files/neg/t10695.scala
+++ b/test/files/neg/t10695.scala
@@ -1,0 +1,14 @@
+
+import X._
+
+object Main extends App {
+
+  val node: raw.Node = null
+
+  Seq().fold(node)(_ => _)
+
+}
+
+object X {
+  def raw(s: String) = ???
+}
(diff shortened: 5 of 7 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -5033,24 +5033,24 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         case Select(qual @ Super(_, _), nme.CONSTRUCTOR) =>
           // the qualifier type of a supercall constructor is its first parent class
           typedSelect(tree, typedSelectOrSuperQualifier(qual), nme.CONSTRUCTOR)
+        case Select(qual, name) if name.isTypeName =>
+          val qual1 = typedTypeSelectionQualifier(qual, WildcardType)
+          if (qual1.isErrorTyped) ErrorUtils.issueNormalTypeError(qual1, "Bad path for selection")
```
hrhino: I suspect this error message might be a bit unclear to some users. With hope that I'm not dooming your PR to months of error-message bikeshedding, what about something like:

```
t10695.scala:6: error: Cannot select from method raw: expected a stable path (value or package).
  val node: raw.Node = null
            ^
```

Which still mentions "path", but at least it sheds some light on what the invalid qualifier is (imagining that it is declared elsewhere and its nature is possibly unclear by looking at the vicinity of the error).

Also, does it deserve to be given a name in `ContextErrors`?

som-snytt: Yes, it's a pretty hard core message. In the bad cycle t6535, do we know what happened? I think I left it vague enough, intending to rely on the caret. I'd be happy to improve the message. `Obviously erroneous.`

lrytz: +1 on ContextError and a more talking message

lrytz: I wonder if the error should not be reported in `def cantAdapt` (in adapt). cc @adriaanm - related to https://github.com/scala/scala/pull/6334, which is assigned to you for review.

adriaanm: I think that could work. We can check `QUALmode` to decide whether this could even be a method application.

adriaanm:
I took the liberty to rework this PR as well. I'd like to book the progress for 2.12.5, but the errors could be improved...

adriaanm:
I'm going to abstain from further reviewing, since I now have a conflict of interest :-)

adriaanm:
I pushed a last-ditch attempt to get this to bootstrap. If that doesn't work, I suggest we take a bit more time and re-assign to 2.12.6

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -1212,10 +1212,15 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
-        case mt: MethodType if mode.typingExprNotFunNotLhs && !hasUndetsInMonoMode && !treeInfo.isMacroApplicationOrBlock(tree) =>
-          instantiateToMethodType(mt)
+        case mt: MethodType =>
+          if (mode.typingExprCanEtaOrApply && mt.isImplicit) adaptToImplicitMethod(mt) // (4.1)
+          else if (mode.typingExprCanEtaOrApply && !hasUndetsInMonoMode && !treeInfo.isMacroApplicationOrBlock(tree)) instantiateToMethodType(mt) // (4.2–3)
+          // The only valid reason to continue with a method-typed tree is if we're specifically looking for a method
+          // A method is not a value, and thus not an expression. The only way to create an expression is either to
+          // apply the method or to lift it to a function via eta-expansion (we tried those above)
+          else if (mode.inFunMode) vanillaAdapt(tree)
+          else UnstableTreeError(tree) // error: we have a method, but not in the
```
adriaanm: ... mood/mode

adriaanm:
Originally, I believe QUALmode was only meant for type checking paths, but it seems ~~it has evolved away from that~~ I was mistaken. Follow up: I asked Martin over lunch and he confirmed that I was wrong :-) QUALmode indicates that packages or references to static objects (which are actually not values) are admitted in addition to regular expressions (stable or not).

lrytz (approved):
When do you ever get to close 3 tickets with a 1-line change..!

hrhino:
👏

adriaanm:
I was just thinking of what could possibly go wrong with this fix -- are there any other situations where implicit search is disabled, where we are ok with marking the tree as error, but where it's not ok to signal an error? I can't think of any. 

In any case, we should introduce a more direct way to identify that we're typing a pattern, so we can tweak the error message accordingly (but, as shown here, both branches should result in emitting an error).

PR #6274, opened 2018-01-21
Better handling of clashes between subpackage and package object member

retronym:
Runtime reflection now works around the clash when it knows it is looking for a package.

The compiler now emits a error, rather than later crashing.


retronym:
Follow up to #6235, thanks to @roman0yurin for the initial PR and detective work to find the problem!

PR #6275, opened 2018-01-22
Eta-expand methods passed as functions to overloaded methods

szeiger:
The stricter automatic eta-expansion of method types to function
types in 2.13 interacts badly with the improved overload resolution
for higher-order methods. When a method reference is passed to such a
method without explicit eta-expansion, the shape-based lubbing
is not applied. Therefore the method is checked without an expected
type and does not eta-expand as it would when passed to a
non-overloaded method.

This is an attempt to fix the problem by using lubs more aggressively.
Unlike function and pattern-matching function literals method references
cannot be detected syntactically as early as we would like (as they are
only Idents without a symbol or type at that point) so we try to lub
all argument types and use that for typing the alternatives unless a
typed function literal is passed (because that might not typecheck
against the lub).

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -3364,9 +3364,30 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
               functionOrPfOrSamArgTypes(relTp)
             }
 
-            def functionProto(argTpWithAlt: List[(Type, Symbol)]): Type =
-              try functionType(funArgTypes(argTpWithAlt).transpose.map(lub), WildcardType)
-              catch { case _: IllegalArgumentException => WildcardType }
+            def sameType(tps: List[Type]): Option[Type] = tps match {
+              case t :: ts => if(ts.forall(_ == t)) Some(t) else None
+              case _ => None
+            }
+
+            def functionProto(argTpWithAlt: List[(Type, Symbol)], allowLub: Boolean = true): Type = try {
+              val argTps = funArgTypes(argTpWithAlt).transpose
+              val unified = argTps.map(ts => if(allowLub) lub(ts) else sameType(ts).getOrElse(WildcardType))
+              functionType(unified, WildcardType)
+            } catch { case _: IllegalArgumentException => WildcardType }
+
+            def samProto(argTpWithAlt: List[(Type, Symbol)]): Type = try {
+              // For SAM types we first check if they are all the same, in which case we use the
+              // type directly. Otherwise we replace all type args by WildcardType and check again.
+              // If they are still different, we use WildcardType. This allows for monomorphic and
+              // polymorphic SAM types but won't push argument types down into the SAM.
+              val alts = argTpWithAlt.map { case (tp, alt) => tp.asSeenFrom(pre, alt.owner).dealiasWiden }
+              sameType(alts).getOrElse {
+                sameType(alts.map {
+                  case TypeRef(t, s, args) => TypeRef(t, s, args.map(_ => WildcardType))
+                  case t => t
+                }).getOrElse(WildcardType)
+              }
+            } catch { case _: IllegalArgumentException => WildcardType }
 
             def partialFunctionProto(argTpWithAlt: List[(Type, Symbol)]): Type =
               try appliedType(PartialFunctionClass, funArgTypes(argTpWithAlt).transpose.map(lub) :+ WildcardType)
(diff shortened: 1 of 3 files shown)


lrytz:
This breaks the following example, which compiles on current 2.13.x:

```
scala> object t { def f(x: Int => Int): Int = 0; def f(y: String => Int): String = "1" }
defined object t

scala> val fun = (x: Int) => x
fun: Int => Int = $$Lambda$1871/736888459@7d59e968

scala> t.f(fun)
           ^
       error: type mismatch;
        found   : Int => Int
        required: Any => ?
```

lrytz:
The only idea I have: we could pass the expected type **only if** all the parameter types from the overloads are the same, i.e., if the lub doesn't widen any types. Or maybe use `WildcardType` instead of an actual lub if we want to push it further.

szeiger:
Hm, I already expected this kind of failure but I tested a similar case and it worked fine. The same thing happened with lambdas which is why I added the exception for typed lambdas in the first place. I probably forgot to use different parameter types in my test.

Could something like `WildcardType => WildcardType` be the appropriate expected type for the non-lambda cases? This should trigger eta-expansion without giving any false type information.

lrytz:
That will probably work, give it a try!

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3409,6 +3409,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                     else if (treeInfo.isPartialFunctionMissingParamType(tree)) {
                       if (argPtAlts.exists(ts => isPartialFunctionType(ts._1))) partialFunctionProto(argPtAlts)
                       else functionProto(argPtAlts)
+                    } else if (!treeInfo.isFunctionWithParamType(tree) && argPtAlts.forall { case (t, _) => isFunctionType(t)}) {
```
lrytz: why `argPtAlts.forall { case (t, _) => isFunctionType(t) }`, not also partial / sam function type?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -3409,6 +3409,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
...
+                      functionProto(argPtAlts)
```
lrytz: Could you add a comment which case is handled here? And while we're at it, it would be great to have a comment explaining why / when we use a `partialFunctionProto` above.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
szeiger: Eta-expansion never yields a PartialFunction but we should consider SAM types.

szeiger:
I've implemented both approaches (no lub + WildcardType params) now and both are useful but don't cover all cases. The WilcardType params version works for all monomorphic methods used as Function args. In cases where the param type are the same for all overloads we can support polymorphic methods by using the real param type instead of a WildcardType.

There does not seem to be an easy solution for the general case (lub required, polymorphic method) though. The source of the problem is the two incompatible uses of "expected type" in typer. It usually means "a type to which the computed type can be assigned" but when pushing arg types down into HOFs for typing it means "a function type whose arg types allow typing of the function". This works well when you know the exact types but the two meanings have opposing variance. In overload resolution where we try to approximate an "expected type" from multiple overloads the first one would require an intersection of the parameter types whereas the second one requires a lub.

I'm still working on SAM support before I push an update.

szeiger:
Here's an update with SAM support. It's not as powerful as for FunctionN because it will always drop parameter types (of polymorphic SAM types). In order to make this better we'd have to work out the relationship of SAM method argument types and SAM type parameter types so we could handle them the same way as for FunctionN. This would only work for simple FunctionN-like SAM types because the relationship can be more complex for SAMs.

adriaanm:
I've been looking into this. I think this makes it clear we need to improve overload resolution, probably along the lines of this PR, but I feel the proposal as it stands is a bit too ad-hoc. I would prefer to treat function types and sam types the same way. Probably something along the lines of "if all arguments in the same position across a set of applicable overloads expect the same-ish[*] type, type the argument with that expected type".

[*] the "ish" is the crux: we should ignore returns types of HOFs, and somehow unify type variables across the overloads, i think.

adriaanm:
Working on a poc

adriaanm:
Promising approach using tree attachments to mimic the idea of propagating richer expected types to where the arguments passed to an overloaded method are actually typed: https://github.com/scala/scala/compare/2.13.x...adriaanm:eta_overload?expand=1

adriaanm:
Then again, maybe we really need to attach this info to the expected type, which is trickier

odersky:
In Dotty, we eta-expand the function before trying to do overloading resolution. This solves the problem in principle. There are some differences, but they have to do with the other differences in the two schemes.

 - dotty does not unify, types must be equal
 - dotty does not consider SAM types, just functions and partial functions.

The modified test case is in https://github.com/lampepfl/dotty/pull/4191

adriaanm:
I'll adopt this one.

PR #6277, opened 2018-01-24
Correct, stable position for mixed-in outer accessors

retronym:
Previously, the position was incorrectly taken from the
outer accessor in the base trait. Not only was this wrong,
but it was only available when jointly compiling the trait
and subclass, so it also a source of unstable output.

References scala/scala-dev#405

Fixes scala/bug#7851


lrytz (approved):
👍

PR #6281, opened 2018-01-25
Update a check file to fix the build

lrytz:
Race condition in the merges of
https://github.com/scala/scala/pull/6270 (added the test)
and
https://github.com/scala/scala/pull/5310 (changes its output).


adriaanm:
Thanks!

PR #6283, opened 2018-01-25
Ensure that parameter names read from bytecode aren't obliterated by generic signatures.

hrhino:
c78d771e added code to parse the `MethodParameters` attribute from Java classfiles. However, if `javac` emits a `Signature` attribute after the `MethodParameters` attribute, the method info (previously parsed from the descriptor) is overwritten with the generic info, which doesn't keep the parameter symbols from the description-based info.

Therefore, collect names in a buffer until all attributes are parsed, then attach them to the parameter symbols in the final info.

Also, use the Java reflection `getParameters` method to populate these parameter symbols in runtime reflection.

Fixes scala/bug#10699.


lrytz (approved):
Nice work, thank you!

hrhino:
:D thanks for the fast review!

PR #6284, opened 2018-01-28
Deprecate static forwarders along with their forwardees.

hrhino:
Evaluates one of the flags that was being dropped on the floor in `addForwarder`, and finds it being unjustly so.

The attached test case shows the issue: javac won't warn on deprecation as scalac would. I'm willing to be told that we shouldn't depend explicitly on the javac output, and change it to a `DirectTest` or similar that inspects the bytecode directly.

Fixes scala/bug#10701.


lrytz (approved):
Thank you!

PR #6285, opened 2018-01-29
Preserve order of decls through pickle/unpickle

retronym:
The pickle format does not explicitly enccode the order of decls. Instead,
symbols are entered into an index in the order that they are found by the
pickler, either as a definition or as a reference. During unpickling,
symbols are read and entered into the owner's decls in that order.

This is a cause of unstable compiler output: a class that mixes in members
from some trait will have a different order if it is compiled jointly with
/ separately from that trait.

This commit modifies the pickler with an initial pass that reserves index
entries for all declarations in the declaration order.

The pickle format and the unpickler are unchanged.

References scala/scala-dev#405

### DIFF
--- a/src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
+++ b/src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
@@ -43,13 +43,15 @@ abstract class Pickler extends SubComponent {
             def shouldPickle(sym: Symbol) = currentRun.compiles(sym) && !currentRun.symData.contains(sym)
             if (shouldPickle(sym)) {
               val pickle = new Pickle(sym)
-              def pickleSym(sym: Symbol) = {
-                pickle.putSymbol(sym)
-                currentRun.symData(sym) = pickle
+              def reserveDeclEntries(sym: Symbol): Unit = {
+                pickle.reserveEntry(sym)
+                if (sym.isClass) sym.info.decls.foreach(reserveDeclEntries)
+                else if (sym.isModule) reserveDeclEntries(sym.moduleClass)
               }
 
               val companion = sym.companionSymbol.filter(_.owner == sym.owner) // exclude companionship between package- and package object-owned symbols.
               val syms = sym :: (if (shouldPickle(companion)) companion :: Nil else Nil)
+              syms.foreach(reserveDeclEntries)
               syms.foreach { sym =>
                 pickle.putSymbol(sym)
                 currentRun.symData(sym) = pickle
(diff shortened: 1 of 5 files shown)


src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
```diff
@@ -35,24 +35,50 @@ abstract class Pickler extends SubComponent {
   class PicklePhase(prev: Phase) extends StdPhase(prev) {
     def apply(unit: CompilationUnit) {
       def pickle(tree: Tree) {
-        def add(sym: Symbol, pickle: Pickle) = {
+        def add(sym: Symbol, pickle: Pickle) = withPickle(sym, pickle){ pickle =>
```
adriaanm: `withPickle` is called only once -- should we just leave it inlined? The `shouldPickle` condition does occur twice, though.

adriaanm: ah, maybe the plan was to factor out the
```
pickle.putSymbol(sym)
currentRun.symData(sym) = pickle
```

part?

src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
```diff
@@ -35,24 +35,50 @@ abstract class Pickler extends SubComponent {
...
-            add(sym, pickle)
-            add(sym.companionSymbol, pickle)
-            pickle.writeArray()
-            currentRun registerPickle sym
+
+            def shouldPickle(sym: Symbol) = currentRun.compiles(sym) && !currentRun.symData.contains(sym)
+            val shouldPickleSym = shouldPickle(sym)
+            if (shouldPickleSym) {
+              val pickle = new Pickle(sym)
+              def reserveDeclEntries(sym: Symbol): Unit = {
```
adriaanm: How about:
```
def reserveDeclEntries(sym: Symbol): Unit =
  if (sym.isClass) sym.info.decls.foreach(reserveDeclEntries)
  else if (sym.isModule) reserveDeclEntries(sym.moduleClass)
  else pickle.reserveEntry(sym)
```
?

adriaanm (changes requested):
A few suggestions / questions.

src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
retronym: Oops, `def add` itself was dead code from a failed refactoring attempt. I'll push a cleanup.

src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
```diff
@@ -35,24 +35,28 @@ abstract class Pickler extends SubComponent {
   class PicklePhase(prev: Phase) extends StdPhase(prev) {
     def apply(unit: CompilationUnit) {
       def pickle(tree: Tree) {
-        def add(sym: Symbol, pickle: Pickle) = {
-          if (currentRun.compiles(sym) && !currentRun.symData.contains(sym)) {
-            debuglog("pickling " + sym)
```
retronym: I didn't think this logging was worth keeping.

src/compiler/scala/tools/nsc/symtab/classfile/Pickler.scala
retronym: Actually, we need to reserve the slots for the class/module symbols themselves, prior to recursing into their decls. Pushed.

retronym:
rebased

PR #6287, opened 2018-01-30
Allow statistics printing after arbitrary phases

retronym:


src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -371,17 +371,7 @@ trait ScalaSettings extends AbsScalaSettings
...
-  val Ystatistics = {
-    val description = "Print compiler statistics for specific phases"
-    MultiChoiceSetting(
-      name    = "-Ystatistics",
-      helpArg = "phase",
-      descr   = description,
-      domain  = YstatisticsPhases,
-      default = Some(List("_"))
-    )
-  }
```
retronym, line 384: I need to tweak this to set the default back to "parser,typer,patmat,erasure,cleanup,jvm"

lrytz: Is this comment a `TODO`? Looks to me like you've done it already.

lrytz (approved):

PR #6290, opened 2018-01-30
drop -Yuse-stupid-types

SethTisue:
this was added by Gilles in 2010 "for internal use" and with the
description "Print the types of inherited members as seen from their
original definition context. Hint: you don't want to do that!"

so, it isn't clear what this was even for originally. but, googling and
searching GitHub turns up no uses.

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/model/ModelFactory.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/model/ModelFactory.scala
@@ -907,15 +907,12 @@ class ModelFactory(val global: Global, val settings: doc.Settings) {
 
   /** */
   def makeTypeInTemplateContext(aType: Type, inTpl: TemplateImpl, dclSym: Symbol): TypeEntity = {
-    def ownerTpl(sym: Symbol): Symbol =
-      if (sym.isClass || sym.isModule || sym == NoSymbol) sym else ownerTpl(sym.owner)
-    val tpe =
-      if (thisFactory.settings.useStupidTypes) aType else {
-        def ownerTpl(sym: Symbol): Symbol =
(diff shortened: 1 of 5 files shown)


SethTisue:
this fruit is hanging so low it's practically underground.

sjrd:
Y use stupid types, indeed? 😂

src/scaladoc/scala/tools/nsc/doc/model/ModelFactory.scala
```diff
@@ -909,13 +909,12 @@ class ModelFactory(val global: Global, val settings: doc.Settings) {
...
       if (sym.isClass || sym.isModule || sym == NoSymbol) sym else ownerTpl(sym.owner)
-    val tpe =
-      if (thisFactory.settings.useStupidTypes) aType else {
-        def ownerTpl(sym: Symbol): Symbol =
-          if (sym.isClass || sym.isModule || sym == NoSymbol) sym else ownerTpl(sym.owner)
-        val fixedSym = if (inTpl.sym.isModule) inTpl.sym.moduleClass else inTpl.sym
-        aType.asSeenFrom(fixedSym.thisType, ownerTpl(dclSym))
-      }
+    val tpe = {
+      def ownerTpl(sym: Symbol): Symbol =
```
mkeskells, line 911: copy and paste? (even if not by you)
isn't this the same as the shadowed ownerTpl on line 910

SethTisue: lol, indeed. I've pushed a revised commit that keeps only one of the two copies.

smarter:
Of course this option should have been called -Yuse-simpleton-types.

adriaanm:
Down with stupid types!

PR #6291, opened 2018-01-31
Avoid allocation of ClassTags in hot code

retronym:
Looking at allocation profiles, I noticed an unreasonable proportion of
classtag creation just to lookup attachments during typechecking.

This suggests that `ClassTag.apply` could be an invokedynamic bootstrap
to be a constant, but that's a job for another day.


Atry:
> This suggests that `ClassTag.apply` could be an invokedynamic bootstrap to be a constant, but that's a job for another day.

Is this feature the `Coherent` type class in https://github.com/lampepfl/dotty/issues/2047 ?

dwijnand:
No it's more like shapeless' `cachedImplict`, except for special-cased in the compiler for `ClassTag`.

SethTisue:
probably needs rebase for Jenkins to pick it up again?

retronym:
rebased

PR #6292, opened 2018-01-31
Refactor typechecking of array instantiation for performance

retronym:
Defining the extractor object `ArrayInstantation` in `typed1` meant
that we needed to thread a `LazyRef` through to `typedApply`
(even if we were dealing with a different AST node!).

This commit moves the extractor to `TreeInfo` after extracting the
core part that relies on `Typer.this`.


retronym:
This is a pretty straight refactoring, so I'm going to merge this now. Happy to address review comments post-hoc, @adriaanm

PR #6294, opened 2018-01-31
simple classpath tidyups

mkeskells:
simple classpath tidyups pulled from https://github.com/scala/scala/pull/5957 before the controversial work

### DIFF
--- a/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
+++ b/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
@@ -82,7 +82,7 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
       }
     }.unzip
     val distinctPackages = packages.flatten.distinct
-    val distinctClassesAndSources = mergeClassesAndSources(classesAndSources: _*)
+    val distinctClassesAndSources = mergeClassesAndSources(classesAndSources)
     ClassPathEntries(distinctPackages, distinctClassesAndSources)
   }
 
@@ -91,8 +91,7 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
    * creates an entry containing both of them. If there would be more than one class or source
    * entries for the same class it always would use the first entry of each type found on a classpath.
    */
-  private def mergeClassesAndSources(entries: Seq[ClassRepresentation]*): Seq[ClassRepresentation] = {
-    // based on the implementation from MergedClassPath
+  private def mergeClassesAndSources(entries: Seq[Seq[ClassRepresentation]]): Seq[ClassRepresentation] = {
     var count = 0
     val indices = collection.mutable.HashMap[String, Int]()
     val mergedEntries = new ArrayBuffer[ClassRepresentation](1024)
@@ -117,7 +116,7 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
         count += 1
       }
     }
-    mergedEntries.toIndexedSeq
+    if (mergedEntries isEmpty) Nil else mergedEntries.toIndexedSeq
   }
 
   private def getDistinctEntries[EntryType <: ClassRepresentation](getEntries: ClassPath => Seq[EntryType]): Seq[EntryType] = {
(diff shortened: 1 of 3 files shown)


src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
```diff
@@ -117,20 +116,19 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
...
+      entry <- getEntries(cp)
     } {
-      entriesBuffer += entry
-      seenNames += entry.name
+      if (seenNames.add(entry.name)) entriesBuffer += entry
     }
-    entriesBuffer.toIndexedSeq
+    if (entriesBuffer isEmpty) Nil else entriesBuffer.toIndexedSeq
   }
 }
```
retronym, line 133: For the next round of changes, `ClassPath#{list,classes, etc}` would compose better (from a performance perspective) if they accepted a buffer as an argument rather than returning collections.

src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
```diff
@@ -108,7 +107,7 @@ case class AggregateClassPath(aggregates: Seq[ClassPath]) extends ClassPath {
 
         if (existing.binary.isEmpty && entry.binary.isDefined)
           mergedEntries(index) = ClassAndSourceFilesEntry(entry.binary.get, existing.source.get)
-        if (existing.source.isEmpty && entry.source.isDefined)
+        else if (existing.source.isEmpty && entry.source.isDefined)
```
retronym: What's the reasoning behind this change? I don't really understand the existing code to reason about the change... 😕

src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
mkeskells: this PR was just the changes that dont affect the behaviour, structure etc 
This change just avoided the filter, a lookup in seenNames, and the builder creation

src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala
mkeskells: I thought that they were mutually exclusive, but this was from the old branch so I have not looked in detail

I think existing has source or binary populated, and so does entry

retronym: I believe that `entry.binary.isDefined && entry.source.isDefined` is always false, because `entry.isInstanceOf[ClassAndSourceFilesEntry]` is always fales. assuming we never wrap `AggregateClasspath`s.

The factory [`AggregateClassPath.apply`](https://github.com/scala/scala/blob/b9520211b22e53d4f80801b0f64cdb6275ac55cf/src/compiler/scala/tools/nsc/classpath/AggregateClassPath.scala#L137-L147) prevents that wrapping, but we don't actually guard against direct calls to `new AggregateClasspath`.

I'd suggest removing the `else` for now to avoid the brain strain of figuring this out, as this implementation detail will likely change as we refactor towards a more performant design.

mkeskells: removed

mkeskells:
running a perf_test on this change shows no measurable effect
```
after 10 90%
                 Run Name	           Wall time (ms)	       All Wall time (ms)	                  CPU(ms)	           Idle time (ms)	           Allocated(MBs)
             01_before-pr	  81	12,737.88 [-3.71% +3.55%]	12,737.88 [-3.71% +3.55%]	12,383.11 [-2.97% +2.96%]	       0.00 [ NaN%  NaN%]	 1,762.38 [-0.19% +0.14%]
             01_cp_simple	  81	12,640.79 [-3.79% +3.32%]	12,640.79 [-3.79% +3.32%]	12,289.90 [-3.76% +2.73%]	       0.00 [ NaN%  NaN%]	 1,756.18 [-0.11% +0.16%]
             02_before-pr	  81	12,733.05 [-3.02% +3.20%]	12,733.05 [-3.02% +3.20%]	12,379.42 [-2.81% +2.74%]	       0.00 [ NaN%  NaN%]	 1,762.18 [-0.14% +0.16%]
             02_cp_simple	  81	12,741.37 [-5.36% +3.59%]	12,741.37 [-5.36% +3.59%]	12,380.96 [-4.47% +2.98%]	       0.00 [ NaN%  NaN%]	 1,762.91 [-0.36% +0.22%]
             03_before-pr	  81	12,691.05 [-4.10% +4.58%]	12,691.05 [-4.10% +4.58%]	12,344.64 [-3.80% +3.66%]	       0.00 [ NaN%  NaN%]	 1,757.09 [-0.16% +0.23%]
             03_cp_simple	  81	12,825.90 [-4.12% +3.24%]	12,825.90 [-4.12% +3.24%]	12,477.96 [-3.96% +2.43%]	       0.00 [ NaN%  NaN%]	 1,758.50 [-0.13% +0.14%]
             04_before-pr	  81	12,701.27 [-3.88% +3.08%]	12,701.27 [-3.88% +3.08%]	12,374.37 [-3.15% +2.78%]	       0.00 [ NaN%  NaN%]	 1,765.64 [-0.29% +0.28%]
             04_cp_simple	  81	12,840.83 [-4.36% +3.36%]	12,840.83 [-4.36% +3.36%]	12,480.06 [-3.35% +2.91%]	       0.00 [ NaN%  NaN%]	 1,766.49 [-0.08% +0.14%]
```

lrytz (approved):

PR #6295, opened 2018-01-31
Toggle comment if no text is selected

Philippus:


lrytz:
Did you test this? On Windows, like the reporter of the issue? For me, on macOS, I can reproduce the issue on Chrome. On Firefox I can select text without the entry collapsing.

lrytz:
Nit: could you move the `fixes scala/scala-lang/issues/553` down from the title into to the commit message? https://github.com/scala/scala/blob/2.13.x/CONTRIBUTING.md#clean-commits-clean-history

PR #6297, opened 2018-02-01
Avoid creation of temporary Lists of imports during Context.lookup

retronym:
This is a major source of allocation pressure during typechecking.

After the refactoring, we only allocate a single cursor (which is
hopefully amenable to escape analyis), which advances the pair
of pointers out the context chain.

I've left my refactoring steps in this branch which might aid review: [retronym:topic/context-imports-unsquashed](https://github.com/scala/scala/compare/2.13.x...retronym:topic/context-imports-unsquashed?expand=1)

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -238,6 +238,7 @@ trait Contexts { self: Analyzer =>
     def imports: List[ImportInfo] = outer.imports
     /** Equivalent to `imports.headOption`, but more efficient */
     def firstImport: Option[ImportInfo] = outer.firstImport
+    protected[Contexts] def importOrNull: ImportInfo = null
     def isRootImport: Boolean = false
 
     /** Types for which implicit arguments are currently searched */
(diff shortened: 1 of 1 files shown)


retronym:
[Benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?orgId=1&from=1517396475678&to=1517468428026&var-scalaSha=7526e45%7Ceac0839f%7C81ad83d7%7Ccf9d508%7C86e001edb%7C79e5101738%7C629ceebfe%7Caac0816&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40) look no worse, maybe marginally better.

mkeskells:
maybe also consider ImportContext. not sure it this can be a val/lazy
```
 override final def imports      = impInfo :: super.imports
```
and some asserts like
```
        assert(imports.tail.headOption == nextOuter.firstImport, (imports, nextOuter.imports))
```
I have not looked in detail though

retronym:
This PR basically bypasses that method. Maybe we should deprecate it.

som-snytt:
Yes, that's a coincidence. I was going to try something similar after https://github.com/scala/scala/pull/5339 and also for the redundant search in https://github.com/scala/scala/pull/5339 ....

The asymmetry in the previous was awkward, witness the tweak for category 4 names introduced by packages.

Edit: I was thinking similar, except tracking all four categories in a pass. The point of the other PR is that if the current available "def sym" is package-defined-not-this-unit, you have to walk all the way to root imports checking for ambiguity due to imports.

retronym:
This is a straight-forward enough refactoring, so I'm going to merge.

PR #6298, opened 2018-02-01
Spot mistaken single-quote more broadly

som-snytt:
When erroring for unclosed char literal, broaden
the previous advice by looking at the source text.
If there are other single-quotes on the line,
then it smells like an intended string literal,
so it's worth adding that ' is not ".

As pushed for https://github.com/scala/bug/issues/10120#issuecomment-362229211


lrytz (approved):
In case you're really motivated (old behavior):

```
scala> 'abc'
<console>:1: error: unclosed character literal (or use " for string literal "abc")
       'abc'
           ^

scala> ''; 'abc'
<console>:1: error: empty character literal
       ''; 'abc'
       ^
<console>:1: error: unclosed character literal (or use " for string literal "'abc")
       ''; 'abc'
               ^

scala> ''; ''; 'abc'
<console>:1: error: empty character literal
       ''; ''; 'abc'
       ^
<console>:1: error: empty character literal
       ''; ''; 'abc'
           ^
<console>:1: error: unclosed character literal (or use " for string literal "''abc")
       ''; ''; 'abc'
                   ^
```

PR #6299, opened 2018-02-02
bundle scala-parser-combinators 1.0.7

SethTisue:
it doesn't matter a whole lot what version we bundle with 2.12.5,
but on the balance, I suggest we bundle the latest 1.0.x release

(there is also a 1.1.0, but let's be conservative)

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -20,7 +20,7 @@ scala.binary.version=2.12
 #  - jline: shaded with JarJar and included in scala-compiler
 #  - partest: used for running the tests
 scala-xml.version.number=1.0.6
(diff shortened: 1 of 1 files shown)


SethTisue:
the PR validator is currently down: https://contributors.scala-lang.org/t/jenkins-is-down/1551/1

gourlaysama (approved):
LGTM

som-snytt:
Fascinated just to see what conservative Seth looks like. Perm link this for future meme reference.

SethTisue:
I don't think we need to release-note this since basically anybody who would care is getting the lib by declaring the dependency, not via the bundling

PR #6301, opened 2018-02-02
Defer parameter alias computation until the end of typechecking

retronym:
Previously, it was done while typechecking super calls, and would fail
to see the fact that a yet-to-be-typechecked super constructor itself
had a parameter aliased by a grand-parent class.

Part of scala/scala-dev#405

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -50,6 +50,13 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
   // Symbols of ValDefs for right-associative operator desugaring which are passed by name and have been inlined
   private val inlinedRightAssocValDefs = new mutable.HashSet[Symbol]
 
+  // For each class, we collect a mapping from constructor param accessors that are aliases of their superclass
+  // param accessors. At the end of the typer phase, when this information is available all the way up the superclass
+  // chain, this is used to determine which are true aliases, ones where the field can be elided from this class.
+  // And yes, if you were asking, this is yet another binary fragility, as we bake knowledge of the super class into
+  // this class.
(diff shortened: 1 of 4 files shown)


src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -49,6 +49,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
   private val rightAssocValDefs = new mutable.AnyRefMap[Symbol, Tree]
   // Symbols of ValDefs for right-associative operator desugaring which are passed by name and have been inlined
   private val inlinedRightAssocValDefs = new mutable.HashSet[Symbol]
+  private val superConstructorCalls: mutable.AnyRefMap[Symbol, collection.Map[Symbol, Symbol]] = perRunCaches.newAnyRefMap()
```
lrytz: Could you add a comment for the structure of the map?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -5919,6 +5917,34 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
       if (phase.erasedTypes) None // OPT save the hashmap lookup in erasure type and beyond
       else transformed remove tree
   }
+
+
+  /** Finish computation of param aliases after typechecking is completed */
```
lrytz, line 5947: It would be nice to have a (very short) summary what happens with param aliases (or say where it's documented)...

lrytz (changes requested):
Looks good! Just asking for a bit of comments.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
retronym: I've added a few comments with my best reverse engineering of the intent.

retronym:
Ping @lrytz

lrytz (approved):

PR #6300, opened 2018-02-02
Stable names for lambda lifted method and fresh names

retronym:
Fresh names are created using a FreshNameCreator, which appends
an increasing number to the given prefix.

```
scala> val fresh = new scala.reflect.internal.util.FreshNameCreator()
fresh: scala.reflect.internal.util.FreshNameCreator = scala.reflect.internal.util.FreshNameCreator@42b84286

scala> List("foo$", "bar$", "foo$").map(fresh.newName(_))
res1: List[String] = List(foo$1, bar$1, foo$2)
```

Each compilation unit had its own fresh name creator, which is used
in the regular compiler. Macros and quasiquotes make use of a global
creator (at least, as of #3401).

Both of these are too broadly scoped to help achieve deterministic
fresh names: if sources are recompiled in a different order or separately
recompiled, the fresh name counters can be different. Methods in a given
compilation unit are not necessarily typechecked in a linear fashion;
they might be typechecked ahead of time to provide an inferred type
to a caller.

This commit:

  - Changes all known fresh name creations within the typer phase (in which
    out-of-order typechecking is a factor) to use a fineer grained fresh
    name creator. How fine grained? A fresh name generated as some position
    `p` shares the fresh name generator scoped at the closest method or
    class that encloses that the outermost enclosing tree at the same
    position. This definition is designed to give a shared fresh name
    creator for all fresh names generated in `macro1(macro2())`, even if
    the fresh names are requiested from with a Typer in the macro enclosed
    by a synthetic method.
  - Changes macro fresh names to use the same fresh naming scheme as the regular
    typechecker. An opt-out compiler option allows the old behaviour, but I'm
    interested to find real-world cases where the new scheme actually causes
    a problem

In addition, a small change is made to lambda lift to lift local methods in the
order that they are encountered during traversal, rather than sorting them
based on `Symbol.isLess` (which include `Symbol.id`, an order-of-typechecking
dependent value).

References scala/scala-dev#405

PR #6302, opened 2018-02-02
Don't copy the OVERRIDE flag in refinement decls

retronym:
`typedRefinement` defers the setting of this flag until the end of the
compilation unit, which means that inferred types that are derived from the
written refinement type can be unstable depending on whether they were
computed before or after the flag was set.

An alternative fix might be to just remove the setting of `OVERRIDE` in
in `typedRefinement.unitToCheck`.

The (last!) part of scala/scala-dev#405

Includes the commits of #6300 as it adds a new test case to `DeterminismTest`. Only the final commit is the subject of this pull request.


adriaanm:
Needs a rebase

lrytz:
Rebased, LGTM. @adriaanm

PR #6303, opened 2018-02-02
Fix scaladoc links for Any{Ref,Val,} and Nothing.

gourlaysama:
The symbols for `Any`, `AnyRef` and friends don't have an
`associatedFile`, so another (hardcoded) symbol is used to make
scaladoc pick the url for scala-library. Same thing is done for
package objects using the `package` member.

But once the external url is picked from the mapping list, we
should link to the real symbol, not the fake one.

That part regressed in #5799 (c6ed953).

Also, using `ListClass` as the replacement symbol doesn't work if the
codebase being documented doesn't reference `List` at all: in that case it
has no `associatedFile` either. The Akka codebase (from the ticket)
obviously uses `List`, so it worked for them, but failed when minimized.
There is some symbol initialization magic here that I don't understand,
but using the root scala package always works and makes more sense
anyway.

Fixes scala/bug#10673.

### DIFF
--- a/src/scaladoc/scala/tools/nsc/doc/model/MemberLookup.scala
+++ b/src/scaladoc/scala/tools/nsc/doc/model/MemberLookup.scala
@@ -9,7 +9,7 @@ trait MemberLookup extends base.MemberLookupBase {
   thisFactory: ModelFactory =>
 
   import global._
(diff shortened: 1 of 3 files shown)


lrytz (approved):
Thanks!

PR #6304, opened 2018-02-02
Add `->` alias for `Tuple2` to Predef

Jasper-M:
I know lately we prefer less possible ways of doing a thing instead of more. However, in the case of `Tuple2` I would say that ship has sailed a long time ago, so it would be nice if at least constructing and destructing a tuple have consistent syntaxes:

```
42 -> "foo" match {
  case key -> value => ???
}
```

It would also help readability (in my opinion) with API's where you can end up with lots of nested tuples, like the `RDD` API in Spark.

By the way, as far as I can tell, bytecode for `case (key, value) =>` and `case key -> value =>`  is exactly the same.

I did __not__ add a corresponding type alias `->`, but can add it if someone strongly feels that it should be included.

ping @SethTisue @Blaisorblade @dwijnand

### DIFF
--- a/test/junit/scala/PredefTest.scala
+++ b/test/junit/scala/PredefTest.scala
@@ -1,0 +1,23 @@
+package scala
+
+import org.junit.Assert._
+import org.junit.Test
+
+
+class PredefTest {
+  @Test
+  def testTuple2Alias: Unit = {
+    val tup = "foobar" -> 3
+    val char = tup match {
+      case str -> i => str.charAt(i)
+    }
+    assertEquals('b', char)
+
+    val a -> b -> c = false -> 42 -> "bazzz"
+    val res: Int = if (a) b else c.length
+    assertEquals(false, a)
+    assertEquals(42, b)
+    assertEquals("bazzz", c)
+    assertEquals(5, res)
+  }
+}
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -167,6 +167,20 @@ object Predef extends LowPriorityImplicits {
   /**  @group aliases */
   val Set         = immutable.Set
 
+  /**
+   * Allows destructuring tuples with the same syntax as constructing them.
+   *
+   * @example {{{
+   * val tup = "foobar" -> 3
+   *
+   * val c = tup match {
+   *   case str -> i => str.charAt(i)
+   * }
+   * }}}
+   * @group aliases
+   */
+  val ->        = Tuple2
+
   // Manifest types, companions, and incantations for summoning
   @annotation.implicitNotFound(msg = "No ClassManifest available for ${T}.")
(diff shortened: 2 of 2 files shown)


Blaisorblade:
The main part of the PR just makes`->` also available as extractor, not just value constructor, for consistency, so I’m in favor (as discussed on Gitter).

> I added the type alias -> for completeness, but I can just as easily leave it out if someone has objections

That’s a new identifier (`->` for *types*) that I’m less sure about, both because it’s good to be conservative on new identifiers, because of plans for Dotty, and because `A -> B` might look like a `Map A B` so it doesn’t seem to add clarity. I’m not sure “make types look like values” (if that’s the reason, I don’t get “completeness”) is a plan we follow often (or a clearly good idea).
EDIT: alternatively, could you elaborate on what you mean by completeness? Not sure I follow you there.

Jasper-M:
Well, for tuples the term syntax mirrors the type syntax: `val a: (Int, String) = (4, "four")`. Similar to how the extraction syntax mirrors the construction syntax. If you add the type alias, that remains true for the alternative syntax: `val a: Int -> String = 4 -> "four"`. But I personally don't really care whether it's there or not... Actually the more I think about it the less I'm convinced it's a good idea.

SethTisue:
yeah I find the type alias part iffy at best

Jasper-M:
I agree, I'll omit it unless told otherwise.

Also, I believe it is theoretically possible to target 2.12.x since scalac compiles away references to `->` from the bytecode when used as an extractor. In that case I could add a `@compileTimeOnly` annotation to make sure nobody does any funny incompatible stuff. But it's probably not worth it to break source compatibility...

Blaisorblade:
Interesting idea, but basically LGTM (IMHO) as-is for 2.13.x, and deferring to @SethTisue for decision.

test/junit/scala/PredefTest.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.Assert._
+import org.junit.Test
+
+
+class PredefTest {
+  @Test
+  def testTuple2Alias: Unit = {
+    val tup = "foobar" -> 3
+    val char = tup match {
+      case str -> i => str.charAt(i)
```
Blaisorblade, line 12: Just realized a small nit: should we test nested tuples? Just to be sure both arrows have the same associativity.

Jasper-M:
That was fun.

Blaisorblade:
@Jasper-M looks like you added nested tuples to your test? Hadn't realized.

LGTM I guess. @SethTisue any open issue with this, or can it be just merged? If so please do, I can't.

SethTisue (approved):
LGTM, but I'd like someone else from Lightbend to provide final signoff (since adding stuff to Predef is something to be especially cautious about)

lrytz (approved):
👍

SethTisue:
nice little improvement, glad to see it get in. thanks Jasper

PR #6305, opened 2018-02-02
Bump up ASM/jline versions in intellij SAMPLE files.

hrhino:
The bump to ASM 6.0.0-scala-1, combined with a deletion of my ivy2 cache recently, meant that IntelliJ was no longer picking up on `scala.tools.asm` classes and dyeing the backend red.

Hopefully this saves someone else the annoyance.

Also bump jline to the version in `versions.properties`, since it's also out of line.


lrytz:
In principle the IntelliJ classpaths don't need to be up-to-date with respect to versions.properties, the `intellij` sbt task updates the classpath entries. But it's not very obvious, once you have the project files the update needs to be triggered manually.

hrhino:
Oh, hey, that works. I actually had not run `intellij` since setting up intellij way back when; I've been running `intellijFromSample` when switching back and forth between 2.12.x and 2.13.x. My bad ...

PR #6306, opened 2018-02-02
Warn also on mirror class clobbering.

hrhino:
... and mention the source file if it's different.

This confused someone on gitter today, and I noticed that the clobbering check misses out on `class Foo; object foo` (whereby `foo`'s mirror will overwrite `Foo`, on a Mac). Slap a call to the check in `genMirrorClass`, and folks are duly warned.

Also drop `neg/case-collision2`, which existed just to test the same test on both JVM backends.

The [commit](https://github.com/scala/scala/commit/e4d1d930693ac75d8eb64c2c3c69f2fc22bec739) that introduced this considered making it an error, and seeing as it leads to runtime crashes with a pretty obscure JVM error, I'm sympathetic to that.


test/files/neg/case-collision.check
```diff
@@ -1,12 +1,27 @@
...
+  Such classes will overwrite one another on case-insensitive filesystems.
 class BIPPY
       ^
-case-collision.scala:8: warning: Class foo.DINGO$ differs only in case from foo.Dingo$. Such classes will overwrite one another on case-insensitive filesystems.
+case-collision.scala:8: warning: Class foo.DINGO$ differs only in case from foo.Dingo$.
+  Such classes will overwrite one another on case-insensitive filesystems.
 object DINGO
        ^
-case-collision.scala:11: warning: Class foo.HyRaX$ differs only in case from foo.Hyrax$. Such classes will overwrite one another on case-insensitive filesystems.
+case-collision.scala:8: warning: Class foo.DINGO differs only in case from foo.Dingo.
```
hrhino: There's probably some way to suppress this for when they're both mirror classes, but I personally don't mind warning twice.

mkeskells:
Hi - in https://github.com/scala/scala/pull/6124 (in review) I am changing the way that case insensitive comparison is made - to run this in the background threads, using ConcurrentHashMap, and a couple of small non functional improvements

In https://github.com/scala/scala/pull/6124/files  looks for 
```
if (!clazz.isArtifact) {
    warnCaseInsensitiveOverwrite(clazz)
    localOptimizations(classNode)
```
I did originally have reports for all files, but changed it because of the test failures :-)

it should be easy to include after that, but this 6124 also moves the the check outside the atifact check
@lrytz can probably advise better what the expectation is

lrytz:
The change LGTM, but let's get it in after #6124

lrytz:
@hrhino could you rebase?

src/compiler/scala/tools/nsc/backend/jvm/PostProcessor.scala
```diff
@@ -75,17 +76,24 @@ abstract class PostProcessor extends PerRunInit {
...
-    val duplicate = caseInsensitively.putIfAbsent(lowercaseJavaClassName, sourceClassName)
-    if (duplicate != null) {
-      backendReporting.warning(
-        clazz.position,
-          s"Class ${sourceClassName} differs only in case from ${duplicate}. " +
-            "Such classes will overwrite one another on case-insensitive filesystems."
+
+    caseInsensitively.get(lowercaseJavaClassName) match {
+      case null =>
+        caseInsensitively.put(lowercaseJavaClassName, (name, clazz.position))
```
lrytz: this code can can be executed concurrently, so we should use an atomic operation

hrhino: Done.

lrytz (approved):
Thanks!

PR #6307, opened 2018-02-03
Avoid IIOBE in phase stack

retronym:
Flatten's info transform ends calls `enteringMixin(enteringFlaten(...))`
all the way up the prefix, which violate my assumption that 128 recursive
atPhase calls should enough for anyone when we encounter heavily nested
code (such as that regrettably still emitted by the REPL when many imports
precede the current line).

This commit switches to using an ArrayBuffer that will grow as needed,
and also disables the entire maintenance of the phase stack when logging
is disabled.

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1098,6 +1098,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
     var currentUnit: CompilationUnit = NoCompilationUnit
 
     val profiler: Profiler = Profiler(settings)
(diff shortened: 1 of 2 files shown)


lrytz (approved):

lrytz (approved):

SethTisue:
@retronym what do you make of this: https://gist.github.com/SethTisue/cb615d34f4d207adb41c38c970618e70

retronym:
mea culpa, off by one. I should have used an `ArrayStack` which supports `pop`. #6327

SethTisue:
twotails is green again in the community build (cc @wheaties)

wheaties:
Cool, thanks guys!

PR #6310, opened 2018-02-05
Iterator.scanLeft is lazy at initial value

som-snytt:
Deliver the initial value before querying the iterator.

Fixes scala/bug#10709

### DIFF
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -590,7 +590,7 @@ trait Iterator[+A] extends TraversableOnce[A] {
   }
 
   /** Produces a collection containing cumulative results of applying the
-   *  operator going left to right.
+   *  operator going left to right, including the initial value.
    *
    *  $willNotTerminateInf
    *  $orderDependent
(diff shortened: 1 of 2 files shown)


src/library/scala/collection/Iterator.scala
```diff
@@ -602,14 +602,20 @@ trait Iterator[+A] extends TraversableOnce[A] {
...
+      case 0 | 2 => true
+      case 3     => false
+      case _     => if (self.hasNext) { state = 2 ; true } else { state = 3 ; false }
+    }
+    def next() = state match {
+      case 0 => state = 1 ; accum
+      case 1 => gen()
+      case 2 => state = 1 ; gen()
+      case 3 => Iterator.empty.next()
+    }
```
lrytz, line 618: For reference, alternative solution in strawman / 2.13: https://github.com/scala/collection-strawman/blob/v0.9.0/collections/src/main/scala/strawman/collection/Iterator.scala#L403

lrytz (approved):

src/library/scala/collection/Iterator.scala
som-snytt: Ichoran got me hooked on this sort of thing.

If `single` were lazy, it could be `Iterator.single(self.next) ++ self` guarding for empty.

PR #6311, opened 2018-02-05
Test that ScriptEngine loads

som-snytt:
Previously, changes around class path handling and a previous
bug with class loading may have conspired to break initialzing
the ScriptEngine.

This is just a test against fix on a related ticket.

Fixes scala/bug#10488

### DIFF
--- a/test/files/run/t10488.scala
+++ b/test/files/run/t10488.scala
@@ -1,0 +1,13 @@
+
+
+import javax.script._
+
+object Test {
+  def run() = {
+    val sem = new ScriptEngineManager()
+    val eng = sem.getEngineByName("scala")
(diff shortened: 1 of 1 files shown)


lrytz:
@som-snytt in the ticket you say "it will break again by 2.13" - should the commit be marked `[nomerge]` then?

som-snytt:
@lrytz I was speaking like Cassandra. Without a test, no one noticed that it broke (previously in time, but in future branch) on 2.13 and then fixed by Adriaan. This test is on 2.12 and will still succeed on current 2.13, because it missed its chance to shine. But I bet this will break again.

lrytz:
👍 thanks

PR #6312, opened 2018-02-06
Stabilizing definitions fixes

### LINKED ISSUE #10714: Only one stabilizing definition is emitted even when more than one required
https://github.com/scala/scala/pull/5999 introduces a synthetic val for the unstable LHS of an implicit method application where the method type depends on the LHS. Unfortunately this only handles the case where an expression requires a _single_ stabilizing val: if the expression involves chained method applications, each requiring a stabilizing val, then only the last of these is emitted resulting in a failure during bytecode generation,

```scala
object Test {
  class Foo {
    type Out
  }
  object Foo {
    implicit def foo: Foo { type Out = Bar } = ???
  }

  class Bar {
    type Baz = Foo
    def foo(implicit foo: Baz): foo.Out = ???
  }

  (new Bar).foo.foo
}
```
```
[[syntax trees at end of                     typer]]
  ...
  {
    <synthetic> <stable> <artifact> val stabilizer$2: shapeless.Test.Bar = stabilizer$1.foo(Test.this.Foo.foo);
    stabilizer$2.foo(Test.this.Foo.foo)
  }
}

java.util.NoSuchElementException: value stabilizer$1
        at scala.collection.mutable.AnyRefMap$ExceptionDefault.apply(AnyRefMap.scala:425)
        at scala.collection.mutable.AnyRefMap$ExceptionDefault.apply(AnyRefMap.scala:424)
        at scala.collection.mutable.AnyRefMap.apply(AnyRefMap.scala:180)
        at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder$locals$.load(BCodeSkelBuilder.scala:392)
...
[error] Error while emitting zipper.scala
[error] value stabilizer$1
[error] one error found
```
This is because the current implementation only records a single stabilizer on the expression tree which is overwritten by later ones.

PR fixing this incoming ...

### LINKED ISSUE #10736: Regression in implicit resolution between 2.12 and 2.13
Here is a self-contained test case. Maybe we can minimize it further:

~~~ scala
object ImplicitRegression {

  import scala.language.{implicitConversions, higherKinds}

  trait Map[K, +V] extends MapOps[K, V, Map, Map[K, V]]

  trait MapOps[K, +V, +CC[_, _], +C]

  trait BuildFrom[-From, -A, +C]

  object BuildFrom {
    implicit def buildFromMapOps[CC[X, Y] <: Map[X, Y] with MapOps[X, Y, CC, _], K0, V0, K, V]: BuildFrom[CC[K0, V0], (K, V), CC[K, V]] = ???
  }

  trait MapDecorator[K, V] {
    val `this`: Map[K, V]
    def zipByKeyWith[W, X, C](that: Map[K, W])(f: (V, W) => X)(implicit bf: BuildFrom[`this`.type, (K, X), C]): C = ???
  }

  implicit def MapDecorator[K, V](map: Map[K, V]): MapDecorator[K, V] { val `this`: map.type } = ???

  def test[A, B, C](map1: Map[A, B], map2: Map[A, B], f: (B, B) => C): Unit = {
    val map3 = map1.zipByKeyWith(map2)(f)/*(BuildFrom.buildFromMapOps)*/
    map3: Map[A, C]
  }

}
~~~

This code compiles fine with M2 but fails with the following error in M3:

~~~
[error] /home/julien/workspace/dev/scala/collection-strawman/collections-contrib/src/main/scala/strawman/collection/ImplicitRegression.scala:25: could not find implicit value for parameter bf: strawman.collection.ImplicitRegression.BuildFrom[stabilizer$1.this.type,(A, C),C]
[error]     val map3 = map1.zipByKeyWith(map2)(f)/*(BuildFrom.buildFromMapOps)*/
[error]                                       ^
[error] one error found
~~~

Enabling `-Xlog-implicits` gives the following:

~~~
[info] /home/julien/workspace/dev/scala/collection-strawman/collections-contrib/src/main/scala/strawman/collection/ImplicitRegression.scala:25: ImplicitRegression.this.BuildFrom.buildFromMapOps is not a valid implicit value for strawman.collection.ImplicitRegression.BuildFrom[stabilizer$1.this.type,(A, C),C] because:
[info] hasMatchingSymbol reported error: polymorphic expression cannot be instantiated to expected type;
[info]  found   : [CC[X, Y] <: strawman.collection.ImplicitRegression.Map[X,Y] with strawman.collection.ImplicitRegression.MapOps[X, Y, CC, _], K0, V0, K, V]strawman.collection.ImplicitRegression.BuildFrom[CC[K0,V0],(K, V),CC[K,V]]
[info]  required: strawman.collection.ImplicitRegression.BuildFrom[stabilizer$1.this.type,(A, C(in method test)),C(in method zipByKeyWith)]
[info]     val map3 = map1.zipByKeyWith(map2)(f)/*(BuildFrom.buildFromMapOps)*/
[info]                                       ^
~~~

It doesn’t compile either with 2.13.0-pre-9afd281-SNAPSHOT (from scala/scala#6308).

milessabin:
Fixes https://github.com/scala/bug/issues/10714. Also fixes https://github.com/scala/bug/issues/10736.

https://github.com/scala/scala/pull/5999 introduces a synthetic val for the unstable LHS of an implicit method application where the method type depends on the LHS. Unfortunately this only handles the case where an expression requires a _single_ stabilizing val: if the expression involves chained method applications, each requiring a stabilizing val, then only the last of these is emitted resulting in a failure during bytecode generation,

```scala
object Test {
  class Bar {
    class Foo {
      type Out
    }
    object Foo {
      implicit def foo: Foo { type Out = Bar } = ???
    }

    def foo(implicit foo: Foo): foo.Out = ???
  }

  (new Bar).foo.foo
}
```
results in a `NoSuchElementException`,
```
java.util.NoSuchElementException: value stabilizer$1
        at scala.collection.mutable.AnyRefMap$ExceptionDefault.apply(AnyRefMap.scala:425)
        at scala.collection.mutable.AnyRefMap$ExceptionDefault.apply(AnyRefMap.scala:424)
        at scala.collection.mutable.AnyRefMap.apply(AnyRefMap.scala:180)
        at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder$locals$.load(BCodeSkelBuilder.scala:392)
...
```
Prior to this PR only a single stabilizer is recorded on the expression tree which is overwritten by later ones. This is fixed here by allowing multiple stabilizers to be recorded and emitted.

In addition to that, previously spurious dependencies were picked up due to insufficient dealiasing. The following,
```scala
object Test {
  class Bar {
    class Foo {
      type Out
    }
    object Foo {
      implicit def foo: Foo { type Out = Bar } = ???
    }

    def foo(implicit foo: Foo): foo.Out = ???
  }

  (new Bar).foo.foo
}
```
triggered the preceding error despite not having a true dependency which would require a stabilizer in the first place. The test for a dependent argument now dealiases to eliminate these spurious dependencies.

Finally we avoid retyping stabilizing definitions when they are emitted, fixing scala/bug#10736.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala
@@ -183,5 +183,22 @@ trait StdAttachments {
    */
   case class OriginalTreeAttachment(original: Tree)
 
-  case class StabilizingDefinition(vdef: ValDef)
+  case class StabilizingDefinitions(vdefs: List[ValDef])
+
+  def addStabilizingDefinition(tree: Tree, vdef: ValDef): Tree =
+    tree.updateAttachment(StabilizingDefinitions(
+      tree.attachments.get[StabilizingDefinitions] match {
+        case Some(StabilizingDefinitions(vdefs)) => vdef :: vdefs
+        case _ => List(vdef)
+      }
+    ))
+
+  def stabilizingDefinitions(tree: Tree): List[ValDef] =
+    tree.attachments.get[StabilizingDefinitions] match {
+      case Some(StabilizingDefinitions(vdefs)) => vdefs
+      case _ => Nil
+    }
+
+  def removeStabilizingDefinitions(tree: Tree): Tree =
+    tree.removeAttachment[StabilizingDefinitions]
 }
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -5050,9 +5050,9 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                 val vsym = insertionContext.owner.newValue(unit.freshTermName(nme.STABILIZER_PREFIX), qual.pos.focus, SYNTHETIC | ARTIFACT | STABLE)
                 vsym.setInfo(uncheckedBounds(qual.tpe))
                 insertionContext.scope enter vsym
-                val vdef = atPos(vsym.pos)(ValDef(vsym, focusInPlace(qual)))
+                val vdef = atPos(vsym.pos)(ValDef(vsym, focusInPlace(qual)) setType NoType)
                 qual.changeOwner(insertionContext.owner -> vsym)
-                insertionContext.tree.updateAttachment(StabilizingDefinition(vdef))
+                addStabilizingDefinition(insertionContext.tree, vdef)
                 val newQual = Ident(vsym) setType singleType(NoPrefix, vsym) setPos qual.pos.focus
                 return typedSelect(tree, newQual, name)
               }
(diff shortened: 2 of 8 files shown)


milessabin:
Demoted to WIP ... it's along the right lines, but needs more work.

milessabin:
Although the multiple stabilizer issue was real, in the actual scenario, shapeless's `Zipper` breaking, the argument type wasn't genuinely dependent ... the type member referenced was an alias of an independent type. Dealiasing appropriately in the test for a dependent argument fixes that.

This is ready for review now ...

milessabin:
Rebased in the hope of unwedging Jenkins.

milessabin:
More synch did it!

milessabin:
Rebased.

lrytz (approved):

soronpo:
I think this PR caused the following regression https://github.com/scala/bug/issues/11397

PR #6314, opened 2018-02-07
Cache classloaders for compiler plugins and macros

jvican:
For the rationale and a detailed performance analysis, please check [this
issue](https://github.com/scala/scala-dev/issues/458).

This pull request continues the work started by @retronym on caching the
classpath, and extends it to caching compiler plugins classloaders. In short,
this change seems to yield compile-time performance improvements ranging from
16% to 35% for projects using compiler plugins.

The approach taken in this pull request is to cache only compiler plugins'
classpaths that contain zip or jar files (to avoid issues with validation). The
approach taken here is a little bit naive and I hope it improves in a follow-up
PR.

There is one missing item that I'll add to this pull request during the next
days: making the cache invalidation algorithm take into account the size of the
classpath entries to make the cache heuristics more robust.

However, as discussed in https://github.com/scala/scala-dev/issues/458, these
heuristics do not work in scenarios where the timestamp is zero'd and the size
does not change (this happens with tools that focus on reproducibility like
Pants and Bazel).

There is only one way we can fix this issue without relying on heuristics:
hashing the contents of the classpath. However, this is an expensive
computation for big classpaths and we want to probably stay away from it.

To solve this issue in a friendlier manner, I propose to add a compiler flag so
that the build tool can share the checksums of every jar they have with the
compiler and let the compiler invalidate the caches based on those. If this
sounds like a good solution to you, I will open another pull request
implementing it.

I'd like this change to target 2.12.5 if possible.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Macros.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Macros.scala
@@ -51,6 +51,9 @@ trait Macros extends MacroRuntimes with Traces with Helpers {
 
   def globalSettings = global.settings
 
+  private final val macroClassLoadersCache =
+    new scala.tools.nsc.classpath.FileBasedCache[ScalaClassLoader]()
+
   /** Obtains a `ClassLoader` instance used for macro expansion.
    *
    *  By default a new `ScalaClassLoader` is created using the classpath
@@ -60,8 +63,24 @@ trait Macros extends MacroRuntimes with Traces with Helpers {
    */
   protected def findMacroClassLoader(): ClassLoader = {
     val classpath = global.classPath.asURLs
-    macroLogVerbose("macro classloader: initializing from -cp: %s".format(classpath))
-    ScalaClassLoader.fromURLs(classpath, self.getClass.getClassLoader)
+    def newLoader = () => {
+      macroLogVerbose("macro classloader: initializing from -cp: %s".format(classpath))
+      ScalaClassLoader.fromURLs(classpath, self.getClass.getClassLoader)
+    }
+
+    import scala.tools.nsc.io.Jar
+    import scala.reflect.io.{AbstractFile, Path}
+    val locations = classpath.map(u => Path(AbstractFile.getURL(u).file))
+    val disableCache = settings.YdisableMacrosClassLoaderCaching.value
+    if (disableCache || locations.exists(!Jar.isJarOrZip(_))) {
+      if (disableCache) macroLogVerbose("macro classloader: caching is disabled by the user.")
+      else {
+        val offenders = locations.filterNot(!Jar.isJarOrZip(_))
+        macroLogVerbose(s"macro classloader: caching is disabled because the following paths are not supported: ${offenders.mkString(",")}.")
+      }
+
+      newLoader()
+    } else macroClassLoadersCache.getOrCreate(locations.map(_.jfile.toPath()), newLoader)
   }
 
   /** `MacroImplBinding` and its companion module are responsible for
--- a/src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
+++ b/src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala
@@ -30,7 +30,7 @@ sealed trait ZipAndJarFileLookupFactory {
   protected def createForZipFile(zipFile: AbstractFile): ClassPath
 
   private def createUsingCache(zipFile: AbstractFile, settings: Settings): ClassPath = {
-    cache.getOrCreate(zipFile.file.toPath, () => createForZipFile(zipFile))
+    cache.getOrCreate(List(zipFile.file.toPath), () => createForZipFile(zipFile))
   }
 }
 
@@ -177,20 +177,24 @@ object ZipAndJarSourcePathFactory extends ZipAndJarFileLookupFactory {
 }
 
 final class FileBasedCache[T] {
+  import java.nio.file.Path
   private case class Stamp(lastModified: FileTime, fileKey: Object)
-  private val cache = collection.mutable.Map.empty[java.nio.file.Path, (Stamp, T)]
+  private val cache = collection.mutable.Map.empty[Seq[Path], (Seq[Stamp], T)]
 
-  def getOrCreate(path: java.nio.file.Path, create: () => T): T = cache.synchronized {
-    val attrs = Files.readAttributes(path, classOf[BasicFileAttributes])
-    val lastModified = attrs.lastModifiedTime()
-    // only null on some platforms, but that's okay, we just use the last modified timestamp as our stamp
-    val fileKey = attrs.fileKey()
-    val stamp = Stamp(lastModified, fileKey)
-    cache.get(path) match {
-      case Some((cachedStamp, cached)) if cachedStamp == stamp => cached
+  def getOrCreate(paths: Seq[Path], create: () => T): T = cache.synchronized {
+    val stamps = paths.map { path =>
+      val attrs = Files.readAttributes(path, classOf[BasicFileAttributes])
+      val lastModified = attrs.lastModifiedTime()
+      // only null on some platforms, but that's okay, we just use the last modified timestamp as our stamp
+      val fileKey = attrs.fileKey()
+      Stamp(lastModified, fileKey)
+    }
+
+    cache.get(paths) match {
+      case Some((cachedStamps, cached)) if cachedStamps == stamps => cached
       case _ =>
         val value = create()
-        cache.put(path, (stamp, value))
+        cache.put(paths, (stamps, value))
         value
     }
   }
--- a/src/compiler/scala/tools/nsc/plugins/Plugin.scala
+++ b/src/compiler/scala/tools/nsc/plugins/Plugin.scala
@@ -8,11 +8,12 @@ package plugins
 
 import scala.tools.nsc.io.Jar
 import scala.reflect.internal.util.ScalaClassLoader
-import scala.reflect.io.{ Directory, File, Path }
+import scala.reflect.io.{Directory, File, Path}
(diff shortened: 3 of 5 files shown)


mkeskells:
Nio path has a hook to jdentify a file. I forget the api and cant find it via phone. Can post the details next week when I am back.
I would be happy to not cache in these situations though as a heavy user of macros and plugins in a parallel multi module build we can aid the testing in this and provide some feedback as to the impact in our environment

mkeskells:
I would also expect to measure a reduction in memory issue particularly in ide

retronym:
[`FileBasedCache`](https://github.com/scala/scala/blob/bcf47b165ccfd8e1827188f70aeb24e2cecfb80f/src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala#L183-L196), which this PR builds upon, uses [`BasicFileAttributes.fileKey`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html#fileKey())

retronym (approved):

mkeskells:
Thats the api that i could not remember

lrytz:
Instead of hashing the jar we could use the hashes in the zip entries

```
scala> import scala.collection.JavaConverters._; new java.util.zip.ZipFile("/Users/luc/scala/scala-2.12/lib/scala-library.jar").entries.asScala.map(_.getCrc).foldRight(0l)(_+_)
res1: Long = 5342018076077
```

Not sure if that's fast enough.

jvican:
@mkeskells Yeah, but remember that `fileKey` keeps being the same even if the contents of the file change.
  
> we can aid the testing in this and provide some feedback as to the impact in our environment
  
I would be very much interested in knowing how this patch affects your environment.

> Instead of hashing the jar we could use the hashes in the zip entries
  
(Those are erased by Bazel too, to the best of my understanding. All metadata in the zip entries is wiped out.)

@lrytz I proposed the same fix for an sbt/zinc regression back in the day: https://github.com/sbt/zinc/issues/433. We can have that too, but I feel that we should tackle the issue at the root of the problem: let the build tool, which knows the hashes for every classpath entry, communicate that to the compiler. This way we can reuse the hashes for other tools like Zinc or even third-party compiler plugins. In case build tools do not provide the hashes, we can fallback on the heuristics.

lrytz:
(Just googlin', but it seems to me the crc is mandatory by the zip spec https://www.loc.gov/preservation/digital/formats/digformatspecs/APPNOTE%2820120901%29_Version_6.3.3.txt)

jvican:
> (Just googlin', but it seems to me the crc is mandatory by the zip spec loc.gov/preservation/digital/formats/digformatspecs/APPNOTE%2820120901%29_Version_6.3.3.txt)
  
Yes, that was my understanding too. I think the bazel approach is just to zero that number. @johnynek Can you shed some light into this?

johnynek:
@Jvican I haven’t looked carefully, but I thought it was only the timestamp that is set to 1980-01-01 (not 1970 for some reason). CRC shouldn’t be a problem from a reproducibility standpoint since you want to invalidate when the contents change.

johnynek:
Using the size of the jar plus the crcs (if they are fast enough to read) seems like a good solution to me, but I’d want to do something like use murmur128 hash of all that data.

mkeskells:
Let's not get too bent out of shape by what these tools do. Optimise for the common idiomatic case

stuhood:
https://github.com/scala/scala/pull/6314#issuecomment-364168839

^ Agree with this as a good middle ground.

But having a generic ability to inject a statically loaded cache interface would be ideal, as that would let someone embedding the compiler (ie zinc, or someone consuming zinc) provide an implementation based on having already fingerprinted the files (most consumers of zinc will have already fingerprinted things before it starts up). See @romanowski's work in zinc with regard to "bring your own fingerprint" usecases.

johnynek:
BTW: thanks for working on this @jvican ... I hope we can ship a cache that is safe to use for all users, and does not assume any special assumptions of others tools, especially if it is in our grasp (CRC).

mkeskells:
Why don't you just use a filewatcher

mkeskells:
If using nio filesystem to modify the content the the file size will increase as it just appends the new index on the end. Index is always last entry.

@romanowski solution works for us for obvious reasons

I generally maintain that we should just focus on environments that present reality. If they lie about the metadata that they they can disable the caching or take the consequences. Just like a build system. If I hide the change that I have made by overwriting the sha codes  in git and messing with the metadata then I get what I deserve😁

Just my 2c

Btw it's 1980 not 1970 as a base date because that's the base for zip files I think.
Zip started as a windows utility so doesn't share the same concept of base time

jvican:
I still believe that we should have a compiler flag to enable this use case for the tools that need it, which are important build tools in the JVM ecosystem (sbt could also benefit from this in the long run). Nevertheless, I'm happy to implement the tradeoff we seem to converge to: "Using the size of the jar plus the crcs". I hope that the performance impact of looking at the CRCs is not significant.
  
I'd like to do that in a follow-up PR to avoid blocking on this one. Does this sound good to you @retronym?

gkossakowski:
Bazel scala rules disable classpath caching: https://github.com/bazelbuild/rules_scala/blob/scala2_12/scala/scala.bzl#L192-L197

They do for the reasons mentioned above: bazel zeroes the metadata that makes the cache invalidation cheap.

I think it's worth a shot to verify whether hashing all crcs is cheap enough. To get a representative sample, I suggest taking 500 jars from `.ivy` and measuring the cost.

Before writing a benchmark, it's worth checking whether perf expectations are within the right orders of magnitude. Let's assume that we want to spend at most 50ms on cache invalidation for 500 jars. Let's assume each jar has around 5000 entries which gives us 2,500,000 entries to read. That's 250,000 entries per 1 ms. I think that would get pretty close to CPU limits of how they can compute anything assuming no I/O is involved. With the added cost of the I/O, I think this idea might be too expensive to implement. I'm curious how my back-of-the-envelope calculations compare to actual benchmarking.

Generally, I'd lean towards asking the build tools to either not zero the meta data (the common case) or to provide fingerprints they already have computed (and cache themselves).

jvican:
@gkossakowski I addressed a regression on classpath hashing a few months ago in Zinc, and I actually have results for hashing 500 jars in the pull request I made: https://github.com/sbt/zinc/pull/452. And I also made the process parallel. There's also my attempt to use xxHash instead of a cryptographic hash (SHA-1) for classpath validation, but I closed it to prioritise other work. https://github.com/sbt/zinc/pull/371
  
Numbers, in a nutshell:

1. No parallel hashing of the first 500 jars in my ivy cache: 1330ms.
1. Parallel hashing of the first 500 jars in my ivy cache: 770ms.
1. Second (cached) parallel hashing of the first 500 jars in my ivy cache: 1ms.
  
> Generally, I'd lean towards asking the build tools to either not zero the meta data (the common case) or to provide fingerprints they already have computed (and cache themselves).
  
We're on the same page, and your last proposition is the one I'm trying to push for because I believe that asking Bazel to keep that metadata is not a plausible solution; if the CRCs inside the zip file have been computed out of the contents of a zip entry that contained some kind of metadata or filestamps and Bazel removes that metadata, it needs to update the CRC checksums and that is unnecessary IO IMO. The easiest is to let build tools share fingerprints with the compiler.

SethTisue:
I don't think Scabot recognizes multiple commands in the same comment (though I'd have to look at the source code to be sure...)

jvican:
CI is green. I'm improving the detection algorithm as discussed here in a follow-up PR today. It will be ready tomorrow.

lrytz:
Follow-up as in another commit here, or a separate PR? This could be merged as-is.

jvican:
I meant in a separate PR :smile:. This can be merged now.

retronym:
I'm not 100% sure, but our nightly build on windows seems to have [started failing](https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-windows/896/) after this change.

Example error:
```
!!  255 - run/macro-duplicate                       [compilation failed]
##### Log file 'Y:\jenkins\workspace\scala-2.12.x-integrate-windows\test\files\run\macro-duplicate-run.log' from failed test #####

Test_2.scala:5: error: macro implementation not found: foo
(the most common reason for that is that you cannot use macro implementations in the same compilation run that defines them)
  Macros.foo
         ^
one error found
```

jvican:
@retronym I've just had a look at the test. One thing I don't understand is what partest is doing to make `macro-duplicate` work. In theory, the test snippet shouldn't even compile because the macro is in the same compilation unit than the test code. Can you shed some light into it?

lrytz:
I'm taking a look

src/compiler/scala/tools/nsc/typechecker/Macros.scala
```diff
@@ -60,8 +63,24 @@ trait Macros extends MacroRuntimes with Traces with Helpers {
...
-    macroLogVerbose("macro classloader: initializing from -cp: %s".format(classpath))
-    ScalaClassLoader.fromURLs(classpath, self.getClass.getClassLoader)
+    def newLoader = () => {
+      macroLogVerbose("macro classloader: initializing from -cp: %s".format(classpath))
+      ScalaClassLoader.fromURLs(classpath, self.getClass.getClassLoader)
+    }
+
+    import scala.tools.nsc.io.Jar
+    import scala.reflect.io.{AbstractFile, Path}
+    val locations = classpath.map(u => Path(AbstractFile.getURL(u).file))
```
adriaanm, line 73: Cross-referencing with regression report in https://github.com/scala/scala-dev/issues/480

fommil:
when I tried out the macro caching on https://gitlab.com/fommil/scalaz-deriving/ it slowed down the build by a few seconds. The compiler flag made no difference.

But it's a tiny project and compiles in under 10 seconds so it's hard to say anything from this data point.

PR #6315, opened 2018-02-08
Deprecate string-building using `+` with a non-`String` type on the left (aka `any2stringadd`)

eed3si9n:
Ref scala/bug#194

1. This deprecates scala.Predef.any2stringadd.
2. Another thing it does is to remove the conversion altogether under `-Xsource:2.14` flag.

### Background

Scala up till 2.12 has had two kinds of String concatenation operator `+` (also known as PLUS or ADD).

a. When the receiver is a non-String, it uses implicit `Predef.any2addstring` to inject `+`.
b. When the receiver is a String, the compiler injects a special method `String_+`.

In other words, `true + "what"` and `"what" + true` use different mechanisms.
This change addresses the first variant that introduces `+` operator into `Any`.
The rationale discussed in scala/bug#194 is that this implicit injection removes too much type safety, leading to confusing results.

Here some examples listed:

```scala
scala> var v = Vector[Int]()
v: scala.collection.immutable.Vector[Int] = Vector()

scala> v += 3
<console>:13: error: value += is not a member of scala.collection.immutable.Vector[Int]
  Expression does not convert to assignment because:
    type mismatch;
     found   : Int(3)
     required: String
    expansion: v = v.$plus(3)
       v += 3
         ^

scala> val xs = List(1, 2, 3)
xs: List[Int] = List(1, 2, 3)

scala> xs foreach { println _ + 1 }
 found   : Int(1)
 required: String
       xs foreach { println _ + 1 }
                                ^
```

Note that in both examples the person who wrote the code were not trying to concatenate String.

### Implementation note

Initially I've attempted to implement what @adriaanm  proposed in 2014 (scala/bug#194 (comment)) and remove all deprecated implicits under `-Xfuture` flag. A simple implementation

```scala
val fd = settings.future && sym.isDeprecated
```

did not work, because the implicit caching is unware of the annotations?

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -853,6 +853,13 @@ trait Contexts { self: Analyzer =>
     private def isQualifyingImplicit(name: Name, sym: Symbol, pre: Type, imported: Boolean) =
       sym.isImplicit &&
       isAccessible(sym, pre) &&
+      !({
+        // [eed3si9n] ideally I'd like to do this: val fd = settings.isScala214 && sym.isDeprecated
+        // but implicit caching currently does not report sym.isDeprecated correctly.
+        val fd = settings.isScala214 && (name ne null) && definitions.isPredefMemberNamed(sym, TermName("any2stringadd"))
+        if (settings.XlogImplicits && fd) echo(sym.pos, sym + " is not a valid implicit value because:\n" + "-Xsource:2.14 removes scala.Predef.any2stringadd")
+        fd
+      }) &&
       !(imported && {
         val e = scope.lookupEntry(name)
         (e ne null) && (e.owner == scope) && (!settings.isScala212 || e.sym.exists)
--- a/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Contexts.scala
@@ -856,7 +856,7 @@ trait Contexts { self: Analyzer =>
       !({
         // [eed3si9n] ideally I'd like to do this: val fd = settings.isScala214 && sym.isDeprecated
         // but implicit caching currently does not report sym.isDeprecated correctly.
-        val fd = settings.isScala214 && (name ne null) && definitions.isPredefMemberNamed(sym, TermName("any2stringadd"))
+        val fd = settings.isScala214 && (sym == currentRun.runDefinitions.Predef_any2stringaddMethod)
         if (settings.XlogImplicits && fd) echo(sym.pos, sym + " is not a valid implicit value because:\n" + "-Xsource:2.14 removes scala.Predef.any2stringadd")
         fd
       }) &&
--- a/src/library/scala/Predef.scala
+++ b/src/library/scala/Predef.scala
@@ -15,6 +15,7 @@ import scala.collection.{ mutable, immutable, ArrayOps }
 import scala.collection.immutable.{ ArraySeq, WrappedString }
 import scala.annotation.{ elidable, implicitNotFound }
 import scala.annotation.elidable.ASSERTION
+import scala.annotation.meta.companionMethod
 
 /** The `Predef` object provides definitions that are accessible in all Scala
  *  compilation units without explicit qualification.
@@ -366,6 +367,7 @@ object Predef extends LowPriorityImplicits {
 
   // scala/bug#8229 retaining the pre 2.11 name for source compatibility in shadowing this implicit
   /** @group implicit-classes-any */
+  @(deprecated @companionMethod)("Implicit injection of + is deprecated. Convert to String to call +", "2.13.0")
   implicit final class any2stringadd[A](private val self: A) extends AnyVal {
     def +(other: String): String = String.valueOf(self) + other
   }
--- a/src/reflect/scala/reflect/internal/Definitions.scala
+++ b/src/reflect/scala/reflect/internal/Definitions.scala
@@ -1504,6 +1504,7 @@ trait Definitions extends api.StandardDefinitions {
       lazy val Predef_classOf      = getMemberMethod(PredefModule, nme.classOf)
       lazy val Predef_implicitly   = getMemberMethod(PredefModule, nme.implicitly)
       lazy val Predef_???          = DefinitionsClass.this.Predef_???
+      lazy val Predef_any2stringaddMethod = getMemberMethod(PredefModule, nme.any2stringadd).suchThat(_.isMethod)
 
       lazy val arrayApplyMethod       = getMemberMethod(ScalaRunTimeModule, nme.array_apply)
       lazy val arrayUpdateMethod      = getMemberMethod(ScalaRunTimeModule, nme.array_update)
--- a/src/reflect/scala/reflect/internal/StdNames.scala
+++ b/src/reflect/scala/reflect/internal/StdNames.scala
@@ -491,7 +491,7 @@ trait StdNames {
       if (isConstructorName(name))
         DEFAULT_GETTER_INIT_STRING + pos
       else
-        name + DEFAULT_GETTER_STRING + pos
+        name.toString + DEFAULT_GETTER_STRING + pos
     )
     // Nominally, name from name$default$N, CONSTRUCTOR for <init>
     def defaultGetterToMethod(name: Name): TermName = (
@@ -654,6 +654,7 @@ trait StdNames {
     val accessor: NameType             = "accessor"
     val add_ : NameType                = "add"
     val annotation: NameType           = "annotation"
+    val any2stringadd: NameType        = "any2stringadd"
     val anyHash: NameType              = "anyHash"
     val anyValClass: NameType          = "anyValClass"
     val apply: NameType                = "apply"
--- a/test/files/jvm/future-spec/FutureTests.scala
+++ b/test/files/jvm/future-spec/FutureTests.scala
@@ -202,13 +202,13 @@ class FutureTests extends MinimalScalaTest {
         a <- future0.mapTo[Int]  // returns 5
         b <- async(a)            // returns "10"
         c <- async(7)            // returns "14"
-      } yield b + "-" + c
+      } yield s"$b-$c"
 
       val future2 = for {
         a <- future0.mapTo[Int]
         b <- (Future { (a * 2).toString }).mapTo[Int]
         c <- Future { (7 * 2).toString }
-      } yield b + "-" + c
+      } yield s"$b-$c"
 
       Await.result(future1, defaultTimeout) mustBe ("10-14")
       assert(checkType(future1, manifest[String]))
@@ -227,13 +227,13 @@ class FutureTests extends MinimalScalaTest {
         Res(a: Int) <- async(Req("Hello"))
         Res(b: String) <- async(Req(a))
         Res(c: String) <- async(Req(7))
-      } yield b + "-" + c
+      } yield s"$b-$c"
 
       val future2 = for {
         Res(a: Int) <- async(Req("Hello"))
         Res(b: Int) <- async(Req(a))
         Res(c: Int) <- async(Req(7))
-      } yield b + "-" + c
+      } yield s"$b-$c"
 
       Await.result(future1, defaultTimeout) mustBe ("10-14")
       intercept[NoSuchElementException] { Await.result(future2, defaultTimeout) }
--- a/test/files/jvm/future-spec/main.scala
+++ b/test/files/jvm/future-spec/main.scala
@@ -65,7 +65,7 @@ trait MinimalScalaTest extends Output with Features {
 
   implicit def objectops(obj: Any) = new {
 
-    def mustBe(other: Any) = assert(obj == other, obj + " is not " + other)
+    def mustBe(other: Any) = assert(obj == other, s"$obj is not $other")
     def mustEqual(other: Any) = mustBe(other)
 
   }
--- a/test/files/neg/implicit-any2stringadd-warning.check
+++ b/test/files/neg/implicit-any2stringadd-warning.check
@@ -1,0 +1,6 @@
+implicit-any2stringadd-warning.scala:2: warning: method any2stringadd in object Predef is deprecated (since 2.13.0): Implicit injection of + is deprecated. Convert to String to call +
+  true + "what"
+  ^
+error: No warnings can be incurred under -Xfatal-warnings.
+one warning found
+one error found
--- a/test/files/neg/implicit-any2stringadd-warning.flags
+++ b/test/files/neg/implicit-any2stringadd-warning.flags
@@ -1,0 +1,1 @@
+-Xfatal-warnings -deprecation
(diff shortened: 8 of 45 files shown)


hrhino:
At last!

In what situation does `settings.future && sym.isDeprecated` not work?

I do think that the "deprecation removes implicit eligibility under `-Xfuture`" feature should be written down somewhere other than a spare comment in `Contexts`. The scaladoc would make sense to me. 

Also, is `-Xfuture` sticking around? I guessed that it might be getting folded in with `-Xsource:2.(n+1)`, but maybe that's just `-Xexperimental`.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -830,6 +830,13 @@ trait Contexts { self: Analyzer =>
     private def isQualifyingImplicit(name: Name, sym: Symbol, pre: Type, imported: Boolean) =
       sym.isImplicit &&
       isAccessible(sym, pre) &&
+      !({
+        // [eed3si9n] ideally I'd like to do this: val fd = settings.future && sym.isDeprecated
+        // but implicit caching currently does not report sym.isDeprecated correctly.
```
eed3si9n: @hrhino In this position, I originally wrote `val fd = settings.future && sym.isDeprecated`, but sym here didn't return `isDeprecated` at least with the tests that I was using. I suspect https://github.com/scala/scala/blob/v2.13.0-M3/src/reflect/scala/reflect/internal/Symbols.scala#L3426-L3434 caching without the annotation.

eed3si9n:
@hrhino 

> Also, is `-Xfuture` sticking around? I guessed that it might be getting folded in with `-Xsource:2.(n+1)`, but maybe that's just `-Xexperimental`.

You're right - https://github.com/scala/scala-dev/issues/430.

I'll change to `-Xsource:2.14`.

dwijnand:
Removing `String#+(Any)` was closed because it lacked a SIP. Does this one not?

https://github.com/scala/scala/pull/5235#issuecomment-280444249

eed3si9n:
@dwijnand Does an implementation come first or the SIP? I thought it got closed because we ran out of time.

sjrd:
`String.+(Any)` doesn't cause confusion, and remains the most cross-platformly efficient way to convert to string + concatenate. It is also quite necessary as a primitive in Scala.js for various reasons. In particular, for `x` a primitive type (e.g., `Int`), `x.toString()` is ultimately implemented in terms of `"" + x`, not the other way around. I would be strongly opposed to its removal.

If someone wants to remove it and prepares an implementation, make sure to also prepare an implementation in Scala.js, so that you feel the pain of its absence ;)

eed3si9n:
@dwijnand @sjrd https://github.com/scala/bug/issues/194 is titled "any2stringadd implicit should be removed from Predef." This PR proposes the following actions:
- Scala 2.13: Deprecate `Predef.any2stringadd`. Remove when used with `-Xsource:2.14`
- Scala 2.14: Remove `Predef.any2stringadd`

Whether the deprecation/removal is good idea or not I think should continue in https://github.com/scala/bug/issues/194 to avoid duplication. (I'd be happy to discuss implementation on this PR here of course)

As it stands, this PR doesn't say anything about the `String#+(Any)` operator; and no one has told me (yet) that the deprecation/removal of `Predef.any2stringadd` must coincide with `String#+(Any)`.

Ichoran:
I find `any2stringadd` a dreadfully type-unsafe nuisance and never have trouble with `String#+(Any)`, so this seems like exactly the right way to go to me.

dwijnand:
I guess I'm the lone idiot who's been accidentally appending `Option[String]`s (and the like) to strings, not realising it was an option of string.

eed3si9n:
Looks like Jenkins is back to operational - https://scala-ci.typesafe.com/job/scala-2.13.x-validate-main/53/execution/node/51/log/?consoleFull

Jasper-M:
I personally find `String#+(Any)` to also be a type-unsafe nuisance. And also a pain in the ass when developing a DSL. At least you can turn off `any2stringadd` by [shadowing its name](https://github.com/Jasper-M/stapl-core/blob/partial-evaluation/main/src/main/scala/stapl/core/dsl.scala#L28). There's no easy way to make `String#+(Any)` go away.

SethTisue:
I suggest we keep this PR focused on `any2stringadd`, and take up `String#+(Any)` separately.

SethTisue:
@eed3si9n would you mind squashing this into a single commit...?

eed3si9n:
@SethTisue Done.

eed3si9n:
Rebased to catch up with https://github.com/scala/scala/pull/6093

SethTisue:
```
[test] !!  786 - run/t10551.scala                          [output differs]
[test] % diff /home/jenkins/workspace/scala-2.13.x-validate-main@2/test/files/run/t10551-run.log /home/jenkins/workspace/scala-2.13.x-validate-main@2/test/files/run/t10551.check
[test] @@ -1,4 +1,3 @@
[test] -warning: there was one deprecation warning (since 2.13.0); re-run with -deprecation for details
```

eed3si9n:
Updated the test with toString.

eed3si9n:
@SethTisue What's next? Is there a process of picking a reviewer?

SethTisue:
> Is there a process of picking a reviewer?

blindfold yourself, then throw a dart at a printout of https://github.com/scala/scala/blob/2.13.x/README.md#get-in-touch ?

I'd suggest @lrytz for a first pass, and then @adriaanm should probably sign off on anything as fundamental as this.

lrytz (commented):
💟

lrytz (approved):

eed3si9n:
Rebased against 2.13.x since a few tests conflicted.

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -838,6 +838,13 @@ trait Contexts { self: Analyzer =>
     private def isQualifyingImplicit(name: Name, sym: Symbol, pre: Type, imported: Boolean) =
       sym.isImplicit &&
       isAccessible(sym, pre) &&
+      !({
+        // [eed3si9n] ideally I'd like to do this: val fd = settings.isScala214 && sym.isDeprecated
+        // but implicit caching currently does not report sym.isDeprecated correctly.
+        val fd = settings.isScala214 && (name ne null) && definitions.isPredefMemberNamed(sym, TermName("any2stringadd"))
```
adriaanm: This will need some finessing, as I assume this is performance critical. I suggest caching the `any2stringadd` symbol and comparing directly against it, potentially along with the other caches just above. I'd also split this part of the condition into a separate method `isDeprecatedImplicit`. I'm going to defer my review to @retronym, since he's more familiar with the performance trade offs here.

lrytz: ping @retronym

retronym: Yes, it should be `sym == currentRun.runDefintions.Predef_ any2stringadd`, after adding `lazy val Predef_ any2stringadd = ....` to `RunDefintions`.

eed3si9n: The `....` part where it's left to the reader as an exercise turned out to be oddly tricky (for me). I could not get `sym` to equal `Predef_any2stringadd` when I defined it as `getMemberMethod(PredefModule, nme.any2stringadd)`.

It turns out that `any2stringadd` is an implicit class, and the implicit class creates both a companion object AND a factory method also named `def any2stringadd[A](a: A)`! This is not normally allowed, right?

What seems to be working is:

```scala
PredefModule.info.decl(nme.any2stringadd.toTermName).suchThat(_.isMethod)
```

eed3si9n:
Rebased against 2.13.x.

lrytz (approved):
Pushed a tiny change to the way the method is looked up (https://github.com/scala/scala/pull/6315/commits/a9d5d38e89c6ddcdfbb74a36785304995920b02a#diff-2b1bf15b53df94495f7155a69f822adeR1507). LGTM, will merge when it goes green!

SethTisue:
@eed3si9n apparently some other code path is involved here:

```
scala> 3 + "foo"
res1: String = 3foo
```

note the absence of deprecation warning.

sjrd:
https://github.com/scala/scala/blob/e33743f31a0ac6c7adefc4a8d4f9d67d43c32dbb/src/library/scala/Int.scala#L45

All primitive types explicitly declare a `def +(s: String): String`, so `prim + s` doesn't use the `implicit any2StringAdd`. You would have to deprecate those explicit `def`s as well.

smarter:
> All primitive types explicitly declare a def +(s: String): String, so prim + s doesn't use the implicit any2StringAdd. You would have to deprecate those explicit defs as well.

I was randomly looking at Int.scala and rediscovered this, so I made a PR for that exact purpose: https://github.com/scala/scala/pull/6755

src/compiler/scala/tools/nsc/typechecker/Contexts.scala
```diff
@@ -853,6 +853,13 @@ trait Contexts { self: Analyzer =>
     private def isQualifyingImplicit(name: Name, sym: Symbol, pre: Type, imported: Boolean) =
       sym.isImplicit &&
       isAccessible(sym, pre) &&
+      !({
+        // [eed3si9n] ideally I'd like to do this: val fd = settings.isScala214 && sym.isDeprecated
+        // but implicit caching currently does not report sym.isDeprecated correctly.
+        val fd = settings.isScala214 && (sym == currentRun.runDefinitions.Predef_any2stringaddMethod)
+        if (settings.XlogImplicits && fd) echo(sym.pos, sym + " is not a valid implicit value because:\n" + "-Xsource:2.14 removes scala.Predef.any2stringadd")
```
retronym, line 860: This logging is pretty noisy. When is intended to be shown?

```
$ qscalac -Xlog-implicits -Xsource:2.14 -d /tmp src/scalap/**/*.scala
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/Classfile.scala:122: materializing requested scala.reflect.type.ClassTag[String] using scala.reflect.`package`.materializeClassTag[String]()
      for ((x, i) <- entries.zipWithIndex ; if x != null) yield
                  ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/scalax/rules/scalasig/ClassFileParser.scala:233: materializing requested scala.reflect.type.ClassTag[Option[Any]] using scala.reflect.`package`.materializeClassTag[Option[Any]]()
  private val values = Array.fill[Option[Any]](size)(None)
                                                    ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/Decode.scala:57: refArrayOps is not a valid implicit value for bytes.type => ?{def take: ?} because:
inferred type arguments [Byte] do not conform to method refArrayOps's type parameter bounds [T <: AnyRef]
      bytes take length
      ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/Main.scala:109: materializing requested scala.reflect.type.ClassTag[String] using scala.reflect.`package`.materializeClassTag[String]()
        classname.split('.').map(NameTransformer.encode _).mkString(".")
                                ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/scalax/rules/Memoisable.scala:50: error: value + is not a member of AnyRef
      if(DefaultMemoisable.debug) println(key + " -> " + other)
                                              ^
src/scalap/scala/tools/scalap/scalax/rules/Memoisable.scala:56: error: value + is not a member of AnyRef
    if(DefaultMemoisable.debug) println(key + " -> " + t + " (" + out + ")")
                                            ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/scalax/rules/scalasig/ScalaSig.scala:33: wrapRefArray is not a valid implicit value for Array[Byte] => scala.collection.IterableOnce[?] because:
inferred type arguments [Byte] do not conform to method wrapRefArray's type parameter bounds [T <: AnyRef]
      case ConstValueIndex(index) => bytesForIndex(index)
                                                  ^
src/scalap/scala/tools/scalap/scalax/rules/scalasig/ScalaSig.scala:34: materializing requested scala.reflect.type.ClassTag[Byte] using scala.reflect.`package`.materializeClassTag[Byte]()
    }.toArray
      ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
src/scalap/scala/tools/scalap/scalax/rules/scalasig/ScalaSig.scala:45: refArrayOps is not a valid implicit value for bytes.type => ?{def take: ?} because:
inferred type arguments [Byte] do not conform to method refArrayOps's type parameter bounds [T <: AnyRef]
        ScalaSigAttributeParsers.parse(ByteCode(bytes.take(length)))
                                                ^
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
method any2stringadd is not a valid implicit value because:
-Xsource:2.14 removes scala.Predef.any2stringadd
two errors found
```

eed3si9n: I intended it to be called only when any2stringadd was invoked, but it seems like that's not the case.

PR #6319, opened 2018-02-09
scala/bug#10033 Remove items deprecated in 2.12 or earlier

### LINKED ISSUE #7308: Revert the deprecation of set in SyncVar
For some very strange reason the set method of SyncVar has been marked for deprecation. Please remove the deprecation and do not deprecate the set method.

The set method is what makes SyncVar a var in the first place. If the SyncVar only has put, then it will effectively become a size 1 bounded queue and is no longer a variable.

SyncVar needs to act like a variable. There must be a way to set the variable twise without reading or clearing it in between.

It must support the following flow:
- Thread 1 set variable to "a"
- Thread 1 set variable to "b" (discards value "a")
- Thread 2 read variable, get value "b"
- Thread 1 set variable to "c" (discards value "b")
- Thread 1 set variable to "d" (discards value "c")
- Thread 2 read variable, get value "d"

Alternatively stop calling it a SyncVar, and rename it to BoundedQueue1

### LINKED ISSUE #9641: timeout of executor.awaitTermination does not work for ForkJoinPool
For the description from API for ForkJoinPoll, when we call shutdown method following the awaitTermination(long,TimeUnit), the main thread should not be blocked when the timeout occurs.
But using scala.concurrent.forkjoin.ForkJoinPool, it does not work.
The forkjoin package in Java Library does's occur that problem.

import scala.concurrent.forkjoin._
val exector = new ForkJoinPool()
  exector.execute( new Runnable {
    override def run(): Unit = {
      while(true){}
    }
  })
  exector.shutdown()
  exector.awaitTermination(3l, TimeUnit.SECONDS)

NthPortal:
Removes most items deprecated in 2.12 and earlier, with the following
exceptions:

`scala.reflect.ClassManifestDeprecatedApis` is not removed, because it
is used by `scala.reflect.Manifest`, which is not deprecated. For the
same reason, the definitions of `ClassManifest` in the `scala.reflect`
package object are also not removed.

`scala.io.Position` is not removed, because it is is used by
`scala.io.Source.Positioner`, which is not deprecated. It may be
removed from the standard library with `scala.io.Source` (and perhaps
un-deprecated then).

Deprecations in `scala.collections` are skipped (unless referenced
outside the package), because the package is being rewritten for 2.13.

`scala.DelayedInit` is not removed, because there is not yet any
replacement implementation for `scala.App`.

`scala.runtime.StringAdd` is not removed, for bootstrap reasons. It can
be removed in 2.13.0-M5.

`scala.concurrent.Future.fold` and `scala.concurrent.Future.reduce` are
not removed, to facilitate cross-building between 2.11/2.12/2.13.

Fixes scala/bug#9641

Closes scala/bug#7308 (as won't fix)


NthPortal:
I've broken the changes up into different commits, per https://github.com/scala/bug/issues/10033#issuecomment-328679219. I'm happy to squash them if desired.

NthPortal:
I've also not marked this PR as fixing 10033 due to the `StringAdd` issue. I'll make a PR for 2.13.0-M5 to remove `StringAdd` and close the issue.

SethTisue:
this looks like an appropriate granularity of commits to me.

retronym:
Just to set expectations here: part of removing deprecated items is evaluating the effect on the community build. If a removal affects a small number of projects, we need to reach out to them and ask them to adapt (perhaps by sending a PR to the project). If causes widespread breakage, we might need to delay the change.

This avoids us from "flying blind" without a community build.

SethTisue:
good point, Jason. I'll try to be proactive about identifying PRs that need community build runs before merge (and about actually doing the runs, too).

NthPortal:
Not sure how to get `combined` to update

NthPortal:
I'm not sure how community builds are run, but I've fixed up all the broken (par)tests at this point, so it's good to go I think.

NthPortal:
While looking into scala/bug#9554, I noticed that the information regarding how certain `Throwable`s are handled is present only in the docs for `onFailure`, and not `onComplete`. I'm going to copy the relevant info into the docs for `onComplete`, so it's not lost with these changes.

NthPortal:
@SethTisue could you run a community build? Thanks!

SethTisue:
queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/951/ ~~(link will be 404 til Jenkins queue clears)~~

NthPortal:
cheers!

SethTisue:
it's still running, but so far:

* scala-async is using `future` (lower case f), `Future.fold`, `Future.reduce`
* macro-paradise is using `tpnme.bridgeAnnot`
* scala-swing is using `Future#onSuccess`
* scalaz is using `PartialFunction.apply`

fixing these may uncover further errors in the same projects or in downstream projects

NthPortal:
So far, nothing looks terrible to fix

- scala-async is easy to fix - perhaps as easy as find-replace
- macro-paradise looks at first glance like it just needs the changes @hrhino made to be ported over; however, it's possible the relevant code is exposed to clients
- scala-swing with `Future.onSuccess` is annoying, because it's not a simple rename, but it still shouldn't be too hard (unless it's unclear what to do in the failure case)
- scalaz should be reasonably easy to fix; it's not a simple rename, but the rewrite isn't too difficult

SethTisue:
remaining failures:

* genjavadoc — I think it's choking on the Scala version number, doesn't have to do with your change. I'll fix and re-run, since there are a number of downstream projects.
* scala-stm — uses `ClassManifest` (I left a comment on https://github.com/nbronson/scala-stm/issues/56)

SethTisue:
new run with genjavadoc fix: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/953/

UPDATE: scala-java8-compat is using `onSuccess`; no other new failures

SethTisue:
scalaz PR: https://github.com/scalaz/scalaz/pull/1654

SethTisue:
@viktorklang how do you feel about https://github.com/scala/scala/pull/6319/commits/b9e4b0b3eda89cfd3efffa01ff896bbf05825dca ? do you want the removals to go forward for 2.13?

NthPortal:
scala-async PR: scala/scala-async#188

I've discovered a problem, which is that `foldLeft` and `reduceLeft` didn't exist until 2.12, so it's difficult to remove `fold` and `reduce` from projects that support 2.11.

Should we hold off on removing `fold` and `reduce` until 2.14?

NthPortal:
scala-swing PR: scala/scala-swing#72 (builds fine)

NthPortal:
`ClassManifest` should be easy to fix as well because

- it's just an alias for `ClassTag` anyway, at this point
- it still exists in `scala.reflect`

I'm working on scala-java8-compat right now

NthPortal:
scala-java8-compat PR: scala/scala-java8-compat#100

viktorklang:
@SethTisue Thanks for the notice, Seth. I've commented.

SethTisue:
@NthPortal anything you add back in response to Viktor should also get a comment, I think

NthPortal:
@SethTisue do you want me to add it back in a separate commit, or rewrite b9e4b0b3eda89cfd3efffa01ff896bbf05825dca?

NthPortal:
@SethTisue @viktorklang what should be done about [`Future.fold` and `Future.reduce`](https://github.com/scala/scala/pull/6319#issuecomment-367558041) - leave them in for now?

NthPortal:
@viktorklang I left replies to some of your comments (and took care of the rest)

SethTisue:
> @SethTisue do you want me to add it back in a separate commit, or rewrite b9e4b0b?

rewrite, but it doesn't have to be right away

some folks prefer to squash/rebase as they go, other folks prefer to wait until all reviewers have signed off, then do it once at the very end. either way is fine.

viktorklang:
@NthPortal What is the problem with (removing) reduce and fold?

SethTisue:
@viktorklang the problem is that the recommended alternatives, `reduceLeft` and `foldLeft`, weren't added until 2.12. so if someone wants to crossbuild their code against 2.11, 2.12, and 2.13, they'll have to find some kind of workaround. this is precisely the situation we hit in https://github.com/scala/scala-async/pull/188

this gets into a broader issue here of how we far we are willing to bend, in general, in order to preserve cross-building across more than two major Scala versions. cross-building against two major versions is definitely a goal of ours to support, against three it's much less clear that it's worth doing.

the argument for three over two is that there is a window — a window we are in _right now_ with 2.13 — where we want library authors to be supporting 2.13.0-Mx, but we don't really expect them to drop 2.11 yet (not until 2.13.0 final is out!)

so, ugh, these decisions are kind of hard, actually. on something minor like this, perhaps the benefit of waiting for 2.14 to remove exceeds the cost of keeping the methods around (where they clutter up the code, clutter up Scaladoc, etc). idk

NthPortal:
The overhead cost for cross-building is non-trivial. It requires creating a utility object for each Scala version, with `foldLeft` or `reduceLeft` methods which forward to the appropriate method depending on Scala version. It's a decent amount of extra work and clutter, and the reason that I *didn't* just do that for scala-async to fix the PR.

viktorklang:
@SethTisue The code for reduce/fold is rather trivial, so any user who wants to maintain the exact same implementation should be able to do so. I don't have a good sense on how frequently used they are.

SethTisue:
@NthPortal what would just inlining the implementation look like, e.g. in scala-async?

NthPortal:
@SethTisue it probably looks like creating a utility object like this

```scala
private object FutureSpec {
  def foldLeft[T, R](futures: scala.collection.immutable.Iterable[Future[T]])(zero: R)(op: (R, T) => R)(implicit executor: ExecutionContext): Future[R] =
    foldNext(futures.iterator, zero, op)

  def reduceLeft[T, R >: T](futures: scala.collection.immutable.Iterable[Future[T]])(op: (R, T) => R)(implicit executor: ExecutionContext): Future[R] = {
    val i = futures.iterator
    if (!i.hasNext) Future.failed(new NoSuchElementException("reduceLeft attempted on empty collection"))
    else i.next() flatMap { v => foldNext(i, v, op) }
  }

  private[this] def foldNext[T, R](i: Iterator[Future[T]], prevValue: R, op: (R, T) => R)(implicit executor: ExecutionContext): Future[R] =
    if (!i.hasNext) Future.successful(prevValue)
    else i.next().flatMap { value => foldNext(i, op(prevValue, value), op) }
}
```

and then changing `Future.foldLeft => FutureSpec.foldLeft` and `Future.reduceLeft => FutureSpec.reduceLeft`

NthPortal:
Actually, I take that back. Having to worry about copyright significantly increases the cost of just inlining the current implementation. You need to worry about either including a copyright notice or changing the implementation so it's not a copy.

SethTisue:
yeah. I withdrew my inlining suggestion, but not until after you'd gotten the notification email, obviously :-/

NthPortal:
oops.

NthPortal:
@SethTisue what do you think about `fold`/`reduce` at this point?

SethTisue:
> what do you think about fold/reduce at this point?

let's hold off and remove `foldLeft` and `reduceLeft` in 2.14. removing stuff is good, but we don't need to rush the timetable unless the old method is actually causing trouble

I'd suggest putting a comment on the methods, something like `// not removed yet in 2.13, to facilitate 2.11/2.12/2.13 cross-building; let's remove it in 2.14; see #6319` so the next boy/girl scout who comes through knows what's up

NthPortal:
Sounds good to me - that was my thought as well

test/files/jvm/future-spec/FutureTests.scala
```diff
@@ -60,14 +60,14 @@ class FutureTests extends MinimalScalaTest {
       val f3 = f2 map { s => s.toUpperCase }
 
       f2 foreach { _ => throw new ThrowableTest("dispatcher foreach") }
-      f2 onSuccess { case _ => throw new ThrowableTest("dispatcher receive") }
+      f2 onComplete { case Success(_) => throw new ThrowableTest("dispatcher receive"); case _ => }
```
NthPortal, line 63: @viktorklang You mentioned that `foreach` would make more sense than `onComplete` here. If that's the case, since the previous line already tests `foreach`, should this line just be removed? I don't know if the original intent was to check if `onSuccess` worked, and if it makes sense to test if `onComplete` works as well.

test/files/jvm/future-spec/FutureTests.scala
```diff
@@ -60,14 +60,14 @@ class FutureTests extends MinimalScalaTest {
...
 
       latch.open()
 
       Await.result(f2, defaultTimeout) mustBe ("success")
 
       f2 foreach { _ => throw new ThrowableTest("current thread foreach") }
-      f2 onSuccess { case _ => throw new ThrowableTest("current thread receive") }
+      f2 onComplete { case Success(_) => throw new ThrowableTest("current thread receive"); case _ => }
```
NthPortal, line 70: @viktorklang Same question as [line 63](https://github.com/scala/scala/pull/6319/commits/691ffa7efc6fe4b3f2a1d1f5ef29564303560ffe#r171052827)

test/files/jvm/future-spec/FutureTests.scala
SethTisue: @NthPortal I think you can safely assume Viktor was just reading through the diffs quickly and making style suggestions without realizing that this is in a special testing context

SethTisue:
@NthPortal I think this might be ready for final review and merge, do you agree?

every downstream project doesn't need to be compatible with this in advance, we just want to assess whether it will be nbd to fix other projects up as needed, and it looks to me like we're fine there.

NthPortal:
@SethTisue sounds good to me

test/files/jvm/future-spec/FutureTests.scala
NthPortal: ok

SethTisue (approved):

SethTisue:
@viktorklang I think this is good to go except for possibly https://github.com/scala/bug/issues/7308 ? that ticket appears not to have been noticed (since it was filed in 2013), I don't know if it has merit.

SethTisue:
🎉

SethTisue:
downstream:

* https://github.com/scala/scala-async/issues/191
* https://github.com/scalamacros/paradise/issues/114
* https://github.com/nbronson/scala-stm/issues/56

once the paradise one is addressed, that may uncover further issues farther downstream. (the other two are not pressing.)

PR #6321, opened 2018-02-13
Added type parameter to Iterator.empty and test cases

kelebra:
This PR addresses issue [#10711](https://github.com/scala/bug/issues/10711) adding type parameter to method `Iterator.empty` and two test cases to `IteratorTest`.

### DIFF
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -38,7 +38,7 @@ object Iterator {
   }
 
   /** The iterator which produces no values. */
-  def empty[T]: Iterator[T] = _empty
+  @inline final def empty[T]: Iterator[T] = _empty
 
   /** Creates an iterator which produces a single element.
    *  '''Note:''' Equivalent, but more efficient than Iterator(elem)
--- a/src/library/scala/collection/Iterator.scala
+++ b/src/library/scala/collection/Iterator.scala
@@ -32,11 +32,13 @@ object Iterator {
     def traversableToColl[B](t: GenTraversable[B]) = t.toIterator
   }
 
-  /** The iterator which produces no values. */
-  val empty: Iterator[Nothing] = new AbstractIterator[Nothing] {
+  private[this] final val _empty: Iterator[Nothing] = new AbstractIterator[Nothing] {
     def hasNext: Boolean = false
     def next(): Nothing = throw new NoSuchElementException("next on empty iterator")
   }
+
+  /** The iterator which produces no values. */
+  def empty[T]: Iterator[T] = _empty
 
   /** Creates an iterator which produces a single element.
    *  '''Note:''' Equivalent, but more efficient than Iterator(elem)
--- a/test/junit/scala/collection/IteratorTest.scala
+++ b/test/junit/scala/collection/IteratorTest.scala
@@ -296,4 +296,16 @@ class IteratorTest {
     assertEquals(v2, v4)
     assertEquals(Some(v1), v2)
   }
+
+  @Test def emptyTypedIteratorsShouldBeEqual(): Unit = {
+    val emptyDoubleIterator = Iterator.empty[Double]
+    val emptyIntIterator = Iterator.empty[Int]
+    assertEquals(emptyDoubleIterator, emptyIntIterator)
+  }
+
+  @Test def emptyIteratorInHigherOrderFunctions(): Unit = {
+    val seqOfIterators = Seq(Seq(1, 2, 3).iterator, Seq(3, 2, 1).iterator, Seq(1, 3, 2).iterator)
+    val unified = seqOfIterators.fold(Iterator.empty[Int])((a, b) => a ++ b)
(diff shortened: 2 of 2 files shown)


src/library/scala/collection/Iterator.scala
```diff
@@ -32,12 +32,14 @@ object Iterator {
...
   }
 
-  /** The iterator which produces no values. */
-  val empty: Iterator[Nothing] = new AbstractIterator[Nothing] {
+  private[this] final val _empty: Iterator[Nothing] = new AbstractIterator[Nothing] {
     def hasNext: Boolean = false
     def next(): Nothing = throw new NoSuchElementException("next on empty iterator")
   }
 
+  /** The iterator which produces no values. */
```
mkeskells, line 40: @retronym we need to ensure this is inlined as a getter. Similar issue to "tl" in :: ?

test/junit/scala/collection/IteratorTest.scala
```diff
@@ -296,4 +296,16 @@ class IteratorTest {
     assertEquals(v2, v4)
     assertEquals(Some(v1), v2)
   }
+
+  @Test def emptyTypedIteratorsShouldBeEqual(): Unit = {
+    val emptyDoubleIterator = Iterator.empty[Double]
+    val emptyIntIterator = Iterator.empty[Int]
+    assertEquals(emptyDoubleIterator, emptyIntIterator)
```
mkeskells: should be assertSame ( asserts that they are eq)

szeiger:
Any collection changes targeting M4 should go into https://github.com/scala/collection-strawman instead. The code there is the same so this should be easy to port but you may consider going further by making `object Iterator` extend `IterableFactory` (if that's possible).

src/library/scala/collection/Iterator.scala
szeiger: These no-runtime-op no-need-to-override methods are all marked as `@inline final` in the strawman. The same should be done for this method.

kelebra: Thanks for suggestion, method now marked as `@inline final`.

kelebra:
@szeiger Got it, will port this change for starters. Should I close this PR then?

Update: transferred fix and added suggested extension in [strawman PR](https://github.com/scala/collection-strawman/pull/450).

src/library/scala/collection/Iterator.scala
mkeskells: That is the solution to a different problem unfortunately
see https://github.com/rorygraves/scalac_perf/issues/46
so this should probably be 
`@inline final def empty[T]: Iterator[T] = _empty: @inline`

test/junit/scala/collection/IteratorTest.scala
kelebra: Fixed, thanks for pointing that out.

kelebra: Thanks, fixed.

src/library/scala/collection/Iterator.scala
kelebra: Thanks a lot for suggestion but it looks like current implementation also gets inlined. This is what I see in the bytecode of client code that uses it:

```
public final class scala.collection.IteratorInlineTest {
  public static scala.collection.Iterator<java.lang.Object> inlineTest2$default$1();
    Code:
       0: getstatic     #16                 // Field scala/collection/IteratorInlineTest$.MODULE$:Lscala/collection/IteratorInlineTest$;
       3: invokevirtual #18                 // Method scala/collection/IteratorInlineTest$.inlineTest2$default$1:()Lscala/collection/Iterator;
       6: areturn

  public static scala.collection.Iterator<java.lang.Object> inlineTest3();
    Code:
       0: getstatic     #16                 // Field scala/collection/IteratorInlineTest$.MODULE$:Lscala/collection/IteratorInlineTest$;
       3: invokevirtual #21                 // Method scala/collection/IteratorInlineTest$.inlineTest3:()Lscala/collection/Iterator;
       6: areturn

  public static scala.collection.Iterator<java.lang.Object> inlineTest2(scala.collection.Iterator<java.lang.Object>);
    Code:
       0: getstatic     #16                 // Field scala/collection/IteratorInlineTest$.MODULE$:Lscala/collection/IteratorInlineTest$;
       3: aload_0
       4: invokevirtual #26                 // Method scala/collection/IteratorInlineTest$.inlineTest2:(Lscala/collection/Iterator;)Lscala/collection/Iterator;
       7: areturn

  public static scala.collection.immutable.List<java.lang.Object> inlineTest1();
    Code:
       0: getstatic     #16                 // Field scala/collection/IteratorInlineTest$.MODULE$:Lscala/collection/IteratorInlineTest$;
       3: invokevirtual #30                 // Method scala/collection/IteratorInlineTest$.inlineTest1:()Lscala/collection/immutable/List;
       6: areturn
}
```

As we can see we __get static field value__ every time (no __invokestatic__) so it looks fine.

PR #6322, opened 2018-02-13
Deeply focus the qualifier before using it as the RHS of the stabilizing val.

hrhino:
...because rangepos expects offset-positioned trees not to contain range-positioned trees. This is the same approach that default arguments use when moving the RHS of the parameter val into a default getter.

Alternatives:
- Remove that assertion: I'm not sure what it's for, but the validation seems quite insistent on that point. I suspect that the IDE or PC or something using positions may break on that.
- Use the tree duplicator (and don't make a focuser traverser). This is how the default args are currently done, but since we're not reusing `qual` here that I can see it doesn't seem too useful to waste the allocations that it would take to duplicate that tree.

The test cases are just slapping `-Yrangepos` on all of the test cases introduced in bcbe993e (where this feature was introduced).

`test/files/pos/t4225.scala` is not thus enhanced, since there's a similar rangeposly error with by-name arguments of right-associative methods.

Fixes scala/bug#10706, references scala/scala-dev#472.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -5038,14 +5038,14 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
                 isInsertionNode(ctx.tree) && !isInsertionNode(ctx.outer.tree)
               }
 
-              if(insertionContext != NoContext) {
+              if (insertionContext != NoContext) {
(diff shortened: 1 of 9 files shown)


hrhino:
/sync

(this didn't report back on a successful build immediately, but /sync seemed to work)

lrytz (approved):

lrytz:
Thank you!

lrytz:
Filed https://github.com/scala/bug/issues/10751 for the right-associative by-name combo.

hrhino:
Thank you!

PR #2342, opened 2013-04-02
SI-7292 Deprecate octal escape literals

soc:

PR #6324, opened 2018-02-14
Remove octal escape literals

eed3si9n:
Ref #2342
Ref scala/bug#7292

This removes octal escape in String literals, which was deprecated in 2.11 and removed under -Xfuture. As part of the effort to move away from -Xfuture flag, I am removing it here.


som-snytt:
Did the workflow change? I couldn't leave a one-line comment by clicking the line.

My comment was: Should error messages use normal capitalization?

test/files/neg/t7292-removal.check
```diff
@@ -1,10 +1,10 @@
-t7292-removal.scala:2: error: Octal escape literals are unsupported, use \u0000 instead.
+t7292-removal.scala:2: error: octal escape literals are unsupported. use \u0000 instead.
   val chr1 = '\0'
               ^
-t7292-removal.scala:3: error: Octal escape literals are unsupported, use \u0053 instead.
+t7292-removal.scala:3: error: octal escape literals are unsupported. use \u0053 instead.
   val str1 = "abc\123456"
                  ^
-t7292-removal.scala:4: error: Octal escape literals are unsupported, use \n instead.
+t7292-removal.scala:4: error: octal escape literals are unsupported. use \n instead.
```
dwijnand: Should error messages use normal capitalisation?

Ichoran: There are several things which are correct, but neither the old version (independent clauses joined by a comma) nor the new version (two lowercase sentences following a colon) follow standard colon usage and/or grammar rules.  These are all arguably correct:

```
octal escape literals are unsupported; use \n instead.
Octal escape literals are unsupported; use \n instead.
Octal escape literals are unsupported.  Use \n instead.
```

SethTisue:
When your PR is ready to merge but then A.P. raises a capitalization issue, that's "Som's nit".

test/files/neg/t7292-removal.check
som-snytt: I didn't mind the comma splice, it's kinda hipsterish.

som-snytt: But it's capitalization like civilization.

eed3si9n: I am used to seeing all lowercase error messages from scalac. Here are some samplings from `*.check` files:

```
abstract-class-2.scala:11: error: object creation impossible, since method f in trait S2 of type (x: P2.this.p.S1)Int is not defined

S.scala:1: error: class S needs to be abstract, since method g in class J of type (y: Int, z: java.util.List)Int is not defined

classmanifests_new_deprecations.scala:2: warning: type ClassManifest in object Predef is deprecated (since 2.10.0): use `scala.reflect.ClassTag` instead

divergent-implicit.scala:5: error: diverging implicit expansion for type Int => String

ambiguous-same.scala:13: error: reference to x is ambiguous;
it is both defined in object X and imported subsequently by
import X.x

serialversionuid-not-const.scala:1: error: annotation argument needs to be a constant; found: 13L.toLong
```

There are certainly capitalized error messages, but the classic ones are mostly all lowercase.

As per punctuations are concerned, could I interest you in colon and no-trailing-period?

```
error: octal escape literals are unsupported: use \n instead
```

This `$clause: use X instead` is the same structure as the deprecation warnings.

som-snytt: I like the explanatory colon. It coordinates in both the grammar and fashion sense.

src/library/scala/StringContext.scala
```diff
@@ -219,20 +218,7 @@ object StringContext {
...
-                oct = oct * 8 + str(idx) - '0'
-                idx += 1
-                if (idx < len && leadch <= '3' && '0' <= str(idx) && str(idx) <= '7') {
-                  oct = oct * 8 + str(idx) - '0'
-                  idx += 1
-                }
-              }
-              idx -= 1   // retreat
-              oct.toChar
+              throw new InvalidEscapeException(str, next)
```
som-snytt: This case is now extraneous, since the default `case _` is identical.

Yes, I read the github doc to realize I have to hit the blue + to comment. Did I always just do that without thinking?

src/library/scala/StringContext.scala
```diff
@@ -187,17 +187,16 @@ object StringContext {
...
    *
    *  @param  str  A string that may contain escape sequences
    *  @return The string with all escape sequences expanded.
    */
-  def treatEscapes(str: String): String = treatEscapes0(str, strict = false)
+  def treatEscapes(str: String): String = treatEscapes0(str)
 
   /** Treats escapes, but disallows octal escape sequences. */
-  def processEscapes(str: String): String = treatEscapes0(str, strict = true)
+  def processEscapes(str: String): String = treatEscapes0(str)
```
som-snytt: We can deprecate one of the methods. I didn't like the name "treatEscapes", as it was not "computer sciencey" enough. There is a supplementary issue that there's no API for adding escapes.

som-snytt:
I took the plunge and started a review instead of drive-by line comments.

Maybe defer the deprecation, otherwise delete that extra case and squash and it looks great.

Answering my previous question, you have to hit the blue + icon. Previously I must have always done that as I have been trained to do since birth.

Also worth adding, it makes you miss soc, whose ideal was Some Other Compiler. The plaque erected in his honor in Lausanne reads, "First to deprecate, first to defect," echoing the old line about Washington, "First in war, first in peace." I almost wrote "defecate" by accident.

eed3si9n:
> Maybe defer the deprecation, 

I want to clarify here. Are you suggesting we continue the deprecation into 2.13, and remove only under `-Xsource:2.14`? I am ok with that. @adriaanm has mentioned that 2.13 is a library release, so he might agree as well.

>  otherwise delete that extra case and squash and it looks great.

Not sure what you mean by "delete that extra case."

> "First to deprecate, first to defect,"

There's something fun and rewarding about gleaning the crops from where the framers left off. A good entry point for first-time contributors like me.

SethTisue:
I think octal escape literals are so obscure a feature, and have been deprecated so long, that I see no need to go the slow route.

Let's see what the community build says before merging, though.

som-snytt:
Sorry, I was highlighting my comment that `treatEscapes` and `processEscapes` are now redundant and one could be deprecated, but maybe we can defer that to another PR. Nothing to do with deprecating octal literals per se.

The extra case is per my other line comment on `StringContext`.

I guess I won't comment on line comments in future.

eed3si9n:
> Sorry, I was highlighting my comment that treatEscapes and processEscapes are now redundant and one could be deprecated, but maybe we can defer that to another PR. Nothing to do with deprecating octal literals per se.

ok, that totally makes sense now. I no idea where the original line comments went.

som-snytt (changes requested):

som-snytt:
Oh wait, I just realized d'oh that I have to hit submit for a review that is not line-comment-only. 

My first time trying github's heavy-weight review process. Sigh.

Probably I prefer light-weight. "You will submit!"

src/library/scala/StringContext.scala
```diff
@@ -187,17 +187,22 @@ object StringContext {
...
    *   control: `\b`, `\t`, `\n`, `\f`, `\r`
    *   escape:  `\\`, `\"`, `\'`
-   *   octal:   `\d` `\dd` `\ddd` where `d` is an octal digit between `0` and `7`.
    *
    *  @param  str  A string that may contain escape sequences
    *  @return The string with all escape sequences expanded.
    */
-  def treatEscapes(str: String): String = treatEscapes0(str, strict = false)
+  @deprecated("use processEscapes", "2.13.0")
+  def treatEscapes(str: String): String = processEscapes(str)
```
eed3si9n, line 195: This is now deprecated.

src/library/scala/StringContext.scala
```diff
@@ -218,21 +223,6 @@ object StringContext {
...
-              if (idx < len && '0' <= str(idx) && str(idx) <= '7') {
-                oct = oct * 8 + str(idx) - '0'
-                idx += 1
-                if (idx < len && leadch <= '3' && '0' <= str(idx) && str(idx) <= '7') {
-                  oct = oct * 8 + str(idx) - '0'
-                  idx += 1
-                }
-              }
-              idx -= 1   // retreat
-              oct.toChar
```
eed3si9n, line 235: This entire `case` no longer hits, so it's removed.

som-snytt (approved):

som-snytt:
LGTM!

Do they still say that for reviews? Thanks for your patience with my newbie reviewing.

eed3si9n:
LGTM is still a thing. Thanks for the review :)

SethTisue:
> LGTM is still a thing

not for long 😈 https://github.com/scala/scabot/pull/74

eed3si9n:
I had no idea there was a whole process you folks associated to it.

som-snytt:
When I see Seth removed the reviewed label and then Jenkins added it back, I imagine them as "Rock'em Sock'em Robots". With Jenkins as Blue Bomber and Seth as Red Rocker, of course.

SethTisue:
community build queued: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/958/ (404 til Jenkins gets in gear)

SethTisue:
community build run: https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/965/

```
[info] >>> The dbuild result is------------: FAILED (failed: scalatest, scala-async)
```

the scala-async thing seems to be pickiness about Scala version numbers, rather than anything to do with the PR

as for ScalaTest, they are actually using octal escapes. it's easy enough to fix, and once ScalaTest is passing other projects will be able to run (about half of the whole build), so I'll submit a ScalaTest PR.

som-snytt:
It's called octal not only because base 8 but because only octogenarians use them.

SethTisue:
I think you and I together make an octogenarian, maybe with enough left over to make a surly teenager with.

SethTisue:
followup run queued:  https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/969/ ~~(404 til Jenkins returns from his smoke break)~~

SethTisue:
my ScalaTest PR was merged and now the only failure is the scala-async one, unrelated to Eugene's changes. so I think this is ready for merge now

SethTisue (approved):

eed3si9n:
Thanks for the merge. Good to see these moving!

PR #3076, opened 2013-10-23
SI-7605 Deprecate procedure syntax

soc:
This commit covers three cases:
- constructor definitions      (def this {...})
- concrete method definitions  (def foo {...})
- abstract method declarations (def foo)

The deprecation is currently hidden behind -Xfuture pending IDE support
for migrating users from procedures to methods.

PR #5647, opened 2017-01-17
SI-10130 Ignore extra parse warnings

som-snytt:
Don't emit warnings when parsing to test for
complete input. The warnings will show up
in the templated code.

```
$ scala
Welcome to Scala 2.12.1 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_111).
Type in expressions for evaluation. Or try :help.

scala> '\060'
<console>:1: warning: Octal escape literals are deprecated, use \u0030 instead.
'\060'
 ^
<console>:2: warning: Octal escape literals are deprecated, use \u0030 instead.
'\060'
 ^
<console>:12: warning: Octal escape literals are deprecated, use \u0030 instead.
       '\060'
        ^
res0: Char = 0

scala> val then = 0
<console>:1: warning: then is a reserved word (since 2.10.0); usage as an identifier is deprecated
val then = 0
    ^
<console>:11: warning: then is a reserved word (since 2.10.0); usage as an identifier is deprecated
       val then = 0
           ^
then: Int = 0

scala> 'abc'
<console>:1: error: unclosed character literal
'abc'
    ^

scala> :quit
apm@mara:~/projects/scala$ ./build/pack/bin/scala -deprecation
Welcome to Scala 2.12.2-20170110-140852-827d69d (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_111).
Type in expressions for evaluation. Or try :help.

scala> '\060'
<console>:12: warning: Octal escape literals are deprecated, use \u0030 instead.
       '\060'
        ^
res0: Char = 0

scala> val then = 0
<console>:11: warning: then is a reserved word (since 2.10.0); usage as an identifier is deprecated
       val then = 0
           ^
then: Int = 0

scala> 'abc'
<console>:1: error: unclosed character literal
'abc'
    ^

scala> 
```

PR #6325, opened 2018-02-14
Deprecate procedure syntax without -Xfuture

eed3si9n:
Ref https://github.com/scala/bug/issues/7605

https://github.com/scala/scala/pull/3076 deprecated the procedure syntax, but only under -Xfuture flag. This deprecates it without it, and drops it under `-Xsource:2.14`.

This PR also migrates all code under `src/compiler/` and `test/` to `: Unit =`-style, mostly automatically by using ScalaFix's ProcedureSyntax (and ExplicitUnit, which will be merged to ProcedureSyntax) rule.


SethTisue:
you're on a roll!

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -2718,18 +2718,20 @@ self =>
...
+              if (settings.isScala214) syntaxError(in.lastOffset, msg("unsupprted"))
+              else deprecationWarning(in.lastOffset, msg("deprecated"), "2.13.0")
               restype = scalaUnitConstr
             }
             newmods |= Flags.DEFERRED
             EmptyTree
           } else if (restype.isEmpty && in.token == LBRACE) {
-            if (settings.future)
-              deprecationWarning(in.offset, s"Procedure syntax is deprecated. Convert procedure `$name` to method by adding `: Unit =`.", "2.12.0")
+            if (settings.isScala214) syntaxError(in.lastOffset, msg("unsupported"))
```
dwijnand: defmsg

eed3si9n: Nice catch.

SethTisue:
to facilitate review, let's not squash until we're totally ready to merge this. (the full, combined diff is........ l-o-n-g.......)

@eed3si9n you didn't write it, but I think the "convert procedure to method" wording in the error message is... not ideal. "procedures" are still methods, they aren't some entirely separate category. also, "add `: Unit`" isn't precisely right, it should be "add `: Unit =`" (with the equals sign). after all it's the equals sign that's 100% necessary, the `: Unit` is actually optional in many cases since it would be inferred anyway. (I'm comfortable with always recommending it regardless.) 

how about replacing "convert..." with simply "instead, add `: Unit =` to method `$name`"

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -2718,18 +2718,20 @@ self =>
         val vparamss = paramClauses(name, contextBoundBuf.toList, ofCaseClass = false)
         newLineOptWhenFollowedBy(LBRACE)
         var restype = fromWithinReturnType(typedOpt())
+        def msg(what: String) = s"procedure syntax is $what. convert procedure `$name` to method by adding `: Unit`."
+        def defmsg(what: String) = s"procedure syntax is $what. convert procedure `$name` to method by adding `: Unit =`."
```
eed3si9n: @SethTisue `=` I think is in the error message when it is a definition. Do you have a negative test link that shows otherwise?

SethTisue: sorry, I missed that, nm.

eed3si9n:
@SethTisue 

### before

```
proc.scala:4: warning: procedure syntax is deprecated: convert procedure `boo` to method by adding `: Unit`
  def boo(i: Int, l: Long)
                          ^
proc.scala:5: warning: procedure syntax is deprecated: convert procedure `boz` to method by adding `: Unit =`
  def boz(i: Int, l: Long) {}
                           ^
```

### after

```
proc.scala:4: warning: procedure syntax is deprecated: instead, add `: Unit` to `boo`
  def boo(i: Int, l: Long)
                          ^
proc.scala:5: warning: procedure syntax is deprecated: instead, add `: Unit =` to `boz`
  def boz(i: Int, l: Long) {}
                           ^
```

With either of these messages, is it be clear for the users to know what to do?

Problem 1-B: Suppose the user, who is otherwise proficient in Scala, has never heard of the term "procedure syntax."

SethTisue:
> Suppose the user, who is otherwise proficient in Scala, has never heard of the term "procedure syntax."

I don't mean to be flip, but: they can Google it?

I don't see a way to do better without writing a substantially longer error message. (doing that wouldn't be a bad idea, but I think the PR is mergeable regardless.)

maybe "procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare`boz`'s return type" ?

eed3si9n:
+1 on "procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `boz`'s return type".

eed3si9n:
Ping?

SethTisue:
It would be good if we expedited this a bit in order for Eugene not to have to do a bunch of annoying rebases. I’ll try to take a final look this week.

adriaanm:
Did some rebasing and squashing to compensate for my tardiness. Hope you won't mind my `push -f`.

src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
```diff
@@ -2727,18 +2727,20 @@ self =>
...
+              if (settings.isScala214) syntaxError(in.lastOffset, msg("unsupported", ": Unit"))
+              else deprecationWarning(in.lastOffset, msg("deprecated", ": Unit"), "2.13.0")
               restype = scalaUnitConstr
             }
             newmods |= Flags.DEFERRED
             EmptyTree
           } else if (restype.isEmpty && in.token == LBRACE) {
-            if (settings.future)
-              deprecationWarning(in.offset, s"Procedure syntax is deprecated. Convert procedure `$name` to method by adding `: Unit =`.", "2.12.0")
+            if (settings.isScala214) syntaxError(in.lastOffset, msg("unsupported", ": Unit ="))
```
adriaanm: The error should probably be positioned at `in.offset`. I'll push a fix once I've updated the check files

adriaanm:
Just realized the last two commits will need another rebase since we'll merge the new collections first.  (One line in src/compiler/scala/tools/nsc/ast/parser/xml/MarkupParserCommon.scala and 29 tests.)

SethTisue:
@eed3si9n sorry, could we ask you for one more rebase, then we'll merge?

SethTisue:
(P.S. if you're worried that our current merge promises are no more trustworthy than previous ones :-/, you might consider splitting this into two PRs, one big one that just removes uses of procedure syntax codebase-wide — since we could easily merge that the moment Jenkins liked it, there's no downside — and then a second, small PR that does the deprecation.)

eed3si9n:
> (P.S. if you're worried that our current merge promises are no more trustworthy than previous ones :-/, you might consider splitting this into two PRs, one big one that just removes uses of procedure syntax codebase-wide — since we could easily merge that the moment Jenkins liked it, there's no downside — and then a second, small PR that does the deprecation.)

Given that the collection library is merged, I am hoping that there's no more blockers to this PR once all the tests pass.

SethTisue:
go Jenkins go!

SethTisue:
tests need updating. even the first commit fails, probably because the collections merge introduced new tests that are now causing deprecation warnings.

eed3si9n:
First and third commits probably need to get squashed together since deprecation affects many of the tests.

SethTisue:
@eed3si9n wouldn't it minimize churn if you first got rid of all the uses, _then_ deprecated? if done in that order, the deprecation commit would be small.

eed3si9n:
I don't think the ordering changes things as long as it affects a bunch of files, and if it stays open for months, assuming that deprecation itself is not blocked/controversial.

eed3si9n:
I originally had it `[ci: last-only]` so I can keep appending more commits to update the tests. Once they are passing, I can squash them all to get greens.

SethTisue:
okay. regardless, I'm standing by to hit "merge" as soon as Jenkins likes it.

eed3si9n:
@SethTisue Rebased and squashed.

SethTisue:
your revenge is that now everybody else has to rebase _their_ PRs 😁

SethTisue:
thank you, Eugene

eed3si9n:
Thanks for the merge!

SethTisue:
this came up at the SIP meeting today. I get a deprecation warning with the M5 compiler:

```
% cat > S.scala
object S { def foo() { } }
% /usr/local/scala/scala-2.13.0-M5/bin/scalac -deprecation -feature S.scala
S.scala:1: warning: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `foo`'s return type
```

but in the REPL, I don't:

```
% /usr/local/scala/scala-2.13.0-M5/bin/scala -deprecation        
Welcome to Scala 2.13.0-M5 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_192).

scala> def foo() { }
foo: ()Unit

scala> object S { def foo() { } }
defined object S

scala> (new Thread).stop()
                    ^
       warning: method stop in class Thread is deprecated: see corresponding Javadoc for more information.
```

SethTisue:
@som-snytt @adriaanm any clue why this might be happening, where an investigation should start?

som-snytt:
From now on, "procedure syntax" will refer only to the order and form of the words you must use at a SIP meeting to get your point across.

SethTisue:
I'll take that as a "no" :-)

som-snytt:
I'll take a closer look. Deprecation works in general. So it's not a lost flag.

BTW `-deprecation -feature` are on by default.

Also `-Xsource:2.14` correctly errors. Luckily, there is pizza in the break room.

Looks like just the parser deprecations aren't forwarded.

```
scala> for (i <- List(1); val j = i + 1) yield j
res0: List[Int] = List(2)

scala> :replay -Xsource:2.14
Replaying: for (i <- List(1); val j = i + 1) yield j
                                ^
       error: `val` keyword in for comprehension is unsupported: instead, bind the value without `val`
```

eed3si9n:
https://github.com/scala/scala/blob/v2.13.0-M5/src/repl/scala/tools/nsc/interpreter/IMain.scala#L1082-L1085
```scala
  def parse(line: String): Either[Result, (List[Tree], Position)] = {
    var isIncomplete = false
    currentRun.parsing.withIncompleteHandler((_, _) => isIncomplete = true) {
      withoutWarnings {
```

REPL parsing code is silencing all warnings.

eed3si9n:
This was added in https://github.com/scala/scala/pull/5647 to fix https://github.com/scala/bug/issues/10130 about multiple warnings showing up.

som-snytt:
It parses again after templating, IIRC. Edit: I see that's my comment on the PR you point out.

I thought maybe I broke something in touching the reporters.

eed3si9n:
On 2.13.x, it seems to work without `withoutWarnings`:

```
> scala -deprecation
....
scala> '\060'
        ^
       error: octal escape literals are unsupported: use \u0030 instead

scala> def foo() { }
                 ^
       warning: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `foo`'s return type
foo: ()Unit
```

som-snytt:
OK, good!

I see, Adriaan keeps the trees and splices it into the template.

PR #6326, opened 2018-02-14
threadsafe simple stats

mkeskells:


lrytz:
Looks good. @jvican what do you think?

The original ticket mentions races around `pushTimer` / `popTimer`. Even though timer stacks are not made thread safe in this PR, the problem should no longer exist since there's a Statistics instance per global, and stacks are only used in single-threaded parts of the compiler. So things should work in sbt / IDEs running multiple compiler threads.

SethTisue:
@mkeskells is this still WIP?

mkeskells:
@SethTisue I have removed WIP.
I marked it as WIP to promote the discussion, and due to my lack of understanding on the use cases of the different timer types.
I made the changes as the back end phases allow MT access, and I have several ideas that would allow MT  acces to be used elsewhere in the compile tool chain. 
I believe that it is essential to support threadsafe statistics and would be happy to see this merged

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -177,22 +179,37 @@ quant)
...
-    var timings = 0
-    def start() = {
-      (nanos, System.nanoTime())
+    private val threadNanos = new ThreadLocal[LongRef] {
+      override def initialValue() = {
+        totalThreads.incrementAndGet()
+        new LongRef(0)
+      }
+    }
+    def nanos = totalNanos.get
```
jvican: Nit: `def` should follow `val` to avoid future NPEs when hacking around this area.

src/reflect/scala/reflect/internal/util/Statistics.scala
```diff
@@ -2,10 +2,12 @@ package scala
 package reflect.internal.util
 
 import scala.collection.mutable
-
 import scala.reflect.internal.SymbolTable
 import scala.reflect.internal.settings.MutableSettings
-import java.lang.invoke.{SwitchPoint, MethodHandle, MethodHandles, MethodType}
+import java.lang.invoke.{MethodHandle, MethodHandles, MethodType, SwitchPoint}
```
jvican: This line did not change.

jvican (commented):
LGTM! Just a few minor comments.

PR #6327, opened 2018-02-15
Fix off by one error regression under -Ylog

retronym:


src/reflect/scala/reflect/internal/SymbolTable.scala
```diff
@@ -182,8 +182,7 @@ abstract class SymbolTable extends macros.Universe
   type RunId = Int
   final val NoRunId = 0
 
-  // sigh, this has to be public or enteringPhase doesn't inline.
```
retronym, line 185: We don't need this to be private anymore as `enteringPhase` doesn't directly access this field.

PR #6328, opened 2018-02-15
add Either.left and Either.right typed as Either

jozic:
motivation: `Left` and `Right` constructors are typed as `Left` and `Right` respectively, but sometimes we want something like `Option.empty` which returns `None` but typed as `Option`.

The test represents a code that would not compile if plain `Left`/`Right` constructors are used 

@SethTisue please take a look

### DIFF
--- a/src/library/scala/util/Either.scala
+++ b/src/library/scala/util/Either.scala
@@ -457,6 +457,16 @@ object Either {
   def cond[A, B](test: Boolean, right: => B, left: => A): Either[A, B] =
     if (test) Right(right) else Left(left)
 
+  /**
+   * Creates `Left` typed as `Either`
+   */
+  def left[A, B](value: A): Either[A, B] = Left(value)
+
+  /**
+   * Creates `Right` typed as `Either`
+   */
+  def right[A, B](value: B): Either[A, B] = Right(value)
+
   /** Allows use of a `merge` method to extract values from Either instances
    *  regardless of whether they are Left or Right.
    *
--- a/test/junit/scala/util/EitherTest.scala
+++ b/test/junit/scala/util/EitherTest.scala
@@ -1,0 +1,36 @@
+package scala.util
+
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.junit.Test
+import org.junit.Assert._
+
+@RunWith(classOf[JUnit4])
+class EitherTest {
+
+  @Test
+  def testLeft: Unit = {
+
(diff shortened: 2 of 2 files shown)


SethTisue:
LGTM in a general way, but I haven't worked in a codebase that makes especially heavy use of `Either`. I asked for feedback on gitter (https://gitter.im/typelevel/general?at=5a85d126ce68c3bc748f61d4)

SethTisue:
what's the value over just doing simple type ascription? e.g. `Left(3): Either[Int, String]`

jozic:
the value is that `no type ascription is needed` :) same as `Option.empty[Int]` vs `None: Option[Int]`, 
code (subjectively) looks a little bit less cluttered without required type ascription
otherwise it's same

SethTisue:
@tpolecat points out that:

> having to repeat the inferable type when you specify the non-inferable one is kind of a grunt

have you considered ways of avoiding that...?

jozic:
the only way I know is already mentioned in gitter, like 
```scala
implicit class LeftOps[A](a: A) extends AnyVal { def left[B]: Either[A,B] = Left[A,B](a) }
```
but I share your concern on selling it, though that would be awesome i think )

LukaJCB:
To get a precise type with something like this (where `Nothing` never gets inferred), we could use the partiallyApplied trick, so something like this:

```scala
object Either {
  def left[A]: FromLeftPartiallyApplied[A] = 
    new FromLeftPartiallyApplied[A]

  def right[A]: FromRightPartiallyApplied[A] = 
    new FromRightPartiallyApplied[A]
}

class FromLeftPartiallyApplied[A](val dummy: Boolean = true) 
  extends AnyVal {
    def apply[B](b: B): Either[B, A] = Left(b)
  }

class FromRightPartiallyApplied[A](val dummy: Boolean = true) 
  extends AnyVal {
    def apply[B](b: B): Either [A, B] = Right(b)
  }
```

This means we can now use it like this: `Either.right[String](42)` and return a fully formed `Either[String, Int]` :)

This technique is described here: https://typelevel.org/cats/guidelines.html#partially-applied-type-params

SethTisue:
@jozic has now also submitted a version of @LukaJCB's suggestion for consideration over at #6329

SethTisue:
closing this one since no one has objected to #6329

SethTisue:
> having to repeat the inferable type when you specify the non-inferable one is kind of a grunt

even if #6329 isn't accepted, I would suggest we not accept this one either on this grounds, actually. but I think #6329 has a good chance

PR #6329, opened 2018-02-15
add Either.left and Either.right typed as Either

jozic:
this is another implementation of #6328 inspired by @LukaJCB and others in gitter(https://gitter.im/typelevel/general?at=5a85d126ce68c3bc748f61d4) (thanks!)

motivation: `Left` and `Right` constructors are typed as `Left` and `Right` respectively, but sometimes we want something like `Option.empty` which returns `None` but typed as `Option`.

The test represents a code that would not compile if plain `Left`/`Right` constructors are used 

@SethTisue please check it again, thanks!
If this looks better (it does to me) then #6328 should be closed

### DIFF
--- a/src/library/scala/util/Either.scala
+++ b/src/library/scala/util/Either.scala
@@ -468,6 +468,35 @@ object Either {
   def cond[A, B](test: Boolean, right: => B, left: => A): Either[A, B] =
     if (test) Right(right) else Left(left)
 
+  /**
+   * Creates `FromLeftPartiallyApplied[B]` that given an argument of type `A`
+   * produces a `Left` typed as `Either[A,B]`
+   * {{
+   *   val l = Either.left[Int]("boo")
+   *   //l: Either[String, Int] = Left("boo")
+   * }}
+   */
+  def left[B]: FromLeftPartiallyApplied[B] = new FromLeftPartiallyApplied[B]
+
+  private[util] final class FromLeftPartiallyApplied[B](private val dummy: Unit) extends AnyVal {
+    def apply[A](value: A): Either[A, B] = Left(value)
+  }
+
+  /**
+   * Creates `FromRightPartiallyApplied[A]` that given an argument of type `B`
+   * produces a `Right` typed as `Either[A,B]`
+   * {{
+   *   val r = Either.right[String](5)
+   *   //r: Either[String, Int] = Right(5)
+   * }}
+   */
+  def right[A]: FromRightPartiallyApplied[A] = new FromRightPartiallyApplied[A]
+
+  private[util] final class FromRightPartiallyApplied[A](private val dummy: Unit) extends AnyVal {
+    def apply[B](value: B): Either[A, B] = Right(value)
+  }
+
+
   /** Allows use of a `merge` method to extract values from Either instances
    *  regardless of whether they are Left or Right.
    *
--- a/test/junit/scala/util/EitherTest.scala
+++ b/test/junit/scala/util/EitherTest.scala
@@ -1,6 +1,7 @@
 package scala.util
 
 import org.junit.{ Assert, Test }
+import org.junit.Assert._
 
 class EitherTest {
 
@@ -13,9 +14,34 @@ class EitherTest {
     val flatrl: Either[String, Int] = rl.flatten
     val flatrr: Either[String, Int] = rr.flatten
 
-    Assert.assertEquals(Left("pancake"), flatl)
-    Assert.assertEquals(Left("flounder"), flatrl)
-    Assert.assertEquals(Right(7), flatrr)
-    
+    assertEquals(Left("pancake"), flatl)
+    assertEquals(Left("flounder"), flatrl)
+    assertEquals(Right(7), flatrr)
+  }
+
+  @Test
+  def testLeft: Unit = {
+
+    def rightSumOrLeftEmpty(l: List[Int]) =
+      l.foldLeft(Either.left[Int]("empty")) {
+        case (Left(_), i) => Right(i)
+        case (Right(s), i) => Right(s + i)
+      }
+
+    assertEquals(rightSumOrLeftEmpty(List(1, 2, 3)), Right(6))
+    assertEquals(rightSumOrLeftEmpty(Nil), Left("empty"))
+  }
+
+  @Test
+  def testRight: Unit = {
+
+    def leftSumOrRightEmpty(l: List[Int]) =
+      l.foldLeft(Either.right[Int]("empty")) {
+        case (Right(_), i) => Left(i)
+        case (Left(s), i) => Left(s + i)
+      }
+
+    assertEquals(leftSumOrRightEmpty(List(1, 2, 3)), Left(6))
+    assertEquals(leftSumOrRightEmpty(Nil), Right("empty"))
   }
 }


src/library/scala/util/Either.scala
```diff
@@ -457,6 +457,35 @@ object Either {
...
+   * Creates `FromLeftPartiallyApplied[B]` that given an argument of type `A`
+   * produces a `Left` typed as `Either[A,B]`
+   * {{
+   *   val l = Either.left[Int]("boo")
+   *   //l: Either[String, Int] = Left("boo")
+   * }}
+   */
+  def left[B]: FromLeftPartiallyApplied[B] = new FromLeftPartiallyApplied[B]
+
+  private[util] final class FromLeftPartiallyApplied[B](private val dummy: Boolean = true) extends AnyVal {
```
LukaJCB: I think these `dummy`s can now be `Unit` since 2.12.3 :)

src/library/scala/util/Either.scala
```diff
@@ -457,6 +457,35 @@ object Either {
...
 
+  /**
+   * Creates `FromLeftPartiallyApplied[B]` that given an argument of type `A`
+   * produces a `Left` typed as `Either[A,B]`
+   * {{
+   *   val l = Either.left[Int]("boo")
+   *   //l: Either[String, Int] = Left("boo")
+   * }}
+   */
+  def left[B]: FromLeftPartiallyApplied[B] = new FromLeftPartiallyApplied[B]
```
NthPortal: Why not just have it use `Left`, and widen the return type?

i.e.

```
def left[A, B](value: A): Either[A, B] = Left(value)
```

src/library/scala/util/Either.scala
xuwei-k: https://github.com/scala/scala/commit/34da49bc31b4ba15d385d3c6112808255d2de2ef

src/library/scala/util/Either.scala
```diff
@@ -457,6 +457,35 @@ object Either {
...
+
+  private[util] final class FromLeftPartiallyApplied[B](private val dummy: Unit) extends AnyVal {
```
SethTisue, line 481: @why `private[util]` here (and 14 lines down), why not fully `private`? (Guess: is it because otherwise you get a warning about the type escaping...?)

src/library/scala/util/Either.scala
```diff
@@ -457,6 +457,35 @@ object Either {
   def cond[A, B](test: Boolean, right: => B, left: => A): Either[A, B] =
     if (test) Right(right) else Left(left)
 
+  /**
+   * Creates `FromLeftPartiallyApplied[B]` that given an argument of type `A`
```
SethTisue, line 472: Since this name appears in the Scaladoc, perhaps a shorter name could be found?

src/library/scala/util/Either.scala
SethTisue: are there any pros/cons to using `Unit` vs `Boolean` in this pattern?

src/library/scala/util/Either.scala
SethTisue, line 479: It's my first time seeing this pattern, so it isn't obvious to me (and might not be obvious to other reviewers) what the downsides of this pattern are. I mean, obviously it's a downside that this extra type is involved (and shows up in the Scaladoc). Are there any concealed footguns here?

SethTisue:
why doesn't Cats use this pattern, why does it have the #6328 version instead?

core/src/main/scala/cats/syntax/either.scala:

```
  def left[A, B](a: A): Either[A, B] = Left(a)
  def right[A, B](b: B): Either[A, B] = Right(b)
```

if this is merged, is it going to be make difficulties for Cats? (offhand, I guess you'd need separate `scala-2.12` and `scala-2.13` source directories to handle it...? not sure I've thought it through properly)

LukaJCB:
That's a good question, though I think most cats users use the more convenient `a.asLeft[A]`. We could probably deprecate those in time for 2.13, I think.

src/library/scala/util/Either.scala
jozic: Do you have any good ideas? I think the name is long, but very descriptive

src/library/scala/util/Either.scala
jozic: that was my first time as well, but it looks safe enough, except the fact you mentioned - it appears in public
like you can do `Either.left[Int]` and get instance of `FromLeftPartiallyApplied[Int]`

src/library/scala/util/Either.scala
jozic: I'm pretty sure I've tried to make it fully `private`, but compiler wasn't happy. I will double check it

jozic: As for Unit, i guess it's just less to type, no need to give it default value. In any case it's just a trick for `AnyVal`. 
Maybe someone has a better idea here, as I don't see much difference otherwise

jozic:
I guess @LukaJCB is right, cats is full of implicits and `3.asLeft[String]` is a no-brainer in Catsland, where standard lib doesn't have much of those patterns. So I'm assuming extensions like `asLeft` have fewer chances to be merged in std lib. Please correct me if i'm wrong here

@LukaJCB you meant deprecate `def left[A, B](a: A): Either[A, B] = Left(a)` and not `asLeft`, right? :)

SethTisue:
> extensions like asLeft have fewer chances to be merged in std lib

yeah I'm not really comfortable with putting extension methods on `Any`

src/library/scala/util/Either.scala
SethTisue: I think I saw somewhere that using `Unit` didn't work until Scala 2.12.3. that would explain it

jozic: as expected
```
[error] /home/jozic/projects/github/scala/src/library/scala/util/Either.scala:468: private class FromLeftPartiallyApplied escapes its defining scope as part of type scala.util.Either.FromLeftPartiallyApplied[B]
[error]   def left[B]: FromLeftPartiallyApplied[B] = new FromLeftPartiallyApplied[B]
[error]                ^
[error] one error found
[error] (library/compile:compileIncremental) Compilation failed
```

src/library/scala/util/Either.scala
LukaJCB: I'm fairly sure there's no footguns here :)

dwijnand:
Needs a rebase, given https://github.com/scala/scala/pull/6681 introduced EitherTest too.

jozic:
rebased on 2.13.x

lrytz:
Another approach would be adding methods to `Left` / `Right` that upcast an instance.

```
final case class Left[+A, +B](value: A) extends Either[A, B] {
  def up[T >: B]: Either[A, T] = this
  ...
}
```

```
scala> Left(1).up[String]
res0: scala.util.Either[Int,String] = Left(1)
```

jozic:
@lrytz I guess that also would work
Do you want me to submit another PR with this approach and let others decide?

lrytz:
I'd say we can just discuss here. But if we go that other way, a new PR is less confusing for future us.

test/junit/scala/util/EitherTest.scala
```diff
@@ -13,9 +14,34 @@ class EitherTest {
...
+
+    assertEquals(rightSumOrLeftEmpty(List(1, 2, 3)), Right(6))
+    assertEquals(rightSumOrLeftEmpty(Nil), Left("empty"))
+  }
+
+  @Test
+  def testRight: Unit = {
+
+    def leftSumOrRightEmpty(l: List[Int]) =
+      l.foldLeft(Either.right[Int]("empty")) {
```
jozic:
```scalal.foldLeft(Right("empty").up[String]) {

adriaanm:
👋 what's the status here?

jozic:
@adriaanm it went through all checks, and to my understanding lacks only acceptance of someone from the team (that could be you :))
@lrytz suggested another approach but so far no one commented on which one is preferred and why

@SethTisue do you have an idea what should the next step be?

SethTisue:
@jozic Lukas's suggestion is a lot simpler and on that grounds alone seems preferable to me. so yes, new PR with that, please. I'll provisionally close this one.

jozic:
@SethTisue new PR is https://github.com/scala/scala/pull/7011

PR #6331, opened 2018-02-16
Fix typo/dead code in grammar production in SLS

Blaisorblade:
Make the production for `SimpleExpr1` match with the ones from the appendix.

@martijnhoekstra said he learned it's in fact dead specification material, but should still be removed: https://gitter.im/scala/contributors?at=5a87302d17b16a237757b79e

### DIFF
--- a/spec/06-expressions.md
+++ b/spec/06-expressions.md
@@ -32,7 +32,7 @@ SimpleExpr1  ::=  Literal
                |  Path
                |  ‘_’
                |  ‘(’ [Exprs] ‘)’
-               |  SimpleExpr ‘.’ id s
+               |  SimpleExpr ‘.’ id
                |  SimpleExpr TypeArgs
                |  SimpleExpr1 ArgumentExprs
                |  XmlExpr


som-snytt:
Folks have asked that the summary appendix always align with the snippets in the sections. Worth automating.

Blaisorblade:
> Folks have asked that the summary appendix always align with the snippets in the sections. Worth automating.

I either agree or fear https://xkcd.com/1319/*. But if we don't automate it, the next best thing is to have a PR review checklist, and add as a step for spec PRs "check that the summary appendix aligns with the snippets".

Using checklists to check for common bugs in this way is a very good software engineering practice—sure, it's a way to work around deficiencies in tools, but what else can you do?

*Automating this means designing (a special case of) Knuth-style literate programming support, since snippets from the sections are weaved together in the appendix (modern literate programming systems tend to skip this since modern languages are flexible enough this is less necessary, Knuth used Pascal for TeX, so weaving was indispensable). Of course, one could also produce text using an (Scala) EDSL and reuse the host language's abstraction support, but I've never seen anybody find this worthwhile enough to implement.

SethTisue (approved):

SethTisue:
@hrhino on Gitter:

> Part of me is scared that this is actually the first letter of "SEND HELP"

Blaisorblade:
@SethTisue can you send help to the CI? After 3 days I guess something's wrong?

PR #6332, opened 2018-02-17
Actually remove @bridge.

hrhino:
Follow-up to scala/scala#6268 / 1eb0fba.

Turns out that `rm src/library/scala/annotation/bridge.scala` isn't the entire story: `Namers` references `bridge` by name, and scaladoc uselessly checks `isBridge` (the compiler doesn't set this flag until `erasure`, when the generated bridges show up).

Removed one of the scaladoc tests which tested that this annotation didn't do more than it was supposed to.

Thanks @SethTisue for keeping me honest about this.


hrhino:
/sync

(Scabot, Y u No Cooperate)?

NthPortal:
This overlaps a bit with #6319. I removed the annotation, but I missed the spots in the compiler where it's referenced indirectly (as a string, or a method containing `bridge`).

NthPortal:
Would you be okay with me cherry-picking this into #6319?

hrhino:
Ah, yeah, so you did. I missed that somehow. I don't mind you cherry-picking this across. LMK if/when you do and I'll close this out. Nice PR, btw.

NthPortal:
Thanks!

I've cherry-picked this across as 1b5429d

hrhino:
👍 thanks!

NthPortal:
I just noticed you originally removed `@bridge` from STARR, so thank you as well!

PR #6334, opened 2018-02-17
[SUPERSEDED] Adapt errors are emitted in patterns

som-snytt:
In patterns, implicits aren't enabled, but
enrichments are. Emit errors in that case;
otherwise, erroneous trees leak.

Fixes scala/bug#10731

[s: waiting for push, better error message]

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -892,7 +892,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         }
 
         def cantAdapt =
-          if (context.implicitsEnabled) MissingArgsForMethodTpeError(tree, meth)
+          if (context.implicitsEnabled || context.enrichmentEnabled) MissingArgsForMethodTpeError(tree, meth)
           else setError(tree)
 
         def emptyApplication: Tree = adapt(typed(Apply(tree, Nil) setPos tree.pos), mode, pt, original)
--- a/test/files/neg/t10474.check
+++ b/test/files/neg/t10474.check
@@ -1,0 +1,11 @@
+t10474.scala:8: error: missing argument list for method Foo in object Test
+Unapplied methods are only converted to functions when a function type is expected.
+You can make this conversion explicit by writing `Foo _` or `Foo(_)` instead of `Foo`.
+    case Foo.Bar ⇒ true
+         ^
+t10474.scala:15: error: missing argument list for method Foo in trait hrhino
+Unapplied methods are only converted to functions when a function type is expected.
+You can make this conversion explicit by writing `Foo _` or `Foo(_)` instead of `Foo`.
+  val Foo.Crash = ???
+      ^
+two errors found
--- a/test/files/neg/t10474.scala
+++ b/test/files/neg/t10474.scala
@@ -1,0 +1,16 @@
+
+object Test {
+  def Foo(a: Int): Char = ???
+
+  object Bar
+
+  def crash[A](): Boolean = Bar match {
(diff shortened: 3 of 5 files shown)


hrhino:
This fixes scala/bug#10474 as well, it seems (see #6042 for the huge buildup of yak hair caused by my attempt). 

```
Welcome to Scala 2.12.5-20180217-103230-bd8b51c (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> def Foo(a: Int): Char = ???
Foo: (a: Int)Char

scala>   def crash[A](): Boolean = Bar match {
     |     case Foo.Bar ⇒ true
     |   }
<console>:12: error: not found: value Bar
         def crash[A](): Boolean = Bar match {
                                   ^
<console>:13: error: missing argument list for method Foo
Unapplied methods are only converted to functions when a function type is expected.
You can make this conversion explicit by writing `Foo _` or `Foo(_)` instead of `Foo`.
           case Foo.Bar ⇒ true
                ^
```

lrytz:
@som-snytt could you include the test case in https://github.com/scala/bug/issues/10474?

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -892,7 +892,7 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         }
 
         def cantAdapt =
-          if (context.implicitsEnabled) MissingArgsForMethodTpeError(tree, meth)
+          if (context.implicitsEnabled || context.enrichmentEnabled) MissingArgsForMethodTpeError(tree, meth)
```
adriaanm, line 895: I never really figured out what's going on here, but my ramblings from https://github.com/scala/scala/pull/6042/files/eec156250d22cdc97dbeb0c66263ff3da74b1268#diff-4eab1aad4533a31c10565971e90f73ea point towards this condition checking that we're not currently type checking a pattern. This change would reverse that. 

In any case (ha!), reporting an error is progress. However, let's provide a better suggestion than eta expansion, which is not allowed in a pattern position.

adriaanm (changes requested):
Needs a more applicable error message.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
som-snytt: Also, "applicable error message" ha!

som-snytt: I tried to look back in history to see when/why the conditional was introduced and failed.

som-snytt: OK, I'll read Adriaan's thread and add some LOC. It means I'll lose my scala center certificate for shortest fix of shortest crasher.

adriaanm:
Today is the last day to shield 2.12.5 from this particular indignity. `</voice ="Herzog">`

adriaanm:
I was feeling protective, so I pulled some slayer-protection into #6273. There's still plenty of room for error message improvement

som-snytt:
@adriaanm I was just reviewing your other long thread; I must have fallen asleep in a late chapter the first time I read it.

I see the crux is fine-tuned messaging, and that you are implementing your ideas for detecting what mode we're in. Do you want me to attempt something today? It's not quite the work hour here yet.

adriaanm:
I think we're too close to the deadline to attempt anything further today. The ideas I pushed to your branches didn't pan out, so I say we sleep on it and try after 2.12.5

adriaanm:
I'm going to close this PR -- I suggest we tackle both issues in the other PR

PR #6335, opened 2018-02-18
Don't accept wildcard type bounds.

### LINKED ISSUE #10733: Parser accepts wildcards too freely
The spec:
```
  SimpleType        ::=  SimpleType TypeArgs
                      |  SimpleType ‘#’ id
                      |  StableId
                      |  Path ‘.’ ‘type’
                      |  ‘(’ Types ‘)’
```

The [doc comment](https://github.com/scala/scala/blob/v2.12.4/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala#L1013-L1020) in `Parsers` repeats that, but adds `WildcardType` as a production.

Therefore, this is accepted:
```scala
scala> tq"L[_ <: _ >: _ <: _ >: Int]"
res0: $r.intp.global.ExistentialTypeTree =
L[_$2] forSome {
  <synthetic> type _$1 >: Int;
  <synthetic> type _$3 <: _$1;
  <synthetic> type _$4 >: _$3;
  <synthetic> type _$2 <: _$4
}
```

If for some reason that tree is at all useful and the spec is overly limiting, the bug becomes that that `ExistentialTypeTree` doesn't pass position validation.

hrhino:
Specifically, wildcards aren't a first-class type according to the spec, which claims that they "... must appear as a type argument of a parameterized type.".

Putting wildcards in places they don't belong usually draws either "unbound placeholder" or "identifier expected" errors, but if we get to `simpleType` with one we can be a bit more specific.

Also scouted the grammar of that spec fragment.

Fixes scala/bug#10733.


som-snytt:
This currently works because the two forms are the same:
```
scala> val f: _ => Int = (x: Any) => 42
              ^
       error: wildcards are only allowed as type arguments

scala> val f: Function1[_, Int] = (_: Any) => 42
f: Function1[_, Int] = $$Lambda$1818/871554897@107bfcb2
```

som-snytt:
Regressive messaging:
```
scala> def f(x: _ <: String) = ???
                ^
       error: not found: type _$1
```

Is it possible to just add the extra advice to the current message?

hrhino:
The messaging isn't really the part that I'm concerned about; I just thought it would be a nice bonus. The real issue here is that `List[_ <: _ <: Int]` turns into `List[x$1] forSome { type x$1 <: x$2; type x$2 <: Int }` and the position for both type declarations in the refinement claim to contain the position of `x$2`. 

I had considered finding a way to patch that up (the position for `type x$1` ranges over `_ <: ` only?) but since this is dubiously useful and likely illegal syntax it seemed more expedient to simply forbid it.

som-snytt:
Yeah I was looking at it again because you linked that other existential ticket. Nobody liked the [commit message](https://twitter.com/som_snytt/status/965451383468994563) I came across, a classic paulpism, except one fake twitter featuring a naked person.

I was about to check out why normal typebounds works out differently. Usually, underscore and ident are pretty symmetrical. The change you point out that introduced it was paulp reducing boolean args.

Probably I'll come close to understanding existentials just as dotty swoops in to eliminate them.

hrhino:
Yeah, the awkward part here is that when parsing `(_ <: Int) => Int` we don't know, at the point of `_`, that we're within a parameter of a function. As far as I understand it the spec forbids that, too, which probably helps with making the grammar context-free. I think it's an important enough use that we should probably support it anyhow.

Possibly the solution is to parse like it does right now, and then reject it at `typ()`  if we're not looking for a type argument. I'll poke at that.

EDIT: paulp's commit messages are half the reason I open the annotation sidebar in IntelliJ. They take the edge off the frustration at times.

test/files/pos/existental-slow-compile2.scala
```diff
@@ -1,7 +0,0 @@
-class C {
-  class L[+A]
```
som-snytt: This is testing something different from slow-compile1.scala, right?

som-snytt:
That's a clever check. I can't tell (off the top of my head) if we'd ever see the current message.

I expected to see just extra advice on the current message like [in my stab](https://github.com/scala/scala/compare/2.12.x...som-snytt:issue/10733).

I wonder if you could add those two finals in my stab? Reading the paulp refactor, it's not obvious what is overridden in the little hierarchy. That has confused me the few times I've ever looked at it.

I was going to add something around that weird arg to `paramType(useStartAsPosition = true)` which doesn't express much about what is going on. Parsing is a small part of the runtime, it should also be a small part of time spent understanding and fixing bugs...

hrhino:
Indeed. Thinking at this now I should probably just use `checkNoEscapingPlaceholders` like elsewhere.

The extra advice is nice.

I've never fully understood the control flow through that part of the parser. I think it may be offended by me poking at its sensitive parts. More finality sounds good.

The current message shows for `val whoknows: _ = 1` and similar, btw.

hrhino:
The two `final`s you added are overridden in `outPattern` and `SeqContextSensitive`. `argType` being sometimes a special `paramType` but only if `argType` is actually `typ` is all boiling down to a bunch of logic that I thought I understand but maybe don't.

I like the error message improvement, though.

som-snytt:
Make them final in `SeqContextSensitive` [here](https://github.com/som-snytt/scala/blob/5065174b2f197ec274770ad4ef75c20292b70af8/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala#L1890).

Not sure if there's a reason not to make the implementations vals [something like](https://github.com/scala/scala/pull/6466/files).

som-snytt:
@hrhino ping, I noticed this on my laptop.

hrhino:
Thanks @som-snytt. I read the comments thread earlier and thought maybe it would be better to start from scratch. After more thought, maybe not. 

I'll reread it for real over the weekend.

smarter:
See also https://github.com/lampepfl/dotty/pull/4422

hrhino:
@smarter nice! do you mind if I reappropriate the test cases there, too?

smarter:
Help yourself.

adriaanm:
:wave:

hrhino:
Fun times:

```scala
hhoughton:~ hhoughton$ scala-launch 2.12.6
Welcome to Scala 2.12.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_161).

scala> type T = _
<console>:11: error: unbound wildcard type
       type T = _
                ^

scala> :quit
hhoughton:~ hhoughton$ scala-launch 2.13.0-M4
Welcome to Scala 2.13.0-M4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_161).

scala> type T = _
                ^
       error: not found: type _$1

scala> :quit
```

SethTisue:
where do we stand, here? tentatively pushing to 2.13.0-RC1

hrhino:
That's fine. I'm revisiting this right now, actually. Incorporating the dotty test showed changes like the one I just posted, making it a not-completely-applicable addition to our test suite (yet). I'll file a bug for that regressive messaging, as well as the one that @som-snytt helpfully yet misleadingly pointed out above.

som-snytt:
I think Scala should accept anything wild, like those poker games that go over my head. Otherwise, I defer to @hrhino 's excellent judgment. In fact, the wilder, the better!

hrhino:
Sprayed some scope-creep-B-gon on this PR. It now addresses exactly and only the type bounds issue, which hopefully makes it acceptable enough.

Filed the regressive messaging bug at scala/bug#11064.

som-snytt:
"Type Bound" was my first Gina Gershon movie. Probably we need more bound and binding puns.

hrhino:
"Wildcards Unbound" sounds like some kind of action movie.

lrytz (approved):

som-snytt:
I just noticed this on my laptop again, I guess Feb 2018 was the beginning of the election cycle just completed?

I have an extra commit with a spec change. After I get some sleep, I'll check whether the spec needs updating.

PR #6336, opened 2018-02-18
Right-associative arguments don't lose singleton type

hrhino:
`x :: y` becomes `{ val rassoc$1 = x ; y.::(rassoc$1) }` during parser. That `val` causes `rassoc$1` to get a non-singleton type, which isn't necessarily what you want. 

The secret handshake, of course, is to make `rassoc$1` a `final val`.

Fixes scala/bug#10726.

### DIFF
--- a/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
+++ b/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala
@@ -848,10 +848,14 @@ self =>
       }
       if (isExpr) {
         if (rightAssoc) {
+          import symtab.Flags._
           val x = freshTermName(nme.RIGHT_ASSOC_OP_PREFIX)
+          val liftedArg = atPos(left.pos) {
+            ValDef(Modifiers(FINAL | SYNTHETIC | ARTIFACT), x, TypeTree(), stripParens(left))
+          }
           Block(
-            List(ValDef(Modifiers(symtab.Flags.SYNTHETIC | symtab.Flags.ARTIFACT), x, TypeTree(), stripParens(left))),
-            Apply(mkSelection(right), List(Ident(x))))
+            liftedArg :: Nil,
+            Apply(mkSelection(right), List(Ident(x) setPos left.pos.focus)))
         } else {
           Apply(mkSelection(left), arguments)
         }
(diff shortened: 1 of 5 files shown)


milessabin (approved):
Yes, this looks good to me.

hrhino:
This only works once:

```scala
> scala
[info] Running scala.tools.nsc.MainGenericRunner -usejavacp
Welcome to Scala 2.13.0-20180216-202027-778b8fd (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).

scala> ("" :: Nil) : List[""]
res0: List[""] = List("")

scala> ("" :: "" :: Nil) : List[""]
           ^
       error: inferred type arguments [""] do not conform to method ::'s type parameter bounds [B >: String]
```

because `"" :: Nil` is typed as `List[String]`. I'm not sure what can be done about that. It's specced like that, so likely nothing, but it feels less than optimal.

Edit: and people say Scala has too many symbols:
```scala
scala> "" :: [""] "" :: [""] Nil
res3: List[""] = List("", "")
```

milessabin:
Why is this being closed without merging?

NthPortal:
I believe you got an answer on Gitter, @milessabin, but in case anyone else comes across this, it was due to legal issues, not because the PR is problematic.

retronym:
@hrhino I've rebased this on your behalf to please the CI gods.

hrhino:
ah, thanks! I tried that once but I forgot to anoint myself with blessed oils and force-push at the wax of the moon.

lrytz (approved):
`-f`-pushed again

hrhino:
🙇 🙇

lrytz:
@hrhino could you check why the patch in this PR (after rebasing) changes the output of `run/t10751.scala`?

hrhino:
Good catch, thanks. Updated, and I added a handful of more cases for the positions around right-associative arguments and final vals.

lrytz:
@hrhino pushed another commit that removes the unreferenced `rassoc$n` ValDefs, in case the references to them get constant-folded. Let me know what you think.

hrhino:
That commit LGTM indeed. I considered doing that but decided it may have been too much scope creep for one (originally tiny) PR. Thanks!

hrhino:
On June 1, 2018 4:28:59 PM EDT, Lukas Rytz <notifications@github.com> wrote:
>Merged #6336.

Thanks!

PR #6337, opened 2018-02-18
Set pointer-events to none in headings

Philippus:


Blaisorblade:
Not tested myself but thanks for the prompt PR!

If you're curious about the syntax for linking, it's a bit fiddly: just write `Fix scala/bug#10728` in the description (not subject) and the link will work (result: scala/bug#10728) ;-).

BTW, if you're into webdev maybe you can also help us at https://github.com/lampepfl/dotty/issues/3952?

hrhino:
I think that scoping this fix to only `#authors-and-contributors` is too narrow. For instance, near the beginning of [Chapter 5](https://www.scala-lang.org/files/archive/spec/2.12/05-classes-and-objects.html), the `<h2>` for section 5.1 also sticks up and over the links: 

screenshot 2018-02-18 18 22 38

I've not dug too deeply into why the `margin-top: -32px; padding-top: 32px` thing is done, but maybe a `main h1, ... main h6 { pointer-events: none }` would work (except, sadly, on [opera mini](https://caniuse.com/#search=pointer-events)...).

Blaisorblade:
Maybe @b-studios has a better idea, though he's probably busy?

I'm very ignorant in webdev, so feel free to correct everything I say next.

> I've not dug too deeply into why the margin-top: -32px; padding-top: 32px thing is done,

Doh. That's to fix the anchor position (https://github.com/scala/scala/commit/62db8d98559766c06d490d9d4c18ff27587fde0b#diff-ceff2600565be2f4c508af28af3e2225R50), since the nav bar is exactly 32px high. And the navbar has `position: fixed`. And the whole thing is a fancy responsive layout.

I wanted to say "we probably don't have an army of frontend devs so let's maybe do something simpler", but I don't see how.

hrhino:
> That's to fix the anchor position, since the nav bar is exactly 32px high.

Ah, this is what I get from only poking around in devtools and not actually reading the source (with comments). I should learn how that works but I'm not even worthy to polish the weapons of the lowest-ranking private in that army, so I'll accept that it's necessary.

b-studios:
Haven't checked it, but maybe it is enough to apply the fix (-32 +32) once on the body and not on the element itself. This way the space which is obscured by the navbar wouldn't be used at all (I hope).

Philippus:
I've applied @hrhino's suggestion

lrytz (approved):

SethTisue:
thanks Philippus!

PR #6340, opened 2018-02-19
Support implicitNotFound on parameters

lrytz:
This allows a more specific message for individual implicit parameters
that are not found.

This allows improving error message, for example in overloading
resolution:

```
scala> trait Iterable[A] { def map[B](f: A => B): Iterable[B] }
scala> trait SortedSet[A] { def map[B: Ordering](f: A => B): SortedSet[B] }

scala> def t(s: SortedSet[Int]) = s.map(x => x) // OK
t: (s: SortedSet[Int])SortedSet[Int]

scala> def t[T](s: SortedSet[T]) = s.map(x => x) // Confusing error
                                        ^
       error: No implicit Ordering defined for T.
```

better:

```
scala> trait SortedSet[A] { def map[B](f: A => B)(implicit @annotation.implicitNotFound("No implicit Ordering[${B}] found to build a SortedSet[${B}]. You can build an unordered Set[${B}] by calling `unsorted` before `map`.") ord: Ordering[B]): SortedSet[B] }
defined trait SortedSet

scala> def t[T](s: SortedSet[T]) = s.map(x => x)
                                        ^
       error: No implicit Ordering[T] found to build a SortedSet[T]. You can build an unordered Set[T] by calling `unsorted` before `map`.
```

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Implicits.scala
@@ -1595,7 +1595,7 @@ trait Implicits {
   }
 
   class ImplicitAnnotationMsg(f: Symbol => Option[String], clazz: Symbol, annotationName: String) {
-    def unapply(sym: Symbol): Option[(Message)] = f(sym) match {
+    def unapply(sym: Symbol): Option[Message] = f(sym) match {
       case Some(m) => Some(new Message(sym, m, annotationName))
       case None if sym.isAliasType =>
         // perform exactly one step of dealiasing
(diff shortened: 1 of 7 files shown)


lrytz:
See also https://github.com/scala/collection-strawman/issues/432

soronpo:
What happens in the following case?
```scala
@implicitNotFound("BarError")
trait Bar
@implicitNotFound("FooError")
trait Foo
object Foo {
  implicit def ev(implicit @implicitNotFound("BarParamError") bar : Bar) : Foo = ???
}

implicitly[Foo]
```

lrytz:
It tells you
```
scala> implicitly[Foo]
                 ^
       error: FooError
```

With `-Xlog-implicits`

```
scala> implicitly[Foo]
                 ^
       Foo.ev is not a valid implicit value for Foo because:
       hasMatchingSymbol reported error: BarParamError
                 ^
       error: FooError
```

So it's picked up here as well.

test/files/neg/t2462b.check
```diff
@@ -6,6 +6,10 @@ t2462b.scala:9: warning: Invalid implicitNotFound message for trait Meh2 in pack
 The type parameter Elem referenced in the message of the @implicitNotFound annotation is not defined by trait Meh2.
 trait Meh2[-From, +To]
       ^
+t2462b.scala:13: warning: Invalid implicitNotFound message for value theC:
+The type parameter Uuh referenced in the message of the @implicitNotFound annotation is not defined by value theC.
```
szeiger: I find this message confusing. The type parameter is supposed to be defined by `m`, not by `theC`.

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1586,8 +1586,14 @@ trait Implicits {
           // #3915: need to quote replacement string since it may include $'s (such as the interpreter's $iw)
       })
 
-    private lazy val typeParamNames: List[String] = sym.typeParams.map(_.decodedName)
-    private def typeArgsAtSym(paramTp: Type) = paramTp.baseType(sym).typeArgs
+    private lazy val typeParamNames: List[String] = {
+      val s = if (sym.isParameter) sym.enclMethod else sym
```
szeiger: Can (and should) we get the full scope here instead of only the method's type parameters? This doesn't work:

```scala
scala> trait I[X, Y]
defined trait I

scala> trait T[A] { def f[B](implicit @scala.annotation.implicitNotFound("Missing I[${A}, ${B}]") i: I[A, B]): Unit = () }
                                                                                                  ^
       warning: Invalid implicitNotFound message for value i:
       The type parameter A referenced in the message of the @implicitNotFound annotation is not defined by value i.
defined trait T
```

lrytz:
Updated. When putting an `@implicitNotFound` on a parameter, the message can now refer to any type paramters in scope. (hmm, that makes me think if it should support type members too.. that might be hard to implement).

test/files/neg/t2462c.check
```diff
@@ -1,7 +1,16 @@
...
   f[Foo[Int]]
    ^
-two errors found
+t2462c.scala:33: error: No C of Foo[Int]
+  g[Foo[Int]]
+   ^
+t2462c.scala:36: error: I see no C[Foo[Int]]
+  h[Foo[Int]]
+   ^
+t2462c.scala:40: error: String List[?] List[C[_]] Int Option[Long] -- .
```
lrytz: The `List[?]` is not great here. [The `formatParameterMessage` method](https://github.com/scala/scala/pull/6340/files#diff-7c03397456d3b987549fd039d6b639c6R1657) could be improved for the case of type constructors. cc @adriaanm.

lrytz:
Maybe we can say this is good enough to merge? It's only about error messages. We can always accept contributions that improve it :) I'd like to not spend much more time on this.

soronpo:
What will happen when an implicit parameter is missing from an implicit conversion? :
```scala
@implicitNotFound("BarError")
trait Bar
@implicitNotFound("FooError")
trait Foo
object Foo {
  implicit def fromBar(bar : Bar)(implicit @implicitNotFound("FooParamError") foo : Foo) : Foo = foo
}

val foo : Foo = new Bar{}
```
Currently (and sadly) scalac does not even report `FooError`. Will it now report `FooParamError` ?

src/compiler/scala/tools/nsc/typechecker/Implicits.scala
```diff
@@ -1628,25 +1628,65 @@ trait Implicits {
...
+      interpolate(msg, Map(symTypeParamNames zip typeArgs: _*))
 
-    def format(typeArgs: List[String]): String =
-      interpolate(msg, Map((typeParamNames zip typeArgs): _*)) // TODO: give access to the name and type of the implicit argument, etc?
+    def formatParameterMessage(fun: Tree): String = {
+      val paramNames = referencedTypeParams
+      val paramSyms = paramNames.map(lookupTypeParam).filterNot(_ == NoSymbol)
+      val paramTypeRefs = paramSyms.map(sym => typeRef(NoPrefix, sym, sym.typeParams.map(_ => WildcardType)))
+      val prefix = fun match {
+        case TypeApply(Select(qual, _), _) => qual.tpe
```
adriaanm: could collapse these two cases (and also deal with repeated Apply) using: `case treeInfo.Applied(Select(qual, _), _, _) => qual.tpe`

adriaanm:
> What will happen when an implicit parameter is missing from an implicit conversion?

Same as before. No (user-defined) error. Having some form of interactive implicit search debugger would be great, but for error messages I don't think we should try to guess secondary issues.

soronpo:
>Same as before. No (user-defined) error. Having some form of interactive implicit search debugger would be great, but for error messages I don't think we should try to guess secondary issues.

In many cases I don't think it's guessing and I would have loved a custom error instead of just a `type mismatch`. More specifically, in case where an implicit conversion only fails due to a missing implicit parameter and there is only one possible implicit conversion definition in scope.

test/files/neg/t2462c.check
adriaanm: consider me sniped :-) I pushed a commit that adds some ad-hoc pretty printing for these. Feel free to `g push -f HEAD^`

adriaanm:
We could discuss that in a ticket and a different PR, but it's definitely out of scope for this one.

adriaanm (approved):

adriaanm:
@lrytz, others: the first commit LGTM, but I'll let you decide on my follow-up :-)

soronpo:
> We could discuss that in a ticket and a different PR, but it's definitely out of scope for this one.

@adriaanm I created a separate ticket for that
https://github.com/scala/bug/issues/10946

With some help I can create a PR, if you'd consider accepting it.

test/files/neg/t2462c.check
```diff
@@ -10,7 +10,7 @@ t2462c.scala:33: error: No C of Foo[Int]
 t2462c.scala:36: error: I see no C[Foo[Int]]
   h[Foo[Int]]
    ^
-t2462c.scala:40: error: String List[?] List[C[_]] Int Option[Long] -- .
+t2462c.scala:40: error: String List [?T0, ZZ] -> List[C[_]] Int Option[Long] -- .
```
lrytz: People will complain why this syntax doesn't work in their code 😏

adriaanm:
Neither do wildcards :-p

adriaanm:
> With some help I can create a PR, if you'd consider accepting it.

let's first discuss on the ticket -- I'm not against on principle or anything, but it's probably worth talking through before embarking on the implementation

PR #6341, opened 2018-02-20
upgrade scala-xml to 1.1.0

SethTisue:
the hope of course is to get scala-xml out of the 2.13 bootstrap, but
I'm not sure we'll get to it before M4. in the meantime, let's dogfood
the new release.

### DIFF
--- a/src/intellij/scala.ipr.SAMPLE
+++ b/src/intellij/scala.ipr.SAMPLE
@@ -201,7 +201,7 @@
         <root url="jar://$USER_HOME$/.ivy2/cache/org.apache.ant/ant/jars/ant-1.9.4.jar!/" />
         <root url="jar://$USER_HOME$/.ivy2/cache/org.apache.ant/ant-launcher/jars/ant-launcher-1.9.4.jar!/" />
         <root url="jar://$USER_HOME$/.ivy2/cache/org.scala-lang.modules/scala-asm/bundles/scala-asm-5.2.0-scala-2.jar!/" />
-        <root url="jar://$USER_HOME$/.ivy2/cache/org.scala-lang.modules/scala-xml_2.13.0-M3/bundles/scala-xml_2.13.0-M3-1.0.6.jar!/" />
+        <root url="jar://$USER_HOME$/.ivy2/cache/org.scala-lang.modules/scala-xml_2.13.0-M3/bundles/scala-xml_2.13.0-M3-1.1.0.jar!/" />
         <root url="jar://$USER_HOME$/.ivy2/cache/jline/jline/jars/jline-2.14.5.jar!/" />
       </CLASSES>
       <JAVADOC />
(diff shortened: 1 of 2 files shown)


retronym:
This seems to break the boostrap build, at least the one used in the benchmarks.

For example, in build [bootstrap-benchmark/637/](
https://scala-ci.typesafe.com/job/bootstrap-benchmark/637/), the [build log](
https://scala-ci.typesafe.com/job/bootstrap-benchmark/637/artifact/logs/builds/*view*/) shows the failure due to the addition of ScalaJS to the scala-xml build.

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.scala-js#scalajs-library_2.13.0-pre-f993373;0.6.22: not found
[warn] 	:: org.scala-js#scalajs-test-interface_2.13.0-pre-f993373;0.6.22: not found
[warn] 	:: org.scala-js#scalajs-junit-test-runtime_2.13.0-pre-f993373;0.6.22: not found
[warn] 	:: org.scala-js#scalajs-compiler_2.13.0-pre-f993373;0.6.22: not found
[warn] 	:: org.scala-js#scalajs-junit-test-plugin_2.13.0-pre-f993373;0.6.22: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

retronym:
@sjrd I remember you experimenting with a way to easily enable/disable the ScalaJS SBT plugin. Did that turn into something we can use?

sjrd:
If you're talking about https://github.com/sjrd/sbt-dynscalajs, it works, yes. But please mind the readme about the expected level of support.

Anyway, I don't think your problem here is that it can't load the plugin. The resolution errors you see only happen if you actually try to `compile` the Scala.js part of the cross project. An easier and more stable fix would simply be to only `compile` (etc.) the JVM part of the cross projects inside scala-xml in the bootstrap build.

ashawley:
If you want to execute just the JVM targets in the scala-xml build, you can use the `xmlJVM` project in the scala-xml repo, so in this case you'd use `xmlJVM/test` and `xmlJVM/publish` and so on.

Why did this come up with upgrading the build to 1.1.0?  The Scala.js changes have been in master for a year or so, now.

retronym:
Our bootstrap build was rebuilding the v1.0.6 tag until this change.

retronym:
> xmlJVM/test and xmlJVM/publish ...

Thanks for this. I'm trying it in: https://github.com/scala/scala/pull/6409

ashawley:
What's the point of that bootstrap script?  There's a good comment at the top on what the script does, but not the _why_.  Does it "Verify the compiler can build itself, including test dependencies and any required Scala modules"?  Is this to verify compatibility issues?

retronym:
An important part of the "why" is that in order to publish `scala-{library,reflect,compiler}` to Maven Central, we need to create doc JARs. This requires scala-xml. If we're in a pre-release of Scala which doesn't make any binary compatibilty guarantees from one miletstone to the next, we can't necessarily use a published build of scala-xml. So we build it ourselves in the script.

ashawley:
Oh, ok.  The point *is* missing dependencies for milestone and new major Scala versions, thus the complete bootstrap.  Thanks for the explanation.  I'm very familiar with the Scaladoc dependency.

PR #6342, opened 2018-02-21
Remove statistics reporting code from some hot paths

retronym:
Recent changes to the implenentation of -Ystatistics had the goal of
having zero overhead in the common case: when statistics had never been
enabled in any instance of Global.

However, I still see evidence of method profile samples in
`areSomeColdStatsEnabled` via `phase`, `flags`, `lookupEntry` and `owner`.

I can't convince myself if this is profiler noise or a bona fide cost.
This commit doesn't show a clear speedup in our benchmarks, for example.

I haven't found those particular counters to be useful in directing performance profiling. Let's remove this a moving part the way we reason about the performance of these hot methods.

### DIFF
--- a/src/reflect/scala/reflect/internal/SymbolTable.scala
+++ b/src/reflect/scala/reflect/internal/SymbolTable.scala
@@ -188,8 +188,6 @@ abstract class SymbolTable extends macros.Universe
 
   final def atPhaseStack: List[Phase] = phStack.toList
   final def phase: Phase = {
-    if (StatisticsStatics.areSomeColdStatsEnabled)
-      statistics.incCounter(statistics.phaseCounter)
     ph
   }
 
@@ -462,7 +460,6 @@ abstract class SymbolTable extends macros.Universe
 trait SymbolTableStats {
   self: TypesStats with Statistics =>
 
-  val phaseCounter = newCounter("#phase calls")
   // Defined here because `SymbolLoaders` is defined in `scala.tools.nsc`
   // and only has access to the `statistics` definition from `scala.reflect`.
   val classReadNanos = newSubTimer("time classfilereading", typerNanos)
--- a/src/reflect/scala/reflect/internal/Symbols.scala
+++ b/src/reflect/scala/reflect/internal/Symbols.scala
@@ -769,7 +769,6 @@ trait Symbols extends api.Symbols { self: SymbolTable =>
     }
 
     final def flags: Long = {
-      if (StatisticsStatics.areSomeColdStatsEnabled) statistics.incCounter(flagsCount)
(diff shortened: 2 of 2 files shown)


retronym:
Already reviewed by @jvican in the context of #6209. I decided to retarget this particular change to 2.12.x

jvican (approved):

jvican:
I've seen a significant difference when applying #6209 on a local branch to make experiments with statistics. Happy to see this landing 2.12.5.

mkeskells:
Can you supply details of this significant difference. Have you looked at the disassembled code?

retronym:
Here are the [benchmarks](https://scala-ci.typesafe.com/grafana/dashboard/db/pr-validation?orgId=1&var-scalaSha=2791989109%7Cea64efaf0a%7Cce95eaf85f%7Ca99e6e3a3e%7C18d79975ce%7Ca7d7a7e132%7Cd1b745c2e9%7C9b89fe9e00%7Cea2482c64d%7C03b0b7de1e%7C241e050a67&from=1503310963352&to=1519208563352&var-source=All&var-bench=HotScalacBenchmark.compile&var-host=scalabench%40scalabench%40) for just this change. The last point is this PR. Some flat, some improvements.

Flight Recorder with `-XX:+DebugNonSafepoint` enabled reported time in `boolean scala.reflect.internal.util.StatisticsStatics.areSomeColdStatsEnabled`. My attempts to disect this with disassembly and JITWatch were inconclusive.

One part I was bit suspicious about is the fact that we can't force JIT inlining of `areSomeColdStatsEnabled` nor of `isEnabledNow`, even though we know that `invokeExact` called on a constant is guaranteed to be inlined. So maybe the efficiency claim of this `AlmostFinalField` pattern was wrong, or we haven't faithfully translated it.

```java
public static boolean areSomeColdStatsEnabled() {
    try {
      return ((BooleanContainer)(Object) COLD_STATS_GETTER.invokeExact()).isEnabledNow();
    } catch (Throwable e) {
      throw new AssertionError(e.getMessage(), e);
    }
  }
```

Or, maybe the problem is that the ["efficient run-time safety check to determine if the symbolic type descriptor encoded at the call site exactly matches the method type descriptor of the MethodHandle"](https://wiki.openjdk.java.net/display/HotSpot/Deconstructing+MethodHandles) (`checkExactType`) still costs enough cycles to make a difference when it is done inside an otherwise trivial method like `phase`.

Or, maybe part if this is one of the [known bugs](https://github.com/jvm-profiling-tools/async-profiler/issues/66) in `AsyncGetCallTrace` which wrongly attributes profiles that land on the end of an assembly routine to some inlined code that was DCE-d.

PR #6343, opened 2018-02-21
Don't add original attachment for constant folded trees

jvican:
Scalac seems to run twice constant-folding in certain scenarios.
Consider the following example:

```scala
object A { final val x = 1 }

object B {
  def main(args: Array[String]) = assert(args(0).toInt == A.x )
}
```

The tree `A.x` in `main` is constant folded because `x` is a constant,
but for some reason the resulting tree `Literal(Constant(1))` get
constant-folded again.

The previous logic would add the original tree attachment on the
constant-folded tree, and would destroy the previous annotation that
gets copied by `treeCopy.Literal`.

Therefore, to avoid this issue, we don't update the original attachment
if there already exists one.

In the future, it would be worthwhile to figure out why scalac tries to
constant-fold twice the same tree.


jvican:
I discovered this issue while addressing https://github.com/sbt/zinc/issues/227. This time I have made sure that I'm able to access the original attachment from Zinc.
  
I'm targeting milestone 2.12.5 because of the triviality of the change and my desire to ship a fix to this solution in a (recent) upcoming version of Zinc.

lrytz:
I took a quick look in the debugger, it's overloading resolution that causes the typer to visit the argument twice. `Int.==` has overloads, the args (`A.x`) are type checked to select the right alternative, then `doTypedApply` is called again with the already typed args. It then just does its thing and goes to `typedArg`, which invokes `typed` on the argument again. That method actually skips type checking if the tree already has a type, which is the case here, but then it still invokes `adapt`, which is where the attachment is updated.

lrytz:
I can't think of a good way to test this.. @retronym do we have something to walk trees after a certain phase?

jvican:
@lrytz Your explanation makes sense. It looks like scalac could be smarter and avoid calling `adaptType` if the target tree is already a literal (in which case it cannot be constant-folded). Do you think this is a petty improvement?  
  
As I didn't find a way to test this, I tested it in my zinc branch where I've been doing some work on this area and confirmed it worked.

retronym:
We could probably test with Junit BytecodeTesting utilities, passing in
compiler args of Ystop-after and then looking through
global.currentRun.units to find the trees.

src/compiler/scala/tools/nsc/typechecker/Typers.scala
```diff
@@ -995,7 +995,8 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         if (sym != null && sym.isDeprecated)
           context.deprecationWarning(tree.pos, sym)
         // Keep the original tree in an annotation to avoid losing tree information for plugins
-        treeCopy.Literal(tree, value).updateAttachment(OriginalTreeAttachment(original))
+        if (tree.hasAttachment[OriginalTreeAttachment]) treeCopy.Literal(tree, value)
+        else treeCopy.Literal(tree, value).updateAttachment(OriginalTreeAttachment(tree))
```
dwijnand: shouldn't this be be adding `OriginalTreeAttachment(original)`?

jvican: `tree == original` in this case, but `original` is a parameter off the enclosing gigantic method while `tree` is the parameter of `adaptType`. I opted for `tree` because it's closer to the use site, and because the original tree from which we copy is `tree` and not `original`.

lrytz:
> test with Junit BytecodeTesting utilities, passing in compiler args of Ystop-after

Great idea, thanks! Here's a test for this PR: https://github.com/scalacenter/scala/compare/topic/fix-original-tree-attachment...lrytz:pr6343?expand=1

@jvican I realized we were adding the attachment to all trees, including literals, which is unnecessary.

lrytz:
We can factor out some helper methods to reduce the (small) boilerplate of tree tests later.

jvican:
Would it make sense that we don't call `adapt` if the target tree is already a literal type? That way, we would avoid copying one extra tree per literal that is adapted. There aren't many literal trees that are adapted, but that would definitely solve the root of the issue rather than work around it (as I do in my PR). Perhaps it would be better to merge this as is (adding tests), and leave the bold change for 2.12.6.

lrytz:
My proposed patch (see above) basically does that, if the orignal tree is a literal it's returned unchanged. I would not move the out and start calling adapt conditionally. Currently every tree that goes through `typed` goes to `adapt`, changing this might have unexpected consequences.

jvican:
> Currently every tree that goes through typed goes to adapt, changing this might have unexpected consequences.
  
I see. Didn't know about this invariant, makes sense. I've cherry-picked your commit, thanks for implementing the test for it! I think this is ready for review.

lrytz:
Could you squash the two commits?

jvican:
I've just squashed the commit, but it lost your attribution -- that's why I wanted to keep them separate. Do you know if there's a way to put you as a co-author?

lrytz:
No need to do it here, but in case you ever need it: https://help.github.com/articles/creating-a-commit-with-multiple-authors

lrytz (approved):

lrytz:
@jvican FYI I pushed a revised version (to your branch), the previous one failed some tests. Here's the diff:
```diff
diff --git a/src/compiler/scala/tools/nsc/typechecker/Typers.scala b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
index 7497ce26cb..8d1fd4ba14 100644
--- a/src/compiler/scala/tools/nsc/typechecker/Typers.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Typers.scala
@@ -994,10 +994,11 @@ trait Typers extends Adaptations with Tags with TypersTracking with PatternTyper
         val sym = tree.symbol
         if (sym != null && sym.isDeprecated)
           context.deprecationWarning(tree.pos, sym)
-        if (tree.isInstanceOf[Literal]) tree
-        else {
-          // If the original tree is not a literal, make it available to plugins in an attachment
-          treeCopy.Literal(tree, value).updateAttachment(OriginalTreeAttachment(tree))
+        tree match {
+          case Literal(`value`) => tree
+          case _ =>
+            // If the original tree is not a literal, make it available to plugins in an attachment
+            treeCopy.Literal(tree, value).updateAttachment(OriginalTreeAttachment(tree))
         }
       }
```

jvican:
Thank you @lrytz!

PR #6344, opened 2018-02-22
Propagate fatal errors during macro expansion.

lrytz:
Fixes scala/bug#10552. Previously, fatal errors got swallowed during macro
expansion.  In the case of implicit blackbox macros, a fatal error (for
example OutOfMemoryException) got reported with the message "exception
during macro expansion". For implicit whitebox macros the fatal error is
not even reported unless -Xlog-implicits is enabled. By default, the
user only sees a cryptic "implicit not found" error message.  See #10649.

This commit changes the error handling of exceptions during macro
expansion to propagate fatal errors. Now fatal errors are left uncaught
and crash compilation with a full stack trace instead of getting
swallowed.

```
error: java.lang.OutOfMemoryError
    at Macros$BlackBox$.materializeImpl(so.scala:8)
```

This change caused the sip-19-macro-revised partest to fail since it
previously relied on triggering a stack overflow to fail expansion if
another implicit SourceContext was in scope. Instead, the test now
guards against infinite recursion itself.

### DIFF
--- a/test/files/run/t10552/Test_2.scala
+++ b/test/files/run/t10552/Test_2.scala
@@ -1,0 +1,14 @@
+import scala.tools.partest._
+
+object Test extends DirectTest {
+  override def extraSettings: String = "-usejavacp -Ystop-after:typer"
+
+  def code = "class C { A.f }"
+
+  def show(): Unit = try {
+    compile()
+    throw new Error("Expected OutOfMemoryError")
+  } catch {
+    case e: OutOfMemoryError if e.getMessage == "OOM" =>
+  }
+}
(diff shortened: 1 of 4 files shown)


test/files/run/t10552/Test_2.scala
```diff
@@ -0,0 +1,14 @@
+import scala.tools.partest._
+
+object Test extends DirectTest {
+  override def extraSettings: String = "-usejavacp -Ystop-after:typer"
+
+  def code = "class C { A.f }"
+
+  def show(): Unit = try {
+    compile()
+    throw new Error("Expected OutOfMemoryError")
```
olafurpg, line 10: Thank you for setting up the test!

olafurpg (approved):
LGTM 👍

PR #6345, opened 2018-02-22
Also augment assertion errors raised in reflect.

hrhino:
And improve the augmentation.

`Global` shadows `Predef`'s assertion methods to give them information about what the compiler was doing when the assertion failed. However, inside the reflect cake layer (`SymbolTable`) these methods weren't visible, so calls to `assert` in reflect went through `Predef.assert` and didn't get augmented. Moving the shadowing methods to `SymbolTable` fixes that.

Also improved some things about the augmentation:
- `supplementErrorMessage` is wrapped in a `try`/`catch` block to avoid erroring while reporting an error. However, there were a few places where exceptions were easy to hit:
  - `sym` can be `null`; don't emit symbol details in that case
  - `lastSeenContext` is `null` in pre-namer phases; make it `NoContext` instead.
- `context_s` is wrapped in its own `try`/`catch` because the file I/O may fail. However, the tree's position's file is a `SourceFile`, so we can just look at its `content` to find the context. Add a `lines` method there to provide the lines context properly. This has the benefit of making sourcefile context work better in the repl.

Thanks to retronym for the idea to use `throwAssertionError` to make the best use of inlining the assertion methods.

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -261,27 +261,6 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
 
 // ------------------ Debugging -------------------------------------
 
-  // Getting in front of Predef's asserts to supplement with more info.
-  // This has the happy side effect of masking the one argument forms
-  // of assert and require (but for now I've reproduced them here,
-  // because there are a million to fix.)
-  @inline final def assert(assertion: Boolean, message: => Any) {
-    // calling Predef.assert would send a freshly allocated closure wrapping the one received as argument.
-    if (!assertion)
-      throw new java.lang.AssertionError("assertion failed: "+ supplementErrorMessage("" + message))
-  }
-  @inline final def assert(assertion: Boolean) {
-    assert(assertion, "")
-  }
-  @inline final def require(requirement: Boolean, message: => Any) {
-    // calling Predef.require would send a freshly allocated closure wrapping the one received as argument.
-    if (!requirement)
-      throw new IllegalArgumentException("requirement failed: "+ supplementErrorMessage("" + message))
-  }
-  @inline final def require(requirement: Boolean) {
-    require(requirement, "")
-  }
-
   @inline final def ifDebug(body: => Unit) {
     if (settings.debug)
       body
@@ -966,7 +945,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
   /** Let's share a lot more about why we crash all over the place.
    *  People will be very grateful.
    */
-  protected var lastSeenContext: analyzer.Context = null
+  protected var lastSeenContext: analyzer.Context = analyzer.NoContext
 
   /** The currently active run
    */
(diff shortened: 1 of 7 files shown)


hrhino:
```
hhoughton:sandbox hhoughton$ cat > t10548.scala
trait T extends (Any{})
hhoughton:sandbox hhoughton$ scalac-launch 2.12.4 t10548.scala
error: java.lang.AssertionError: assertion failed: transformCaseApply: name = Any! tree = Any / class scala.reflect.internal.Trees$Ident
	at scala.tools.nsc.typechecker.RefChecks$RefCheckTransformer.transform(RefChecks.scala:1704)
	at scala.tools.nsc.typechecker.RefChecks$RefCheckTransformer.transform(RefChecks.scala:107)
	at scala.reflect.api.Trees$Transformer.$anonfun$transformTrees$1(Trees.scala:2559)
	at scala.reflect.api.Trees$Transformer.transformTrees(Trees.scala:2559)
	at scala.reflect.internal.Trees.itransform(Trees.scala:1416)
	at scala.reflect.internal.Trees.itransform$(Trees.scala:1348)
	at scala.reflect.internal.SymbolTable.itransform(SymbolTable.scala:16)
        ...
hhoughton:sandbox hhoughton$ ../build/quick/bin/scalac t10548.scala
error: java.lang.AssertionError: assertion failed:
  transformCaseApply: name = Any! tree = Any / class scala.reflect.internal.Trees$Ident
     while compiling: t10548.scala
        during phase: refchecks
     library version: version 2.12.5-20180221-215140-f821fe3
    compiler version: version 2.12.5-20180221-215140-f821fe3
  reconstructed args:

  last tree to typer: CompoundTypeTree
       tree position: line 1 of t10548.scala
            tree tpe: Any
              symbol: null
           call site: trait T in package <empty>

     1 trait T extends (Any{})
     2
	at scala.reflect.internal.SymbolTable.throwAssertionError(SymbolTable.scala:157)
	at scala.reflect.internal.SymbolTable.assert(SymbolTable.scala:140)
	at scala.tools.nsc.typechecker.RefChecks$RefCheckTransformer.transform(RefChecks.scala:1704)
	at scala.tools.nsc.typechecker.RefChecks$RefCheckTransformer.transform(RefChecks.scala:107)
	at scala.reflect.api.Trees$Transformer.$anonfun$transformTrees$1(Trees.scala:2559)
	at scala.collection.immutable.List.loop$1(List.scala:176)
        ...
```

This could be better by tracking the current tree in the transformer during non-analyzer phases rather than the last tree we gave to a typer, but it's still better.

src/reflect/scala/reflect/internal/util/SourceFile.scala
```diff
@@ -187,6 +190,12 @@ class BatchSourceFile(val file : AbstractFile, content0: Array[Char]) extends So
     lastLine
   }
 
+  override def lines: Iterator[String] =
```
mkeskells: for this use case line(i:Int) would be more efficient
or lines(start, end) 

I would be concerned leaving this in the common API as it as it with churn memory, e.g. in the way it is used.

src/compiler/scala/tools/nsc/Global.scala
```diff
@@ -966,7 +945,7 @@ class Global(var currentSettings: Settings, var reporter: Reporter)
   /** Let's share a lot more about why we crash all over the place.
    *  People will be very grateful.
    */
-  protected var lastSeenContext: analyzer.Context = null
+  protected var lastSeenContext: analyzer.Context = analyzer.NoContext
```
lrytz, line 948: https://github.com/scala/scala/blob/v2.12.4/src/compiler/scala/tools/reflect/ToolBoxFactory.scala#L63 should be updated too

src/reflect/scala/reflect/internal/util/SourceFile.scala
lrytz: Agree, let's add `(start: Int = 0, end: Int = lineIndices.length)` parameters

lrytz (dismissed):

src/compiler/scala/tools/nsc/Global.scala
hrhino: Good catch.

src/reflect/scala/reflect/internal/util/SourceFile.scala
hrhino: Do you mean returning a `String` or an `Iterator[String]`? I can't imagine (I'll check) that `lines(a, b).mkString` in the latter case would be too different than `(lines take a drop b).mkString`; the iterator is lazy so it will only make strings for the given lines. I'll change it to the former; we only need it in this context anyways.

lrytz: Dropped elements are still evaluated:
```
scala> (0 to 10).iterator.map(x => {println(x); x}).drop(3).take(3).toList
0
1
2
3
4
5
res0: List[Int] = List(3, 4, 5)
```

hrhino: ... of course. Sorry, it's early here :P

lrytz: ☕️  🥐

lrytz (approved):
Thank you!

SethTisue:
@hrhino in the 2.12 community build I'm seeing, e.g. in kind-projector:

```
[kind-projector] [error] Unspecified value parameter message.
[kind-projector] [error]       require(i >= 0)
[kind-projector] [error]              ^
```

was it intentional that you removed the one-argument overload of `require`?

hrhino:
It was; apologies. I should have made that clearer in the commit message for the gracious reviewers. It was completely unused in the compiler and I didn't consider that plugins may be using it. Paul's comment inspired me to encourage the use of assertion messages.

I see you've [made a few PRs](https://github.com/lightbend/genjavadoc/pull/130) in that direction. I'm willing to do the dirty work if you want.

Sorry about that.

hrhino:
Actually, I just checked Jenkins and noticed that this breaks scala.js and possibly singleton-ops too. Would you rather I just put it back?

SethTisue:
@hrhino I think it's better to put it back for 2.12.x, yeah. (perhaps deprecated? no strong feeling either way.)

we are free to change the compiler API in minor Scala versions without notice, and compiler plugins are expected to cross-compile against specific minor versions. regardless, it's kinder to delay such changes until the next major version unless the delay is really costing us.

hrhino:
That's fair. scala/scala#6414 does exactly that.

PR #6346, opened 2018-02-23
8 threads

lrytz:
PR just for running the benchmarks


Ichoran:
Why do you want eight threads running benchmarks?  That seems counter to normal advice about benchmarking (i.e. "make sure your system isn't doing anything else" not "make sure you have seven other computationally intensive things going on at the same time")?  Maybe I'm misunderstanding something.

SethTisue:
@Ichoran see #6124

PR #6347, opened 2018-02-24
Introduce Loan object, a Scala implementation of try-with-resources

varming:
The try-with-resources construction in Java (introduced in Java 7) is too useful to not have a counterpart in the Scala standard library (IMHO). WDYT? Are you up for adding something like Loan?

I also added a method `orValue` to `Try`. You can see the usefulness of that method in my implementation of `Loan`.

I wasn't sure how you handle Java classes leaking into the Scala library. Is it permissible to use java.lang.AutoCloseable?

### DIFF
--- a/src/library/scala/util/Loan.scala
+++ b/src/library/scala/util/Loan.scala
@@ -1,0 +1,112 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2008-2018, LAMP/EPFL             **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+package scala
+package util
+
+/**
+ * The `Loan`s apply method implements Java's try-with-resources. You can loan
+ * out a resource to a function `f` and the resource will be closed when `f`
+ * returns regardless of the way `f` returns.
+ *
+ * See https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3
+ *
+ * In the following example `f` is applied to an input stream if it is
+ * non-null. If the input stream is null, then the RuntimeException is thrown.
+ * If `f` throws an exception, then the input stream is closed and the
+ * exception is rethrown.  If furthermore, closing the input stream throw an
+ * exception then that exception is suppressed on the exception thrown by `f`.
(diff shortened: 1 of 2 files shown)


hvesalai:
There are may other names for this construct. For example: `using` and `cleanly`. I find `loan` not to be the best name for this.

SBT io defines https://github.com/sbt/io/blob/1.x/io/src/main/scala/sbt/io/Using.scala

I find that name to be much more descriptive.

lrytz:
We should consider extending the native `try` to support resources. It would avoid the allocation / indirection of having it in a library. The change could go along with an equivalent update to `scala.util.Try` to support resources too.

varming:
@lrytz adding try-with-resource to the core language would be great. The program transformation in https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3 shouldn't be hard to implement in the compiler.

That said, I never worried about the allocation of the `LoanX` objects in the code in this PR. The cost of those allocations is tiny compared to dealing with resources (native calls, system calls, file handle allocations, etc).

SethTisue:
I would suggest posting on https://contributors.scala-lang.org to call attention to this PR. We may end up merging some form of this, but it needs community discussion. (Posting on Discourse will reach people who don't watch this repo.)

nafg:
> I wasn't sure how you handle Java classes leaking into the Scala library. Is it permissible to use java.lang.AutoCloseable?

The scala library depends on Java classes in many places and ways. The strategy so far has been to cross-compile Scala ports of the Java classes for Scala.is and scala-native.

SethTisue:
https://contributors.scala-lang.org/t/adding-try-with-resource-to-scala/1641

ocadaruma:
What do you think about implementation by type class ?
For example: https://gist.github.com/piyo7/a82da2bcb5ede5a53d6d93354d7cea6f

hvesalai:
@ocadaruma Closer-pattern would allow the scala implementation to be independent of AutoCloseable. Ofcourse an implementation of a `Closer[AutoCloseable]` would be provided and the type bounds should be adjusted accordingly

adriaanm:
Heads up: if this is going to make it into 2.13, it will need to be ready to merge by M5 (due Aug 10).

SethTisue:
@varming I'm going to close this for inactivity, since we are trying to hard to cut down on the number of open PRs scala/scala. (closing the PR for now doesn't mean anything about its eventual chances of success.)

SethTisue:
the action's at https://github.com/scala/scala/pull/6907 now

PR #2861, opened 2013-08-21
Deprecate early type defs

densh:
This is rehash of #2856 

Review by @adriaanm

PR #6349, opened 2018-02-26
Remove early type members under -Xsource:2.14

eed3si9n:
Ref https://github.com/scala/bug/issues/3572
> I shudder to think of the possible consequences this could have for type soundness.
>
> We agreed to deprecate them instead.

Ref https://github.com/scala/scala/pull/2861


adriaanm (approved):

adriaanm:
We should do the same for early initializers (even though there won't be a replacement until trait params land in 2.14...)

PR #6350, opened 2018-02-26
Fix error message for overriding type alias

### LINKED ISSUE #6001: Confusing first error message that overriding type members is possible
When trying to override a concrete type member without a override keyword the compiler returns an error saying that the override keyword must be used.

```
scala> :paste

class MyString
trait A { type Hello}
class AClass extends A{ type Hello = MyString}

defined class MyString
defined trait A
defined class AClass


scala> class B extends AClass{ type Hello = MyString with Serializable}
<console>:13: error: overriding type Hello in class AClass, which equals MyStrin
g;
 type Hello needs `override' modifier
       class B extends AClass{ type Hello = MyString with Serializable}
                                    ^
```

But after applying the overriding keyword the compiler says that the type member is incompatible. It could already be said before.
Maybe also a more unambiguous message like "Overriding concrete type members is not allowed".
"type Hello has incompatible type" suggests too much that the error is in this case the relation between MyClass and MyClass with Serializable.
                                   
```
scala> class B extends AClass{ override type Hello =MyString with Serializable}
<console>:13: error: overriding type Hello in class AClass, which equals MyStrin
g;
 type Hello has incompatible type
       class B extends AClass{ override type Hello =MyString with Serializable}
                                           ^
```

eed3si9n:
Fixes https://github.com/scala/bug/issues/6001

This fixes the error message for overriding a concrete type alias.

### before

```scala
override-concrete-type.scala:7: error: overriding type A1 in class Foo, which equals Something;
 type A1 has incompatible type
class Bar extends Foo { override type A1 = Something with Serializable }
                                      ^
one error found
```

### after

```scala
override-concrete-type.scala:7: error: overriding type A1 in class Foo, which equals Something;
 overriding concrete type alias type A1 is not allowed except when equivalent
class Bar extends Foo { override type A1 = Something with Serializable }
                                      ^
one error found
```

This error message was suggested in https://github.com/scala/bug/issues/6001#issuecomment-292407825.

### DIFF
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -336,7 +336,7 @@ abstract class RefChecks extends Transform {
           analyzer.foundReqMsg(pair.lowClassBound, pair.highClassBound)
         )
 
-        val indent = " "
+        val indent = "  "
         def overrideErrorMsg(msg: String): String = {
           val isConcreteOverAbstract =
             (other.owner isSubClass member.owner) && other.isDeferred && !member.isDeferred
--- a/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
@@ -336,20 +336,19 @@ abstract class RefChecks extends Transform {
           analyzer.foundReqMsg(pair.lowClassBound, pair.highClassBound)
         )
 
+        val indent = " "
         def overrideErrorMsg(msg: String): String = {
           val isConcreteOverAbstract =
             (other.owner isSubClass member.owner) && other.isDeferred && !member.isDeferred
           val addendum =
             if (isConcreteOverAbstract)
-              ";\n (Note that %s is abstract,\n  and is therefore overridden by concrete %s)".format(
-                infoStringWithLocation(other),
-                infoStringWithLocation(member)
-              )
+              s";\n${indent}(note that ${infoStringWithLocation(other)} is abstract,\n" +
+              s"${indent}and is therefore overridden by concrete ${infoStringWithLocation(member)})"
             else if (settings.debug)
               analyzer.foundReqMsg(member.tpe, other.tpe)
             else ""
 
-          s"overriding ${infoStringWithLocation(other)};\n ${infoString(member)} $msg$addendum"
+          s"overriding ${infoStringWithLocation(other)};\n$indent$msg$addendum"
         }
         def emitOverrideError(fullmsg: String) {
           if (member.owner == clazz) reporter.error(member.pos, fullmsg)
@@ -358,14 +357,15 @@ abstract class RefChecks extends Transform {
 
         def overrideError(msg: String) {
           if (noErrorType)
-            emitOverrideError(overrideErrorMsg(msg))
+            emitOverrideError(overrideErrorMsg(infoString(member) + " " + msg))
         }
 
         def overrideTypeError() {
           if (noErrorType) {
             emitOverrideError(
               if (member.isModule && other.isModule) objectOverrideErrorMsg
-              else overrideErrorMsg("has incompatible type")
+              else if (!other.isDeferred && other.isAliasType) overrideErrorMsg(s"overriding concrete type alias ${infoString(member)} is not allowed except when equivalent")
+              else overrideErrorMsg(s"${infoString(member)} has incompatible type")
             )
           }
         }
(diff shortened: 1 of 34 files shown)


test/files/neg/override-concrete-type.check
```diff
@@ -0,0 +1,5 @@
+override-concrete-type.scala:7: error: overriding type A1 in class Foo, which equals Something;
+ overriding concrete type alias type A1 is not allowed except when equivalent
```
som-snytt: Is one-space indentation intentional? I can't say that ten times fast.

som-snytt: I see there is an existing one-space indent.

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -358,14 +358,15 @@ abstract class RefChecks extends Transform {
...
+            emitOverrideError(overrideErrorMsg(infoString(member) + " " + msg))
         }
 
         def overrideTypeError() {
           if (noErrorType) {
             emitOverrideError(
               if (member.isModule && other.isModule) objectOverrideErrorMsg
-              else overrideErrorMsg("has incompatible type")
+              else if (!other.isDeferred) overrideErrorMsg(s"overriding concrete type alias ${infoString(member)} is not allowed except when equivalent")
+              else overrideErrorMsg(s"${infoString(member)} has incompatible type")
```
som-snytt: My limited ability to compose strings in my head suggests we could use two test cases in the test, for each branch.

At least I can do code coverage in my head.

som-snytt:
The use case I know about, where the knot must be tied:

```
scala> trait A { type X = String }
defined trait A

scala> trait B { type X = String }
defined trait B

scala> class C extends A with B
<console>:13: error: class C inherits conflicting members:
  type X in trait A, which equals String  and
  type X in trait B, which equals String
(Note: this can be resolved by declaring an override in class C.)
       class C extends A with B
             ^

scala> class C extends A with B { type X = String }
<console>:13: error: overriding type X in trait B, which equals String;
 type X needs `override' modifier
       class C extends A with B { type X = String }
                                       ^

scala> class C extends A with B { override type X = String }
defined class C
```

It currently does one-space indent, but maybe it should be two?

It will say "this can be resolved" even if it cannot?

test/files/neg/override-concrete-type.check
eed3si9n: I am happy to change this to two space indent.

eed3si9n:
@som-snytt 

> It will say "this can be resolved" even if it cannot?

I am guessing that it's saying that you need an override with the `override` modifier, in 085a3c9 I've changed the message to:

```
t5358.scala:3: error: class C inherits conflicting members:
  method hi in trait A of type => String and
  method hi in trait B of type => String
  (note: this can be resolved by declaring an `override' in class C.)
class C extends A with B
      ^
```

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
```diff
@@ -336,20 +336,21 @@ abstract class RefChecks extends Transform {
...
         )
 
+        val indent = "  "
         def overrideErrorMsg(msg: String): String = {
           val isConcreteOverAbstract =
             (other.owner isSubClass member.owner) && other.isDeferred && !member.isDeferred
           val addendum =
             if (isConcreteOverAbstract)
-              ";\n (Note that %s is abstract,\n  and is therefore overridden by concrete %s)".format(
+              ";\n${indent}(note that %s is abstract,\n${indent}and is therefore overridden by concrete %s)".format(
```
som-snytt: I'm sorry `-Xlint` is not turned on for the build to catch the missing interpolator. Probably a good moment to get rid of the format call, since it's not doing much formatting, so `s""` is good enough. But if you choose `f`, the newline should be `f"%n"` for platform line separator, sigh.

test/files/neg/t1364.check
```diff
@@ -1,5 +1,5 @@
 t1364.scala:9: error: overriding type T in trait A with bounds <: AnyRef{type S[-U]};
- type T has incompatible type
+  type T has incompatible type
  type T = { type S[U] = U }
```
som-snytt, line 3: The weird indentation here is because the test source is weirdly indented.

som-snytt:
Also run/StubErrorTypeDef.scala needs update-check.

src/compiler/scala/tools/nsc/typechecker/RefChecks.scala
eed3si9n: Interesting. I didn't catch that locally because neg/t5358.scala passed.
I'll fix this.

som-snytt:
Just needs squashing for history hygiene.

I had a girlfriend, Jeanne, who besides the joke on "hi, Jeanne," whenever I said, "I love you, Jeanne," she'd say, "Who's Eugene?"

som-snytt (approved):
I don't know if the person doing a merge likes to squash-and-merge or prefers pre-squashed. I won't attempt S&M today. @SethTisue

eed3si9n:
> I had a girlfriend, Jeanne, who besides the joke on "hi, Jeanne," whenever I said, "I love you, Jeanne," she'd say, "Who's Eugene?"

That was my college life, as I was only given the first syllable at birth.

> I don't know if the person doing a merge likes to squash-and-merge or prefers pre-squashed.

I am guessing the ideal would be split the commit into fix related things and indentation space=2 changes, and make sure both would pass the test.

SethTisue:
thank you, Gene!

PR #6255, opened 2018-01-09
remove Ant support, remove Eclipse support

SethTisue:
remove Ant tasks for building user code using Ant.  this is
now historical. hard to be sure absolutely nobody is using it anymore,
but regardless, it ought to be a separate project, if anyone still
wants it to exist in 2.13+ world.

support for building Scala itself using Eclipse is removed and
replaced with a note in the readme that says "Support for Eclipse has
decayed and been removed, but could be resurrected by a volunteer"

PR #6354, opened 2018-02-28
Minor sbt build improvements

szeiger:
- Remove old `.desired.sha1` files that were only used by the ant build
- Remove `.desired.sha1` files that are used for partest and add the dependencies directly to `build.sbt`
- Don't build unnecessary subprojects for stability tests


SethTisue:
@szeiger this overlaps a bit with #6255

lrytz (approved):

PR #6355, opened 2018-02-28
Explicit type annotations and !isEmpty replace with nonEmpty

sh0hei:


SethTisue (approved):

adriaanm:
I think this is good for inclusion in 2.12.6

src/compiler/scala/tools/nsc/ast/parser/xml/Utility.scala
```diff
@@ -77,7 +77,7 @@ object Utility {
       else sb append c
     }
 
-    if(!sb.isEmpty) // flush buffer
+    if(sb.nonEmpty) // flush buffer
```
som-snytt, line 80: oh man, missed opportunity to normalize spacing before the LPAREN. I promise to contribute more reviews on lunch breaks. Or maybe on snack breaks. Also, in a demoralizing blow, retronym recently came out against `nonEmpty`, in some context. Or is this scala StringBuilder with native nonEmpty, and why do I have to ask?

adriaanm: Good question -- I didn't think to check, but we get `TraversableOnce's` version here: `def nonEmpty: Boolean = !isEmpty` (according to IntelliJ). So, neither `isEmpty` nor `nonEmpty` uses the potentially more efficient `length` on the underlying Java `StringBuilder`.

PR #6358, opened 2018-03-01
Merge 2.12 again

lrytz:
To include https://github.com/scala/scala/pull/6354, then we can rebase https://github.com/scala/scala/pull/6255.


lrytz:
It's greener than it looks!

PR #6359, opened 2018-03-01
Report an error when using an unstable qualifier in a type selection

lrytz:
The `treeInfo.admitsTypeSelection` check got lost in the refactoring
in 0b055c6cf697.

In 2.12.4, `typedType` of `newOuter.Inner` produces `Outer#Inner`:

    scala> val x: newOuter.Inner = null
    x: Outer#Inner = null

Fixes scala/bug#10619

### DIFF
--- a/test/files/neg/t10619.check
+++ b/test/files/neg/t10619.check
@@ -1,0 +1,10 @@
+t10619.scala:4: error: stable identifier required, but Test.this.newOuter found.
+  val a: newOuter.Inner = { val o = newOuter; new o.Inner }
+         ^
+t10619.scala:5: error: stable identifier required, but Test.this.newOuter found.
+  val b: newOuter.Inner = a
+         ^
+t10619.scala:12: error: stable identifier required, but Test.this.newOuter found.
+  val f = new newOuter.Inner
+              ^
+three errors found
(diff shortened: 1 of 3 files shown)


test/files/neg/t10619.check
```diff
@@ -0,0 +1,18 @@
+t10619.scala:4: error: stable identifier required, but Test.this.newOuter found.
+  val a: newOuter.Inner = { val o = newOuter; new o.Inner }
+         ^
+t10619.scala:5: error: stable identifier required, but Test.this.newOuter found.
+  val b: newOuter.Inner = a
+         ^
+t10619.scala:8: error: type mismatch;
```
lrytz: This error doesn't show in 2.12.3, we should probably try to get rid of it.

test/files/neg/t10619.check
```diff
@@ -0,0 +1,18 @@
...
+ found   : Outer#Inner
+ required: Test.o.Inner
+  val c: o.Inner = b
+                   ^
+t10619.scala:12: error: stable identifier required, but Test.this.newOuter found.
+  val f = new newOuter.Inner
+              ^
+t10619.scala:12: error: Outer is not a legal prefix for a constructor
```
lrytz: Same here, this one doesn't show in 2.12.3, and it's confusing.

adriaanm (approved):
My bad! Thanks.

PR #6360, opened 2018-03-01
Forward more methods for reverse orderings

NthPortal:
Forward most comparison methods (not just abstract
ones) in default implementations of `Ordering.reverse`
and `PartialOrdering.reverse`.

This may yield a small performance benefit if comparison
methods are overridden in the outer `Ordering`, and shouldn't
hurt performance if they are not.

This change may also be necessary depending how
scala/bug#10511 is resolved.

### DIFF
--- a/src/library/scala/math/Ordering.scala
+++ b/src/library/scala/math/Ordering.scala
@@ -109,6 +109,13 @@ trait Ordering[T] extends Comparator[T] with PartialOrdering[T] with Serializabl
   override def reverse: Ordering[T] = new Ordering[T] {
     override def reverse = outer
     def compare(x: T, y: T) = outer.compare(y, x)
+    override def lteq(x: T, y: T) = outer.lteq(y, x)
+    override def gteq(x: T, y: T) = outer.gteq(y, x)
+    override def lt(x: T, y: T) = outer.lt(y, x)
+    override def gt(x: T, y: T) = outer.gt(y, x)
+    override def equiv(x: T, y: T) = outer.equiv(y, x)
+    override def max(x: T, y: T) = outer.min(x, y)
+    override def min(x: T, y: T) = outer.max(x, y)
   }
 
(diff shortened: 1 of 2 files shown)


NthPortal:
review @Ichoran

SethTisue:
ping @Ichoran

NthPortal:
I originally expected/intended this to be merged before #6410 was even fully written; now that it *is* complete however, I'm not sure if it's better to have this as its own PR, or to just have it as a separate commit in #6410.

As its own PR, it's not waiting on #6468, so it an be merged now potentially ¯\\\_(ツ)_/¯

NthPortal:
Also, #6410 is a superset of this PR's changes, and has already been reviewed, so reviewing this PR should be trivial.

szeiger:
What's the hold-up on https://github.com/scala/scala/pull/6410 since it's green and approved? Should we merge this one for M4 instead?

SethTisue:
I have misgivings about #6410. I'll comment there.

NthPortal:
@szeiger I'm happy to get this merged into M4.

The hold-up on #6410 is that I'm waiting on #6468 to get merged for two reasons:

- I don't want to make more work for Eugene by requiring them to rebase and fix #6468 to work on top of #6410. It's not a ton of work, but still
- rebasing on top of #6468 reduces the diff size of #6410 

Neither of those reasons is *super* strong to me, and if you want to merge #6410 sooner rather than later, I'm happy to squash and rebase it now so it can be merged.

NthPortal:
I actually would like to have this commit for #6576 as well (otherwise it will create merge conflicts) - are you amenable to merging this now @szeiger?

NthPortal:
I don't know why it says there's no corresponding job on jenkins - the link still goes to the build, which passed

szeiger (approved):

PR #6361, opened 2018-03-02
A pair of developer convenience enhancements.

hrhino:
In case anyone else wants them.

- `typed` in power mode had an odd signature, wherein it assumed that the type of a typed tree is the same as the type of the tree that was typed, which isn't necessarily true. Never again will I type `typed(tq"Int":Tree)` to avoid a `ClassCastException`.

- `Global#printTypings` was a `var`. I can't tell why; it started out as a `var`, became a `val`, re-became a `var`, was mutated in `withSuppressedSettings` to work around the fact that setting its setting didn't set it itself, and now has become a `def` like you probably expected. Now, `:settings -Ytyper-debug` turns the verbosity on, and `:settings -Ytyper-debug:false` turns it off.

I doubt the latter change is going to cause a performance impact, but if it does it's obviously not worth the benefit to my convenience. If you'd like me to run one I will, otherwise I'm inclined to let the mergely bencher bench on merge.

Also, I [read around](https://github.com/scala/bug/issues/9026#issuecomment-363590768) that maybe keeping copyright years up to date is a good thing. 2013 was a pretty okay year but we're in the brave new world now, so unless people object I'm going to start swapping `3` for `8` on files I touch.

### DIFF
--- a/src/compiler/scala/tools/nsc/Global.scala
+++ b/src/compiler/scala/tools/nsc/Global.scala
@@ -1,5 +1,5 @@
 /* NSC -- new Scala compiler
- * Copyright 2005-2013 LAMP/EPFL
+ * Copyright 2005-2018 LAMP/EPFL
  * @author  Martin Odersky
  */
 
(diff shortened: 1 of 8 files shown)


SethTisue:
@hrhino is this one you plan to revive...?

hrhino:
The performance implications concern me. I'd like to get the `-Xdev` PR in for a baseline first, but yes.

SethTisue:
(#7174, for those keeping score)

PR #6362, opened 2018-03-02
JEP-238 (Multi-Release JAR) and JEP-247 (Compiler for older JDKs) support

retronym:
Introduces a `-release N` compiler flag which accepts a JDK major version `N` (8, 9, 10). This flag only has an effect when the compiler runs on a JDK >= 9.

Specifying `-release` has two effects:
  - For multi-release JARs ([JEP-238](http://openjdk.java.net/jeps/238)) on the classpath, the visible classfiles correspond to the selected JDK version
  - The JDK on the classpath corresponds to the selected JDK version ([JEP-247](http://openjdk.java.net/jeps/247)).

Technical detail: JDKs >= 9 ship with a `ct.sym` file, which is a zip file containing stripped-down classfiles with no code, only signatures, renamed with a `.sig` file extension. The folder structure within `ct.sym` allows selecting the `sig` files for a specific JDK major version.

This option is currently incompatible with the `scaladoc` tool.

### DIFF
--- a/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
+++ b/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
@@ -14,6 +14,9 @@ import scala.reflect.io.{AbstractFile, PlainFile, PlainNioFile}
 import scala.tools.nsc.util.{ClassPath, ClassRepresentation}
 import FileUtils._
 import scala.collection.JavaConverters._
+import scala.collection.immutable
+import scala.reflect.internal.JDK9Reflectors
+import scala.tools.nsc.classpath.PackageNameUtils.{packageContains, separatePkgAndClassNames}
 
 /**
  * A trait allowing to look for classpath entries in directories. It provides common logic for
@@ -125,13 +128,35 @@ trait JFileDirectoryLookup[FileEntryType <: ClassRepresentation] extends Directo
 
 object JrtClassPath {
   import java.nio.file._, java.net.URI
-  def apply(): Option[ClassPath] = {
-    try {
-      val fs = FileSystems.getFileSystem(URI.create("jrt:/"))
-      Some(new JrtClassPath(fs))
-    } catch {
-      case _: ProviderNotFoundException | _: FileSystemNotFoundException =>
-        None
+  def apply(release: Option[String]): Option[ClassPath] = {
+    import scala.util.Properties._
+    if (!isJavaAtLeast("9")) None
+    else {
+      // TODO escalate errors once we're sure they are fatal
+      // I'm hesitant to do this immediately, because -release will still work for multi-release JARs
+      // even if we're running on a JRE or a non OpenJDK JDK where ct.sym is unavailable.
+      //
+      // Longer term we'd like an official API for this in the JDK
+      // Discussion: http://mail.openjdk.java.net/pipermail/compiler-dev/2018-March/thread.html#11738
+
+      val currentMajorVersion: Int = JDK9Reflectors.runtimeVersionMajor(JDK9Reflectors.runtimeVersion()).intValue()
+      release match {
+        case Some(v) if v.toInt < currentMajorVersion =>
+          try {
+            val ctSym = Paths.get(javaHome).resolve("lib").resolve("ct.sym")
+            if (Files.notExists(ctSym)) None
+            else Some(new CtSymClassPath(ctSym, v.toInt))
+          } catch {
+            case _: Throwable => None
+          }
+        case _ =>
+          try {
+            val fs = FileSystems.getFileSystem(URI.create("jrt:/"))
+            Some(new JrtClassPath(fs))
+          } catch {
+            case _: ProviderNotFoundException | _: FileSystemNotFoundException => None
+          }
+      }
     }
   }
 }
@@ -161,11 +186,7 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
   /** Empty string represents root package */
   override private[nsc] def hasPackage(pkg: String) = packageToModuleBases.contains(pkg)
   override private[nsc] def packages(inPackage: String): Seq[PackageEntry] = {
-    def matches(packageDottedName: String) =
-      if (packageDottedName.contains("."))
-        packageOf(packageDottedName) == inPackage
-      else inPackage == ""
-    packageToModuleBases.keysIterator.filter(matches).map(PackageEntryImpl(_)).toVector
+    packageToModuleBases.keysIterator.filter(pack => packageContains(inPackage, pack)).map(PackageEntryImpl(_)).toVector
   }
   private[nsc] def classes(inPackage: String): Seq[ClassFileEntry] = {
     if (inPackage == "") Nil
@@ -188,8 +209,8 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
   def findClassFile(className: String): Option[AbstractFile] = {
     if (!className.contains(".")) None
     else {
-      val inPackage = packageOf(className)
-      packageToModuleBases.getOrElse(inPackage, Nil).iterator.flatMap{x =>
+      val (inPackage, _) = separatePkgAndClassNames(className)
+      packageToModuleBases.getOrElse(inPackage, Nil).iterator.flatMap { x =>
         val file = x.resolve(className.replace('.', '/') + ".class")
         if (Files.exists(file)) new scala.reflect.io.PlainNioFile(file) :: Nil else Nil
       }.take(1).toList.headOption
@@ -197,6 +218,72 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
   }
   private def packageOf(dottedClassName: String): String =
     dottedClassName.substring(0, dottedClassName.lastIndexOf("."))
+}
+
+/**
+  * Implementation `ClassPath` based on the $JAVA_HOME/lib/ct.sym backing http://openjdk.java.net/jeps/247
+  */
+final class CtSymClassPath(ctSym: java.nio.file.Path, release: Int) extends ClassPath with NoSourcePaths {
+  import java.nio.file.Path, java.nio.file._
+  type F = Path
+  private val javaHome = System.getProperty("java.home")
+  private val javaSpecVersion = scala.util.Properties.javaSpecVersion
+  private val fileSystem: FileSystem = FileSystems.newFileSystem(ctSym, null)
+  private val root: Path = fileSystem.getRootDirectories.iterator().next
+  private val roots = Files.newDirectoryStream(root).iterator().asScala.toList
+
+  // http://mail.openjdk.java.net/pipermail/compiler-dev/2018-March/011737.html
+  private def codeFor(major: Int): String = if (major < 10) major.toString else ('A' + (major - 10)).toChar.toString
+
+  private val releaseCode: String = codeFor(release)
+  private def fileNameMatchesRelease(fileName: String) = !fileName.contains("-") && fileName.contains(releaseCode)
+  private val subset: List[Path] = roots.filter(root => fileNameMatchesRelease(root.getFileName.toString))
+
+  // e.g. "java.lang" -> Seq(/876/java/lang, /87/java/lang, /8/java/lang))
+  private val packageIndex: scala.collection.Map[String, Seq[Path]] = {
+    val index = collection.mutable.AnyRefMap[String, collection.mutable.ListBuffer[Path]]()
+    subset.foreach(root => Files.walk(root).iterator().asScala.filter(Files.isDirectory(_)).foreach{
+      p =>
+        if (p.getNameCount > 1) {
+          val p1 = if (scala.util.Properties.isJavaAtLeast("9")) p.subpath(1, p.getNameCount) else p
+          val packageDotted = p1.toString.replace('/', '.')
+          index.getOrElseUpdate(packageDotted, new collection.mutable.ListBuffer) += p
+        }
+    })
+    index
+  }
+
+  /** Empty string represents root package */
+  override private[nsc] def hasPackage(pkg: String) = packageIndex.contains(pkg)
+  override private[nsc] def packages(inPackage: String): Seq[PackageEntry] = {
+    packageIndex.keysIterator.filter(pack => packageContains(inPackage, pack)).map(PackageEntryImpl(_)).toVector
+  }
+  private[nsc] def classes(inPackage: String): Seq[ClassFileEntry] = {
+    if (inPackage == "") Nil
+    else {
+      packageIndex.getOrElse(inPackage, Nil).flatMap(x =>
+        Files.list(x).iterator().asScala.filter(_.getFileName.toString.endsWith(".sig"))).map(x =>
+        ClassFileEntryImpl(new PlainNioFile(x))).toVector
+    }
+  }
+
+  override private[nsc] def list(inPackage: String): ClassPathEntries =
+    if (inPackage == "") ClassPathEntries(packages(inPackage), Nil)
+    else ClassPathEntries(packages(inPackage), classes(inPackage))
+
+  def asURLs: Seq[URL] = Nil
+  def asClassPathStrings: Seq[String] = Nil
+
+  def findClassFile(className: String): Option[AbstractFile] = {
+    if (!className.contains(".")) None
+    else {
+      val (inPackage, classSimpleName) = separatePkgAndClassNames(className)
+      packageIndex.getOrElse(inPackage, Nil).iterator.flatMap{x =>
+        val file = x.resolve(classSimpleName + ".sig")
+        if (Files.exists(file)) new scala.reflect.io.PlainNioFile(file) :: Nil else Nil
+      }.take(1).toList.headOption
+    }
+  }
 }
 
 case class DirectoryClassPath(dir: File) extends JFileDirectoryLookup[ClassFileEntryImpl] with NoSourcePaths {
--- a/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
+++ b/src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
@@ -216,8 +216,6 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
       }.take(1).toList.headOption
     }
   }
-  private def packageOf(dottedClassName: String): String =
-    dottedClassName.substring(0, dottedClassName.lastIndexOf("."))
 }
 
 /**
@@ -225,9 +223,7 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
   */
 final class CtSymClassPath(ctSym: java.nio.file.Path, release: Int) extends ClassPath with NoSourcePaths {
   import java.nio.file.Path, java.nio.file._
-  type F = Path
-  private val javaHome = System.getProperty("java.home")
-  private val javaSpecVersion = scala.util.Properties.javaSpecVersion
+
   private val fileSystem: FileSystem = FileSystems.newFileSystem(ctSym, null)
   private val root: Path = fileSystem.getRootDirectories.iterator().next
   private val roots = Files.newDirectoryStream(root).iterator().asScala.toList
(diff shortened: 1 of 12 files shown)


retronym:
@lrytz I've optimistically assigned this to 2.12.5. I'll rebase and clean up ASAP. The Akka build is suffering some pain that the JEP-247 support would help.

ktoso:
It would be most excellent to have this in 2.12.5. We've jumped a bit ahead and are supporting JDK9 types in our build and making multi-release-jars since quite a while, hoping for Scala to provide this feature sooner rather than later. Nowadays our build has to include manual steps and juggling of class files, with this we'd be able to get back to a normal release mode -- same as anyone who's looking to build multi-release-jars, which perhaps is not all that common, but may become increasingly coming very soon.

Thanks a ton for the work @retronym and I hope it'd be possible to make this happen, fingers crossed :shipit:

retronym:
Our CI runs on Java 8 and the new tests will skip themselves.

I've run locally:

```
⚡ (java_use 9; sbt)
[info] Loading global plugins from /Users/jz/.sbt/0.13/plugins
[info] Compiling 2 Scala sources to /Users/jz/.sbt/0.13/plugins/target/scala-2.10/sbt-0.13/classes...
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by sbt.ivyint.ErrorMessageAuthenticator$ (file:/Users/jz/.sbt/boot/scala-2.10.7/org.scala-sbt/sbt/0.13.17/ivy-0.13.17.jar) to field java.net.Authenticator.theAuthenticator
WARNING: Please consider reporting this to the maintainers of sbt.ivyint.ErrorMessageAuthenticator$
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
...
[info] Test scala.tools.nsc.classpath.MultiReleaseJarTest.mrJar started
jar:file:/var/folders/b7/xcc2k0ln6ldcv247ffpy2d1w0000gp/T/mr-jar-test-5443454358858279065.jar
jar:file:/var/folders/b7/xcc2k0ln6ldcv247ffpy2d1w0000gp/T/mr-jar-test-9666552433935531747.jar
[info] Test scala.tools.nsc.classpath.MultiReleaseJarTest.ctSymTest started
[info] Test run finished: 0 failed, 0 ignored, 2 total, 3.077s
```

I'm also trying a manual CI run with Java 9 [here](https://scala-ci.typesafe.com/job/scala-2.12.x-validate-main/230/console)

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -79,6 +79,14 @@ trait ScalaSettings extends AbsScalaSettings
       domain  = languageFeatures
     )
   }
+  val release = StringSetting("-release", "<release>", "Compile for a specific JVM version. Supported targets: 6, 7, 8, 9", "").withPostSetHook { (value: StringSetting) =>
+    if (value.value != "" && !scala.util.Properties.isJavaAtLeast("9")) {
+      errorFn.apply("-release is only supported on Java 9 and higher")
```
retronym: - [ ] Typo: should be "--release"

src/reflect/scala/reflect/internal/JDK9Reflectors.java
```diff
@@ -0,0 +1,54 @@
+package scala.reflect.internal;
+
+import java.io.IOException;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.util.jar.JarFile;
+
+public final class JDK9Reflectors {
```
retronym, line 9: This pattern lets us compile on 8 but call APIs of 9 reflectively. The performance overhead of reflection is close to zero when calling through static final method handles.

retronym:
- [ ] Needs a more detailed PR description
 - [ ] Tests for the non-happy cases

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
```diff
@@ -79,6 +79,14 @@ trait ScalaSettings extends AbsScalaSettings
       domain  = languageFeatures
     )
   }
+  val release = StringSetting("-release", "<release>", "Compile for a specific JVM version. Supported targets: 6, 7, 8, 9", "").withPostSetHook { (value: StringSetting) =>
```
lrytz: Maybe `Compile with a specific version of the Java API on the classpath.`

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
lrytz: I think we should keep `-release`, as we don't use double dashes elsewhere. (for reference, `javac` 8 uses single dashes, but in 9+ there's a mix).

src/reflect/scala/reflect/internal/JDK9Reflectors.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final MethodHandle RUNTIME_VERSION_PARSE;
+    private static final MethodHandle NEW_JAR_FILE;
+
+    static {
+        RUNTIME_VERSION_PARSE = lookupRuntimeVersionParse();
+        NEW_JAR_FILE = lookupNewJarFile();
+    }
+
+    public static /*java.lang.Runtime.Version*/ Object lookupRuntimeVersionParse(String string) {
```
lrytz: Rename to `runtimeVersionParse`

lrytz: Also this method could be made private and invoked directly from `newJarFile`.

src/reflect/scala/reflect/io/ZipArchive.scala
```diff
@@ -117,15 +120,28 @@ abstract class ZipArchive(override val file: JFile) extends AbstractFile with Eq
...
+
+  protected def getDir(dirs: mutable.Map[String, DirEntry], entry: ZipEntry, entryName: String): DirEntry = {
+    if (entry.isDirectory) ensureDir(dirs, entryName, entry)
+    else ensureDir(dirs, dirName(entryName), null)
   }
 }
 /** ''Note:  This library is considered experimental and should not be used unless you know what you are doing.'' */
-final class FileZipArchive(file: JFile) extends ZipArchive(file) {
+final class FileZipArchive(file: JFile, release: Option[String]) extends ZipArchive(file, release) {
+  val x = getClass
```
lrytz: // dbg

and some more of those, i guess you're aware

src/reflect/scala/reflect/io/ZipArchive.scala
```diff
@@ -169,20 +186,30 @@ final class FileZipArchive(file: JFile) extends ZipArchive(file) {
...
-              new LeakyEntry(zipFile, zipEntry)
-
-          dir.entries(f.name) = f
+        if (!zipEntry.getName.startsWith("META-INF/versions/")) {
+          val zipEntryVersioned = if (release.isDefined) {
+            if (zipEntry.getName.contains("Versioned"))
+              getClass
+            zipFile.getEntry(zipEntry.getName)
+          } else zipEntry
+          val entryName = zipEntry.getName
```
lrytz: should it be `zipEntryVersioned.getName` here?

src/reflect/scala/reflect/io/ZipArchive.scala
```diff
@@ -169,20 +186,30 @@ final class FileZipArchive(file: JFile) extends ZipArchive(file) {
...
+          if (entryName != zipEntry.getName)
+            getClass
+          val dir = getDir(dirs, zipEntry, entryName)
+          if (zipEntry.isDirectory) dir
```
lrytz: the reference to `dir` seems to be dead code (already before this PR)

test/junit/scala/tools/nsc/classpath/MultiReleaseJarTest.scala
```diff
@@ -0,0 +1,88 @@
...
+
+    // TODO test fails if both Global runs look at the same JAR on disk. Caching problem in our classpath implementation?
+    // val temp2 = temp1
+    val temp2 = Files.createTempFile("mr-jar-test-", ".jar")
+
+    try {
+      def code(newApi: String) = s"package p1; abstract class Versioned { def oldApi: Int; $newApi }"
+
+      val oldC = compileToBytes(code("")).head._2
+      val newC = compileToBytes(code("def newApi: Int")).head._2
```
lrytz, line 30: nice that we can re-use that testing infra

test/junit/scala/tools/nsc/classpath/MultiReleaseJarTest.scala
```diff
@@ -0,0 +1,88 @@
...
+    val manifestBytes = os.toByteArray
+    manifestBytes
+  }
+  private def createZip(zipLocation: Path, content: List[(String, Array[Byte])]): Unit = {
+    val env = new java.util.HashMap[String, String]()
+    Files.deleteIfExists(zipLocation)
+    env.put("create", String.valueOf(true))
+    val fileUri = zipLocation.toUri
+    val zipUri = new java.net.URI("jar:" + fileUri.getScheme, fileUri.getPath, null)
+    println(zipUri)
```
lrytz: // dbg

src/compiler/scala/tools/nsc/classpath/ZipArchiveFileLookup.scala
```diff
@@ -18,13 +18,14 @@ import scala.tools.nsc.util.{ClassPath, ClassRepresentation}
  */
 trait ZipArchiveFileLookup[FileEntryType <: ClassRepresentation] extends ClassPath {
   val zipFile: File
+  val release: Option[String] = null
```
lrytz: I guess this is WIP, should it just be an abstract val? Then it needs to be defined (as `None` should be fine) in `ZipArchiveSourcePath`.

This (likely) caused the test failure
```
[test] !! 1867 - run/various-flat-classpath-types.scala    [non-zero exit code]
[test] ##### Log file '/home/jenkins/workspace/scala-2.12.x-validate-main/test/files/run/various-flat-classpath-types-run.log' from failed test #####
[test] 
[test] error: java.lang.NullPointerException
[test] 	at scala.reflect.io.FileZipArchive.x$2$lzycompute(ZipArchive.scala:190)
[test] 	at scala.reflect.io.FileZipArchive.x$2(ZipArchive.scala:180)
```

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -125,13 +126,31 @@ trait JFileDirectoryLookup[FileEntryType <: ClassRepresentation] extends Directo
...
+        val fs = FileSystems.getFileSystem(URI.create("jrt:/"))
+        Some(new JrtClassPath(fs))
+      } catch {
+        case _: ProviderNotFoundException | _: FileSystemNotFoundException =>
+          None
+      }
+    }
+    release match {
+      case None => jrt
+      case Some(v) if v == javaSpecVersion =>
```
lrytz: On Java < 9, `scala.util.Properties.javaSpecVersion` has a leading `1.`, while the `-release` flag asks for `6, 7, 8, 9, ...`

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -125,13 +126,31 @@ trait JFileDirectoryLookup[FileEntryType <: ClassRepresentation] extends Directo
...
+        if (isJavaAtLeast("9")) jrt else None
+      case Some(older) =>
+        if (isJavaAtLeast("9")) {
+          try {
+            Some(new CtSymClassPath(release))
+          } catch {
+            case _: Throwable =>
+              // TODO more fine grained error handling
+              None
```
lrytz: we probably want to report an error if we cannot satisfy what the users asks for using the `-release` flag

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
     dottedClassName.substring(0, dottedClassName.lastIndexOf("."))
 }
 
+/**
+  * Implementation `ClassPath` based on the $JAVA_HOME/lib/ct.sym backing http://openjdk.java.net/jeps/247
+  */
+final class CtSymClassPath(release: Option[String]) extends ClassPath with NoSourcePaths {
```
lrytz: `release` doesn't need to be an `Option`, when `CtSymClassPath` is created we know it's defined (or maye it's practical for testing?)

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
...
+  import java.nio.file.Path, java.nio.file._
+  type F = Path
+  private val javaHome = System.getProperty("java.home")
+  private val file = Paths.get(javaHome).resolve("lib").resolve("ct.sym")
```
lrytz: Should it fail early if the file doesn't exist? It seems only JDKs provide it (can the scala compiler run on a JRE? Probably yes?).

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
...
+  private val javaSpecVersion = scala.util.Properties.javaSpecVersion
+  private val fileSystem: FileSystem = FileSystems.newFileSystem(file, null)
+  private val root: Path = fileSystem.getRootDirectories.iterator().next
+  private val roots = Files.newDirectoryStream(root).iterator().asScala.toList
+  private val effectiveRelease = release.getOrElse(javaSpecVersion)
+  private val subset: List[Path] = roots.filter(_.getFileName.toString.contains(effectiveRelease))
```
lrytz: Assuming they keep their naming scheme, this might break at some point, for example `876910111213` contains `1` or `21`.. it's not a very robust scheme - is it described somewhere?

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
...
+
+  // e.g. "java.lang" -> Seq("/modules/java.base")
```
lrytz: This comment should be expanded. The v9 `ct.sym` contains files like

```
     4023  02-27-2018 02:14   76/java/lang/StringBuffer.sig
     4023  02-27-2018 02:14   8/java/lang/StringBuffer.sig
     4414  02-27-2018 02:14   9/java/lang/StringBuffer.sig
      158  02-27-2018 02:14   8769/java/lang/Cloneable.sig
     3133  02-27-2018 02:14   9-modules/java.base/module-info.sig
      247  02-27-2018 02:14   9-modules/java.sql/module-info.sig
```

So IIUC the map for this part would be

```
java.lang -> Seq(9/java/lang, 8769/java/lang)
java.base -> Seq(9-modules/java.base)
java.sql -> Seq(9-modules/java.sql)
```

Does that look as intended?

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
...
+        Files.list(x).iterator().asScala.filter(_.getFileName.toString.endsWith(".sig"))).map(x =>
+        ClassFileEntryImpl(new PlainNioFile(x))).toVector
+    }
+  }
+
+  override private[nsc] def list(inPackage: String): ClassPathEntries =
+    if (inPackage == "") ClassPathEntries(packages(inPackage), Nil)
+    else ClassPathEntries(packages(inPackage), classes(inPackage))
+
+  def asURLs: Seq[URL] = ???
```
lrytz: it seems this method is used in a few places..

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
...
+  def asClassPathStrings: Seq[String] = Nil
+
+  def findClassFile(className: String): Option[AbstractFile] = {
+    if (!className.contains(".")) None
+    else {
+      val inPackage = packageOf(className)
+      val classSimpleName = if (inPackage == "") className else className.substring(inPackage.length + 1)
```
lrytz: given the `if` 3 lines above, the one here should alwyas be false

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
```diff
@@ -199,6 +218,75 @@ final class JrtClassPath(fs: java.nio.file.FileSystem) extends ClassPath with No
...
+      packageIndex.getOrElse(inPackage, Nil).iterator.flatMap{x =>
+        val file = x.resolve(classSimpleName + ".sig")
+        if (Files.exists(file)) new scala.reflect.io.PlainNioFile(file) :: Nil else Nil
+      }.take(1).toList.headOption
+    }
+  }
+  private def packageOf(dottedClassName: String): String =
+    dottedClassName.substring(0, dottedClassName.lastIndexOf("."))
```
lrytz: code duplication with the one in `JrtClassPath`

lrytz (commented):
I think we can still get this in, it's a new flag, worst case there are bugs with it, but it doesn't affect existing features.

Ideas about the binary compatibility issues..?

```
scala-reflect: found 4 potential binary incompatibilities while checking against org.scala-lang:scala-reflect:2.12.0  (filtered 307)

method getDir(scala.collection.mutable.Map,java.util.zip.ZipEntry,java.lang.String)scala.reflect.io.ZipArchive#DirEntry in class scala.reflect.io.ZipArchive does not have a correspondent in other version
filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.io.ZipArchive.getDir")

method this(java.io.File,scala.Option)Unit in class scala.reflect.io.ZipArchive does not have a correspondent in other version
filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.io.ZipArchive.this")

method x()java.lang.Class in class scala.reflect.io.FileZipArchive does not have a correspondent in other version
filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.io.FileZipArchive.x")

method this(java.io.File,scala.Option)Unit in class scala.reflect.io.FileZipArchive does not have a correspondent in other version
filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.reflect.io.FileZipArchive.this")
```

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: My understanding is that it is a JVM implementation detail. I have just asked about the future plans for the encoding of the set of version numbers in the path name: http://mail.openjdk.java.net/pipermail/compiler-dev/2018-March/thread.html#11736

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: This PR should exclude the `-modules` directories for now, we'll need to consider them when we add JPMS support.

I'll revisit this code and revise the comment.

src/compiler/scala/tools/nsc/settings/ScalaSettings.scala
lrytz: Maybe the message should mention both uses: the Java API and the selection from multi-version jars

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: The summary of the discussion so far is that the plan is to encode `10=A`, `11=B`, ...

I'll update this PR along these lines.

We also discussed how we could enhance `javax.tools.StandardJavaFileManager` implementation to serve as a foundation us build the Scala compiler's support for `--release` (and all things class/module path related), insulating us from the details of `ct.sym`. That might subsume this first cut at directly dealing with ct.sym, but I'll target those efforts to 2.13.x.

src/reflect/scala/reflect/internal/JDK9Reflectors.java
retronym: > Also this method could be made private and invoked directly from `newJarFile`.

I'd prefer to keep these methods signature identical to the methods they reflect, rather than mixing concerns of the reflection with facades.

src/reflect/scala/reflect/io/ZipArchive.scala
retronym: No, this is intentional. I'll add a comment.

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: We actually want to fall through to the `None` case on major version < 9, so this as a benign problem. I'll rework this method in any case to exit quickly on older JDKs and to use the new `Runtime.version.major` to get the current JVM major version.

```scala
  def apply(release: Option[String]): Option[ClassPath] = {
    import scala.util.Properties._
    if (!isJavaAtLeast("9")) None
    else {
      // TODO escalate errors once we're sure they are fatal

      val currentMajorVersion: Int = JDK9Reflectors.runtimeVersionMajor(JDK9Reflectors.runtimeVersion()).intValue()
      release match {
        case Some(v) if v.toInt < currentMajorVersion =>
          try {
            Some(new CtSymClassPath(release))
          } catch {
            case _: Throwable => None
          }
        case _ =>
          try {
            val fs = FileSystems.getFileSystem(URI.create("jrt:/"))
            Some(new JrtClassPath(fs))
          } catch {
            case _: ProviderNotFoundException | _: FileSystemNotFoundException => None
          }
      }
    }
  }

```

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: I've moved the check for existence out into the factory that constructs `CtSymClassPath`, but haven't escalated the error yet, as per the comment I'm adding:

```
      // TODO escalate errors once we're sure they are fatal
      // I'm hesitant to do this immediately, because -release will still work for multi-release JARs
      // even if we're running on a JRE or a non OpenJDK JDK where ct.sym is unavailable.
      //
      // Longer term we'd like an official API for this in the JDK
      // Discussion: http://mail.openjdk.java.net/pipermail/compiler-dev/2018-March/thread.html#11738
```

src/compiler/scala/tools/nsc/classpath/DirectoryClassPath.scala
retronym: `= Nil` is probably the best we can do, I'll change to that.

retronym:
Thanks for the detailed review of the slightly messy implementation, @lrytz. I've pushed a commit to address most of your comments. I'll give the patch another round of review and polish tomorrow.

lrytz (approved):
LGTM! I pushed a small cleanup commit, please take a look.

retronym:
All LGTM, thanks for that.

ktoso:
Awesome, LGTM :-)

ktoso (approved):

retronym:
FTR, the OpenJDK team has logged our feature request to support `ct.sym` through the `javax.tools` compiler APIs as [JDK-8199521](https://bugs.openjdk.java.net/browse/JDK-8199521). With that, we could avoid coding against the unspecified format of ct.sym.

ijuma:
Nice, looking forward to 2.12.5. :)

